PR #20, opened 2015-10-25
i18n for manifests: org.jkiss.utils , org.jkiss.wmi

ruspl-afed:
Line endings was converted to LF for MANIFEST.MF, build.properties, bundle.properties


serge-rider:
Let's try..

Issue #146, opened 2016-01-04
Cannot run SELECT .. FOR UPDATE query


basteln3rk:
Because dbeaver automatically adds a limit clause to SQL query, it is not possible to run this query in dbeaver:

```sql
select * from innodb_deadlock_maker where a = 0 for update;
```

Error message:

![screenshot_2016-01-04_11-32-51](https://cloud.githubusercontent.com/assets/1027746/12087888/e80f8d5a-b2d6-11e5-9c7b-a5108d637e5d.png)

In the mysql terminal, this query runs just fine, since no LIMIT clause is appended

(this problem occurs on version 3.5.8)

r-richmond:
Try changing the limit to 0.

![image](https://cloud.githubusercontent.com/assets/9246654/12096324/6e7d97fa-b2c9-11e5-9678-fc4464b18075.png)

codebling:
@r-richmond nice workaround! 

I'm preparing a fix.

codebling:
The fix also works for `LOCK IN SHARE MODE`.

serge-rider:
BTW standard workaround for all problems with explicit LIMIT is to disable query modifications in preferences. Just disable "Use SQL to limit fetch size" option:
![rs-limit-config](https://cloud.githubusercontent.com/assets/6398845/12221576/649ae3fe-b7b1-11e5-86c5-9604093dd88b.png)

serge-rider:
SELECT .. FOR UPDATE issue was fixed.
Other LIMIT problems should be tracked in different issues.

codebling:
@serge-rider Ok if I open a new issue for the issues discussed in the PR?

serge-rider:
Sure, it is always ok to open new issues if they are reproducible and/or make sense.

PR #151, opened 2016-01-06
#146 No limit on queries for update

codebling:


codebling:
Actually I am not sure that this is the right fix for issue #146. 

This fix prevents `LIMIT` from being appended to the query string run by DBeaver, but DBeaver will still run `SET SQL_SELECT_LIMIT=x` with this fix applied. 

This will have unintended side effects for the user if the total number of rows selected by the user's query is greater than the number of rows that DBeaver wants to display (in that all of the rows > than the number of rows that DBeaver wants to display will **not** be locked). 

Probably in the case of a `FOR UPDATE`, the limiting should be disabled in `org.jkiss.dbeaver.model.DBUtils#prepareStatement`. 

What do you think?  I can make this change.

codebling:
There are a few other cases where limiting should be disabled for `SELECT`. 

Here is the complete list: 

* `SELECT...PROCEDURE`
* `SELECT...INTO`
* `SELECT...FOR UPDATE`
* `SELECT...LOCK IN SHARED MODE`

serge-rider:
This seems to a valid fix, thanks!

serge-rider:
Now SELECT INTO, SELECT FOR UPDATE and LOCK IN SHARED MODE are covered by plain select detector.
Could you give me an example of SELECT...PROCEDURE query?

codebling:
Those statement types do seem to be picked up by Jsqlparser, but when testing my fix, I found that (at least for the MySQL plugin) it still generated `SET SQL_SELECT_LIMIT=x` before the query. I think that this still affects these (`...PROCEDURE`, `...INTO`, `...FOR UPDATE`, `...LOCK IN SHARED MODE`) queries, doesn't it?  In that case it would be necessary to bypass the `SET SQL_SELECT_LIMIT=x` statement and I haven't looked into that yet. 

I have not used `SELECT...PROCEDURE` (I have tried but I've been getting an error), but it is supported by both [MySQL](http://dev.mysql.com/doc/refman/5.7/en/select.html) and [Oracle](https://docs.oracle.com/cd/E17952_01/refman-5.0-en/select.html). I haven't been able to confirm its status in the SQL standard. 

MySQL documentation states that "A PROCEDURE clause names a procedure that should process the data in the result set." 

The PROCEDURE clause is supposed to come after LIMIT
```
SELECT
...
    [LIMIT {[offset,] row_count | row_count OFFSET offset}]
    [PROCEDURE procedure_name(argument_list)]
    [INTO OUTFILE 'file_name'
        [CHARACTER SET charset_name]
        export_options
      | INTO DUMPFILE 'file_name'
      | INTO var_name [, var_name]]
    [FOR UPDATE | LOCK IN SHARE MODE]]
```

My understanding is that it should be applicable to any SELECT statement. It will simply take the results of the SELECT and send them to the named PROCEDURE, but I am not sure how this works with the PROCEDURE ins and outs. I will have to do more research.

PR #184, opened 2016-02-02
Update MySQLDataSourceProvider.java

giulio-quaresima:
Apparently the ~/.keystore is the default location for the keystore expected by keytool. And, moreover, D:/ is sooo windows only!!!
I would suggest also to use "changeit" instead of "password", because it seems "changeit" is the de-facto default password used with keytool, see for example https://tomcat.apache.org/tomcat-7.0-doc/ssl-howto.html and https://community.oracle.com/thread/1049953?tstart=0


serge-rider:
Great, thanks! 
I think these Windows-based pieces of code left from some testing. Honestly SSL support wasn't announced as a working feature in MySQL extension.

giulio-quaresima:
Very thanks you.
I did it because yesterday I have tested your wonderful DBeaver to connect to mysql with SSL and the exception was a FileNotFoundException "D:/temp/ssl/test-cert-store". Then I followed [this](http://dbeaver.jkiss.org/forum/viewtopic.php?f=2&t=1401) and it worked for me.

PR #189, opened 2016-02-04
Stored procedures source code for DB2 on Z/OS

ghost:
I suggest to add feature to read store procedures code for DB2 on Z/OS.
The following select, tested and verified, extracts code :

select NAME as SP_NAME, cast(TEXT as varchar(32767)) as SP_SOURCE_CODE
from SYSIBM.SYSROUTINES
where SCHEMA = '_schema_name_'


serge-rider:
Sorry, this is not a correct pull request.
Pull requests should be made against "devel" branch and should contain some valuable code.
If you want to make a feature request then create one on issues tab.

PR #196, opened 2016-02-11
Grammatical correction in result set filter

r-richmond:
Should be "Enter a SQL expression to filter results" not "Enter an SQL expression to filter results"


serge-rider:
Ok, thanks!

PR #227, opened 2016-02-24
do not add 1 to hour when generating display string for time and datetime column

zbelial:
When generating INSERT statement, the HOUR of a time or datetime column is wrong imho, because it is added 1.


serge-rider:
Thanks!

PR #269, opened 2016-03-06
Add Codacy badge to README

codebling:
Don't know if you care -- I did this in a while ago and forgot about it.


serge-rider:
I'm still not sure about Codacy.
I made my own project (https://www.codacy.com/app/serge/dbeaver/) there and receive info letters after each commit. Generally it is convenient but particular errors/warnings are not very useful in most cases (especially code style which is always controversial).
In any case I don't think it should be in readme. It's better to put a link somewhere in dev docs (when they will appear in wiki).

codebling:
ah! You needn't have bothered creating your own, the one I created was under a 'DBeaver' org and I gave you full ownership (iirc). Haha anyways I agree, not all of the warnings are super helpful, are though they are all configurable. Might be useful much later on if there's ever time to put a magnifying glass to each line of code.

Issue #320, opened 2016-03-30
Oracle source problem for objects names with quotes


lluccia:
When looking at some Oracle procedure codes from the edit screen, in some cases the code was being truncated, as the following example:

**expected SQL:**
```sql
CREATE OR REPLACE PROCEDURE "SCHEMA"."PROC_NAME" (param1 in  number						
                                  ,param2 out number )						
is						
	v_an_procedure          varchar2(30)   := 'PROC_NAME';					
						
End;
```
**actual SQL:**
```sql
CREATE OR REPLACE procedure "SCHEMA".varchar2(30)   := 'PROC_NAME';					
						
End;
```

The problem occurs when the procedure name is enclosed in quotes.
It truncates part of the code if there is a word "procedure" in the rest of the code, as the variable name in the example. Otherwise, the CREATE OR REPLACE command is not inserted.

The method causing the problem is ```org.jkiss.dbeaver.ext.oracle.model.OracleUtils.insertCreateReplace() ```that is using a regular expression to match the procedure declaration for later including "CREATE OR REPLACE" and schema name:
```java
    public static String insertCreateReplace(OracleSourceObject object, boolean body, String source) {
        String sourceType = object.getSourceType().name();
        if (body) {
            sourceType += " BODY";
        }
        Pattern srcPattern = Pattern.compile("(" + sourceType + ")\\s+(\\w+)", Pattern.CASE_INSENSITIVE);
        Matcher matcher = srcPattern.matcher(source);
        if (matcher.find()) {
            return
                "CREATE OR REPLACE " + matcher.group(1) + " " +
                DBUtils.getQuotedIdentifier(object.getSchema()) + "." + matcher.group(2) +
                source.substring(matcher.end());
        }
        return source;
    }
```

I fixed this problem changing the regular expression to include optional quotes and matches from the beginning of text:
```java
Pattern srcPattern = Pattern.compile("^(" + sourceType + ")\\s+(\"{0,1}\\w+\"{0,1})", Pattern.CASE_INSENSITIVE);
```

PR #321, opened 2016-03-30
#320 fixes CREATE OR REPLACE text insertion on Oracle object sources

lluccia:
…declaration when procedure name is enclosed with "


serge-rider:
Thanks!

Issue #435, opened 2016-05-11
Implement fuzzy search


ihabunek:
Here's a wish: Could you implement sublimetext-like fuzzy search when giving column/table/etc auto-completion suggestions?

For example, in a WERE clause,  when i type "pid" and pres Ctrl+Space, the popup window should contain **p**erson_**id** and image_**p**rofile_**id** since they match (fuzzily).

I feel this would be a great addition.

serge-rider:
Good point, I like it.

carlosjrcabello:
It's really nice! 

It seems with auto complete implemented in Eclipse Mars.

![screen207](https://cloud.githubusercontent.com/assets/637767/15178974/4f6f654c-174f-11e6-9fd5-c9d0f7a9342e.png)

PR #466, opened 2016-05-21
Match any where proposal

tarcisiojr:
New feature for SQL completion proposal. The user can select if he wants look for suggestions basead on word parts that appear in anywhere on proposal template strings.


serge-rider:
Isn't it the same as fuzzy search (#435) ? It will be added in 3.6.8.

tarcisiojr:
Yes, it is the same feature. I going to use my implementation until you release the 3.6.8 version.

Thanks.

serge-rider:
Ok then

Issue #403, opened 2016-04-25
Array type column edit support


zinking:
should support editing array data types through the data grid. 

some of my attempts: 
https://github.com/serge-rider/dbeaver/compare/devel...zinking:arrayedit?expand=1

can:edit existing array elements
cannot:add/remove new array elements
cannot:support cancel the changes.

loskultos:
I second this. I have a column type text array. can't edit exisiting entries nor add or remove elements

zinking:
@serge-rider you can explain a bit which direction you want this feature to be? I can try to help.
like:
- maybe should use the Table for UI
- maybe should separate a different value controller etc..

serge-rider:
It is complicated.
Complex type editor uses table for UI but it has a few disadvantages. And it doesn't work if arrays item is a complex type too. Main idea is to integrate array editor in the Grid editor. But it requires serious effort and much time (which I don't have at the moment).
But Table editor may work as a cheap (and temporary) workaround for simple types (strings, numbers, etc) editor.
I'll return to this issue after 3.6.7 release.

If you have any UI ideas please tell.

serge-rider:
Some basic arrays edit support has been added. It works with PostgreSQL arrays, at least.
It'd be great is somebody would test it with other databases before version release.

zinking:
quickly checked Phoenix HBase, array still in read only mode. since the support is already added, I assume it's straight forward to adapt to phoenix HBase, will see what could be done. 

thanks @serge-rider

serge-rider:
Hm,
can you use add/remove element actions?

![array_edit](https://cloud.githubusercontent.com/assets/6398845/15646797/77467dd0-2668-11e6-9acc-192d42b05ce1.png)

zinking:
@serge-rider  I can edit/add/remove array types in phoenix/HBASE.  with the PR amended above

with one boundary case, when I removed all array elements it behaves correctly to set the value to `null`. but on top of null, can't add element to a null array value. or do other edit actions.

!ENTRY org.jkiss.dbeaver.model 4 0 2016-06-01 14:45:34.377
!MESSAGE Can't find children items for add
2016-06-01 14:45:39.785 - Can't find children items for add

anatoliykmetyuk:
Unable to use the array editor with H2 database. When I press the "+" button to add a new entry, nothing happens.

curala70:
For PostgreSQL the array editor does not work properly.
[...]

PR #498, opened 2016-06-01
enable phoenix/habse array edit

zinking:
#403


serge-rider:
Look reasonable, 10x.
Maybe I'll make it default implementation for array value handler.

Issue #759, opened 2016-09-23
App icon and name don't display properly on macOS


g4brielvs:
Remove black background from app icon and fix app name renaming Dbeaver.app to DBeaver.app

It displays as follows on 3.7.4:

<img width="100" alt="screenshot 2016-09-23 00 07 23" src="https://cloud.githubusercontent.com/assets/4053416/18774386/e2b9914e-812e-11e6-8030-8fcb45dd3516.png">

Any ideas how I can be of any assistance?

/cc @serge-rider

serge-rider:
For now I don't have Mac for UI testings. This leads to problems with Mac icons (I made ICNS from set of PNG images using one of online icon generators).
If you will create good-looking ICNS icon for DBeaver it will be great.

ICNS must contain icons with following resolution (at least): 16x16, 32x32, 48x48, 64x64 and 256x256.
Source images (PNG) are here: https://github.com/serge-rider/dbeaver/tree/devel/plugins/org.jkiss.dbeaver.core.application/icons

All of them have transparent background. But for some reason it was lost during translation (you can find current dbeaver.icns in DBeaver installation). I believe that native Mac tool should make it better.

Thanks in advance.

josemodena:
I've created PR #842 to fix this issue.

serge-rider:
Great, thank you!
I can't test it on different Mac versions but generally new icns looks fine.
Closing.

PR #842, opened 2016-10-13
Fixed icon background on macOS

josemodena:
Serge, first of all thank you for the great app.

To address issue #759, I created a new dbeaver.icns and added new source icons based on the original ones.

I tested the new dbeaver.icns in my local copy of dbeaver and the issue is solved.


serge-rider:
Great, thank you!

PR #845, opened 2016-10-14
New Database Object Connection and Privilege Checks added

Sargul:
Hi,

from the feedback of my Beta Testers I made these changes to exasol:

*) new Object Type Connections is now supported
*) added "visibleif" checks to verify the user has sufficient privileges to access the meta tables for users, connections and roles objects.
*) Reformat source using dbeaver formatter

Kind Regards

Karl


serge-rider:
Looks good!

PR #901, opened 2016-10-27
Increase Performance for Exasol Security Object Views

Sargul:
Implemented Security Objects as Caches on Datasource Level to increase performance in an Exasol databases with many objects


serge-rider:
Looks good.
note that next release will be on Monday

PR #941, opened 2016-11-03
Fix Apache Drill sample URL

laurentgo:
The current sample URL for Apache Drill has port after schema.
According to http://drill.apache.org/docs/using-the-jdbc-driver/,
port is part of the zookeeper host name.


serge-rider:
Thanks!

PR #943, opened 2016-11-05
Exasol: Support for High Speed CSV Export/Import

Sargul:
- Added Support for High Speed Exasol CSV Import/Export
- Cleaned up unnecessary Imports
- removed Execute Keyword from Dialect (Exasol does not support PrepareCall
- Generate DDL for Schema


Sargul:
error in CI

PR #1077, opened 2016-12-17
Performance Optimization and Cleanups

Sargul:
- Heavily increase MetaData queries for Tables by setting read on access properties to expensive
- Fixed missing close cursor statement
- Enable edit of description on schema and table objects
- bug in meta data query for views


serge-rider:
Cool, thanks.
Note: when you declare "expensive" properties you need to add extra parameter (DBRProgressMonitor monitor) in property getter. Do not use VoidProgressMonitor.INSTANCE because operation will stuck forever if anything will go wrong. Real progress monitor allows to cancel most of IO or JDBC operations.

serge-rider:
I've just fixed ExasolTable expensive properties. This was easy, in some other cases obtaining real progress monitor is tricky. Anyhow VoidProgressMonitor should be used only if you 100% sure that operation won't perform remote database request.

PR #1078, opened 2016-12-18
Add maven repository for Exasol JDBC Driver

Sargul:
Exasol JDBC drivers are now available in a maven repository added url in plugin.xml

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.exasol/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.exasol/plugin.xml
@@ -321,20 +321,16 @@
                         sampleURL="jdbc:exa:{host}[:{port}][;schema={database}]"
                         defaultPort="8563"
                         description="Exasol JDBC driver"
-                        webURL="https://www.exasol.com/portal/display/DOWNLOAD/5.0">
+                        webURL="https://maven.exasol.com/artifactory/exasol-releases/com/exasol/exasol-jdbc/">
 
                     <replace provider="generic" driver="exasol"/>
+                    <file
+                          path="https://maven.exasol.com/artifactory/exasol-releases/com/exasol/exasol-jdbc/5.0.17/exasol-jdbc-5.0.17.jar"
+                          type="jar">
+                    </file>
                     <property name="querytimeout" value="600"/>
                     <property name="clientname" value="DBeaver"/>
                     <property name="connecttimeout" value="40"/>
-                    <property name="@dbeaver-default-dataformat.type.timestamp.pattern"
-                              value="yyyy-MM-dd-HH.mm.ss.ffffff"/>
-                    <library path="C:\Program Files (x86)\EXASOL\EXASolution-5.0\JDBC\exajdbc.jar"/>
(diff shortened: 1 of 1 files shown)


serge-rider:
Is there any reason to add direct link on jar file instead of maven artifact reference?
Usually artifact is better because users can update to the recent driver version without DBeaver upgrade.

serge-rider:
Compare with this patch:
```
===================================================================
--- ../plugins/org.jkiss.dbeaver.ext.exasol/plugin.xml	(revision 90211f5c297aca10e7b574cb61c09d70c5f6878d)
+++ ../plugins/org.jkiss.dbeaver.ext.exasol/plugin.xml	(revision )
@@ -327,14 +327,7 @@
                     <property name="querytimeout" value="600"/>
                     <property name="clientname" value="DBeaver"/>
                     <property name="connecttimeout" value="40"/>
-                    <property name="@dbeaver-default-dataformat.type.timestamp.pattern"
-                              value="yyyy-MM-dd-HH.mm.ss.ffffff"/>
-                    <library path="C:\Program Files (x86)\EXASOL\EXASolution-5.0\JDBC\exajdbc.jar"/>
-                    <fileSource url="https://www.exasol.com/portal/display/DOWNLOAD/5.0"
-                                name="Exasol JDBC Drivers"
-                                instruction="Download Exasol JDBC (JCC) client driver. Extract downloaded archive and copy exajdbc.jar in some directory. Then add exajdbc.jar as a library in DBeaver driver editor dialog.">
-                        <file name="exajdbc.jar" description="JDBC driver"/>
-                    </fileSource>
+                    <file type="jar" path="maven:/com.exasol:exasol-jdbc:RELEASE"/>
                 </driver>
             </drivers>
             <views>
@@ -440,6 +433,10 @@
             </objectType>
          </tool>
       </tools>
+    </extension>
+
+    <extension point="org.jkiss.dbeaver.mavenRepository">
+        <repository id="exasol-repo" name="Exasol Repository" url="https://maven.exasol.com/artifactory/exasol-releases/"/>
     </extension>
 
 </plugin>
```

Sargul:
Hi Serge,

the problem is that there are some bugs in Version 6 of the jdbc driver. V6 is the RELEASE Version.

I tried adding the exasol maven repository to my plugin
`  <repositories>
        <repository>
            <id>maven.exasol.com</id>
            <url>https://maven.exasol.com/artifactory/exasol-releases</url>
            <snapshots>
                <enabled>false</enabled>
            </snapshots>
        </repository>  
  </repositories>`

and use `maven:/com.exasol:exasol-jdbc:5.0-SNAPSHOT` as path. However this does not seem to work.

So I hard coded to 5.0.17.

Regards

Karl

serge-rider:
I see. There is similar problem with MySQL drivers.
In that case you can use regex pattern to match only good versions:
`<file type="jar" path="maven:/com.exasol:exasol-jdbc:{.*5\.0\..*}"/>`

serge-rider:
BTW I mean dynamic driver configuration in DBeaver (for samples see generic/postgresql/mysql drivers config in plugin.xml) - not hard dependency in pom.xml.

PR #1144, opened 2017-01-12
Bugfix for Required Column Attribute and German Translation

Sargul:
- bug in display of required for Exasol
- show DBeaver and Version in Connection Attributes in Exasol
- Workaround for different handling of result set and table meta data in Exasol JDBC Driver V6
- german translation for dbeaver.core and exasol (more to come)


serge-rider:
A few notes:
1. There is a bug in plugins/org.jkiss.dbeaver.lang/.classpath diff. You can just remove org.jkiss.dbeaver.lang plugin from build, it is not used.
2. To provide client information it is better to use another approach. As I understood Exasol uses special connection properties (clientname & clientversion) to identify client. Oracle does the same. Check OracleDataSource.getInternalConnectionProperties method. This is a standard way to identify client while patching connection URL doesn't look very good.
Additionally DBUtils.getClientApplicationName allows to redefine client name in preferences (in some cases users want to mimic some other database client while working in DBeaver).

Sargul:
add provide client information:

with the exasol jdbc driver this can only be done in the url.  The properties clientname and clientversion can only be set in the url.

I've verified this with Exasol Support. They plan to allow setting these parameters as properties in the future.

serge-rider:
Ok, I see.
I'll pull this patch. Thanks!

I'd still recommend to improve client application name obtain (see DBUtils.getClientApplicationName source). 
If client name is overridden in preferences then you need to use it instead of real DBeaver product name.

Sargul:
OK Serge,

I will change the application name asap. 

As the DBUtils.getClientApplicationName always retrieves the name including version number I will have to fill the clientversion field with some other strings. Otherwise "Unkown" will be show in the session information.

Regards

Charly

PR #1154, opened 2017-01-17
Quote schema query to prevent an IllegalArgumentException.

EusthEnoptEron:
Using a schema name that contains a dollar symbol ($) would throw a `java.lang.IllegalArgumentException: Illegal group reference`.

[String#replaceFirst()](http://docs.oracle.com/javase/6/docs/api/java/lang/String.html#replaceFirst%28java.lang.String,%20java.lang.String%29):
> Note that backslashes (\) and dollar signs ($) in the replacement string may cause the results to be different than if it were being treated as a literal replacement string; see Matcher.replaceFirst(java.lang.String). Use Matcher.quoteReplacement(java.lang.String) to suppress the special meaning of these characters, if desired.


serge-rider:
Thanks for fix!

PR #1395, opened 2017-03-05
Switch from PreparedStatement to Statement with literals

Sargul:
Switch to Statement with literals for queries of the system objects on Exasol. 
PreparedStatements on System Objects are incredibly slow (see  [EXASOL-1948](https://www.exasol.com/support/browse/EXASOL-1948) for details if you have an Exasol Support account).


serge-rider:
Good.
BTW I've made some cleanup of statements create functions.

serge-rider:
Note: you can use DBCSession.prepareStatement(...) with type DBCStatementType.SCRIPT - it will call createStatement() + setQueryString. No need to make this manually in every execution.

(I think I'll rename DBCSession.prepareStatement to something else - it is ambiguous name because in fact it may produce prepared, callable or plain statement).

PR #1427, opened 2017-03-13
Mysql dump advance option support

yiitz:
add advanced mysql dump options for expert user.
for example,some one only want to dump data structure but not data,the need to give a "--no-data" option


serge-rider:
Yep, this makes sense.
Thanks

PR #1530, opened 2017-04-05
Lock viewer.

AndrewKhitrin:
It looks like a table in "Administer" node with session locks and lock details, also it have the lock dependency graph for sessions waiting and holding locks with the possibility of killing the selected session.

Now works for Oracle <11g and PostgreSQL <9.x. 

A complicated graph for testing can be generated with attached java utils - dont forget to setup your db host\port\user directly in code (sorry), additionally table "USR" with pack of test data will be created.

[LockGenerator.java.txt](https://github.com/serge-rider/dbeaver/files/900335/LockGenerator.java.txt)
[LockGeneratorOracle.java.txt](https://github.com/serge-rider/dbeaver/files/900338/LockGeneratorOracle.java.txt)

-----------------------------------

Please review and send me comments.

Regards,
Andrew


serge-rider:
Wow, great job!

I like the code, looks like DBeaver model/UI is clear for you.
Comments:
- Could you extract org.jkiss.dbeaver.ui.views.lock* into separate plugin? Let's say org.jkiss.dbeaver.ext.ui.locks. This will make you a plugin owner, will simplify its support and will remove gef/draw2d dependency from core module. You can do this similarly to erd plugin. Or, if you wish I can do this by myself. Note: additions to model, oracle and postgresql plugins should remain where they are. I am not a fan of tons of tiny plugins.
- Add license/copyright header in all *.java files. It presents in some of them but missed in others.
- Add yourself in copyright section (see sample in files like DB2Schema.java and other which were contributed by other people)
- Write me on serge@jkiss org - I will add you in the contributor list on dbeaver.jkiss.org web site.

And I have one concern. Although DBeaver standalone now requires Java 8 (because of Eclipse Neon dependency) its sources are still in Java 7. This makes sense because Eclipse Plugin version can be installed in Eclipse prior to Neon (which depends on Java 7 or 6). Maybe it is a time to move to Java 8 world completely but I'm still not sure about that. There are a lot of people who stuck in early 10s.

BR,
Serge

AndrewKhitrin:
Serge,

Many thanks for the review. Thanks for your feedback. Dbeaver is very important project for me, I have a lot of plans to create the functionality I need, and I hope it can be usefull for all.

Please feel free to criticize my code and give recommendations.

I push the following changes into the request branch:

-  Corrected copyright message and contributor's name
-  Lock viewer placed in separate plugin org.jkiss.dbeaver.ext.ui.locks, but now Oracle and Postgres plugin depends on ext.ui.locks, please correct it if you vision differ from mine


About Java 6, I dont think its possible to build DBeaver on JDK6 even at first glance there is a lot of JDK7 construction (diamond syntax for example) so I think JDK7 is reality now.

We ought to pull out a lot of people from early 10s ! **:)**

_Thanks,
Andrew_

serge-rider:
Andrew,

Yes, postgresql and oracle will depend on this new plugin. Don't forget to include it in the root pom file and in the core feature (just like erd plugin).
Yes, Java6 is not an option (way too old).
I will make merge today evening.

Thanks

AndrewKhitrin:
Done.

Add all files (root pom.xml plugin.xml and so). Now build with maven and tycho without error.

Sorry for many commits.

_Andrew._

PR #1626, opened 2017-05-03
Lock Viewer for Exasol

Sargul:
* Lock Viewer for Exasol

I've changed the timeout in the ObjectListControl from 4 seconds to 12 seconds as lock queries in exasol may take relativly long.


Sargul:
Hi  Serge, 

I've changed the timeout in the ObjectListControl to 12 seconds (for exasol lock viewer).

That's why I won't just merge the pull request. Could you have a look?

Regards

Karl

serge-rider:
I've made this timeout configurable (in fact it is refresh timeout and it is not used in regular data load)

PR #1914, opened 2017-07-25
Update plugin.xml

ghost:
Hi there, 

I have edited some of configuration. Would it be updated soon?


serge-rider:
Hi,

it will be updated in the next version (hopefully after 2 weeks)

PR #1973, opened 2017-08-06
Inserts instead of COPY option

gamikun:
I added an option to backup PostgreSQL with SQL INSERT instead of the default COPY format.

It implements the --inserts flag in pg_dump.
PostgreSQL documentation: https://www.postgresql.org/docs/9.1/static/app-pgdump.html


serge-rider:
Good enhancement, thanks!

PR #2127, opened 2017-09-08
add clean option to Postgres restore dialog

twooton61:
Would be nice to have the option to add the `-c` option for `pg_restore` so the command will "Clean (drop) database objects before recreating them". Otherwise a backup and restore will throw a lot of errors regarding duplicate keys and tables


serge-rider:
Good point, thanks

PR #2326, opened 2017-10-13
Complete org.jkiss.dbeaver.core/OSGI-INF/l10n/bundle_zh.properties

liuyuanyuan:
Complete Chinese Localization of org.jkiss.dbeaver.core/OSGI-INF/l10n/bundle_zh.properties.


serge-rider:
Great, thanks!

PR #2345, opened 2017-10-16
Update plugin's i18n and l10n for  org.jkiss.dbeaver.core.application

liuyuanyuan:
Complete i18n for plugin.xml and l10n for bundle.properties, bundle_zh.properties of  org.jkiss.dbeaver.core.application.


liuyuanyuan:
Because this request has too many conflicts, I will pull request next time more clear.

PR #2367, opened 2017-10-19
update org.jkiss.dbeaver.ext.generic/plugin_zh.properties

liuyuanyuan:
Complete l10n for update org.jkiss.dbeaver.ext.generic/plugin_zh.properties


serge-rider:
Thanks!

PR #2377, opened 2017-10-20
Update CoreResources_zh.properties

liuyuanyuan:
Update CoreResources_zh.properties to add bundle about:
Menu/Toolbar/SQL Editor context menu item, 
New Connection Wizard ,
Preference/Properties Dialog.

(By the way, I always add bundles and never delete from the original file.
The compare results is always caused by  the format, like space beside = .)


serge-rider:
Umm, something went wrong. Looks like this PR conflicts with the first PR for CoreResources l10n.

liuyuanyuan:
Yes,  indeed so.  I have resolved the conflicts now  : )

PR #2418, opened 2017-10-27
Dbeaver 2138, hexView size based on 16 bytes, remind part based on the 8, 4, 2.

serjiokov:


serge-rider:
I see some customization in code but it is not clear how to activate this in UI. At the moment Hex editor uses the same old behavior (16 bytes align).
And there is no preference option or some adaptive behavior...

PR #2432, opened 2017-10-30
Clean non-working code from plugin.xml of org.jkiss.dbeaver.core

liuyuanyuan:
Clean non-working code from plugin.xml of org.jkiss.dbeaver.core.


liuyuanyuan:
This pull request include:
some more l10n for preference in class file and properties files ,
some copyright modification in  l10n properties files except org.jkiss,dbeaver.core ,
and some l10n properties file supplement.
Please konw this~

serge-rider:
Generally looks good.
However most of preference page classes (e.g. PostgreSSLConfigurator.java) looks like they were completely replaced by new versions - although there are just a few changes inside. It is very hard to understand what was changed exactly.
I wondering why we have so weird diff.. Do you do commits directly from Eclipse or manually from command line?

liuyuanyuan:
I commit from Eclipse generally, but recently failed several times caused by network, so I have to push them later in command line...Feel sorry.

liuyuanyuan:
I also want to know why some file has weird diff,  your words remind me , I'll have some test to confirm this problem.

ruspl-afed:
The diff may be huge because of line endings, you can enable "ignore whitespace" mode with "w=1" flag like
https://github.com/serge-rider/dbeaver/pull/2432/files?w=1

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageAbstract.java
```diff
@@ -104,7 +104,7 @@ protected void createDriverPanel(Composite parent) {
 
         if (!site.isNew() && !site.getDriver().isEmbedded()) {
             Link netConfigLink = new Link(panel, SWT.NONE);
-            netConfigLink.setText("<a>Network settings (SSH, SSL, Proxy, ...)</a>");
+            netConfigLink.setText("<a>" + CoreMessages.dialog_connection_edit_wizard_conn_conf_network_link + "</a>");
```
ruspl-afed: I'm not sure regarding this approach, for me the whole text (including "<a>" and "</a>") should go to resources

liuyuanyuan: Of course, I know what you say, but here the code <a> </a> is used for a link,  so I just translate the text.

liuyuanyuan:
Just now I commited in eclipse and I saw the diff is normal.

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageAbstract.java
ruspl-afed: If you have any concern please refer to platform sources
'
package org.eclipse.debug.internal.ui.launchConfigurations;
...
public class SelectLaunchersDialog extends AbstractDebugListSelectionDialog {
...
	protected void addCustomHeaderControls(Composite parent) {
...
		Link link = new Link(comp, SWT.WRAP);
		link.setText(LaunchConfigurationsMessages.SelectLaunchersDialog_4);
...
'
and LaunchConfigurationsMessages.properties contains
'
SelectLaunchersDialog_4=<a href="ws">Change Workspace Settings...</a>
'

serge-rider:
Looks good,
thanks guys :)

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageAbstract.java
liuyuanyuan: Thanks for your example, then I know the standard practice.
Also thanks for your solution to the diff, I really learned.

PR #2489, opened 2017-11-08
Add i18n and l10n for driver and SQL editor result set filter panel

liuyuanyuan:
Only add i18n and l10n for driver and SQL editor result set filter panel（files changed are clear this time）.
Note: I removed unused imports from ResultSetFilterPanel.java .


serge-rider:
Looks good.
Thanks!

PR #2498, opened 2017-11-09
Add main eclipse nl fragments to nls.feature and  move all l10n properties file to core.nls

liuyuanyuan:
Add main eclipse nl fragments to org.jkiss.dbeaver.runtime.nls.feature for eclipse standard parts,
and  move all l10n properties file from org.jkiss.dbeaver.core to org.jkiss.dbeaver.core.nls .


ruspl-afed (changes requested):
Default *.properties resource bundles should be kept inside the host bundle
1) bundle.properties should be kept on original location
2) CoreResources.properties should be kept on original location

The eclipse platform nl fragments should be a part of another feature that we didn't introduced yet, now the fragments you refernced are not available during build
3) please revert the changes for feature.xml

liuyuanyuan:
I'll move default *.properties resource bundles back and restore feature.xml.

But Eclipse platform nl fragments just need to install eclipse babel, their host plug-ins are already in DBeaver :)

ruspl-afed (approved):

ruspl-afed:
Thanks for contribution!
You are right regarding Eclipse nl fragments - the host part is already inside DBeaver. But we need to tune our build scripts before we can include Babel fragments. This why I asked you to revert feature.xml for now.

liuyuanyuan:
OK, ruspl-afed~ Thanks for your reply~
When you include Babel, please inform me and I'll add the fragments then ^_^

Issue #2502, opened 2017-11-10
Provide NLS fragments and NLS features for DBeaver bundles and features


ruspl-afed:
The should be NLS fragment and NLS feature for each DBeaver bundle and [feature](https://github.com/serge-rider/dbeaver/wiki/Features-structure) to contain localization resources.

ruspl-afed:
#2506 merged

PR #2517, opened 2017-11-14
#2502 remove empty /src folders for NLS fragments

ruspl-afed:


liuyuanyuan:
Hi, ruspl-afed~ empty src for org.jkiss.dbeaver.ext.sample.database.nls still not removed, please check.

ruspl-afed:
Just checked https://github.com/serge-rider/dbeaver/tree/devel/nls/bundles/org.jkiss.dbeaver.ext.sample.database.nls
1) there is no reference to /src in .classpath
2) there is no reference to /src in build.properties
You may still have /src locally as git is file-based and doesn't process empty folders

Issue #2213, opened 2017-09-26
Use environment variables in connection settings


rmerren:
I am very happy with DBeaver...thanks for all your work on it.

I was wondering if it was possible (or could be made possible) to use environment variables in my connection settings. For example, my username and password on a system are stored in an environment variable for various tasks I undertake. When I have to change that password, I need to change it in all of the DB entries every time. I would prefer to just change it in the environment variable and let DBeaver read the variable (e.g. $MY_SYSTEM_PASSWORD). 

Alternatively, it would be convenient to store user credentials securely within the app and refer to them from multiple connections so that I would only have to change the password in one place.

Either way, having a single place to change my password every few months which would affect 12 different connections would be great!

serge-rider:
I think this could be a good feature.
Although I don't think that having user password in env variable is a secure way to keep passwords.
But generally it would be helpful to use env variables in different connection properties.

ruspl-afed:
We need to utilize org.eclipse.core.variables functionality, ${env_var:VARIABLE_NAME} should work for requested case

IgorNovozhilov:
I think, the use of environment variables for storing passwords is not safe. It is better to use files of profiles for users, which implement the storage of shared variables and their use in the connection settings as templates, https://github.com/serge-rider/dbeaver/issues/2092
Also, the profile files need to be encrypted.

For example:
![2017-09-27_09h12_58](https://user-images.githubusercontent.com/3082812/30898453-11e9adb6-a364-11e7-9d61-b3b736037270.png)

serge-rider:
![image](https://user-images.githubusercontent.com/6398845/33289923-2140d066-d3d2-11e7-848f-88eba33169af.png)

dlsxd:
Hi, what about SSH credentials, seems they are not supported for now
![image](https://user-images.githubusercontent.com/5175552/35148683-311f4938-fd24-11e7-89a6-9bcad3ab91c9.png)

efradkin:
@dlsxd you are right - they aren't supported for now. But this feature will be included into the next DBeaver version (ssh host, port, user name and password).
Thank you for your participation!

glarkou:
@serge-rider @dlsxd @efradkin do you know if it will be available here as well?

Hammond95:
Is there a way to set password using a before connection script?

PR #2559, opened 2017-11-21
#2213-env-vars-in-connection-settings

efradkin:
Environment variables are substituted in connection settings (url, user name and password).
The info label is added only on generic and PostgreSQL connection settings pages (because I have such the datasources now).


serge-rider:
Looks good. The only concern - getActualConnectionConfiguration() method becomes too expensive (as it copies many objects including 3 hashmaps) while it is called very often in some cases.
Maybe we could cache this somehow..

PR #2581, opened 2017-11-30
create navicat import connection

juliardi:
Add import connection feature from Navicat Connection Export (NCX) file


ruspl-afed:
Please use https://github.com/serge-rider/dbeaver/blob/devel/docs/codestyle/eclipse-formatter-profile.xml to format the sources.

juliardi:
okay, I will update it when I have access to my computer, maybe in the next few days

serge-rider (commented):
Looks good

PR #2582, opened 2017-11-30
Pg10 partitions

AndrewKhitrin:
Not ready yet. Need to fix DDL for child and partitions and add some props.


ruspl-afed (approved):
The only concern is JDBCPreparedStatementCachedImpl , we need to resolve it later

serge-rider (commented):
Great job

PR #2617, opened 2017-12-08
add password decryption feature on navicat connection import

juliardi:
Password descryption feature on navicat connection import


serge-rider:
Good job!

Issue #1808, opened 2017-06-23
Generate report from selected rows/columns of the Resutlset


geuss:
Hi,
this is only a new feature request.... Could be very useful generate a HTML table (in clipboard) of the selected rows/fields selected in the Result View.

Thanks, and good work!!!!

efradkin:
![image](https://user-images.githubusercontent.com/16664546/34364960-38c65946-ea9c-11e7-8ea4-1e266dc23ed3.png)
Now you can set the wizard to export the selected columns/rows only.

PR #2653, opened 2017-12-16
#1808 Export selected rows only

efradkin:


serge-rider:
1. I don't think that changing DBSDataContainer interface is a good idea. It has a dozen of implementations (note: check EE sources) and we can make all of them take care about row numbers
2. We can't fetch ALL rows and export a few of them (selected ones). Because table may have millions of rows, it will take too much time

What I think we should do:
Make new implementation of DBSDataContainer which will wrap ResultSetViewer. It won't do actual server-side selects - but it will operate with already fetched data (by using ResultSetModel).
Besides that it may filter fetched data in any fashion. 
Also this new option (export only selected rows) should be available only for this data producer.

serge-rider (changes requested):
Refactoring is needed

PR #2655, opened 2017-12-18
2556 postgres debug repackage

ruspl-afed:


serge-rider (approved):
Let's merge it.
Although we still have too much stuff in postgresql.debug.*.

PR #2687, opened 2017-12-26
Update l10n for ext.postgresql

liuyuanyuan:
Update l10n for ext.postgresql.
(replace procedure with function, PG only has function.)


serge-rider:
Thanks Yuanyuan, great as always!

The only concern - why to rename PG procedures to functions? Of course they are mostly functions but generally they are `stored procedures`.

liuyuanyuan:
I thought Function seems clear before , but it need lots of work to replace, I will restore it soon.

PR #2721, opened 2018-01-03
Date Time editor modification

serjiokov:
Was added new widget editor for date, time, timestamp data types.


serge-rider:
Visible issues:
- On Windows initial timestamp controls layout is still broken (day of month is not visible)
- Edit control is too wide (gets over the screen edge if timestamp column is the last one). Perhaps time editor should be under date editor
- LONG date format isn't convenient (it is too different from value in grid). MEDIUM is more suitable.
- Panel editor shouldn't fill vertically (it looks ugly in value view panel).
- We still need to support plain-text editor (as it is done now). Many people prefer this editor over fancy UI.

serge-rider:
Also: no need to declare formatter constants in CustomTimeEditor. They are already declared in DBDDataFormatter.

PR #2731, opened 2018-01-04
"Export resultset as Json" use column label instead of column name.

beonit:
Fixed column label specified as "AS" to be output to json export.
Json export use column label instead of column name.

Query : SELECT CTGRY_ID AS categoryId FROM TSPC_CTGRY LIMIT 2;
Old result : [{"CTGRY_ID" : 1001}, {"CTGRY_ID" : 1002}]
Change result : [{"categoryId" : 1001}, {"categoryId" : 1002}]
fixed : [{"categoryId" : 1001}, {"categoryId" : 1002}]


ruspl-afed:
Let's do it configurable
You can add to the plugin.xml near the 
            <processor
                id="stream.json"

something like                     

<property id="useColumnLabel" label="%dataTransfer.processor.json.property.useColumnLabel.label" type="boolean" defaultValue="true"/>

And then use its valye during JSON export

serge-rider:
I'm not sure about config option. Using label is the default behavior of all exporters.
However I'll add extra check for empty label.

serge-rider:
@beonit Thanks!

PR #2747, opened 2018-01-07
2556 postgres generics

ruspl-afed:


serge-rider (approved):
Good.

serge-rider:
We'll need to fix code style. 5+ imports of the same package should be transformed into star (*).
I know that this is not safe in some cases. But it simplifies life.

Issue #2254, opened 2017-10-02
Date Picker for Date Field


samyem:
Can the grid data editor show date picker when editing date fields similar to what some other similar tools do like:
![image](https://user-images.githubusercontent.com/323450/31139768-92db976e-a840-11e7-94cc-04f467be2dc6.png)

serge-rider:
DBeaver had date/time picker a time ago. Eventually we replaced it with text string editor because of multiple requests.
We could try to combine both plain text editor and datetime picker.

YassienW:
Which settings is used to enable the date picker?

weeman1337:
Does anyone know how this works? For me no picker appears.

pdimh:
It seems the last couple commits of this issue never got merged. The PR got merged before them.

cbush06:
Check `Activate advanced datetime editor` under the `Miscellaneous` section here: 
![image](https://user-images.githubusercontent.com/15720146/210601746-d99850b2-0862-4dc8-9d5c-a4b61208bc8d.png)

PR #2766, opened 2018-01-09
#2254 DateTime Editor Support

serjiokov:
Were added new editor for DateTime Format data, integrated preferences
selection between  old text editor and fancy DateTime editor.


serge-rider:
This is much better but still:
- Tab key not working (after switching to time editor it stucks somewhere). It should work like with other inline editors.
- In panel DateTime editor is stretched and looks ugly
- In date editor day of month is still not visible. I'm not sure that it is possible to solve this easily but on Windows it makes this control very hard to use.

See screenshot:
image

serge-rider:
Looks like whole date editor control is shifted to the left for 10-20 pixels. Perhaps some known SWT issue..

PR #2787, opened 2018-01-14
 Fix error in plugin.xml of ext.mssql . 

liuyuanyuan:


serge-rider:
Missed that one, thanks!

Issue #2619, opened 2017-12-10
Suggested Improvement - Simplify resultset filtering


0biWanKenobi:
I'd like to suggest a simpler, quicker way to filter a resultset by column(s) value(s). Currently the user has to right click a column, and choose Order\Filter, or use F11, while I propose a filter icon (permanent or on hover) as a quicker way to access filters. Also, I think directly listing all possible values (as you can do with the IN.. filter) would make filtering much quicker. The screenshot below is taken from dbForge: their strategy is to tuck the "advanced" filters in a separate popup, while giving immediate acces to null, not null, and a list of distinct values. I've been able to test that solution for a while, it' really effective.

![image](https://user-images.githubusercontent.com/4061104/33800821-11a06268-dd49-11e7-8a2c-473feb53667b.png)

serge-rider:
This makes sense. But we'll need to think a bit more about UI consistency. Column header is already divided on 3 zones (draggable icon, selectable title and sortable icon). Adding one more might be a bit confusing.

BTW Inline combo with list of possible (existing) values is also useful for cell editor.

0biWanKenobi:
Yeah, I agree.. maybe the show-on-hover (or just dim and highight?) could help there. Great idea about the cell editor!

0biWanKenobi:
I decided to give it a try, this is the result so far:
![image](https://user-images.githubusercontent.com/4061104/34427036-a5c1b3c2-ec3d-11e7-93aa-edc3434a543d.png)

I still need to bind value selecting with actual resultset filtering, but it looks nice to me!

serge-rider:
Thanks for contributing!
Additional enhancements were also pushed

Issue #1748, opened 2017-06-08
Bug: Cannot filter results by column alias


encynator:
It is not possible to filter results using column alias used in select.

**Reproduction scenario:**
1. Perform select where column has alias within select over Oracle DB, eg.:
   select product_id, technical_status **as tech_stat** from BA.PRODUCT;

2. Try to filter result using column alias instead (tech_stat = 'A') of it's name -> ORA-00904 - see attached picture
It works if real column name is used (technical_status = 'A') -> Result correctly filtered

**Version:**
Community version 4.0.8 with builtin Java

**Expected behavior:**
DBeaver either suggest column name in auto completion, but more likely it should be able to filter using column alias.
As it would not be really comfortable to use real column name within complicated selects.

**As you can see on attached picture:**
1. On auto-complete request DBeaver suggests alias
2. Once aliased used, it throws ORA-00904
3. If real name is used, DBeaver successfully filters results

![dbeaver_filter_bug](https://user-images.githubusercontent.com/11492235/26919434-fa207532-4c35-11e7-9239-bebaa63a4d88.png)

serge-rider:
To perform filter DBeaver executes SQL query. In standard SQL you can't use column alias to filter.
But you can try to use nested selects:
```sql
select * from
(select product_id, technical_status as tech_stat from BA.PRODUCT) x
```
In this case filtering and ordering will work fine with aliases.

encynator:
Alright, but in that case - DBeaver should auto-complete regular column name instead of alias, right?
Now auto-complete suggests alias name. It would be fine if DBeaver suggest real/full column name for auto-complete.

serge-rider:
This is because database driver doesn't provide information about underlying column names. And this makes sense because you can use any SQL expression after alias. E.g.
```sql
select col1, col_num * 25 + 10 as some_alias from some_table
```
The only way to use it in filters is to use subselects.

encynator:
I see. Alright then.

esequiel-ferreira:
I see two solutions to this issues:

1.  check real column name by parsing the query
2.  filter local Resultset (optionally when original filtering fails)

If you guys face this feature as a "WHERE clause automation", the approach is to check real column name by parsing the original query would be more appropriate and the solution number 1 could be implemented.

However filtering local Resultset would be less disappointing. It would work as a post-execution filter, not as a WHERE clause automation.

PR #2788, opened 2018-01-14
2619 simplify resultset filtering

0biWanKenobi:
## Target

tl;dr Implementing a quicker way to filter results with a dedicated icon.

Currently, filtering a resultset by column value is a bit of a lenghty process. You must 

- click on a cell ( or the IN.. filter won't work)
- left click to open context menu
- choose submenu filters
- select IN..
- proceed with filtering

As proposed in issue #2619, I implemented a pop up menu with the list of a column distinct values. The user is able to open it via a new icon, place side by side with the sorting icon. This means that the user can filter like this:

- click filter icon
- choose cell value

and that's it! Much quicker ;)

image

## Implementation

I added a new icon alongside the sorting one: click detection is handled exactly the same way. Event triggering, too, happens the same. 

image

The list of column values is a `Composite` holding a `TableViewer` with gray bg and no lines or headers, so that it looks like a menu. Under the list there is a textbox to filter results. 
The list of values is obtained the same way as with `FilterValueEditDialog` class, only in this case the `TableViewer` has just the labels column; this way it's more lightweight for the user. 

## Major project changes

Only one really: I refactored `FilterValueEditDialog` in a new package called `org.jkiss.dbeaver.ui.controls.resultset.valuefilter`. This has been done because there's a lot of common code between the new menu and the existing dialog used for the "IN.." filter, so I wanted to avoid duplicate code and prevent methods and variables from being exposed to all classes in the previous package. Aside from moving code from place to place, no breaking changes have been done and the preexisting code works exactly the same as before. 

## Implementation details

Modified resources:

`GridColumn`, `GridColumnRenderer`, `LightGrid`, `Spreadsheet`, `SpreadsheetPresentation`, `IResultSetController`, `ResultSetViewer`, `FilterValueEditMenu`, `FilterValueEditDialog`, `GenericFilterValueEdit`.

- _GridColumn_: 
  - Since there's a new icon, increase header height to avoid cluttering;
  - With the same logic of the sort icon mouse position detection, add detection for the new filter icon;
  - Adjust sort arrow detection for new x coordinate.

- _GridColumnRenderer_:
   - Since there's a new icon, increase header height to avoid cluttering;
  - Define icon image and size;
  - in paint(), adjust header width for text considering new icon; compute bounds a draw new icon.

- _LightGrid_:
  - With the same logic of the sort icon click detection, add detection for the new filter icon;
  - When a click happens, trigger new Event "Event_FilterColumn".

- _Spreadsheet_:
  - Let Eclipse reorganize imports;
  - Added listener for new event triggered from LightGrid by filter icon click;
  - Call ResultSetViewer's method thru SpreadSheetPresentation when click event is handled.

- _SpreadsheetPresentation_:
  - Before calling ResultSetViewer, deselects any cell in current or other columns 

- _IResultSetController:
  - Abstract method to call when user clicks filter icon.

- _ResultSetViewer_:
  - Concrete implementation of interface method to handle user click;
  - Uses new class FilterValueEditMenu to show a list of column values.

- _FilterValueEditMenu_:
  - New class.  Since it had a lot in common with FilterValueEditDialog, a new package has been created, along with a generic class holding common code;
Package name is `org.jkiss.dbeaver.ui.controls.resultset.valuefilter`;
  - Uses TableView without checkboxes, it allows single selection only;
  - Is modal, user can use Esc button to close it without filtering the resultset.

- _FilterValueEditDialog_:
  - Has been moved to `org.jkiss.dbeaver.ui.controls.resultset.valuefilter`;
  - Common methods have been moved to a generic class;
  - Works exactly the same as before.

- _GenericFilterValueEdit_:
  - Holds all common code and variables between `FilterValueEditMenu` and `FilterValueEditDialog`;


serge-rider:
Wow, great job.
I will review this shortly (hopefully on Monday)

serge-rider:
Found a couple of issues:

1. Filter drop-down fails with NPE when you click on the filter icon second time (click, select value, click on the same column icon again):
```java
java.lang.NullPointerException
	at org.jkiss.dbeaver.ui.controls.resultset.valuefilter.FilterValueEditMenu.createMultiValueSelector(FilterValueEditMenu.java:135)
	at org.jkiss.dbeaver.ui.controls.resultset.valuefilter.FilterValueEditMenu.init(FilterValueEditMenu.java:59)
	at org.jkiss.dbeaver.ui.controls.resultset.valuefilter.FilterValueEditMenu.<init>(FilterValueEditMenu.java:41)
	at org.jkiss.dbeaver.ui.controls.resultset.ResultSetViewer.showDistinctFilter(ResultSetViewer.java:1598)
	at org.jkiss.dbeaver.ui.controls.resultset.spreadsheet.SpreadsheetPresentation.showFiltering(SpreadsheetPresentation.java:1062)
	at org.jkiss.dbeaver.ui.controls.resultset.spreadsheet.Spreadsheet.handleEvent(Spreadsheet.java:324)
```
2. It is better to make drop-down like composite (FilterValueEditMenu) as non-modal dialog. Similarly to how schema selector dialog was made (it sets focus in selector table and it closes when table looses focus): see DataSourceManagementToolbar.showDatabaseSelector and SelectObjectDialog.java . It is much more user friendly.


Also it makes sense to add standard copyright header in all new source files.

Or I can do this after PR merge. And add some other extra improvements as: shortcut, filter icon width reduce, last filter value selection save and maybe some others.

0biWanKenobi:
I'll take a look at the NPE and make the drop-down close when focus is lost. Sorry about forgetting the copyright header, I'll add that too! 
I'll leave the improvements to you, if that's ok with you 😄

0biWanKenobi:
I can't seem to reproduce the bug you mention when clicking the same filter icon multiple times. It happened once, seemingly at random, but I wasn't debugging.. 😕  can you help with some more details?

serge-rider:
NPE is gone 👍 
But I still think that drop-down should be a dialog. I'll change this by myself, 
thanks a lot for contribution!

0biWanKenobi:
You're welcome! It's been fun 😄  I was thinking, maybe we should make the dropdown list only visible cells? It would help composite filtering a lot!

I went through the code and realized that the whole `KeyLoadJob` purpose in `GenericFilterValueEdit` is to retrieve the values for all cells, even those hidden by the current filter.. Maybe it could be avoided altogether?

serge-rider:
Yep, perhaps. Also I think it should allow multiple values selection (with checboxes).
Please check my changes (non-modal dialog and others)

0biWanKenobi:
Yeah I've seen them, great work! After your changes everything is much more refined. I had totally missed the integration with the filtering option in DBeaver settings 😅   

Regarding multiple selection, it would bring in the need of some sort of confirmation, and I don't really like the idea of introducing buttons. So I implemented ctrl+click for multi selection (see video in attached zip), how do you like it? Could be a good idea?

[Demo multiselection.zip](https://github.com/dbeaver/dbeaver/files/1640648/Demo.multiselection.zip)

As a personal preference, I would have the dropdown look like this, without the header..
image
Because I feel it's more of a menu than a dialog.. but then again, it's just my POV!

esequiel-ferreira:
nice feature! Its very useful. I would suggest a little improvement: When there are many possible values, the popup shows a textbox for search in the possible values. 

Could you guys set focus to textbox when its visible? Also, when press Up key inside this textbox, it would be nice to set focus to Grid of possible values, so we would be able to navigate then with arrows keys.

The point is: it would great if I was able to do all the process with keyboard after clicking in the Filter button.

serge-rider:
@esequiel-ferreira good point. Check it in the next version.

esequiel-ferreira:
Nice! 

Just one more thing: When I run a select statement with alias, the filter button is still there and when clicking on it a popup with possible values is shown. However, when applying a value to filtering an error happens. 

Its clear from #1748 that filtering alias column is currently not possible, but to get this feature (filter column popup) at state of art, i think this issue should be handled.

PR #2790, opened 2018-01-15
2556 support postgre session methods

ruspl-afed:


serge-rider (commented):
Do not declare both DBGException and DBException in one method. DBGException extends DBException.
Also I think we should use DBGException in all debug API methods and shouldn't use plain DBException.

PR #2802, opened 2018-01-17
Some missed PG keywords and functions

AndrewKhitrin:
I've added some Postgres SQL keywords and functions not all of course but ....


serge-rider:
👍

PR #2951, opened 2018-02-09
2556 close debug session

ruspl-afed:


AndrewKhitrin (approved):
Terminate looks horribly but it will work

PR #3156, opened 2018-03-16
Preventing PostgreSQL from dump privileges and owners.

gamikun:
I added two options in backup dialog:

1. "Do not backup privileges (GRANT/REVOKE)".
If you need the structure but you will use different permission schema, you can export without them (uses --no-privileges from pg_dump).

2. "Discard objects owner"
If you will not use the same user for the environment you will import the exported database. (uses --no-owner from pg_dump).

backup_y_dbeaver_5_0_1


serge-rider:
Looks great, 
thanks!

PR #3309, opened 2018-04-11
#3302 optimized UUID ordered format transformator

focbenz:
As requested in Feature Request #3302 here is a display transformator for the optimized storage format for UUIDs mentioned in article: https://www.percona.com/blog/2014/12/19/store-uuid-optimized-way/

Sorry for the messed up indentation ... tried to correct it online but failed.
Would create a new Pull Request if necessary because I was missing the correct formatter options.


serge-rider:
Looks good!

PR #3386, opened 2018-04-26
Add i18n and l10n for undo and redo

liuyuanyuan:


serge-rider:
@liuyuanyuan Great job, as always 👍 
However there is a small issue. Some files looks like completely changed (every line is first deleted and then added). E.g. CoreResources_zh.properties. This is because Unix/Windows linefeeds change - usually this happens if your IDE (Eclipse I guess) automatically converts linefeeds before commit. You can disable this behavior in preferences.

Could you please fix this in committed files? Otherwise we'll loose changes history.

liuyuanyuan:
@serge-rider Thanks, I'll try to solve the linefeeds  problem ~~

PR #3394, opened 2018-04-27
typo of translation for sql_editor_resultset_filter_panel_control_execute_to_see_reslut

ihipop:
should be `执行查询 {0} 或者脚本 {1} 来查看结果`
image
image

--------------

And,why there is additional ` single` `double quote` at end of these line ?
@liuyuanyuan 
image
I didn't change these quote


serge-rider:
Thanks!
This trailing quote character seems to be a bug (or just incorrect usage of resources). It should be something like `Add node {0}`.

PR #3441, opened 2018-05-07
OracleSchedulerJob - Added a run command

crowne:
Added a run command to execute the job asynchronously and added a panel for viewing the job action which could be a PLSQL block up to 4000 characters long.

I'm fairly new to github and eclipse development, so I apologise in advance for any errors.
I've enjoyed learning through making these changes to DBeaver, and hope to contribute more in future.
Please review and let me know if this is a successful PR.
Is it an issue that I developed the change in the master branch?
Should changes first be done in devel?


serge-rider:
Thanks,
I'll review it and write some comments.
It is better to work with develop always. Master is always a bit outdated (and sometimes not a bit).

serge-rider:
Generally looks good.
There are a couple of issues with formatting (mixed tabs and spaces) and one strange if/else instead of switch. But that's easy to fix.
The only thing which concerns me is quite big handler class which is very similar to existing CompileHandler. Perhaps we need some base class for them..

PR #3463, opened 2018-05-13
Support for export in DbUnit dataset format - improved formating

henriklen:
I did more extensive testing on the export and realised that I could not rely on the format used by the original XML export as much as I had hoped since it did locale dependent formating in order to be more human readable. I've added DbUnit-compatible formating and verified the corresponding java classes mapping.


serge-rider:
👍

PR #3488, opened 2018-05-17
Spanish translation

faibistes:
Hi, this is my first PR ever, the spanish translation. It seems that I can't change "core" eclipse messages, tho. Is there any way?


serge-rider:
Great job 👍 
One note - no need to keep copyright notices in resource files - you should insert your name there instead of original one :)

Regarding core eclipse messages. They can be localized only from Babel package (all strings are already translated on all languages, you don't need to do this again). I'll add Spanish language in default locale list in DBeaver build procedure. You can test it in EA version (https://dbeaver.io/files/ea/)

serge-rider:
If this will work then I'll include Spanish in installers as well.

PR #3664, opened 2018-06-20
PostgreSQL 10+ sequence show in tree fix

akos2:
For PostgreSQL version 10+, it's impossible to get sequence properties from sequence directly. Catalog pg_sequence view must be used. Version check and SQL select for model was added.


serge-rider:
Thank you. 
However it is the fix of Generic driver, main PostgreSQL driver already contains this improvements.
Is there any specific reason why do you use Generic driver? We plan to remove it in some future version.

akos2:
Thanks for explanation. It was not obvious for me that i can use JDBC from https://jdbc.postgresql.org/download/postgresql-42.2.2.jar and and use other type than Generic.
I've just tested your suggestion - it works fine. Modern features like multilevel partitions and identity columns are also OK.

Thanks a lot.

PR #3834, opened 2018-07-20
Move to V6 Driver and bug fixes

Sargul:
* bug fix of add column when default value and not null is specified (special order in exasol)
* move to V6 driver as incompatibility with dbeaver result set edit is no more
* bug fix in alter column when mixed case column names where used
* bug for new columns as not all necessary properties where set


serge-rider:
👍

PR #4052, opened 2018-08-22
Fixed TABLE_ROWS column name in mysql partitions

liumengjun:


serge-rider:
Yep, current constant is a bit confusing, thanks.

PR #4068, opened 2018-08-23
Removed .trim to ListToString in InformixUtils

orlandoboletini:
NOT APPLY '.TRIM' TO VARIABLE IN METHOD ListToString, PROBLEM TO FORMAT 'PROCEDURE' BECAUSE LINE DELIMITED CRLF and LF GENERATE 'Sintax error'


serge-rider:
Thanks!

Issue #2768, opened 2018-01-10
auto indent not working on new line after semicolon


wthys:
I'm using DBeaver CE 4.3.2 on Ubuntu 16.04 and have noticed since 4.3.1 that when I start a new line in Oracle PL/SQL code the autoindent as in 4.2.x is only replicated when the previous line was not ended with a semicolon (';'). If the previous line was ended with a semicolon, the new line will start at column 1, not the at the previous indentation level.

There's no autoindent option in the GUI that I can find under preferences so I'm inclined to think this is a regression on previous behaviour

wthys:
It seems it was a deliberate feature added in aedd8cda20f807d6b260517f729bce6a64b3a032. Any chance in getting a config option to turn this on/off?

serge-rider:
Yes, this probably should be configurable.

serge-rider:
Or, better, we shouldn't reset indentation but we should do single "unindent" after semicolon. This makes sense for complicated query/procedures with multiple indentations.

PR #4087, opened 2018-08-26
#2768 Removed auto unindent on new line after delimiter.

michelvosje:
This is a fix for issue #2768.

My reasons for removing the auto unindentation is because:
* It doesn't work properly/consistently.
* It forces unindentation instead of letting the user keep control.


serge-rider:
Agreed, this "unindention" was a bit messy.

Issue #4105, opened 2018-08-28
Formatting already formatted SQL code changes the format


michelvosje:
# How to reproduce.
1. Write some SQL code
```
insert into public.test (value2) values ('1'), ('2');
```
2. Format it using **Ctrl + Shift + F**.
 The code is formatted.
```
insert
	into
		public.test ( value2 )
	values ( '1' ),
	( '2' );
```
3. Format it again using **Ctrl + Shift + F**.
 The format changes in a slightly different format.
```
insert
	into
		public.test (value2)
	values ('1'),
	('2');
```
4. Format it again using **Ctrl + Shift + F**.
 The format is the same as in step 2.
```
insert
	into
		public.test ( value2 )
	values ( '1' ),
	( '2' );
```

DBeaver version: Current devel branch version and in 5.1.6.
OS: Windows 10
Driver: PostgreSQL

serge-rider:
@michelvosje Thanks for the improvement!

PR #4122, opened 2018-08-30
#4105 Formatting SQL code is now consistent.

michelvosje:
Related to ticket: #4105.

Changes:
* Reformatting already formatted SQL code should not change.
* Spaces are not inserted after '(' and before ')'.

Note that i am not 100% sure about the effects of the following part:
```
if (prev.getType() != TokenType.SPACE &&
                token.getType() != TokenType.SPACE &&
                !prev.getString().equals("(") &&
                !token.getString().startsWith("(") &&
                !prev.getString().equals(")") && <-- this line
                !token.getString().equals(")"))
```


serge-rider:
Thank you for the report and the fix.

PR #4174, opened 2018-09-08
Rename Elastic Search to Elasticsearch (official name)

costin:
Thank you for adding support for the Elasticsearch JDBC driver.
I've made a small PR to use the proper/official name (Elasticsearch - one word, no camel case) and updated the JDBC driver to the latest version.

P.S. Each version of Elasticsearch comes with its own JDBC driver. Not sure whether it is possible, but it would be great in the maven artifact to use a placeholder like latest to avoid having to manually update the jar on each release: 
`<file type="jar" path="maven:/org.elasticsearch.plugin:jdbc:{latest}"/>`

Cheers!

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.es.ui/META-INF/MANIFEST.MF
+++ b/plugins/org.jkiss.dbeaver.ext.es.ui/META-INF/MANIFEST.MF
@@ -1,6 +1,6 @@
 Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
-Bundle-Name: Elastic Search UI
+Bundle-Name: Elasticsearch UI
 Bundle-SymbolicName: org.jkiss.dbeaver.ext.es.ui;singleton:=true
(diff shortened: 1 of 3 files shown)


serge-rider:
Ah, apparently artifact name was also changed (from `jdbc` to `x-pack-sql-jdbc`). Fixed.

Unfortunately I can't set version id to RELEASE. Elasticsearch Maven (Nexus?) doesn't support version listing (there is no maven-metadata.xml file in artifact root folder). That's why particular version has to be set.
(dbeaver works with Maven according to http://maven.apache.org/ref/3.2.5/maven-repository-metadata/)

costin:
For some reason I didn't get notified on your comments - sorry about that.
I'll see what can be done regarding the `maven-metadata.xml` and report back (or can create a new issue if you prefer that).

Cheers,

serge-rider:
Yep, new issue would be preferable. 
Regards

Issue #4117, opened 2018-08-30
Tip of the day popup


serge-rider:
I think we need Tip Of The Day popup. DBeaver has too many features and this could be very useful for users.

tati-kru:
Create a separate DBeaver plugin which will show Tip-of-the-day Nebula widget. Tips should be stored in the some localized xml-file. Tips will be created separately.

PR #4257, opened 2018-09-21
#4117: Added tip of the day functionality.

tchibatron:
I'm not sure that I need add another plugin to solve this issue. I guess my solition is also fine, but if I'm wrong can make it as a separate plugin.
Also I haven't found any ways to add config to General tab. To my mind it is not extensible. Because of this I've looked at other configs and found out, that all general custom configs are on PrefPageDatabaseGeneral, but may I'm also wrong :) I am open to any discussions.

Issue: Tip of the day popup #4117


tchibatron:
I've created a separate plugin and changed the way of parsing tips file. Current parser supports only primitive html tags. 
And I also use an URL.openConnection to read the file with tips. There are some pros and cons. What do you think, is it OK?

ruspl-afed:
Do you think it is possible to use just standard .properties file for tips definition?
```properties
tip_id1=Some <b>rich text</b> content
tip_id2=Another <u>cool</u> tip
```
This may simplify both implementation and localization.

tchibatron:
If it is a property file, the support will be more difficult. Each time, when you need to add new tip, you will have to add a constat to java code, to get this tip as a property or think about the solution to dermine the count of tips and configure property id in runtime. I recon xml file is better in this case.

plugins/org.jkiss.dbeaver.core.application/plugin.xml
```diff
@@ -50,6 +50,7 @@
             <property name="applicationCSSResources" value="platform:/plugin/org.eclipse.platform/images/"/>
             <property name="rendererFactoryUri" value="bundleclass://org.jkiss.dbeaver.core/org.jkiss.dbeaver.ui.e4.DBeaverRendererFactory"/>
             <property name="preferenceCustomization" value="plugin_customization.ini"/>
+            <property name="tipsFile" value="platform:/plugin/org.jkiss.dbeaver.ext.tipoftheday/tips.xml"/>
```
ruspl-afed: Please use `platform:/plugin/org.jkiss.dbeaver.ext.tipoftheday/$nl$/tips.xml` to support translation

plugins/org.jkiss.dbeaver.ext.tipoftheday/src/org/jkiss/dbeaver/ext/tipoftheday/TipOfTheDayInitializer.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+        return result;
+    }
+
+
+    private InputStream getTipsFileStream() throws IOException {
+        String pathToTipsFile = Platform.getProduct().getProperty("tipsFile");
+
+        URL url = new URL(pathToTipsFile);
+        return url.openConnection().getInputStream();
```
ruspl-afed: Please have a look at
`org.eclipse.core.runtime.FileLocator.find(URL)`

ruspl-afed:
Constant in java code is not required to work with .properties, you can just use all the properties loaded with some `InputStream` like you do for .xml now.
Anyway, please do not forget to support localization of tips with `$nl$` segment in your URL

serge-rider:
Going to merge but can't. DBeaverPreferences.java and DBeaverPreferencesInitializer.java were changed in every line (different line feeds I guess) so any change in them leads to a conflict :(

PR #4333, opened 2018-10-04
Issue#4170

fangkung:
Format SQL : Added a line break before the closing bracket and the added indentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ buildNumber.properties
 
 *.iml
 /idea/
+.idea
 
 #############
 ## Windows detritus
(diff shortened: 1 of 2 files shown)


serge-rider:
As you can see in the git history this code was commented for a reason. It produces too verbose SQL in some cases (like in MySQL view definition) making it very inconvenient.
I think that by default it should work as is. 
We need new formatter option on this page:

image

It is here: https://github.com/dbeaver/dbeaver/blob/e0f3d52c8c10119847f0f7ab464125438b73db9e/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/editors/sql/format/tokenized/SQLTokenizedFormatterConfigurationPage.java

PR #4394, opened 2018-10-13
Update plugins/org.jkiss.dbeaver.core/OSGI-INF/l10n/bundle_de.properties

Bullnados:
Start of German Translation Core Command/Props (colorDefinition)


serge-rider:
👍

PR #4432, opened 2018-10-19
Update plugins/org.jkiss.dbeaver.core/OSGI-INF/l10n/bundle_zh.properties

fengchao:
Update translation of zh translation for dbeaver.core.


serge-rider:
Great, thanks!
BTW you could add yourself in the copyright section (on the top)

PR #4501, opened 2018-10-27
German translation

Bullnados:


Bullnados:
Now only oracel (2/3) and db2 (near all) is missing

serge-rider:
Great, as always 👍

Issue #4500, opened 2018-10-26
Filter Database Objects not allowing multiple or negative conditions


KipperK:
I am trying to setup database filtering for one of my MariaDB connections in DBeaver 5.2.3 but it is failing due to syntax errors.

Multiple "Include" filters:
```
SQL Error [1064] [42000]: (conn:1105793) You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near '( LIKE 'database1' OR  LIKE 'database2')' at line 1
Query is: show databases( LIKE ? OR  LIKE ?), parameters ['database1','database2']
```
The query should be:
```
show databases WHERE (`Database` LIKE 'database1' OR `Database` LIKE 'database2');
```

Similarly, "Exclude" conditions also have a syntax error.
```
SQL Error [1064] [42000]: (conn:1105793) You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'databasesNOT ( LIKE 'database_prefix%')' at line 1
Query is: show databasesNOT ( LIKE ?), parameters ['database_prefix%']
```
The query should instead be:
```
show databases WHERE (`Database` NOT LIKE 'database_prefix%');
```
Please let me know if I should provide any more information.

serge-rider:
Duplicate of #4554

KipperK:
I jest tested f8258307a5916ef3574fe4c6d01752d73224e761 and can confirm it is working.  Thanks!

PR #4543, opened 2018-11-05
Issue #4500

rubengonzalezasencio:
Filter Database Objects error


serge-rider:
Please write a bit more info. In my test env it works fine with current filter.
Query "show databases like ?" is just fine. 
What is the point to replace it with 
```sql
show databases WHERE `Database` LIKE ?
```
?

ydq:
dbeaver

PR #4546, opened 2018-11-05
Refactor SQLFormatterTokenized. Extracted new class for indent format…

tchibatron:
…ting. Added unit tests.


serge-rider:
Hey :)
I think this all makes sense. I'll merge it, thanks for looking into it.

However it'll much better to move all Maven configs from particular bundle pom (org.jkiss.dbeaver.model at the moment) to some shared place. Maybe  somewhere in root pom or in sum-module in /tests folder.
WDYT?

tchibatron:
Root pom sounds good for me. I'll do it near future.

PR #4563, opened 2018-11-06
German translation continued

Bullnados:
Found some untranslated properties in the depths of the folder structure 😄


serge-rider:
👍  as usual )

PR #4584, opened 2018-11-08
Fix unit test. Added new test for case when script starts with comment. Mo…

tchibatron:
…ved test dependencies to root pom and added also source root for tests and source code there.


serge-rider:
👍

Issue #4170, opened 2018-09-07
Add option to change brackets style for formatter


kufii:
Right now the brackets style like this:

```sql
SELECT
	(
	SELECT
		thecol
	FROM
		thetable )
FROM
	dual
```

I would personally much prefer a brackets style like this, with a line break before the closing bracket. Is this a feature you could add?

```sql
SELECT
	(
		SELECT
			thecol
		FROM
			thetable
	)
FROM
	dual
```

fangkung:
This issue want to add the button option to change brackets style or Can I modify code for new expected formatter.

PR #4596, opened 2018-11-10
DBVR-4170: Added special config to do indent and linebreak before clo…

tchibatron:
…se bracket.

#4170


serge-rider (approved):
Looks good 👍 
However we need to think about general tests organization. Moving them to a separate modules and still try to use SWT Bot (I know it is not working well but at least it can test something).

PR #4697, opened 2018-11-27
optimize meta data query performance and organize scripts

Sargul:
- speed up connect time by minimizing meta queries
- optimize meta queries for constraints on systems with a lot of objects
- optimize meta queries for schema objects on systems with a lot of objects
- split scripts into procedures, udfs and adapter scripts


serge-rider:
Looks great 👍

Sargul:
Forgot new class

PR #4759, opened 2018-12-08
firebird database plan

tomas303:
Hello , I added possibility to display plan for firebird database. Firebird plan can be received via api only as series of subqueries' plans and main query plan. That info is displayed in top level nodes, but in addition is parsed to tree nodes and statistics for index fields are retrieved.
dbeaverfbplan


serge-rider:
Wow, looks great 👍

serge-rider:
Generally it works just fine. But in trivial case (SELECT * FROM TableName) it fails in plan parser (`Error parsing plan - expected NATURAL at position 14 but got NATURAL
PLAN (COUNTRY^^^NATURAL)`).

I've removed one check and exception thrower. Please check https://github.com/dbeaver/dbeaver/commit/05d141ea21fea1a7be8bd1446156b6465fcfa1d7

tomas303:
Yes, unfortunately this trivial case I didn't test. I checked 05d141e  and  suggest uncomment line with check and add jump before it. But it is only cosmetic change - just to check that parsing was ok.

                    tokenMatch.jump();
                    tokenMatch.checkToken(FireBirdPlanToken.RIGHTPARENTHESE);

I had tested it on my other queries as well and was without problems too. 
In cases where index is involved instead natural, plan ends with more right parentheses so checkToken didn't raise exception. 
And thank you for taking my changes.

serge-rider:
Ok. 
See it in 5.3 :)

PR #4763, opened 2018-12-10
Office plugin change

Arty68:
Only selected rows and columns are displayed in excel if nothing selected everything is exported to excel


serge-rider:
Good one.
However I've patched it a bit. Export of selected rows/cols will work only if more than one row is selected (because you can't "unselect" a single cell if you already selected it)

Arty68:
Hi
Thank you for merging.

Le lun. 10 déc. 2018 à 23:07, Serge Rider <notifications@github.com> a
écrit :

PR #4766, opened 2018-12-10
Outlook Plugin

Arty68:
Outlook : open the selected rows in an new Outlook mail.
This need the powershell script (CopyHtml.ps1) in the main dbeaver directory

grafik

grafik


serge-rider:
I guess - do we really need yet another separate export command in toolbar/context menu? Maybe it'd be better to implement this as yet another data export format?

Excel export is special as it may be used very often in some cases.

As I understand the main difference is that regular data export doesn't launch some appropriate program (like Outlook) to open exported file. We can add such feature to stream exporters though. It will be more consistent as for me.

WDYT?

Arty68:
Hi

do we really need yet another separate export command in toolbar/context
menu?   -> I would say yes It's a bit same as for excel
But maybe we can also add some new context Menu named "Quick export"

Maybe it'd be better to implement this as yet another data export format? -
>
It's not easy to implement another export Format for Outlook everything is
binary microsoft proprietary format.
That's why I use a powershell script to interface it .
The outlook  toolbar/context menu export everything in html format , and
start the powershell script that open a new outlook mail and "paste it"
html to the bodyhtml property.
So an other export format I would say no expect if you know opensource
ibrabry that write directly outlook binary format ?




Le lun. 10 déc. 2018 à 23:15, Serge Rider <notifications@github.com> a
écrit :

Arty68:
Hi ! Maybe I will use HSMF (Horrible Stupid Mail Format ou Format horrible, stupide de courrier)  from Apache POI.
Will it possible then to integrate it in the office plugin ?

p91paul:
looks like hsmf from poi is read only... can't outlook be made to open an html mail somehow?

Arty68:
> looks like hsmf from poi is read only... can't outlook be made to open an html mail somehow? perhaps eml format, which can be understood by thunderbird as well

Yes with powershell but that is what i have done with my other template :D

Arty68:
Maybe with this : https://github.com/ctabin/jotlmsg

serge-rider:
Closing as really outdated.
We still need Outlook format. Thanks for the hint. Probably will add it closer to 7.3 version.

Arty68:
Hi it would be to have this Outlook format

PR #4789, opened 2018-12-11
Fix color for connection

tchibatron:
Added special css class for elements, which should have different color, when connection is colored. Added special handlers for supporting this class in runtime. Also added changing CSS scheme in runtime for applying styles for TabbedFolder component.


serge-rider:
Wow, sounds like magic 👍 

Will take a look shortly..

tchibatron:
Yes, but unfortunately this does not change background of all needed components:
image

serge-rider:
Generally we could give it a try. but we need to:
1. Have ability to disable this feature (enable by default + option in preferences)
1. Move CSS related stuff like `xxx.setData(CSSSWTConstants.CSS_CLASS_NAME_KEY, COLORED_BY_CONNECTION_TYPE)` into some utilities class. E.g. CSSUtils.setControlStyles(control, COLORED_BY_CONNECTION_TYPE). In future we can call this function from some base UIUtils functions.
1. Rename classes ConnectionSpecifiedCompositeElementHandler  into something more general. Potentially we may have multiple different color schemas, not just connection-related.
1. Move CSS constants into some constant class (e.g. org.jkiss.dbeaver.ui.DBStyles).

tchibatron:
@serge-rider  Everything is done.

serge-rider:
Ok, lets try it.
BTW what is buildNumber.properties?

tchibatron:
I don't know what buildNumber.properties is. I've added only ./idea to gitignore. 

What purpose can be to colorizing tabs except connection?

Issue #4346, opened 2018-10-08
SQL Workbench/j formatter error


ced75:
Hi,

I'm trying to configure Sql workbench/j formatter as SQL formarter.
I'm using DBeaver 5.2.2 under Linux Mint 19, with Java 8

I've downloaded and unzip SQL Workbench/j (latest version).
In DBeaver preferences, I select "Sql workbench/j formatter" as formatter and point the directory containing SQL Workbench/J jar file in "SQL Workbench/J path".
When I click on the "Apply" button, I get the following error message:
Error formatting with SQL Workbench/J

Motif:
Error calling SQL Workbench/J formatter

Here is the log entry associated with this error:
2018-10-08 08:29:01.348 - null
org.jkiss.dbeaver.DBException: Error calling SQL Workbench/J formatter
	at org.jkiss.dbeaver.ext.format.sqlworkbenchj.SQLWorkbenchJManager.format(SQLWorkbenchJManager.java:94)
	at org.jkiss.dbeaver.ext.format.sqlworkbenchj.SQLWorkbenchJFormatter.format(SQLWorkbenchJFormatter.java:46)
	at org.jkiss.dbeaver.ui.editors.sql.syntax.SQLFormattingStrategy.format(SQLFormattingStrategy.java:63)
	at org.eclipse.jface.text.formatter.ContentFormatter.format(ContentFormatter.java:501)
	at org.eclipse.jface.text.formatter.ContentFormatter.formatRegion(ContentFormatter.java:412)
	at org.eclipse.jface.text.formatter.ContentFormatter.format(ContentFormatter.java:364)
	at org.eclipse.jface.text.source.SourceViewer.doOperation(SourceViewer.java:992)
	at org.eclipse.jface.text.source.projection.ProjectionViewer.doOperation(ProjectionViewer.java:1378)
	at org.jkiss.dbeaver.ui.preferences.PrefPageSQLFormat.formatSQL(PrefPageSQLFormat.java:306)
	at org.jkiss.dbeaver.ui.preferences.PrefPageSQLFormat.performApply(PrefPageSQLFormat.java:259)
	at org.eclipse.jface.preference.PreferencePage.lambda$1(PreferencePage.java:286)
	at org.eclipse.swt.events.SelectionListener$1.widgetSelected(SelectionListener.java:81)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:249)
[...]

PR #4793, opened 2018-12-12
#4346: Use MySQL formatter type for WBFormatter, when there is not con…

tchibatron:
…figured connections.

Fix for issue #4346


serge-rider:
Thanks 👍

PR #4797, opened 2018-12-12
Adds WITH clause to table DDL for GreenPlum DB.

TheBoysAreZachInTown:
In GreenPlum, users can use the create table command containing a WITH clause, but when viewing the DDL in DBeaver for such a table, the WITH clause was missing. To make it easier for a user to create an identical table, the DDL now includes the WITH clause and any attribute key-value pairs within.


plugins/org.jkiss.dbeaver.ext.postgresql/tests/org/jkiss/dbeaver/ext/postgresql/model/PostgreWithClauseAppenderTest.java
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    @Test
+    public void generateWithClause_whenTableIsGreenPlumTableWithOidsWithMultipleOptions_shouldDisplayWithClauseWithOidsAndAllTheOptions() {
+        table = mock(GreenplumTable.class);
+
+        when(serverExtension.supportsOids()).thenReturn(true);
+        when(dataSource.getServerType()).thenReturn(serverExtension);
+        when(table.getDataSource()).thenReturn(dataSource);
+        when(table.isHasOids()).thenReturn(true);
```
tchibatron: Looks like each test has similar code blocks. Why don't you extract this code blocks to setup method.

plugins/org.jkiss.dbeaver.ext.postgresql/tests/org/jkiss/dbeaver/ext/postgresql/model/PostgreWithClauseAppenderTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+import static org.jkiss.dbeaver.ext.postgresql.model.PostgreWithClauseAppender.generateWithClause;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class PostgreWithClauseAppenderTest {
+
+    private PostgreTableRegular table;
+    private PostgreTableBase tableBase;
```
tchibatron: You could get rid of mocking table and tableBase in each test using Mock and RunWith(MockitoJunitRunner.class) annotations.

serge-rider:
Also, I think we should move PostgreWithClauseAppender somewhere in PostgreServerExtension interface (feel free to add new function there and add real implementation in PostgreServerGreenplum).
All GP-specific code should be in org.jkiss.dbeaver.ext.postgresql.model.impls.greenplum package.

Otherwise we couple base PostgreSQL table manager with Greenplum-dependent functionality.

plugins/org.jkiss.dbeaver.ext.postgresql/pom.xml
```diff
@@ -11,4 +11,53 @@
   <artifactId>org.jkiss.dbeaver.ext.postgresql</artifactId>
   <version>2.1.71-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
+  <build>
+    <sourceDirectory>src</sourceDirectory>
+    <testSourceDirectory>tests</testSourceDirectory>
```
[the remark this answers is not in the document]
TheBoysAreZachInTown: Correct. The sourceDirectory and testSourceDirectory tags are not needed and have now been removed.

plugins/org.jkiss.dbeaver.ext.postgresql/tests/org/jkiss/dbeaver/ext/postgresql/model/PostgreWithClauseAppenderTest.java
gshaw-pivotal: Refactored the test file to move this setup to a private setup helper method. The logic can't cleanly go into a '@Before' method as the when mocks for supportOids and isHasOids do change. Now each test calls setupGeneralWhenMocks and pass two booleans to indicate how it wants these to when mocks setup.

The alternative is to have the '@Before' set default when mocks and then have those tests that want different behavior override the default(s) by setting their own when.

plugins/org.jkiss.dbeaver.ext.postgresql/tests/org/jkiss/dbeaver/ext/postgresql/model/PostgreWithClauseAppenderTest.java
gshaw-pivotal: Refactored the unit tests to use the MockitoJunitRunner as suggested. This removed the need for a `@Before` annotated setup method.

Some of the unit tests require that `table` be a mock of the `GreenplumTable` class and thus they set directly set this overriding the global mock of PostgreTableRegular.

gshaw-pivotal:
So to be sure I understand your suggestion:

PostgreTableManager would change to have something akin to `ddl.append(tableBase.getDataSource().getServerType().createWithClause(table))` in the `appendTableModifiers` method.

Then PostgreServerExtension would add `String createWithClause(PostgreTableRegular table)`

PostgreServerGreenplum would have an implementation similar to what is in the PostgreWithClauseAppender class that is in the PR at this point in time (this class could end up going away or become some kind of helper to PostgreServerGreenplum.

And the following would need some kind of implementation of `createWithClause()` due to them all extending `PostgreServerExtension`:
PostgreServerCockroachDB
PostgreServerPostgreSQL
PostgreServerTimescale
PostgreServerRedshift
PostgreServerYellowBrick

Their implementation could be a copy of the original behavior that was found in PostgreTableManager
```
if (table.getDataSource().getServerType().supportsOids() && table.isHasOids()) {
    ddl.append("\nWITH (\n\tOIDS=").append(table.isHasOids() ? "TRUE" : "FALSE");
    ddl.append("\n)");
}
```

However, if each of them did this there would be duplication so maybe they all use a helper that does this for all of them (like the `PostgreWithClauseAppender` class).

Does this sound like it fits in with your thinking?

plugins/org.jkiss.dbeaver.ext.postgresql/tests/org/jkiss/dbeaver/ext/postgresql/model/impls/greenplum/GreenplumWithClauseBuilderTest.java
```diff
@@ -0,0 +1,112 @@
...
+        String withClause = generateWithClause(table, tableBase);
+        assertEquals("\nWITH (\n\tOIDS=TRUE, appendonly=true, orientation=column\n)", withClause);
+    }
+
+    private void setupGeneralWhenMocks(boolean supportOids, boolean hasOids) {
+        when(serverExtension.supportsOids()).thenReturn(supportOids);
+        when(dataSource.getServerType()).thenReturn(serverExtension);
+        when(table.getDataSource()).thenReturn(dataSource);
+        when(table.isHasOids()).thenReturn(hasOids);
+    }
```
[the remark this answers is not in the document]
gshaw-pivotal: No they are not needed. I missed cleaning them up during the refactor as those particular tests moved to another test class.

Apologies for that.

serge-rider:
> And the following would need some kind of implementation of `createWithClause()` due to them all extending `PostgreServerExtension`:
> ...
> PostgreServerYellowBrick
> 
> Their implementation could be a copy of the original behavior that was found in PostgreTableManager

Yep. Base implementation can be moved into PostgreServerExtensionBase. And overrided in PostgreServerGreenplum.

However I'd better make it even more basic. Whole function PostgreTableManager.appendTableModifiers can be moved into String PostgreServerExtension.getTableModifiers(DBRProgressMonitor monitor, PostgreTableBase tableBase, boolean alter)

serge-rider:
Looks good, thanks. I'll merge this later today.
Further plan is to move GP-specific stuff in the new plugin + make Tycho configs more general (move it in root pom).

PR #4817, opened 2018-12-17
Update plugin versions

ddubson:
- Bump `tycho-p2-repository-plugin` version to `1.3.0` (from `1.1.0` and `1.1.0-SNAPSHOT`)
- Bump `jetty-maven-plugin` version to `8.1.16.v20140903` (from `8.1.5.v20120716-SNAPSHOT`)

Reduce dependence on SNAPSHOT repositories which are out-of-date and unstable, due to the nature of SNAPSHOTs themselves.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
     <properties>
         <dbeaver-version>5.3.1</dbeaver-version>
         <dbeaver-product>DBeaver</dbeaver-product>
-        <tycho-version>1.1.0</tycho-version>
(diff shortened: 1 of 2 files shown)


serge-rider:
Thanks,

However localRepository is not used in DBeaver build directly. It prepares p2 dependencies (converts from maven format into p2 repository) so it doesn't really matter what Tycho or Jetty version is used there. 
But root Tycho version matters.

PR #4841, opened 2018-12-20
Hide partition tables from the table list under schemas for greenplum.

gshaw-pivotal:
For Greenplum partition and sub-partition tables are no longer shown as siblings of the root table they are part of.

They are discoverable via the `Child Table` option of their parent table.


serge-rider:
👍

PR #4875, opened 2018-12-29
Rename German translation of label ...core.window.preferences.name

DerTomm:
The German translation of the label "command.org.jkiss.dbeaver.core.window.preferences.name" is currently "Eigenschaften" which actually means "properties" in English. A better translation is the word "Einstellungen" which is commonly used for "preferences" in German application language. In fact I had to search some time until I found the user preferences dialog due to the current translation.


serge-rider:
Thanks for fixing this

Bullnados:
Mmh for me the English settings was "Einstellungen", 
So where is the difference between settings and preferences?
I used for all preferences = Eigenschaften and for settings = Einstellungen, so u should not translate only one single string.

DerTomm:
"Settings" [1] and "preferences" [2] can both be translated with "Einstellungen". On the other side "properties" refer to the properties of an object - not the preferences of the user.

>so u should not translate only one single string, but rather all files where this combo exists

I checked the other german language bundle files and did not find any other occurences of "preferences" but only "properties" which where all correct. Do you see occurences which are not translated consistently now?

[1] https://www.dict.cc/?s=settings
[2] https://www.dict.cc/?s=preferences

PR #4895, opened 2019-01-02
GP: Add DDL generation for Greenplum's concept of external tables

gshaw-pivotal:
- Greenplum has a concept of external tables that store data outside of the primary Greenplum store.
- DBeaver Greenplum plugin had no concept of DDL generation specifically for external tables
- This commit adds basic external table DDL generation

Example external table DDL:

```sql
CREATE EXTERNAL TABLE ext_customer
   (id int, name text, sponsor text)
   LOCATION ( 'gpfdist://filehost:8081/*.csv' )
   FORMAT 'CSV' ( DELIMITER ',' );
```


serge-rider:
Looks good 👍 

Please add header with Apache license in new files. Add yourself and/or Pivotal in author area.

Notes: 
- We usually do not use static imports
- As for me wide usage of String.format in combination is with StringBuilder is a bit odd and makes code less readable. I'd suggest to use StringBuilder and/or simple strings concatenations. String.format doesn't make much sense unless you are using some specific formatting options.

gshaw-pivotal:
Thanks.

Commit added to address issues in the Greenplum plugin.

serge-rider:
Awesome :)

Issue #4840, opened 2018-12-20
Adding an external table folder under the Schema tree folder [Greenplum]


Eulerizeit:
Greenplum has the concept of an external table which is similar to the idea of the Postgres FDW. Tables that are external are at the same lever (within the schema) at regular tables and the DDL statements differ depending on the external table source and access method. 

Currently DBeaver displays external tables as regular tables with slightly mangled DDL. We'd like to implement a new folder under the schema that displays external tables separately from regular tables with the proper DDL.

serge-rider:
We added support of GP external tables in some 6.x version.

PR #4917, opened 2019-01-04
Add 'External Tables' category under schema in the database navigator pane

kindbeing:
Greenplum allows for external tables. In the hierarchy shown in the
database navigator pane we now separate external tables into their own
category (External Tables) instead of grouping them under the existing
Tables category.

This solves issue: #4840


serge-rider:
Generally looks good. 
One note: full replace of database navigator tree (in plugin.xml) isn't the best way to add extra nodes. Besides other issues it won't pull changes made for base PostgreSQL navigator tree.
Better way is to use treeInjection  instead of tree elements. For example:
```xml
            <treeInjection path="postgresql/database/schema" after="table">
                <folder type="org.jkiss.dbeaver.ext.greenplum.model.GreenplumExternalTable" label="External tables" icon="#folder_table">
                    <items label="%tree.externalTables.node.name" path="externalTable" property="externalTables" icon="#table">
                    </items>
                </folder>
            </treeInjection>
```

Note: I've added attribute "after" support just now.

ddubson:
Updated commit as per your suggestions. 👍 Let us know if there are any other potential issues.

@serge-rider, @kindBeing @gshaw-pivotal and I are really appreciative of your suggestions and comments on our PRs. Thanks so much for taking the time to give us more direction in how to integrate code into DBeaver!

PR #4946, opened 2019-01-08
Add additional keywords to support External Tables DDL generation.

kindbeing:
- Add keyword "TEMPORARY" to the table ddl generation.
- Add "READABLE" as a keyword.


serge-rider:
👍 
BTW I think we could make these properties (writable, temporary, etc) public, i.e. visible in table editor. Just mark them as @Property.

PR #4947, opened 2019-01-08
Renamed New Connection to New Database Connection in main menu.

riednyko:


serge-rider:
Good, but please do not commit launch config changes (DBeaver.product.launch)

Issue #2010, opened 2017-08-10
Exasol: add comments support


serge-rider:
We need comments support (COMMENT ON?) for Exasol.

Sargul:
Yeah it is comment on. I can do this for most objects

Currently it's only supported for Tables,Columns and Scripts.

Schemas, virtual schemas and Connections are still open.

Still wonder why you open these requests. Has someone from exasol been talking to you? ;-)

Regards

Charly

serge-rider:
Hey Charly,
As some other FRs this comes from email. But it is not from Exasol itself :)

I'm not quite sure how comments are currently working in Exasol. Looks very similar to Oracle. Anyhow - it is a reminder ticket.

Sargul:
Hi,

COMMENT Support was already there for Table, Column, Script objects.

With the Pull Request #2014  I've added comment Support for Schema and Connection objects.

Sargul:
Impelmented in #2028

steven4320555:
This is a closed ticket, but my issue is closely related to the comment function. [It was an issue for me in both previous DBeaver as well as the latest version 5.3.2. ]

Basically, I find the quotation marks are not added to SQL generated for Exasol in column part, the table, schema are corrected quoted. So the comment function works very well for table, but not working for column, because of the wrong quotation. I have tried to set up the meta data both locally and globally, by different combinations,but, no luck. 
![image](https://user-images.githubusercontent.com/16542163/50866550-39fca880-13a2-11e9-910f-1886097a0f53.png)
  
![image](https://user-images.githubusercontent.com/16542163/50866084-83e48f00-13a0-11e9-8e36-e8b23513d70c.png)]

For your information, I have tested MS SQL server as well as PostgreSQL, which works fine.
![image](https://user-images.githubusercontent.com/16542163/50866143-aecee300-13a0-11e9-8bf0-cc2210cd26f9.png)

A side issue, for the SQL editor, unlike other database, the column names are not populated automatically. It seems the automation level for Exasol is at table level, whereas MS SQL server and PostgresSQL is at column level.

Sargul:
Fixed in 5.3.3

Sargul:
@steven4320555: you can test the fix in version 5.3.3 when it is released

steven4320555:
Hi Sargul. Thanks very much for the help, and sorry for my late reply. I will make sure to open a new ticket instead of comment in a closed ticket. 

[...]

PR #4949, opened 2019-01-09
Fix comment handling and edit/view of scripting objects

Sargul:
* fix issue where column name is not quoted correctly #2010 
* fix issue when handling single quote in comment string
* fix issue when trying to edit adapter and udf scripts


serge-rider:
One suggestion: use SQLDialect.escapeString instead of ExasolUtils.quoteString

PR #4974, opened 2019-01-10
GP: Add `Replicated` keyword for tables in GP v6

gshaw-pivotal:
- Adds support for the `Replicated` keyword for Greenplum v6 table DDL generation.


tchibatron:
I guess that using PowerMock is not the best idea. If you have to mock private methods of your test class, than that means that you have some code quality issues (breaking Single responsibility principle, for example). Using PowerMock you loose strong connection between tests and java code, because you have to call methods using string constants. This would be a compilation error if somebody renamed original method. 
You could extract new class from GreenPlumpTable or prepare a special data for real calling private method. This could be better instead of using PowerMock.

gshaw-pivotal:
We have updated the PR and been able to successfully test without needing PowerMock (and thus removed it from the plugin).

serge-rider:
👍

PR #5094, opened 2019-01-24
Add Dremio driver

mal:
Add knowledge of Dremio and it's JDBC driver. I think I got the categories correct, but happy to defer to someone more familiar with the project if that's not the case. 🙂


serge-rider:
Looks good! 👍

PR #5103, opened 2019-01-25
Hide SSH banner in JSCH implementation.

jportner-hw:
Currently, the SSH tunnel JSCH implementation displays the SSH banner in a separate popup window when connecting.
This change will suppress showing that SSH banner.

This is similar to the default behaviour in MySQL Workbench and other tools.


serge-rider:
Good point. 
We don't need dummy banner but we should log it in in debug log. 
I believe shouldn't be any secure info in that message.

PR #5138, opened 2019-01-28
Optimize toArray

re-thc:
replace toArray(new T[size]) with toArray(new T[0]) for better performance

https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_conclusion


serge-rider:
I'm scared to make so many automatic code fixes. We won't be able to test these changes.

I know that new T[0] works a bit faster but it is not an issue in 99% cases. So I'd prefer to fix them one by one once I'll find them in code.

Thanks for submission anyway.

PR #5146, opened 2019-01-29
ru_RU interface - PostgreSQL

mkgrgis:
null -> NULL, order abc, User Mapping etc.


mkgrgis:
Transfered in https://github.com/dbeaver/dbeaver/pull/5307

PR #5265, opened 2019-02-12
Exasol: Fix auto complete for columns in EXA_STATISTICS schema

Sargul:
special treatment for meta data of exa_statistics schema is necessary as it is a system schema.


serge-rider:
👍

Issue #5419, opened 2019-03-06
[PostgreSQL] regnamespace type shows error in data view


Vlczech:
#### System information: 
- Mageia 7
- DBeaver 5.3.5
- Java version 

#### Connection specification:
- PostgreSQL 10
- Driver name
- Do you use tunnels or proxies (SSH, SOCKS, etc)? 

#### Describe the problem you're observing:
Table with column defined as regnamespace shows error `SQL Error [22003]: Špatná hodnota pro typ int : dbg_test` in displayed data of table. (right button on table, edit table from context navigation)
(Dbeaver 5.1(.6? i think) show nothing for this column - `[NULL]` was here)

psql shows this correctly as it is shown in screenshot.

#### Steps to reproduce, if exist:
```sql
CREATE SCHEMA dbg_test;
CREATE TABLE public.schemas(
  schema_name    regnamespace
);
INSERT INTO public.schemas (schema_name) VALUES ('dbg_test'::regnamespace);
```

#### Include any warning/errors/backtraces from the logs
```
org.jkiss.dbeaver.model.exec.DBCException: SQL Error [22003]: Špatná hodnota pro typ int : dbg_test
	at org.jkiss.dbeaver.model.impl.jdbc.data.handlers.JDBCAbstractValueHandler.fetchValueObject(JDBCAbstractValueHandler.java:55)
	at org.jkiss.dbeaver.ui.controls.resultset.ResultSetDataReceiver.fetchRow(ResultSetDataReceiver.java:122)
	at org.jkiss.dbeaver.model.impl.jdbc.struct.JDBCTable.readData(JDBCTable.java:204)
	at org.jkiss.dbeaver.ui.controls.resultset.ResultSetJobDataRead.lambda$0(ResultSetJobDataRead.java:102)
	at org.jkiss.dbeaver.model.DBUtils.tryExecuteRecover(DBUtils.java:1635)
	at org.jkiss.dbeaver.ui.controls.resultset.ResultSetJobDataRead.run(ResultSetJobDataRead.java:100)
	at org.jkiss.dbeaver.model.runtime.AbstractJob.run(AbstractJob.java:101)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
Caused by: org.postgresql.util.PSQLException: Špatná hodnota pro typ int : dbg_test
	at org.postgresql.jdbc.PgResultSet.toInt(PgResultSet.java:2844)
[...]

PR #5420, opened 2019-03-06
#5419 FIX of regnamespace error

Vlczech:
Regnamespace is OID type => now the regnamespace is shown as string correctly.


Vlczech:
(if it is another 5.* release planned, it would be good to cherry-pick this fix to it ;) )

serge-rider:
👍 
We'll include it in 6.0 (next release)

PR #5472, opened 2019-03-12
database_navigator renamed

Sertarus:


serge-rider:
The task is to change "Database Navigator" view title, not the icon file name.

PR #5476, opened 2019-03-13
dbeaver-corp/dbeaver-i18n#146 fix1

riednyko:
Added missing translation


serge-rider:
Please localize all constraint types in this ticket (not just ones reported in 146)

PR #5501, opened 2019-03-14
To do PR

MayneKeen:
I've finally done this. I'm sorry this've taken so much time. This won't happen anymore.


serge-rider:
Generally it is ok (correct fix was done) but the problem is in commit itself. 
It changes every line in a file. although only one line has to be changed. Please check it and commit again.

PR #5522, opened 2019-03-18
dbeaver-corp/dbeaver-i18n#123

riednyko:
Added missing translation


serge-rider:
Terminology has to be fix in russian - https://postgrespro.ru/docs/postgresql/9.6/sql-grant

PR #5594, opened 2019-03-27
dbeaver-corp/dbeaver-i18n#108 fix1

riednyko:
Added missing translation


serge-rider:
This functionality (attach/detach) is obsolete. I'll drop it from the source code.

PR #5623, opened 2019-03-30
Use constants to control port values

juherr:


serge-rider:
This makes sense, thanks

PR #5648, opened 2019-04-04
Add unit tests for String functions in org.jkiss.utils.CommonUtils

JohnLBergqvist:
Hi,

I've analysed your code base and noticed that `org.jkiss.utils.CommonUtils` in the `bundles` module is not fully tested.

I've written some tests for the functions in this class with the help of [Diffblue](https://www.diffblue.com/) [Cover](https://www.diffblue.com/products).

Due to the size of the class, the tests in this PR just cover the functions that return a `String` value.

Hopefully these tests should help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other particular classes that you consider important.

### DIFF
--- a/bundles/org.jkiss.utils/tests/org/jkiss/utils/CommonUtilsTest.java
+++ b/bundles/org.jkiss.utils/tests/org/jkiss/utils/CommonUtilsTest.java
@@ -1,0 +1,233 @@
+package org.jkiss.utils;
+
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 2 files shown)


serge-rider:
Looks great, thank you very much!
We constantly lack unit tests, it is awesome when someone contribute one :)

You could take a look at org.jkiss.utils package in general. There are a few other very generic classes like XMLBuilder, IntKeyMap and LongKeyMap.

PR #5695, opened 2019-04-10
Add unit tests for org.jkiss.utils.SecurityUtils

JohnLBergqvist:
Hi,

Further to my initial PRs for the `utils` module, I have written tests that cover the `org.jkiss.utils.SecurityUtils` class with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

Once again, these tests should help you detect any regressions caused by future code changes.

### DIFF
--- a/bundles/org.jkiss.utils/tests/org/jkiss/utils/SecurityUtilsTest.java
+++ b/bundles/org.jkiss.utils/tests/org/jkiss/utils/SecurityUtilsTest.java
@@ -1,0 +1,119 @@
+package org.jkiss.utils;
+
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyInt;
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.isA;
+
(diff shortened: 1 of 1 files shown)


serge-rider:
These tests really make sense, thanks

serge-rider:
Unfortunately I had to disable latest tests. They were dependent on powermock and we can't use it.
Despite the fact that Powermock is evil by itself it doesn't work well on Java 11 (at least with default Maven settings). So I've commented all tests which depend on Powermock.
Maybe this can be done with base junit and Mockito?

JohnLBergqvist:
@serge-rider Unfortunately there is no way to mock the static classes used here such as `System`, `INetAddress`, `MessageDigest`, or to mock the constructors of `Random`/`SecureRandom`, with just Mockito, hence why Powermock was used. Is there a different mocking framework such as EasyMock or Jmockit, that you would consider instead?

serge-rider:
I can't say for sure as I'm not fluent in different testing frameworks. 
We could try it on Java 12 to see how it goes.

@tchibatron WDYT?

tchibatron:
Maybe just simple refactor this class? :)

PR #5699, opened 2019-04-11
dbeaver-corp/dbeaver-i18n#5

riednyko:
Added missing translation


serge-rider:
I've fixed the translation a bit

PR #5876, opened 2019-05-08
getMostPrecise CoordinateOperation

### LINKED ISSUE #5729: GIS: support different projections
Part of #4222
We need to support different projections (SRIDs)  besides 4326.

Examples: fmedata database tables most of which are in SRDID=26910 (North American Datum, Transverse_Mercator) which is not compatible with standard GPS coordinates.

Acceptance criteria: ability to see maps with SRIDs different that 4326 (default)

9und70:
fixes https://github.com/dbeaver/dbeaver/issues/5729#issuecomment-489536801
use most precise CoordinateOperation, not all


serge-rider:
Thanks for fixing this!
I didn't had a chance to test against multi-operational CRSes.

PR #5923, opened 2019-05-17
# 5902 tuned bug in SQL construction for sequence renaming in postgre…

ocrisurbainski:
…sql database


serge-rider:
Great, thanks!

Issue #5632, opened 2019-04-02
Wrong DDL when dropping index on SQL Server


davidjb:
#### System information: 
- Operating system (distribution) and version ``macOS 10.14``
- DBeaver version ``6.0.1.201903251041``
- Java version `OpenJDK 12.33`
- Additional extensions `None`

#### Connection specification:
- Database name and version ``Microsoft SQL Server 11.00.2218``
- Driver name:  ``Microsoft JDBC Driver 7.1 for SQL Server 7.1.4.0``
- Do you use tunnels or proxies (SSH, SOCKS, etc)?  ``No``

#### Describe the problem you're observing:

The DROP INDEX syntax produced by DBeaver for this database looks like:

```sql
DROP INDEX schema_name.index_identifier GO
```

(where schema and index_identifier are replaced accordingly).  Running this SQL leads to an error because the identifier cannot be found and given that identifier:

```
Cannot drop the index 'schema_name.index_identifier', because it does not exist or you do not have permission.
```

However, the documentation (https://docs.microsoft.com/en-us/sql/t-sql/statements/drop-index-transact-sql?view=sql-server-2017) for `DROP INDEX` indicates that it should be like so for compatibility with SQL Server, Azure, etc:

```sql
DROP INDEX index_name ON schema_name.table_name
```

The current syntax _appears_ to be following the deprecated, backwards-compatible syntax, but failing since it is using `schema_name.index_name` isn't a valid identifier, as opposed to `schema_name.table_name.index_name` which would work.  In any case, the current syntax is both broken and in need of updating for SQL Server/Azure compatibility.

#### Steps to reproduce, if exist:

1. Go to table with index (or create if necessary)
2. Right-click index in the Database navigator and choose `Delete`
3. Click `Save` to commit
4. Observe not-found error as above and below
5. Run the latter `DROP INDEX ... ON` DDL and note it works

#### Include any warning/errors/backtraces from the logs

```
Cannot drop the index 'schema_name.index_identifier', because it does not exist or you do not have permission.
```

djong0:
I got the same problem.

![image](https://user-images.githubusercontent.com/3746553/56343339-0de89300-61ed-11e9-9bc9-99a7e4c277c6.png)

uslss:
verified

PR #6004, opened 2019-05-26
Fixed drop index ddl #5632

riednyko:


serge-rider:
Please fix the query construction. Use getFullyQualifiedName of table (you can't rely on toString).

serge-rider:
Just noticed that you have fixed base model class instead of SQLServerIndexManager.

This is database-specific feature, it has to be changed in database-specific manager.

PR #6006, opened 2019-05-27
fix Korean typo

liumin38:
'Edit data' is translated to 데이터 수정, not 데티터 수정.
You can verify this with google translator.


serge-rider:
Thanks!

Issue #5921, opened 2019-05-16
[Oracle] Add DDL tab to index view


p91paul:
Just like there is a DDL tab in the table view, it would be useful in the index view. It can be obtained with this query:

    SELECT DBMS_METADATA.GET_DDL('INDEX','IDX_BROD_ISO_PAYM_IDU','BRODOGG_OWN') TXT FROM DUAL

Currently index DDL is visible in table metadata, but only if DBMS_METADATA is not accessible and DDL is generated by dbeaver. Info from DBMS_METADATA is much more useful if available, and anyway having DDL in the index view is valuable by itself.

Table:
![immagine](https://user-images.githubusercontent.com/2157006/57861745-7714f380-77f7-11e9-9e33-056b1ac24d49.png)

Index:
![immagine](https://user-images.githubusercontent.com/2157006/57861788-898f2d00-77f7-11e9-9ca4-329675b6ff5c.png)

uslss:
verified

PR #6028, opened 2019-05-30
Index view DDL tab #5921

riednyko:


serge-rider:
Interface OracleScriptObject  seems to be unused, please remove it

Issue #5558, opened 2019-03-22
";" is duplicate when creating view


uslss:
#### System information: 
- Operating system (distribution) and version win 10 x64
- DBeaver version 6.0.1

#### Describe the problem you're observing:
[View creation] If query is ended with ";", it is duplicated when trying to save created view
![image](https://user-images.githubusercontent.com/31996417/54850042-0f5e9280-4cf7-11e9-9b8f-69d466dc4e82.png)

uslss:
verified

PR #6036, opened 2019-06-02
Fixed duplicate delimiter #5558

riednyko:


serge-rider:
You can't use -1.
Delimiter may be multicharacter, you need to use exact delimiter length

PR #6054, opened 2019-06-05
Excel column autosize

AndrewKhitrin:
Column autosize now option default value is off for perfomance reason


AndrewKhitrin:
No need option

Issue #5810, opened 2019-04-26
SAP HANA - Support for Trigger


correiadn:
#### System information: 
- Operating system (distribution) and version: Windows 10
- DBeaver version: 6.0.3
- Java version: 1.8.0_192
- Additional extensions: No

#### Connection specification:
- Database name and version: SAP HANA
- Driver name: SAP_HANA_CLIENT (ngdbc.jar)
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No

#### Describe the problem you're observing:
Will Trigger be supported for SAP HANA DB as new feature?

#### Steps to reproduce, if exist:
N/A

#### Include any warning/errors/backtraces from the logs
N/A

correiadn:
@uslss Please find below the code to show the triggers on Hana DB:
https://github.com/correiadn/dbeaver/commit/d4286b026637242794fc6e8f6a864bde32f8679c
I tested here with success, is it possible to include it in the next release?
Thanks in advance.

serge-rider:
This looks good 👍 
Could you make a pull request? 
I can copy-paste the code but PR would be much better.

correiadn:
@serge-rider done.
Thanks to consider this change it will be very helpful to me and to others as well. :)

uslss:
closed without testing

PR #6079, opened 2019-06-11
Issue 5810 - Show Triggers on SAP HANA DB- Tested

correiadn:
https://github.com/dbeaver/dbeaver/issues/5810
Could you please include in next release? I already tested and it seems to be working fine on SAP HANA DB.
Thanks in advance


serge-rider:
👍 
Thanks for the contribution

Issue #6118, opened 2019-06-18
Please support EnterpriseDB (EDB) derivative of PostgreSQL


rbellamy:
**Is your feature request related to a problem? Please describe.**
Please support EnterpriseDB (EDB) derivative of PostgreSQL

**Describe the solution you'd like**
EnterpriseDB has a derivative of PostgreSQL which includes the following challenges when working with ANY database tool OTHER THAN pgAdmin4, and specifically for DBeaver:

1. Custom JDBC connection URL - in DBeaver you MUST create customer driver based on the "Generic" type. In order to leverage the already excellent PostgreSQL support, a new model with a custom JDBC connection URL is needed.
2. EnterpriseDB has packages, PostgreSQL does not.
3. EnterpriseDB has complex types that PostgreSQL does not, including collection types (`TABLE` and `ARRAY`) and composites (`OBJECT` types with fields of type `TABLE`).
4. EnterpriseDB has a custom implementation of the Oracle
PL/SQL language implemented as an overlay to pgSQL called SPL.

**Describe alternatives you've considered**
I tried creating a custom driver definition of type "PostgreSQL" with a URL Template of `jdbc:edb://{host}:{port}/{database}` and using the `com.edb.Driver` JDBC driver class. DBeaver throws an exception, which turns out to be because it ends up using `jdbc:postgresql://{host}:{port}/{database}` instead.

**Additional context**
I understand that items 2 through 4 above are a significantly heavy ask. However, using the PostgreSQL JDBC driver with EDB is not a viable option, as it has additional functionality which can cause unexpected behavior for "normal" PostgreSQL operations. DBeaver's PostgreSQL support is already robust and useful, and if I want to be able to use the EDB JDBC driver with DBeaver, I see no other alternative than to provide a custom `PostgreEdbModel` implementation.

garyshell:
This would be VERY useful to me.  Our current testing doesn't expose editing of functions and procedures.  Beyond that support of packages would be appreciated as well.

uslss:
Done in the scope of #6135

PR #6119, opened 2019-06-18
Initial commit to support EnterpriseDB - #6118

rbellamy:
EnterpriseDB uses a custom JDBC connection URL - #6118 
`jdbc:edb://<HOST>:<PORT>/<DB>`

The current Driver Settings do not allow for "inheriting" from the PostgreSQL standard with a new driver and JDBC connection string. Thus, in order to support EnterpriseDB and their proprietary driver, DBeaver must be made aware of this via a custom model implementation and plugin configuration.

TODO:

1. Icons are all just copies of the edb_logo.png file, which was scraped from the EnterpriseDB site. They need to be replaced with files that provide the resolution and pixel density they claim.
2. Support EnterpriseDB packages.
3. Support EnterpriseDB complex types.
4. Support EnterpriseDB SPL - a custom implementation of the Oracle PL/SQL language implemented as an overlay to pgSQL.


serge-rider:
Looks great! Thhank you, I had a few feature requests about EDB support but not much expertise in it.
Specifically EDB packages is one of the most wanter features.

It is not really clear what is the problem with URL. You can just make it imilar to Redshift (different driver, different URL but it inherits from standard PG driver). Looks like you already have this in your PR..

There are a few misconfigs in driver declaration i'll fix them

serge-rider:
Fixed redundant driver configs.
Pease see #6135

Firt problem is that EDB driver can't be loaded from public Maven central report.
Do you know where it can be downloaded (directly, maybe some proprietary Maven report)?

PR #6143, opened 2019-06-22
update ja properties

monomonomoe:
Hello

I am Japanese.

Updated `CoreResources_ja.properties` because there was a language that is not appropriate for Japanese.
Please merge if there is no problem

sorry. This sentence is machine translated.


serge-rider:
Thank you for the fix.
Unfortunately we can't validate Japanese translation by ourselves. If you will see any other issue with translatin please create a PR or simply tell me (by creating  a ticket) what exactly should be replaces with what.
Best regards.

PR #6250, opened 2019-07-09
set GIS viewer for GEOMETRY data type

Sargul:
* set GIS viewer for GEOMETRY data type
* enhance performance when reading table meta data


Sargul:
all changes apply to exasol

PR #6265, opened 2019-07-11
Add PrestoSQL driver config (prestosql.io)

guyco33:
Solves issue https://github.com/dbeaver/dbeaver/issues/6270

New UUID type is supported by PrestoSQL (https://prestosql.io/docs/current/language/types.html) and not by facebook prestoDB (http://prestodb.github.io/docs/current/language/types.html)

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
@@ -876,10 +876,27 @@
                     class="com.facebook.presto.jdbc.PrestoDriver"
                     sampleURL="jdbc:presto://{host}:{port}/{database}"
                     defaultPort="8080"
-                    description="PrestoDB JDBC driver"
+                    description="Facebook Presto (prestodb) JDBC driver"
                     webURL="https://prestodb.io/"
                     categories="sql,analytic,bigdata">
                     <file type="jar" path="maven:/com.facebook.presto:presto-jdbc:RELEASE"/>
+
+                    <parameter name="use-search-string-escape" value="true"/>
+                </driver>
+
+                <!-- PrestoSQL -->
+                <driver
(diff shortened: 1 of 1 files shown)


serge-rider:
Looks good 👍 
However we need some meaningful driver(s) description otherwise it is not clear why we will have two drivers for the same DB. 
E.g. Community vs Facebook made.

guyco33:
Make sense. Thanks @serge-rider for your comment.
I will add the following description:
For PrestoDB: `Facebook Presto (prestodb) JDBC driver` 
For PrestoSQL: `Community Presto (prestosql) JDBC driver`

findepi:
@guyco33 "Community Presto" is an unofficial term I observed it is sometimes misleading (as in "enterprise vs community", which is _not_ the case here).
Maybe call it just "Presto".

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -876,14 +876,31 @@
...
+                <!-- PrestoSQL -->
+                <driver
+                        id="prestosql_jdbc"
+                        label="PrestoSQL"
+                        icon="icons/presto_icon.png"
+                        iconBig="icons/presto_icon_big.png"
+                        class="io.prestosql.jdbc.PrestoDriver"
+                        sampleURL="jdbc:presto://{host}:{port}/{database}"
+                        defaultPort="8080"
+                        description="Community Presto (prestosql) JDBC driver"
```
findepi:
```suggestion
                        description="Presto (prestosql) JDBC driver"
```

guyco33: Agree. Community in this context may be misleading

serge-rider:
👍

PR #6274, opened 2019-07-12
Update to use information_schema to get db names

brerenat:
Using the information_schema to get the db names mitigates some issues users have with the show databases query which requires more privileges to run


serge-rider:
Some MySQL forks don't support information_schema or support it poorly (e.g. some schemas are missing there). While show databases is the most standard way to read schema list.

We used to use select from information_schema.schemata but then changed it to the current behavior.
Maybe this should be done configurable?

Issue #6363, opened 2019-07-28
Improve "sql to java code" feature


stalent:
Hi there,
When we use copy sql as java code feature in "Choose Format" panel, Java code results something like this:

```
"select\n" + 
"*\n" + 
"from\n" + 
"table;\n";
```

May we improve this feature with StringBuilder instead of using "+"? 

```
query.append("select");
query.append("*");
query.append("from");
query.append("table;")
```

Or may we add  customization options in this panel?

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/convert/impl/JavaSQLConverter.java

```
for (int i = 0; i < sourceLines.length; i++) {
            String line = sourceLines[i];
            result.append('"').append(CommonUtils.escapeJavaString(line)).append(lineDelimiter).append('"');
            if (i < sourceLines.length - 1) {
                result.append(" + \n");
            } else {
                result.append(";");
            }
        }
 
```

serge-rider:
This can be improved by adding extra config options (in additions to existing two options). E.g. "Use StringBuilder".

loro2:
Hi there,
I'm looking forward to contribute some code to this project, and this issue one seems like an ideal starting point. I'd be happy to take it.

uslss:
verified
![image](https://user-images.githubusercontent.com/31996417/62368701-0c41a180-b536-11e9-9322-88e280ea559f.png)

stalent:
Hi there,
Firstly, Thanks for the job.
May we remove the new line character "\n" from each line? You know as a developer, when we use stringbuilder for create query, it is not necessary using new line character. Besides it has '\n' character, it will fail to execute directly and we will have to remove it again.
Don't you agree with?

loro2:
You can set the "line delimiter" value to blank and it will remove the '\n' characters.

PR #6410, opened 2019-08-02
Implementation of #6363 - add StringBuilder option to Copy as Source panel for Java

loro2:
Implementation of #6363 

Quite straightforward implementation. I decided to change the method signature of convertSourceLines() in SourceCodeSqlConverter.java - this way makes the Copy as Source panel easily extensible with language-specific options.


serge-rider:
Looks good, thank you!
We'll test it and probably will include this in 6.1.4 release.

PR #6505, opened 2019-08-12
An indefinite article helps the readability of this sentence.

annawinkler:
I upgraded my DBeaver today, and noticed this sentence could use the indefinite article `a` for readability.


serge-rider:
Agreed )

Issue #6345, opened 2019-07-25
remember tree hierarchy in db navigator


rdp:
I'm always frustrated when I have to traverse a hierarchy 4 levels deep to get to Postgres DB's (server -> db name -> schemas -> public -> Tables)

I wish it would remember the "Tree opened" on the left for me so I don't have to reopen it every time I pull open dbeaver.

Another option to avoid the "very tiered tree" is to have the user specify the database as part of the connection, and make schema a "dropdown" then it could display the tables immediately.  (psequel does this)
<img width="762" alt="psequel" src="https://user-images.githubusercontent.com/12496/61836831-50022000-ae3f-11e9-90d6-42919587bbef.png">

rdp:
It does remember which tabs were open which is nice (already) :)

Spoelle:
I also find this frustrating, always open the tree. Nice future request !

loro2:
I created a PR for this FR. Though you might want to look into the current option 'Window -> Preferences -> Database -> Navigator -> Expand navigator tree on connect'. It expands your tree at least some levels when you connect to a DB (which can be automatic on startup).

serge-rider:
I've improved it a bit to avoid too long program startup. Now nav views are expanded (with progress) in detached mode only after UI was initialized.

kseniiaguzeeva:
Verified

rdp:
Is there a preference setting to turn this on? I tried git master :
database -> navigator -> "Restore navigator state up to depth: 10" and it
seems to still just auto expand to the database name level only [Postgres
in this case]?  Thanks!

serge-rider:
No other pref settings, it should work.
You could try it in the EA version (https://dbeaver.io/files/ea).
Or maybe you have special circumstances/config/filters. Please provide more details.

rdp:
Hmm seems to be working fine this morning, so thank you!

JGKle:
Odd, I'm not seeing this - found this issue when searching for the same, as a first-time user, one of the first things I noticed is that it forgets the state of the tree in Database Navigator each & every time I quit.  Using 21.1.5 on Linux. Also tried the settings mentioned above, but it never restores the tree state.

JGKle:
...After messing with this for awhile, it seems like it only restores the tree if there are no tabs opened. So either you can restore tabs or the tree, but not both...?

alejoar:
Same here, tree resets on app restart (tabs remain). On linux.

PR #6590, opened 2019-08-22
Implementation of #6345 - remember tree hierarchy in db navigator

loro2:
Implementation of #6345.

The code might be a bit complicated, though I tried to make it as de-coupled from the rest of the codebase as possible. The feature is hidden behind an option, and using the default value means no change in behavior and very little additional code execution.


serge-rider:
Looks like a good feature.
I will test it in a branch and probably will include in one of the near version releases.
Thanks!

PR #6636, opened 2019-08-27
Add support for Open Distro Elasticsearch SQL

dansimpson:
There are a few elasticsearch SQL plugins.  Elasic has their X-Pack driver, which requires a license to use.  Open Distro offers an open version which has similar features.  The JDBC driver is a bit different, and this PR simply adds support for that driver.


serge-rider:
This looks fine. It would be great to have OS version of elasticsearch driver.
A couple of minor things: please name plugin `org.jkiss.dbeaver.ext.opendistro` (without ui) and set bundle version (in Manifest.mf and in pom.xml) to `1.0.0`.

dansimpson:
@serge-rider Can you explain what you mean by 

> OS version of elasticsearch driver.

I'll fix up the namespace and version.

serge-rider:
Thanks!

PR #6660, opened 2019-08-30
Add support for Kognitio

elocemearg:
Add a new connection template for Kognitio, along with icons and default connection settings for the Kognitio JDBC driver.
Also add a Kognitio plugin which provides table and view DDL.


serge-rider:
Looks great 👍 
One thing: please move driver definition  and icons to the new plugin (from generic plugin). This will resolve conflicts and make everything less coupled.
For sample see how Firebird extension is formed.

serge-rider:
Thanks!

Issue #6630, opened 2019-08-27
Enquoting selected text deletes it instead


david-cortes:
**Steps to reproduce:**
* Open an SQL editor.
* Write some select statement.
* Double-click some non-quoted word from the select statement.
* Press quote character.

**Expected behavior:** should enclose in quotes the text that was selected.

**Actual behavior:** replaces the text with empty quotes.
![dbeaver_enquote](https://user-images.githubusercontent.com/11860098/63765055-ec9d5f00-c8d0-11e9-9f08-161abb5b0a31.gif)

serge-rider:
It is not a bug, most of text editors behave this way.
But this could be a good feature.

JuanLH:
Hi, which branch I have to clone to get that case?

david-cortes:
> Hi, which branch I have to clone to get that case?

Happens on the release version, so I guess all of them would have this behavior.

p91paul:
@JuanLH AFAIK, most work is done on the devel branch

JuanLH:
Hi, I'm trying to run the proyect for find the bug but i'm facing errors for missing dependencies, i  guess.
I did the following:

- Clone the repo  [devel] branch.
- Added existing maven proyect and i searched the git repo downloaded.
- Updated maven proyect.

And i have this
![Error for dependencies](https://user-images.githubusercontent.com/10179221/64491507-3d994580-d237-11e9-849d-971c6878c7a0.png)

JuanLH:
Ok, I found [https://github.com/dbeaver/dbeaver/wiki/Develop-in-Eclipse](url) I fix that error, I was the plugins

JuanLH:
I don't have the same behavior in the SQL editor, look:
![enquoting dbeaver 2](https://user-images.githubusercontent.com/10179221/64493411-ef437100-d24d-11e9-928b-30514e3a0771.gif)
I cloned the devel branch. 

And I don't have options for connections, maybe I have to include some data, right?
![connections](https://user-images.githubusercontent.com/10179221/64493457-71339a00-d24e-11e9-8569-1f6ecc1b40c1.gif)

david-cortes:
> I don't have the same behavior in the SQL editor, look:
> ![enquoting dbeaver 2](https://user-images.githubusercontent.com/10179221/64493411-ef437100-d24d-11e9-928b-30514e3a0771.gif)
> I cloned the devel branch.
> 
> And I don't have options for connections, maybe I have to include some data, right?
[...]

PR #6714, opened 2019-09-10
#6630 Enquoting selected text deletes it instead

vivekworks:
Added a private method to get the selection text from the editor and appended it along with the document when a single or double quote character is entered.


serge-rider:
Looks good 👍 
Thanks for contribution

vivekworks:
> Looks good 👍
> Thanks for contribution

Welcome

Issue #5453, opened 2019-03-11
Oracle package compile doesn't show compile errors


yaoguoba:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 5.3.5+
- Java version 1.8
- Additional extensions

#### Connection specification:
- Database name and version Oracle 12c
- Driver name ojdbc8
- Do you use tunnels or proxies (SSH, SOCKS, etc)?  No

#### Describe the problem you're observing:
Do not show any compile error or compile log when I save&persist the package. Compile always success.

#### Steps to reproduce, if exist:
~5.3.4 is OK
![image](https://user-images.githubusercontent.com/18393410/54095892-62c20e00-43e4-11e9-9880-1ef7f15fffb7.png)

5.3.5&6.0.0
![image](https://user-images.githubusercontent.com/18393410/54095814-124ab080-43e4-11e9-8548-0f395dc2d2e1.png)

serge-rider:
Since 5.3.5 we don't compile packages along with package save. 
So saving package do not validate package contents.

You need to perform compile separately (ctrl+f9 or click on Compile in the context menu)

yaoguoba:
Push ctrl+f9, also success.

p91paul:
When you say "Since 5.3.5 we don't compile packages along with package save.", is this by design or is it a bug that is going to be fixed? If the former, may I ask why is that?

adeorato:
> Since 5.3.5 we don't compile packages along with package save.

As far as I can see (SQL-Log-Viewer), this is not correct. In the "old world" the commands
```
ALTER PACKAGE PKG_FSRI_DATA_LOAD COMPILE;
ALTER PACKAGE PKG_FSRI_DATA_LOAD COMPILE BODY;
```
and the errors are displayed. Now DBeaver only calls the first compilation (no more body). The first command always returns "true" and the errors have to be checked explicitly (show errors). It seems that this does not (anymore) happen. 

But the display of "Compiled successfully" (screenshot) leads to two possibilities:
* It is compiled and errors are ignored.
* The message is wrong

Either way, it's a critical behaviour.

DragonStorm97:
I Also get the same (agreed, Critical) issue. it looks like a `ALTER PACKAGE SOME_PACKAGE COMPILE BODY` is never called. so it only compiles the spec and not the body, then when going to call the package it fails as it tries to actually compile the body.

this seems to be an easy fix of just adding the compile body clause.
[...]

PR #6752, opened 2019-09-16
#5453: compile Oracle package body and throw errors on save

p91paul:
Commenting line 175 is necessary because `sourceDefinition` is always `null` after saving a package, because after save and before compile the `refreshObject` method is called.

Alternatives to commenting line 175 are: 
- passing a `DBRProgressMonitor` object to `getCompileActions`, to allow calling `getExtendedDefinitionText`
- avoid calling `refreshObject`, or avoid setting `sourceDefinition` to null inside `refreshObject`, but this can lead to stale objects.

However, does a package without body even make sense? Stated otherwise, if you create a package without a body, and you get an error for that, you are unlikely to get angry.


serge-rider:
As far as I know it is very possible that package doesn't have a body (e.g. a package which contains only data types declaration). And this will be a valid package.

Anyhow, I've add progress monitor to compileActions method.

p91paul:
Closing as I rebased and force pushed and github seemingly won't update.

PR #6765, opened 2019-09-17
Compile Oracle package body and throw errors on save (#5453)

p91paul:
I'm unsure about error handling here. Currently I just logged at warning level, however the unability to read package source might be considered an error, and if so it should throw an exception to the user.

With current code the user will see a "package compiled successfully" message while that might not be true. However I'm inclined to leave this to you since I'm not very familiar with proper error handling in dbeaver.


serge-rider:
Let's try it.
Thank you.

PR #6780, opened 2019-09-19
added HANA plugin features:

kai-morich:
- driver download from maven repository
- search views,tables,column,procedures
- explain
- enable synonyms
- error position
- set application as client property
- show 'Unit' property for SYS view columns


kai-morich:
should I increase the version number in the plugin pom file?

serge-rider:
Wow, looks awesome 👍 
I'll make some testing, probably we'll include these HANA improvements in version 6.2.2.
Regarding version number - you can increase major version to 1.1 (in both manifes.mf and pom.xml) We increase minor versions automatically on each release.

Also - you could/should add yourself in the copyright section of source files. Feel free to modify this PR or to make a new one.

serge-rider:
Thanks. We'll track HANA improvements in #6791

PR #6857, opened 2019-09-28
Hana02

kai-morich:
more HANA enhancements


serge-rider:
I think you should keep driver definition in Generic plugin. It will be replaced by new driver if hana extension installed.

kai-morich:
ok, kept driver as 'HANA (old)'

serge-rider:
👍

PR #6863, opened 2019-09-29
add author and remark into firebird plan sources

tomas303:
Hello,
I added author and small info based on other examples I found in sources. I hope it is ok. If not, you can change it or give me a suggestion.
regards
Tomas


serge-rider:
👍

Issue #5164, opened 2019-01-31
SSH tunnel using keyboard-interactive AND publickey auth possible?


ceandreasen:
#### System information: 
- Ubuntu 18.04.1 LTS
- 5.3.3
- openjdk 1.8.0_191
- No extensions

#### Connection specification:
- Postgres 10
- PostgreSQL jdbc driver
- SSH tunnel 

#### Describe the problem you're observing:
I've recently changed how ssh is set up on my "tunnel" host to use keyboard-interactive auth in addition to public key authentication (A 2fa solution.) Now, I don't seem to be able to authenticate the ssh tunnel with dbeaver, it simply says auth fail. (I am using jsch, haven't installed anything else for ssh)

Am I correct in surmising that because of [this line](https://github.com/dbeaver/dbeaver/blob/0dde5c3459c5ebfd9e7c9bd3a03a440d5172369c/plugins/org.jkiss.dbeaver.net.ssh.jsch/src/org/jkiss/dbeaver/model/net/ssh/SSHImplementationJsch.java#L62) there is no way to use keyboard-interactive and public-key auth together? Perhaps this is a rare setup, but I don't think it's especially exotic. Would any special changes be necessary to enable this setup, or would just adding keyboard-interactive to the first string there pretty much do it?

Thanks,
Clark

ceandreasen:
I've tested changing the line mentioned above to say "publickey,keyboard-interactive", and it seems to work fine. I'm not well versed in contributing on github but I'll try and get a PR together when I have a chance.

serge-rider:
I'll push the fix by myself, thanks for investigating this!

FBlass:
Hi,
following up on this problem: In my case the server first asks for publickey authentication. After successful publickey authentication the server only asks for password, which is a short lived passcode.

ssh -v log: 
debug1: Authentications that can continue: publickey
debug1: Next authentication method: publickey
debug1: Offering public key: ###
debug1: Server accepts key: ###
Authenticated with partial success.
debug1: Authentications that can continue: password
debug1: Next authentication method: password

I am only asked for the 2factor passcode when adding 'password' to the PreferredAuthentications aswell.

PR #6892, opened 2019-10-01
SSH Auth: Add password as possible 2Factor AuthN when using Publickey AuthN

FBlass:
Follow up of Issue #5164

Server might also ask for 'password' as second auth, not only 'keyboard-interactive'.


serge-rider:
Thanks for PR 👍

PR #6944, opened 2019-10-08
add colspan so title line is above all columns not just first

jcszymansk:
In the HTML output from "Copy HTML" table title - query -  is only above the first column, usually some kind of ID. This change adds colspan attribute, so table title spans the entire table width.


serge-rider:
👍

Issue #6481, opened 2019-08-08
DBeaver should have a way to do PostgreSQL EXPLAIN without ANALYSE


rafafr:
DBeaver should have an option to disable the "ANALISE" option when doing the internal Execution plan on PostgreSQL databases.

When DBeaver does an EXPLAIN on a PostgreSQL database, the command sent by default to the server is "EXPLAIN (FORMAT XML, ANALYSE) <query>". This way the query is actually executed and we have more precise run time statistics.

For small queries it may be OK to run EXPLAIN this way, but during the development of big, slow queries, this can result in some not good huge queries send executed if the developer press "CTRL+SHIFT+E" by mistake.

For example, I had to do an UPDATE of about 2000 rows over a big table (almost 100 million rows) with JOINs to another two tables to locate the correct rows. I wanted to see the explain plan before runnnig the UPDATE to look for errors. There was an error on my query generating a CROSS JOIN of some trillion rows and the Explain Execution Plan locked my connection for about one minute until I stopped the execution. After some frustrating tries and re-tries of the Explain Execution Plan the DBA asked what the **** I was doing running big UPDATEs just to cancel them.

If the command being explained does something that can't be ROLLBACKed, this EXPLAIN can cause problems too.

There should be an option to run the Explain Execution Plan "without run time" ou "with no detailed statistics" to avoid those problems.

serge-rider:
Good point. Perhaps we should show some kind of config dialog before running actual explain plan.
Having separate commmands or some unobvious preferences doesn't look like a good ui..

serge-rider:
Done

uslss:
verified

PR #6958, opened 2019-10-10
Remove ANALYZE from Explain Execution Plan for Postgres

### LINKED ISSUE #2307: Option for Explain exec plan in Postgres
"Explain execution plan" always do EXPLAIN **ANALYZE** ...  that executes the query during analyse, but when the query is rather big and complex and not yet optimized, it may take very much time to wait.
The additional option is needed here - do the ANALYZE or not.

bobobobo:
This fixes #2307


henkeberggren:
I would like if this PR get merged.

serge-rider:
We can't just remove ANALYZE. Many people rely on it.
So, to solve performance issue we need to make EXPLAIN configurable. 
Some extra UI is needed (e.g. options dialog with checkbox "Use ANALYZE").
Or something like that.

serge-rider:
This was resolved in #6481

Issue #4516, opened 2018-10-31
SSH Tunnel to server thru a jump server


dhavalkshah:
Hi,

I have setup where i have jump server in between
So it is like: Client --SSH--> Jump Server --SSH--> Server
In my mac i have used ProxyCommand in .ssh/config to create a ssh tunnel to the server
Thus my config file has server defined as:

```
Host jump
        HostName <JUMP IP>
        User dhaval
        IdentityFile <PATH TO KEY>
Host app
        HostName <APPIP>
        User dhaval
        IdentityFile <PATH TO KEY>
        ForwardAgent yes
        ProxyCommand ssh jump nc %h %p 2> /dev/null
```
Thus I am able to do a ssh to server using command `ssh app` 
 
Now the queries:
1. I am not able to use app as host name in Dbeaver, although it is mentioned in .ssh/config
2. When I try IP of jump server and test ssh tunnel in DBeaver it goes thru but if I use IP of server it doesn't go thru

Any pointers will be helpful

~DKS

ericmjalbert:
I'm having a similar problem as this. Is there any update or workaround for it?

svanscho:
Any update?

yutaro1985:
I also have a similar issue.
Is there any way to use ssh/config?

jgomo3:
Maybe there is a possible workaround if we could write the exact shell command to use for SSH connection, so in the example mentioned in this ticket, instead of filling the form we could opt to write: `ssh app` or `/usr/bin/ssh app`.

Or adding support for the `-J` flag in OpenSSH.

Meanwhile... what workaround would you suggest?

jgomo3:
As a feature request:

1. Add support for the -J flag in OpenSSH.
2. Allow a free form field to just write the ssh command. E.g. `ssh app`.

Meanwhile... what workaround would you suggest?

Nimeshka:
I have the same problem, and there's no way to connect. So I switched back to PgAdmin until there's a solution for this in Dbeaver. It works in PgAdmin.

serge-rider:
Probably this issue will be resolved by #7164
Thanks to @bsv798

uslss:
verified

Jormangeud:
OP wanted:
Client --SSH--> Jump Server --SSH--> Server.
Use case: ssh through a gateway into a virtual machine's sshd, and use port forward from there.

#4516 implements this:
Client --SSH--> Remote host --port forward--> Server.
[...]

PR #7164, opened 2019-11-05
Ability to set remote host and port in ssh tunnel configuration

bsv798:
This patch allows advanced ssh tunnel configuration:

- tunnel through jump server: `ssh -L 5432:pg.example.com:5432 jump.example.com`
- remote port specification (if DBMS is runnning on non-default port): `ssh -L 5432:127.0.0.1:15432 pg.example.com`
- bind address specification (if DBMS is listening on specific interface): `ssh -L 192.168.5.11:5432:192.168.5.11:5432 pg.example.com`

I have added several fields in SSHTunnelConfiguratorUI.java and set default values in case these fields are not set in SSHImplementationAbstract.java:

- Local host (bind address). Default value: 127.0.0.1
- Local port. Default value: any free port with number > 10000
- Remote host. Default value: value of local host field
- Remote port. Default value: JDBC driver default port

Also, resource translations to german, french and chinese was made via google translate and may require manual correction.


serge-rider:
Interesting. I'll try to test it.

serge-rider:
Customizing localhost makes much sense. No questions here.

But jump server is questionable.
We will have 3 host/port pairs:
1. SSH server
2. Jump server
3. Database server

I'm not quite understand how jump server works. Is it a kind of second tunnel between ssh server and database server?

Like this: Client -> SSH Server -> Jump Server -> Database Server.

p91paul:
Corporate security experts will never get tired of adding hops for people to jump through. Just add another layer of indirection they said, it will solve all your problems they said. :grin:

bsv798:
@serge-rider 
Consider following diagram:
servers-diagram
All three servers are in the same network and **manager** server is the only one than can be accessed through the internet. You do not have direct access to **db** server through the internet. So you will have to establish ssh connection to **manager** server first and then use it to "jump" to **db** server (hence the name) and forward port from there to your host.
Configuration in dbeaver that allows you to do so is like this:
dbeaver-config
And appropriate ssh command: `ssh -L superuser@222.223.224.225 5432:192.168.5.3:5432`.

"Jump server" is just a fancy name. Such advanced tunnel can be configured with local host/port and remote host/port pairs.
And, of course, if your DBMS is on the same host that ssh daemon is, then you do not have to specify remote host/port values in dbeaver (they will be configured automatically anyway).

bsv798:
Accidently closed PR. Sorry :)

serge-rider:
This makes sense. And sources look good, thanks for contribution.
I will probably update UI a bit (as there are too many different options) after the merge.

BTW I think this PR will resolve ticket #4516. At least partially.

serge-rider:
I've fixed a couple of bugs (NPE in parsing existing SSH config) and a couple of labels.
Unfortunately I can't fix labels in languages I don't know (de, fr, zh). Could you fix them?

`Remote port for tunnel. If set to <=0 then default port of JDBC driver will be acquired` is not exactly correct. It is database server port, not JDBC driver port.

bsv798:
> Unfortunately I can't fix labels in languages I don't know (de, fr, zh). Could you fix them?
No, I do not know these languages either.

yash3492:
@bsv798 
I have exactly same use-case like you described here. But for some reason it doesn't work with dbeaver Version 7.1.4.202008021649.

Below is command line version which is working for me:
```
ssh -L 54321:pg-rds.example.com:5432 deploy@jump-ec2-server.example.com
psql -h localhost -p 54321 -U deploy db_name
```
Within dbeaver SSH tunnel connection check is returning success, but thereafter DB connection test is timing out.

Many thanks in advance!

PR #7213, opened 2019-11-12
Fix a small typo in popup error message

rafael747:
"No databases fond on the server" -> "No databases found on the server"


serge-rider:
thanks)

PR #7217, opened 2019-11-13
WIP: Disconnect for folders

maratkalibek:


serge-rider:
We don't need to change DBNLocalFolder  here, as well as any other model classes.

Issue #7087, opened 2019-10-24
Update OSM tiles server


ebocher:
#### System information: 
- Operating system  :  Linux
- DBeaver version : 6.2.2 and greater
- Additional extensions : GIS viewer, H2GIS

#### Describe the problem you're observing:

The leaflet viewer uses the mapbox server to provide tiles. It seems that the tiles are not up to date. See the two screenshots one in dbeaver the other from openstreetmap.org

#### Steps to reproduce, if exist:

Run the following query on a H2GIS connection.

```sql
SELECT st_setsrid('LINESTRING (450735.8542908922 5407155.854499144, 450733.1367910038 5407156.567704182, 450735.8542908922 5407155.854499144)'::geometry, 32631) AS the_geom
```
and open the map viewer. You will see the diff.

#### Proposal

In org.jkiss.dbeaver.data.gis.view/web/view_template.html replace 

```
  let baseLayersObj = {
                'Street': L.tileLayer('https://api.tiles.mapbox.com/v4/{id}/{z}/{x}/{y}.png?access_token=pk.eyJ1IjoibWFwYm94IiwiYSI6ImNpejY4NXVycTA2emYycXBndHRqcmZ3N3gifQ.rJcFIG214AriISLbB6B5aw', {
                    maxZoom: 18,
                    attribution: 'Map data &copy; <a href="https://www.openstreetmap.org/" target="_new">OpenStreetMap</a> contributors ' +
                        '<a href="https://creativecommons.org/licenses/by-sa/2.0/" target="_new">CC-BY-SA</a>, ' +
                        'Imagery <a href="https://www.mapbox.com/" target="_new">Mapbox</a>',
                    id: 'mapbox.streets'
                }),
```
by 

```

let baseLayersObj = {
                'Street': L.tileLayer(''https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png'', {
                    maxZoom: 18,
                   attribution: '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors'
                }),

```
OSM
![osm](https://user-images.githubusercontent.com/1820572/67477479-73b14e00-f65a-11e9-9107-ea85f5171adc.png)

DBeaver
![dbeaver_osm](https://user-images.githubusercontent.com/1820572/67477480-73b14e00-f65a-11e9-8d74-b87a89dd7909.png)

serge-rider:
Good point.
Current template was originally taken from leaflet samples. 
[...]

PR #7293, opened 2019-11-26
Replace the mapbox tiles server by OSM tiles server

ebocher:
About #7087


serge-rider:
Thanks!
I'll test it and probably include the fix in the next version.

serge-rider:
I made it customizable. Reused just OSM URI.

PR #7322, opened 2019-11-30
Saving templates in UTF-8

### LINKED ISSUE #7294: SQL Editor templates UTF8 support
#### System information: 
- DBeaver version: 6.2.5
- Java version: openjdk version "11.0.5"

SQL Editor templates are not persisted using UTF-8 encoding.

1) save a new SQL Editor template, containing non-ASCII characters
2) use a template by typing its name in SQL Editor, followed by a TAB -> everything works fine
3) restart DBeaver
4) use a template again, the same way used in step 2) -> non-ASCII characters are not shown right

ArtyomPanfutov:
Fixes #7294


serge-rider:
Thanks!
This makes much sense.

PR #7351, opened 2019-12-03
Changing HashSet to LinkedHashSet for deterministic iterations 

contextshuffling:
Test `org/jkiss/utils/CommonUtilsTest.java` uses a `HashSet` to test the functionalities of `CommonUtils.getItem`, which uses the iterator of the argument. Currently, the test depends on the iteration order of the `HashSet` (i.e., assuming the second object is `b`). However, `HashSet` does not guarantee any specific order of entries. Tests can fail if the iteration order changes.

This PR proposes to fix the test by changing `HashSet` to `LinkedHashSet` that make the iteration order deterministic.


serge-rider:
👍

PR #7368, opened 2019-12-04
HANA spatial types support

stefanuhrig:
This change adds a HANA value handler for geometries so that geometries
can be viewed in DBeaver's integrated Leaflet panel.


stefanuhrig:
Hi,

It would be nice if this pull request could be merged. I'm happy to rework if something needs to be fixed.

Please pay special attention to my changes GISLeafletViewer.java. I moved the attribute SRID detection in front of the transformation loop so that the attribute's SRID can be used for the transformation.

Additionally, I added GSON type adapter for DBDContent. Without the adapter, GSON used the reflection to jsonize the instance, which lead to an infinite recursion.

I'm also not sure about the versioning conventions.

Oh, and I just see that Codacy already complained... I'll check.

serge-rider:
Looks great 👍 

I see that quite a lot of work has been done. As it is a big PR I'll need some time to review it and to check than nothing was broken in spatial rendering for other databases.

Anyhow there is a big chance that we'll include this PR in the next CE version. Probably with some modifications.

E.g. some Codacy complains make sense, for example "Fields should be declared at the top of the class, before any method declarations, constructors, initializers or inner classes." should be addressed.

stefanuhrig:
Thanks for the feedback. I'll work on the codacy complaints and will update the pull request.

serge-rider:
Thanks. I'll merge it right after 6.3 code freeze (next week).

Note: we are going to unify license headers in all dbeaver source files. Make it more "standard", similar to all Eclipse sources. Basic template is here: https://github.com/dbeaver/dbeaver/blob/devel/docs/license_header.txt

So main part and copyright notice will remain the same for all files.
Contributors section will contain all actual contributors. No need to add dates there as they can be get from Git history.
We will automatically update/convert all project files closer to 2020. But you can do this by yourself if you like.

stefanuhrig:
Thanks for the information. I'll update the headers and then push again.

PR #7380, opened 2019-12-06
Quote identifiers in HANAMetaModel#getTableDDL

stefanuhrig:
Identifiers must be quoted when calling get_object_definition.
Otherwise the DDL of tables with a name containing lower-case letters
(or other special characters) cannot be determined.


serge-rider:
Thanks for the contribution!

PR #7426, opened 2019-12-16
Oracle Sessions

SergDzh:
Add collumns sqlid and childnumber in list
image


serge-rider:
Thanks!

I'll test it right after 6.3.1 release.

PR #7427, opened 2019-12-16
Oracle Session New Tab

SergDzh:
Add in detail view new tab with Execute plan from dbms_xplan
image


serge-rider:
I've created a ticket for this PR (generally we always create tickets before PRs as all commit messages must have ticket number in the beginning)

PR #7429, opened 2019-12-16
Fix initial datasource loading to refresh metadata folder contents

datallah:
If the metadataFolder has been changed while DBeaver is closed, the changes will not be noticed without the refresh.


serge-rider:
Thanks!

I'll test it right after 6.3.1 release.

serge-rider:
Well, let's try it.
It shouldn't  make any harm, the only thing which concerns me is potential cyclic listeners (project refresh may cause datasources reloading which may cause project refresh). However this shouldn't happen in this case as project is not loaded at the moment of DataSourceRegistry instantiation.

Issue #7442, opened 2019-12-18
[PostgreSQL] Rename sequence is missing schema qualification


cpburnz:
#### System information: 

- OS: Ubuntu 18.04 64bit
- Java: OpenJDK 11.0.4
- DBeaver: 6.3.0.201912011827
- Additional extensions: No

#### Connection specification:

- DB: PostgreSQL 9.4.7
- Driver: PostgreSQL JDBC Driver
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No

#### Describe the problem you're observing:

When renaming a sequence either through the context menu or through the properties pane, the generated DDL is missing the schema.

This is different from the fixed "Unable to rename sequence in Postgresql" #5902 because it was using the new name for the source name.

#### Steps to reproduce, if exist:

Create a schema and sequence:

```sql
CREATE SCHEMA test_schema;

CREATE SEQUENCE test_schema.test_seq;
```

Open the sequence's properties pane, change its name to "test_seqqq", click "Save", and the Persist Changes dialog box will have the following DDL:

```sql 
ALTER SEQUENCE test_seq RENAME TO test_seqqq;
```

This SQL statement is missing the schema qualification. It should be:

```sql
ALTER SEQUENCE test_schema.test_seq RENAME TO test_seqqq;
```

#### Include any warning/errors/backtraces from the logs

When trying to persist the incorrect DDL the following error details is displayed:

```
SQL Error [42P01]: ERROR: relation "test_seq" does not exist
  ERROR: relation "test_seq" does not exist
```

helmd8d:
I observed the same behaviour for the RENAME VIEW statement. Returned error is "Object ... does not exist".
When adding schema name manually in generated SQL it is possible to execute the statement successfully. Dbeaver 2.6.5 community/Windows. Jdbc/Exasol DB.

uslss:
thanks for the bug report

serge-rider:
Fixed

uslss:
The error appears when trying to rename sequence

> ALTER SEQUENCE public.actor2_actor_id_seq RENAME TO public.actor21_actor_id_seq;

![image](https://user-images.githubusercontent.com/31996417/71590008-0927fd00-2b38-11ea-9136-561bc6a75c3e.png)

> ALTER SEQUENCE public.actor2_actor_id_seq RENAME TO actor21_actor_id_seq;
works as expected

cpburnz:
[...]

PR #7517, opened 2020-01-03
Rename view missing schema qualification

helmd8d:
Hi! I fixed the Rename View missing schema qualification problem for the Exasol plugin. Works fine for me now.
See my comment on https://github.com/dbeaver/dbeaver/issues/7442

Could anyone check my other commit 203fa59 regarding the Manifests-File. I was not able to compile it without removing the two lines.


serge-rider:
Thanks!

Please remove two other files from commit.
What exactly compile error do you have?

helmd8d:
> Please remove two other files from commit.

Done

> What exactly compile error do you have?

I got the error: Package 'org.jkiss.dbeaver.model.sql.parser.tokens' does not exist in this plug-in

serge-rider:
I believe I've already fixed the problem with manifest.
Thanks

PR #7601, opened 2020-01-10
Fixed tablespace option selection in the PostgreSQL database creation…

pekhterev:
When creating a new database, the selected table space is not used. This is due to the addition of an  default value to tablespaceCombo. ([here](https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/PostgreCreateDatabaseDialog.java#L179))


serge-rider:
Thanks!

pekhterev:
Thank you for great tool!

PR #7605, opened 2020-01-11
Fixed closing the database selection dialog for PostgreSQL

pekhterev:
Hi. I think this is a dirty solution, but I don’t know Java, so I can’t offer something better.

When you try to select another base for the connection, the dialog just closes, because the main dialogue loses focus. I highlighted the problem element in red.

dbeaver

This introduced by [commit](https://github.com/dbeaver/dbeaver/commit/e5a61f9c9242a14d0e762eb0b64a3e96f12793a5)

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/dialogs/SelectDatabaseDialog.java
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/dialogs/SelectDatabaseDialog.java
@@ -108,8 +108,6 @@ public class SelectDatabaseDialog extends SelectObjectDialog<DBNDatabaseNode>
                         objectList.loadData();
                     }
                 });
-
-                closeOnFocusLost(instanceCombo);
             }
         }
     }
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/dialogs/SelectObjectDialog.java
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/dialogs/SelectObjectDialog.java
@@ -220,8 +220,6 @@ public class SelectObjectDialog<T extends DBPObject> extends AbstractPopupPanel 
 
         objectList.loadData();
 
-        closeOnFocusLost(objectList.getItemsViewer().getControl(), objectList.getSearchTextControl());
-
         return group;
     }
 


serge-rider:
I can't reproduce the problem - at least on Windows or MacOS.
Probably it may happen on some linuxes because of different focus management.

Anyway, the solution could be to remove focus listener from database combo. But not from other controls. WDYT?

pekhterev:
I am using KDE. Other controls work as expected (In any case, I had no problems). For example, I do not have this problem when creating the database, but there is a modal dialog box. Eclipse controls themselves may be affected. I will try under Gnome and XFCE and let you know.

pekhterev:
> I can't reproduce the problem - at least on Windows or MacOS.
> Probably it may happen on some linuxes because of different focus management.

The results are sad, it does not work under Linux. I tried Ubuntu and CentOS with KDE, Gnome and XFCE. All of them have this issue.

> Anyway, the solution could be to remove focus listener from database combo. But not from other controls. WDYT?

I do not exactly understand what you mean by other elements. Dialog not closed only if i remove focus listener in both places.

serge-rider:
> I do not exactly understand what you mean by other elements. Dialog not closed only if i remove focus listener in both places.

You commented the following line:

> closeOnFocusLost(objectList.getItemsViewer().getControl(), objectList.getSearchTextControl());

Dialog won't close on focus lost - no matter what control was in focus.
It will fix one problem but will disable popup auto-close.

We need another solution then. E.g. don't treat combo dropdown as focus lost in general.

serge-rider:
Somewhere over here:
https://github.com/dbeaver/dbeaver/blob/d56fec40294cff90891b484526a0738fb139f5cc/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/dialogs/AbstractPopupPanel.java#L84

pekhterev:
> You commented the following line:
> 
> > closeOnFocusLost(objectList.getItemsViewer().getControl(), objectList.getSearchTextControl());
> 
> Dialog won't close on focus lost - no matter what control was in focus.
> It will fix one problem but will disable popup auto-close.
> 
this is strange, but the dialog closes if you don't remove the second line closeOnFocusLost(instanceCombo);
I tested it. In a modal dialog, everything works, for example, in the database creation dialog.

pekhterev:
> Somewhere over here:
> 
> https://github.com/dbeaver/dbeaver/blob/d56fec40294cff90891b484526a0738fb139f5cc/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/dialogs/AbstractPopupPanel.java#L84

For combo getShell() return parent window; getShell().getDisplay().getFocusControl() returns null

serge-rider:
Perhaps this problem was fixed in 6b2721b1fbdc56a058cbeb4c2242e747b42c57e9 - please check Early Access version (https://dbeaver.io/files/ea)

pekhterev:
> Perhaps this problem was fixed in [6b2721b](https://github.com/dbeaver/dbeaver/commit/6b2721b1fbdc56a058cbeb4c2242e747b42c57e9) - please check Early Access version (https://dbeaver.io/files/ea)

Fixed, thank you very much!!!

PR #7614, opened 2020-01-13
Hex Editor Colors

maratkalibek:


maratkalibek:
Added functionality to change following hex editor properties:
- text color
- caret line color
- text font

Issue #5014, opened 2019-01-15
DBeaver: Table grid NAVIGATE -> REFERENSES does not sort tables alphabetical


pdanie:
DBeaver - **Version 5.3.3 CE** - **EA 14-Jan-2019 21:17**
DBeaver driver: **MS SQL Server  / Microsoft Driver** _(the new Microsoft Driver)_
Operating System: **Windows 7 / Windows 8.1 / Windows 10**
Database Server: **Microsoft SQL Express 2014, 2016, 2017**

Unfortunately the **EA 15-Jan-2019 20:50** contained a small bug but a very annoying one. That the list showing **Table Foreign Keys** and **Table References** shown with **Navigate to Referenses** is broken when using the **New Microsoft Driver** in that the table names are not sorted alphabetical.

Both the table names in **Table Foreign Keys** and in **Table References** ought to be sorted.

***PS**. This is an extra fix for the work done in #5009 issue.

![2019-01-15 22-45-04](https://user-images.githubusercontent.com/31191374/51212189-14c6e780-1918-11e9-986a-1aba96ee4af2.png)

pdanie:
I have tried it in **EA 16-Jan-2019 07:28** and it seems to work **OK**. Great work.

PR #7719, opened 2020-01-22
Sort entries in References menu

minus7:
With a PostgreSQL database, the context menu for references to/from a
table did not list entries in a sorted order. #5014 seems to have fixed
that for MSSQL databases only. This commit sorts the entries in the UI
instead.


serge-rider:
This makes sense, 
thanks!

PR #7728, opened 2020-01-23
Fix session comment string

rduivenvoorde:
While looking into code how DBeaver managers to create Datetime types, I happen to see that this string is apparently a copy of the same class used by the Postgres driver.

It's trivial, I know, but thought to 'fix' it.


serge-rider:
Thanks!

PR #7843, opened 2020-02-05
Fix WKB parsing issue in HANA plugin

stefanuhrig:
MultiPoint WKBs are not parsed properly by the WKB parser in the HANA
plugin.

Instead of parsing each member of the multi-point as an independent
point, the parser interprets the following data as raw coordinates.

We had the same issue in geotools
(https://osgeo-org.atlassian.net/browse/GEOT-6509).


serge-rider:
Thanks for the fix 👍

PR #7849, opened 2020-02-05
Fix bad unicode for Spanish

albfernandez:


serge-rider:
Thank your for validating and fixing this!
Apparently some non-latin characters were broken during one of automatic source code update routines.

PR #7941, opened 2020-02-18
Fix issues with HANA geometry data transfer

stefanuhrig:
There were three issues with HANA geometry data transfers:
 - Binding a geometry would always throw
 - Polygons with more than one hole would have a faulty serialization
 - The SRID would not be added to the type


serge-rider:
👍

PR #7946, opened 2020-02-18
FIX: Cast for PostgreSQL reg* types to avoid error of uncompatible varchar and reg* & OID types refactoring

Vlczech:
Fix for following bug [PostgreSQL], which occurred during saving changes in table which contains for example regclass primary key as same as for example some other regclass column, which is not primary key.

regclass vs. varchar


serge-rider:
I like this feature.
Thanks!

serge-rider:
I'll reuse the same dialect function for type cats in data filter conditions generator.

Vlczech:
You're welcome. Use it as you need - it was purposely designed (and implemented) for the possibility of wider use (because other DB systems have for example no cast operators, but have the cast functions)  :-)

Vlczech:
BTW @serge-rider : Will be the 6.3.6 with these fixes? And when? (Maybe I will push another one.)

PR #7967, opened 2020-02-20
ERDiagram table nameLabel contrast color

rajter:
Enhances visibility of table name.
Sets the table name label color to a contrast color based on the background color of the table.

Preview:
Label color is the same regardless of the background color:
Digram-Before
Label color is determined based on the background color:
Digram-After


serge-rider:
This looks awesome.
Thank you!

rajter:
You're welcome.

Issue #4925, opened 2019-01-06
[feature] Support OmniSci (MapD) GPU SQL Analytics Engine


archenroot:
Hi,

I would like to use dbeaver for connections to OmniSci platform via JDBC. 

Is there any temporary workaround to inject just JDBC and try it or it will require all the data type mapping, DDL/DML language to be mapped into some general model of dbeaver?

Thx for response.

serge-rider:
No, only JDBC driver (jar file) + driver configuration is required. n most cases no any data type mappings are required.
If you will succeed in OmniSci configuration - please share driver configuration (URL, class ,etc). We'll include it in one of near DBeaver versions.

archenroot:
@serge-rider - ok, so connection is established:
![image](https://user-images.githubusercontent.com/553512/51456042-a2b23080-1d65-11e9-8a07-3a93f147bd3c.png)

archenroot:
And additionally I attach XML config:
```
<provider id="generic">
                <driver id="B7E68812-B745-F526-374A-D9084F050A77" categories="" custom="true" embedded="false" n$
                        <library type="jar" path="${home}/mapdjdbc-4.3.0.jar" custom="true"/>
                        <parameter name="supports-indexes" value="true"/>
                        <parameter name="omit-single-catalog" value="false"/>
                        <parameter name="all-objects-pattern" value="%"/>
                        <parameter name="quote-reserved-words" value="true"/>
                        <parameter name="omit-single-schema" value="false"/>
                        <parameter name="supports-scroll" value="false"/>
                        <parameter name="omit-catalog" value="false"/>
                        <parameter name="script-delimiter-redefiner" value=";"/>
                        <parameter name="supports-struct-cache" value="true"/>
                        <parameter name="supports-truncate" value="true"/>
                        <parameter name="script-delimiter" value=";"/>
                        <parameter name="legacy-sql-dialect" value="false"/>
                        <parameter name="script-delimiter-after-query" value="false"/>
                        <parameter name="use-search-string-escape" value="false"/>
                        <parameter name="omit-type-cache" value="false"/>
                        <parameter name="split-procedures-and-functions" value="false"/>
                        <parameter name="supports-multiple-results" value="false"/>
                        <parameter name="supports-stored-code" value="true"/>
                        <parameter name="script-delimiter-after-block" value="false"/>
                        <parameter name="supports-references" value="true"/>
                        <parameter name="ddl-drop-column-short" value="false"/>
[...]

PR #7972, opened 2020-02-21
Fix MapD/OmniSci plugin/configuration

randyzwitch:
(Background: I'm Senior Director of Community/Open Source at OmniSci)

Currently, the MapD configuration is out of date. MapD rebranded to OmniSci last year, and version 5 of the database has been released. This PR attempts to rebrand, add the maven jar location for the JDBC driver, and fix the port and connection details.

One thing that isn't accounted for is additional necessary jars. On creating a connection, the dialog box comes up asking if you want to download the additional jars referenced by `omnisci-jdbc`. If you would like me to add those jars to the configuration, so that users don't get hit with that message, I'm happy to.

It's also not clear to me right now whether any additional parameters need to be set per #4925. I have built dbeaver locally using `mvn package` and it seems to work fine, but would love any guidance here about setting other parameters.


randyzwitch:
Here are the jars I am talking about:

image

maheshwarigagan (approved):

serge-rider:
Hi, nice to meet you :)
Thanks for the PR. Most likely we'll include this configuration in the next version.

Extra jars are Maven artifact dependencies. DBeaver download them once along with main jar.
If dependencies are not really needed then you can add `ignore-dependencies="true"` parameter in artifact configuration.

I'll test this config and merge pr in next couple of days.

serge-rider:
Icons 16x16 and 32x32 are unreadable :(

randyzwitch:
@serge-rider unreadable in what way?

serge-rider:
It is impossible to read the text them because of icons bad quality.

https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ext.generic/icons/omnisci_icon%402x.png

Perhaps text can be replaced with some image?

randyzwitch:
Oh, okay. Yes, I'll submit a new PR with better icons

PR #7978, opened 2020-02-22
Fix: generated SQL using new cast() (needed for pgsql reg* columns)

Vlczech:
Fix for generated queries for example for these cases:
* context menu above table -> Generate SQL -> INSERT
* context menu above row(s) -> Advanced Copy -> Copy As SQL

_[build on / using #7946]_


serge-rider:
Please update your codebase, I've changed API a bit (cast->getTypeCastClause)

Vlczech:
@serge-rider: And did you already rename `cast()` or should I update also sources from #7946 ? At this moment I don't see any changes of `cast()`...

serge-rider:
It is already in the `devel` branch: https://github.com/dbeaver/dbeaver/blob/19c13f62f241da3cb024037ccbbfc4e0a610ebb6/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialect.java#L264

Vlczech:
@serge-rider: Sorry, I looked into branch `master`.. Code from this PR updated.

Issue #7988, opened 2020-02-24
Public Snapcraft config


serge-rider:
We need to put snapcraft.yaml in public repo.
As it is used for dbeaver-ce publishing and external PRs may be useful.

galgalesh:
:+1: for this. I was investigating why the `dbeaver-ce` snap starts so slow and I found a few issues.

* It includes a bunch of icon themes in the snap itself without caches. These caches are generated when the snap first starts up.
* It includes a bunch of libraries in the snap which are already available in the `gnome-3-28-1804` content snap. This increases startup time because the dynamic linker has a bunch more libraries to iterate over and the snap size is a lot bigger which increases decompression time.

I'm happy to fix these issues in the `snapcraft.yaml` file myself when it's online somewhere.

riednyko:
snapcraft.yaml now in the public repo:
https://github.com/dbeaver/dbeaver/blob/devel/product/snapcraft/snapcraft.yaml

Issue #7521, opened 2020-01-03
Access Denied: Cannot Read /usr/bin When Installed as Snap


dsmithhayes:
#### System information: 
- Elementary OS 5.1
- DBeaver 6.3.1

#### Connection specification:
- MariaDB
- MySQL

#### Describe your question:
I am not able to read my system's `/usr/bin` path within DBeaver when it is installed through the Snap packaging system. I have A MySQL and MariaDB client on my local machine in that directory and would like to use DBeaver to dump databases. Is there any insight some one can provide?

I apologize if this is a technical limitation of Snap and does not belong here.

CitizenBeta:
I can reproduce this on Ubuntu 19 when installed via "Ubuntu Software"

galgalesh:
@dsmithhayes Thanks for your bug report. I'm looking to include the native clients in the snap. Are both the MySQL _and_ MariaDB clients required or is it ok to only include the MariaDB client?

dsmithhayes:
@galgalesh Thank you, I am personally fine wit the MariaDB client, but I think it may be more valuable to include both.

serge-rider:
MariaDB client included in new snap build. Thanks to @galgalesh

uslss:
verified

gsevla:
Same with postgres.

dveaver have no permission to access /usr/bin.

galgalesh:
@gsevla Are you saying the dbeaver snap can't dump a postgres database?

gsevla:
> @gsevla Are you saying the dbeaver snap can't dump a postgres database?

I didn't try to dump pg db. I'm saying dbeaver snap can't connect to pg db, because it's runner is located in / usr / bin, where dbeaver snap have no permission to access.

I already tested dbeaver flatpak and it haven't access to /usr/bin too. But I can install postgres dbeaver io from flatpak and access pg bin from flatpak folders so it works. If dbeaver snap have something like it, I doesn't know.

galgalesh:
@gsevla Can you open a new issue for this? Specifically about you no being able to connect to postgres using the dbeaver snap. It's very unlikely dbeaver will ever get access to `/usr/bin` but all the required clients should be shipped inside of the snap.

neuberoliveira:
same here, installed via snap install dbeaver-ce trying to dump a mysql database with no success.

[...]

Issue #6704, opened 2019-09-09
On Ubuntu snap installation is not working


compalmanel:
I'm running Ubuntu 19.04. After using Ubuntu Software (a snap frontend) to install DBeaver, it fails to start with the following error:

> (process:8713): Gtk-WARNING **: Locale not supported by C library.
> 	Using the fallback 'C' locale.
> Failed to connect to Mir: Failed to connect to server socket: No such file or directory
> Unable to init server: Could not connect: Connection refused
> Dbeaver: Cannot open display: 
> Failed to connect to Mir: Failed to connect to server socket: No such file or directory
> Unable to init server: Could not connect: Connection refused
> Invalid Configuration Location:
> The configuration area at '/home/pmfa/.eclipse/1738816236_linux_gtk_x86_64/configuration' > could not be created.  Please choose a writable location using the '-configuration' command > > line option.

aurelijusrozenas:
Same on Ubuntu 18.10.
> (process:7538): Gtk-WARNING **: Locale not supported by C library.
	Using the fallback 'C' locale.
Dbeaver: Cannot open display: 
Invalid Configuration Location:
The configuration area at '/home/infro/.eclipse/1738816236_linux_gtk_x86_64/configuration' could not be created.  Please choose a writable location using the '-configuration' command line option.

bkolasa:
Exactly the same on Ubuntu 18.04.3 LTS

leandropaz:
Same issue on Ubuntu 19.10.

cristiancmello:
Same issue on Ubuntu 19.10.

gogosofta:
same here (ubuntu 19.10)

tayfunyasar:
same here. dbeaver isnt in latest version in snap store. it is still in 5.2.5

audacioustux:
same in manjaro kde
```
11:04:30:0140 Gtk Locale not supported by C library.

11:04:30:0141 Gtk Failed to load module "canberra-gtk-module"
No protocol specified
Unable to init server: Could not connect: Connection refused
11:04:30:0146 Gtk cannot open display: :0
```

serge-rider:
Fixed by @galgalesh PR

uslss:
verified

alacevedo:
Se agrega delay a la descarga de documentos

PR #8117, opened 2020-03-06
Switch snap to gnome-3-28 extension

### LINKED ISSUE #7548: Snap package doesn't have native client
#### System information: 
- Operating system (distribution) and version: **Ubuntu 18.04** (up-to-date)
- DBeaver version: **Version 6.3.2.202001051908**
- **Snap installation**

- Database name and version: **MySQL 5.6.10**
- Driver name: **MySQL Connector Java mysql-connector-java-5.1.44 ( Revision: b3cda4f864902ffdde495b9df93937c3e20009be )**
- Do you use tunnels or proxies (SSH, SOCKS, etc)? **No**

For dumping a database, the application needs a native MySQL client. But the snap package doesn't contain that. As intended, non-classic snap packages cannot execute something on the system. So the snap package should contain native clients too.

Try to dump something with the snap package.

merlijn-sebrechts:
[As promised](https://github.com/dbeaver/dbeaver/issues/7988); a big refactor of the snap.

* Switch to `base: core18`. So now the snap is using Ubuntu 18.04 as a base. This also makes it a lot easier to build the snap because it now automatically uses `multipass` to create a VM to build the snap in. Just run `snapcraft` in the `product/snapcraft` folder and **it will create an Ubuntu 18.04 VM to build the snap in**. You can now build the snap on any Linux version as long as you have the `snapcraft` and `multipass` snaps installed.
* Switch to gnome-3-28 extension. This removes a lot of boilerplate and pulls in the latest fixes to the `desktop-launch` script.
* Added "cleanup" part to remove duplicated files. This reduces the snap size from `208MB` to **`108MB`** and reduces the time until you see the spash screen on a cold start from 6 seconds to 3 seconds!
* Fallback to xwayland when running on wayland.
* Adopt metadata from AppStream appdata.xml file. Metadata like `name`, `summary`, `license`, the `.desktop` launcher and the icon are now extracted from the AppStream file.
* Get version from jar: the version number is extracted from the `jar` and added as metadata. So now the snap version shows the actual dbeaver version!
* Add MySQL and MariaDB clients to the snap. Fixes https://github.com/dbeaver/dbeaver/issues/7548 and https://github.com/dbeaver/dbeaver/issues/7521

This will probably also fix all of the issues in https://github.com/dbeaver/dbeaver/issues/6704


serge-rider:
Thanks!
We'll try to adapt it in 7.0.1 version.

PR #8152, opened 2020-03-10
FIX for granting on schema [fixes #8134]

### LINKED ISSUE #8134: [PostgreSQL] Granting schema generates grant table <schema_name>.<schema_name> query
#### System information: 
- Operating system (distribution) and version: Windows 10 x64
- DBeaver version: 7.0
- Additional extensions: -

- Database name and version: PgSQL 11.7
- Driver name: PgSQL JDBC

Open user under <some database> - Roles - <specific user>. Under permissions tab, select some schema and for example usage permissions and save it. It generates this BUGGED query:
GRANT USAGE ON **TABLE** test **.test** TO someuser;

instead of right query:
GRANT USAGE ON **SCHEMA test** TO someuser;

...see above...

obrazek

Vlczech:


serge-rider:
Thanks, this patch will be included in the next dbeaver version

PR #8175, opened 2020-03-12
SAP HANA Spatial: Patch rings with less than 4 points

stefanuhrig:
HANA Spatial accepts degenerated polygon rings with less than 4 points.
However, such polygons are rejected by the Java Topology Suite used by
DBeaver.

We patch those degenerated polygon rings now by adding the last point
multiple times.


serge-rider:
This patch will be included in the next dbeaver version

Issue #8216, opened 2020-03-18
Meta data queries executed on editor connection and not meta cnonection


Sargul:
#### System information: 
- Operating system (distribution) and version: Windows 10
- DBeaver version: DBeaver 6.3.5 and 7.0
- Additional extensions: Office Plugin

#### Connection specification:
- Database name and version: Exasol 6.1.6
- Driver name: Exasol
- Do you use tunnels or proxies (SSH, SOCKS, etc)? 
No

#### Describe the problem you're observing:
certain meta data queries for objects are executed on the editor connection for Exasol

#### Steps to reproduce, if exist: die 
```sql
SELECT * from "$ODBC"(CTRL+space to trigger autocomplete); -- don't execute it's just a lookup that can't return a result
flush statistics;
select * from exa_statistics.exa_dba_audit_sql where session_id = current_session and sql_text like '%table_schem%';
```
The select from the audit shows SQLs that are executed on the current session (Editor Window).
![image](https://user-images.githubusercontent.com/5595511/76954001-e70c7280-690f-11ea-8d80-d8243d3165bd.png)
As you can see the lookup for the object for the auto complete was done in editor session.

Looks like the session given to ExasolStructureAssistant in searchTables is not in the DBCExecutionPurpose.META context even though it set in the class

Sargul:
This seems not to happen if 
![image](https://user-images.githubusercontent.com/5595511/76959146-520e7700-6919-11ea-8841-ce184ea2e950.png)
is disabled

serge-rider:
Since 6.3.x we use editor's connection for SQL completion lookups. This happens only if "Open separate connection for each editor" option is enabled (since 6.3 it is enabled by default).

This was made on purpose. One of the reasons is that each SQL editor has its own context (e.g. active schema) and thus may return different results.

Is this a problem? If so - please write more details.

Sargul:
In Exasol I use a special connection type for the execution of command (with reduced locking). 

But there is a change on the way (in Exasol software) to enable this on statement level. I think I will go forward an implement this technique in all meta data queries to be sure and independent of Meta vs Editor connections.

Sargul:
I've assigned the task to myself

serge-rider:
I see, ok, this makes sense.

PR #8237, opened 2020-03-21
Enhancement for issue in Exasol plugin #8216

Sargul:
This is an enhancement  for #8216 

When using separate connection for every editor window meta data queries are executed in the data connection of that editor.

As of Exasol Version >= 6.2.5  there is a special comment that will start the sql in snapshot transaction mode in Exasol. In older versions this will just be ignored like a regular comment.

[EXASOL-2646](https://www.exasol.com/support/browse/EXASOL-2646)


Sargul:
And forgot to mention that I fixed a minor bug in ExasolStructureAssistant.java

Issue #8301, opened 2020-03-31
Add Exasol build in functions to auto completion


Sargul:
**Is your feature request related to a problem? Please describe.**
currently some exasol build in functions are show as resevered word.
Example would be: STDDEV_POP

**Describe the solution you'd like**
show as functions like postgre JSONB_BUILD_OBJECT

**Describe alternatives you've considered**
None

**Additional context**
should be fairly easy as there is a hidden meta table containing these functions

Sargul:
Ok - this is basically working. However I'm not sure why no brackets will be added at the end of the functions. 

@serge-rider: Do I have to implement this somewhere?

![auto_complete_new_keywords](https://user-images.githubusercontent.com/5595511/78013139-5473c680-7346-11ea-8d02-e72c190956cc.gif)

serge-rider:
@Sargul I think dbeaver just can't find this function in database metadata. So it is just a reserved keyword.

Sargul:
I've added it to the Dialect as Function. Shouldn't this be enough?

serge-rider:
Dialect provides just a list of keywords/function names which are used for syntax highlighting.

In order to get function signature dbeaver needs implementation of DBSProcedure.
There is method ExasolSchema.getProcedures which returns stored scripts. 
But it doesn't return aggregate functions. So dbeaver knows nothing about signature of these functions.

Generally I don't think that we have support of aggregate functions signature auto-complete for any database. Maybe for PostgreSQL though (as they are similar to regular functions there).
So I'm afraid it won't be that easy to implement this feature.

serge-rider:
I've merged PR so at least we can auto-complete function names without signature.
We'll need another FR for this.

serge-rider:
Or do you mean just brackets without parameters?

Sargul:
Brackets would be sufficient

serge-rider:
It was a bug in standard SQL keyword list. As STDDEV* are functions, not keywords.

PR #8302, opened 2020-03-31
#8301 Add Aggregate functions to auto complete for Exasol

Sargul:
This should implement #8301


serge-rider:
(note we don't put PRs into project, only issues).

PR #8346, opened 2020-04-06
Update Japanese translation

ScratchBuild:


serge-rider:
Thank a lot, looks great.

One note: could you remove lines with original English values? If these strings cannot be translated or they are not translated yet then it will make further translation difficult. 
Eclipse can automatically find all "untranslated" strings. It won't find them because they already have values.

serge-rider:
It is a problem because automatic translation software (any software, not just Eclipse) won't be able to find untranslated strings.
Although generally it will work.

serge-rider:
Thanks!

PR #8354, opened 2020-04-06
Fix templates

VASilaev:
Исправляется ошибка с вводом переменных в шаблон. Если шаблон многострочный и начинается не сначала строки.


serge-rider:
Thanks for PR!

Could you post a couple of examples to test templates with variables.
Also, please do not use non-English comments. dbeaver is multilingual project and base language is English.

VASilaev:
`declare
  ${x} record;
begin
  for ${x} in (${select})
    loop
      ${};
    end loop;
end;`

serge-rider:
You made another PR (#8374) and the closed it - was on purpose?

PR #8386, opened 2020-04-08
GIS data viewer : popular TileLayers

mkgrgis:
Topomaps + 1 satellite images


serge-rider:
Notes: 
- you can't use `&copy;` (it is not supported by xml parser, it comes from html). Use `&amp;copy;` instead.
- you need to escape all `<` characters in the tile definition as `&lt;`. It is another xml parser requirement. See other tiles definition.

PS OpenTopoMap looks good 👍

mkgrgis:
It will be v2

PR #8547, opened 2020-04-24
fix CASE ... WHEN ... END indent

kai-morich:
-indent dection was case sensitive
-unindent relied on SQLDialect.getBlockBoundStrings() returning END
-unindent was one level to much


serge-rider:
Indent formatter is a bit messy (or even worse), thanks for looking into it ;)

uslss:
verified

PR #8618, opened 2020-05-04
hana: provide more driver information

kai-morich:
- link to more specific driver documentation. use help.sap.com search instead of deep link
- show some frequently used driver properties


kai-morich:
to be reworked

Issue #8348, opened 2020-04-06
Show Indexes in Exasol


Sargul:
**Is your feature request related to a problem? Please describe.**
Indexes are automatically maintained (create/drop) in Exasol

**Describe the solution you'd like**
Show Indexes including management (manual drop/create -> even if this is not encouraged by Exasol,it is sometimes necessary)

**Describe alternatives you've considered**
-

**Additional context**
-

Sargul:
was implemented

PR #8659, opened 2020-05-09
Implement #8348

Sargul:
Implement #8348 
- create/drop still missing as I wait for the refactor of UIclasses into separate plugin


Sargul:
Don't know how to fix this new codacy issue as this exact problem is in every *TableIndex class in the other plugins

serge-rider:
It is a very minor warning by Codacy. In fact we need to make better warnings filter there.
Thanks!

Issue #8720, opened 2020-05-17
[Oracle] Metadata search doesn't work for tables if CONSTRAINT and/or FOREIGN KEY is also checked in Object type settings


kseniiaguzeeva:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 7.0.5 EA

#### Connection specification:
- Database name and version Oracle

#### Describe the problem you're observing:
Metadata search doesn't work for tables if CONSTRAINT and/or FOREIGN KEY is also checked in Object type settings.

#### Steps to reproduce, if exist:
1. Open Metadata search
2. select Oracle.HR
3. enter **test** in Object name field (or another table name)
4. Check TABLE, CONSTRAINT, FOREIGN KEY in Object types
5. Press **Search**

Actual result: the object is not found.
![BXJGNNgay2](https://user-images.githubusercontent.com/51405061/82146478-686e7b80-9853-11ea-874f-bb507d192032.gif)

kseniiaguzeeva:
verified

PR #8778, opened 2020-05-22
#8720 ability to search added in case other object types

LonwoLonwo:
when besides FOREIGN KEY/CONSTRAINTS other oracle object types are selected
in Oracle DB Metadata Search


plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleStructureAssistant.java
```diff
@@ -121,6 +121,9 @@ public OracleStructureAssistant(OracleDataSource dataSource)
             if (ArrayUtils.contains(objectTypes, OracleObjectType.CONSTRAINT, OracleObjectType.FOREIGN_KEY) && objects.size() < maxResults) {
                 // Search constraints
                 findConstraintsByMask(session, schema, objectNameMask, objectTypes, maxResults, objects);
+                if(!containsOnlyConstraintOrFK(objectTypes)){
```
ThangLeQuoc, line 124: I would suggest to add space before and after parenthesis
`if (!containsOnlyConstraintOrFK(objectTypes)) {`

Issue #8795, opened 2020-05-26
Option "Force object refresh on editor open" does not work on Oracle packages


simasch:
#### System information: 
- Window 10
- 7.0.5

#### Connection specification:
- Oracle 12.2.0.1.0
- Oracle JDBC

#### Describe the problem you're observing:
Oracle Packages are not refreshed when the option "force object refresh on editor open" is set.

This is a severe error because it may lead to lost updates on package declarations!

#### Steps to reproduce, if exist:
Edit a package in another editor like SQL Developer and open the package in DBeaver. You will see that the changes are not visible.

#### Include any warning/errors/backtraces from the logs
No entries found in the log files

simasch:
The problem are these lines in NavigatorHandlerObjectOpen

```
if (DBWorkbench.getPlatform().getPreferenceStore().getBoolean(NavigatorPreferences.NAVIGATOR_REFRESH_EDITORS_ON_OPEN)) {
    if (databaseObject instanceof DBSObjectContainer) {
        // do not auto-refresh object containers (too expensive)
    } else {
        refreshDatabaseNode(dnNode);
    }
}

```

OraclePackage implements DBSObjectContainer and will not be refreshed.
This is a mistake for Packages.

uslss:
verified

PR #8799, opened 2020-05-26
#8795: Option "Force object refresh on editor open" does not work on …

simasch:
…Oracle packages: Removed exclusion of DBSObjectContainer from refresh

I don't see a problem when refreshing DBSObjectContainer . We have very large Packages in Oracle and I didn't found an issue.


p91paul:
I'm not a dbeaver developer and I don't know why containers were excluded to begin with. However I think it might be safer to add a condition specifically for oracle packages instead of refreshing all DBSObjectContainer. Especially since you are touching code that seems shared between all database providers, and it's unlikely you know the implications for all DBSObjectContainer subclasses in all DBMS.

simasch:
@p91paul I can only see two DB specific implementations for DBSObjectContainer MySQL and Oracle. And therefore I removed the check. But you are probably right.

serge-rider:
I can't remember when and why this condition was added. Perhaps it doesn't make any sense now.
Let's try it (fix will be added and tested after 7.1).

Issue #8898, opened 2020-06-04
Support For YugabyteDB Distributed SQL Database


nchandrappa:
**Is your feature request related to a problem? Please describe.**
Please support the YugabyteDB Distributed SQL database in DBeaver. 

**Describe the solution you'd like**
YugabyteDB supports ANSI SQL, fully-relational geo-distributed SQL with PostgreSQL compatibility. As we continue to bring the PostgreSQL features into geo-distributed SQL, we will need to only enable DBeaver PostgreSQL features that are currently supported by YugabyteDB. As part of the this solution, we will configure a Yugabyte Specific JDBC Datasource. 

**Describe alternatives you've considered**
We have tried creating a custom driver definition of type "PostgreSQL", by default PostgreSQL driver uses `postgres` user and `postgres` database, however, Default user and database for YugabyteDB is `yugabyte`. Additionally, we have to disable PostgreSQL features like `MaterializedViews` which are yet to be enabled in YugabyteDB.

serge-rider:
Please see comments in #8899

serge-rider:
Done in #8899

uslss:
Incorrect logo size in connection create dialog
![image](https://user-images.githubusercontent.com/31996417/84877711-8406b500-b091-11ea-8551-09536cd29fd3.png)

serge-rider:
@nchandrappa could you fix logo icon. Max dimensions are about 300x100.

nchandrappa:
Sorry folks, I didn't see the comment. I'll update the logo with the correct dimensions and revert back. Thanks.

nchandrappa:
I've submitted a PR to fix this - #9090 .

serge-rider:
@nchandrappa Thanks!

Moving t oQA

PR #8899, opened 2020-06-04
Initial commit to Support YugabyteDB - #8898

nchandrappa:
Changes for adding YugabyteDB support in DBeaver. Included the YugabyteDB specific changes to include YugabyteDB logos, describe and YugabyteDB specific DataSource file.


serge-rider:
Thanks for submission.
Generally it is good but database icons aren't.
We use 4 icons: 16x16, 32x32, 64x64 and 128x128.

First two icons are used in database navigator and they are really small (16x16 is used on regular monitors, 32x32 on HighDPI monitors).
Second two icons are used in connection wizard/settings dialogs and they a bigger.

So, you need to fix small icons, current ones will be unreadable on regular FullHD monitors.

nchandrappa:
@serge-rider Thank you for the feedback. I have updated the Yugabyte logo to match the recommended resolutions `16x16, 32x32, 64x64, and 128x128`.

PR #8916, opened 2020-06-08
Default to using MySQL's longer connection timeout var

angusholder:
MySQL and MariaDB have a connection property called `interactiveClient`, which is intended to be set on connections that are in direct use by a human, and causes MySQL to make the connection timeout be the (typically longer) `interactive_timeout` variable instead of `wait_timeout`.

As an example, our database has `wait_timeout` as 10 minutes, and `interactive_timeout` as 8 hours. Since DBeaver wasn't setting `interactiveClient` by default, my connection was being killed after 10 minutes, causing my temporary table I was working with to be dropped, which is pretty annoying and unnecessary.

A stock install of MySQL  has both `interactive_timeout` and `wait_timeout` set as 8 hours, so many people probably don't run into this snag until they start tweaking some of the database defaults, set `wait_timeout` lower to keep so fewer connections sitting around, then don't realize it's the cause of their frustrations that DBeaver's connections die very quickly.


serge-rider:
Let's try it in 7.1.5.

Thanks for the contribution!

Issue #8304, opened 2020-03-31
oid::regclass returns NULL


michalk-k:
Doday I found that cast of **srrelid** column of **pg_subscription_rel** table returns NULL.
By documentation, srrelid column is type of oid (the same as in pg_class)

The same query run in pgAdmin3 returns table names which leads me to conclusion this is DBeaver issue.

![obrazek](https://user-images.githubusercontent.com/7868445/78040755-286d3b00-7370-11ea-9e84-902e0a4e96ad.png)

![obrazek](https://user-images.githubusercontent.com/7868445/78040810-3e7afb80-7370-11ea-8353-cab441a02c7c.png)

If I cast one of IDs directly in SELECT statement, it works as expected:
![obrazek](https://user-images.githubusercontent.com/7868445/78041419-f3151d00-7370-11ea-8704-af43f06e51e4.png)

A bit strange is that cast performed in similar way on pg_class works as expected (SELECT oid::regclass FROM pg_class).

Anyway I'm almost sure it worked properly in DBeaver v6

michalk-k:
Here is similar issue with regrole:

![obrazek](https://user-images.githubusercontent.com/7868445/78041939-aaaa2f00-7371-11ea-9c62-9eb743af66b8.png)

Of course this query is proper working as expected in native postgresql tools

serge-rider:
@michalk-k That's weird. DBeaver converts column type using postgresql data type definition. regrole is defined as numeric type in pg_types. Thus fetch fails.

Perhaps we should make regrole a string data type. It looks like a hack but may work.

serge-rider:
Yeah, in fact I read this so many times )

I found a bug, by some reason `regrole` was the only OID type which wasn't marked as a special alias. Fixed.

michalk-k:
one issue solved, one left: what about issue from the first posting?

serge-rider:
I can't reproduce first issue so far (as I don't have a test db with subscriptions).
Queries like `SELECT oid::regclass,* FROM pg_class` work fine..

michalk-k:
As far as i remember pg_publication_rel behaves the same. so it's enough to initialize a publication with some tables

michalk-k:
Today I spotted new version v7.0.2 with change log entry about fixed OIDs.
I've checked regrole stuff. Right now there is no exception in record but NULL value while text should be displayed. Since consistent with my the first report.

SELECT oid, oid::regrole FROM pg_catalog.pg_roles 
[...]

PR #8962, opened 2020-06-14
Check data kind while matching column for query result

mlt:
This was reported on #postgresql IRC channel and I felt like fixing it.
The following query shows an empty result in DBeaver
```sql
create table post (post_id integer generated by default as identity);
ALTER TABLE post ADD tags _text NOT NULL DEFAULT '{}'::text[];
insert into post values (1, array['Apples']);
select tags[1] from post
```
https://www.db-fiddle.com/f/hDNfD3uVQXwYhTyvVB7T6G/1
But `select tags[1] as foo from post` works.
Before
image
After
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
@@ -673,6 +673,7 @@ public class DBExecUtils {
                     }
 
                     if (tableColumn != null &&
+                        tableColumn.getDataKind().isComplex() == attrMeta.getDataKind().isComplex() &&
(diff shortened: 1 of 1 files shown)


mlt:
I'm actually not sure whether it is the way. I can see also [haveEqualsTypes](https://github.com/dbeaver/dbeaver/blob/d69a75e63bf0a00e37f6b4ab9c9aa4fcaa0ded23/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/DBDAttributeBindingMeta.java#L234) that has that check but setEntityAttribute still changes things.

serge-rider:
I'm afraid it is more complicated issue.
Overwriting resulset metadata with real table column metadata causes different problems in different cases. We can't disable this feature (as it will disable many useful features). And we can't leave it as is.

serge-rider:
The issue is similar to the #8304

PR #9007, opened 2020-06-17
Update small SQream DB icons to be more legible at 16x16 as requested

arnons1:
@kseniiaguzeeva - I've replaced the SQream DB icons to be more legible at 16x16


kseniiaguzeeva:
Verified. Thank you!

Issue #1089, opened 2016-12-21
Add links from text editor to Google search


lacanoid:
This is a simple usability improvement suggestion;

In the context menu in the text editor, please add "Search Web with Google..." item, which would run google search with the selected text (in a default web browser). Xcode has this. It is quite handy for quickly finding documentation online. Now one has to copy/paste stuff around to do this, which is rather clunkier (1 click vs. about 4).

Perhaps this also applies well to the resultset, log & output, not only the text editor.

GiantCrocodile:
I like this feature request but I suggest that we can configure which URL is queried per GET-request. I would set some SQL wiki and not Google.

serge-rider:
I like this

lacanoid:
One wouldn't want to overcomplicate it.

Perhaps an editable list of links, customizable somewhere in preferences. Or something similiar to macos services. Maybe there is a suitable eclipse plugin already.

serge-rider:
As far as I know there is no such plugin available. There is a couple pretty old ones which are integrated with JDT but not with general text editor.

lacanoid:
@GiantCrocodile Did you have a particular wiki/website in mind?

GiantCrocodile:
I could imagine something like this @lacanoid: https://wiki.postgresql.org/wiki/Special:Search?search=Concat&go=Go. Problem is when you use non English versions of DBMS and you try to search for German error message in Google. That never works so I prefer that I can search for English keywords easily without having to look at Google's results but have one site like the wiki of my DBMS where I find it for sure and even with correct syntax etc for my specific DBMS.

serge-rider:
Implemented in https://github.com/dbeaver/dbeaver/pull/9100

uslss:
verified

PR #9100, opened 2020-06-24
Issue 1089 work in progress version of adding a new menu item

halitanildonmez:
Hello! I have been working on this issue: https://github.com/dbeaver/dbeaver/issues/1089

I managed to add a basic functionality yet I am not sure at the moment if I am following the guidelines to add such a thing properly. 

Though here is a WIP of the feature. Note that it is not done but I wanted to share the current work and get some feedback and hopefully amend my mistakes sooner than later.

Also the https://github.com/dbeaver/dbeaver/blob/devel/.github/CONTRIBUTING.md file seems to be empty.


halitanildonmez:
I have fixed the styling issues and this is no longer WIP (I hope 😄 ) 

@serge-rider will you be proceeding with this PR ? Should some one else test it ? 

Is google the default search engine ? Should that be a property user can modify ? 

If you do not want to continue with this or something else I can close this PR since it has been open for a long time 😄

serge-rider:
Thanks for the contribution!
I like the idea, it can be very handy.

What I don't like:
- It is in the core plugin. It'd be better in org.jkiss.dbeaver.ui.editors.sql
- The action is contributed in the SQL editor context menu. Which is already too heavy. Perhaps we need some sub-menu with SQL tools. And put all tools there. Like this one or UUID generator or text formatter/morpher.

halitanildonmez:
I can absolutely do those! 

This was my first swing with dbeaver code so apologies for the silly mistakes 😄 

Yet if you like it I can update the code.

Question about sub-menu with SQL tools: do you mean creating a new package ? I am going to investigate but if you already have an opinion I can do that! 

Thanks again for the input!

halitanildonmez:
I have moved the plugin from core, it should fix the first point, 

going to investigate the second point as I am still not sure what I have to do 😄

serge-rider:
Unfortunately it doesn't compile now.
See https://github.com/dbeaver/dbeaver/pull/9100/checks?check_run_id=1227939790

halitanildonmez:
@serge-rider Thanks for looking! 

It was a strange thing, I have java 11 in the path and that was the problem. 

It looks better now though 😄 Let me know if there is something else I need to fix!

serge-rider:
Now it looks much better 👍 
A few other notes:
- Use https protocol for google links
- Trim selected text
- Use UIUtils.launchProgram(URL) instead of native Java AWT utility.

halitanildonmez:
@serge-rider thanks again! Hope this looks better now ?

serge-rider:
We'll include it in 7.2.5 version. 
Thank you!

PR #9127, opened 2020-06-29
feat: add support for the official Cloud Spanner JDBC driver

### LINKED ISSUE #9126: Use official Cloud Spanner JDBC driver
DBeaver currently support Google Cloud Spanner using the community open source driver (https://github.com/olavloite/spanner-jdbc). This driver is no longer actively developed, and will not support new features that Cloud Spanner might introduce in the future.

DBeaver should preferably (also) support the use of the open source JDBC driver supported by Google (https://cloud.google.com/spanner/docs/use-oss-jdbc).

Continue to use the community driver, and accept that users might not be able to access all features of Cloud Spanner through DBeaver.

**Additional context**
I'm the author of the community open source JDBC driver for Cloud Spanner, and while I'm obviously very happy that you chose to include my driver in DBeaver, I also know that it in the long term would be better to move to the officially supported driver.

(I'm creating this feature request so I can attach a PR to it that I already have prepared.)

olavloite:
Adds support for the open source JDBC driver for Google Cloud Spanner that is backed by Google (https://cloud.google.com/spanner/docs/use-oss-jdbc). This makes sure recent and future features that are added to Cloud Spanner are also accessible through DBeaver.

Fixes #9126


serge-rider:
Thanks for the contribution!

But how it differs from our existing Google Spanner extension?
https://github.com/dbeaver/dbeaver/tree/devel/plugins/org.jkiss.dbeaver.ext.spanner

olavloite:
> Thanks for the contribution!
> 
> But how it differs from our existing Google Spanner extension?
> https://github.com/dbeaver/dbeaver/tree/devel/plugins/org.jkiss.dbeaver.ext.spanner

## Features

It uses the official Google Cloud Spanner open source JDBC driver (https://mvnrepository.com/artifact/com.google.cloud/google-cloud-spanner-jdbc/1.15.0) instead of the community driver that I developed on my own initiative back when Cloud Spanner beta was first released (https://mvnrepository.com/artifact/nl.topicus/spanner-jdbc/1.1.6). The most important differences between the two drivers at this moment is:
1. The community (nl.topicus...) driver does not support backend DML. Instead, all DML statements are parsed internally in the JDBC driver and then translated into Cloud Spanner specific RPCs. The reason for this is that the first releases of Cloud Spanner did not support DML. The official driver has support for normal server side DML statements.
1. The official driver also has support for [Partitioned DML](https://cloud.google.com/spanner/docs/dml-partitioned).
1. The official driver supports foreign keys. (Those were not implemented when the community driver was developed)
1. The official driver has a built-in retry mechanism for read/write transactions. Cloud Spanner may abort any read/write transaction, and the official driver will automatically retry the transaction when it detects this. If no differences are detected between the original transaction and the retried transaction, the abort will be transient for the user. The community driver will just throw an exception in such a case.

And more in general: The official driver continues to be updated and will support new features of Cloud Spanner. The community driver will only receive patches for existing functionality.

## Technical differences
The connection URL format of the official driver significantly differs from the connection URL format used by the community driver. The format for the official driver is `jdbc:cloudspanner:[//host[:port]]/projects/project-id[/instances/instance-id[/databases/database-name]][\?property-name=property-value[;property-name=property-value]*]?`. The official driver requires you to specify the project, instance and database names as part of the URL, while the community driver expects those values to be given as properties. The two drivers also support different connection properties. This makes it a little difficult to let one be a direct replacement of the other.

serge-rider:
This sounds reasonable. But does it make sense to create a separate Spanner extension while we can just add new driver definition? Similarly to Presto - we have 2 different drivers (Facebook and Community). While database server is the same and SQL dialect is the same.

Maybe adding new driver definition in the existing Spanner plugin will be enough?

olavloite:
> Maybe adding new driver definition in the existing Spanner plugin will be enough?

I'll have a look and see if that is doable. I'm not super familiar with the plugin infrastructure of DBeaver (as in; first timer :-)), but based on the above example I'll give it a try.

olavloite:
@serge-rider I've refactored the change so that there is only one Spanner plugin containing two different JDBC drivers. I've kept the community driver with its existing ID, so existing connections should continue to work, and added the officially supported open source driver. PTAL.

serge-rider:
@olavloite Thanks!
Will test and probably include in 7.1.3 version 👍

Issue #9205, opened 2020-07-08
ValueValidator is not working


LonwoLonwo:
For example in ORACLE Gather table statistics tool. Numbers can be entered, but they do not appear in query
![2020-07-08 12_30_00-DBeaver 7 1 2 - _orcl_ Script-4](https://user-images.githubusercontent.com/45152336/86919458-f0288600-c130-11ea-9b8f-f3acd0a3ae53.png)

kseniiaguzeeva:
verified

PR #9217, opened 2020-07-09
#9205 Oracle validator fixed

LonwoLonwo:
valueValidator in OracleToolTableGatherStatisticsSettings fixed


serge-rider:
Please use CommonUtils.toInt instead of Integer.valueOf(value.toString())

Issue #9008, opened 2020-06-18
Can't execute query in new tab when a query is running


nicchristeller:
### System information: 
- Operating system (distribution) and version
Windows 10
- DBeaver version
7.1.0

#### Connection specification:
Any
#### Describe the problem you're observing:
"execute SQL in new tab" fails with "Cannot execute more than one query in one tab. Try to use multiple tabs for simultaneous query execution." when the current tab as an ongoing query.
#### Steps to reproduce, if exist:
- execute a long running query
- attempt to execute another query while  the first is still running
- **both "execute SQL statement" and "execute SQL in new tab" actions result in "Cannot execute more than one query in one tab. Try to use multiple tabs for simultaneous query execution."**
- especially problematic when you only have one resultset tab currently open, as I can't see how to run another query without cancelling the ongoing one

uslss:
thanks for the bug report

mlazatinph:
https://github.com/dbeaver/dbeaver/pull/9226 fixes this

uslss:
Reopened.
For now new tab is opened, but both queries are executed, so result for the second query is shown only after first query is executed.
Steps to reproduce:
1. Open sql editor (e.g. for postgres connection)
1. Execute select pg_sleep(1000); using ctrl+enter
1. Than execute some select query using ctrl+\ (e.g. SELECT * FROM actor )

serge-rider:
It depends on database driver. Some drivers (including PG and MySQL) lock active connection on query execute so you can't execute multiple queries simultaneously. Others don't (e.g. Clickhouse, SQLite).

This particular fix resolves the problem on client (DBeaver) side.

PR #9226, opened 2020-07-10
Execute in new tab: create a new tab if current query processor has jobs running

mlazatinph:
When the user issues an "Execute in new tab" command, we need to check if the current query processor is running a query. If it is, we want to create a new tab.


mlazatinph:
this is an attempt to fix https://github.com/dbeaver/dbeaver/issues/9008

serge-rider:
@mlazatinph Thanks!

PR #9234, opened 2020-07-10
Issue 4729: adding a new action that save a temp file of a blob and p…

halitanildonmez:
Adds a new button to DBeaver when the user views the blob values. When pressed the file content is saved as octet-stream as a temp file and the OS propts the user to open the file choosing a viewer program. 

If the file is null nothing will happen, Not using a mime detector since the capability is limited and my trials shows to me that it is simpler to go this way though if need be I can use mime detectors (like probeContentType)

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
+++ b/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
@@ -250,6 +250,7 @@ public class DBeaverApplication extends BaseApplicationImpl {
                 return EXIT_OK;
             }
 
+            
             // if the exit code property has been set to the relaunch code, then
             // return that code now, otherwise this is a normal restart
             return EXIT_RELAUNCH.equals(Integer.getInteger(PROP_EXIT_CODE)) ? EXIT_RELAUNCH
--- a/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
+++ b/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
@@ -16,6 +16,14 @@
  */
 package org.jkiss.dbeaver.ui.app.standalone;
 
+import java.io.File;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintStream;
+import java.net.URL;
+import java.util.Properties;
+
 import org.apache.commons.cli.CommandLine;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.equinox.app.IApplication;
@@ -55,10 +63,6 @@ import org.jkiss.utils.ArrayUtils;
 import org.jkiss.utils.CommonUtils;
 import org.jkiss.utils.IOUtils;
 import org.jkiss.utils.StandardConstants;
-
-import java.io.*;
-import java.net.URL;
-import java.util.Properties;
 
 /**
  * This class controls all aspects of the application's execution
--- a/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
+++ b/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
@@ -16,14 +16,6 @@
  */
 package org.jkiss.dbeaver.ui.app.standalone;
 
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.io.PrintStream;
-import java.net.URL;
-import java.util.Properties;
-
 import org.apache.commons.cli.CommandLine;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.equinox.app.IApplication;
@@ -63,6 +55,10 @@ import org.jkiss.utils.ArrayUtils;
 import org.jkiss.utils.CommonUtils;
 import org.jkiss.utils.IOUtils;
 import org.jkiss.utils.StandardConstants;
+
+import java.io.*;
+import java.net.URL;
+import java.util.Properties;
 
 /**
  * This class controls all aspects of the application's execution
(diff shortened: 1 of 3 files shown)


rojares:
I cloned git@github.com:halitanildonmez/dbeaver.git
Then I ran "mvn install"
How can I now run the compiled program?

halitanildonmez:
So what I do is following here (developing in Eclipse): https://github.com/dbeaver/dbeaver/wiki/Develop-in-Eclipse and debug the application

Yet I think you can do the steps here: https://github.com/dbeaver/dbeaver/wiki/Build-from-sources

You only have to run `mvn package` and it will be here: `product/standalone/target/products`

rojares:
I found the executable in C:\Users\ramio\dv\dbeaver\product\standalone\target\products\org.jkiss.dbeaver.core.product\win32\win32\x86_64\dbeaver
When I execute dbeaver.exe I get the following dialog.
image

halitanildonmez:
I am looking into this now. Sorry about the delay

halitanildonmez:
This works for me so I assume that it may be your setup ? 

Here is what I did: 

1. clone https://github.com/halitanildonmez/dbeaver.git
2. cd dbeaver
3. checkout "feature_open_blob_external"
4. run "mvn clean install"
5. run "mvn package"
6. run the exe. 

I have java version: 

java version "11.0.5" 2019-10-15 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.5+10-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.5+10-LTS, mixed mode)

Maven version: 

Apache Maven 3.6.3

OS: Windows 10 64 bit.

rojares:
Followed your steps but I still get the same error.
My java is 1.8
I downloaded the newest community edition in zip package, extracted and ran the exe and got the same error.
But my version installed using the installer works alright.

halitanildonmez:
Well I dont know to be honest. 

I googled your error and it seems like there are some suggestions: https://javahungry.blogspot.com/2018/10/solved-java-started-returned-exit-code-13-error-eclipse.html

Have you looked into it ? Also can you confirm if the development branch is working ?

halitanildonmez:
Another thing to try is to clone the repo again but this time trying to build the master or the devel branch since they wont have the changes I made. 
This way we can figure out if it is the changes or something else

halitanildonmez:
Would it also work if I share the executable ?

rojares:
I updated my java to the newest openjdk and launching the exe file started to work.
I tried opening a blob column value both using the icon (3 dots) and from context menu (Opne in external editor) but nothing happened.

halitanildonmez:
Glad to know it is working! Thank you for checking! :D 

About the test data: is blob empty ? What does it contain ? Just to know how  I can reproduce the problem

rojares:
I know that you mean well Halitanildonmez, but this is really becoming too big burden for me...

I created 1 table with three columns: int, varchar and blob.
I inserted 2 rows.
The first row had an image in the blob column and the second one a pdf file.
The internal dbeaver value viewer was able to show the image and the pdf it showed as ascii.
Open in external editor did not do anything.

halitanildonmez:
ok sorry about it. Promise I am bothering for the final time. 

After fixing I will update the merge request. You won't have to do anything unless you want to. 

Apologies for this being a burden.

halitanildonmez:
New commit, it is ok if you can not test it. Not even sure if I can get it to be approved. 

I hope it works

rojares:
This is what I get when I try to open a BLOB containing pdf file with the Open in external editor.
image
Have you tested this feature yourself?
Is it working on your machine?

rojares:
After the first dialog a second dialog was shown...
image

rojares:
Ok so I made some tests because the NullPointerException got me thinking.
I added a new column that was varbinary.
The difference to BLOB is that you can get the whole value directly from varbinary.
From blob you get an inputstream and read the value from the input stream.
So blob returns a null when trying to get the value directly.
When I was opening the pdf from a varbinary column, this is what I received:
image
image
So all binaries are opened as .octet-stream file.
It should be clear that you have still a long way to go with your development.
You have to use some kind of library that can recognize the mime type of a binary file just by reading part of it's beginning.
Here is an article where you can find ideas.
https://www.baeldung.com/java-file-mime-type

rojares:
Only Apache Tika, which is mentioned last, will work in this case.

halitanildonmez:
Ok I am taking a look again

halitanildonmez:
Ok So I am no longer getting the contents. Very bad implementation on my part. 

I have been investigating the code and I have added a new way to handle the input. I will continue working on this but just wanted to give a heads up. Now you should at least be able to open both types. 

About mime types: 

When I was investigating this, I tried the link you pasted, checked URL library as well and they were not detecting the mime types correctly and/or they were really limited. The best way to go according to me was to simply have a hashmap<mime_name, file start string> and check there. 

Then I saw what you wrote (save as octet) and to me that was a better way to go instead of maintaining a list or adding an external library. 

Another question: I usually can not reproduce the errors you get and I always try in my system before pushing. It works. And what I have is an SQL table with 3 columns (string, string, blob) and I use the data editor and upload a pdf file to a BLOB. That is how I test. Is this how you test as well ?

halitanildonmez:
OK I think it is done now. I am able to open multiple types of files.

rojares:
I was able to open a pdf file so I guess that's good enough for me.
You could skip the extra dialog that says "Opening the file at ...."

halitanildonmez:
Ok going to make some minor tweaks. Sorry for being late a bit.

halitanildonmez:
@serge-rider maybe your feedback is needed here ?

Arslan007:
I have tested and I really like that additional button to Export/Save Binary(BLOB) data to a file.

serge-rider:
Thanks!
I'll review it before 7.2.1

serge-rider:
The feature is useful but code doesn't match our policies:
Please
1. Remove DBeaverApplication.java and ContentPanelEditor.java from PR: they has nothing to do with the feature
1. Do not reformat class imports (we use star iport when number of packlage classes >= 5)
1. We use spaces instead of tabs (1 tab = 4 spaces). Tabs break source formatting.

halitanildonmez:
Ok I am going to fix it! Thank you for the input!

halitanildonmez:
@serge-rider Again thanks for the input! 

I have fixed the tabs yet I am not sure if my styling rules are correct. For example, what is the allowed number of characters per row ? 

I have removed an extra dialog @rojares said I can just remove. 

Please let me know if you have another issue with this and I can fix it asap!

serge-rider:
Thanks 👍 
Now it looks good

Issue #9010, opened 2020-06-18
Success status is displayed for MySQL,DB2 tool task if a table, mentioned in the task, is deleted


kseniiaguzeeva:
#### System information:  
- Operating system (distribution) and version Win10
- DBeaver version 7.1.1 EA
- MySQL

Success status is displayed for MySQL tool task if a table, mentioned in the task, is deleted

Steps:
1. Create a MySQL tool task for an eisted table (e.g. Kseniia.MyISAM_4)
2. Execute the task
3. Delete the table
4. Execute the task again using btn on Database Task Panel.

Expected: status **Error** appears is the task status (on the right of Database Task Panel)
Actual result: status **Success** is displayed in the task status (on the right of Database Task Panel)

kseniiaguzeeva:
Actual for DB2 too.

uslss:
verified

PR #9247, opened 2020-07-13
#9010 error status for tasks on deleted objects added

LonwoLonwo:
Success status is not displayed anymore if a table, mentioned in the task, is deleted
2020-07-13 21_41_47-DBeaver 7 1 3 - run_202007132135_2 log


serge-rider:
We must show original error (which led to empty object list) and show it in UI.

Issue #8713, opened 2020-05-15
Colored Foreign Schemas in ERD


rafardenas:
#### Suggestion
When visualizing the ER of a table, one can see the relationships between them. Would be nice, if the tables that are on another schema be colored on another way to better identify them 
 :)

uslss:
thanks for suggestion

uslss:
verified

PR #9255, opened 2020-07-14
Colored Foreign Schemas in ERD #8713

SA357:
Now tables in different schemas are colored differently

image


serge-rider:
Don't use JDBCTable unless you really need it

Issue #9306, opened 2020-07-18
DBeaver shows wrong array values from postgres - array values are correct in database


TryAndErr0r:
#### System information: 
- W10
- Version 7.1.2.202007051953
- Additional extensions - none

#### Connection specification:
- Postgresql
- org.postgresql.Driver
- Do you use tunnels or proxies (SSH, SOCKS, etc)? -no 

#### Describe the problem you're observing:
When using an array collumn of decimal values DBeaver failes to format the values right.

When using psql in bash:
30637 | 2020-07-18 11:06:18.71624 | 2020-07-18 09:15:07.515012+00 |   2068 | {{1.0001000000000000,15351.0700000000000000},{1.0002000000000000,12050.0000000000000000},{1.0003000000000000,9650.0000000000000000},{1.0004000000000000,2479.6200000000000000},{1.0005000000000000,1966.5000000000000000}} | {{1.0000000000000000,13929.9100000000000000},{0.9999000000000000,19707.1400000000000000},{0.9998000000000000,12668.5400000000000000},{0.9997000000000000,37557.6100000000000000},{0.9996000000000000,4150.3500000000000000}}

When using DBeaver text output 
30637|2020-07-18 11:06:18|2020-07-18 11:15:07|  2068|{{1.0001000000000000,153510700000000000000,{1.0002000000000000,120500000000000000000,{1.0003000000000000,96500000000000000000,{1.0004000000000000,24796200000000000000,{1.0005000000000000,19665000000000000000}|{{1.0000000000000000,139299100000000000000,{0.9999000000000000,197071400000000000000,{0.9998000000000000,126685400000000000000,{0.9997000000000000,375576100000000000000,{0.9996000000000000,41503500000000000000}|

#### Steps to reproduce, if exist:
1. Step create table:
CREATE TABLE public.orderbookrecords (
	id int4 NOT NULL GENERATED ALWAYS AS IDENTITY,
	msgtimestamp timestamp NULL,
	databasetimestamp timestamptz NULL DEFAULT CURRENT_TIMESTAMP,
	asks _numeric NULL,
	bids _numeric NULL,
	CONSTRAINT orderbookrecords_pk PRIMARY KEY (id)
);

2. Step create record:
Insert record:
[...]

PR #9321, opened 2020-07-20
Issue 9306: Fixed object parser to support multi-dimensional array for viewer

llama90:
#9306  

query panel

query result panel


serge-rider:
Thanks!

We'll to add unit tests for PG structs/arrays parser. It become too complicated.

hsseo0501:
Hi, @serge-rider This issue will be resolved by yourself? 

if that is right, I will try to find other issues for contribution.

serge-rider:
I'll merge this PR after we'll add several unit tests. And then close #9306

hsseo0501:
@serge-rider Thank you for your reply.

serge-rider:
We did it in a bit different way in #9398

Issue #9343, opened 2020-07-22
Task configuration: hide empty properties


serge-rider:
We need to hide properties panel if there are no properties for the task.

kseniiaguzeeva:
closed

PR #9354, opened 2020-07-23
Task configuration: hide empty properties #9343

SA357:
taskOptionsViewer is invisible if there are no item options

image


serge-rider:
Use getItemCount function

PR #9358, opened 2020-07-24
8859 oracle default numeric

llama90:
The column type is changed to INTEGER

image

And then ctrl + S

image

Click the persist 

image

Also DECIMAL and SMALLINT are same as INTEGER


hsseo0501:
Bugs were found, fixed, and should be PR again.

p91paul:
well, integer smallint and decimal are just aliases for different NUMBER sizes in Oracle...

Issue #9046, opened 2020-06-20
Shape field has weird characters


ernestocd:
System information:
•	Windows 10 Pro
•	DBeaver version: Version 7.1.0.202005311732
Connection specification:
•	SQL Server Management Studio	18.5
Describe your question:
•	I am new to Dbeaver, I have imported the database successfully but when reviewing the shape field, it appears to have weird characters, therefore, I am unable to use the spatial components the platform offers.

kseniiaguzeeva:
Could you please send a screenshot of the problem?
Which data type does it affect?

ernestocd:
Here is the screenshot illustrating the shape field with the weird characters I mention above. Not sure if it is useful to know but this is an MS SQL spatial database created with ArcGIS, i.e., Esri software.

![image](https://user-images.githubusercontent.com/4702142/86789363-38fb1480-c035-11ea-933b-9b9b44e808c1.png)

jlrodasr:
I have the same problem, any SELECT against MS SQL Server (2016 in my setup) with a geometry column does not get recognized by DBeaver as a query with a spatial column. For example, the following table:
```sql
CREATE TABLE [dbo].[ORDENES_DE_TRABAJO](
	[OBJECTID] [int] NOT NULL,
	[Shape] [geometry] NULL,
 CONSTRAINT [R1273_pk] PRIMARY KEY CLUSTERED 
(
	[OBJECTID] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, FILLFACTOR = 75) ON [PRIMARY]
) ON [PRIMARY] TEXTIMAGE_ON [PRIMARY];

ALTER TABLE [dbo].[ORDENES_DE_TRABAJO]  WITH NOCHECK ADD  CONSTRAINT [g1248_ck] CHECK  (([SHAPE].[STSrid]=(32717)));

ALTER TABLE [dbo].[ORDENES_DE_TRABAJO] NOCHECK CONSTRAINT [g1248_ck];
```
With the following data:
```sql
INSERT INTO [dbo].[ORDENES_DE_TRABAJO](OBJECTID, Shape) VALUES(1, geometry::STGeomFromText('POINT (730121.23979999963 9684060.2467)', 32717));
[...]

PR #9359, opened 2020-07-24
Issue 8859: oracle default numeric

llama90:
image

Column types changed to decimal, integer, and smallint.
The decimal type needs the size, which is automatically changed to 38 on the SQL viewer.

image

And then ctrl + S.

image

The SQL is generated as shown above.

And persist is processed. 

image


hsseo0501:
I tried to delete a branch on an issue #9046 but accidentally deleted this branch.

serge-rider:
It is better to fix it in metadata read routine.
We shouldn't use max length in bytes (22). We should set max precision to the default (38) unless precision is specified explicitly.

PR #9391, opened 2020-07-29
Added personal-files interface for snap

riednyko:
personal-files provides access to the specified files in the user’s home. This interface gives privileged access to the user’s data.


Wouter0100:
I see that currently only the `.ssh`-folder is requested. Wouldn't the [`ssh-keys` interface](https://snapcraft.io/docs/ssh-keys-interface) be more appropriate?

Issue #9387, opened 2020-07-29
PostgreSQL table does not show comment in DDL.


smcdl:
#### System information: 
- Operating system (distribution) and version: MacOS
- DBeaver version: 7.1.3
- Additional extensions: Nothing

#### Connection specification:
- Database name and version: PostgreSQL 11.8
- Driver name
- Do you use tunnels or proxies (SSH, SOCKS, etc)? : No

#### Describe the problem you're observing:
It does not show comment in DDL of Properties tab and Generate SQL -> DDL context menu.

#### Steps to reproduce, if exist:
Please refer to description.

kseniiaguzeeva:
Thank you for the bug report.

hsseo0501:
@smcdl @kseniiaguzeeva  Hello, Maybe you can fix that by following the link [issue-8329](https://github.com/dbeaver/dbeaver/issues/8329)

kseniiaguzeeva:
Thank you, indeed. I close the ticket as a duplicate of https://github.com/dbeaver/dbeaver/issues/8329

smcdl:
@kseniiaguzeeva It is not the same question. Generate SQL dialogs does not show 'COMMENT ON COLUMN' sql script and it also dose not have 'Show comments' button yet.

![image](https://user-images.githubusercontent.com/4975064/88785711-1e473600-d1c4-11ea-953d-215f0d2ec2a3.png)

kseniiaguzeeva:
You are right. 

(It can be tested on postgres.Xeniia.9294:
Select Generate SQL>DDL in the table context menu.
Expected: comments for the table and column "pk" are shown in the table DDL.
Actual result: there are no comments at all).

serge-rider:
@SA357 I think we need to add extra options in DDL generate dialog. Otherwise we will use existing options for a very different purposes and this will produce a mess.

I'd suggest to add new object definition options similar to PostgreSQL:
OPTION_INCLUDE_NESTED_OBJECTS
OPTION_INCLUDE_COMMENTS
OPTION_INCLUDE_PERMISSIONS

And add corresponding checkboxes in Generate SQL dialog.

To determine whether an option supported by underlying database object we can use DBPScriptObjectExt2 interface.
Existing implementations of DBPScriptObjectExt2 must be patched.

Then we can reuse this feature in other databases.

uslss:
verified

PR #9401, opened 2020-07-29
[#9387] Fixed generating ddl logic to show comment for postgres

llama90:
This logic works well without disturbing existing logic (DDL tab + Show comments toggle).

I tried to create a show comment tab in the Generated SQL viewer, but Oracle shows the comment when schema contains the comment. So I referenced that. 

# DDL tab
image

# DDL tab (with Show comments)
image

# Generate DDL
image


serge-rider:
I turned #9387 into FR. We need new model options to support this properly (and not just for PostgreSQL).
Closing this PR.
@hsseo0501 thanks for contribution anyways!

PR #9417, opened 2020-07-31
Oracle: Use URL template for custom connections

insightfuls:
This allows you to use {host} and {port} so dynamically-assigned local SSH ports work.

I need this in my context because we use an Oracle database with an SSL connection but which we can only access through an SSH tunnel. We need a horrible looking TNS connection string as part of the JDBC URL with the host and port embedded in it: `jdbc:oracle:thin:@(DESCRIPTION=(ADDRESS=(PROTOCOL=TCPS)(PORT={port})(HOST=localhost))(CONNECT_DATA=(SID=ORCL)))`

Using this string, I can set the remote endpoint of the tunnel to the database host and port, and leave the local end with port 0 for a dynamically chosen free one, and the Oracle driver connects to the allocated port.


insightfuls:
@serge-rider I'm sure you're terribly busy, looking at how many commits you are authoring here. I'm guessing you're responsible for maintaining the project. As the old adage goes, however, "If you want something done, ask a busy person."

Could you please help me understand how to pursue getting this change merged? I couldn't see any particular instructions in the README or contribution instructions or anything. Do I just need to give it time, or is there something more you need from me?

serge-rider:
PR looks good, probably we'll include it in 7.1.5 release after some testing.
Thanks for the contribution!

insightfuls:
Thanks a lot @serge-rider ! I'll look forward to seeing it in a release. If you need any test evidence from me, or anything further, please don't hesitate to reach out.

PR #9471, opened 2020-08-06
PostgreSQL table comment  DDL. #9387

SA357:
Added table and column comments to ddl, they can be turned off by pressing "Compact"-button.

This PR was inspired by https://github.com/dbeaver/dbeaver/pull/9401 .

image

image


hsseo0501:
Is second commit related to compact option?
Why oracle comment is printed by default, but not postgre? ..

serge-rider:
I'm afraid we have a sort of mess in DDL options. 
This is very similar to #9401

vandroiycl:
Can't see comments in a Postgres function source or DDL either, 
a similar "Show comments" button would be useful there too, 
should be a new feature or bug request?

serge-rider:
@vandroiycl Yes, please create another FR ticket for that.

serge-rider:
I turned #9387 into FR. We need new model options to support this properly (and not just for PostgreSQL).
Closing this PR.

PR #9545, opened 2020-08-14
Colored Foreign Schemas in ERD #8713

SA357:
If changing colors is not supported table's background remain yellow


SA357:
Also added DBSCatalog support for this feature

PR #9621, opened 2020-08-25
Fix Athena plugin to script view definitions

edharper01:
`SHOW CREATE TABLE` does not generate view definitions in the current version of Athena.

This change modifies the `getViewDDL` definition to run the correct command (`SHOW CREATE VIEW`) by abstracting out the shared logic to a helper method.


serge-rider:
Thank you for the fix!

Issue #4768, opened 2018-12-10
Navigate Association fails on MySQL 8 with 'columns differ from referenced constraint'


DaPoco1986:
I use DBeaver Version 5.2.1
Mysql 8 under Manjaro Linux
I have 3 schemas containing the same database structure. When I want to use the 'navigate association' feature within a resultset I get following error:
Entity [xxx.branch] association [FK_TpResource_IsBranch] columns differs from referenced constraint [PRIMARY] (3<>1)

When I use a database having only one schema, it works.
I assume the resolver queries the database schema without filtering on the schema itself. So in my case, 3 TableReferences are found for the navigating column and DBeaver does not know which to chose.

serge-rider:
I can't reproduce that. In fact DBeaver uses schema to search reference tables/keys.
I can't figure out how this coudl be possible (looks more like a corrupted schema).
Could you try to this on the latest DBeaver version?

DaPoco1986:
So, I upgraded to the current version, but the error still exists.
Nevertheless I was able to narrow it down. I switched back to an older JDBC Driver (Version 5.1.47) and here it works! The Version I used before was 8.0.11, so I guess there is something wrong with that.

serge-rider:
Ah, this is possible. DBeaver MySQL extension uses native JDBC driver functions to read foreign keys info. So this might be a bug in MySQL driver itself..

DaPoco1986:
That is what I'm thinking. Thanks for your help, I will close this issue.

Dimas-sc:
Same happens with MariaDB. Any solutions?

PR #9884, opened 2020-09-26
MySQL 8: fix duplicates in foreign keys

tmalahie:
In MySQL 8, the function metaData.getImportedKeys() may return twice the same column in the result set.
This is a known bug that has been reported here: https://bugs.mysql.com/bug.php?id=95280.

This issue can be a bit annoying in the case of DBeaver, especially because it breaks the foreign key link feature in the data viewer (It shows an error dialog with message "Entity <db.table> association <fk> columns differs from referenced constraint"). Cf issue #4768 which is exactly what I'm referring to.

I just created a patch to fix this issue. Basically when building the FK list, it checks if the column is already in the list before registering it. Hopefully this pull request gets accepted :)


serge-rider:
Thanks for the contribution 👍

PR #9969, opened 2020-10-05
Fix Sybase stored proc text order

germondc:
When retrieving the source for a Sybase stored procedure that has multiple rows in the syscomments table the order can come back incorrectly and hence what is displayed is incorrect.

The change simply adds an order by to the SQL used to fetch the comment text.


serge-rider:
Thank for the fix!

Issue #9635, opened 2020-08-26
Existing columns are mentioned as new when trying to import db to db


uslss:
#### System information: 
- Operating system (distribution) and version win 10
- DBeaver version 7.2.0

#### Connection specification:
- Database name and version SQLite, Postgres

#### Steps to reproduce, if exist:
1. Export table from sqlite to postgresql - New table is created in Postgresql
1. Start import from table that was created at step 1 to source table from step 1. - Table is mentioned as existing, but all the columns are shown as new (but they are existing) so the import ends with an error
![2020-08-26_15h31_45](https://user-images.githubusercontent.com/31996417/91304588-67ce7500-e7b2-11ea-98f1-59459bff471f.gif)

serge-rider:
This happens because PG keeps unquoted identifiers in lower case while SQLite keeps them in mixed case. DT wizard can't find corresponding columns in target table because it search for lower-case identifiers.

Solution: use case insensitive metadata search for SQLite.

mbelvadi2:
I'm seeing exactly this same problem when importing a tsv file into an existing sqlite table. Even though the import sequence knows the table is existing, and the column headers are in row one of the tsv, because their labels have different case, they show up as Create instead of existing.

akilovich:
This issue appears to be a duplicate of issue #9933 and is already resolved.

Issue #9933, opened 2020-10-01
Preview for export from statement doesn't work


kseniiaguzeeva:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 7.2.1

Steps:
1. Execute a SELECT statement in SQL Editor (or create export from query via tasks)
2. Open Export from query wizard
3. Select database type
4. Press Preview btn in Tables mapping step.

Actual result:
Error appears:
Target attribute for [last_update] wasn't resolved
eclipse.buildId=unknown
java.version=11.0.5
java.vendor=AdoptOpenJDK
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en
Command-line arguments: -os win32 -ws win32 -arch x86_64

![JBUDtcq7N0](https://user-images.githubusercontent.com/51405061/94791605-90b0de00-03e0-11eb-81ad-d2053899f59f.gif)

kseniiaguzeeva:
Still can be reproduced using steps above on   DBeaver 7.2.3.202010120615.

kseniiaguzeeva:
verified

PR #10029, opened 2020-10-09
#9635 Ignore case when finding target attribute by name

ShadelessFox:


ShadelessFox:
This is actually fix for #9933, but it also can fix #9635.

Issue #6289, opened 2019-07-16
Introduce InterSystems IRIS driver


evshvarov:
Hi dbeaver developers!
you already have InterSystems Caché support.
But InterSystems have introduced a new data platform InterSystems IRIS which has similar ODBC/JDBC driver.
IRIS is faster and has more capabilities.
Could you please add the support for IRIS?

IRIS Community verison on docker container could be found [here](https://hub.docker.com/_/intersystems-iris-data-platform)
Any questions could be discussed on InterSystems developers [forum](https://community.intersystems.com)

serge-rider:
It seems that JDBC driver is not available for public download (there is no Maven repo or some direct download link). Maybe you are aware of any community resource for that?

Of course we can just make users to download IRIS by themselves and then to configure driver manually but this is actually a blocker for many people :(

eduard93:
Here's [the driver](https://github.com/intersystems/quickstarts-multimodel-java/blob/master/lib/intersystems-jdbc-3.0.0.jar)

The schema is:
```
                <driver
                    id="iris"
                    label="InterSystems IRIS"
                    icon="icons/cache_icon.png"
                    iconBig="icons/cache_icon_big.png"
                    class="com.intersystems.jdbc.IRISDriver"
                    sampleURL="jdbc:IRIS://{host}:{port}/{database}"
                    defaultPort="51773"
                    webURL="http://www.intersystems.com/try/"
                    description="InterSystems IRIS Data Platform"
                    categories="sql,object">
                    <file type="jar" path="repo:/drivers/cache/CacheDB.jar"/>
                    <file type="license" path="repo:/drivers/cache/LICENSE.txt"/>
                </driver>
```

tom-dyar:
I am happy to make a pull request, if that is what is needed to get this in. If someone can list a concise set of changes that need to be made in addition to the xml snippet above, that would be very helpful. For example, where would we put the jar? I forked the repo, but that seems to have broken "maven package", which builds fine from the base repo. Do I need to develop in Eclipse? Thanks!

serge-rider:
We'll try include it in the next version.
XML config is enough, thanks.

tom-dyar:
That is great, thanks!!

tom-dyar:
We do have a new jar file, and wil be posting to maven at some point. Please do not use the jar linked above!

tom-dyar:
[...]

PR #10093, opened 2020-10-17
Simple InterSystems IRIS support

daimor:
Resolves the issue #6289

Replaced Cache driver with IRIS, as it's compatible.
InterSystems replaced Caché with a new product IRIS Data Platform.


serge-rider:
I don't think that replacing existing driver is a good idea. 
People may have existing connection configurations, they may be broken by such replacement.
I thin we just need to add the new IRIS driver.

BTW are you aware of any Maven-like repository for driver jars? We can live with direct link but it is not good in general (it doesn't allow users to upgrade driver version by themselves).

daimor:
I've checked the replacement of cache to iris, and it works well, so I did it, and it have backward compatibility, and fortunately dbeaver uses uri, which just can be replaced. Cache will not receive any new version any way.

Unfortunately, I'm not responsible for publishing to maven, but we are working on it, and as soon as it will be published there, I'll do a new PR with the fix.

daimor:
May it have replacement and cache driver definition at the same time?

serge-rider:
Big number of people know Cache as Cache though. They will search for Cache, not for IRIS. 
Thus we can't just remove this old driver.

Also, we can use the same icons for both drivers.

Note: adding some Cache reference in IRIS driver description will let users to find both drivers by using quick search.

daimor:
@serge-rider so, I've returned back Cache

serge-rider:
Great!
Please let know when some Maven repository will be available.

PR #10165, opened 2020-10-25
more HANA features - do not merge

kai-morich:
* show error when schemas not loadable (e.g. alter password required)
* show dependencies for HANA views, procedures and functions


serge-rider:
Thanks!
I'd suggest following improvements:
- Write schema read errors in the log (log.error(...)).
- Perhaps it is better to throw error in loadSchemas - dbeaver will show error dialog instead of strange schemas in the list. WDYT?
- In HANADependency.getName() you should enquote identifiers. As they may contain special characters. DBUtils.getFullyQualifiedName will do the job.

kai-morich:
1. the original error is already written to the log [here](https://github.com/dbeaver/dbeaver/blob/234c8e483affb3aa17f5e3c801ef018932bd95a8/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/meta/GenericMetaModel.java#L244). With `SELECT * FROM DUMMY` i just try to get the error message again
2. throwing an error from loadSchemas does not work. It is logged and discarded [here](https://github.com/dbeaver/dbeaver/blob/234c8e483affb3aa17f5e3c801ef018932bd95a8/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericDataSource.java#L517)
3. done

changing code in 1. and 2. that currently discards the error might have side effects for other databases that have half baken schema support, so I had choosen the minimal impact solution with error as pseudo schema

BTW: for non-resolvable dependencies I created HANADependency.DummyObject entries. Is there already a solution that could be used to show just the name without hyperlink?

kai-morich:
`loadSchemas()` fails for HANA if the DB is started in restricted mode, which can happen if password or license is expired. 

Throwing an error from `loadSchemas()` does not show an error dialog as several layers on top catch, log and discard the error. Commenting out the try/catch in the [next layer](https://github.com/dbeaver/dbeaver/blob/234c8e483affb3aa17f5e3c801ef018932bd95a8/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericDataSource.java#L517) only creates an error dialog when refreshing the schema list with F5, but not on initial connect.

Any idea how to proceed?

kai-morich:
Failing `loadSchemas`is now logged as error and thrown as exception. The logging in generic code is only as warning, but for HANA it is an error.

serge-rider:
Sorry for the long delay. 

I've added new function GenericMetaModel.isSchemasOptional. It is true by default. For HANA I set it to false.
If it is false then Generic driver throws real error when schema read fails. See #10510
Hope this will help in HANA.

Also - please add standard Apache license/copyright header in new classes.

Thanks!

kai-morich:
I will split up this PR, as error handling still has issues

Issue #10266, opened 2020-11-06
Hana PUBLIC SYNONYMS are not offered as autocomplete option


wernerdaehn:
#### System information: 
- Operating system (distribution) and version: Windows
- DBeaver version: 7.2.4
- Additional extensions: none

#### Connection specification:
- Database name and version: Hana
- Driver name
- Do you use tunnels or proxies (SSH, SOCKS, etc)? 

#### Describe the problem you're observing:

When you type "select * from PUBLIC." no elements are shown. But when you expand the Synonym Node of PUBLIC, there are lots of synonyms.

Root cause seems to be that in GenericObjectContainer the method [getChildren](https://github.com/dbeaver/dbeaver/blob/62f12b8003b600a3a856c659f1c762de4fac04f2/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericObjectContainer.java#L374) does return tables only.

I considered creating a new HanaPublicSchema class that returns synonyms as well, but the we have the same issue with regular synonyms. Modifying the GenericObjectContainer might potentially have side effects. One kind of side effects is when a class calls getChildren but expects tables only, e.g. in the ERD diagram. But there an explicit filter is added to select primary objects (=tables) only.
Apart from that databases which do support synonyms and use the Generic classes might be impacted. Oracle for example does not use the Generic* classes as foundation but implements the interfaces by itself.

wernerdaehn:
https://github.com/dbeaver/dbeaver/pull/10272

@serge-rider @kai-morich

LonwoLonwo:
duplicate of https://github.com/dbeaver/dbeaver/issues/9710

PR #10272, opened 2020-11-06
#10266 (Hana) Public Synonyms

wernerdaehn:
I had multiple issues with Hana around public synonyms and the case did spread a bit into the Generic* package.

The starting point was the following: Hana supports synonyms and public synonyms. Public synonyms are used a lot for data dictionary, e.g. PUBLIC.TABLES points to SYS.TABLES. A normal synonyms are used often for cross schema access.

Examples:
1 select * from PUBLIC.TABLES; (schema + **synonym**)
2 select * from SYS.TABLES; (schema + table/view)
3 select * from TABLES; (public **synonym**)
4 select * from MYSCHEMA.TABLE1; (schema + table/view)
5 select * from TABLE1; (table/view in selectedschema)
6 select * from MYSCHEMA.REMOTETABLE1; (schema + **synonym**)
7 select * from REMOTETABLE1; (**synonym** in selectedschema)

For all these cases
* auto complete
* column binding (when executing the query);

must work.

When testing only case 2), 4) and 5) did work, and these are the ones where neither a local nor a public synonym had been involved.

The first required change was to add an indication what the public schema is. When a query without a schema is executed, the logic tries to find the table in the selected schema and if not found, must now try for public synonyms. So we must know what schema the public synonym is.

For that I followed the isVirtual() approach and added an interface DBSPublicSynonymSchema and GenericSchema implements it.
Next is the getChildren() and getChild(). They return the TableCache only but must return synonyms as well. Hence I modified the GenericObjectContainer. Problem is, the variables tableCache and synonyms are different. In Oracle a SynonymCache class was added. Instead I changed the synonyms from ArrayList to Map (sorted). This is sort of a poor man's cache. If an ArrayList is fine, a simple Map should be as well.
Finally I had to modify the autocomplete code in certain points to support the missing 4 combinations from above list of seven queries.

One thing I did not however: When a object name without a schema is typed, e.g. "select * from tab..." I do not add the public synonyms to the list. It depends on the persona. An administrator will hate to be forced typing "select * from public.tab..." because he is using these mostly. A normal developer would hate to get the 2000 synonyms shown, all with generic names, instead of the objects in his schema. I decided to focus on the latter persona.

Regarding side effects, I have tested as much as I can. Hana builds on the Generic classes so does SQL Server and two more. As they do not add a public synonym schema (yet), there shouldn't be a negative side effect. Does any of these databases even support public synonyms? For local synonyms it depends on how those database supporting those (SQL Server only) behave. It might actually be that they lack from the same issue as Hana did with synonyms and then this PR would allow the same enhancement there as well.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANAStructureAssistant.java
+++ b/plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANAStructureAssistant.java
@@ -65,7 +65,10 @@ public class HANAStructureAssistant extends JDBCStructureAssistant<JDBCExecution
         GenericSchema parentSchema = parentObject instanceof GenericSchema ? (GenericSchema) parentObject : null;
 
         if (objectType == RelationalObjectType.TYPE_TABLE)
-            findTablesByMask(session, parentSchema, objectNameMask, caseSensitive, maxResults, result);
+        	/* 
+        	 * A TYPE_TABLE is requested in any kind of select statement. Hence it should return tables, views and (public) synonyms
+        	 */
+        	findObjectsByMask(session, parentSchema, objectNameMask, caseSensitive, maxResults, result);
         if (objectType == RelationalObjectType.TYPE_VIEW)
             findViewsByMask(session, parentSchema, objectNameMask, caseSensitive, maxResults, result);
         if (objectType == RelationalObjectType.TYPE_PROCEDURE)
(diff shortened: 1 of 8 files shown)


kai-morich (changes requested):
deleted. see next comment at exact code location

plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANAStructureAssistant.java
```diff
@@ -65,7 +65,10 @@ protected void findObjectsByMask(JDBCExecutionContext executionContext, JDBCSess
         GenericSchema parentSchema = parentObject instanceof GenericSchema ? (GenericSchema) parentObject : null;
 
         if (objectType == RelationalObjectType.TYPE_TABLE)
-            findTablesByMask(session, parentSchema, objectNameMask, caseSensitive, maxResults, result);
+        	/* 
+        	 * A TYPE_TABLE is requested in any kind of select statement. Hence it should return tables, views and (public) synonyms
+        	 */
+        	findObjectsByMask(session, parentSchema, objectNameMask, caseSensitive, maxResults, result);
```
kai-morich, line 71: The object types are choosen in the 'DB metadata' search dialog, so here should not find other types.
If searching without schema might be ok to include synonyms with correct target object type in the result.

A completely different solution would be adding 'Synonym' as searchable object type.

kai-morich (changes requested):

plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANAStructureAssistant.java
wernerdaehn: Done.

serge-rider:
We have similar issues with Oracle public synonyms.
I need to explore it a bit more.

wernerdaehn:
Hmmmm. Yes and No. I have made changes in the Generic* classes but the Oracle classes are not using those. There you use your own classes, implementing the interfaces directly.
Hence I would assume these two solutions do not share anything and there is no "global" solution solving both at once.
The root cause are the getChildren()/getChild() methods in the interface definition. Everybody can implement whatever he wants there. For a proper solution we would need to remove those and have methods like

- getTable()
- getView()
- getSynonym()
- getPublicSynonym()
- getFromCandidate() .... which is the union of above

Looking forward to hear how close to reality I came with my statement. ;-)

serge-rider:
I need to review it after 8.0 release. Too major code impact and we'll probably change synonyms model by that time.

LonwoLonwo:
So. Based on your code, another PR was created. https://github.com/dbeaver/dbeaver/pull/11825
Fix will be available in version 21.0.2  [April 05, 2021]
The main tasks for synonyms in autocompletion were solved.
And also synonyms were added to the list of DB Metadata Search. 
Thank you for your contribution!

Issue #10099, opened 2020-10-19
MSSQL + DBeaver creating indexes...


ravenlost:
Hello, 

How can I create a non-clustered index ? When I right click on the Indexes folder from a table, select **Create New Index**, I cannot change the Type. It's grayed out on "Other". 

Also, it would be nice if we could give a custom name to our indexes right from that screen.

P.S. running DBeaver 7.2.3 on Linux.

Thanks!!
Pat

uslss:
seems it is not implemented.
Your issue is marked as a feature request.
Thanks for suggestion

ravenlost:
> seems it is not implemented.
> Your issue is marked as a feature request.
> Thanks for suggestion

Thank you.

LonwoLonwo:
I added two general types to choose from - non-clustered and clustered. And left the default (that is, in fact, non-clustered)

ravenlost:
> I added two general types to choose from - non-clustered and clustered. And left the default (that is, in fact, non-clustered)

Awesome! Thank you :+1:

uslss:
verified

PR #10333, opened 2020-11-12
#10099 ability to choose Index type via Index UI creation added

LonwoLonwo:
2020-11-12 20_02_04-DBeaver Enterprise 7 3 0 - TestTable


serge-rider:
Please turn Other into Default and make it the default option

Issue #10432, opened 2020-11-19
Folding doesn't refresh after connection is changed


ShadelessFox:
#### System information: 
- DBeaver 7.2.5

#### Connection specification:
- Database name and version: Oracle 12
- Do you use tunnels or proxies (SSH, SOCKS, etc)? SSH

#### Describe the problem you're observing:
Block folding doesn't refresh according to the current connection's dialect.

#### Steps to reproduce, if exist:
1. Connect to Oracle database
2. Open SQL editor and paste this code snippet: 
```
DECLARE
   TYPE timerec IS RECORD (hours SMALLINT, minutes SMALLINT);
   TYPE meetin_typ IS RECORD (
      date_held DATE,
      duration  timerec,
      location  VARCHAR2(20),
      purpose   VARCHAR2(50));
BEGIN
  NULL;
END;
````
3. Save script, but don't close it, instead close DBeaver itself
4. Launch DBeaver, script will be opened

* Expected: Folding works as expected, showing only one anchor
![image](https://user-images.githubusercontent.com/35821147/99648319-f7a75680-2a63-11eb-9769-d721b87c9b60.png)

* Actual: Folding shows three anchors in total, ignoring current dialect's block-related redefinitions
![image](https://user-images.githubusercontent.com/35821147/99648526-30473000-2a64-11eb-90ae-b1449d55edb5.png)

#### Workaround
* Update script file by typing any symbol, forcing folding to update
* Disable and enable again folding manually in script's context menu "Folding Enabled" (right click on the line numbers)

akilovich:
Fixed.

PR #10458, opened 2020-11-22
#10432 reconcile sql file on datasoure change

flambchap:


serge-rider:
Very unclear logic in SQLReconcilingStrategy.getQueries
Starting from method name (it definitely not just gets queries but do some other work)

PR #10486, opened 2020-11-24
Fix navigate association not working on mysql 8 in some cases

tmalahie:
In mysql 8 driver, navigation association does not work well when the foreign key is not referencing to the primary key of the target table. You can reproduce it by creating a MySQL 8 database and running [this script](https://timothe.malahieude.net/dbeaver-bug-mysql8.sql). The foreign key link for customer_bill_id in table product does not work.
This is due to a bug in mysql 8 driver itself, (it always return PRIMARY as PK_NAME in getImportedKeys method).
I managed to fix this by changing the way the referenced pk is retreived in DBeaver. I basically switched 2 blocks of code and added an extra check to increase the chance of actually retreiving the right pk.
Feel free to merge :)


serge-rider:
Good workaround.
Thanks for the contribution!

kseniiaguzeeva:
verified (using example above)

PR #10532, opened 2020-11-28
MySQL user editor: make elements on user properties tab resizeable.

SubOptimal:
On the properties tab in the MySQL user editor the elements had a fixed width and were not resizable.
dbeaver_20201128-090853

This PR makes them resizeable.
dbeaver_20201128-090937


serge-rider:
We need to test it on different OSes more.

PS Thanks for the contribution!

PR #10622, opened 2020-12-06
allow to filter connection aliases fetched from TNS file

bdshadow:
in our company, we have a tnsnames.ora file which is almost 26 000 lines and has more than 5 000 connections. It's very inconvenient to search for the name you want because no filtering is provided in the swt Combo.

Looking forward to your suggestions and improvements.

Signed-off-by: Dmitrii Bocharov <bdshadow@gmail.com>


bdshadow:
although works fine when launched, unfortunately org.eclipse.swt.internal.gtk.GTK can't be used and i didn't manage to add it to the manifest. Another solution needed

serge-rider:
GTK can't be used directly. DBeaver is multiplatform. 
This code won't work under Windows or Mac.

serge-rider:
Possible workaround is to add a button which will open a popup dialog with TNS name list + filter.

bdshadow:
@serge-rider got it. Thanks for the advice. I will do it

Issue #10020, opened 2020-10-08
Use Data Formatting for Generated SQL


uslss:
**Is your feature request related to a problem? Please describe.**
For now timestamp format for a data source affects only the display format not the format for generated sql.
![image](https://user-images.githubusercontent.com/31996417/95474498-d5fb7f80-098d-11eb-92d9-c862dbd9b2b7.png)

**Describe the solution you'd like**
Date, Timestamp, and other data format in generated SQL should be similar to data format in Grid

sovittyagi:
Hi All- Can we consider it on Priority and make it available in early release. Thanks for your help and support.

LonwoLonwo:
New checkbox added in SQL Generate Insert and SQL Generate Update dialogues. 
![102359358-bbfba000-3fc1-11eb-9d84-100716208058](https://user-images.githubusercontent.com/45152336/102398167-fa0fb880-3fef-11eb-8553-c8254ead14b4.png)

Enabling this setting will change the data in the dialog, but only if the default formatting setting in preferences is unchecked.
![2020-12-16 22_47_36-Preferences](https://user-images.githubusercontent.com/45152336/102398874-e6188680-3ff0-11eb-8bb8-a710fa7e8258.png)

uslss:
verified

LonwoLonwo:
Implemented in 7.3.1.
Will be available [here](https://dbeaver.io/) at 2020-12-21

sovittyagi:
Thanks to accept my request and consider for implementation.

Once again thank you.

LonwoLonwo:
@sovittyagi 🎄

PR #10706, opened 2020-12-16
#10020 data format in Generate SQL insert/update dialog added

LonwoLonwo:
2020-12-16 17_03_45-Window


serge-rider:
WE need additional config option to toggle ui/native formatting

PR #10819, opened 2020-12-28
HANA specific connection page

kai-morich:
- calculate port number from edition (+ instance number)
- database name

examples:
image001
image002


serge-rider:
Thanks for the contribution!
Looks like a very handy feature for HANA (as it has a tricky connection params).
We'll try to include it in the next release.

serge-rider:
I've added a couple of minor UI fixes in the connection page

PR #10827, opened 2020-12-29
Fix npe on DBSTypedObject getTypeName

reugn:


serge-rider:

PR #10925, opened 2021-01-11
Add Trino driver configuration

mosabua:
I can add the icon as well, but will need clarification if it needs a transparent background or can have a white one. And if transparent.. is there a background color this has to be optimized for. 

We can use the square icon from https://github.com/trinodb or I can get a special one made by the designer team. 

Icons added now. 

Presto SQL has been rebranded to Trino. For more info https://trino.io/blog/2020/12/27/announcing-trino.html
The release 351 is all Trino. 350 was the last Presto version. I am not sure how you want to take care of that in the UI. 

Also related... let me know if I can help with CloudBeaver addition..


mosabua:
Related to https://github.com/dbeaver/dbeaver/pull/6265

findepi (approved):
LGTM.

Is there a way to mark PrestoSQL driver as deprecated?

note that Trino driver is not a drop-in replacement, as the protocol had to be changed in backwards incompatible manner.

serge-rider:
There is no way to mark a driver as deprecated.
Driver can be replaced (there is a tag `<replace provider="..." driver="..."/>`). It hides original driver and switches all configured connections to a new driver.

However, as I understood, it is not applicable to this new driver as we should keep old Presto driver for backward compatibility.
In fact we have two Presto drivers: PrestoDB (official) and PrestoSQL (community). I suppose both of them remain untouched.

serge-rider:
@mosabua 
In regards to icon: they must have transparent background. Can be optimized for white background but still should be readable on a dark background.

If you think that this PR is fine then I'll merge it in the next release branch.

Presto community drivers will be added in the next major version of CloudBeaver community (along with some other community drivers). So, Trino driver will be also added there.

PS Thanks for the contribution!

serge-rider:
However I think we need to add some Presto reference in the Trino driver description. This will let users to find it in the driver list (if they are not aware about this rebranding).

mosabua:
* yes, at this stage both Presto drivers should remain (well personally PrestoDB should go but ... hahahah)
* @aliloney do you think the images are fine or should we improve them based on the feedback above?
* with regards to the description, how long can/should that be, add something like 

```
                   description="Trino JDBC driver, equivalent to Presto 351+"
```

* I could also update the Presto community one that 350 is last version and with a note to look at Trino one for newer driver?

Maybe @martint or @findepi have suggestions for these texts?

findepi:
"Trino JDBC Driver (supersedes PrestoSQL JDBC Driver)"?

@mosabua i'd avoid word "equivalent" not to suggest backward compatibility which we couldn't maintain. Not sure whether my "supersedes" is any better.

nineinchnick:
The icons don't have a transparent background and will look weird when a dark theme is used:
obraz

serge-rider:
Yes. Icons must have transparent background.
Another option: make some sort of border/substrate, similar to db2 icons.

mosabua:
Let me update the PR after I chat with @aliloney and work with her to get the images updated as needed. I will also update the text.

mosabua:
Images and wording updated now @serge-rider

uslss:
@mosabua could you check if driver works in EA version https://dbeaver.io/files/ea/ ?

mosabua:
Screen Shot 2021-01-28 at 18 05 40

Tried with local Trino 351.

* create connection
* search trino
* download jdbc driver
* show catalogs
* run query

All works as expected at this stage. Awesome!

Issue #9526, opened 2020-08-13
Geopackage (sqlite+spatial) as true DBeaver citizen


rduivenvoorde:
**Is your feature request related to a problem? Please describe.**

Geopackage, is a sqlite-based OpenData (http://www.geopackage.org/) format used as a filebased spatial personal database.
But there is no native 'geopackage' connection possible, and we cannot view the geometries in it (like we do in Postgis) :-)
You CAN open a Geopackage file (*.gpkg) using the sqlite driver, but there is no 'spatial' knowledge in it.

**Describe the solution you'd like**
It would be supercool to be able to handle a Geopackage file just like a Postgis database, so: create a connection to it, fire some queries (maybe some triggers to automagically do some metadata registration).
And a spatial data viewer tab in the data tab would be supercool too

**Describe alternatives you've considered**
You can open geopackages using gdal-based tools (like QGIS (qgis.org)). But given I manage most of my databases (Postgis, Oracle etc) with DBeaver it would be nice if there is some 'native' driver

**Additional context**
Attached a Geopackage file (crs: EPSG:28992, Dutch National coordinate system) with 3 layer (point, line, polygon) create with QGIS (by 'exporting' some old data to the aa.gpkg)

If you need more information, feel free to contact me.

[nl.zip](https://github.com/dbeaver/dbeaver/files/5070852/nl.zip)

kseniiaguzeeva:
Thank you for the suggestion

ShadelessFox:
Basic support for GeoPackage was added:
![java_N2VBDvnbuo](https://user-images.githubusercontent.com/35821147/105096620-67fc4380-5ab8-11eb-9ea4-fed1bd0ddae2.png)

uslss:
verified

uslss:
verified

PR #11048, opened 2021-01-19
#9526 Add support for GeoPackage geometry data

ShadelessFox:


mbernasocchi:
@ShadelessFox do I see correctly that the support is for DML only? or did I miss something and DDL statements (like adding a geometry column) should be supported as well?

ShadelessFox:
It's not related to DML/DDL. In SQLite, you can specify any column time even if it's not supported by SQLite.

Regarding geometry data, editing is not supported.

mbernasocchi:
> In SQLite, you can specify any column time even if it's not supported by SQLite.

ah, thanks a lot for the quick answer I got confused by the dropdown showing only "standard" types
image

so you can basically just type `MULTIPOLYGON` in the type column and then manually run the SQL to add metadata into gpkg_contents (with meaningful numbers for min and max x and y):
`insert into gpkg_contents values ('column1', 'features','some description','sample_attribute_with_geometry','2019-02-19T10:49:06.022Z',74572.400000,6631280.000000,732294.000000,7774890.000000,3856);`

and add the column metadata into gpkg_geometry_columns

`insert into gpkg_geometry_columns values ('column1','geometry','MULTIPOLYGON',3856, 0, 0 )`

is there any plans to make this automatically for geopackages? (I looked the issue queue but could find only #9526 )

ShadelessFox:
Besides metadata, there's data. GeoPackage uses a special binary storage format; you can't simply insert strings like `POINT(1 2)`.

In #9526, only a read-only data view was implemented. Adding editing support is a difficult task, and perhaps it's something DBeaver is not supposed to support.

PR #11079, opened 2021-01-21
use recent version of mariadb-connector-j

eworm-de:
The MariaDB connector used to select an odd collation for the
connection when the server is configured to use non-mb4 utf8.
This was fixed upstream [0] and made its way into version
2.6.1. That's the minimum version I need, but let's just bump to the
recent version 2.7.1.

[0] https://github.com/mariadb-corporation/mariadb-connector-j/commit/6edc091e7e034b6ba3a764f30391de827179c96a


serge-rider:
Thanks for the important update. 
We'll include it in the next dbeaver version.

Issue #10694, opened 2020-12-15
Wrong export format of postgis data


ulrobix:
#### System information: 
- Operating system (distribution) and version
Windows 10 20H2 Build 19042.685

- DBeaver version
7.3.0
- Additional extensions
No

#### Connection specification:
- Database name and version
PostgreSQL 12.4 (Ubuntu 12.4-1.pgdg18.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0, 64-bit
from dockerhub: mdillon/postgis
- Driver name
- Do you use tunnels or proxies (SSH, SOCKS, etc)? 
yes ssh
db start in docker on wsl 2

#### Describe the problem you're observing:
Im export results of query to sql format, but export postgis geometry data in wrong format:

```
INSERT INTO points (id,route_group_id,"name",point,"type",parent_point) VALUES
	 (97507,1782,'Остановка "ГПТУ-№3"','''SRID=4326;POINT (49.566987 54.237836)''::geometry','B',97703),
```

look at '''SRID=4326;POINT (49.566987 54.237836)''::geometry'
wrong quotes
When I Import this sql i have error:

```
 SQLSTATE[XX000]: Internal error: 7 ERROR:  parse error - invalid geometry
  LINE 3:   (97507,1782,'Остановка "ГПТУ-№3"','''SRID=4326;POINT (49.5...
                                              ^
  HINT:  "'S" <-- parse error at position 2 within geometry

```

in old version 7.2.3 correct format:
```
INSERT INTO points (id,route_group_id,"name",point,"type",parent_point) VALUES
	 (97507,1782,'Остановка "ГПТУ-№3"','SRID=4326;POINT (49.566987 54.237836)'::geometry,'B',97703),

```

I tried 7.2.5 version, it have the same problem.

#### Steps to reproduce, if exist:

- Make sql query with postgis data "SELECT * FROM points"
- Click to export button in bottom of results frame.
- Chose sql format, but not change config of export
- Select export to clipboard
- Run export.
- Paste  exported data to notepad

kseniiaguzeeva:
Thank you for the report

flambchap:
Fixed.

kseniiaguzeeva:
verified

PR #11080, opened 2021-01-21
#10694 do not escape dbgeometry

flambchap:


serge-rider:
We need better type check. 
- What OTHER types need to be quoted as strings?
- Can we just check for `'` character in the beginning of strValue?

Issue #10570, opened 2020-12-01
GIS viewer support for PrestoSQL


samueltqx:
Is it possible to support Spatial Data viewer for PrestoSQL? 
The `H2GISGeometryValueHandler` can be reused for handling Geometry type from Presto. Since Presto is in the Generic plugins, can the GeometryValueHandler be added to it, or should it be a separate plugin?

kseniiaguzeeva:
Yes, it is possible. We are going to implement it in 7.3.3 release.

kseniiaguzeeva:
More tickets about gis support in this milestone:
https://github.com/dbeaver/dbeaver/issues/10661
https://github.com/dbeaver/dbeaver/issues/10660
https://github.com/dbeaver/dbeaver/issues/10573

LonwoLonwo:
Hello @samueltqx 
What kind of column type are you using for spatial data?
I can add map support for binary data types.

samueltqx:
Hi @LonwoLonwo 
I am using both text and binary for spatial data but can convert to Geometry object in Presto/Trino.

LonwoLonwo:
@samueltqx Could you provide some examples for testing please? Table DDL with some data.

LonwoLonwo:
@samueltqx also can you try change view/format settings for text columns like on the picture?

![2021-01-25 14_49_49-](https://user-images.githubusercontent.com/45152336/105702414-d019b200-5f1c-11eb-91a0-5a775a5fc0a1.png)

In theory, when switching to the "geometry" type, a "Spatial" tab should appear.

samueltqx:
@LonwoLonwo yes I'm able to switch to geometry type from text column in Presto and the Spatial viewer works! Was this feature added recently? I was looking at the GenericValueHandlerProvider which is why I assumed there isn't support for Spatial yet.

samueltqx:
However I did notice an odd bug where I was only able to switch to Geometry type when I alias the column name. 

Without alias, there's no option to set format.
![without alias](https://user-images.githubusercontent.com/12363062/105706616-aacf7880-5f09-11eb-88e9-d639caf8f630.png)

With alias, I am able to. 
![with alias](https://user-images.githubusercontent.com/12363062/105706784-e8cc9c80-5f09-11eb-82d2-4919bfebaf7a.png)

The metadata for both results shows the type as Varchar. But without aliasing, it seems to show the column as Unknown type in the Grid.

LonwoLonwo:
[...]

PR #11099, opened 2021-01-22
#10570 varbinary into geometry case added

LonwoLonwo:
varbinary into geometry case added

2021-01-25 16_40_27- metadata - DBeaver Enterprise 7 3 3 - address


serge-rider:
We need to convert displayString as well.

PR #11101, opened 2021-01-22
feat: add gcloud PostgreSQL driver

olavloite:
Adds a driver for Google Cloud SQL PostgreSQL instances. Connecting to such instances requires generating SSL certificates and IP whitelisting, or using the [cloud-sql-jdbc-socket-factory](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory#postgresql-1). This driver automatically includes the cloud-sql-jdbc-socket-factory and allows users to connect to a cloud instance by merely adding the instance and database name.

Screenshot from 2021-01-22 22-03-40


olavloite:
The same can also be done for Google Cloud SQL MySQL and SQL Server, but I thought I would hold off on any PR's for those until this has been approved (or denied). Please let me know if you would prefer that I submit PR's for those as well already now.

serge-rider:
Thanks for the contribution!

We'll probably include this improvement in dbeaver 7.3.5.
We need to to test UI changes and test with real GC PG.

serge-rider:
I think we need to patch PostgreDataSourceProvider.getConnectionURL for cloud PG (remove explicit port number from URL).

Also please merge from devel, there is a dummy merge conflict (you need to include all changes).

olavloite:
> Also please merge from devel, there is a dummy merge conflict (you need to include all changes).

I've merged with devel and fixed the merge conflict.

> I think we need to patch PostgreDataSourceProvider.getConnectionURL for cloud PG (remove explicit port number from URL).

I have the impression that that is already taken care of here: https://github.com/dbeaver/dbeaver/blob/3a7892bd3ff2bcbfcd171c9f9a96252f0c6e58f8/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreDataSourceProvider.java#L89-L92

I also didn't run into any problems with it when I tested this with a Google Cloud PostgreSQL instance. I was able to get a connection and run queries and updates on it. Or am I missing something here?

serge-rider:
Makes sense.
I think we can merge it and make more testing.

PR #11171, opened 2021-01-29
Fix error when pasting more rows than current number of rows in ResultSetViewer

tmalahieude:
When I use Advanced Paste in a table to add new rows, I have to ensure that I leave enough room for the rows I am creating, otherwise it shows me an error dialog:
image
It was not the case in 7.3.2 version of dbeaver. The error seams to come from this commt: https://github.com/dbeaver/dbeaver/commit/f740fb41e970a382eddefd3c49465cb8144f1e25

I found a way to fix it by adding an out of bounds check in LightGrid.posToCell (make sure that target position is less than grid length). Not sure if it's the most relevant fix but it seams to make sense since there is already a check in the other way (make sure that position is greater than 0).

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/LightGrid.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/LightGrid.java
@@ -4254,7 +4254,9 @@ public abstract class LightGrid extends Canvas {
         }
         List<GridCell> cells = new ArrayList<>(selectedCells.size());
         for (GridPos pos : selectedCells) {
-            cells.add(posToCell(pos));
+        	GridCell cell = posToCell(pos);
+        	if (cell != null)
+        		cells.add(cell);
         }
         return cells;
     }
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/LightGrid.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/LightGrid.java
@@ -671,6 +671,9 @@ public abstract class LightGrid extends Canvas {
         if (pos.col < 0 || pos.row < 0) {
             return null;
         }
+        if (pos.col >= columnElements.length || pos.row >= rowElements.length) {
+            return null;
+        }
         return new GridCell(columnElements[pos.col], rowElements[pos.row]);
     }
 


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/LightGrid.java
```diff
@@ -671,6 +671,9 @@ public GridCell posToCell(GridPos pos)
         if (pos.col < 0 || pos.row < 0) {
             return null;
         }
+        if (pos.col >= columnElements.length || pos.row >= rowElements.length) {
+            return null;
+        }
```
ShadelessFox, line 676: Please add sanity check that returned _cell_ is not null at:
https://github.com/dbeaver/dbeaver/blob/b4e57852deeeb175ca6ffb43357d23fa13c2ec1e/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/LightGrid.java#L4256-L4258
```

ShadelessFox (changes requested):

bnickelsen:
Is there an easy work around for this issue, until a fix is released?

ShadelessFox:
Hello @bnickelsen,

I don't think so.

Also I'm having hard time reproducing this issue, can you share exact steps, so I can reproduce it? Because currently I'm just hoping for the best.

bnickelsen:
Starting with a table open in the Data view.
image
Use the add line icon at the bottom to add one line.
image
I then have a spread sheet with all the data that I want to put into this table. So high ight the spread sheet and copy. Then in the data view Select the first cell of the line you added and paste. that is when I get the error. The issue seems to be that with this last update when you past more lines then you have added you get this error. For some reason it is not adding the new lines. the work around for me was to add enough line to fit eth data I was pasting 1027 row later. no error.

So again a work around is to add enough lines to fit the data you are pasting.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/LightGrid.java
tmalahieude: Thnaks for the insight @ShadelessFox. Do you know if there are other sanity checks to do? The method seams to be used at many places... Maybe I should actually revert my changes and put the out of bounds check at a higher level in the code to avoid breaking something else?

ShadelessFox: If you're using Intellij IDEA, you can highlight method and press `Alt+F7` to find all the usages of it. I've done it already, so this is the only place where null check is missing.
This method is marked as _nullable_ and there's a similar check above, so everything is OK. Actually, your PR is self-contained, you didn't break anything, but adding this little sanity check would be nice.

ShadelessFox:
Thank you for response, I will investigate later.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/LightGrid.java
tmalahieude: Ok thanks :) then if you're confident I'll just add the extra sanity check you're requesting. I double checked the method usages and you seam to be right, there is already a null check everywhere else.

tmalahieude: Done

Issue #10573, opened 2020-12-01
Allowing users to have custom Map Tile source


chungcf2006:
**Is your feature request related to a problem? Please describe.**
I got a project which I need to plot the spatial data on a specific map (not in the list shipped with DBeaver)

**Describe the solution you'd like**
Having a simple Map Tile Manager can solve the problem.
The manager should allow users to customize the tile source URL and the zoom level (as what the user can set in TileLayer under leaflet.js) 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

kseniiaguzeeva:
Thank you for the suggestion.

mkgrgis:
Duplicate https://github.com/dbeaver/dbeaver/issues/10531 ?

flambchap:
Yes, it is. Thank you.

flambchap:
Implemented.

kseniiaguzeeva:
verified

PR #11184, opened 2021-02-01
#10573 add leaflet tiles management

flambchap:


flambchap:
I believe that the current complaints from Codacy are false-positives.

PR #11234, opened 2021-02-04
Data Virtuality Driver added

sebcom-gmbh:


plugins/org.jkiss.dbeaver.ext.datavirtuality.ui/src/org/jkiss/dbeaver/ext/datavirtuality/views/DataVirtualityConnectionPage.java
```diff
@@ -0,0 +1,290 @@
...
+        if (dbText != null) {
+            String databaseName = connectionInfo.getDatabaseName();
+            if (CommonUtils.isEmpty(databaseName)) {
+                databaseName = DataVirtualityConstants.DEFAULT_DB_NAME;
+            }
+            dbText.setText(databaseName);
+        }
+        // TODO: need to fix for checkbox
+        if (sslCheckbox != null) {
+            sslCheckbox.setSelection(CommonUtils.notEmpty(connectionInfo.getProviderProperty(DataVirtualityConstants.PROP_SSL))=="mms" ? true : false);
```
aimiebell, line 207: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [ES: Comparison of String objects using == or != (ES_COMPARING_STRINGS_WITH_EQ)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#es-comparison-of-string-objects-using-or-es-comparing-strings-with-eq).
This code compares java.lang.String objects for reference equality using the == or != operators. Unless both strings are either constants in a source file, or have been interned using the String.intern() method, the same string value may be represented by two different String objects. Consider using the equals(Object) method instead.

serge-rider:
Generally looks great 👍 

- Please fix warning reported above (usage of == to compare strings)
- Pass dialect ID to GenericDialect (this is a recent change)
- Keywords can't have spaces inside ("SIMILAR TO" - these are two different keywords).
- - If possible please rename plugin `org.jkiss.dbeaver.ext.datavirtuality.core` into `org.jkiss.dbeaver.ext.datavirtuality`. We no longer use core suffix for headless extensions (and going to rename all remaining ones).
- 
Thanks for the contribution!

serge-rider:
@sebcom-gmbh any update?

sebcom-gmbh:
Hi @serge-rider , sorry I did not noticed that my pull request already landed in your repos. I will fix the mentioned points and repull

sebcom-gmbh:
Hi @serge-rider, where can I find the warning you mentioned  (usage of == to compare strings)? Thanks

sebcom:
Hi @serge-rider, I did the requested changes. Please let me know if something is still missing, I'm wondering where should the link to the JDBC driver point to? Can we upload the driver to the maven repo. Or shall I ask the vendor for an official download link? Sebastian

serge-rider:
Hi @sebcom
Great job, now it looks much better 👍 

A few minor issues:
- Please remove changes in file org.jkiss.dbeaver.cmp.simple/pom.xml (probably added by accident)
- Delete driver `datavirtuality_generic` definition from generic plugin. It doesn't makes sense as you declare it separately in the new plugin.
- Update copyrights to 2021 year (2010-2021)
- DataVirtualityConnectionPage should work with icon like it was done in HANAConnectionPage. This will also make class DataVirtualityUIActivator useless - you could remove it (and remove reference on it from manifest).

Regarding driver jar.
For now you could use direct https link (on some particular version on page https://datavirtuality.com/download-driver/#). You cannot use paths like C:\Users\admin\.

But Maven artifact is much more preferred (it will let to choose driver version from UI and update it automatically).
You cannot add maven artifact by yourself, you should ask vendor to do this.

sebcom-gmbh:
Hi @serge-rider 

Regardin g the download URL, the vendor came up with the idea to point to the DB server instance directly. Because in case of DV they have an built-in web portal with all the needed drivers and software. So like this it would always point to the correct versioin of the JDBC driver.

Sebastian

serge-rider:
> Regardin g the download URL, the vendor came up with the idea to point to the DB server instance directly. Because in case of DV they have an built-in web portal with all the needed drivers and software. So like this it would always point to the correct versioin of the JDBC driver.

Unfortunately it won't work this way. Parameter {host} is not available at the moment when DBeaver downloads a driver. Moreover host can be set to localhost in case of SSH tunnel usage.

So I think we should use direct link on the public  website (until public Maven artifact will appear).

The rest looks fine.

sebcom-gmbh:
Hi @serge-rider 

I added now the link to the latest driver, but their is also an official website with the links to the diffrent versions:

https://datavirtuality.com/download-driver/

What do you think is better?

Sebastian

serge-rider:
Link on exact driver version is ok for now. DBeaver will download driver automatically and will be able to connect without any additional configuration.
Note: currently link is broken: `http://https://datavirtuality.com/wp-content/uploads/sites/2/2020/03/datavirtuality-jdbc_2.3.8.jar`, please remove http in the beginning.

Issue #11081, opened 2021-01-21
Display Sequences for MariaDB


Mosk915:
I believe this is similar to #2534 but for MariaDB databases.  Would it be possible to display sequences in the navigator pane? This is for the community edition of DBeaver.

LonwoLonwo:
https://mariadb.com/kb/en/create-sequence/
https://mariadb.com/kb/en/sequence-overview/

LonwoLonwo:
Implemented in DBeaver 7.3.5

uslss:
verified

PR #11235, opened 2021-02-04
#11081 sequence support for mariaDB > 10.3 added

LonwoLonwo:


serge-rider:
We need to filter out sequence tables

Issue #8506, opened 2020-04-21
GIS data view: restore position and zoom level after changing tile layer


mkgrgis:
**Is your feature request related to a problem? Please describe.**
If there is switching of different layers, we must focused on the same area for each layer.

**Describe the solution you'd like**
For each GIS data view map we can use some variables for center coordinates (phi/lam) and z coefficient. This variables must initialize new layer and must be actualized by moving/setZoom/setCenter events of leaflet map.

**Additional context**
GIS data viewer. The simplest environment for diagnostic is PostGIS.

uslss:
thanks for suggestion

LonwoLonwo:
Hello @mkgrgis 
Сould you provide more information about your idea please?
I don't quite understand what is meant by switching layers. Is it a map type switch? Or switching Street/Topography?
Because in these two options, it seems that we are already centered.
![2021-01-25 11_26_28-runtime-DBeaverEnterprise product - DBeaver Enterprise 7 3 3 - planet_osm_polygo](https://user-images.githubusercontent.com/45152336/105680879-6dff8380-5f01-11eb-824f-b63c32857183.png)

mkgrgis:
> I don't quite understand what is meant by switching layers. Is it a map type switch? Or switching Street/Topography?

Both of them. The word layers meant as in [leaflet](https://ru.wikipedia.org/wiki/Leaflet) documentation. There was problem if anybody try to focus on area of object. Then if any other layer (for ex. OpenTopoMap) was selected, focused area haven't  saved for new map layer.

> Because in these two options, it seems that we are already centered.

The problem was not only in centred map, but in reconfiguration of observation window in case of changing of map type. For example it can be selected area without the current DB object, if user selected such area. There is similar feature request for autozoom by DB object if the first map layer initialized. But than current observation window must be saved between different map layers (providers).

LonwoLonwo:
@mkgrgis Okay, I got the point. We'll see what we can do about it.

mkgrgis:
In my tests 7.3.4 has saved z and centre of map between different layers, but **not between different providers**, where the map parameters always set for focused object from GISs data.
![изображение](https://user-images.githubusercontent.com/41448637/106749420-89703a00-6637-11eb-9c90-3dd804b7a7ff.png)
I'm waiting on behaviour with layer's management system, was added some day ago.

mkgrgis:
[...]

PR #11244, opened 2021-02-06
Tiles#10573

flambchap:


mkgrgis:
There is [some other notices](https://github.com/dbeaver/dbeaver/issues/8506) about management of tiles and tile providers.

PR #11259, opened 2021-02-08
Postgre types#10684

LonwoLonwo:
2021-02-08 01_50_32-Edit attribute _Column1_


serge-rider:
We must move data type cache to database

Issue #11333, opened 2021-02-13
PostgreSQL: Connection setting "Show all databases" make session connect to wrong database


haugs:
#### System information: 
- Windows 10 Version 2004
- DBeaver version 7.3.0
- no extensions

#### Connection specification:
- PostgreSQL 11.6 (Debian 11.6-0+deb10u1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
- Driver name: PostgreSQL JDBC Driver
- corporate proxy

#### Describe the problem you're observing:
When I connect to a PostgreSQL Database host that runs multiple database (yes databases, not schemas) the session is not connected to database specified in the connection settings ("Main"). Instead it is connected to a different database as I found out using the statemet `SELECT current_database();`
The GUI marks the database from the connection settings in **bold** which is even more misleading.
I believe the problem only arises when the option "Connection settings > PostgreSQL > Show all databases" is selected.

#### Steps to reproduce, if exist:
1. Connect to a database host that is running multiple databases with a DBA role (e.g. postgres) with the option "Connection settings > PostgreSQL > Show all databases" selected. You may leave the field "Connection settings > Main > Database" empty for now.
2. Verify which database the session is connected to with `SELECT current_database();` It is understandable that any database might be the current one as no specific one was specified.
3. Edit the connection settings and specify a different database ("Connection settings > Main > Database") and reconnect.
4. Observe that the setting from step 3 did not take effect by running `SELECT current_database();`

#### Include any warning/errors/backtraces from the logs
none available

kseniiaguzeeva:
It is the expected behavior at this time. But we should make an enhancement in the future.

haugs:
Thanks for looking at this issue.
I don't really understand how it can be the expected behavior. In the connection settings I select database A and I end up being connected to database B.
I understand although that technically this could be not as simple as it looks for the user.
I guess that "show all databases" changes the way DBeaver connects with the database and that "selecting a database" then needs to be handled differently per type of database.

E1izabeth:
Can't reproduce against DBeaver 23.0.5
If you still experience this issue, please let me know

PR #11334, opened 2021-02-14
#11333 postgres connection still using old database name when config …

dextok:
Fixed postgres using database name form old instance when open new connection

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
@@ -365,7 +365,7 @@ public class PostgreDataSource extends JDBCDataSource implements DBSInstanceCont
         JDBCRemoteInstance instance = context == null ? null : context.getOwnerInstance();
         Connection pgConnection;
         if (instance != null) {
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -376,7 +376,7 @@ protected Connection openConnection(@NotNull DBRProgressMonitor monitor, @Nullab
             try {
                 // Patch URL with new database name
                 if (CommonUtils.isEmpty(conConfig.getUrl()) || !CommonUtils.isEmpty(conConfig.getHostName())) {
-                    conConfig.setDatabaseName(instance.getName());
+                    //conConfig.setDatabaseName(instance.getName());
```
samuei: Let's not leave commented code laying around. If we're removing a line, let's remove the line.

dextok: It has been removed, thanks for your review.

samuei: 👍

serge-rider:
Will get back to it after version 21 release

Issue #11361, opened 2021-02-16
Tip of the day is not displayed on startup


kseniiaguzeeva:
#### Steps to reproduce, if exist:
1. Be sure Show tip on startup is enabled
2. Relaunch DBeaver

Expected: tips are displayed
Actual result: tips are not shown and can be seen via Help menu

flambchap:
It's a macOS-specific issue.

flambchap:
Fixed. Now we do not attempt to show the TotD window if DBeaver is started for the very first time.

kseniiaguzeeva:
verified

PR #11383, opened 2021-02-17
#11361 add ensureDataSourceRegistryLoaded method to DBPProject

flambchap:


serge-rider:
We can't change project metadata load process

PR #11456, opened 2021-02-23
Fix spelling and replacing some words with better alternatives (pt-BR)

brunobastosg:


serge-rider:
Thanks for the contribution 👍

PR #11574, opened 2021-03-02
Fix for webview renderer on Windows if system defaults to old IE browser without javascript enabled

henkeberggren:


serge-rider:
This makes sense.

PR #11637, opened 2021-03-06
feat: add drivers for Google Cloud MySQL

olavloite:
Adds a driver for Google Cloud SQL MySQL instances. Connecting to such instances requires generating SSL certificates and IP whitelisting, or using the [cloud-sql-jdbc-socket-factory](https://github.com/GoogleCloudPlatform/cloud-sql-jdbc-socket-factory/blob/main/docs/jdbc-mysql.md). This driver automatically includes the cloud-sql-jdbc-socket-factory and allows users to connect to a cloud instance by merely adding the instance and database name.

The PR adds drivers for both MySQL8+ as well as MySQL5.x which are the versions that are supported by Google Cloud SQL.

This feature is similar to #11101.

Screenshot from 2021-03-06 14-25-03


olavloite:
@serge-rider I noticed that this PR and also the similar one for SQL Server were closed. Is this feature something you plan to support in some other way? Or is there anything I could/should do to get these PRs reconsidered?

Issue #6558, opened 2019-08-19
Semi-Automatic Updates


Lonniebiz:
In DBeaver Version 6.1.5.201908191054, I have a suggestion that isn't urgent, but may be worthy of future consideration.

I update DBeaver each time a new version is released, and the process of updating is already simple, but the steps of updating could be automated.

Upon selecting "Update" from the help menu, these steps could be done:

1) Check for updates, if none exist, Tell the user they're up-to-date. If updates exist then:
2) Prompt the user to save any unsaved work.
3) Determine how DBeaver is currently installed (version, location of files, etc).
4) Download the correct version based on #3's info.
5) Auto-launch the installer and then immediately close DBeaver.

Again, this isn't urgent to me, and if it doesn't get done at all, that's ok. This is just a suggestion.

serge-rider:
In 6.1.5 we already have new version download. But we don't do 3 and 5.

jspeis:
Would be great to see this! (Came to request the same). I do think 3 & 5 would be nice to have if possible.

nicoschl:
It would be great if everything could be a seamless update of the core of DBeaver plus any of the installed DBeaver plugins without having to do all the manual steps. I know DBeaver is based on Eclipse, where this functionality already exist, I don't need to download a new Eclipse everytime there is a new release or update to a plugin, Eclipse take care of everything.

You would make many people happy if this feature request and #6522 could be implemented.

ethanherbertson:
Just upgraded to 6.3.4 (CE) and upon running the new version for the first time, I was automatically prompted to update/reinstall the plugin I had installed. Seems like progress!

nicoschl:
How did you upgrade @ethanherbertson ? Did you install/extracted the new version "over" the existing version or into a new/clean directory?

I'm asking because I just upgraded to 6.3.5 (CE) by extracting the linux tar.gz package into a new directory and when I run dbeaver I don't get the prompt you are referring to. I'm running Arch Linux with KDE, but I doubt the OS or KDE have an influence here.

ethanherbertson:
I used the Windows installer. I did not do a clean uninstall/reinstall.

ShadelessFox:
Now it is possible to install an update directly from DBeaver on Windows and macOS.

Unfortunately, there is no an easy way to test this until the next version will come out.

nicoschl:
> Now it is possible to install an update directly from DBeaver on Windows and macOS.

Does this include Linux? 

[...]

PR #11668, opened 2021-03-10
#6558 Automatic update mechanism

ShadelessFox:


plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/update/VersionUpdateDialog.java
```diff
@@ -294,4 +378,59 @@ public static boolean isSuppressed(VersionDescriptor version) {
...
+            return "exe";
+        }
+    }
+
+    private static final class MacintoshInstaller implements PlatformInstaller {
+        @Override
+        public void run(@NotNull File file, @NotNull Log log) throws Exception {
+            Runtime.getRuntime().exec(new String[]{
+                "/bin/sh", "-c",
+                "open -F -W " + file.getAbsolutePath() + " && rm " + file.getAbsolutePath()
```
ShadelessFox: To be really sure this probably should be quoted.

PR #11720, opened 2021-03-14
Translation: Use Stringbuilder

CodyBuilder-dev:
Hello! I'm a korean user and I'm really enjoying using dbeaver.
I tried to find a way to contribute and I just found that dbeaver accpets translations for korean.
So I just made my first little translation for test, and it works fine for me.
It would be great if I could keep contributing for dbeaver! Thank you!


serge-rider:
Thanks for the contribution 👍

PR #11774, opened 2021-03-18
Add jdbc driver for APM (Jennifer)

KyongSik-Yoon:
I'm [Jennifer](https://www.jennifersoft.com) APM(Application Performance Management) product developer.
Recently we prepared JDBC driver for provide convenient search to users.

Jennifer use mmap file database made by own.
It's very fast and easy to handle for us. 
But It couldn't provide free for search because we didn't have common language for search.
Surely, Jennifer suggest alternative ways as like Open API, RDB migration.

However Someone don't want to do that.
So I decided to make JDBC driver and finally almost done.
After phase, I'm investigating to JDBC client for user who want free software and I found this nice project.

We have various users and customers In South Korea.
If you approve register new category for us It will be very helpful for us and we suggest this tool to them as default JDBC client tool.


serge-rider:
Hi,
Thanks for the contribution!

A few notes:
- There is no reference on jar download. Maven artifact is the preferred way but direct https link will also work
- I don't think that new plugin is neccessary. It doesn nothing database-specific so you can just add new driver definition in the plugin.xml of Generic plugin. We usually add new plugin only we provider some database-specific functionality (Java code)

KyongSik-Yoon:
@serge-rider 

I wanted just give to users a convenient way without additional configuration.
Because I couldn't find a way how to explain usage for register new driver definition.
They are not developer but manager and meeting with all of them is not possible due to Corona (As you know..)
Actually I want show our driver ready to use. (Include some icon which show identity of product).

Is it possible add this plugin if I put direct link?
Please think about one more time.

serge-rider:
New driver will work fine without new plugin. Generic plugin contains dozens of driver definitions. New plugin just doesn't make any sense as there is no extra code or configuration there.
Just move `<driver ...>` definition to https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ext.generic/plugin.xml.

So pull request will consists of one change in a single file.

KyongSik-Yoon:
@serge-rider Thanks for reply. I didn't know what's role of generic plugin.
Let me check about it please.

KyongSik-Yoon:
@serge-rider Now my separated redundant plugin removed and definition of it placed in generic plugin.
I found this in SQL category after build on my local.
Please review modification.

serge-rider:
All good 👍

PR #11787, opened 2021-03-18
LIMIT Query Transformations - Added support to not transform the queries for some databases, then SELECT does not have a FROM part. Enabled this exclusion for Vertica connections.

vityank:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -1278,18 +1278,22 @@ public static DBCStatement makeStatement(
...
         throws DBCException
     {
+        DBCQueryTransformProvider transformProvider = DBUtils.getAdapter(DBCQueryTransformProvider.class, session.getDataSource());
+        
         // We need to detect whether it is a plain select statement
         // or some DML. For DML statements we mustn't set limits
         // because it sets update rows limit [SQL Server]
         boolean selectQuery = sqlQuery.getType() == SQLQueryType.SELECT && sqlQuery.isPlainSelect();
+        // For some DB's like Vertica LIMIT can't be applied to queries without FROM part - i.e calling meta functions.
+        if (sqlQuery.isPlainSelectWithoutFrom() && transformProvider.getClass().getSimpleName().equals("VerticaDataSource"))
```
ShadelessFox: Thanks for the contribution!

I can suggest you to extend [SQLDialect.java](https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialect.java) and add convenient method to determine whether the `LIMIT` without `FROM` is allowed or not, e.g:
```java
boolean supportsLimitWithoutFrom();
```
And then override it to return `true` exclusively for [VerticaSQLDialect.java](https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ext.vertica/src/org/jkiss/dbeaver/ext/vertica/model/VerticaSQLDialect.java). It would be more generic approach.

After that, you can obtain the dialect and check for its "ability" instead of using hard-coded string.

vityank: Thanks for suggestion, that's definitely better approach. I need to find which of objects available in the modified method are of SQLDialect type. If you have some pro hint of obtaining the dialect from present variables, it would save much time from searching for :)

ShadelessFox: You can obtain dialect from the data source (which is obtained from the available `session` variable in this case):
```java
final SQLDialect dialect = SQLUtils.getDialectFromObject(session.getDataSource());
```

vityank: I think I found it using slightly other access:

session.getDataSource().getSQLDialect()

Is it good too?
Checked it - Builds and works. Pushed updated commit.

ShadelessFox: That's okay too :)

vityank:
Added small change - Moved declaration transformProvider back to original position, as it's no longer used in our feature conditional check.

serge-rider:
@vityank After all I've decided to make new limit transformer specifically for Vertica. See 0d654fd21696ecaa4e5b70c684bb99ecd96cf989 .
I partially used code from your PR.

Issue #11796, opened 2021-03-19
Terminate DBeaver gracefully after starting updater


ShadelessFox:
We've implemented automatic updates in #6558, but the termination routine may lead to the workbench corruption.

ShadelessFox:
Please make sure that installer asks user to close all unsaved data.

kseniiaguzeeva:
The installer doesn't ask it and changes are not saved (tested on Mac)

kseniiaguzeeva:
veirified on Mac and Win10. Separate ticket is created https://github.com/dbeaver/dbeaver/issues/11859

PR #11797, opened 2021-03-19
#11796 Gracefully terminate the workbench before updating

ShadelessFox:


ShadelessFox:
I believe this analysis done by Codacy is outdated, since this entire PR is intended to eliminate the usage of `System.exit`.

PR #11819, opened 2021-03-22
chore: update official driver name and status of community driver

olavloite:
The official driver name should be `Google Cloud Spanner` to correspond with the name of the database and the corresponding JDBC driver.

The community driver has been deprecated and does not receive any updates anymore. Users should therefore whenever possible use the offically supported driver instead of the community driver. See also https://github.com/olavloite/spanner-jdbc#deprecated

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.spanner/OSGI-INF/l10n/bundle.properties
+++ b/plugins/org.jkiss.dbeaver.ext.spanner/OSGI-INF/l10n/bundle.properties
@@ -1,3 +1,3 @@
 Bundle-Vendor = DBeaver Corp
-Bundle-Name = DBeaver Google Spanner Extension
+Bundle-Name = DBeaver Google Cloud Spanner Extension
 datasource.spanner.description = Spanner datasource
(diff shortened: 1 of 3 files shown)


serge-rider:
Thanks for the update.
Is there any value in the community driver? 
If it is 100% replaceable by Google driver then we can add tag
`<replace provider="google_spanner" driver="spanner_jdbc"/>` to the Google driver definition. It will replace community driver automatically (and it won't accessible anymore).

olavloite:
> Is there any value in the community driver?

No, not really. It should not be used anymore and all its features are supported by the Google driver. 

> If it is 100% replaceable by Google driver then we can add tag
> `<replace provider="google_spanner" driver="spanner_jdbc"/>` to the Google driver definition. It will replace community driver automatically (and it won't accessible anymore).

I'm not exactly sure how that works, but I assume that there would be an issue with the fact that the connection string formats between the two drivers differ significantly. So a connection string that worked for the community driver will not work for the Google driver.

serge-rider:
Ok, I see. Then let's leave Community driver there.

PR #11820, opened 2021-03-22
Denedo plugin

sebcom-gmbh:
First version of the Denodo plugin


sebcom-gmbh:
Remark: _JDBC driver is only availble from a login protected HTTP URL!_

sebcom-gmbh:
Funny, my local version compiles error free...

PR #11882, opened 2021-03-25
Denodo Plugin

sebcom-gmbh:


serge-rider:
The plan is to include it in 21.0.3.

We also should test it somehow.

serge-rider:
@sebcom-gmbh driver download link doesn't work. It requires authorization which can't be done in headless mode.
Is there any other link which is publicly available?

Issue #11116, opened 2021-01-26
GIS data is displayed without Z and M values


kseniiaguzeeva:
#### System information: 
- Operating system (distribution) and version Mac
- DBeaver version 7.3.4 EA

#### Connection specification:
- Database name and version PostgreSQL

#### Describe the problem you're observing:
GIS values are cut to show 2D data only.

#### Steps to reproduce, if exist:
Add to Geometry column:

- LINESTRING M (1 1 0, 1 2 0, 1 3 1, 2 2 0)
- POLYGON Z ((0 0 0, 0 1 0, 1 1 0, 1 0 0, 0 0 0))
- POINT Z (-8.6964351 40.3307427 1)
- POINT ZM (1 2 3 4)

Expected:
the values are displayed completely in cells, Value panel, during data transfer.

Actual result:
the values are cut:
- LINESTRING (1 1, 1 2, 1 3, 2 2)
- POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))
- POINT (-8.6964351 40.3307427)
- POINT (1 2)

ShadelessFox:
By some reason, the geometry data like:
```
LINESTRING M (1 1 0, 1 2 0, 1 3 1, 2 2 0)
``` 
is rendered as 
```
LINESTRING (1 1, 1 2, 1 3, 2 2)
```
without zeroes. This may confuse user.

Also `POINT` lacks 3rd dimension if specified.

ShadelessFox:
Fixed:
![image](https://user-images.githubusercontent.com/35821147/113787497-ce9e0e80-9743-11eb-97d1-b08669cb4bfa.png)

kseniiaguzeeva:
verified

PR #12002, opened 2021-04-06
#11116 PostGIS trimmed geometry data fix

ShadelessFox:
java_izJzOfGfIT

### DIFF
--- a/bundles/org.jkiss.utils/src/org/jkiss/utils/BeanUtils.java
+++ b/bundles/org.jkiss.utils/src/org/jkiss/utils/BeanUtils.java
@@ -21,6 +21,7 @@ import org.jkiss.code.NotNull;
 import org.jkiss.code.Nullable;
 
 import java.lang.reflect.*;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.StringTokenizer;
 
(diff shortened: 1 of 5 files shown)


plugins/org.jkiss.dbeaver.data.gis/src/org/jkiss/dbeaver/model/gis/DBGeometry.java
```diff
@@ -68,12 +69,21 @@ public DBGeometry(Object rawValue, int srid, Map<String, Object> properties) {
...
     }
 
+    @Nullable
     public String getString() {
-        return rawValue == null ? null : CommonUtils.toString(rawValue);
+        if (rawValue == null) {
+            return null;
+        }
+        if (rawValue instanceof Geometry) {
+            // Use all possible dimensions (4 stands for XYZM) for most verbose output
```
flambchap: "the most"

bundles/org.jkiss.utils/src/org/jkiss/utils/BeanUtils.java
```diff
@@ -366,4 +369,21 @@ public static Object invokeStaticMethod(Class<?> objectType, String name, Class<
...
+     */
+    @Nullable
+    public static Class<?> findClassFromInstance(@NotNull Object instance, @NotNull String className) {
+        for (Class<?> cls = instance.getClass(); cls != null; cls = cls.getSuperclass()) {
+            if (cls.getName().equals(className)) {
+                return cls;
+            }
+        }
+        return null;
+    }
```
flambchap: I am forced to say that neither the name of this util method nor its comments are informative. It's absolutely clear from the code that it finds an ancestor with the specified `className`, but unfortunately, there is nothing about that in the name or the comments.

Although the method is quite simple, it would be nice to see a small unit test.

I also believe that `{@code}` blocks are substantially more readable that `<code></code>`.

LonwoLonwo (approved):

bundles/org.jkiss.utils/src/org/jkiss/utils/BeanUtils.java
ShadelessFox: It may not be an ancestor, e.g. supplied `instance` is already an instance of required class. Suggest better name.

serge-rider: isInstanceOf would be better

+ move class name to constant
+ include class name in the exception message

ShadelessFox: The purpose of this method is to locate concrete class named `className` from the supplied `instance`, so `isInstanceOf` does not reflect that and sounds like we want to produce a boolean result. We decided to use name `findAncestorClass` instead.

The other two notes are fair.

flambchap (approved):
👍

Issue #10543, opened 2020-11-29
Exasol: incorrect data is displayed at the beginning when creating a new column


kseniiaguzeeva:
#### System information: 
- DBeaver version 7.3.0 EA

#### Connection specification:
- Database name and version Exasol, DB2

#### Steps to reproduce, if exist:
Create a column Integer for a table

**Expected**: Column with correct type and length appears in the column list
**Actual result**: the column has length 50 at first and changes it on 36 only after saving
![AGBny8oyKf](https://user-images.githubusercontent.com/51405061/100544161-debd5300-3264-11eb-8c49-3954ed3b0a8a.gif)

LonwoLonwo:
also DB2 problem

LonwoLonwo:
https://www.ibm.com/docs/en/db2-for-zos/11?topic=columns-string-data-types DB2 types which support modifiers after data type name.

LonwoLonwo:
Fixed for Exasol and DB2. Now length/precision of column are changing in case of column type (from text to numeric, from date to binary etc.) modification.

uslss:
verified

PR #12024, opened 2021-04-07
#10543 change data type parameters in column modification case

LonwoLonwo:
2021-04-07 16_50_30-DBeaver Enterprise 21 0 2 - TEST_MODIFIER_10543


plugins/org.jkiss.dbeaver.ext.db2/src/org/jkiss/dbeaver/ext/db2/manager/DB2TableColumnManager.java
```diff
@@ -73,7 +73,7 @@
     @Override
     public boolean canEditObject(DB2TableColumn object)
     {
```
ShadelessFox, line 75: Consider moving `{` onto the previous line for all method declarations according to the code style

flambchap (approved):

PR #12026, opened 2021-04-07
#8506 Restore position and zoom after changing tile layer

ShadelessFox:


mkgrgis:
Useful change. I'm waiting this feature as PostGIS user. Thanks.

Issue #12036, opened 2021-04-08
[PostgreSQL] return main array types in data types list.


LonwoLonwo:
PostgreSQL is a special database that has a huge list of data types. 
Now we read standard data types (int4, varchar, time etc.), add aliases (like serial) and user data types.
Yo can enable reading of all data types in connection settings.

Now we need to return to the main list of types - basic array data types (_int4, _varchar, _time etc.).

LonwoLonwo:
Implemented. Also array types is now in the end of possible data types list for the column.

kseniiaguzeeva:
verified

PR #12039, opened 2021-04-08
#12036 return main array types in standard data types list

LonwoLonwo:
2021-04-08 11_08_54-DBeaver Enterprise 21 0 3 - auto_increment

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreAttribute.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreAttribute.java
@@ -37,8 +37,10 @@ import org.jkiss.dbeaver.model.runtime.VoidProgressMonitor;
 import org.jkiss.dbeaver.model.struct.DBSEntity;
 import org.jkiss.dbeaver.model.struct.DBSTypedObjectEx;
 import org.jkiss.dbeaver.model.struct.DBSTypedObjectExt4;
+import org.jkiss.utils.ArrayUtils;
 import org.jkiss.utils.CommonUtils;
 
+import java.sql.Types;
 import java.util.*;
 
 /**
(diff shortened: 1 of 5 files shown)


ShadelessFox (approved):

flambchap (approved):

flambchap (approved):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreAttribute.java
```diff
@@ -434,7 +436,9 @@ public boolean allowCustomValue()
                 log.debug("Can't get data types from database schemas", e);
                 types.addAll(column.getDatabase().getLocalDataTypes());
             }
-            return types.toArray(new PostgreDataType[0]);
+            PostgreDataType[] notArrayTypes = types.stream().filter(type -> type.getTypeID() != Types.ARRAY).sorted(Comparator.comparing(JDBCDataType::getTypeName)).toArray(PostgreDataType[]::new);
+            PostgreDataType[] arrayTypes = types.stream().filter(type -> type.getTypeID() == Types.ARRAY).sorted(Comparator.comparing(JDBCDataType::getTypeName)).toArray(PostgreDataType[]::new);
+            return ArrayUtils.concatArrays(notArrayTypes, arrayTypes);
```
ShadelessFox: Instead of this madness, you can use this instead:

```java
return types.stream()
    .sorted(Comparator
        .comparing((DBSTypedObject type) -> type.getTypeID() != Types.ARRAY)
        .thenComparing(DBSTypedObject::getTypeName))
    .toArray(PostgreDataType[]::new);
```

flambchap: Since `types` is a List, there is no need to create a stream, you can sort the list itself.

flambchap (changes requested):

ShadelessFox (approved):

Issue #11095, opened 2021-01-22
Option to edit COPY AS parameters


JamesInform:
DBeaver 7.3.3, Office Support

I have a delete sql statement with a returning clause that runs inside a transaction that I have startet clicking on the 'T' menu button.

After execution I click the Excel export button at the bottom of the data to get the returned rows as Excel. Now the query seems to be re-executed inside its own new transaction and persists the deletes so that they cannot be rollbacked.

I think the Excel export must run in the same transaction as the sql editor. If this is not possible the user should definitely get a warning when trying to do Excel export while currently being in a transaction.

I would judge this severe, if you work on data that you don't want to loose.

kseniiaguzeeva:
It happens, because **copy** option (which includes Excel export button) has default settings now and **open new connection** is used by default. We should add possibility to customize **copy** as we do for data transfer.

As a workaround while you can use export in XLSX https://github.com/dbeaver/dbeaver/wiki/Extension-Office with disabled **open new connection** 
![dbeaver_aOP09FHnd9](https://user-images.githubusercontent.com/51405061/106999795-dc610300-6797-11eb-80ab-c477c72457d6.png)

JamesInform:
This doesn't work on PostgreSQL. 

Please repeat the following steps to reproduce the issue on a PostgreSQL database:

```sql
drop table if exists my_test_table;
create table my_test_table as select * from generate_series(1,10) as my_data;

-- You should now see that the table has 10 records
select * from my_test_table;

-- Start transaction
begin;

-- Select this statement and do an export to XLSX (with "Open new connection(s)" unticked)
delete from my_test_table returning *;

-- Rollback transaction
rollback;

-- This should show the 10 records because of the rollback.
-- But it's empty, so it seems the XLSX export hat committed changes
select * from my_test_table;
```

kseniiaguzeeva:
I could reproduce it, thank you.

flambchap:
The ability to edit 'copy as' parameters has been added.

flambchap:
Fixed the original issue + added configuration dialog for 'Copy as' commands.

kseniiaguzeeva:
Transactions are not commited when you press Excel in the bottom toolbar in Manual or Smart commit modes now.
Please note, Excel button still commits transactions in Auto-commit (you can use Production connection type by default to avoid unexpected commits).

Now **Copy as** format settings can be configured via the context menu.
[...]

PR #12047, opened 2021-04-08
#11095 add possibility to save 'copy as' configurations from the context menu in result set

flambchap:
Screenshot 2021-04-08 162114
Screenshot 2021-04-08 162136


ShadelessFox (approved):
This tiny dialog is so cute, I like it.

LonwoLonwo (approved):

ShadelessFox (approved):

PR #12073, opened 2021-04-12
#11095 do not commit transaction in data transfer if transaction was not in auto commit mode before transfer started

flambchap:


LonwoLonwo (approved):

ShadelessFox (approved):

serge-rider:
Why do we commit transaction there? 
I though we don't need to perform commit at all (only switch back to auto-commit if needed)

flambchap:
You are right, we don't need it here.

ShadelessFox (approved):

PR #12088, opened 2021-04-13
Colorize selected rows#10960

flambchap:


LonwoLonwo (approved):

ShadelessFox (approved):

LonwoLonwo:
this setting deserves true as default value))

ShadelessFox:
I agree with suggestion above. What it looks like on dark theme, by the way?

PR #12098, opened 2021-04-14
dbeaver/dbeaver-ee#923 use upper case aliases option added

LonwoLonwo:
2021-04-14 15_42_11-DBeaver Enterprise 21 0 3 - _localhost MySQL 5 5_ Script-105

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
@@ -748,7 +748,7 @@ public final class SQLUtils {
         return null;
     }
 
-    public static String generateEntityAlias(DBSEntity entity, boolean useUpperCase, DBRFinder<Boolean, String> aliasFinder) {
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -748,7 +748,7 @@ public static String getScriptDescription(@NotNull IFile sqlScript)
         return null;
     }
 
-    public static String generateEntityAlias(DBSEntity entity, DBRFinder<Boolean, String> aliasFinder) {
+    public static String generateEntityAlias(DBSEntity entity, boolean useUpperCase, DBRFinder<Boolean, String> aliasFinder) {
```
ShadelessFox: You can use [`DBPIdentifierCase`](https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBPIdentifierCase.java) that already shipped with the case transformer in it.

ShadelessFox (changes requested):

ShadelessFox (approved):

serge-rider:
I thought the idea was to reuse existing configuration from SQL formatting?
Screenshot 2021-04-14 at 21 43 23

Issue #11480, opened 2021-02-25
ctrl-f Find defaults to cell under cursor


fossildoc:
- Operating system (distribution) and version
- ubuntu 18.04
- DBeaver version
- 7.3.5
- Java version 
- 1.8.0_282

- Database name and version
- postgresql  12.6
- Driver name
- jdbc postgresql
- Do you use tunnels or proxies (SSH, SOCKS, etc)? 
no
#### Describe the problem you're observing:
ctrl+f (Find) search field defaults to the contents of the cell under the cursor. If the search fails and I switch to another table, the search field changes to whatever happens to fall under the cursor. Even if the dropdown list is used to restore the search field, or even if I retype it, the Find system continues to reference the original table. This requires that the Find window be closed and re-opened in order for it to reference the new, second, table. Instead of all this, the search field should have no default and should not change automatically for any reason, and Find should change its reference to whatever table the user switches to, without having to close and re-open it.

uslss:
thanks for the report

ShadelessFox:
Fixed / Implemented:

https://user-images.githubusercontent.com/35821147/115051541-9afe7980-9ee5-11eb-9936-1b082654074b.mp4

kseniiaguzeeva:
verified

fossildoc:
This bug has not been fixed. The system behaves exactly as it did when I made the o.p.

ShadelessFox:
Can you please provide more information and demonstrate your issue by capturing a video?

PR #12125, opened 2021-04-16
#11480 Update find/replace target every time the search is initiated

ShadelessFox:
https://user-images.githubusercontent.com/35821147/115051541-9afe7980-9ee5-11eb-9936-1b082654074b.mp4


LonwoLonwo (approved):

flambchap (approved):

flambchap (approved):

serge-rider:
It is bad to have static singleton which references spreadsheet (and thus all database model objects).
Probably it is better to determine active spreadsheet on every IFindReplaceTarget API request

PR #12133, opened 2021-04-18
Updated colortheme url

agodlydeciple:
when trying to build from source the build fails when trying to get to `https://eclipse-color-theme.github.com/update` when going to that link it throws a 404.  Changing the link to `https://eclipse-color-theme-github.io/update` fixes the build and allows it to proceed to completion.


LonwoLonwo:
https://github.com/dbeaver/dbeaver/discussions/11707

Issue #10059, opened 2020-10-13
Db2 "Generate SQL"/CALL for user defined functions generates wrong syntax


gerhard81:
Hi,

Generate SQL Call for Db2 (LUW) user defined functions generates wrong syntax.
Syntax generated for i.e. SYSPROC.ADMIN_GET_INTRA_PARALLEL
`SELECT  "SYSPROC".ADMIN_GET_MEM_USAGE(:null)
`

_org.jkiss.dbeaver.model.sql.DBSQLException: SQL Error [42601]: An unexpected token "END-OF-STATEMENT" was found following "ADMIN_GET_MEM_USAGE()".  Expected tokens may include:  "<table_expr>".. SQLCODE=-104, SQLSTATE=42601, DRIVER=4.26.14_

The correct syntax must be
`SELECT * FROM TABLE ( "SYSPROC".ADMIN_GET_MEM_USAGE(:null))`

Regards,
Gerhard

uslss:
Thanks for the bug report

LonwoLonwo:
@gerhard81 

`SELECT * FROM TABLE ( "SYSPROC".ADMIN_GET_MEM_USAGE(:null))` does not work for me

but `SELECT SCHEMA_NAME.FUNCTION_NAME(:PARAM1) FROM SYSIBM.SYSDUMMY1` is much better. 

This call function method is mentioned in the documentation on z/os https://www.ibm.com/docs/en/db2-for-zos/11?topic=objects-creation-user-defined-functions but works for LUW to.

LonwoLonwo:
Fixed in DBeaver 21.0.4

![115261817-a0093600-a13c-11eb-9c7d-422962fc4639](https://user-images.githubusercontent.com/45152336/115666816-b508e380-a34d-11eb-9abf-bcaf39ffb902.png)

uslss:
verified

gerhard81:
@LonwoLonwo

Your implementation is correct for scalar functions.
But it is not correct when the function is a table function which returns a table as resultset.
In this case the function has to be called in the following way:

`SELECT * FROM TABLE ( "SYSPROC".ADMIN_GET_MEM_USAGE(:null))`

The documentation link you mentioned is for Db2 z/OS but when you call a table function in Db2 (LUW = Linux, Unix, and Windows) the function call is different .

[...]

PR #12160, opened 2021-04-19
#10059 change CALL function statement

LonwoLonwo:
2021-04-19 18_21_03-DBeaver Enterprise 21 0 3 - _none_ Script-108

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.db2/src/org/jkiss/dbeaver/ext/db2/DB2SQLDialect.java
+++ b/plugins/org.jkiss.dbeaver.ext.db2/src/org/jkiss/dbeaver/ext/db2/DB2SQLDialect.java
@@ -112,6 +112,7 @@ public class DB2SQLDialect extends JDBCSQLDialect {
     @Override
     protected String getProcedureCallEndClause(DBSProcedure procedure) {
         if (procedure.getProcedureType() == DBSProcedureType.FUNCTION) {
+            // Only "Select from function_name" doesn't work for user-defined functions. See #10059
             return "FROM SYSIBM.SYSDUMMY1";
         }
         return super.getProcedureCallEndClause(procedure);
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ext.db2/src/org/jkiss/dbeaver/ext/db2/DB2SQLDialect.java
```diff
@@ -106,6 +108,15 @@ protected void loadFunctions(JDBCSession session, JDBCDatabaseMetaData metaData,
         }
     }
 
+    @NotNull
+    @Override
+    protected String getProcedureCallEndClause(DBSProcedure procedure) {
+        if (procedure.getProcedureType() == DBSProcedureType.FUNCTION) {
+            return "FROM SYSIBM.SYSDUMMY1";
```
ShadelessFox, line 116: Probably you should leave a comment with a link to the source this goofy table comes from

https://www.ibm.com/docs/en/db2-for-zos/12?topic=tables-sysdummy1

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -743,6 +743,11 @@ protected String getStoredProcedureCallInitialClause(DBSProcedure proc) {
         }
     }
 
+    @NotNull
+    protected String getProcedureCallEndClause(DBSProcedure procedure) {
+        return "";
+    }
+
```
ShadelessFox, line 750: Consider moving this method to the super interface [`SQLDialect`](https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialect.java)

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
LonwoLonwo: I'm not sure about this.
getStoredProcedureCallInitialClause not in SQLDialect.

ShadelessFox: Okay, I see.

ShadelessFox (approved):

Issue #11201, opened 2021-02-03
"Can't navigate" error show up when reference panel is open after query execution 


SamFrost85:
#### System information: 
- Operating system (distribution) and version: Ubuntu 18.04.5 LTS
- DBeaver version: 7.3 (but I had this issues in past versions of DBeaver)
- Java version : 11
- Additional extensions

#### Connection specification:
- Database name and version: Oracle Database 18c 
- Driver name: oracle jdbc 12.2.0.1
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No

#### Describe the problem you're observing:
Often, when I have the Reference panel open the new references aren't loaded and I receive an error in this format:
Attribute ACC_ID, type=NUMBER, pos=2 is missing in result set

#### Steps to reproduce, if exist:
- Run a normal query (E.G. SELECT  * FROM ACCOUNT ab)
- open the reference panel and select one item from the drop down menu
- run another query this one is giving me the problem consistently:
SELECT DISTINCT (acc_id) FROM ACCOUNT_BALANCE ab
- click on any of the result rows
- A popup with the menu title "Can't navigate" will show up, only way of stopping it is to close the panels.

#### Include any warning/errors/backtraces from the logs
eclipse.buildId=unknown
java.version=11.0.9.1
java.vendor=Ubuntu
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_GB
Command-line arguments:  -os linux -ws gtk -arch x86_64

org.jkiss.dbeaver.model
Error
Wed Feb 03 10:55:10 GMT 2021
Attribute ACC_ID, type=NUMBER, pos=2 is missing in result set

uslss:
Hi @SamFrost85 ,
can you provide tables ddl?

SamFrost85:
@uslss , I created a new test schema with just two tables:

`CREATE TABLE "TEST"."ACCOUNT" 
   (	"ACC_ID" NUMBER(10,0) NOT NULL ENABLE, 
	"ACC_NAME" VARCHAR2(50) NOT NULL ENABLE, 
	"ACC_CODE" VARCHAR2(50), 
	 CONSTRAINT "ACCOUNT_PK" PRIMARY KEY ("ACC_ID")
  USING INDEX PCTFREE 10 INITRANS 2 MAXTRANS 255 COMPUTE STATISTICS 
[...]

PR #12213, opened 2021-04-22
#11201 refresh reference key list in ReferencesResultsContainer if new entities appeared

flambchap:


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/panel/references/ReferencesResultsContainer.java
```diff
@@ -221,7 +241,7 @@ private void refreshReferenceKeyList() {
             }
         }
 
-        if (!allEntities.isEmpty()) {
+        if (allEntities != null && !allEntities.isEmpty()) {
```
LonwoLonwo: may be "!CommonUtils.isEmpty(allEntities)" as usual?

LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/panel/references/ReferencesResultsContainer.java
flambchap: Totally forgot about this method, thanks.

ShadelessFox (approved):

serge-rider:
Do not use static cache of DBSObject references in UI controls.
We need better way to determine that parent entity/query has changed

Issue #11489, opened 2021-02-25
Wrong default length for numeric type


kseniiaguzeeva:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 21.0.0 EA

#### Connection specification:
- Database name and version SQL Server

#### Steps to reproduce, if exist:
1. Create a numeric column
2. Save it

**Expected**: the column is created
**Actual resul**t: the column is not created and the error appears: 
`SQL Error [2750] [S0001]: Column or parameter #7: Specified column precision 100 is greater than the maximum precision of 38.`

LonwoLonwo:
Fixed for numeric and decimal columns.

uslss:
verified

PR #12215, opened 2021-04-22
#11489 check scale and precision for numeric/decimal columns

LonwoLonwo:
2021-04-22 16_22_01-TestTableOne - Persist Changes

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDialect.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDialect.java
@@ -184,7 +184,7 @@ public class SQLServerDialect extends JDBCSQLDialect {
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDialect.java
```diff
@@ -219,6 +219,20 @@ public String getColumnTypeModifiers(DBPDataSource dataSource, @NotNull DBSTyped
             }
         } else if (ArrayUtils.contains(PLAIN_TYPE_NAMES , typeName)) {
             return null;
+        } else if (dataKind == DBPDataKind.NUMERIC &&
+                (SQLServerConstants.TYPE_NUMERIC.equals(typeName) || SQLServerConstants.TYPE_DECIMAL.equals(typeName))) {
+            // numeric and decimal - are synonyms in sql server
+            // The numeric precision has a range from 1 to 38. The default precision is 38.
+            // The scale has a range from 0 to p (precision). The scale can be specified only if the precision is specified. By default, the scale is zero
+            Integer precision = column.getPrecision();
```
ShadelessFox, line 227: Avoid additional boxing operations by using `int` instead of `Integer`

ShadelessFox (approved):

PR #12242, opened 2021-04-25
#11201 use result of toString() method to differentiate different underlying queries in parent data container in references result container

flambchap:


ShadelessFox:
Why don't use a reference to the data container instead of this bug-prone manipulations with its string representation?

flambchap:
https://github.com/dbeaver/dbeaver/pull/12213#issuecomment-825772768

ShadelessFox (approved):

Issue #11783, opened 2021-03-18
"Open Database Object" omits Foreign Tables


fjf2002:
**Is your feature request related to a problem? Please describe.**
The Dialog "Open Database Object" omits Foreign Tables - at least with PostgreSQL.

**Describe the solution you'd like**
The Dialog "Open Database Object" should also take Foreign Tables into consideration.

**Describe alternatives you've considered**
None.

**Additional context**
Tested on a PostgreSQL database.

fjf2002:
By the way, NOT listing foreign tables is inconsistent in my opinion, because in the DB Navigator, foreign tables ARE listed as tables.

Thank you for considering my issue.

LonwoLonwo:
Fixed in DBeaver 21.0.5.
Now "Open Database Object" returns Foreign Tables.

uslss:
verified

PR #12305, opened 2021-04-29
#11783 add foreign tables in object search dialog

LonwoLonwo:
2021-04-30 01_18_58-DBeaver Enterprise 21 0 4 - _none_ Script-9


mkgrgis:
Thanks. It will be very useful for full automatic mapping (`IMPORT FOREIGN SCHEMA`) from Firebird or others RDBMS with fdw for Postgres.

Issue #12260, opened 2021-04-27
With the latest version column names cannot be retrieved in Vertica 


isak-kaz:
#### System information: 
- macOS Big Sur
- DBeaver version 21.0.3.202104181339

#### Connection specification:
- Vertica 
- Driver Vertica 10.1.0

#### Describe the problem you're observing:

In object explorer when retrieving column list from a table drop down following error occurs: 

![image](https://user-images.githubusercontent.com/60605981/116255530-9bfb9a80-a772-11eb-9c82-4420bccbae3a.png)

#### Steps to reproduce:

Connect to VerticaDB and try getting columns names from a drop down list in object explorer

isak-kaz:
This affects autocomplete object name as well

LonwoLonwo:
Hello @isak-kaz 
Thanks for the bug report.
We already fixed this problem here https://github.com/dbeaver/dbeaver/issues/12127 , I hope.
Can you please test [Early Access version](https://dbeaver.io/files/ea/)?

isak-kaz:
Hi thanks, I don’t see a mac early access on the list in dmg form.

isak-kaz:
> Hello @isak-kaz
> Thanks for the bug report.
> We already fixed this problem here #12127 , I hope.
> Can you please test [Early Access version](https://dbeaver.io/files/ea/)?

I can confirm with that with Version 21.0.4.202105021514 this issue has **not** been resolved.

apkerr:
I can also confirm that the issue has not been resolved as of 21.0.4.202105021514.
I'm on Windows 10, using vertica-jdbc-9.3.1-0 driver.

SQL Error [2656] [42703]: [Vertica][VJDBC](2656) ERROR: Column com.child_object does not exist
  [Vertica][VJDBC](2656) ERROR: Column com.child_object does not exist
  [Vertica][VJDBC](2656) ERROR: Column com.child_object does not exist

Error pops up when you go to the Properties tab, and no columns are listed.

In the Data tab, the columns are present. However, it's no longer possible to add/delete rows or edit existing data.

LonwoLonwo:
Fixed. We added a child_object column check in the "comments" system table. 
Unfortunately, Vertica documentation does not contain enough information, which Vertica versions have column ("child_object") in the system table "comments".

LonwoLonwo:
Fixed in 21.0.5
Will be available [here](https://dbeaver.io/) at 2021-05-17

[...]

PR #12367, opened 2021-05-05
#12260 check on child_object column in comments table added

LonwoLonwo:


serge-rider:
Make childObjectColumnAvailable private.

Issue #11357, opened 2021-02-15
Oracle: error when editing trigger/package body/procedure/function and trying to persist it after creation


F0xD8rk:
- Operating system (distribution) and version: Windows 10 version 20H2 Build 19042.804
- DBeaver version: 7.3.5.202102141704

- Database name and version: Oracle database 12c
- Driver name: Oracle Client 12 - 12.2.0.1.0 - oracle.jdbc.OracleDriver
- Do you use tunnels or proxies (SSH, SOCKS, etc)?  No

#### Description/Steps to reproduce the problem

First, I created a trigger by executing the corresponding code in the SQL editor as a SQL statement. (with the ";" character at the end). After creation, I needed to adjust the trigger, so I open the trigger by:
- Navigating to the corresponding table of the trigger using the navigator.
- I click on the triggers tab
- I right click on the the just created trigger and choose "View trigger"
- I perform an edit on the trigger
- I press ctrl-s to save the trigger
- In the next window, I click on "Persist" after which the compilation fails with the error:
_PLS-00103: Encountered the symbol "end-of-file" when expecting one of the following:
   ;
The symbol ";" was substituted for "end-of-file" to continue.
Compile error at line 69, column 0_

After you then edit the trigger again and also place the ";" ending character manually, the persist works fine. The next time you edit the trigger using the GUI again, the problem is gone. It seems the problem only occurs the first time you edit the trigger using the GUI after creation using a sql editor pane.

Not a big issue, but it would be nice if it was not needed to place the ";" character manually at the end of the trigger after first creation. Thanks!

F0xD8rk:
In the meanwhile I also noticed this problem when creating a package.
When using the SQL editor:
- When creating a package using _CREATE OR REPLACE PACKAGE ..._, the delimiter is set correctly in the declaration when viewing the properties.
- When creating a package body using _CREATE OR REPLACE PACKAGE BODY_, the delimiter is _not_ set automatically (same problem as the described issue above) and you have to add it manually at the end before saving (and persisting) in tthe the _BODY_ tab of the properties.

F0xD8rk:
I just noticed that the same problem occurs after creation of a procedure or a function. The delimiter is also not set automatically after creation of the procedure/function using the SQL editor. 
The problem does not occur anymore after:
- manually putting the delimiter at the end of the procedure/function using the GUI
- Persisting the changes

In short, the problem occurs only after first creation of the procedure/function.

[...]

PR #12410, opened 2021-05-11
#11357 fix semicolon trim when executing selected oracle query

flambchap:


ShadelessFox (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -390,6 +388,16 @@ public static String trimQueryStatement(SQLSyntaxManager syntaxManager, String s
...
+    private static boolean containsWhitespaces(@NotNull CharSequence charSequence) {
+        for (int i = 0; i < charSequence.length(); i++) {
+            char c = charSequence.charAt(i);
+            if (Character.isWhitespace(c)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
```
ShadelessFox: IMO it could be replaced with one-liner: `afterEnd.chars().noneMatch(Character::isWhitespace)`

ShadelessFox: Also, the logic of this method does not reflect its name.

It _says_ that it checks for presence of whitespace character, but, in fact, it check if there's no whitespaces whatsoever.

flambchap: Fair point.

ShadelessFox (approved):

LonwoLonwo (approved):

Issue #10737, opened 2020-12-20
Edit host in exist user


ZhouJiaZhi:
The host field can only be modified when creating a new user. After the user is successfully created, this field cannot be modified.
I hope this field can be modified, so that I don’t need to modify this field by deleting the user then create a new user.
![image](https://user-images.githubusercontent.com/20750357/102707574-481a0a00-42d7-11eb-8c4b-109c728954fa.png)

uslss:
thanks for the suggestion

flambchap:
We've decided to add a possibility to rename the user from the navigator tree instead, which solves the original problem as well.

uslss:
verified

PR #12432, opened 2021-05-12
#10737 add renaming facility for mysql user

flambchap:
This PR adds renaming functionality for MySQL users.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/config/MySQLCommandChangeUser.java
+++ b/plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/config/MySQLCommandChangeUser.java
@@ -17,6 +17,8 @@
  */
 package org.jkiss.dbeaver.ext.mysql.ui.config;
 
+import org.jkiss.code.NotNull;
+import org.jkiss.code.Nullable;
 import org.jkiss.dbeaver.DBException;
 import org.jkiss.dbeaver.ext.mysql.MySQLUtils;
 import org.jkiss.dbeaver.ext.mysql.model.MySQLDataSource;
@@ -31,9 +33,8 @@ import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
 import org.jkiss.dbeaver.model.sql.SQLUtils;
 import org.jkiss.utils.CommonUtils;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
+import java.util.function.Supplier;
 
 /**
  * Grant/Revoke privilege command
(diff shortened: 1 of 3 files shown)


LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLUser.java
```diff
@@ -303,4 +309,30 @@ public DBSObject refreshObject(@NotNull DBRProgressMonitor monitor) throws DBExc
...
+    public void setPersistedHost(@Nullable String persistedHost) {
+        this.persistedHost = persistedHost;
+    }
+
+    public String getPersistedFullName() {
+        return getFullName(persistedUserName, persistedHost);
+    }
+
+    private static String getFullName(@Nullable String userName, @Nullable String host) {
+        return "'" + userName + "'@'" + host + "'";
```
ShadelessFox, line 336: Does this makes sense to allow nullable values here? In the worst case, it will result in `'null'@'null'`

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLUser.java
```diff
@@ -72,7 +76,9 @@ public MySQLUser(MySQLDataSource dataSource, ResultSet resultSet) {
         if (resultSet != null) {
             this.persisted = true;
             this.userName = JDBCUtils.safeGetString(resultSet, "user");
+            persistedUserName = userName;
```
ShadelessFox, line 79: Add `this.` for consistency

plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/config/MySQLCommandChangeUser.java
```diff
@@ -86,27 +101,36 @@ public void afterExecute(DBCSession session, Throwable error)
...
+        if (sql != null) {
+            actions.add(new SQLDatabasePersistAction(MySQLUIMessages.edit_command_change_user_action_update_user_record, sql));
+        }
+    }
+
+    @Nullable
+    private String generateRenameScript() {
+        MySQLUser mySQLUser = getObject();
+        if (mySQLUser.isPersisted() && (!Objects.equals(mySQLUser.getUserName(), mySQLUser.getPersistedUserName()) || !Objects.equals(mySQLUser.getHost(), mySQLUser.getPersistedHost()))) {
+            return "RENAME USER " + mySQLUser.getPersistedFullName() + " TO " + getObject().getFullName() + ";"; //$NON-NLS-1$ //$NON-NLS-3$
```
ShadelessFox: Missing NLS marker for the second string

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLUser.java
flambchap: It totally makes sense since this is exactly what's going to happen when we create a new user. The fields `persistedUserName` and `persistedHost` are null in this case. Moreover, IMHO `@Nullable`'s meaning is to indicate that it is _safe_ to pass null as an argument.

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLUser.java
flambchap: To be quite honest, it's tempting to delete `this` qualifier from other statements rather than add it to new ones.

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLUser.java
ShadelessFox: OK.

plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/config/MySQLCommandChangeUser.java
flambchap: Woops.

Issue #12396, opened 2021-05-10
DBeaver 21.0.4: Not all columns showing in DB Navigation Pane/Generated SQL - View


Cwest99:
#### System information: 
- Operating system: Windows 10 v1909
- DBeaver version: 21.0.4

#### Connection specification:
- Database name and version: AWS Redshift
- Driver name: com.amazon.redshift.jdbc42.Driver, using local PostgreSQL Binaries
- Using SSL

#### Describe the problem you're observing:
None of the date/time fields are showing in the Navigation Pane for Views. Also, when generating SELECT, none of the date/time fields in the DDL are generated.

Example: DDL for View

CREATE OR REPLACE VIEW ei_ops.atom_prod_tasks_data AS
 SELECT atom_prod_tasks_data.id_nbr
, atom_prod_tasks_data.lni_text
, atom_prod_tasks_data.status_text
, atom_prod_tasks_data.has_hist_ind
, atom_prod_tasks_data.create_dts
, atom_prod_tasks_data.updt_dts
, atom_prod_tasks_data.msg_dts
   FROM ops.atom_prod_tasks_data with no schema binding;

The three *_dts fields are not showing in the navigation pane, or in the generated SELECT. (see attached images)

Is there a fix for this bug?
![DB_Generated_View_SQL](https://user-images.githubusercontent.com/83968374/117682079-f253da80-b16f-11eb-9d55-84a2269cc309.jpg)
![DB_Nav_Pane](https://user-images.githubusercontent.com/83968374/117682092-f7188e80-b16f-11eb-95aa-d28688ac4e29.jpg)

LonwoLonwo:
Hello @Cwest99 
Can you provide DDL of ops.atom_prod_tasks_data table?

LonwoLonwo:
connected with https://github.com/dbeaver/dbeaver/issues/12372 and https://github.com/dbeaver/dbeaver/issues/12204

Cwest99:
Sure. The DDL for the tableis:

CREATE TABLE IF NOT EXISTS ops.atom_prod_auto_tasks_data
(
	id_nbr INTEGER   ENCODE RAW
	,lni_text VARCHAR(270)   ENCODE zstd
	,status_text VARCHAR(270)   ENCODE zstd
	,has_hist_ind INTEGER   ENCODE az64
	,ver_nbr INTEGER   ENCODE az64
	,type_name VARCHAR(120)   ENCODE zstd
[...]

Issue #12372, opened 2021-05-06
View columns not visible with 'NO SCHEMA BINDING' views on Redshift connection


mkosater:
#### System information: 
- Windows
- DBeaver version: 21.0.3.202104181339

#### Connection specification:
- Amazon Redshift 
- com.amazon.redshift.jdbc42.Driver 

#### Describe the problem you're observing:
When dbeaver displays columns for redshift views that have been defined with 'NO SCHEMA BINDING', it does not display the columns in all cases.  

#### Steps to reproduce, if exist:
Define a redshift view using  'NO SCHEMA BINDING' columns will not show. 

create or replace view public.tst_no_binding_uc
as 
select 'testc1' c1, 999 c2
WITH NO SCHEMA BINDING
;

Define a view using  'no schema binding' and the columns will show.

create or replace view public.tst_no_binding_lc
as 
select 'testc1' c1, 999 c2
with no schema binding
;

appears that the check done for this condition is using case sensitive compare.

#### Include any warning/errors/backtraces from the logs
no error are thrown.

LonwoLonwo:
Thanks for the bug report

uslss:
verified

PR #12437, opened 2021-05-12
Redshift missing columns when view is configured as a non-binding.

mkosater:
This change address two issues: 

- #12372 View columns not visible with 'NO SCHEMA BINDING' views on Redshift connection
- #12396 DBeaver 21.0.4: Not all columns showing in DB Navigation Pane/Generated SQL - View

It address a few aspects, first it will allow the column to display in the navigator and also allow the columns to show in 'generate sql' functions.


LonwoLonwo (approved):

serge-rider:
Thank you for the contributions 👍

Issue #11289, opened 2021-02-11
Smart commit starts a transaction when executing explain


ionutursuleanu:
#### System information: 
- macOS Catalina 10.15.7
- dbeaver 7.3.4

#### Connection specification:
- PostgreSQL 11.9

#### Describe the problem you're observing:
Smart commit starts a transaction when executing explain
#### Steps to reproduce, if exist:
Run ```explain select 1;```

uslss:
thanks for the report

flambchap:
Fixed.

uslss:
verified

PR #12447, opened 2021-05-13
#11289 do not trigger transaction start in smart commit mode on explain statement

flambchap:


LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -48,6 +48,8 @@
     private static final String[] QUERY_KEYWORDS = new String[] { SQLConstants.KEYWORD_SELECT };
     private static final String[] EXEC_KEYWORDS = new String[0];
     private static final String[] DDL_KEYWORDS = new String[0];
+    private static final Collection<String> TRANSACTION_NON_MODIFYING_KEYWORDS =
```
ShadelessFox, line 51: Can we have a plain array here? There's not much entries to search from, so performance is not so crucial here, I believe. Also, it's an additional overhead.

bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
```diff
@@ -875,4 +875,12 @@ public static boolean isDigit(int ch, int radix) {
...
         return false;
     }
+
+    @NotNull
+    @SafeVarargs
+    public static <T> Set<T> unmodifiableSet(@NotNull T... vararg) {
+        Set<T> set = new HashSet<>(vararg.length * 4 / 3 + 1); //Adjusting to account for the default load factor
+        set.addAll(Arrays.asList(vararg));
+        return Collections.unmodifiableSet(set);
+    }
```
ShadelessFox: That just looks like an unreasonable complication

Issue #10731, opened 2020-12-19
The content of HEX Editor settings partly disappeared in Preferences on Mac


kseniiaguzeeva:
#### System information: 
- Operating system (distribution) and version Mac Catalina
- DBeaver version 7.3.1 EA

#### Steps to reproduce, if exist:
Open Window>Preferences>Editors>Data Editor>Binary Editors>HEX Editor

Expected: all tab fields have a data
Actual result: the data is not visible in some fields:
![Screenshot 2020-12-19 at 12 33 01](https://user-images.githubusercontent.com/51405061/102686154-c652af80-41f6-11eb-853e-e51ef655cedd.png)

flambchap:
Fixed.

kseniiaguzeeva:
verified

PR #12516, opened 2021-05-18
#10731 check if font monospaced by measuring chars width in range [0x21; 0x7F)

flambchap:


plugins/org.jkiss.dbeaver.ui.editors.hex/src/org/jkiss/dbeaver/ui/editors/binary/pref/HexPreferencesManager.java
```diff
@@ -356,6 +346,24 @@ private void populateFixedCharWidthFontsAsync()
...
     }
 
+    private boolean isMonospacedFont(@NotNull FontData fontData, boolean isScalable) {
+        int height = 10;
+        if (!isScalable) {
+            height = fontData.getHeight();
+        }
+        Font font = new Font(Display.getCurrent(), fontData.getName(), height, SWT.NORMAL);
+        fontsGc.setFont(font);
+        char ch = ' ';
```
ShadelessFox: Can you please extract these obscure characters (`' '` and `'\u007F'`) to a class constants?

plugins/org.jkiss.dbeaver.ui.editors.hex/src/org/jkiss/dbeaver/ui/editors/binary/pref/HexPreferencesManager.java
```diff
@@ -317,18 +318,7 @@ private void populateFixedCharWidthFontsAsync()
         FontData fontData = getNextFontData();
         if (!fontsRejected.contains(fontData.getName())) {
```
ShadelessFox, line 320: What about filtering-out Windows-specific fonts that start with `@`? They're vertically aligned, so they can't be used as a monospace font in this situation.

https://devblogs.microsoft.com/oldnewthing/20120719-00/?p=7093

ShadelessFox (changes requested):

ShadelessFox (approved):

LonwoLonwo (approved):

ShadelessFox (approved):
That's way better :^)

Issue #12246, opened 2021-04-26
New tests for Oracle DDL


LonwoLonwo:
We need more tests for Oracle.

For Alter column statements first at all.
Check data types parameters and correct syntax.

PR #12712, opened 2021-06-02
#12246 add Oracle table columns tests

LonwoLonwo:
add Oracle table columns tests

### DIFF
--- a/test/org.jkiss.dbeaver.ext.oracle.test/src/org/jkiss/dbeaver/ext/oracle/model/OracleAlterTableColumnTest.java
+++ b/test/org.jkiss.dbeaver.ext.oracle.test/src/org/jkiss/dbeaver/ext/oracle/model/OracleAlterTableColumnTest.java
@@ -1,0 +1,142 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2021 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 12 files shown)


test/org.jkiss.dbeaver.ext.oracle.test/src/org/jkiss/dbeaver/ext/oracle/model/OracleAlterTableColumnTest.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.Collections;
+import java.util.List;
+
+@RunWith(MockitoJUnitRunner.class)
+public class OracleAlterTableColumnTest {
+
+    private static final Log log = Log.getLog(OracleAlterTableColumnTest.class);
+
+    @Mock
+    private DBRProgressMonitor monitor;
```
ShadelessFox, line 47: You don't need to mock the monitor here, you can use `VoidProgressMonitor`

ShadelessFox:
According to the log, you forgot to commit `org.jkiss.dbeaver.model.impl.edit.TestCommandContext.java`.

ShadelessFox (changes requested):
I don't think the `TestCommandContext.java` class should be exposed in the model.
Consider moving it to the tests or better make it inner class of `OracleAlterTableColumnTest.java`.

ShadelessFox (approved):

ShadelessFox (approved):

Issue #4979, opened 2019-01-11
Decompression of blobs


KOLANICH:
Sometimes it is needed to store compressed data in the DB. Unfortunately not all the DBs have built-in compression and FUSE compressed FSes are not available for every OS. So it may make sense to store compressed binary blobs in the DB.

Unfortunately when one sees them in DBeaver he sees them compressed, but often they are needed uncompressed. So it'd be nice to have a feature to decompress the contents of certain columns.

This may be done by injecting an external function into SQL doing decompression.

serge-rider:
Sound too specific. There are dozens of different compression algorithms moreover it is not clear how particular data types should be compressed (e.g. numbers, timestamps) - in some binary format, plain text or somehow else.

KOLANICH:
>There are dozens of different compression algorithms moreover

The most popular are: deflate, lzma, lzma2, brotli, zstd. Data stream type can be either determined by header, or can be provided (some compression libs allow to put raw data).

>moreover it is not clear how particular data types should be compressed

Compression algorythms usually work on binary data. So the transformation proposed is `blob (compressed) <-> blob (decompressed)`. For other types they should be (de)serialized from/to binary form first.

serge-rider:
I think this can be easily (relatively) implemented as a custom value transformer (`View As` context menu).

flostadler:
I'd like to tackle this issue if it is still relevant.

sachinmukherjee:
Is this issue still open?

KOLANICH:
As you see, it is still open. And I still use my python lihrary for compressing databases (now with an additional LMDB backend).

alex-rus-msk:
I cannot decompressing navision 2009R2, somebody can help?

BLOB fields starts with magic number 0x01447C5A
if throw a way 0x01447C5A results still not decompressed or error because unsupported compression method

in nav2009R2 my fields in HEX:
01447C5A000A386D40C0B870B879A973C406F7B061B0

it is simple txt file - with text:
testtest20212021

compressed string look like:
�D|Z
8m@A?p?y©sA�?°a°

i try to decompress string with code (can see below) but unsuccefuly because error... (System.IO.InvalidDataException: "The archive entry was compressed using an unsupported compression method.")

can somebody help to decompress NAV2009R2 blob?

mycode:
Private Sub Decompress(ByVal fi As FileInfo)
    Using inFile As FileStream = fi.OpenRead()
        Dim curFile As String = fi.FullName
[...]

PR #12720, opened 2021-06-03
Implement prototype auto-decompress AttributeTransfomer, #4979

lambdaupb:
#4979

### DIFF
--- a/features/org.jkiss.dbeaver.appstore.feature/feature.xml
+++ b/features/org.jkiss.dbeaver.appstore.feature/feature.xml
@@ -26,6 +26,10 @@
         <import plugin="org.apache.commons.logging" match="greaterOrEqual"/>
         <import plugin="com.jcraft.jsch" match="greaterOrEqual"/>
         <import plugin="org.objectweb.asm" match="greaterOrEqual"/>
+        <import plugin="org.apache.commons.compress" match="greaterOrEqual"/>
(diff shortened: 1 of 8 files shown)


serge-rider:
I think it should be implemented in a separate plugin. 
We don't want to add Bouncycastle as a mandatory DBeaver CE dependency.

You also don't need to modify appstore.feature configuration. It is specifically designed for Windows Store.
What is the reason to add tukaani.xz as a dependency?

lambdaupb:
Bouncycastle? I assume you mean commons-compress.
tukaani.xz is an optional dependency of commons-compress for XZ support.
Sadly brotli is not available as an eclipse osgi package, or it would be added as well.

Yes I agree, this is only a PoC and I created the pull request to provide the skeleton to base something on, mostly showing that this feature is reasonably easy to implement.

In the Issue I stated that the extension APIs this uses are not the best place to do this, as it also determines how the contents of the compressed blob are interpreted. (afaict)

I have no clue about the codebase overall, and I didn't  see an extension point for AttributeTransformers for a plugin, so I went for the most direct place to put it.

serge-rider:
Please ignore bouncycastle comment, it is an optional dependency already.
Tukaani is already included so it is also not a problem.

I lake general idea. However getDisplayString override isn't the primary goal of such a transformer. It has to provide access to uncompressed binary data in a stream form (it could be a picture, or xml or whatever). To view it in the value panel.
Function getValueFromObject in the value handler usually handles this.

Issue #11760, opened 2021-03-17
Show info message when database node is not deleted because of open editor


flambchap:
Currently, when we have a navigator node opened in some editor and we attempt to delete it via the navigator tree, we have to confirm deletion using the `save` button in the editor. This is somewhat confusing, so we should display a message box that informs the user about that.

akilovich:
We already show the type of objects. Well, that's convinient.

Matvey16:
Verified

PR #12724, opened 2021-06-03
#11760 show message box when object has been deleted only from the navigator

flambchap:


serge-rider:
- Use `instanceof NavigatorViewBase` instead of exact class check.
- Activate editor after messagebox close

PR #12726, opened 2021-06-03
Workspaces dequadratized#12150

flambchap:


plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -310,50 +315,89 @@ private void markLocationReadOnly(Location instanceLoc) {
...
+        ChooseWorkspaceData launchData = new ChooseWorkspaceData(instanceLoc.getDefault());
+        String[] arrayOfRecentWorkspaces = launchData.getRecentWorkspaces();
+        Collection<String> recentWorkspaces;
+        if (arrayOfRecentWorkspaces == null) {
+            recentWorkspaces = new ArrayList<>();
+        } else {
+            recentWorkspaces = Arrays.asList(arrayOfRecentWorkspaces);
+        }
+        int maxSize = recentWorkspaces.size();
+        recentWorkspaces.removeIf(Objects::isNull);
```
ShadelessFox, line 357: This call will fail because `Arrays.asList` returns an unmodifiable list.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
flambchap: Well, it's not modifiable, but `removeIf` will indeed fail. Thank you for noticing.

Also, the `removeIf` method of this particular `List` implementation would have worked with O(n^2), which is not what I desire.

ShadelessFox (approved):

Issue #12746, opened 2021-06-05
Map: button to copy the position of the center of the viewport


chapmanjacobd:
Sometimes I need to update point geometry in the database. The Map viewer is very helpful but it doesn't allow me to copy the correct coordinate so that I can update it. I have to open a browser and go to an online map, pan to the correct location and copy the latlng.

**Describe the solution you'd like**
I would like to have a button which inserts the centroid of the map viewport (in the current coordinate system) into the clipboard.

**Describe alternatives you've considered**
Maybe that is a little hacky but I can't think of a better way other than to make the geometry movable (direct manipulation) to update the geometry within the results panel (but that seems like a lot more work without much additional benefit).

**Additional context**

ShadelessFox:
Hello @chapmanjacobd,

Thanks for the suggestion!

ShadelessFox:
Implemented basic functionality:

![xgXwX2R1fI](https://user-images.githubusercontent.com/35821147/121094062-adff3d00-c7f6-11eb-803c-9c0c45de1a3e.png)

It copies the coordinates formatted as `latitude, longitude` to the clipboard.

Please note that it intentionally works only when viewed inside DBeaver.

Matvey16:
Wrong coordinates are copied, may always be from EPSG:3857

Matvey16:
Verified.
Please note that the coordinates are copied according to EPSG:4326 CRS and are just raw numbers. You may need to remove comma and switch latitude and longitude places to correctly insert it into a database.

Issue #12244, opened 2021-04-26
Refactor web map to flatgeobuf


mkgrgis:
[xf:gis]
**Is your feature request related to a problem? Please describe.**
The _fastest_ `leaflet.js` view and the _smallest_ web applications and _universal_ export of GIS data based on GeoJSON format. Now DBeaver's web map freeze if there is more than ~2500 points/other objects. `Leaflet.js` can support not less than ~10000..20000 points depends on CPU/RAM.
Now in DBeaver's web app there is `sourceValues` (wkx format) and `sourceTips`(array). In complex there is full data model for the geotable. This two variables is _calculated base_ for `leaflet.js` through `wkx.js`, **not geodata "as is"**, which is geoJSON for `leaflet.js`.

I think, DBaver's team can refactor web application (especially for PostGIS, I don't know how there is in OracleSpatial and other ) from wkx directly to geoJSON. Web app will be simpler and faster and there will be no problems with other useful function "export geodata in GeoJSON format".

How to compose both data and geometry to geoJSON?

Example
```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {
        "start_date": 1792,
        "reconstruction_date": [1850, 1947, 2013]
      },
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              30.452706813812256,
              59.68983687874888
            ],
            [
              30.45325398445129,
              59.68956073111389
            ],
            [
              30.453307628631595,
              59.68958238983421
            ],
            [
              30.45274972915649,
              59.68985312265652
            ],
            [
              30.452706813812256,
              59.68983687874888
            ]
          ]
        ]
      }
    },
    {
      "type": "Feature",
      "properties": {
        "name": "Затвор из дерева",
        "год реконструкции": 2013
      },
      "geometry": {
        "type": "Point",
        "coordinates": [
          30.45335054397583,
          59.68954448706445
        ]
      }
    }
  ]
}
```
for this objects:
![изображение](https://user-images.githubusercontent.com/41448637/116035344-818bc900-a66d-11eb-9a64-18b13bee09f0.png)

**Describe the solution you'd like**
Web application files structure
```
├── inc
│   ├── layers.png
│   ├── leaflet.css
│   ├── leaflet.js
│   └── webApp.html
└── *.geoJSON
```

**Describe alternatives you've considered**
WebApp based on two files: static **webApp.html** and dynamic ***.geoJSON** for separating geodata.

[...]

PR #12767, opened 2021-06-07
#12746 Allow copying coordinates under cursor

ShadelessFox:
xgXwX2R1fI


mkgrgis:
There is a similar instrument on http://retromap.ru/0519961_z16_55.751933,37.671840 with some useful functions.

ShadelessFox:
Hello @mkgrgis,

We definitely don't want to draw a cross over the entire screen, but I like the idea of displaying address under the cursor.
It can be achieved with https://wiki.openstreetmap.org/wiki/Nominatim.

Or perhaps you wanted to point at some specific thing?

mkgrgis:
Thanks for answer, @ShadelessFox ! I think copying of coordinates isn't about drawing, of course. Nominatim is really useful for special string near map, activated by a button or for leaflet `Popup`. It's hard to compose good ergonomics for me.
On [osm.org](osm.org) there is complex simple location service, not only Nominatim, see net queries by button 
изображение
and result of this queries.

ShadelessFox:
As long as there's an open-source API for gathering such information we can implement that. The real question is do we really need such features? DBeaver is not a geologically-specialized tool, I'm not even sure if I spelled this word correctly :)
Of course small features like copying coordinates would be cool, but something else is an overkill IMO.

mkgrgis:
I think at this moment it's important to implement simple and fast web application [with full geoJSON data model](https://github.com/dbeaver/dbeaver/issues/12244) Then it will be not a problem to add a functions like we discussed. It will be easy for me to propose implementation of some features via `leaflet.js` incl. Nominatim.

Issue #10734, opened 2020-12-19
JSON, HTML export from query add extra signs near <,>,=


kseniiaguzeeva:
#### System information: 
- Operating system (distribution) and version Mac
- DBeaver version 7.3.1 EA

**Steps:**
1. Execute a select with a condition (e.g. for PostgreSQL.dvdrental.publc: select language_id from language where language_id<4;)
2. Export it as JSON or HTML

**Expected**: the query is displayed in the exported file like in SQL Editor
**Actual result**: extra signs appear in the file
![Screenshot 2020-12-16 at 17 17 04](https://user-images.githubusercontent.com/51405061/102688691-69142980-4209-11eb-95c1-40c4e6b1bc09.png)

P.S. It is actual for CloudBeaver too.

kseniiaguzeeva:
verified.

PR #12797, opened 2021-06-09
Fixed issue#10734 HTML file showing HTML entities

shawnsp:
Hi DBeaver Community,

This is a fix for issue #10734. Exported HTML file is now able to display <, &, and >. I could not reproduce the same issue on the JSON file. JSON file actually works fine on my end. 

after

### DIFF
--- a/plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterHTML.java
+++ b/plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterHTML.java
@@ -167,6 +167,7 @@ public class DataExporterHTML extends StreamExporterAbstract {
         if (value == null) {
             out.write("&nbsp;");
         } else {
+        	value = value.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;");
             out.write(value);
         }
         out.write("</th>");
@@ -178,6 +179,7 @@ public class DataExporterHTML extends StreamExporterAbstract {
         if (value == null) {
             out.write("&nbsp;");
         } else {
+        	value = value.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;");
             out.write(value);
         }
         out.write(header ? "</th>" : "</td>");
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterHTML.java
```diff
@@ -167,7 +167,6 @@ private void writeTableTitle(String value, int columns) {
         if (value == null) {
             out.write("&nbsp;");
         } else {
-            value = value.replace("<", "&lt;").replace(">", "&gt;").replace("&", "&amp;");
```
ShadelessFox, line 170: Non-escaped `<` and `>` may break HTML and allow potential injection.

The problem here is that these special characters are escaped first and the produced `&` characters are replaced by `replace("&", "&amp;")` too, resulting in broken escape sequence: `<&` &rArr; `&lt;&` &rArr; `&amp;lt;&amp;`.

So the fix is to move `replace("&", "&amp;")` to be the beginning, just before other ones.

ShadelessFox (changes requested):
Thanks for the contribution, anyway!

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterHTML.java
shawnsp: Thank you for the detailed explanation.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterHTML.java
```diff
@@ -167,7 +167,7 @@ private void writeTableTitle(String value, int columns) {
...
+        	value = value.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;");
```
ShadelessFox: One last moment: please use 4 spaces instead of tabs. Let's not mix them 🙂

ShadelessFox (changes requested):

ShadelessFox (approved):

Issue #12676, opened 2021-05-31
Different column is exported to CSV


KES777:
How to reproduce:
when result have columns with same name and I select second column, but first column is exported into CSV

uslss:
cannot reproduce an issue.
what database and dbeaver version do you use?
could you provide a video with the issue?

KES777:
`select 1 as first, 2 as first, 3 as first`

After run your query by rightclicking column select first and third column
export to CSV

at csv you will get `1` and `2` value, instead of expected `1` and `3`

Version is 21.0.3, PostgresDB

KES777:
Here I select only two, but three columns =)

This is a bit different issue, but root of the problem is same: columns with similar names
![image](https://user-images.githubusercontent.com/799021/120331682-b4714e80-c2f6-11eb-9094-f464f7240cb1.png)

uslss:
thanks for the report and clarification

LonwoLonwo:
Export with enabled setting "Selected columns only" and columns with equals names fixed in DBeaver 21.2.0 .

Matvey16:
Verified
Need to fix another issue with same-named columns #13622

PR #12800, opened 2021-06-10
Fixed Issue#12676 Different column is exported to CSV when columns have the same name

pkakria:
Hi,
I found out that the root cause of the issue #12676 was that in ResultSetDataContainer.java, the filtering of the ResultSet based on the selected columns was being done only based on column name. This caused any column that had the same name as a selected column to be put inside the filtered array. 
I put in an additional condition to compare the ordinal positions of a candidate attribute against those of the selected columns. Now only the columns in the selected positions are exported.
To make this change, I added a new private member List to the ResultSetDataContainerOptions.java and added get and set methods for the same. I also updated the code wherever a ResultSetDataContainerOptions object was being created to set the new member List so that no existing code calls filterAttributeBindings() without setting the list of selected column ordinal positions and raises an exception. 
Please review if the changes are appropriate. I have attached some screenshots validating the fix here.

---------------
_ResultSet= [1,2,3]_. Select only the first column containing 1.

---------------

---------------

New expected behaviour

Earlier behaviour with the same input
previous
---------------
This also corrects the behaviour of OpenWith option.
browserSelection

---------------
ViewOnBrowser

### DIFF
--- a/plugins/org.jkiss.dbeaver.data.office.ui/src/org/jkiss/dbeaver/data/office/ui/handlers/OpenSpreadsheetHandler.java
+++ b/plugins/org.jkiss.dbeaver.data.office.ui/src/org/jkiss/dbeaver/data/office/ui/handlers/OpenSpreadsheetHandler.java
@@ -69,12 +69,16 @@ public class OpenSpreadsheetHandler extends AbstractHandler
                 selectedRows.add((long) selectedRow.getRowNumber());
             }
             List<String> selectedAttributes = new ArrayList<>();
+            // this list stores the ordinal positions of the selected columns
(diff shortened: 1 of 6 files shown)


pkakria:
Hi Dbeaver team, May I follow-up on of this PR ?
Thanks.

serge-rider:
Thanks for the PR 👍 

However this issue may be resolved in #13299 
We'll test it during version 21.1.4 and review this PR again.

LonwoLonwo:
Thanks for your PR again!

The problem was solved a little bit in another way.
https://github.com/dbeaver/dbeaver/issues/12676

Issue #12660, opened 2021-05-31
Remove single quotes for filtered values in context menu


kseniiaguzeeva:
#### System information: 
- DBeaver version 21.1.0

#### Describe the problem you're observing:
When we select a value to filter via context menu, it is surrounded by single quotes. We have to remove them.

Skydiveroid:
We can't do that.

PR #12801, opened 2021-06-10
Fix for issue #12660 remove single quotes

smk-dev20:
Hello DBeaver Community,

Submitting a fix for issue https://github.com/dbeaver/dbeaver/issues/12660
In ResultSetViewer.java I only commented out the relevant line because I wasn't sure what the motivations were behind the if condition. 
Please let me know if I need to change it.

Thanks


serge-rider:
I don't think that we should remove quotes always. We need them for string literals but not for numbers..

Issue #12314, opened 2021-04-30
[MSSQL] Show more trigger fields


kseniiaguzeeva:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver 21.0.4

#### Connection specification:
- Database name and version SQL Server

#### Describe the problem you're observing:
We should show more trigger fields in Metadata Editor.

LonwoLonwo:
Implemented in DBeaver 21.1.1

![2021-06-16 13_14_48-Window](https://user-images.githubusercontent.com/45152336/122246782-86336780-cecf-11eb-9507-e82f08be56f9.png)

uslss:
verified

PR #12848, opened 2021-06-15
#12314 add more parameter fields for triggers

LonwoLonwo:
2021-06-16 13_14_48-Window


plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerTriggerBase.java
```diff
@@ -49,8 +50,17 @@
...
+    private String triggerTypeDescription;
+    private String eventGroup;
+    private Date createDate;
+    private Date modifyDate;
     private boolean insteadOfTrigger;
     private volatile int disabled;
+    private boolean isMsShipped;
+    private boolean isNotForReplication;
+    private boolean isFirst;
+    private boolean isLast;
```
ShadelessFox: Since these fields are assigned in constructor and they can't be changed later, mark them `final`.

ShadelessFox (changes requested):

ShadelessFox (approved):

ShadelessFox (approved):

Issue #11013, opened 2021-01-16
Table triggers folder is opening for 2 minutes or more


kseniiaguzeeva:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 7.3.3 EA
- Additional extensions

#### Connection specification:
- Database name and version Informix
#### Describe the problem you're observing:
I have the only trigger in Table Triggers folder and it takes more then 2 minutes to open the folder

LonwoLonwo:
Fixed for: HSQL, Informix, Teradata, Firebird, Sybase, SQLite, Old PostgreSQL
More parameters for Informix Trigger and HANA Trigger added
HSQL view and trigger definitions improved
Implemented in DBeaver 21.1.2

uslss:
verified

PR #12885, opened 2021-06-18
#11013 add support of trigger cache for generic triggers

LonwoLonwo:
more parameters for InformixTrigger and HANATrigger
HSQL view and trigger definitions improved

2021-06-18 02_49_51-DBeaver Enterprise 21 1 1 - _SYS_TRIGGER_REF_INS_163735_#0_#

2021-06-18 02_50_30-DBeaver Enterprise 21 1 1 - upqty_i

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.firebird/src/org/jkiss/dbeaver/ext/firebird/model/FireBirdMetaModel.java
+++ b/plugins/org.jkiss.dbeaver.ext.firebird/src/org/jkiss/dbeaver/ext/firebird/model/FireBirdMetaModel.java
@@ -151,10 +151,7 @@ public class FireBirdMetaModel extends GenericMetaModel
         int type = JDBCUtils.safeGetInt(dbResult, "RDB$TRIGGER_TYPE");
         String description = JDBCUtils.safeGetStringTrimmed(dbResult, "RDB$DESCRIPTION");
         int systemFlag = JDBCUtils.safeGetInt(dbResult, "RDB$SYSTEM_FLAG");
-        boolean isSystem = true;
-        if (systemFlag == 0) { // System flag value 0 - if user-defined and 1 or more if system
-            isSystem = false;
-        }
+        boolean isSystem = systemFlag > 0; // System flag value 0 - if user-defined and 1 or more if system
 
         return new FireBirdTableTrigger(
                 parent,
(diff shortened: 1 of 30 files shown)


plugins/org.jkiss.dbeaver.ext.firebird/src/org/jkiss/dbeaver/ext/firebird/model/FireBirdMetaModel.java
```diff
@@ -128,11 +128,77 @@ public boolean supportsTriggers(@NotNull GenericDataSource dataSource) {
...
+            return null;
+        }
+        int sequence = JDBCUtils.safeGetInt(dbResult, "RDB$TRIGGER_SEQUENCE");
+        int type = JDBCUtils.safeGetInt(dbResult, "RDB$TRIGGER_TYPE");
+        String description = JDBCUtils.safeGetStringTrimmed(dbResult, "RDB$DESCRIPTION");
+        int systemFlag = JDBCUtils.safeGetInt(dbResult, "RDB$SYSTEM_FLAG");
+        boolean isSystem = true;
+        if (systemFlag == 0) { // System flag value 0 - if user-defined and 1 or more if system
+            isSystem = false;
+        }
```
ShadelessFox: Can be inlined because it looks prettier to me that way (IMHO):

```java
// System flag value 0 - if user-defined and 1 or more if system
boolean isSystem = systemFlag > 0;
```

Instead, you can do this directly in the QUERY itself:
```sql
SELECT ... RDB$SYSTEM_FLAG > 0 AS IS_SYSTEM ...
```

plugins/org.jkiss.dbeaver.ext.informix/src/org/jkiss/dbeaver/ext/informix/model/InformixTrigger.java
```diff
@@ -0,0 +1,92 @@
...
+    private final String beforeValue;
+    private final String afterValue;
+    private final String collation;
+
+    public InformixTrigger(@NotNull GenericTableBase container, String name, @NotNull JDBCResultSet resultSet) {
+        super(container, name, null);
+        String eventTypeLetter = JDBCUtils.safeGetString(resultSet, "event");
+        if (CommonUtils.isNotEmpty(eventTypeLetter)) {
+            this.eventType = TriggerEventType.valueOf(eventTypeLetter);
+        }
```
ShadelessFox: This may lead to unchecked exceptions appearing all over the place. Use `CommonUtils.valueOf` instead with supplied default value.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.informix/src/org/jkiss/dbeaver/ext/informix/model/InformixTrigger.java
LonwoLonwo: Not sure about the default value for Event. Null is the best default value for this situation :)

ShadelessFox: You can specify `null` as default value :^)

Issue #11767, opened 2021-03-17
Import CSV - Error no integer size determination


Jrmy003:
Hello everyone,

It looks like the size of integer fields are not auto determined as the size of varchar filed is, the default integer size is used which can lead to issue during import.
Attached a file containing csv data lead to an issue while using csv import functionality.

I'm using SQLServer as SGBD.

[test.log](https://github.com/dbeaver/dbeaver/files/6157235/test.log)

![image](https://user-images.githubusercontent.com/3603542/111486897-2d212f80-8738-11eb-82fd-1e2e3b7513e4.png)

This error happened

Data transfer failed during batch insert
(you can disable batch insert in order to skip particular rows).

Reason:
SQL Error [8115] [S0002]: Une erreur de dépassement arithmétique s'est produite lors de la conversion de expression en type de données int.

uslss:
Hi @Jrmy003 ,
could you try to uncheck Disable batches option in import settings and check if it helps?
![image](https://user-images.githubusercontent.com/31996417/111966754-53fcae80-8b08-11eb-87a7-b8702111f4d8.png)

ShadelessFox:
@uslss,

Unfortunately, this will not help.

This issue is created from #11759. We don't determine column length for numeric values during import from csv as we do for textual values.

flambchap:
Hi @Jrmy003. I've just successfully imported the CSV file you provided using the latest version of DBeaver. Moreover, the length of the integer field was set to 8. Can you confirm that the problem is gone? Thanks.

flambchap:
Never mind, successfully reproduced. Working on it.

PR #13246, opened 2021-07-19
#11767 find wide enough int type when performing csv -> database transfer

flambchap:


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/importer/DataImporterCSV.java
```diff
@@ -111,18 +111,38 @@ public DataImporterCSV() {
...
                                     columnInfo.setMaxLength(length);
                                 }
                                 break;
                             case NUMERIC:
+                                if (prevDataKind == DBPDataKind.UNKNOWN) {
+                                    columnInfo.setDataKind(dataType.getFirst());
+                                    columnInfo.setTypeName(dataType.getSecond());
+                                } else if (DatabaseTransferUtils.DATA_TYPE_INTEGER.equals(dataType)) {
+                                    columnInfo.setDataKind(dataType.getFirst());
+                                    long val = Long.parseLong(line[i]);
```
serge-rider, line 130: Long.parseLong throws exception in case of invalid input. Use try/catch.

Issue #11917, opened 2021-03-29
Copy table & column rename creates wrong DDL


elring:
OS: Linux Mint 20
Dbeaver: 21.0.2 EA (and 21.0.0)
DB: PostgreSQL 12

After table copy (for example Ctrl+C and Ctrl+V in table list) new table is "pre-created". When I immediately rename some column in this table (without previous saving), the result DDL is wrong. The table is created with new column name, that's ok. But then follows wrong additional rename SQL clause. This rename SQL should not be created at all.

See the table DDL (original column was testchar, new column name is testchar_newname).

```
CREATE TABLE public.testtable_1 (
	id serial NOT NULL,
	testchar_newname varchar NULL,
	CONSTRAINT testtable_pkey_1 PRIMARY KEY (id)
);
ALTER TABLE public.testtable_1 RENAME COLUMN testchar TO testchar_newname;
```

uslss:
thanks for the report

flambchap:
Fixed.

Matvey16:
Verified
#13358 New related issue opened

PR #13319, opened 2021-07-26
#11917 take into account that we can rename newly created object

flambchap:


LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/SQLObjectEditor.java
```diff
@@ -453,8 +453,19 @@ public String getNewName() {
...
+                if (object.equals(getObject()) && object instanceof DBPNamedObject2) {
+                    DBPNamedObject2 namedObject = (DBPNamedObject2) object;
+                    namedObject.setName(newName);
+                    return prevCommand;
+                }
+            }
+
             // We need very first and very last rename commands. They produce final rename
-            final String mergeId = "rename" + getObject().hashCode();
+            String mergeId = "rename" + getObject().hashCode();
```
ShadelessFox: So rude 🙁

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/SQLObjectEditor.java
serge-rider: Not sure that we need to  call setName here. It is a work for command reflector

serge-rider (changes requested):
Not sure that we need to call setName here. It is a work for command reflector

flambchap:
More than that, we don't even need to do anything about it. The only thing we need to care about is to throw away all rename commands.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/SQLObjectEditor.java
```diff
@@ -453,8 +453,22 @@ public String getNewName() {
...
         public DBECommand<?> merge(DBECommand<?> prevCommand, Map<Object, Object> userParams) {
+            // We need to dismiss all rename commands if there is a create command in the command queue.
+            // Otherwise we issue redundant rename commands
+            // See https://github.com/dbeaver/dbeaver/issues/11917
+            int hashCode = getObject().hashCode();
+            String createId = "create#" + hashCode;
+            Object createCmd = userParams.get(createId);
+            if (createCmd instanceof SQLObjectEditor.ObjectCreateCommand) {
+                return (SQLObjectEditor<?, ?>.ObjectCreateCommand) createCmd;
+            }
```
ShadelessFox: Knowing that `userParams.get(createId)` will always result in object of type `SQLObjectEditor.ObjectCreateCommand` (because you're adding it to the map later), this instance check is redundant and `createCmd` can be safely casted to `SQLObjectEditor.ObjectCreateCommand`. After that, use simple null check.

ShadelessFox (approved):

Issue #12734, opened 2021-06-04
NPE appears when a user without permissions closes a table


kseniiaguzeeva:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 21.1 EA
#### Connection specification:
- Database name and version PostgreSQL
#### Steps to reproduce, if exist:
1. Login by a user without permissions to see a table data
2. Open and close a table

**Expected**: the table is closed
**Actual result**: the table is closed with NPE error

https://user-images.githubusercontent.com/51405061/120793848-17a1f180-c540-11eb-8414-c60f7c1775df.mov

```
!ENTRY org.jkiss.dbeaver.model 4 0 2021-06-04 13:36:27.301
!MESSAGE Event loop exception
!SUBENTRY 1 org.jkiss.dbeaver.model 4 0 2021-06-04 13:36:27.303
!MESSAGE 
!STACK 0
java.lang.NullPointerException
    at org.eclipse.ui.operations.OperationHistoryActionHandler.dispose(OperationHistoryActionHandler.java:221)
    at org.eclipse.ui.texteditor.AbstractTextEditor.registerUndoRedoAction(AbstractTextEditor.java:5606)
    at org.eclipse.ui.texteditor.AbstractTextEditor.dispose(AbstractTextEditor.java:4366)
    at org.eclipse.ui.texteditor.AbstractDecoratedTextEditor.dispose(AbstractDecoratedTextEditor.java:392)
    at org.jkiss.dbeaver.ui.editors.text.BaseTextEditor.dispose(BaseTextEditor.java:104)
    at org.jkiss.dbeaver.ui.editors.sql.SQLEditorBase.dispose(SQLEditorBase.java:511)
    at org.eclipse.swt.widgets.RunnableLock.run(RunnableLock.java:40)
    at org.eclipse.swt.widgets.Synchronizer.runAsyncMessages(Synchronizer.java:185)
    at org.eclipse.swt.widgets.Display.runAsyncMessages(Display.java:4179)
    at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3839)
    at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1157)
    at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:338)
    at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1046)
[...]

PR #13328, opened 2021-07-27
#12734 Don't dispose editor asynchronously

ShadelessFox:


flambchap (approved):

serge-rider:
There was a reason to dispose it in aync mode.
Need to test it in modal/non-modal dialogs.

Issue #12649, opened 2021-05-29
Objects without headers - postgres


plotn:
Hi! Still remains old bug, and I possibly found a way to reproduce it.

1. We work with database as commons.
2. The connection lost (in my case there is normal network and internet, but vpnis swithched off, so db is unreachable)
3. I expand database object tree, select and double click object - editor appears.
4. Object is shown without header (I never switch off the headers), e.g.:

declare l_pid int8 = pg_backend_pid(); 
declare l_ts timestamp = clock_timestamp();
declare l_cnt int8;
begin
....

And we cannot do anything with it

uslss:
@plotn , what dbeaver version do you use?

plotn:
always maximally fresh

elring:
Hi. I have this problem regulary too. Usually when I suspend my PC with db left open and next day I continue my work... 
At first the function source has only text "-- Loading 'test_function(params)' source..." and then (after it reconnects to db) the function source has only the function body. When I close the function and reopen it, it still has only body. Possible solution is to refresh the functions in Database Navigator.

plotn:
exactly the same, plus this effect: https://github.com/dbeaver/dbeaver/issues/12559 - "disconnected windows (tabs) remains disconnected forever"

flambchap:
Fixed.

Matvey16:
Verified

@plotn I would like to ask you to also verify if this was fixed. Our tests were artificial and don't cover complete picture, because we were not able to reproduce this issue naturally. New version with this fix will be released on Monday.

PR #13362, opened 2021-07-29
#12649 reconnect is connection is lost when we read procedure body

flambchap:


LonwoLonwo (approved):

serge-rider:
This should be done for all databases, not just postgresql. 
Use `tryExecuteRecover`

Issue #13360, opened 2021-07-29
Support of the INSERT with default and non-default values


uslss:
#### System information: 
- Operating system (distribution) and version win 10
- DBeaver version 21.1.4

#### Describe the problem you're observing:
the side-effect of the #12786 .
we need to support Insert for rows that contains default and non-default values. for now it is fails with an error: 
```
[92] ERROR:  null value in column "id" of relation "invoices" violates not-null constraint
[92] DETAIL:  Failing row contains (null, null, null, null).
[92] STATEMENT:  INSERT INTO accounting.invoices (id,created_at,updated_at,deleted_at)
VALUES ($1,$2,$3,$4)
```
Table DDL:

```
CREATE TABLE accounting.invoices (
	id uuid NOT NULL DEFAULT uuid_generate_v4(),
	created_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
	updated_at timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
	deleted_at timestamp NULL,
	CONSTRAINT invoices_pkey PRIMARY KEY (id)
);
```

Matvey16:
Verified

PR #13373, opened 2021-07-30
#13360 check the columns on nullability for insert default values

LonwoLonwo:
2021-07-30 10_15_14-Window


serge-rider:
Not sure..

Issue #13399, opened 2021-08-02
Buttons too close to each other in user properties for MySQL


Matvey16:
When editing properties for user in MySQL `Check All` and `Clear All` buttons are too close to each other and almost merge into one button.

![dbeaver_Yzo4emPgem](https://user-images.githubusercontent.com/82543000/127823533-359d9ec8-351e-41b5-934d-d08d60c17287.png)

Matvey16:
Verified

PR #13400, opened 2021-08-02
#13399 Add padding between buttons in MySQL user editor

ShadelessFox:
image


serge-rider:
Use createComposite instead

Issue #11268, opened 2021-02-09
PostgreSQL connection settings advanced user role doesn't work as expected


giulio-quaresima:
#### System information: 
- Linux 4.19.0-14-amd64 #1 SMP Debian 4.19.171-2 (2021-01-30) x86_64 GNU/Linux
- OpenJDK Runtime Environment (build 11.0.9.1+1-post-Debian-1deb10u2)
- DBeaver Version 7.3.4.202101311042

#### Connection specification:
- PostgreSQL Server 9.3.25
- PostgreSQL JDBC 42.2.5
- sslmode=prefer

#### Describe the problem you're observing:
I have a user with GRANTS to query some views in a given schema. The grants are assigned to a role which I own, so to query the views I have to execute the command `SET ROLE 'MY_ROLE';` before querying the views.

When I configured the connection, I have put 'MY_ROLE' in the Connection settings > Advanced > User role field, expecting the `SET ROLE` command to be called immediately after the connection opening, but if I try to surf the views from the navigator, showing their content using the "View data" function, I get a "permission deneid" error. On the other hand, if in the SQL Editor I execute the `SET ROLE` command, then I can `SELECT` all the views without errors.

I'm not pretty sure whether this is a bug or I misinterpreted the use of the "User role" setting. If this is a "work as expected" case, how can I workaround the issue? For example, is there a way to configure an initialization SQL script to put the `SET ROLE` command in?

Thank you

kseniiaguzeeva:
Thank you for your report.

kseniiaguzeeva:
I cannot reproduce the issue in [the latest 21.1.3 version](http://dbeaver.io/)? Could I ask you to check the issue from your side?

giulio-quaresima:
Just tried: same error for version 21.1.3.202107181810

flambchap:
Does SET ROLE 'MY_ROLE' statement appear in the [query manager](https://github.com/dbeaver/dbeaver/wiki/Query-Manager) when you connect to the database? Make sure to tick "Utility functions" before checking that. I also want to ask you to send [logs](https://github.com/dbeaver/dbeaver/wiki/Log-files) which would be very helpful. Thank you.

giulio-quaresima:
Dear Nikita,

following your suggestion, I carefully read the logs and I found the
solution: it is a quotation problem.

It happens that if one set a role in the "user role" field of the
connection properties, say MY_GREAT_ROLE, when DBeaver opens the connection,
[...]

PR #13412, opened 2021-08-03
#11268 always enquote session role & change session role field name and tooltip

flambchap:
https://github.com/dbeaver/dbeaver-ee/pull/1057

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
@@ -267,14 +267,15 @@ public class PostgreExecutionContext extends JDBCExecutionContext implements DBC
         }
     }
 
-    void setSessionRole(final DBRProgressMonitor monitor) throws DBCException {
+    private void setSessionRole(@NotNull DBRProgressMonitor monitor) throws DBCException {
         final String roleName = getDataSource().getContainer().getConnectionConfiguration().getProviderProperty(PostgreConstants.PROP_CHOSEN_ROLE);
         if (CommonUtils.isEmpty(roleName)) {
             return;
         }
         try (JDBCSession session = openSession(monitor, DBCExecutionPurpose.UTIL, "Set active role")) {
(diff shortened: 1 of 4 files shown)


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
```diff
@@ -267,14 +267,15 @@ private void setSearchPath(String path) {
...
-    void setSessionRole(final DBRProgressMonitor monitor) throws DBCException {
+    private void setSessionRole(@NotNull DBRProgressMonitor monitor) throws DBCException {
         final String roleName = getDataSource().getContainer().getConnectionConfiguration().getProviderProperty(PostgreConstants.PROP_CHOSEN_ROLE);
         if (CommonUtils.isEmpty(roleName)) {
             return;
         }
         try (JDBCSession session = openSession(monitor, DBCExecutionPurpose.UTIL, "Set active role")) {
             try (JDBCStatement dbStat = session.createStatement()) {
-                dbStat.executeUpdate("SET ROLE " + roleName);
+                String sql = "SET ROLE " + getDataSource().getSQLDialect().getQuotedIdentifier(roleName, false, true);
```
ShadelessFox, line 277: Let's use prepared statements here instead 🙂

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
flambchap: Let's not do that for the following reasons: 

1. In the general case, you can not bind everything using JDBC's `?`. AFAIK, it works 100% reliably only when binding column values;
2. There is no point in defending against SQL injection here;
3. Prepared statements are generally more expensive to create which makes them pointless if they are not reused.

ShadelessFox: Good points, okay.

ShadelessFox (approved):

Issue #10918, opened 2021-01-11
Tables and views from another project could be added to ERD and saved, but after ERD reopening it is not shown


uslss:
#### System information: 
- Operating system (distribution) and version win 10 x64
- DBeaver version 7.3.3 EA

#### Describe the problem you're observing:
Tables and views from another project could be added to ERD and saved, but after ERD reopening they are not shown
#### Steps to reproduce, if exist:
1. Create ERD using File - New - ERD in Project 1
2. Add tables and/or views from Connection from Project 2 to created ERD and save it - No warnings that objects from another project will not be saved are shown
3. Close ERD and then open it - No previously added objects are visible.

tati-kru:
We must restrict adding entities from other projects.

ShadelessFox:
Implemented. Now you can't add objects for project that differs from diagram's project

DariaMarkaryan:
Verified

PR #13413, opened 2021-08-03
#10918 Load/save data sources from different projects

ShadelessFox:
java_QhgQcavjrT

### DIFF
--- a/plugins/org.jkiss.dbeaver.erd.model/src/org/jkiss/dbeaver/erd/model/ERDPersistedState.java
+++ b/plugins/org.jkiss.dbeaver.erd.model/src/org/jkiss/dbeaver/erd/model/ERDPersistedState.java
@@ -54,6 +54,7 @@ public class ERDPersistedState {
     public static final String ATTR_NAME = "name";
     public static final String ATTR_TIME = "time";
     public static final String ATTR_ALIAS = "alias";
+    public static final String ATTR_PROJECT = "project";
     public static final String ATTR_ID = "id";
     public static final String ATTR_ORDER = "order";
     public static final String ATTR_TRANSPARENT = "transparent";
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.erd.model/src/org/jkiss/dbeaver/erd/model/ERDPersistedState.java
```diff
@@ -96,10 +97,19 @@
...
                 for (Element dsElem : XMLUtils.getChildElementList(entitiesElem, TAG_DATA_SOURCE)) {
+                    String dsProjectName = dsElem.getAttribute(ATTR_PROJECT);
+                    DBPProject project = projectMeta;
+                    if (!CommonUtils.isEmpty(dsProjectName)) {
+                        // Get owner project, if any
+                        project = DBWorkbench.getPlatform().getWorkspace().getProject(dsProjectName);
+                        if (project == null) {
+                            continue;
+                        }
+                    }
```
flambchap, line 108: I believe it makes sense to extract this code into a helper funtion.

flambchap (changes requested):

plugins/org.jkiss.dbeaver.erd.model/src/org/jkiss/dbeaver/erd/model/ERDPersistedState.java
ShadelessFox: I don't think it deserves a helper function, since it is used just in one place and performs a particular job.

What the helper function will look like?

```java
public static DBPProject getProjectByName(@NotNull DBPProject defaultProject, @Nullable String projectName) {
    if (CommonUtils.isNotEmpty(projectName) {
        return DBWorkbench.getPlatform().getWorkspace().getProject(projectName);
    }
    return defaultProject;
}
```

I think it's pointless.

flambchap (approved):

serge-rider:
Not sure. Project can be renamed..

Issue #10984, opened 2021-01-15
Display column pin state in RSV filter settings


ShadelessFox:
Currently we're not displaying pin state of columns in filter settings dialog, so pinning/unpinning multiple columns may be tedious.

ShadelessFox:
Implemented (you are allowed to pin only top-level attributes):

![image](https://user-images.githubusercontent.com/35821147/128236774-b9eba187-165b-41f3-89bf-e4935bc7fcd1.png)

uslss:
verified

PR #13421, opened 2021-08-03
#10984 Display pin state in filter settings dialog

ShadelessFox:
image


serge-rider:
👍 
Column controller rocks.

Issue #12926, opened 2021-06-21
Only one entity highlighted when search items by name part in ERD


DariaMarkaryan:
#### System information: 
- Windows 10
- DBeaver 21.1.1 CE
#### Steps to reproduce:
 -Go to ERD with many tables in it
 -Click on the magnifier btn in the bottom ERD menu 
![image](https://user-images.githubusercontent.com/43252954/122768945-c4080580-d2ac-11eb-8475-b02a3602ddab.png)
 -Type in letters that contains in several items in the opened ERD and press Enter

 ##### Expected:
All elements that contains entered letters highlighted 
 ##### Actual:
Only one item highlighted 
![image](https://user-images.githubusercontent.com/43252954/122768394-39270b00-d2ac-11eb-864d-a9324a579171.png)

flambchap:
Fixed.

Matvey16:
Verified

PR #13473, opened 2021-08-08
#12926 vastly simplify searcher in erd editors; add support for multiple selected editParts

flambchap:


ShadelessFox (approved):
deletions > additions, cool!

Issue #11425, opened 2021-02-19
Use multi-row INSERTs in data transfer


serge-rider:
We need this to speedup data load.
Currently it works terribly slow on some platforms like Redshift.
Using multi-row batches may increase performance dramatically.

deepaksenthilkumar:
Is this planned for Monday's release team?

LonwoLonwo:
Nope, sorry @deepaksenthilkumar 
It will be implemented in the 21.1.4 release (02 August).
Now runs the testing stage.

LonwoLonwo:
We added the ability to use a multi-row insert with data import for better performance. You can turn it on on the data transfer settings page:

![2021-07-21 14_24_57-DBeaver Enterprise 21 1 3 - mood_data](https://user-images.githubusercontent.com/45152336/126481368-6f23b0ca-d8fd-490b-876b-d1faeccef11a.png)

It is the database-specific feature because not all databases support multi-row insert syntax like:

```sql
INSERT INTO table_name (column1, column2) VALUES
         (1,'First'),
	 (2,'Second'),
	 (3,'Third');
```

So it works with: 
- PostgreSQL
- Redshift
- Greenplum
- Cockroach
- MySQL/MariaDB
- SQLServer
- SQLite
- DB2
- Exasol

You can't use a multi-row insert with the "Disable batches" setting.

deepaksenthilkumar:
Thanks @LonwoLonwo. Which version of DBeaver has this functionality. Is there an early build that we can use now?

LonwoLonwo:
Hello @deepaksenthilkumar 
The current early version doesn't contain this feature, maybe a few days later it will be updated. You can check it here: https://dbeaver.io/files/ea/ (Release date must be at least 21.07)

Matvey16:
Some inconsistency in the UI, checkbox becomes available after returning with **Back** button

![uljX5oXwdb](https://user-images.githubusercontent.com/82543000/126604303-fb4d771d-db15-41a6-b76d-6782dd47aee5.gif)

LonwoLonwo:
Fixed.

Matvey16:
Verified

deepaksenthilkumar:
We built this from the developer branch and I tried to import an excel file with 1.5 Million records and 3 columns (for Redshift)

col1 varchar(180)
loaddate varchar(18)
cnt (int4)

[...]

Issue #13283, opened 2021-07-21
Error obtaining geometry type from typmod: fff7f


gtorinsd:
Hello Everybody,
I can't view GEOMERTY column in DBeaver (Redshift)
There is a error:

**Error obtaining geometry type from typmod: fff7f** 

OS: Winddows 10
DBeaver v 21.1.3.202107181810, community
Database: Amazon redshift, no ssh, socks etc.

I can select records, view geomerty objects as a line etc, but I can't view the column in columns list,
The same situation was in a previos versions of DBeaver.

![image](https://user-images.githubusercontent.com/11867927/126501514-4bdcc01c-4646-4450-baf2-170dd67dd409.png)
[bak_0.log](https://github.com/dbeaver/dbeaver/files/6856246/bak_0.log)
To reproduce the issue, just open columns list.

Script to create table is:

**drop table public.some_table_name;
CREATE TABLE IF NOT EXISTS public.some_table_name
(
	id BIGINT  DEFAULT "identity"(1285223, 0, ('1,1'::character varying)::text) ENCODE az64
	,field1 GEOMETRY   ENCODE RAW
	,name VARCHAR(150)   ENCODE lzo
	,value DOUBLE PRECISION   ENCODE RAW
	,created_at timestamp default CURRENT_TIMESTAMP
)
DISTSTYLE AUTO
;
ALTER TABLE public.some_table_nameowner to kuku;**

Thanks!
Best wishes,
Torin.

ShadelessFox:
Hello @gtorinsd,

Thanks for the report.

May I ask you to execute following query and provide the result:
```sql
select * from pg_catalog.pg_attribute where attrelid = <table_oid_here>;
```

It does not contain any sensitive data, only metadata. `Object ID` can be found under <kbd>Properties</kbd> tab of the problematic table.

ShadelessFox:
I did some research and it seems that it's invalid value. 

Also, pgAdmin4 was unable to display the value too:

![](https://i.imgur.com/58OcFi9.png)

ShadelessFox:
Hello @gtorinsd,

Is the issue still relevant to you? Please take look at https://github.com/dbeaver/dbeaver/issues/13283#issuecomment-884228194

ShadelessFox:
My guess is that this specific column has type `geometry(geometrycollection)`, but I need more information.

deepaksenthilkumar:
CREATE TABLE test_geom (c1_geometry geometry)

[...]

PR #13482, opened 2021-08-10
Changes for performance of Redshift multirow insert

deepaksenthilkumar:
The following issue didn't get fixed properly for Redshift 
https://github.com/dbeaver/dbeaver/issues/11425

The problem is in the Redshift JDBC Driver class PGEscaper which becomes exponentially slow with the number of question marks in the INSERT statement. 

Added RedshiftValueHandlers that implement DBDValueBinder that prevents the use of ? in the code.

This gives a huge performance improvement in Redshift for 1 M records with 3 columns. 

It took 17 hours (with multirow insert--Only possible setting is batch size 100 anything more hangs and commit every 100k records) to 4 mins now (batch size 10k, commit every 100k records).

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreValueHandlerProvider.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreValueHandlerProvider.java
@@ -122,6 +122,7 @@ public class PostgreValueHandlerProvider extends JDBCStandardValueHandlerProvide
 	validNumericTypes.add("integer");
 	validNumericTypes.add("bigint");
 	validNumericTypes.add("decimal");
+	validNumericTypes.add("numeric");
 	validNumericTypes.add("double precision");
 	validNumericTypes.add("real");
 	validNumericTypes.add("float");
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreValueHandlerProvider.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreValueHandlerProvider.java
@@ -45,6 +45,25 @@ import org.jkiss.dbeaver.model.struct.DBSTypedObject;
  * PostgreValueHandlerProvider
  */
 public class PostgreValueHandlerProvider extends JDBCStandardValueHandlerProvider {
+    
+    private static List<String> validRSNumericTypes = new ArrayList<String>();
+    
+    static {
+	validRSNumericTypes.add(PostgreConstants.TYPE_INT2);
+	validRSNumericTypes.add(PostgreConstants.TYPE_INT4);
+	validRSNumericTypes.add(PostgreConstants.TYPE_INT8);
+	validRSNumericTypes.add(PostgreConstants.TYPE_FLOAT4);
+	validRSNumericTypes.add(PostgreConstants.TYPE_FLOAT8);
+	validRSNumericTypes.add("smallint");
+	validRSNumericTypes.add("integer");
+	validRSNumericTypes.add("bigint");
+	validRSNumericTypes.add("decimal");
+	validRSNumericTypes.add("numeric");
+	validRSNumericTypes.add("double precision");
+	validRSNumericTypes.add("real");
+	validRSNumericTypes.add("float");
+    }
+    
     @Nullable
     @Override
     public DBDValueHandler getValueHandler(DBPDataSource dataSource, DBDFormatSettings preferences,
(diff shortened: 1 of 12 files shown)


deepaksenthilkumar:
In the regression of dbeaver#13482
Noticed fff7f is the hexstring returned by Redshift for geometry.

Added code to handle that scenario.

ShadelessFox:
Thanks for the contribution, but unfortunately we can't accept it.

**Please create a new PR that will contain code that is related to the issue itself.**

1. Instead of creating a bunch of dummy classes and a separate method `getRedshiftValueHandler` in [`PostgreValueHandlerProvider`](https://github.com/dbeaver/dbeaver/pull/13482/files#diff-4229eeda6249bc37a298ada521202bde5681285d186e26300b738e8034800101R131), you can achieve the same result using  `PostgreDialect#escapeScriptValue`, which is more convenient.
2. What's the point of modifying `JDBCTable` [here](https://github.com/dbeaver/dbeaver/pull/13482/files#diff-ec8c005fda2c83e87dd065ef3de7def455f6254f6aecf327ead339b27d199671R367-R373)? By doing that, you broke multirow insert in SQL Server.
3. The issue with geometry type is unrelated, so it must not be included here. Although you can provide more information about the problem in #13283

Also, looks like something went wrong with the formatting, fix it too. We use spaces instead of tabs, please follow our code style.

ShadelessFox:
My eye can't catch a particular change that led to this improvement in performance. Can you highlight me the key changes so we can help cooperate with you?

deepaksenthilkumar:
The key change is to not use ? and ind variables directly (it's a redshift Postgres version bug). If we bind it directly then it caused issue in Data viewer insert (JDBC table). 

@ShadelessFox -Do you mean we can get the dialect and edit PostgreStringValueHandler instead of adding a RedshiftStringValueHandler? This bug is specific to Redshift. So, I thought it's better to just handle it for Redshift classes.

ShadelessFox:
What do you mean by the bug? The issue was in a poor performance of multirow insert. Have you tried installing different driver versions? Is the performance the same?

I don't get what you mean by `bind values directly`, isn't `?` the only way to achieve that in prepared statements?

deepaksenthilkumar:
The issue is when you use multi row inserts, the number of question marks increase in the prepared statement and the Redshift driver (Postgres flavor) becomes terribly slow (exponentially proportional to the number of question marks) at handling a regex check they have. 

The biggest problem is we can’t use 2.x drivers as the current implementation of DBeaver is reliant on Postgres flavor of the driver that has this performance issue. Using  2.x redshift driver breaks many other things like navigator, generation of DDL, etc. And even in the 1.x version, DBeaver has currently recommended to stop upgrade due to other issues.

Based on what I have tested I could see performance of 1M inserts is extremely fast (4 mins vs 17 hours) by directly setting the values in the insert instead of using the ? and binding prepared statement. I didn’t want to change anything for Postgres, as the issue is specific to Redshift. I can change my fix to use Postgres dialect and the value handler classes for redshift which will skip the use of bind variables. I assume that’s why DBDBinder exists? 

Can we do this just for multi row insert of Redshift without impacting any other database?

Pasting my comments that I posted last week last week in https://github.com/dbeaver/dbeaver/issues/11425
 
I checked further and noticed the problem is in the Redshift JDBC Driver class PGEscaper which becomes exponentially slow with the number of question marks in the INSERT statement. I made a couple of crude changes (just to confirm that) and noticed now I am able to insert 1M records in 8 mins. (with a setting of 10k multi-row insert batch size and commit after 10k records)

In `ExecuteInsertBatchImpl` hardcode the string (should handle it better)
```java
if (valueHandler instanceof DBDValueBinder) {
    valuesPart.append(((DBDValueBinder) valueHandler).makeQueryBind(attribute, attributeValues[i]));
} else if (useMultiRowInsert) {
    valuesPart.append(attributeValues[i] == null ? null : "'" + attributeValues[i].toString() + "'");
} else {
    valuesPart.append("?"); //$NON-NLS-1$
}
```

Avoided binding in `ExecuteBatchWithMultipleInsert`:
```java
private void bindAndFlushStatement(DBDValueHandler[] handlers, DBCStatistics statistics, DBCStatement batchStatement, Object[] allMultiInsertValues) throws DBCException {
    statistics.setQueryText(batchStatement.getQueryString());
    statistics.addStatementsCount();
    if (batchStatement.getQueryString().contains("?")) {
        bindStatement(handlers, batchStatement, allMultiInsertValues);
    }
    batchStatement.addToBatch();
    flushBatch(statistics, batchStatement);
}
```

Any suggestion on how we make this fix proper would be helpful. Thanks for your help in introducing this feature.

ShadelessFox:
The main problem with manual binding is SQL injection. For example, string value may contain some malicious payload, and we don't want to allow user to shoot himself in the foot. I don't think we want to follow Redshift approach and invent our own version of `PGescaper` just to escape values. This is the primary reason of my concern.

The idea of passing values directly without binding them first is a great, since bind operation is expensive and it can drastically improve data transfer performance not just for Redshift but for other databases too.

I can suggest you to provide a new option in Data Transfer Wizard (so user can toggle it) to bypass binding values. You can pass this option in `prepareQueryForStatement` and if it is set, then pass values directly (or at least use `SQLDialect#escapeScriptValue` to somehow escape values) without binding them. And this option must be exposed in the UI with descriptive tip with its cons and pros.

You can create a separate feature request and work on it.

deepaksenthilkumar:
Thank @ShadelessFox that makes total sense. I will do that.

PR #13485, opened 2021-08-10
#12649 always throw exception when we are unable to read procedure body

flambchap:


ShadelessFox (approved):

serge-rider:
Need to test this on different PG forks (Redshift, Cockroach, Greenplum)

ShadelessFox (approved):

PR #13488, opened 2021-08-10
#10918 Use currently active project

ShadelessFox:


serge-rider:
Not sure it should work like this..

plugins/org.jkiss.dbeaver.erd.model/src/org/jkiss/dbeaver/erd/model/DiagramObjectCollector.java
```diff
@@ -213,11 +211,8 @@ private boolean aliasExist(String alias) {
                 }
             }
             final DBPProject project = ((DBSObject) object).getDataSource().getContainer().getProject();
-            if (currentProject == null) {
-                currentProject = project;
-            }
-            if (currentProject != project) {
-                diagram.addErrorMessage("Can't add object '" + DBUtils.getObjectFullName(object, DBPEvaluationContext.UI) + "' from different project '" + project.getName() + "' (current project is '" + currentProject.getName() + "')");
+            if (activeProject != project) {
+                diagram.addErrorMessage("Can't add object '" + DBUtils.getObjectFullName(object, DBPEvaluationContext.UI) + "' from a non-active project '" + project.getName() + "' (active project is '" + activeProject.getName() + "')");
```
flambchap: If I understand this correctly, if the user selects multiple objects from a non-active project and tries to add them to a diagram, we will show them a bunch of consequent identical error messages. That doesn't look good.

ShadelessFox:
This PR does not handle situations when we're editing ERD from another project.

Maybe we should ask the diagram what project it's from instead of relying on currently active project?

plugins/org.jkiss.dbeaver.erd.model/src/org/jkiss/dbeaver/erd/model/DiagramObjectCollector.java
ShadelessFox: Improved:

java_T1KZzbUIOR

PR #13522, opened 2021-08-13
introduce MessageBoxModern

flambchap:
Screenshot 2021-08-13 112527


plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/dialogs/MessageBoxBuilder.java
```diff
@@ -0,0 +1,103 @@
...
+    public MessageBoxBuilder setPrimaryImage(@NotNull DBPImage image) {
+        dialog.setPrimaryImage(image);
+        return this;
+    }
+
+    @NotNull
+    public MessageBoxBuilder setReplies(@NotNull Reply... replies) {
+        this.replies = new ArrayList<>(Arrays.asList(replies));
+        return this;
+    }
```
ShadelessFox, line 64: Maybe use `EnumSet` here? Otherwise user may supply duplicated elements

ShadelessFox (commented):
API looks really clean 🙂

Issue #13498, opened 2021-08-11
Performance improvement of multi-row inserts with the bypass of binding values


deepaksenthilkumar:
**Is your feature request related to a problem? Please describe.**
Performance issues in multi-row inserts of millions of records for the Redshift database. The issue is because of the time the Redshift 1.x JDBC driver takes to prepare a statement (which is exponentially proportional to the number of question marks in the statement we are trying to prepare). 

**Describe the solution you'd like**
Provide a new option in Data Transfer Wizard (so the user can toggle it) to bypass binding values. You can pass this option in prepareQueryForStatement and if it is set, then pass values directly (or at least use SQLDialect#escapeScriptValue to somehow escape values) without binding them. And this option must be exposed in the UI with the descriptive tip with its cons and pros.

**Describe alternatives you've considered**
The single row insert is slow too, so we do need to implement this solution.

**Additional context**
N/A

Matvey16:
@deepaksenthilkumar Something went wrong because with this option toggled on, only one row is inserted millions of times
![6B53Kp6xbx](https://user-images.githubusercontent.com/82543000/130256525-7468521e-82d0-4196-8285-599e3fb83972.gif)

serge-rider:
Fixed.
Note: multi-row mode and value binding are independent parameters now.

Binding disable + big multi-row value increases data transfer performance dramatically.

deepaksenthilkumar:
I noticed the bug. Thanks @serge-rider for fixing it, but using replace of ? is 6x slower during the replace. I will work on appending it to stringbuilder directly.

deepaksenthilkumar:
Submitted the PR for a faster/accurate version of skip bind values. Thanks @Matvey16  for finding the issue!

serge-rider:
Umm, from my expirience 99% of time it spends in insert itself and in network operations. Optimizing StringBuilder operations doesn't make much sense.

My concern is that we have a lot of different modes (batch, single/multi row, disabled bindings, databases which don't support batches, etc). All these cases must be tested.

deepaksenthilkumar:
In this case, the slowness is because of how the replace method internally works vs. just append. I measured timing and it was just doing replace for 50-70 seconds for every run (100k multirow insert size)

I had the same concern and tested all the below options as I didn't want to break anything.

These are the options that I tested.
1. Single record (multirow insert+no skip bind value, disable batches+no skip bind value, multirow insert+skip bind value, disable batches+skip bind value)
[...]

PR #13535, opened 2021-08-15
Added option to skip bind values during multi-row insert

deepaksenthilkumar:
Has changes for the below feature request
https://github.com/dbeaver/dbeaver/issues/13498

Thanks, @ShadelessFox for providing the approach. I have implemented the same.


deepaksenthilkumar:
Hi @LonwoLonwo 

I noticed @akilovich mentioned in another OceanBase plugin thread that he would be on a vacation.

Do you know tentatively when you would be able to review this PR and which release this can be target to be released on? 

It will be a huge performance improvement for our end users who are using DBeaver to access Redshift.

serge-rider:
Generally looks good. I think that we may enable this parameter by default for some drivers (like Redshift).

I'll review it in details later this week.

deepaksenthilkumar:
@serge-rider  Thanks.

serge-rider:
Let's merge it. We can improve UIX later.

Issue #13536, opened 2021-08-15
Redshift: No Schema Binding views column datatypes are inconsistent and length are missing in the navigator


deepaksenthilkumar:
#### System information: 
- Operating system (distribution) and version: Windows 10
- DBeaver version: 21.1.4

#### Connection specification:
- Database name and version: Redshift 1.0.29551 
- Driver name: https://s3.amazonaws.com/redshift-downloads/drivers/jdbc/1.2.45.1069/RedshiftJDBC42-no-awssdk-1.2.45.1069.jar
- Do you use tunnels or proxies (SSH, SOCKS, etc) SSH

#### Describe the problem you're observing:
Table and Normal View have the right column data types/length
![image](https://user-images.githubusercontent.com/36084950/129468718-d09009df-c129-4ca2-b6e5-2c167b0a24b5.png)
![image](https://user-images.githubusercontent.com/36084950/129468726-4c639ff9-750c-4614-89ee-50f588706696.png)

No schema binding View has a different data type (variant) and column length (precision/scale) are missing/different
![image](https://user-images.githubusercontent.com/36084950/129468764-bcab7092-f584-45a6-987e-9a6168616048.png)

#### Steps to reproduce, if exist:

Create the following table, views, and check the output.

Table:
CREATE TABLE IF NOT EXISTS test_all_types
(
	c1 SMALLINT   ENCODE az64
	,c2 INTEGER   ENCODE az64
	,c3 BIGINT   ENCODE az64
	,c4 NUMERIC(18,0)   ENCODE az64
	,c5 REAL   ENCODE RAW
	,c6 DOUBLE PRECISION   ENCODE RAW
	,c7 BOOLEAN   ENCODE RAW
	,c8 CHAR(1)   ENCODE lzo
	,c9 VARCHAR(256)   ENCODE lzo
	,c10 DATE   ENCODE az64
	,c11 TIMESTAMP WITHOUT TIME ZONE   ENCODE az64
	,c12 TIMESTAMP WITH TIME ZONE   ENCODE az64
	,c15 TIME WITHOUT TIME ZONE   ENCODE az64
	,c16 TIME WITH TIME ZONE   ENCODE az64
)
DISTSTYLE AUTO
;

No Schema Binding View:
CREATE VIEW v_test_all_types_nb AS
[...]

PR #13541, opened 2021-08-15
Resolved the data type names and column type length, precision/scale for Redshift

deepaksenthilkumar:
Fix for https://github.com/dbeaver/dbeaver/issues/13536

Use the below scripts
CREATE TABLE IF NOT EXISTS test_all_types_2
(
	c1 SMALLINT   ENCODE az64
	,c2 INTEGER   ENCODE az64
	,c3 BIGINT   ENCODE az64
	,c4 NUMERIC(18,0)   ENCODE az64
	,c5 REAL   ENCODE RAW
	,c6 DOUBLE PRECISION   ENCODE RAW
	,c7 BOOLEAN   ENCODE RAW
	,c8 CHAR(1)   ENCODE lzo
	,c9 VARCHAR(256)   ENCODE lzo
	,c10 DATE   ENCODE az64
	,c11 TIMESTAMP WITHOUT TIME ZONE   ENCODE az64
	,c12 TIMESTAMP WITH TIME ZONE   ENCODE az64
	,c15 TIME WITHOUT TIME ZONE   ENCODE az64
	,c16 TIME WITH TIME ZONE   ENCODE az64
	,c17 NUMERIC(18,7)   ENCODE az64
)
DISTSTYLE AUTO
;

CREATE VIEW usr_senthidx.v_test_all_types_b AS
SELECT a.c1, a.c2, a.c3, a.c4, a.c5, a.c6, a.c7, a.c8, a.c9, a.c10, a.c11, a.c12, a.c15, a.c16,a.c17 FROM test_all_types_2 a;

CREATE VIEW usr_senthidx.v_test_all_types_nb AS
SELECT a.c1, a.c2, a.c3, a.c4, a.c5, a.c6, a.c7, a.c8, a.c9, a.c10, a.c11, a.c12, a.c15, a.c16,a.c17 FROM test_all_types_2 a
WITH no SCHEMA binding;

After applying the fix table, binding view, no schema binding view-all look the same!

Table:
image

image

It's a simple fix of using viewColumn.setFullTypeName to get the column length/precision, scale!

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
@@ -19,6 +19,7 @@ package org.jkiss.dbeaver.ext.postgresql.model.impls.redshift;
 import org.jkiss.code.NotNull;
 import org.jkiss.dbeaver.DBException;
 import org.jkiss.dbeaver.Log;
+import org.jkiss.dbeaver.ext.postgresql.PostgreConstants;
 import org.jkiss.dbeaver.ext.postgresql.model.PostgreDataType;
 import org.jkiss.dbeaver.ext.postgresql.model.PostgreSchema;
 import org.jkiss.dbeaver.ext.postgresql.model.PostgreTableColumn;
@@ -97,44 +98,22 @@ public class RedshiftView extends PostgreView
                         String colName = JDBCUtils.safeGetString(dbResult, "col_name");
                         String colType = JDBCUtils.safeGetString(dbResult, "col_type");
                         int colNum = JDBCUtils.safeGetInt(dbResult, "col_num");
-
-                        PostgreTableColumn viewColumn = new PostgreTableColumn(this);
-                        viewColumn.setName(colName);
-                        viewColumn.setPersisted(true);
-
-                        String colTypeName;
-                        int colPrecision = -1, colScale = -1;
-                        int divPos = colType.indexOf("(");
-                        if (divPos != -1) {
-                            colTypeName = colType.substring(0, divPos);
-                            int divPos2 = colType.lastIndexOf(")");
-                            if (divPos2 < 0) {
-                                log.error("Bad column type name '" + colType + "'");
-                                continue;
-                            }
-                            String typeMod = colType.substring(divPos + 1, divPos2).trim();
-                            divPos2 = typeMod.indexOf(",");
-                            if (divPos2 < 0) {
-                                colPrecision = CommonUtils.toInt(typeMod);
-                            } else {
-                                colPrecision = CommonUtils.toInt(typeMod.substring(0, divPos2).trim());
-                                colScale = CommonUtils.toInt(typeMod.substring(divPos2 + 1).trim());
-                            }
-                        } else {
-                            colTypeName = colType;
-                        }
+                        String colTypeName = DBUtils.getTypeModifiers(colType).getFirst();
+                        String resolvedColTypeName = PostgreConstants.DATA_TYPE_ALIASES.get(colTypeName);
+                        colTypeName = resolvedColTypeName == null ? colTypeName : resolvedColTypeName;
 
                         PostgreDataType dataType = (PostgreDataType) getDataSource().resolveDataType(monitor, colTypeName);
                         if (dataType == null) {
                             log.error("Column type name '" + colType + "' not found");
                             continue;
                         }
+                        PostgreTableColumn viewColumn = new PostgreTableColumn(this);
+                        viewColumn.setName(colName);
+                        viewColumn.setFullTypeName(colType);
+                        viewColumn.setPersisted(true);
                         viewColumn.setDataType(dataType);
-                        if (colPrecision >= 0) viewColumn.setPrecision(colPrecision);
-                        if (colScale >= 0) viewColumn.setScale(colScale);
                         viewColumn.setOrdinalPosition(colNum);
                         columns.add(viewColumn);
-
                     }
                     return columns;
                 }
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/PostgreServerRedshift.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/PostgreServerRedshift.java
@@ -20,7 +20,6 @@ import org.jkiss.code.NotNull;
 import org.jkiss.code.Nullable;
 import org.jkiss.dbeaver.DBException;
 import org.jkiss.dbeaver.Log;
-import org.jkiss.dbeaver.ext.postgresql.PostgreConstants;
 import org.jkiss.dbeaver.ext.postgresql.model.*;
 import org.jkiss.dbeaver.ext.postgresql.model.impls.PostgreServerExtensionBase;
 import org.jkiss.dbeaver.model.DBPErrorAssistant;
@@ -40,7 +39,6 @@ import org.osgi.framework.Version;
 import java.sql.SQLException;
 import java.util.Arrays;
 import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -294,14 +292,6 @@ public class PostgreServerRedshift extends PostgreServerExtensionBase implements
     @Override
     public boolean supportSerialTypes() {
         return false;
-    }
-
-    @Override
-    public Map<String, String> getDataTypeAliases() {
-        Map<String, String> aliasMap = new LinkedHashMap<>(super.getDataTypeAliases());
-        aliasMap.put("character", PostgreConstants.TYPE_BPCHAR);
-        aliasMap.put("character varying", PostgreConstants.TYPE_VARCHAR);
-        return aliasMap;
     }
 
     @Override


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
```diff
@@ -98,43 +99,34 @@ private boolean isViewVwithNoSchemaBinding(DBRProgressMonitor monitor) throws DB
...
+			if (divPos != -1) {
+			    colTypeName = colType.substring(0, divPos);
+			    int divPos2 = colType.lastIndexOf(")");
+			    if (divPos2 < 0) {
+				log.error("Bad column type name '" + colType + "'");
+				continue;
+			    }
+			} else {
+			    colTypeName = colType;
+			}
```
ShadelessFox: Instead of manually doing this, I suggest you to use `DBUtils#getTypeModifiers` instead. See its javadoc for usage

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
```diff
@@ -98,43 +99,34 @@ private boolean isViewVwithNoSchemaBinding(DBRProgressMonitor monitor) throws DB
...
+                        
+			String resolvedColTypeName = PostgreConstants.DATA_TYPE_ALIASES.get(colTypeName);
```
ShadelessFox: Better to use `PostgreServerExtension#getDataTypeAliases` here. I doubt it can be so simple, there's some special cases with `serial` data types and so on 🤔

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
```diff
@@ -98,43 +99,34 @@ private boolean isViewVwithNoSchemaBinding(DBRProgressMonitor monitor) throws DB
...
                             log.error("Column type name '" + colType + "' not found");
                             continue;
                         }
+                        PostgreTableColumn viewColumn = new PostgreTableColumn(this);
+                        viewColumn.setName(colName);
+                        viewColumn.setFullTypeName(colType);
+                        viewColumn.setPersisted(true);
                         viewColumn.setDataType(dataType);
-                        if (colPrecision >= 0) viewColumn.setPrecision(colPrecision);
-                        if (colScale >= 0) viewColumn.setScale(colScale);
```
ShadelessFox, line 134: I guess complex type handlers are not working anymore, so you can't, for example, edit string values.

You can at least use `PostgreTypeHandlerProvider#getTypeHandler` here and then `setTypeMod` on `viewColumn` using `PostgreTypeHandler#getTypeModifiers` using your newly acquired type handler. No need to update precision and scale after that though

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
deepaksenthilkumar: That piece of code was already there in that class. But will refactor. Thanks for the suggestion.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
deepaksenthilkumar: @ShadelessFox 
Redshift doesn’t have serial. I couldn’t test geometry as it has a different bug in DBeaver. Is there any other data type that comes to your mind?

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
deepaksenthilkumar: @ShadelessFox 
I can make this change. Can you pls elaborate on what you mean by edit of string values don’t work?

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
ShadelessFox: There's no other special cases, so maybe that's OK.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
ShadelessFox: There are several `PostgreTypeHandler` that are responsible for extracting precision/scale and other parameters from `typmod`. If the latter is not set, data type may be displayed in a weird way. The best example would be `numeric(15, 5)` type.

deepaksenthilkumar: Typemod was not available in the late binding view. That’s the reason I went with this approach. I do see numeric is displayed right (see screenshot  above in my first comment). 

Redshift already gives the data in a displayable format. 

The output from Redshift itself is numeric(18, 7).

While they retained typemod approach for tables and normal views, for late binding views all they provide is col_type. 

In hindsight, Why complicate it with type modifier, when current approach is working.

ShadelessFox: I see, okay. I forgot that `PostgreAttribute#setFullTypeName` already does that I just described to you.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
deepaksenthilkumar: Completed that change.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java
deepaksenthilkumar: Thanks @ShadelessFox. I completed the other change you suggested. Hope we are all good with this PR.

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreConstants.java
```diff
@@ -153,6 +153,9 @@
         DATA_TYPE_ALIASES.put("time without time zone", TYPE_TIME);
         DATA_TYPE_ALIASES.put("timestamp with time zone", TYPE_TIMESTAMPTZ);
         DATA_TYPE_ALIASES.put("timestamp without time zone", TYPE_TIMESTAMP);
+        
+        DATA_TYPE_ALIASES.put("character", TYPE_BPCHAR);
```
LonwoLonwo: https://github.com/dbeaver/dbeaver/blob/9679ae7766b65aa0544527a4f2712707890e79e3/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/PostgreServerRedshift.java#L269

I think, we already have it here.

deepaksenthilkumar: Yes, but that's not visible and the BPCHAR and VARCHAR are applicable to both Postgres and Redshift.
@LonwoLonwo-I have removed the Override in the PostgreServerRedshift.java and let it be handled from PostgreConstants

deepaksenthilkumar: @LonwoLonwo Does this change look good or was there any specific reason to keep varchar and bpchar in PosgreServerRedshift class alone?

deepaksenthilkumar: @serge-rider Any inputs on my question above. Should I do something different in the code?

deepaksenthilkumar: @LonwoLonwo Did you get a chance to look at my question? Was there any specific reason to keep varchar and bpchar in PosgreServerRedshift class alone?

ShadelessFox: @deepaksenthilkumar Actually, I already added these aliases some time ago, so you don't have to:
https://github.com/dbeaver/dbeaver/blob/ab82296b92eaf6ea7a135e4da017617078b26179/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreConstants.java#L148-L151

deepaksenthilkumar: @ShadelessFox-Cool. Then I will do one thing... Rebase the code based on current devel branch and then just change 
[RedshiftView.java](plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/)

deepaksenthilkumar: @ShadelessFox - I have completed that. Now the changes are just in 

[RedshiftView](plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/RedshiftView.java)
And
[PostgreServerRedshift](plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/PostgreServerRedshift.java)

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreConstants.java
ShadelessFox: 👍🏻

Issue #13544, opened 2021-08-16
Redshift Data Editor-Editing of Date/Timestamp columns fails


deepaksenthilkumar:
#### System information: 
- Operating system (distribution) and version: Windows 10
- DBeaver version: 21.1.4

#### Connection specification:
- Database name and version: Redshift 1.0.29551 
- Driver name: https://s3.amazonaws.com/redshift-downloads/drivers/jdbc/1.2.45.1069/RedshiftJDBC42-no-awssdk-1.2.45.1069.jar
- Do you use tunnels or proxies (SSH, SOCKS, etc) SSH

#### Describe the problem you're observing:
Data Editor doesn't work for insert or Update of Date/Timestamp columns in Redshift

#### Steps to reproduce, if exist:

Create below table
CREATE TABLE IF NOT EXISTS test_all_types
(
	c1 SMALLINT   ENCODE az64
	,c2 INTEGER   ENCODE az64
	,c3 BIGINT   ENCODE az64
	,c4 NUMERIC(18,0)   ENCODE az64
	,c5 REAL   ENCODE RAW
	,c6 DOUBLE PRECISION   ENCODE RAW
	,c7 BOOLEAN   ENCODE RAW
	,c8 CHAR(1)   ENCODE lzo
	,c9 VARCHAR(256)   ENCODE lzo
	,c10 DATE   ENCODE az64
	,c11 TIMESTAMP WITHOUT TIME ZONE   ENCODE az64
	,c12 TIMESTAMP WITH TIME ZONE   ENCODE az64
	,c15 TIME WITHOUT TIME ZONE   ENCODE az64
	,c16 TIME WITH TIME ZONE   ENCODE az64
	,"-c10" VARCHAR(10)   ENCODE lzo
)
DISTSTYLE AUTO
;

Open Data Editor, try to add date/timestamp data. Following error occurs

![image](https://user-images.githubusercontent.com/36084950/129505756-3dc605b8-891f-486e-8041-bff7ae8b21a9.png)

deepaksenthilkumar:
I was fixing the PostgreDateTimeValueHandler but noticed the Redshift JDBC driver 1.x doesn't work for bind variables with timestamptz. It works for Date, Timestamp but not for timestamptz. 

Probably we should have a skip bind value option for data editor too. 
@ShadelessFox @serge-rider @LonwoLonwo Thoughts?

Probably just fix date, timestamp at this point...

serge-rider:
[...]

PR #13605, opened 2021-08-21
Fix for Redshift Data Editor-Editing of Date/Timestamp column

deepaksenthilkumar:
https://github.com/dbeaver/dbeaver/issues/13544

The Date/Timestamp columns are fixed now. Redshift doesn't know about Types.OTHER, it has to be Types.VARCHAR.

The 1.x Redshift JDBC driver doesn't work fine for timestamp with timezone, time, timetz columns edits from data editor

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreDateTimeValueHandler.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreDateTimeValueHandler.java
@@ -17,6 +17,8 @@
 package org.jkiss.dbeaver.ext.postgresql.model.data;
 
 import org.jkiss.code.NotNull;
+import org.jkiss.dbeaver.ext.postgresql.model.PostgreDataSource;
+import org.jkiss.dbeaver.ext.postgresql.model.impls.redshift.PostgreServerRedshift;
 import org.jkiss.dbeaver.model.data.DBDFormatSettings;
 import org.jkiss.dbeaver.model.exec.DBCException;
 import org.jkiss.dbeaver.model.exec.DBCSession;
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreDateTimeValueHandler.java
```diff
@@ -64,15 +66,18 @@ public Object getValueFromObject(@NotNull DBCSession session, @NotNull DBSTypedO
...
+		if (((PostgreDataSource) session.getDataSource()).getServerType() instanceof PostgreServerRedshift) {
+		    ((JDBCPreparedStatement) statement).setObject(index + 1, value.toString(), Types.VARCHAR);
+		} else {
+		    ((JDBCPreparedStatement) statement).setObject(index + 1, value.toString(), Types.OTHER);
+		}
+	    } catch (SQLException e) {
+		throw new DBCException(ModelMessages.model_jdbc_exception_could_not_bind_statement_parameter, e);
+	    }
+	    return;
+	}
```
ShadelessFox: Please reformat indentation using spaces instead of tabs and then squash these commits into just one

ShadelessFox (changes requested):

ShadelessFox (approved):

ShadelessFox (approved):
Didn't notice this one :)

serge-rider:
We use Types.OTHER i nmany other PostgreSQL value handlers.
Also I don't like `instanceof` checks in base PG code.

Probably we need some more general solution. E.g. add add `bindObjectParameter` function in PostgreServerExtension.

deepaksenthilkumar:
@serge-rider Agree with you. I just copied the same pattern in [PostgreValueHandlerProvider
](https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreValueHandlerProvider.java)

Can you elaborate further on the approach, can we have Redshift specific Value handler?

serge-rider:
I've add extra API in ServerExtension.

Fix pushed in https://github.com/dbeaver/dbeaver/commit/8f4b7c3f6f5240652a4e71b57a5e4a25fe1886ec

Issue #13606, opened 2021-08-21
Redshift specific functions and keywords aren't highlighted in a different color


deepaksenthilkumar:
**Is your feature request related to a problem? Please describe.**
Some of the Redshift-specific functions, keywords are not indicated in a different color in the SQL editor.

Some of the examples are:
"SYSDATE"
"COPY",
"DATASHARE",
"LIBRARY",
"MATERIALIZED",
"MODEL",
"REFRESH",
"UNLOAD",
"VACUUM"
"NVL",
"NVL2"
**Describe the solution you'd like**
We can add PostgreDialect to support Redshift-specific keywords and functions.

Current Behavior:
![image](https://user-images.githubusercontent.com/36084950/130331409-1346a09c-4260-416b-a3a3-f1694d12358b.png)

Expected Behavior:
![image](https://user-images.githubusercontent.com/36084950/130331422-c1216d0b-4297-43f2-998a-4a59a518f2f2.png)

Initially, we can add the keywords/functions listed here and eventually cover all the keywords/reserved words/functions in
https://docs.aws.amazon.com/redshift/latest/dg/cm_chap_SQLCommandRef.html

uslss:
verified

PR #13607, opened 2021-08-21
Added Redshift specific functions, keywords in PostgreDialect

deepaksenthilkumar:
Fix for Redshift specific functions and keywords aren't highlighted in a different color #13606

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDialect.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDialect.java
@@ -44,6 +44,7 @@ import org.jkiss.utils.CommonUtils;
 
 import java.sql.Types;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 
 /**
@@ -756,6 +757,11 @@ public class PostgreDialect extends JDBCSQLDialect implements TPRuleProvider {
 
         // #12723 Redshift driver returns wrong infor about unquoted case
         setUnquotedIdentCase(DBPIdentifierCase.LOWER);
+    }
+    
+    @Override
+    public void addKeywords(Collection<String> set, DBPKeywordType type) {
+        super.addKeywords(set, type);
     }
 
     @NotNull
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDialect.java
```diff
@@ -716,6 +737,14 @@ public void initDriverSettings(JDBCSession session, JDBCDataSource dataSource, J
...
 
         addExtraKeywords(POSTGRE_EXTRA_KEYWORDS);
+
+        // Add Extra keywords from Redshift
+        if (((PostgreDataSource) dataSource).getServerType() instanceof PostgreServerRedshift) {
+            addExtraKeywords(REDSHIFT_EXTRA_KEYWORDS);
+            addKeywords(Arrays.asList(REDSHIFT_OTHER_TYPES_FUNCTION), DBPKeywordType.OTHER);
+            addExtraFunctions(REDSHIFT_FUNCTIONS_CONDITIONAL);
+        }
+
```
ShadelessFox: You can move this snippet a bit lower, so you can use `serverExtension` field from `PostgreDialect` itself:

https://github.com/dbeaver/dbeaver/blob/b32d53917b49383a5987051c9c5492e016fd0023/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDialect.java#L781-L784

ShadelessFox: Actually, it would be better to override `PostgreServerExtension#configureDialect` in `PostgreServerRedshift` and add keywords to the dialect there

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDialect.java
deepaksenthilkumar: Good idea! Done and commits are squashed.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDialect.java
```diff
@@ -757,6 +758,11 @@ public void initDriverSettings(JDBCSession session, JDBCDataSource dataSource, J
         // #12723 Redshift driver returns wrong infor about unquoted case
         setUnquotedIdentCase(DBPIdentifierCase.LOWER);
     }
+    
+    @Override
+    public void addKeywords(Collection<String> set, DBPKeywordType type) {
+        super.addKeywords(set, type);
+    }
```
ShadelessFox, line 765: I don't think it's needed

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDialect.java
```diff
@@ -44,6 +44,7 @@
 
 import java.sql.Types;
 import java.util.Arrays;
+import java.util.Collection;
```
ShadelessFox, line 47: I don't think it's needed

ShadelessFox (changes requested):

ShadelessFox (approved):
👍🏻

PR #13609, opened 2021-08-22
Skip Bind Values performance tuning.

deepaksenthilkumar:
As part of #13498 

Instead of setting ? and then replacing, this technique of appending is 6x faster (on the string manipulation piece).

For 1.2 Million records import of 3 columns, it takes 4 mins with this technique (I verified the import with different data in each line this time around) vs. setting it to ? and then performing replace takes 25 mins.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
@@ -16,6 +16,7 @@
  */
 package org.jkiss.dbeaver.model.impl.data;
 
+import org.eclipse.core.runtime.Assert;
 import org.jkiss.code.NotNull;
 import org.jkiss.code.Nullable;
 import org.jkiss.dbeaver.model.DBPDataSource;
@@ -32,7 +33,10 @@ import org.jkiss.dbeaver.model.struct.DBSDataManipulator;
 import org.jkiss.dbeaver.model.struct.rdb.DBSTable;
 import org.jkiss.utils.CommonUtils;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
+import java.util.StringJoiner;
 
 public class ExecuteInsertBatchImpl extends ExecuteBatchImpl {
 
@@ -110,6 +114,9 @@ public class ExecuteInsertBatchImpl extends ExecuteBatchImpl {
         DBSTable table,
         boolean useMultiRowInsert,
         Map<String, Object> options) throws DBCException {
+        
+        Assert.isLegal(attributes.length == handlers.length);
+        Assert.isLegal(useMultiRowInsert || attributes.length == attributeValues.length);
 
         // Make query
         String tableName = DBUtils.getEntityScriptName(table, options);
@@ -143,6 +150,7 @@ public class ExecuteInsertBatchImpl extends ExecuteBatchImpl {
             return query;
         }
         boolean hasKey = false;
+        List<Integer> usedAttributes = new ArrayList<Integer>();
         for (int i = 0; i < attributes.length; i++) {
             DBSAttributeBase attribute = attributes[i];
             if (DBUtils.isPseudoAttribute(attribute) || (!useMultiRowInsert && (!allNulls && DBUtils.isNullValue(attributeValues[i])))) {
@@ -157,56 +165,33 @@ public class ExecuteInsertBatchImpl extends ExecuteBatchImpl {
                 attributeName = DBUtils.getObjectFullName(table.getDataSource(), attribute, DBPEvaluationContext.DML);
             }
             query.append(attributeName);
+            usedAttributes.add(i);
         }
         query.append(")\n\tVALUES "); //$NON-NLS-1$
-        StringBuilder valuesPart = new StringBuilder(64);
-        valuesPart.append("(");
-        hasKey = false;
+        
+        int usedAttributesArr[] = usedAttributes.stream().mapToInt(Integer::intValue).toArray();
+
+        StringJoiner valuesPart = new StringJoiner(",");
         boolean skipBindValues = CommonUtils.toBoolean(options.get(DBSDataManipulator.OPTION_SKIP_BIND_VALUES));
-        for (int i = 0; i < attributes.length; i++) {
-            DBSAttributeBase attribute = attributes[i];
-            if (DBUtils.isPseudoAttribute(attribute) || (!useMultiRowInsert && (!allNulls && DBUtils.isNullValue(attributeValues[i])))) {
-                continue;
-            }
-            if (hasKey) valuesPart.append(","); //$NON-NLS-1$
-            hasKey = true;
-
-            DBDValueHandler valueHandler = handlers[i];
-            
-            if (valueHandler instanceof DBDValueBinder) {
-                valuesPart.append(((DBDValueBinder) valueHandler).makeQueryBind(attribute, attributeValues[i]));
-            } else {
-                valuesPart.append("?"); //$NON-NLS-1$
-            }
-        }
-        valuesPart.append(")"); //$NON-NLS-1$
-        if (useMultiRowInsert) {
-            for (int i = 0; i < attributeValues.length / attributes.length; i++) {
-                if (i != 0) {
-                    query.append(",");
+        for (int i = 0; i < attributeValues.length / attributes.length; i++) {
+            StringJoiner rowValuesPart = new StringJoiner(",", "(", ")");
+            for (int j : usedAttributesArr) {
+                DBSAttributeBase attribute = attributes[j];
+                DBDValueHandler valueHandler = handlers[j];
+                int k = i * attributes.length + j;
+                if (valueHandler instanceof DBDValueBinder) {
+                    rowValuesPart.add(((DBDValueBinder) valueHandler).makeQueryBind(attribute, attributeValues[k]));
+                } else if (skipBindValues) {
+                    rowValuesPart.add(SQLUtils.convertValueToSQL(session.getDataSource(), attribute, valueHandler, attributeValues[k], DBDDisplayFormat.NATIVE));
+                } else {
+                    rowValuesPart.add("?");
                 }
-                query.append(valuesPart);
             }
-        } else {
-            query.append(valuesPart);
+            valuesPart.add(rowValuesPart.toString());
         }
 
-        if (skipBindValues) {
-            // Here we replace all binding marks (?) with real values
-            int curPos = 0;
-            for (int i = 0; i < attributeValues.length; i++) {
-                String sqlValue = SQLUtils.convertValueToSQL(
-                    session.getDataSource(),
-                    attributes[i % attributes.length],
-                    handlers[i % handlers.length],
-                    attributeValues[i],
-                    DBDDisplayFormat.NATIVE);
-
-                int qmPos = query.indexOf("?", curPos);
-                query.replace(qmPos, qmPos + 1, sqlValue);
-            }
-        }
-
+        query.append(valuesPart);
+        
         String trailingClause = method.getTrailingClause(table, session.getProgressMonitor(), attributes);
         if (trailingClause != null) {
             query.append(trailingClause);


deepaksenthilkumar:
@serge-rider 
Reposting from the issue as it's closed now.

In this case, the slowness is because of how the replace method internally works vs. just append. I measured timing and it was just doing replace for 50-70 seconds for every run (100k multirow insert size-with 3 columns)

I had tested all the below options as I didn't want to break anything.

These are the options that I tested.

1. Single record (multirow insert+no skip bind value, disable batches+no skip bind value, multirow insert+skip bind value, disable batches+skip bind value)
2. 2k+ records (multirow insert+no skip bind value, disable batches+no skip bind value, multirow insert+skip bind value, disable batches+skip bind value)
3. 1M+ records (multirow insert+skip bind value)
The modified version functionally does the same thing as what and you had for skip bind values and @LonwoLonwo for multirow respectively. It just binds the variable in a loop vs. doing replace later.

This does the same work as replace but in a quicker way.

Pls consider this change as it will help in faster data imports.

deepaksenthilkumar:
@ShadelessFox @serge-rider Did you get a chance to check this. This gives significant performance improvement for the skip bind values functionality

ShadelessFox:
Sorry for long response. I'll take a look on it later today.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -160,52 +160,43 @@ StringBuilder prepareQueryForStatement(
...
+                } else {
+                    if (skipBindValues) {
+                        valuesPart.append(SQLUtils.convertValueToSQL(session.getDataSource(), attributes[i], valueHandler, attributeValues[k], DBDDisplayFormat.NATIVE));
+                    } else {
+                        valuesPart.append("?"); //$NON-NLS-1$
+                    }
                 }
-                query.append(valuesPart);
             }
-        } else {
```
ShadelessFox, line 190: What is the point of such changes? Can't we just add a new `if else` branch here checking for `skipBindValues`?

deepaksenthilkumar: We don't want to add ? and then do replace, that is slow. Instead, we can just do it in the original for loop with an append which is extremely fast.

deepaksenthilkumar: Or, we can ignore the valuespart for loop in the beginning and do it in a totally new if outside, which might be cleaner?

ShadelessFox: You can loop over `attributeValues` instead of `attributes` to bind values properly and also support `useMultiRowInsert`. Both `handler` and `attribute` can be received using `array[i % array.length]`.

deepaksenthilkumar:
Was thinking to avoid same modulus calculated over and over again.

The complexity is the same if you have 10 records and 3 attributes, I do
10*3 and you are suggesting 1*30?
-- 
SD

ShadelessFox:
Calculating modulus is not a really expensive operation on modern CPUs. At least you can make proof of concept to see whether it will perform good or bad.

deepaksenthilkumar:
@ShadelessFox I didn't see any difference in performance using modulus. So, I modified the code just to change the skipBindValues portion. I see a difference in performance close to 6.5x-7x improvement. 30 mins for 1 Million 3 columns vs. 4 mins!

deepaksenthilkumar:
@ShadelessFox-Also noticed the replace is exponentially slow with number of columns increasing to 25+. The append is extremely fast!

ShadelessFox:
I don't really the `query.indexOf("\n\tVALUES ")` part. Maybe you can move `skipBindValues` check above, so it will not generate `VALUES` clause in the first place.

Actually, I was talking about rewriting and embedding this logic in the main loop above to eliminate this additional loop entirely.

deepaksenthilkumar:
I don’t like that part too, but ended up doing it to not touch other part
of the code. I will do that. --
SD

deepaksenthilkumar:
@ShadelessFox I embedded this logic in the main loop now and eliminated the additional loop totally.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -160,53 +160,47 @@ StringBuilder prepareQueryForStatement(
...
         query.append(")\n\tVALUES "); //$NON-NLS-1$
         StringBuilder valuesPart = new StringBuilder(64);
-        valuesPart.append("(");
         hasKey = false;
         boolean skipBindValues = CommonUtils.toBoolean(options.get(DBSDataManipulator.OPTION_SKIP_BIND_VALUES));
-        for (int i = 0; i < attributes.length; i++) {
-            DBSAttributeBase attribute = attributes[i];
-            if (DBUtils.isPseudoAttribute(attribute) || (!useMultiRowInsert && (!allNulls && DBUtils.isNullValue(attributeValues[i])))) {
+        for (int i = 0; i < attributeValues.length; i++) {
+            DBSAttributeBase attribute = attributes[i % attributes.length];
```
ShadelessFox: Let's avoid repeating `i % attributes.length` by introducing local variable containing the result of this expression.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -160,53 +160,47 @@ StringBuilder prepareQueryForStatement(
...
+            if (DBUtils.isPseudoAttribute(attribute) || (!useMultiRowInsert && (!allNulls && DBUtils.isNullValue(attributeValues[i % attributes.length])))) {
                 continue;
             }
```
ShadelessFox: I don't like repeating this either, the same statement is located just a several lines above. Also the latter one is wrong and does not respect nulls from rows > 0.

Can we filter attributes only one and do it the right way and just use this filtered list later on?

As a bonus, we will able to avoid having `hasKey`, since it's just an index of the first attribute.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -160,53 +160,47 @@ StringBuilder prepareQueryForStatement(
...
-            hasKey = true;
 
-            DBDValueHandler valueHandler = handlers[i];
-            
+            if (i % attributes.length == 0) {
+                valuesPart.append("(");
+                hasKey = true;
+            }
+
+            DBDValueHandler valueHandler = handlers[i % attributes.length];
```
ShadelessFox: Either assert that `handlers.length == attributes.length` either use `i % handlers.length`. I prefer the first one.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
deepaksenthilkumar: Went with i % handlersf.length

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
deepaksenthilkumar: Added 3 new attributes to have the filtered attributes, atrributevalues, handlers there by we dont have to check it in the loop over and over. 
        DBSAttributeBase[] attributesf = new DBSAttributeBase[attributes.length - skipAttributes.size()];
        Object[] attributeValuesf = new Object[attributeValues.length - (attributes.length * skipAttributes.size())];
        DBDValueHandler[] handlersf = new DBDValueHandler[handlers.length - skipAttributes.size()];

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
deepaksenthilkumar: Done.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -159,54 +163,76 @@ StringBuilder prepareQueryForStatement(
...
+                if (skipAttributes.contains(j)) {
+                    continue;
+                }
+                attributeValuesf[++k] = attributeValues[i];
+            }
+        } else {
+            attributesf = attributes;
+            handlersf = handlers;
+            attributeValuesf = attributeValues;
+        }
```
ShadelessFox: Perhaps it may work, it's really bug-prone.

There's two ways to improve readability:
* Use `List<>` for storing filtered objects, that way you don't have to track the last indexes.
* Use `Arrays.stream(...)` and perform filtering in-place. Then use `.toArray` to convert stream into arrays.

And please use less-cryptic names for variables holding filtered objects. You can borrow original names and rename input parameters (just for this method) for something like `blabla` &rArr; `allBlaBla`

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
deepaksenthilkumar: Done!

deepaksenthilkumar: Done now! Applied the changes you suggested.

deepaksenthilkumar:
@ShadelessFox - Take a look when you can, I have updated the code.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -143,9 +148,11 @@ StringBuilder prepareQueryForStatement(
             return query;
         }
         boolean hasKey = false;
+        List<Integer> skipAttributes = new ArrayList<Integer>();
```
ShadelessFox: Why can't we have a list of attributes themselves, not just their indexes?

Update: I see why. Then let's inverse the meaning of this list by having _used_ attributes instead of _skipped_.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -159,54 +166,77 @@ StringBuilder prepareQueryForStatement(
...
+        Object[] filteredAttributeValues = null;
+
+        if (skipAttributes.size() > 0) {
+            List<DBSAttributeBase> attributesList = new ArrayList<DBSAttributeBase>(Arrays.asList(attributes));
+            filteredAttributes = 
+                    IntStream.range(0, attributesList.size())
+                            .filter(i -> !skipAttributes.contains(i))
+                            .mapToObj(attributesList::get)
+                            .collect(Collectors.toList())
+                            .stream().toArray(DBSAttributeBase[]::new);
```
ShadelessFox: See previous comment.

We can do following instead:
```suggestion
            filteredAttributes = usedAttributes.stream()
                .map(x -> attributes[x])
                .toArray(DBSAttributeBase[]::new);
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -159,54 +166,77 @@ StringBuilder prepareQueryForStatement(
...
+                            .collect(Collectors.toList())
+                            .stream().toArray(DBDValueHandler[]::new);
+            
+            List<Object> attributeValuesList = new ArrayList<Object>(Arrays.asList(attributeValues));
+            filteredAttributeValues = 
+                    IntStream.range(0, attributeValuesList.size())
+                            .filter(j -> !skipAttributes.contains(j % attributes.length))
+                            .mapToObj(attributeValuesList::get)
+                            .collect(Collectors.toList())
+                            .stream().toArray();
```
ShadelessFox: We have indexes of used attributes, what's the point of this list, anyway? Using some math we can calculate the desired index based on current attribute.

Pseudocode:
```java
for (int i = 0; i < attributes.length; i++) {
    for (int j = 0; j < values.length / attributes.length; j++) {
        var value = values[i * attributes.length + j];
    }
}

File: plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java [hunk 9]
```diff
@@ -159,54 +166,77 @@ StringBuilder prepareQueryForStatement(
...
-                    handlers[i % handlers.length],
-                    attributeValues[i],
-                    DBDDisplayFormat.NATIVE);
-
-                int qmPos = query.indexOf("?", curPos);
-                query.replace(qmPos, qmPos + 1, sqlValue);
+            if (j == filteredAttributes.length - 1) {
+                valuesPart.append(")"); //$NON-NLS-1$
+                if (i < filteredAttributeValues.length - 1) {
+                    valuesPart.append(","); //$NON-NLS-1$
```
ShadelessFox: Let's use `StringJoiner` for the purpose of joining values with comma. It can also help appending `(` and `)` manually.

ShadelessFox (changes requested):
We were discussing possible suggestions and improvements related to the code. I feel bad for requesting too much changes.

deepaksenthilkumar:
Fixed

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
deepaksenthilkumar: Fixed it to remove all extra filtered variables. Using usedAttributes for indexes of attributes that matter and used StringJoiner. Hope this looks good. I don't feel bad all the changes at all. And I do these in my spare time, so I don't try to code the most optimal way, which I would probably do going fwd :)

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
deepaksenthilkumar: Stopped using filtered attributes

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
deepaksenthilkumar: Yeah removed all filtered fields.

ShadelessFox:
I like this iteration. Have you tested it?

deepaksenthilkumar:
Yup with different settings and combinations. --
SD

flambchap (approved):

flambchap:
Hi @deepaksenthilkumar. The PR generally looks nice 👍

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -157,56 +161,36 @@ StringBuilder prepareQueryForStatement(
...
-            }
-        }
-        valuesPart.append(")"); //$NON-NLS-1$
-        if (useMultiRowInsert) {
-            for (int i = 0; i < attributeValues.length / attributes.length; i++) {
-                if (i != 0) {
-                    query.append(",");
+        for (int i = 0; i < attributeValues.length / attributes.length; i++) {
+            StringJoiner rowValuesPart = new StringJoiner(",", "(", ")"); //$NON-NLS-1$
+            for (int j : usedAttributes) {
```
flambchap: I have one idea in mind. Here in the inner loop, we continuously unbox `Integer` references. Maybe it's worthwhile to unbox them by hand once outside those loops into an `int[]` array and then iterate over it instead of `usedAttributes` list. However, I'm not sure if it will give us a substantial performance gain. Could you please try and test this?

deepaksenthilkumar: Good suggestion @akilovich int[] is faster than unboxing once we have at least 5 attributes (tested with million rows and 100 columns 100 times). I added that logic now!

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -157,56 +161,38 @@ StringBuilder prepareQueryForStatement(
...
+        for (int i = 0; i < attributeValues.length / attributes.length; i++) {
+            StringJoiner rowValuesPart = new StringJoiner(",", "(", ")"); //$NON-NLS-1$
+            for (int j : usedAttributesArr) {
+                DBSAttributeBase attribute = attributes[j];
+                DBDValueHandler valueHandler = handlers[j];
+                int k = i * attributes.length + j;
+                if (valueHandler instanceof DBDValueBinder) {
+                    rowValuesPart.add(((DBDValueBinder) valueHandler).makeQueryBind(attribute, attributeValues[k]));
+                } else {
+                    if (skipBindValues) {
```
ShadelessFox: Please inline this branch into single `else if` instead of having nested branches

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -157,56 +161,38 @@ StringBuilder prepareQueryForStatement(
...
-        }
-        valuesPart.append(")"); //$NON-NLS-1$
-        if (useMultiRowInsert) {
-            for (int i = 0; i < attributeValues.length / attributes.length; i++) {
-                if (i != 0) {
-                    query.append(",");
+        StringJoiner valuesPart = new StringJoiner(","); //$NON-NLS-1$
+        boolean skipBindValues = CommonUtils.toBoolean(options.get(DBSDataManipulator.OPTION_SKIP_BIND_VALUES));
+        for (int i = 0; i < attributeValues.length / attributes.length; i++) {
+            StringJoiner rowValuesPart = new StringJoiner(",", "(", ")"); //$NON-NLS-1$
```
ShadelessFox: The `$NON-NLS-1$` comment only covers the first string literal `","`. Also, it's obsolete and you can remove it entirely.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -157,56 +161,38 @@ StringBuilder prepareQueryForStatement(
...
-                    handlers[i % handlers.length],
-                    attributeValues[i],
-                    DBDDisplayFormat.NATIVE);
-
-                int qmPos = query.indexOf("?", curPos);
-                query.replace(qmPos, qmPos + 1, sqlValue);
+            valuesPart.add(rowValuesPart.toString());
+            if (!useMultiRowInsert) {
+                break;
             }
```
ShadelessFox: If multi-row insert is disabled, values and attributes should have same length. I think this statement can be safely avoided by adding an assertion in the beginning of the method (along with `handlers.length == attributes.length` mentioned some time ago), just to be extra sure. 

You can use `org.eclipse.core.runtime.Assert#isLegal` for this purpose. Assertions are better than assumptions.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
deepaksenthilkumar: Done

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
deepaksenthilkumar: Done

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
deepaksenthilkumar: Done

ShadelessFox (approved):
Looks very promising and clean :)

I'll test this on Monday.

deepaksenthilkumar:
I did some testing and fixed the assert logic! 
Also note the current code in 21.1.2 fails if there are ? in the file that we are trying to import. So this is not just a performance fix, but also a bug fix!

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
```diff
@@ -110,6 +114,11 @@ StringBuilder prepareQueryForStatement(
         DBSTable table,
         boolean useMultiRowInsert,
         Map<String, Object> options) throws DBCException {
+        
+        Assert.isLegal(attributes.length == handlers.length);
+        if (!useMultiRowInsert) {
+            Assert.isLegal(attributes.length == attributeValues.length);
+        }
```
ShadelessFox:
```suggestion
        Assert.isLegal(useMultiRowInsert || attributes.length == attributeValues.length);
```

Either we use multi row insert and don't care about this assumption, otherwise we use "single row insert" and length must be the same :)

ShadelessFox:
> Also note the current code in 21.1.2 fails if there are ? in the file that we are trying to import. So this is not just a performance fix, but also a bug fix!

Do you mean question mark inside values? Now I noticed this too. Blindly replacing `?` with actual values is a bit suspicious thing because it can lead to unforeseen consequences when value contains `?` too. Nice catch, by the way!

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteInsertBatchImpl.java
deepaksenthilkumar: Done!

ShadelessFox (approved):

deepaksenthilkumar:
@serge-rider @ShadelessFox - Is this targeted for 21.2.1?

ShadelessFox:
We haven't discussed it yet, unfortunately.

deepaksenthilkumar:
Mainly because of imports with ? Failing this might need to be a priority
fix!--
SD

ShadelessFox:
We are going to review it tomorrow.

serge-rider:
Let's try it.
@deepaksenthilkumar, @ShadelessFox  Thanks for reviewing and patching 👍

Matvey16:
Verified

Issue #13611, opened 2021-08-22
In Redshift, during data import the column length has to be based on the byte length/octet length for string columns instead of length of the columns.


deepaksenthilkumar:
**Is your feature request related to a problem? Please describe.**
In Redshift, during data import, the column length has to be based on the byte length/octet length for string columns instead of the length of the columns.

**Describe the solution you'd like**
Calculate byte length and use if for field lengths.

deepaksenthilkumar:
@ShadelessFox Any suggestion on where to start. Is this configurable by database?

ShadelessFox:
I'm not sure what you're talking about. You're talking about the case when you import data to non-existing table in Redshift?

deepaksenthilkumar:
Yes.

ShadelessFox:
The target column's length is determined using the source attribute's length. Unfortunately it's a generic code and it can't handle such specific cases. You still can take look at it and maybe add yet another function to `SQLDialect`.

deepaksenthilkumar:
Can you point me to where the length is calculated now, I will see if we can modify the SQLDialect for Redshift.

ShadelessFox:
Sure, see `DBStructUtils#mapTargetDataType`.

deepaksenthilkumar:
Made it generic, as it might benefit any other database that needs to use this option.

Matvey16:
@deepaksenthilkumar I've tested this feature and found few issues.
1. The option doesn't work if you go back in import wizard and change it, it remembers the first choice.

![h5ZI3Q2dRr](https://user-images.githubusercontent.com/82543000/130787720-5f40c809-1c58-49d6-b733-628660a95f04.gif)

2. I performed an import with this option disabled (which obviously led to a fail, because it tried to insert values with multiple 2 byte symbols after creating column with lenght based on number of characters) But in any additional import after that, this feature was not working (see gif below, the longest value is 77 bytes, not 60)

![1VERxM2lta](https://user-images.githubusercontent.com/82543000/130788253-20cc765c-eb91-4def-95a1-0a5aec5225a8.gif)


@ShadelessFox do you think, those issues can be solved in scope of only this ticket?

deepaksenthilkumar:
@Matvey16 - I will check and fix this.

deepaksenthilkumar:
This seems to be an existing problem (refresh of target DDL when settings change). Test it just based on column sample count
Have 3 records
1
10
100

[...]

PR #13615, opened 2021-08-22
Add ability to use byte length instead of a string length for column length as an Yes/No option in import

deepaksenthilkumar:
Fix for #13611 

Made it generic!


ShadelessFox:
But this fix only applies to CSV importer, what about other importers?

deepaksenthilkumar:
What other importers are there? I didn’t look at them. But can add there as
well. I was going by usage of Column Length. Didn’t find those constants
used elsewhere.--
SD

ShadelessFox:
There's also XLSX, XML and data transfer from other table (the latter is the most interesting one, but I doubt we should select sample the data just to figure its max length).

deepaksenthilkumar:
I didn’t realize you had xlsx import in the community edition. Do you have
it?

I will look into it.

ShadelessFox:
Nope, we don't, sorry!

deepaksenthilkumar:
So, only the csv data importer I suppose. And we wouldn’t check sample
length when we take it from another database.

deepaksenthilkumar:
@ShadelessFox-Should I do anything else then?

ShadelessFox:
I think that's it, then.

Issue #13432, opened 2021-08-04
[mssql] default value constraint(s) are not displayed


bobocaca:
My request is similar to #8448 , I faced the same issue when trying to delete field with default value (DF_* constraint depend on field, so can not delete), then I realize I didn't see the constraint in the tree view.

the following picture is the view from dbeaver
![image](https://user-images.githubusercontent.com/64453466/128102421-4b1d48db-68bf-4699-8643-7d89aaee8168.png)

view from SQL management studio
![image](https://user-images.githubusercontent.com/64453466/128102511-41efcb41-fe37-4377-8249-aca703ee3b8a.png)

I'm keen to have it also display in dbeaver, so I can perform my development work on different DB with the same SQL tool.
If the feature already available, please ignore my request.
Thanks.

LonwoLonwo:
https://github.com/dbeaver/dbeaver/issues/5896

ShadelessFox:
We still don't display default constraints in a separate tab, but now they're respected when dropping columns.

uslss:
verified

PR #13714, opened 2021-08-27
#13432 Drop dependent default constraint before dropping column

ShadelessFox:


LonwoLonwo (approved):

serge-rider:
I think we should cache constraint name in column object and not query it before delete.

Issue #11850, opened 2021-03-23
Password doesn't meet requirements


staze:
Just ran into this issue: https://github.com/dbeaver/dbeaver/issues/5175

Found the issue is dbeaver is trying to create the user with a blank password, then alter the user to add the password as two separate operations. The first one fails because the of password complexity rules. 

Removing the alter piece, and tying into the create fixes the issue. This is on MySQL 8, and current dbeaver (Version 21.0.1.202103211229).

uslss:
Hi @staze ,
did you try to use Mysql(not Mysql 5) driver and check if it helps?
![image](https://user-images.githubusercontent.com/31996417/112297301-c60afa00-8ca6-11eb-9c79-0cb3aaeac129.png)
for me works ok:
![image](https://user-images.githubusercontent.com/31996417/112297244-b5f31a80-8ca6-11eb-91db-f620a62af5c4.png)

staze:
It is using that driver. 

I'm connecting to an MySQL 8 Community server over SSH. This is how it generates the create user strings. 

![image](https://user-images.githubusercontent.com/1100483/112551899-0d64b980-8d7f-11eb-9fc3-601537bb3fee.png)

ShayArtzi:
I faced this issue as well with MySql 8 over SSH. As a workaround, I manually added the `IDENTIFIED BY` clause to the `CREATE USER` statement.

uslss:
verified

PR #13734, opened 2021-08-30
#11850 Include `IDENTIFIED BY` clause in `CREATE USER` statement

ShadelessFox:


serge-rider:
No clear how it fixes the problem.

ShadelessFox:
As the title says, it includes `IDENTIFIED BY` clause in `CREATE USER` statement if password is specified. This is the what the issue was about. When I was making other ticket about <kbd>Simple Password Check</kbd> extension, I thought that this syntax is only valid when the plugin is enabled. I was wrong and this syntax is available in MySQL 5.5 or even lower.

Issue #10149, opened 2020-10-23
Content Viewer data handling improvements


Zbambus:
I just jumped from community editiion 7.2.0 to 7.2.3.

The one fature change that surprised me was that the default view mode of content of a CLOB now was binary mode (hitting Shift+Enter when cell is chosen in result grid) . This **probably** makes sense as plain text is supposed to be handled by a CLOB column instaed of a BLOB.

Unfortunately I have inherited a database where the BLOB column is used to store both text and binary files in a ratio 99% text and 1% binary, and BLOB was chosen in order to accomodate the 1% cases, but in 99% of the cases when I want to see the content of the column, text is what I want to see.  Yes, not the best solution, but there it is!

I looked through preferences up to see if I could find somewhere to set the default mode of presentation, but was unable to locate this. That would have solved this situation nicely as I could just set such an parameter/option to "Text".

My next search involved if there is a key combination that would let me toggle between the modes ("Text"/"Binary"). That would involve one more command entered on the keyboard after opening the view of the content, but that was fine as opposed to now having to use the mouse to point on the "Text" tab and activate this, which is very time consuming as the mouse pointer might be anywhere on my 3-screen setup. I was unable to locate this as well.

So to sum it up : 

1) Is there a way to determine what presentation mode is to be default when viewing content from a BLOB column?
2) Is there a way to use the keyboard to toggle between the presentation modes when viewing content from a BLOB/CLOB column?
3) Is there a way to determine what presentation modes are to be available when viewing content from a BLOB/CLOB column? Today BLOB comes with "Text" and "Binary", while CLOB comes with "Text", "Binary", "JSON" and "XML".  There is really no reason to not include "JSON" and "XML" for BLOB presentation as such a column might contain Json and XML content as well.

If the answer is **no** to the questions above, could it be considered as candidates for a future release? As of now I'm probably downgrading to 7.2.0 again because my daily work very much involves having to look at the content of said column and the extra effort of using the mouse to change presentation mode becomes a huge obstacle.  Every action, that can not be solved with key combinations on the keyboard but involves a mouse, is to be avoided in the name of efficiency, in my opinion.

Thanks for the effort to develop dBeaver which is a fantastic piece of software!

kseniiaguzeeva:
Thanks for the suggestion. Additional view configuration needs to be added.

Zbambus:
Hi good people at dBeaver.
Any desicion taken, upon which release, this request might be added to the configuration of dBeaver?

[...]

PR #13735, opened 2021-08-30
#10149 Remember last used value manager in content viewer panel

ShadelessFox:


serge-rider:
1. Not sure we should use virtual entities as UI settings storage
2. It looks like this code won't work if virtual entity doesn't exist (i.e. was never used before).

PR #13736, opened 2021-08-30
#13283 Redshift geometry type handling fix

ShadelessFox:
I don't really like this approach, but it's simple. Postgres plugin must not know about its concrete extensions. 

In future, after we move Redshift into separate plugin, we can add logic for retrieving type handler provider from the data source instead of accessing it via static method.


serge-rider:
Please add comment describing empty type handler behavior.

Issue #13406, opened 2021-08-02
Cannot add comment to foreign table in Postgres


plotn:
Hi!
I have foreign table:

CREATE FOREIGN TABLE XXX (
	id_ varchar(64) NOT NULL,
	rev_ numeric(38) NOT NULL,
	resource_type_ varchar(255) NOT NULL,
	name_ varchar(255) NOT NULL,
	owner_ varchar(255) NULL,
	query_ bytea NULL,
	properties_ bytea NULL
)
SERVER YYY
OPTIONS (schema 'ZZZ', table 'KKK');

Then I try to add comment and the SQL is generated as "COMMENT ON TABLE XXX" but this doesnt work, we need "COMMENT ON foreign TABLE" instead.
When I added comment manually, the metadata still shows "COMMENT ON TABLE XXX" - it is still incorrect, but comment is displayed

uslss:
thanks for the report

LonwoLonwo:
Fixed in DBeaver 21.2.1

Matvey16:
Can't delete a comment for a foreign table

![aXUuyouxp8](https://user-images.githubusercontent.com/82543000/132187774-f0df80c0-eed9-47d8-93b3-02d77968848f.gif)

Matvey16:
Verified

PR #13737, opened 2021-08-30
#13406 change comment foreign table statement

LonwoLonwo:
2021-08-30 23_45_04-DBeaver Ultimate 21 2 0 - my_table


serge-rider:
Don't mix different type checks. Use instanceof PostgreViewBase instead of isView.

serge-rider:
Also I think we should add function getTableType() in PostgreTableBase to avoid this kind of checks all around the code.

Issue #13699, opened 2021-08-27
Check All for Database Dump Ignores Tables - Manual Check Works Fine


teamayu:
Operating system (distribution) and version: Win 10, 64 Bit
DBeaver version: 21.1.5
Database: MySQL, 5.7.29

Right Click a table in DBeaver and select 'Tools > Dump Database'. The dialog opens with the selected table highlighted. 

![27-08-2021_104822](https://user-images.githubusercontent.com/10624092/131064106-44e81f48-602d-4856-a66b-21b333b22f56.jpg)

If you select 'All' and then 'Start', it only backs up the table that was initially selected, regardless of the selections. If you check the boxes manually without using 'All', it works fine.

LonwoLonwo:
Thanks for the bug report!

Doesn't work with PostgreSQL dump too.

As a workaround - right-click on tables container and choose Tools->Dump.

![2021-08-27 10_41_52-DBeaver Ultimate 21 2 0 - _dvdrental_ Script-2](https://user-images.githubusercontent.com/45152336/131091420-cd49080c-f150-4045-8c6c-2e6bc957612b.png)

teamayu:
Great, thank you.

LonwoLonwo:
Fixed in DBeaver 21.2.1

uslss:
verified

PR #13800, opened 2021-09-05
#13699 "all" and "none" buttons event handled; localized

LonwoLonwo:
2021-09-05 19_14_53-Backup


plugins/org.jkiss.dbeaver.tasks.native.ui/src/org/jkiss/dbeaver/tasks/ui/nativetool/AbstractNativeToolWizardPage.java
```diff
@@ -76,6 +77,9 @@ protected void updateState() {
         setPageComplete(true);
     }
 
+    protected void updateTableCheckedStatus(Table table, boolean check) {
+    }
+
```
ShadelessFox: I'd rather refactor `updateState` to accept two additional parameters instead of creating a separate method.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.tasks.native.ui/src/org/jkiss/dbeaver/tasks/ui/nativetool/AbstractNativeToolWizardPage.java
LonwoLonwo: In my opinion, here refactoring will be overwhelmed. I need arguments from the method, but I do not want to throw them through the "updateState" method. It is used in 8 classes, but I'm interested in only two.

ShadelessFox: Introduction of a new method creates a mess. If it is used in 8 classes, so what? They may eventually need such functionality. Instead of overloading another method, they would use these parameters instead.

LonwoLonwo: In 8 classes and in 22 places. I continue to assume that this is too big refactoring for such a small case.

ShadelessFox: I'd refactor it further to avoid having `checkedObjects` in every inheritor. Move it to base class, add generics here and there. Don't you want to make our codebase cleaner?

At least consider adding `@NotNull` annotations to parameters in the introduced method.

ShadelessFox (commented):
You can squash these commits now.

Issue #13820, opened 2021-09-07
Redshift drivers: upgrade driver to 2.x and switch to Maven repository


serge-rider:
Also we need to be able to force optional Maven dependencies load. Redshift driver has all AWS SDK deps as optional.

uslss:
verified

deepaksenthilkumar:
@serge-rider 
@ShadelessFox @akilovich @Matvey16 
This breaks a number of functionalities within Redshift as DBeaver relies on the Postgres flavor of the driver. It should use 1.x driver not 2.x driver

ShadelessFox:
Hello again @deepaksenthilkumar,

What kind of functionality is broken?

deepaksenthilkumar:
One thing on top of my head is
>>>>Right click and get DDL.

It won’t return Postgre object. Using 2.x means we can’t take advantage of
all the Postgre extensions in the code, it has to be 1.x driver for that.
-- 
SD

ShadelessFox:
There's an issue for this problem: #13523. We should investigate and it fix.

Issue #13523, opened 2021-08-13
Redshift generete full DDL: Unsupported vector type: com.amazon.redshift.util.RedshiftObject


archenroot:
#### System information: 
- Operating system (distribution) and version: Gentoo
- DBeaver version - 
- Additional extensions

#### Connection specification:
- Database name and version: AWS Redshift
- Driver name: redshift
- Do you use tunnels or proxies (SSH, SOCKS, etc)?: ssh

#### Describe the problem you're observing:
When using generate DDL for Schema in redshift database when I click button complete DDL i get the following message:
![image](https://user-images.githubusercontent.com/553512/129329442-7c669f86-642d-41dd-9acd-4653da961c78.png)

#### Steps to reproduce, if exist:
right-click on schema and generate ddl, use complete ddl option checkbox.
NOTE: some schemas doesn't fail
#### Include any warning/errors/backtraces from the logs
[dbeaver.log](https://github.com/dbeaver/dbeaver/files/6981113/dbeaver.log)

LonwoLonwo:
Hello @archenroot 
When DBeaver generates schema DDL it uses DDL from all tables. So the problem is with specific tables. 
Can you please find this specific table for example? Its DDL will be very helpful for issue reproducing.

deepaksenthilkumar:
@LonwoLonwo -I have noticed this error earlier. It occurs when the 2.x Redshift JDBC Driver is used instead of 1.x driver as 2.x expects RedshiftObject instead of PostgreObject
@archenroot-What's the JDBC driver used?

uslss:
thanks for the report. Reproducible against 2.0.0.7 driver.

deepaksenthilkumar:
AFAIK-DBeaver doesn’t work on Redshift 2.x drivers.--
SD

deepaksenthilkumar:
I checked further on this and the issue is for tables with unique key

CREATE TABLE public.temp_unique_key (col1 varchar(10),unique(col1)) 

Then try to get the DDL for this. It won't work.

serge-rider:
Fixed. Now we support RedshiftObject values.

@kseniiaguzeeva please test it with geometry data types.

deepaksenthilkumar:
Thanks for fixing this team! It works very well with 2.x driver compared to 1.x driver.

Matvey16:
Verified

PR #13906, opened 2021-09-15
Fix for Unsupported vector type exception while using 2.x redshift

deepaksenthilkumar:
Fix for Unsupported vector type: com.amazon.redshift.util.RedshiftObject
When there is a unique constraint in a table

This is not the most efficient fix, but a good enough solution I guess.


deepaksenthilkumar:
Related to #13820 and #13523

LonwoLonwo:
Simple and effective :+1:

LonwoLonwo (approved):

deepaksenthilkumar:
Thanks. Overall I see 2.x driver work better for things like import,
insert, update etc.

Not sure if anything else has to be tested, I don’t check the geometry,
hllsketch type of data types as we don’t use it.

Might be better to test those though!
-- 
SD

serge-rider:
@deepaksenthilkumar Thank you for the bugfix and problem investigation.
I've fixed it in ticket #13523 . We have more PgObject references in PG extension and I covered all of them now. Geometry and other types should work ok now.

PR #13930, opened 2021-09-17
Use returned value

thomasleplus:
The method replace() returns the expected results instead of modifying the current string.


ShadelessFox (approved):
Nice catch, thanks for the contribution!

flambchap (approved):

PR #13983, opened 2021-09-22
Update OracleCLOBValueHandler.java

da-z:
2000 is safer in case there are multibyte characters. interpolates just twice as many ')||TO_CLOB(' in worst case scenario.


serge-rider:
Good point 👍

Issue #8728, opened 2020-05-18
[VERTICA] "Smart autocomplete" doesnt work on Vertica when using ${variables}


ashdam:
#### System information: 
- Windows 10 64bit
- 7.0.4

#### Connection specification:
- Vertica 9.1
- JDBC 9.1

#### Describe the problem you're observing:

"Smart autocomplete" doesnt work on vertica when using ${variables} with values. I've tested this feature on Oracle and it perfectly works.

It might be related: same happens when you want to inspect a table on ${schema}.table using CTRL+click on "table"

#### Steps to reproduce, if exist:

1) Execute:
select *
from ${schema}.table
2) Try to use CTRL+SPACE to autocomple when writing your "table"

kseniiaguzeeva:
Thank you for the report.

LonwoLonwo:
Fixed for Vertica and for Oracle.

![2021-09-30 11_11_58-DBeaver Ultimate 21 2 2 - _Partner11DB_Script-66](https://user-images.githubusercontent.com/45152336/135725297-e3cd3551-3643-4e79-8091-ad2849d61031.png)

![2021-09-30 11_12_44-](https://user-images.githubusercontent.com/45152336/135725299-5cc09b25-6438-4cff-8e0d-66d79b479eee.png)
e.

Matvey16:
Verified

PR #14065, opened 2021-09-30
#8728 filter tokens list to avoid empty tokens

LonwoLonwo:
2021-09-30 11_11_58-DBeaver Ultimate 21 2 2 - _Partner11DB_Script-66

### DIFF
--- a/bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
+++ b/bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
@@ -21,10 +21,7 @@ import org.jkiss.code.NotNull;
 import org.jkiss.code.Nullable;
 
 import java.lang.reflect.Array;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.List;
+import java.util.*;
 
 /**
  * Common utils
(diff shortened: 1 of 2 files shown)


bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
```diff
@@ -259,4 +256,14 @@ public static void main(String[] args) {
...
+
+    public static String[] getFilteredStringArray(String[] array) {
+        List<String> filteredList = new ArrayList<>();
+        for (String token : array) {
+            if (CommonUtils.isNotEmpty(token)) {
+                filteredList.add(token);
+            }
+        }
+        return filteredList.toArray(new String[0]);
+    }
```
ShadelessFox, line 268: Let's refactor this function to be useful in other cases:

```java
    @NotNull
    public static <T> T[] filter(T[] array, Predicate<T> predicate) {
        List<String> filteredList = new ArrayList<>();
        for (String item : array) {
            if (predicate.test(token)) {
                filteredList.add(token);
            }
        }
        return filteredList.toArray(new String[0]);
    }
```

Also, I would suggest pre-allocating `filteredList` to at most the size of input `array` and then making a required slice of it.

The problem is that by generalizing it, it becomes just a dummy way to write:

```java
Arrays.stream(...).filter(...).toArray(T[]::new)
```

ShadelessFox (changes requested):

bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
LonwoLonwo: Maybe streams are in a general better solution.

PR #14110, opened 2021-10-02
dbeaver/dbeaver-ee#899 add universal code to handle tree with checkboxes

flambchap:


serge-rider:
It is not clear fixTree does. Requirement to call it manually for every tree looks poor (you could do this in constructor).

flambchap:
I believe I've found a better approach.

PR #14115, opened 2021-10-04
code cleanup

LonwoLonwo:


serge-rider:
You cannot remove confirm_* constants. They are used indirectly.

serge-rider:
As well as actions_ContentAssistProposal_description

Issue #14138, opened 2021-10-05
Configurators i18n


serge-rider:
We still have several metaobjects configurators which use hardcoded strings instead of E i18n.

LonwoLonwo:
These objects localized to English and Russian:
- DB2 create schema dialog
- Exasol Main Connection Page (password, URL, etc.)
- Exasol create Role Dialog
- Mysql create a database dialog
- Oracle create schema dialog
- Oracle sessions - Show background/Show inactive actions with tooltips

uslss:
verified

PR #14139, opened 2021-10-05
#14138 OracleSchemaConfigurator localization

IlyaCh1zhov:
image


ShadelessFox (approved):

plugins/org.jkiss.dbeaver.ext.oracle.ui/src/org/jkiss/dbeaver/ext/oracle/ui/internal/OracleUIMessages.java
```diff
@@ -139,4 +139,9 @@ private OracleUIMessages() {
     public static String dialog_schema_edit_user_name;
     public static String dialog_schema_edit_user_password;
     public static String dialog_schema_edit_label;
+
+    public static String views_session_manager_viewer_show_background;
+    public static String views_session_manager_viewer_show_background_tasks_tool_tip_text;
```
ShadelessFox: `_tool_tip_text` is a bit verbose prefix. We often use just `_tip`. Same for other occurrences

ShadelessFox (approved):

Issue #13481, opened 2021-08-09
Navigating references lists relationships multiple times


simmermonr:
#### System information: 
- Mac OS Big Sur - V 11.4
- DBeaver Community Version 21.1.4.202108020335

#### Connection specification:
- PostgreSQL 10.10 (Debian 10.10-1.pgdg90+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516, 64-bit (On Docker image)
- Driver name: PostgreSQL JDBC Driver
- Installed on Docker Image clkao/postgres-plv8:10-2

#### Describe the problem you're observing:
When pressing Ctrl + 1 on results record, or going to Navigate --> References, the "child" table references get listed multiple times for the same FK.

![image](https://user-images.githubusercontent.com/82680848/128754280-60c4b799-a0da-4c24-8bd9-361f20918a62.png)

Matvey16:
@simmermonr Thank you for the report

contibru:
Occurs in the References tab to

![image](https://user-images.githubusercontent.com/30509284/133099865-9364c691-b734-49ca-86d0-fca98292206d.png)

LonwoLonwo:
Fixed in DBeaver 21.2.3

Matvey16:
Verified

PR #14157, opened 2021-10-06
#13481 avoid schemas names duplicates for reference query

LonwoLonwo:
Was:
2021-10-06 17_33_37-DBeaver Ultimate 21 2 2 - customer

Now:
2021-10-06 17_33_10-DBeaver Ultimate 21 2 2 - customer


ShadelessFox (approved):
Nice catch 🙃

PR #14159, opened 2021-10-06
Fix for SQL Server to avoid extra "use database;"

okeuday:
* The connection property "databaseName" is used to set the database.
  This change will allow relying on the database already set due to
  the connection property so an extra "use database;" request can be
  avoided.


okeuday:
Thanks!

Issue #13745, opened 2021-09-01
Hana: create procedure fails with syntax error at END if there is a loop inside.


wernerdaehn:
#### System information: 
- Win10
- DBeaver 21.2.0

#### Connection specification:
Sap Hana

#### Describe the problem you're observing:
Simple procedure cannot be created due to supposed syntax errors.

#### Steps to reproduce, if exist:
Run the SQL Script:

```
create or replace procedure abc
      (in x SMALLINT,
       in y SMALLINT)      
as 
begin
    declare loop_x SMALLINT;
    declare loop_y SMALLINT;
    
    declare maxID BIGINT;
    
    for loop_y in 1..:y do
        for loop_x in 1..:x do
            maxID := maxID + 2;
        end for;
    end for;    
end;
```

It seems the parser sends only the string
```
create or replace procedure abc
      (in x SMALLINT,
       in y SMALLINT)      
as 
begin
    declare loop_x SMALLINT;
    declare loop_y SMALLINT;
    
    declare maxID BIGINT;
    
    for loop_y in 1..:y do
        for loop_x in 1..:x do
            maxID := maxID + 2;
        end for
```
truncating the statement after the end-for's semicolon char.

Error message is
```
SQL Error [257] [HY000]: SAP DBTech JDBC: [257] (at 302): sql syntax error: line 14 col 13 (at pos 302)
```

and the create statement is obviously valid, it works in Hana Studio.

LonwoLonwo:
Hello @wernerdaehn 

Thanks for the bug report!

You can try to highlight your procedure statement in SQL Editor and use Ctrl+Enter to avoid block parsing problems.

Now we, unfortunately, cannot test this case, because the connection to the database you gave us is no longer working. If you have the opportunity to back our access to the database - it would be very cool!

wernerdaehn:
I have shutdown the server but can start it at any time. It is just that I do no longer need it 24x7. Please send me an email when and for how long I should start the server and I will do so.

LonwoLonwo:
Thanks for your help. Your database was very useful for us.

Anyway. I see a blank line in your statement.
Can you please tell me - the "Blank line is statement delimiter" is enabled or disabled in your preferences?

![2021-09-01 18_13_04-DBeaver Ultimate 21 2 0 - _orcl_ Script-8](https://user-images.githubusercontent.com/45152336/131697461-af599285-0a56-428b-b6f6-5d40331425f2.png)

[...]

PR #14254, opened 2021-10-14
support HANA while loop

kai-morich:
same as #13745 but other control structure


LonwoLonwo:
Hello @kai-morich 
Query examples will be very helpful.
Also, why you changed "{"IF", SQLConstants.BLOCK_END + " IF"}" string? It is always "END IF" at the end of the block? Or it can be just "END"?

kai-morich:
Fails without final IF. Always has to be END IF as documented [here](https://help.sap.com/viewer/de2486ee947e43e684d39702027f8a94/2.0.05/en-US/46e875a0d4e446ae9ab08334f13cd1aa.html). 

example:
```
do BEGIN
	DECLARE i integer;
	DECLARE j integer = 1;
	DECLARE k integer = 1;
	FOR i IN 0..10 do
		j = j*2;
	END FOR;
	while k < 10 do
		k = k*2;
	END while;
	IF 1=1 THEN
		i = 42;
	END if;
	SELECT i i, j j, k k FROM dummy;
END;
```
returns
```
I |J   |K |
--+----+--+
42|2048|16|
```

BTW: DBeaver shows a small (-) in front of the collapsible IF block
grafik
Doesnt work for any more for `END ...` blocks in my patched version:
grafik

LonwoLonwo (approved):

PR #14307, opened 2021-10-20
dbeaver/dbeaver-ee#1105 set Athena to read-only meta data mode

LonwoLonwo:
and CSV driver too

2021-10-20 11_51_57-DBeaver Ultimate 21 2 3 - testtable1

2021-10-20 11_53_20-DBeaver Ultimate 21 2 3 - cloudfront_logs

2021-10-21 01_17_41-DBeaver Ultimate 21 2 3 - _postgres_ Script-10

### DIFF
--- a/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/EditConnectionNavigatorSettingsDialog.java
+++ b/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/EditConnectionNavigatorSettingsDialog.java
@@ -92,10 +92,12 @@ public class EditConnectionNavigatorSettingsDialog extends BaseDialog {
                 1);
 
 
-            boolean mergeEntitiesEnabled = false;
+            boolean mergeEntitiesEnabled;
             if (dataSourceDescriptor != null) {
                 mergeEntitiesEnabled = dataSourceDescriptor.getDriver().getProviderDescriptor().getTreeDescriptor().supportsEntityMerge();
                 mergeEntities.setEnabled(mergeEntitiesEnabled);
+            } else {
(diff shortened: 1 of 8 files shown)


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/EditConnectionNavigatorSettingsDialog.java
```diff
@@ -89,17 +91,44 @@ protected Composite createDialogArea(Composite parent) {
                 navigatorSettings.isMergeEntities(),
                 1);
 
+
+            boolean mergeEntitiesEnabled = false;
```
ShadelessFox: I suggest adding an `else` branch and initializing this variable in it, so it can be marked as `final`.

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/EditConnectionNavigatorSettingsDialog.java
```diff
@@ -89,17 +91,44 @@ protected Composite createDialogArea(Composite parent) {
...
+            });
+
             hideFolders = UIUtils.createCheckbox(
                 miscGroup,
                 CoreMessages.dialog_connection_wizard_final_checkbox_hide_folders,
                 CoreMessages.dialog_connection_wizard_final_checkbox_hide_folders_tip,
                 navigatorSettings.isHideFolders(),
                 1);
+
+            boolean finalMergeEntitiesEnabled = mergeEntitiesEnabled;
```
ShadelessFox: See previous note

ShadelessFox (changes requested):

Issue #13012, opened 2021-06-28
Database Navigator filter always resetting


CounterClops:
#### System information: 
- Windows 10 21H1
- DBeaver 21.1.1.202106210824

#### Describe the problem you're observing:
In the Database navigator window, changing the filter away from **"Filter tables, procedures, etc."** to one of the other 2 options isn't remembered when restarting DBeaver.

This is annoying when we want the Database navigator search to only ever filter by connection. We are expecting this option to be remembered between restarts.

#### Steps to reproduce, if exist:

1. Press the arrow next to the Filter button in the Database Navigator view
2. Select a different option from default (**"Filter connections"** is the option we use)
3. Completely close the DBeaver application
4. Open DBeaver and check the filter, and it'll now be set back to **"Filter tables, procedures, etc."**

You can see the below options for reference.
![image](https://user-images.githubusercontent.com/19741838/123596830-39338780-d825-11eb-9757-1b9559d05e20.png)

uslss:
thank you for the report

ShadelessFox:
Implemented:

![java_u4cKAyDmjl](https://user-images.githubusercontent.com/35821147/138711775-1da39df6-51d6-41fe-a7cb-bd352d536285.png)

uslss:
verified

PR #14382, opened 2021-10-25
#13012 Save/restore database navigator filter + code refactor

ShadelessFox:
java_u4cKAyDmjl


serge-rider:
Let's name it "Save quick filter value"

PR #14404, opened 2021-10-26
Incorrect Clickhouse table statistics, it includes inactive parts

den-crane:
It needs a filter active to account only active parts otherwise it shows inflated numbers.


serge-rider:
Thanks for the fix 👍

Matvey16:
Verified

Issue #12131, opened 2021-04-17
SQL Server - can't update views (which I can update using other tools)


aisbergde:
#### System information: 
- Operating system (distribution) and version Windows 10
- DBeaver version
- Additional extensions

#### Connection specification:
- Database name and version: SQL Server 2019
- Driver name
- Do you use tunnels or proxies (SSH, SOCKS, etc)? no 

#### Describe the problem you're observing:

Sometimes I use views to update content in tables, for example if I want to add some lookup content.

I can update these views in SSMS or in Azure Data Studio, but I can't update them in dbeaver.

First I am asked to define a virtual PK to identify a unique row.
But even with a virtual key it is not possible to update views. I get this kind of errors: 
![image](https://user-images.githubusercontent.com/9020187/115119482-d1e99380-9fa8-11eb-80d2-945bf332472d.png)

![image](https://user-images.githubusercontent.com/9020187/115119686-ec703c80-9fa9-11eb-91f6-4135f69167a9.png)

The error message is wrong, the object exists. But it looks like dbeaver want to update only tables and thinks the view doesn't exist. But I edited content in this view, it exists.

This is only one example. I was never able to update views.

#### Steps to reproduce, if exist:

code of the tables and of the view:

```sql
CREATE TABLE [repo].[Index_virtual](
	[index_guid] [uniqueidentifier] NOT NULL,
	[has_managedName] [bit] NOT NULL,
	[index_name] [nvarchar](128) NULL,
	[index_type] [tinyint] NOT NULL,
	[is_index_disabled] [bit] NOT NULL,
	[is_index_primary_key] [bit] NOT NULL,
	[is_index_unique] [bit] NOT NULL,
	[parent_RepoObject_guid] [uniqueidentifier] NOT NULL,
	[referenced_index_guid] [uniqueidentifier] NULL,
	[RowNumberInReferencing] [int] NULL,
 CONSTRAINT [PK_Index_virtual] PRIMARY KEY CLUSTERED 
(
	[index_guid] ASC
)WITH (PAD_INDEX = OFF, STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, ALLOW_ROW_LOCKS = ON, ALLOW_PAGE_LOCKS = ON, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
) ON [PRIMARY]
GO

Create Table [repo].[Index_Settings](
[...]

PR #14426, opened 2021-10-28
#12131 support of sql_variant type binding

LonwoLonwo:
2021-10-28 00_50_28-Event


serge-rider:
Its better to create new SQLServerVariantValueHandler than reuse SQLServerMoneyValueHandler

LonwoLonwo:
hmm

PR #14472, opened 2021-11-02
Permissons pg#13993

LonwoLonwo:
2021-11-02 15_09_05-DBeaver Ultimate 21 2 4 - category

2021-11-02 15_14_51-json_test_json - Persist Changes


mkgrgis:
There was an error with non-quoted roles in Pg. I have searched for this problem. With new source code there is no error. Thanks.

LonwoLonwo:
@mkgrgis 
Yes. This broke recently because of the fixes of another problem :) how it always happens.

Issue #13596, opened 2021-08-20
Oceanbase plugin development


serge-rider:
Most job was done by  @LuckyLeoZZ in https://github.com/dbeaver/dbeaver/pull/13449

serge-rider:
@LuckyLeoZZ new plugin merged in devel. It is ready to be inbcluded in release cycle. However there are several issues with it:
- No driver jars specified. It is not clear where to get them. Perfect solution would be to specify some Maven artifact in driver config. Or some https download link at least. We usually do not add new database extensions without drivers because for end users it is hard to use them.
- Please add icons for high res monitors. I mean ob_icon and ob_icon_big. Just add ob_icon@2x.png and ob_icon_big@2x.png with resolution 32x32 and 128x128
- On the connection page pre-fill default database port, host (localhost?) and database name

LuckyPickleZZ:
@serge-rider Thank you very much. For these questions you said, I 'll make improvements in the near future.
And I would like to tell you that I maybe request for a PR for OB oracle mode. Sorry for increasing the workload of review for you.

LuckyLeoZZ:
@serge-rider Thank you very much. For these questions you said, I 'll make improvements in the near future.
And I would like to tell you that I maybe request for a PR for OB oracle mode. Sorry for increasing the workload of review for you.

LuckyPickleZZ:
@serge-rider Hello,  our team discussed it and decided to give up the PR of Oracle mode for the time being. Could you please add our plugin to the next release and let our users use it first?

LuckyPickleZZ:
@tati-kru Hello, our team discussed it and decided to give up the PR of Oracle mode for the time being. Could you please add our plugin to the next release and let our users use it first?

tati-kru:
@LuckyLeoZZ we cannot include your plugin to the release without the fixes mentioned above. Please, fix them and we can move forward.

LuckyPickleZZ:
> @LuckyLeoZZ we cannot include your plugin to the release without the fixes mentioned above. Please, fix them and we can move forward.

Thanks al lot. I see what you mean. We'll fix these soon.

LuckyPickleZZ:
> @LuckyLeoZZ we cannot include your plugin to the release without the fixes mentioned above. Please, fix them and we can move forward.

https://github.com/dbeaver/dbeaver/pull/14607
Please check this. I've made some impovements.

serge-rider:
[...]

PR #14607, opened 2021-11-18
Fix problems of OceanBase plugin.

LuckyPickleZZ:
https://github.com/dbeaver/dbeaver/issues/13596


serge-rider:
Let's try it in the next release.

BTW is there any cloud service where we can test the driver?

PR #14610, opened 2021-11-18
#14950 return ability to create indexes via UI for SQLite

LonwoLonwo:
2021-11-18 15_49_32-DBeaver Ultimate 21 2 5 - Album


LonwoLonwo:
#14590

Issue #14556, opened 2021-11-15
MariaDB: NPE when creating restore task


DariaMarkaryan:
#### System information: 
- Win10
- DBeaver CE 21.2.5

#### Connection specification:
- MariaDB (Driver version 2.7.1)

#### Describe the problem you're observing:
Null Pointer Exception is thrown when creating restore db task

#### Include any warning/errors/backtraces from the logs
![image](https://user-images.githubusercontent.com/43252954/141743628-4033f6b9-ec76-4431-93b6-0a20313d8bc3.png)

.log:
```
java.lang.NullPointerException
	at org.jkiss.dbeaver.ext.mysql.tools.MySQLWizardPageSettings.createSecurityGroup(MySQLWizardPageSettings.java:49)
	at org.jkiss.dbeaver.ext.mysql.tools.MySQLScriptExecuteWizardPageSettings.createSecurityGroup(MySQLScriptExecuteWizardPageSettings.java:1)
	at org.jkiss.dbeaver.ext.mysql.tools.MySQLScriptExecuteWizardPageSettings.createControl(MySQLScriptExecuteWizardPageSettings.java:121)
	at org.jkiss.dbeaver.ui.dialogs.MultiPageWizardDialog.changePage(MultiPageWizardDialog.java:278)
	at org.jkiss.dbeaver.ui.dialogs.MultiPageWizardDialog.showPage(MultiPageWizardDialog.java:381)
	at org.jkiss.dbeaver.tasks.ui.wizard.TaskConfigurationWizardDialog.buttonPressed(TaskConfigurationWizardDialog.java:216)
	at org.eclipse.jface.dialogs.Dialog.lambda$0(Dialog.java:619)
	at org.eclipse.swt.events.SelectionListener$1.widgetSelected(SelectionListener.java:84)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:252)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4237)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1060)
	at org.eclipse.swt.widgets.Display.runDeferredEvents(Display.java:4054)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3626)
	at org.eclipse.jface.window.Window.runEventLoop(Window.java:823)
[...]

PR #14630, opened 2021-11-22
#14556 fix restore task exception

Destrolaric:


Destrolaric:
Bug is related to calling wizard.initializeWizard() only on dialogue creation. It works for direct restore through tools menu. But breaks on restoring task. Fix is to add initialization for newly added wizards

flambchap (approved):

flambchap:
Sounds reasonable

ShadelessFox (approved):

ShadelessFox:
Seems to be working.

Issue #14599, opened 2021-11-18
NPE appears in case of typing in sql editor for not reconnected connection


uslss:
#### System information: 
- Operating system (distribution) and version win10
- DBeaver version 21.2.5

#### Connection specification:
- Database name and version postgresql
- Driver name
- Do you use tunnels or proxies (SSH, SOCKS, etc)? 

#### Steps to reproduce, if exist:
1. Connect to database and wait until it disconnects
2. Open SQL editor for mentioned connection (connection is not invalidated, N/A is shown instead of schema)
3. Type some symbols in SQL editor

NPE appears for every symbol in error log
![7uoSl2hl8Q](https://user-images.githubusercontent.com/31996417/142393071-5f00b713-08b1-426e-bc1b-9c4cf6f67aa2.gif)

.
#### Include any warning/errors/backtraces from the logs
```
eclipse.buildId=unknown
java.version=11.0.12
java.vendor=Eclipse Foundation
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -data file:/C:/Users/Dziyana/AppData/Roaming/DBeaverData/workspace6/

org.jkiss.dbeaver.model
Error
Thu Nov 18 12:47:11 MSK 2021


java.lang.NullPointerException
	at org.jkiss.dbeaver.model.DBUtils.getSelectedObject(DBUtils.java:1707)
	at org.jkiss.dbeaver.model.DBUtils.getActiveInstanceObject(DBUtils.java:1701)
	at org.jkiss.dbeaver.model.sql.completion.SQLCompletionAnalyzer.runAnalyzer(SQLCompletionAnalyzer.java:248)
	at org.jkiss.dbeaver.model.sql.completion.SQLCompletionAnalyzer.runAnalyzer(SQLCompletionAnalyzer.java:98)
	at org.jkiss.dbeaver.model.sql.completion.SQLCompletionAnalyzer.run(SQLCompletionAnalyzer.java:82)
	at org.jkiss.dbeaver.model.sql.completion.SQLCompletionAnalyzer.run(SQLCompletionAnalyzer.java:1)
	at org.jkiss.dbeaver.model.exec.DBExecUtils.tryExecuteRecover(DBExecUtils.java:171)
[...]

PR #14649, opened 2021-11-23
#14599 fix NullPointerException when typing while reconnecting

Destrolaric:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -1702,17 +1702,19 @@ public static DBSObject getActiveInstanceObject(@NotNull DBCExecutionContext exe
...
-        DBCExecutionContextDefaults contextDefaults = context.getContextDefaults();
-        if (contextDefaults != null) {
-            DBSSchema defaultSchema = contextDefaults.getDefaultSchema();
-            if (defaultSchema != null) {
-                return defaultSchema;
-            }
-            DBSCatalog defaultCatalog = contextDefaults.getDefaultCatalog();
-            if (defaultCatalog != null) {
-                return defaultCatalog;
+    public static DBSObject getSelectedObject(DBCExecutionContext context)
```
ShadelessFox: Where did the annotation go? If it is nullable indeed, wouldn't it be better to locate the source of `null` instead of fixing the receiver?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -1702,17 +1702,19 @@ public static DBSObject getActiveInstanceObject(@NotNull DBCExecutionContext exe
...
+    {
```
ShadelessFox: Please fix formatting.

ShadelessFox (changes requested):

LonwoLonwo (approved):

Issue #14477, opened 2021-11-02
Show time SQL completed?


rdp:
Sometimes when I turn to DBeaver in the morning, connections will have been lossed.
Then when I execute a query, it "pauses" for awhile, then reconnects, then issues the query.  In the meantime I've gone off to do something else instead of staring intently at the DBeaver app to see "when it will come back."
When I get back to DBeaver, it's not clear if the query results were from "yesterday" or from "today"

What I'd like is for the "query status" short bar to say "7 row(s) fetched - 116ms, 1 min ago" or "7 row(s) fetch - Nov 2 10:50 AM"

**Describe alternatives you've considered**
I can rerun the query, but it's confusing.

Thanks!

Matvey16:
@rdp Thank you for the suggestion

ShadelessFox:
Implemented:

![image](https://user-images.githubusercontent.com/35821147/149334519-0de16110-1148-4247-abb7-ee760ec658d4.png)

rdp:
Thanks!

Matvey16:
Verified

Matvey16:
Verified

PR #14684, opened 2021-11-25
proposed solution to #14477

mlazatinph:
This is proposed solution to #14477


ShadelessFox:
Hello @mlazatinph,

Thanks for the contribution.

Unfortunately, it's not a general solution. Also, the output is not human-friendly.

Issue #14669, opened 2021-11-24
Babelfish issues


uslss:
#### System information: 
- Operating system (distribution) and version win10
- DBeaver version 21.3.0

#### Connection specification:
- Database name and version Babelfish

#### Describe the problem you're observing:
Case 1:
Objects created in non-default database aren't visible after refresh
**STR**
1. Connect to database (database in preferences = master)
2. Create table in another non-default db (e.g. demo2)
3. Save table and refresh node
Actual result:
Created object aren't visible.

Case 2:
Synonyms, Sequences, Triggers and Database Triggers folders cannot by open due to error
#14808
Case 3:
Objects are always created in the default database
![image](https://user-images.githubusercontent.com/31996417/143225984-6b0c7395-db37-45cc-b3ce-59fb2bb87282.png)

Case 4:
Data import from csv isn't possible
![image](https://user-images.githubusercontent.com/31996417/143229678-bcdebc4b-b5cf-45b6-b16a-0acb69bf3e00.png)

Destrolaric:
For Case 2 new issue is opened #14808

E1izabeth:
To avoid switching between databases with `use dbname;` before and after every user action, only default database is shown for now.

uslss:
verified

PR #14709, opened 2021-11-26
#14669 Add missing Babelfish queries

okeuday:
@serge-rider Would it be possible to put this into 21.3.0 ?


serge-rider:
Unfortunately no, we are in the code freeze stage now.
New changes will be available in the 21.3.1 and in the Early Access version.

It also may be included in DBeaver EE 21.3.0.

Matvey16:
@okeuday I can't see sequences in Navigator or schema properties
dodHTssXLq
TZeEebidic

okeuday:
@Matvey16 That is normal due to Babelfish currently lacking `sys.sequences`.  The [query used instead](https://github.com/dbeaver/dbeaver/blob/425cf111277372253a6cc6d98b01a99e7e1f986d/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/SQLServerUtils.java#L179) uses `TOP 0` to provide the column types without rows.  When `sys.sequences` is added on the server-side, the client-side query can be removed.

Matvey16:
Verified

Issue #14474, opened 2021-11-02
Checkboxes for permissions disappear in table properties


Matvey16:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 21.2.4

#### Connection specification:
- Database name and version PostgreSQL

#### Describe the problem you're observing:
When checking permissions for different roles all checkmarks except select disappear when switching back and forth between the roles. Generated script is still correct.

#### Steps to reproduce, if exist:
1. Open permission tab in table properties
2. Check some of the permissions for one role
3. Click on another role and check some permissions
4. Click back on the first role

**Expected:** Checkmarks for checked permissions are present
**Actual:** All checkmarks except select disappear

![hHJaxBUHaE](https://user-images.githubusercontent.com/82543000/139867490-a99e337b-2fd3-4f35-baae-61e823262258.gif)

Lkledu:
i guess this could be a good first issue that i'll try to resolve

Lkledu:
I finished to resolve it.

PR #14713, opened 2021-11-28
#14474 Fix check component for privilege roles of postgresql

Lkledu:
This is proposed solution to #14474 
the issue was added to the 21.3.5 milestone

- Changed the `setPermission` method from `PostgrePrivilege` so it not overwrite the permissions anymore.
- Changed the `updateCurrentPrivileges` method from `PostgresRolePrivilegesEditor` so the `if(permission == null)` doesn't skip the addition of new permissions before user add the first permission.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/PostgresRolePrivilegesEditor.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/PostgresRolePrivilegesEditor.java
@@ -309,7 +309,7 @@ public class PostgresRolePrivilegesEditor extends AbstractDatabaseObjectEditor<P
                                 currentObject.getName(),
                                 Collections.singletonList(privGrant));
                     } else {
-                    	permission.setPermission(privilegeType, grant);
+                    	permission.setPermission(privilegeType, grant, currentUser);
                     }
                     
                 }
--- a/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/PostgresRolePrivilegesEditor.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/PostgresRolePrivilegesEditor.java
@@ -254,7 +254,18 @@ public class PostgresRolePrivilegesEditor extends AbstractDatabaseObjectEditor<P
                     // No permission - nothing to revoke
                     continue;
                 }
-                if (isRoleEditor()) {
+                
+            } else if (privilegeType != null) {
+                // Check for privilege was already granted for this object
+                boolean hasPriv = permission.getPermission(privilegeType) != PostgrePrivilege.NONE;
+                if (grant != hasPriv && !grant) {
+                    permissionMap.remove(permission.getName());
+                }
+            }
+            
+            
+            {
+            	if (isRoleEditor()) {
                     PostgrePrivilegeOwner permissionsOwner = (PostgrePrivilegeOwner) currentObject;
                     PostgrePrivilegeGrant.Kind kind;
                     String objectName;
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgrePrivilege.java
```diff
@@ -163,9 +163,16 @@ public short getPermission(PostgrePrivilegeType privilegeType) {
...
+            if (permission.privilegeType != privilegeType) {
                 if (permit) {
                     permission.permissions |= GRANTED;
+                    ObjectPermission[] tempPermission = new ObjectPermission[this.permissions.length+1];
+                    for(int i = 0; i < this.permissions.length; i++) {
+                	tempPermission[i] = this.permissions[i];
+                    }
+                    tempPermission[this.permissions.length] = new ObjectPermission(privilegeType, permission.getGrantor(), permission.permissions);
+                    this.permissions = tempPermission;
+
```
ShadelessFox: Can be replaced with `org.jkiss.utils.ArrayUtils#add`.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/PostgresRolePrivilegesEditor.java
```diff
@@ -413,7 +423,7 @@ private void updateObjectPermissions(List<DBSObject> objects) {
             for (TableItem item : permissionTable.getItems()) {
                 PostgrePrivilegeType privType = (PostgrePrivilegeType) item.getData();
                 short perm = currentPermissions[0] == null ? PostgrePrivilege.NONE : currentPermissions[0].getPermission(privType);
-                item.setChecked((perm & PostgrePrivilege.GRANTED) != 0);
+                item.setChecked((perm & PostgrePrivilege.GRANTED) != 0 );
```
ShadelessFox: Can be replaced with `org.jkiss.utils.CommonUtils#isBitSet`.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgrePrivilege.java
```diff
@@ -163,9 +163,16 @@ public short getPermission(PostgrePrivilegeType privilegeType) {
 
     public void setPermission(PostgrePrivilegeType privilegeType, boolean permit) {
         for (ObjectPermission permission : permissions) {
-            if (permission.privilegeType == privilegeType) {
+            if (permission.privilegeType != privilegeType) {
```
ShadelessFox, line 180: Is this change intended? This will bloat `permissions` with multiple `privilegeType`-s.

Also it updates `permissions` array while iterating through it which is bad.

ShadelessFox (changes requested):
Thanks for the contribution!

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgrePrivilege.java
Lkledu: `org.jkiss.utils.ArrayUtils#add` just add int arrays, i've a ObjectPermission array

Lkledu: i change the for statement for `ArrayUtils.concatArrays()`

ShadelessFox: There's generic overload; usage: `ArrayUtils.add(ObjectPermission.class, permissions, new ObjectPermission(...))`.

Please see my other comment below.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgrePrivilege.java
Lkledu: it's intended. Before this implementation the permission array was overwritten, so alway when the userr change role and goes back, dbeaver would only check the last privilege checkbox, but let all the previous grant privilege statements on the permissionMap (so when user clicks on save button it shows all grant statement queries correct)

Lkledu: i had wrote a message before asking your opinion about using another object to store permissions list just for use with check UI instead to change the current permission object, but i don't know why the message get missed.

Issue #1651, opened 2017-05-10
Accessing SQL Server Jobs 


jrich8573:
I am using Eclipse Neon, with the dbeaver plugin, I am looking to access SSMS SQL job agent. I do not see a link or directory within the database navigator.  Is there a known method for access the SQL job agent in Eclipse using the dbeaver plugin?  

Thank you

islamanwar:
Is this available in dbeaver?

MarkFergusonTN:
Is there a way to "upvote" this feature?

LonwoLonwo:
Hello @MarkFergusonTN 

I've already added this ticket in 21.3.1 Milestone, which is dedicated to the metadata. It means, that our team takes this task to work.

ShadelessFox:
Implemented as a part of the Enterprise Edition.

uslss:
verified

MarkFergusonTN:
Just to make sure I understand, the ability to see/work with SQL Server Agent jobs is now part of the Enterprise edition of DBeaver?

ShadelessFox:
@MarkFergusonTN yes.

LonwoLonwo:
@MarkFergusonTN and part of DBeaver Lite Edition too.

MarkFergusonTN:
Great! Thank you so much!

MarkFergusonTN:
Is there any documentation on how to do this in the Lite Edition yet?

PR #14740, opened 2021-11-30
#1651 SQL Server jobs support

ShadelessFox:


plugins/org.jkiss.dbeaver.ext.mssql.ui/plugin.xml
```diff
@@ -56,6 +56,7 @@
             <objectType name="org.jkiss.dbeaver.ext.mssql.model.SQLServerDatabaseTrigger"/>
             <objectType name="org.jkiss.dbeaver.ext.mssql.model.SQLServerDatabaseTrigger"/>
```
LonwoLonwo: Hmm... String 57 looks precisely like string 56)

LonwoLonwo:
Why does no one loves pictures :(

plugins/org.jkiss.dbeaver.ext.mssql.ui/plugin.xml
ShadelessFox: Not related to my PR to be quite honest.

PR #14746, opened 2021-11-30
HANA procedures/functions with table-type parameters

kai-morich:
here some test statements:

```sql
CREATE SCHEMA s1;

DROP TABLE s1.tab1;
CREATE TABLE s1.tab1(i int, s varchar(5));

DROP TYPE s1.typ1;
CREATE TYPE s1.typ1 AS TABLE (i int, s varchar(5));

DROP VIEW s1.view1;
CREATE VIEW s1.view1 AS SELECT * FROM s1.tab1;


CREATE OR replace PROCEDURE s1.proc1(a nchar, b nvarchar(1), c tinyint, d bigint, e varbinary(1), f BINARY, g char, 
					 h decimal, i decimal(8,2), j smalldecimal, k integer, l SMALLINT, m REAL, n double,
					 o varchar(1), p boolean, q date, r time, s seconddate, t timestamp, 
					 u st_geometry, v blob, w nclob) AS 
BEGIN
END;

CREATE OR replace PROCEDURE s1.proc2(i int, OUT j int, INOUT k int, l int DEFAULT 42) AS 
BEGIN
END;

CREATE OR replace PROCEDURE s1.proc3(i int, j s1.tab1, k s1.typ1, l s1.view1, m TABLE(i int), n TABLE(...)) AS 
BEGIN
END;

CREATE OR replace PROCEDURE s1.proc4(t TABLE(i int, s varchar(50), d decimal(8,2), b boolean, c clob)) AS 
BEGIN
END;

CREATE OR replace FUNCTION s1.func1(i int, j s1.tab1, k s1.typ1, l s1.view1, m TABLE(i int, s varchar(5))) RETURNS o int AS
BEGIN
	o:=42; 
END;

CREATE OR replace FUNCTION s1.func2(i int, j sys.dummy, l int DEFAULT 42) RETURNS TABLE(dummy varchar(1)) AS
BEGIN
	RETURN SELECT * FROM :j; 
END;


do BEGIN 
	DECLARE j int;
	DECLARE k int;
	CALL s1.proc2(1, j, k);
END;

CALL s1.proc3(1, s1.tab1, s1.tab1, s1.tab1, s1.tab1, s1.tab1);

SELECT s1.func1(1, s1.tab1, s1.tab1, s1.tab1, s1.tab1) FROM dummy;

SELECT * FROM s1.func2(1, sys.dummy);


SELECT * FROM sys.PROCEDURE_PARAMETERS WHERE SCHEMA_NAME='S1';
SELECT * FROM sys.PROCEDURE_PARAMETER_COLUMNS WHERE SCHEMA_NAME='S1';
SELECT * FROM sys.function_PARAMETERS WHERE SCHEMA_NAME='S1';
SELECT * FROM sys.FUNCTION_PARAMETER_COLUMNS WHERE SCHEMA_NAME='S1';
```


plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANAProcedure.java
```diff
@@ -17,24 +17,169 @@
...
+                try (JDBCResultSet dbResult = dbStat.executeQuery()) {
+                    while (dbResult.next()) {
+                        String columnName = dbResult.getString(1);
+                        String typeName = dbResult.getString(2);
+                        int columnSize = dbResult.getInt(3);
+                        int scale = dbResult.getInt(4);
+                        int position = dbResult.getInt(5);
+                        String parameterTypeStr = dbResult.getString(9);
+                        boolean hasInplaceTableType = BOOLEAN_TRUE.equals(dbResult.getString(8));
+                        boolean hasDefaultValue = BOOLEAN_TRUE.equals(dbResult.getString(10));
```
ShadelessFox: Please use `JDBCUtils.safeGetXXX` to silently log the error instead of failing. Boolean comparison can also be replaced with an appropriate utility function.

It also removes the need for having `try-catch` block around it.

plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANAProcedure.java
```diff
@@ -17,24 +17,169 @@
...
+                            parameterType = DBSProcedureParameterKind.IN;
+                        } else if(parameterTypeStr.equals(PARAMETER_TYPE_INOUT)) {
+                            parameterType = DBSProcedureParameterKind.INOUT;
+                        } else if(parameterTypeStr.equals(PARAMETER_TYPE_OUT)) {
+                            parameterType = DBSProcedureParameterKind.OUT;
+                        } else if(parameterTypeStr.equals(PARAMETER_TYPE_RETURN)) {
+                            parameterType = DBSProcedureParameterKind.RETURN;
+                        } else {
+                            parameterType = DBSProcedureParameterKind.UNKNOWN;
+                        }
```
ShadelessFox: I'd suggest replacing it with `switch` statement.

plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANAProcedure.java
```diff
@@ -17,24 +17,169 @@
...
+                            } else {
+                                String tableTypeSchema = dbResult.getString(6);
+                                String tableTypeName = dbResult.getString(7);
+                                GenericSchema schema = getDataSource().getSchema(tableTypeSchema);
+                                if (schema != null) {
+                                    tableType = schema.getTable(monitor, tableTypeName);
+                                }
+                            }
+                        }
+                        GenericProcedureParameter column = new HANAProcedureParameter(
```
ShadelessFox: Just a minor issue, variable `column` is redundant here and can be inlined.

plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANAProcedureParameter.java
```diff
@@ -0,0 +1,95 @@
...
+    public DBSObject getTableType() {
+        return tableType;
+    }
+
+    @Property(viewable = true, order = 26, optional = true)
+    public Boolean getHasInplaceTableType() {
+        if(inplaceTableType != null)
+            return true;
+        return null;
+    }
```
ShadelessFox: DBeaver recognizes `hasXXX` getters just fine.

```suggestion
    public boolean hasInplaceTableType() {
        return inplaceTableType != null;
    }
```

The method duplicated below could be removed.

ShadelessFox (changes requested):
Thanks for contributing!

ShadelessFox (approved):
👍🏻

Matvey16:
Verified

Issue #14729, opened 2021-11-30
Querying empty string without specifying alias causes an error rendering the result


feeva:
#### System information: 
- Operating system (distribution) and version - **Windows 11**
- DBeaver version - **21 and later**
- Java version  - **11.0.12**

#### Connection specification:
- Database name and version - **mysql**

#### Describe the problem you're observing:
Querying empty string without specifying alias causes an error rendering the result. For example, `select name, created_date, '' from my_table` doesn't render the result table and then the ctrl+enter shortcut won't work.

#### Include any warning/errors/backtraces from the logs

```
java.lang.IllegalArgumentException: Argument cannot be null
	at org.eclipse.swt.SWT.error(SWT.java:4874)
	at org.eclipse.swt.SWT.error(SWT.java:4808)
	at org.eclipse.swt.SWT.error(SWT.java:4779)
	at org.eclipse.swt.graphics.GC.drawStringInPixels(GC.java:2159)
	at org.eclipse.swt.graphics.GC.drawString(GC.java:2154)
	at org.jkiss.dbeaver.ui.controls.lightgrid.GridColumnRenderer.paint(GridColumnRenderer.java:144)
	at org.jkiss.dbeaver.ui.controls.lightgrid.LightGrid.paintColumnsHeader(LightGrid.java:2519)
	at org.jkiss.dbeaver.ui.controls.lightgrid.LightGrid.paintColumnHeader(LightGrid.java:2506)
	at org.jkiss.dbeaver.ui.controls.lightgrid.LightGrid.paintHeader(LightGrid.java:2475)
	at org.jkiss.dbeaver.ui.controls.lightgrid.LightGrid.onPaint(LightGrid.java:2212)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:234)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4237)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1060)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1084)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1069)
	at org.eclipse.swt.widgets.Composite.WM_PAINT(Composite.java:1536)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4810)
[...]

PR #14775, opened 2021-12-02
#14729 avoid to compare nullable values

LonwoLonwo:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
```diff
@@ -569,8 +569,8 @@ public static DBSEntityAssociation getAssociationByAttribute(DBDAttributeBinding
     public static boolean equalAttributes(DBCAttributeMetaData attr1, DBCAttributeMetaData attr2) {
         return
             attr1 != null && attr2 != null &&
-            SQLUtils.compareAliases(attr1.getLabel(), attr2.getLabel()) &&
-            SQLUtils.compareAliases(attr1.getName(), attr2.getName()) &&
+            SQLUtils.compareAliases(CommonUtils.notEmpty(attr1.getLabel()), CommonUtils.notEmpty(attr2.getLabel())) &&
+            SQLUtils.compareAliases(CommonUtils.notEmpty(attr1.getName()), CommonUtils.notEmpty(attr2.getName())) &&
```
ShadelessFox, line 573: Maybe it would be better to make `SQLUtils.compareAliases` accept nullable values?

LonwoLonwo: I thought about it, but I'm not sure.

Issue #14791, opened 2021-12-04
[21.3.0] Partition sizes aren't displayed anymore in Database Navigator


Hermholtz:
#### System information: 
- Operating system (distribution) and version: Windows 11
- DBeaver version: 21.3.0
- Additional extensions: not used

#### Connection specification:
- Database name and version: PostgreSQL 14
- Driver name: default
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No

#### Describe the problem you're observing:
In version 21.2.0 partition sizes were correctly displayed. Today I upgraded to 21.3.0 and the place in the Database Navigator where they used to be is empty. Please see the screenshot. I've tried to lookup a setting that could be responsible for that, but failed.

#### Steps to reproduce, if exist:
- Open the Database Navigator
- Navigate to a PostgreSQL partitioned table
- Expand the "Partitions" subtree
- Observe lack of sizes

![image](https://user-images.githubusercontent.com/15248889/144725535-52f2955f-61c2-4a68-afd4-bad2cb47228d.png)

LonwoLonwo:
@Hermholtz Hello

Thanks for the bug report!

Hermholtz:
@LonwoLonwo Thank you very much for quick reaction and the fix!

LonwoLonwo:
Fixed in DBeaver 21.3.1

uslss:
verified

PR #14801, opened 2021-12-05
#14791 remove unnecessary condition - return partition tables statistics

LonwoLonwo:
2021-12-06 01_03_40-DBeaver Ultimate 21 3 0 - ACCOUNT_csv


ShadelessFox (commented):
I think this solution is too general 😕

LonwoLonwo:
@ShadelessFox I left this (removed) condition there :)

Issue #13112, opened 2021-07-06
r-click results grid->filter ->customize filters--ability to select multiple columns at once to move up/down instead of selecting one-by-one & clicking move-up/-down


gthbusr0:
**Is your feature request related to a problem? Please describe.**
it's time consuming to move columns one-by-one clicking move-up multiple times once for each col we'd like to move up to show these cols first in the results grid
**Describe the solution you'd like**
it saves time to enable multiple column selection to move fields up in the results grid ..shift-select-all-cols we'd like to move & just click move-up/move-down once
**Describe alternatives you've considered**
selecting in SQL is considered but for tables with 100s of cols it's timeconsuming
**Additional context**

Matvey16:
@iDeepMind  Thank for the suggestion, we will implement it.

Matvey16:
Verified

PR #14802, opened 2021-12-06
#13112 add multi selection for row filter settings

Destrolaric:
Adds multi-selection for row filter settings. Now you can select and move multiple fields at one time.


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/FilterSettingsDialog.java
```diff
@@ -506,6 +527,21 @@ protected void okPressed()
         super.okPressed();
     }
 
+    private void updateButtons() {
+        TreeItem[] selection = getSelection(columnsViewer.getTree());
+        if (selection == null || selection.length == 0) {
```
LonwoLonwo: ArrayUtils.isEmpty() can be helpful also

ShadelessFox (approved):

Issue #6806, opened 2019-09-23
Change file name


Ferrox12:
Hello,
how can I change the name of my script in relation of the project I'm working about?
Es. project's name is "House"; when I save my sheet, DBeaver save it as "Script-1". Can i change the output name?

serge-rider:
No, there is no such config option.
Probably we should add it. 
Any ideas what parameters we should use on new script names?

Ferrox12:
I think this will be a big improvement of the software. For the moment I have no other ideas. Congratulations for the work you've done. Dbeaver in my opinion is the best SQL client that I have tried

Matvey16:
Verified

Matvey16:
If preferences are opened and closed with "Apply and close" button, file name pattern will be erased

![QdFfNheiZW](https://user-images.githubusercontent.com/82543000/145195790-617aeeff-e744-4f7e-b930-7bcc7e127bbe.gif)

Matvey16:
Verified

PR #14813, opened 2021-12-06
#6806 Add project name to script

Destrolaric:
Added an ability to set custom preset for the script name patterns are supported. Refactored methods used previously to create title pattern into new Class - ScriptVariablesResolver. The field for script template is added in Preferences - Editor - SQL Editor - Scripts


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/internal/SQLEditorMessages.java
```diff
@@ -235,6 +235,7 @@
     public static String pref_page_sql_editor_checkbox_blank_line_delimiter_tip;
     public static String pref_page_sql_editor_checkbox_enable_sql_parameters;
     public static String pref_page_sql_editor_title_pattern;
+    public static String pref_page_sql_editor_file_name_patter;
```
LonwoLonwo: You missed "n" in the end.

LonwoLonwo:
6806 <> 6808

Issue #1000, opened 2016-11-24
Missing foreign keys


Azenet:
Hi,

I am not sure this is easily reproducible, if reproducible at all. I have an issue with DBeaver 3.8, sometimes the foreign keys will not load when I open a table, so fields that actually have foreign keys will not show the button to go to the linked row in another table.

Steps that seems to have the best reproducibility for me:
1. Click once a disconnected connection
2. Press Enter, a SQL editor will open and the connection will connect
3. Open a table that contains foreign keys
4. Foreign keys are not loaded until you go in Properties > Foreign keys and press F5.

temporaryaccount:
I can confirm, I had the same issue today on a MySQL connection.

serge-rider:
I saw this once but can't reproduce again. Will try again later.
Maybe it is a result of recent object editor redesign..

serge-rider:
I think this problem was a result of bug in metadata cache reading. It was fixed in https://github.com/serge-rider/dbeaver/commit/e267ec0ca45dd52384fc7b4897447c415c3bcaf1

PR #14831, opened 2021-12-07
kai-morich/dbeaver#10263 show only SYNOMYMs at HANA PUBLIC schema

kai-morich:
and boolean cleanup

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANASchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANASchema.java
@@ -26,6 +26,7 @@ import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
 import org.jkiss.dbeaver.model.struct.DBSEntity;
 
 public class HANASchema extends GenericSchema {
+	
     public HANASchema(@NotNull GenericDataSource dataSource, @Nullable GenericCatalog catalog, @NotNull String schemaName) {
         super(dataSource, catalog, schemaName);
     }
@@ -34,5 +35,9 @@ public class HANASchema extends GenericSchema {
     @Override
     public Class<? extends DBSEntity> getPrimaryChildType(@Nullable DBRProgressMonitor monitor) throws DBException {
         return HANATable.class;
+    }
+    
+    public boolean hasOnlySynonyms() { 
+    	return HANAConstants.SCHEMA_PUBLIC.equals(getName()); 
     }
 }
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCUtils.java
@@ -270,6 +270,17 @@ public class JDBCUtils {
         }
     }
 
+    public static boolean safeGetBoolean(@NotNull ResultSet dbResult, int columnIndex, @NotNull String trueValue)
+    {
+        try {
+            final String strValue = dbResult.getString(columnIndex);
+            return strValue != null && strValue.startsWith(trueValue);
+        } catch (Exception e) {
+            debugColumnRead(dbResult, columnIndex, e);
+            return false;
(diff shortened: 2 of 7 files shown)


plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANASchema.java
```diff
@@ -35,4 +38,8 @@ public HANASchema(@NotNull GenericDataSource dataSource, @Nullable GenericCatalo
     public Class<? extends DBSEntity> getPrimaryChildType(@Nullable DBRProgressMonitor monitor) throws DBException {
         return HANATable.class;
     }
+    
+    public boolean isPUBLIC() {
```
ShadelessFox: Please use `camelCase` for method names.

plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANASysViewColumn.java
```diff
@@ -23,6 +23,10 @@
 
 public class HANASysViewColumn extends HANATableColumn {
 
+	// boolean like columns in SYS schema are stored as VARCHAR(5) with values TRUE, FALSE
+    static final String TRUE = "TRUE";
+    static final String FALSE = "FALSE";
```
ShadelessFox: Perhaps we should move these constants into their own class, `HANAConstants`?

It should be marked as `final` and contain a `private` constructor since it's a utility class.

There are also several usages of these constants in the `HANAProcedure` class.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCUtils.java
```diff
@@ -270,6 +270,17 @@ public static boolean safeGetBoolean(ResultSet dbResult, int columnIndex)
         }
     }
 
+    public static boolean safeGetBoolean(ResultSet dbResult, int columnIndex, String trueValue)
```
ShadelessFox: I understand that you have copy-pasted this method, but let's add some annotations to season it well.

Let's mark `dbResult` and `trueValue` as `@NotNull`, as well move `{` onto the same line where the method's name lays.

plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANASchema.java
```diff
@@ -26,6 +26,9 @@
 import org.jkiss.dbeaver.model.struct.DBSEntity;
 
 public class HANASchema extends GenericSchema {
+	
+    static final String PUBLIC = "PUBLIC"; // pseudo schema for PUBLIC SYNONYMs
```
ShadelessFox: See my comments about `TRUE` and `FALSE` constants above.

ShadelessFox (changes requested):
Code refactoring is a great thing.

Also, please rename your commit messages according to our style: it should begin with an issue number (`#1000`) with a short message after it. After doing so, you can force-push to this branch.

plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANASchema.java
kai-morich: here it is not about a schema beeing public, but a schema having the exact name PUBLIC, so i had chosen upperaces.

ShadelessFox: I believe it's in the upper case for some legacy reasons. It's not an abbreviation, so it should follow general code style conventions.

kai-morich: i completely renamed the method

ShadelessFox (approved):
Looks good 👍🏻

Matvey16:
Verified

Issue #12864, opened 2021-06-16
Show details about LOCATION from EXTERNAL TABLES


fernandodelago:
I'm try to use only DBeaver, but always want to use EXTERNAL TABLES I have been feel upset. SSMS and Azure Data Studio are worst when compare with DBeaver, but, I have to use it, because I can't see information about data LOCATION from external tables on DBeaver.
I attached a print screen to show what I talk about.

DBeaver could show the information about external tables like SSMS, I my vision. It is the best form to show it.

A alternative option maybe show a additional  tab on table properties with this information. But, when the table was not a external table, only show "not applied for this object".

Another possibility to help the developers is create a subfolder inside the "tables" folder on left view "Database Navigator" to show only external tables for a better organization. It's a suggestion.

![compare_dbeaver_x_data_studio](https://user-images.githubusercontent.com/39306731/122221141-61b39d00-ce87-11eb-8b17-49a535675d78.jpg)

fernandodelago:
After some tests, I got the location with a SELECT.
That the result of this query  I want see at the  end of CREATE TABLE.

    select top 1 t1.name as name_table
         , T2.name AS data_source
         , T1.location
         , t3.name as file_format
         , T1.reject_type
         , T1.reject_value 
    from sys.external_tables T1,
         SYS.EXTERNAL_DATA_SOURCES T2,
         sys.external_file_formats T3,
         INFORMATION_SCHEMA.COLUMNS t4
    where T1.data_source_id = T2.DATA_SOURCE_ID
    and t1.file_format_id = t3.file_format_id
    AND upper(T1.name) = UPPER(T4.TABLE_NAME)
    AND UPPER(t4.TABLE_SCHEMA) = <input_schema_name>
    AND UPPER(T4.TABLE_NAME) = <input_table_name>

LonwoLonwo:
Thanks for the suggestion and script.

ShadelessFox:
Implemented:

![](https://user-images.githubusercontent.com/35821147/145281208-d362742d-82b0-4f59-bbc2-9b457d9dc9ed.png)

Matvey16:
Verified

PR #14846, opened 2021-12-08
#12864 Show external table information in DDL

ShadelessFox:
image


LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.ext.mssql/OSGI-INF/l10n/bundle.properties
```diff
@@ -183,6 +187,8 @@ tree.synonyms.node.name            = Synonyms
 tree.table.node.name               = Table
 tree.tableTriggers.node.name       = Triggers
 tree.tables.node.name              = Tables
+tree.externalTable.node.name       = External Tables
```
LonwoLonwo: Maybe "External Table"? :upside_down_face:

PR #14847, opened 2021-12-08
#6077 pgAgent jobs support

### LINKED ISSUE #6077: Create jobs by GUI
Hi,
In DBEaver there is some way to create a job through the GUI interface. Same thing exists in PGAdmin?

image

ShadelessFox:
image


uslss:
scheduler settings disappear after refresh
dbeaver_RiYQe3ItKy

it is the unobvious error message when trying to save the scheduled state with (Jan and April, 31st) or (Feb 29th,30th,31st) checked 
image

uslss:
New Job, Step and Schedule always have the same name, even if the step with the same default name exists

image

uslss:
It is needed to reload parent Job node to see newly created Steps and Schedules
dbeaver_H1jnjhVcyc

uslss:
Changed Job properties are reset after refresh
dbeaver_O1qq48YHH6

uslss:
It is impossible to disable any step due to the error
image

uslss:
Description cannot be deleted - update query isn't generated
bQ435aU9Ov

Issue #12503, opened 2021-05-17
Error when exporting table to JSON with a DATE type column 


Matvey16:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 21.0.5

#### Connection specification:
- Database name and version
- Driver name DB2
- Do you use tunnels or proxies: SSH

#### Steps to reproduce, if exist:
1. Create a table with a DATE type column
2. Fill the column with at least one date
3. Select `Export data` from the context menu
4. Select JSON format
5. Turn on "Format dates in ISO 8601" option
6. Start export.

**Expected result:** Export is completed

**Actual result:** Error appears

![2021-05-17_16h08_13](https://user-images.githubusercontent.com/82543000/118493848-358eea00-b72a-11eb-82f9-e228770f8888.png)

Matvey16:
Verified

PR #14848, opened 2021-12-08
#12503 fix error when exporting date or time type column

E1izabeth:
Handle cases when there is no time component or no date component in sql timestamp representation.


ShadelessFox (approved):
👍🏻

PR #14917, opened 2021-12-15
dbeaver/dbeaver-ee#1108 add ability to bind bracket type for in statement

Destrolaric:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -56,6 +56,7 @@
     private static final String[][] DEFAULT_BEGIN_END_BLOCK = new String[0][];
     private static final String[] CORE_NON_TRANSACTIONAL_KEYWORDS = new String[0];
     public static final String[] DML_KEYWORDS = new String[0];
+    public static final Pair<String, String> STRING_PAIR = new Pair<>("(", ")");
```
ShadelessFox:
```suggestion
    public static final Pair<String, String> IN_CLAUSE_PARENTHESES = new Pair<>("(", ")");
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialect.java
```diff
@@ -385,6 +385,9 @@
     @Nullable
     String getTestSQL();
 
+    @NotNull
+    Pair<String, String> getInClauseBrackets();
```
ShadelessFox:
```suggestion
    Pair<String, String> getInClauseParentheses();
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -118,6 +119,12 @@ protected void addSQLKeywords(Collection<String> allKeywords) {
         }
     }
 
+    @NotNull
+    @Override
+    public Pair<String, String> getInClauseBrackets() {
+        return STRING_PAIR;
+    }
```
ShadelessFox:
```suggestion
    public Pair<String, String> getInClauseParentheses() {
        return IN_CLAUSE_PARENTHESES;
    }
```

ShadelessFox (changes requested):

ShadelessFox (changes requested):
Please squash these commits into one. There is no point in separating them, as compilation will fail anyway.

PR #14889, opened 2021-12-13
#14808 Hide triggers, synonyms, sequences folders for babelfish

E1izabeth:
Hide triggers, synonyms, sequences folders for babelfish connection over SQLServerDataSource. Also some indentation fixed.

PR #14932, opened 2021-12-17
#14669 Hide non-default databases for babelfish

E1izabeth:
Only default database is shown for now.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDataSource.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDataSource.java
@@ -64,11 +64,13 @@ public class SQLServerDataSource extends JDBCDataSource implements DBSInstanceCo
     private String serverVersion;
 
     private volatile transient boolean hasStatistics;
+    private boolean isBabelfish;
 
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDataSource.java
```diff
@@ -93,7 +98,7 @@ public String getServerVersion() {
         return serverVersion;
     }
 
-    private boolean isNotBabelfishDatasource() {
+    public boolean isNotBabelfishDatasource() {
         boolean isBabelfish = SQLServerUtils.isDriverBabelfish(getContainer().getDriver());
         return !isBabelfish;
     }
```
ShadelessFox: I don't quite follow the reason why this method is negated. Also, what is the point of having this method? It's just a wrapper that calls a utility function. Can we inline it?

E1izabeth: The reason for this method is that the expression has already been shared between three different indirectly referenced methods implementing visibility conditions. (see [this commit](https://github.com/dbeaver/dbeaver/pull/14889/commits/e45306f3b42e33cb78eb1e413530ed30acfb028d))

Issue #14187, opened 2021-10-11
SQL script editor bug when delete


lutischan-ferenc:
#### System information: 
- Win 10
- 21.2.2.202110030714
- Additional extensions -

#### Connection specification:
- Postgresql 13
- org.postgresql.Driver (42.2.20)

#### Describe the problem you're observing:
1. E.g. "SELECT id, c1, c2, c3 FROM table;" from SQL script window.
2. select a row and mark it for delete.
3. browse over the row limit (e.g. 200)
4. press "save"
5. the row doesn't deleted

Matvey16:
@lutischan-ferenc Thank you for the report

Matvey16:
Happens not only with row deletion, but also if value in any cell was changed. After fetching additional rows, all of the changes are automatically reverted and the script with changes appears empty, so saving it does nothing. Reproducible in data editor.

ShadelessFox:
1. The confirmation dialog pops up after data is fetched:
![image](https://user-images.githubusercontent.com/35821147/146428242-cb3b5e02-9846-417d-9c3a-9049061290dd.png)

2. The confirmation dialog can't be closed by pressing `Cancel`:
![jexsbXi5f8](https://user-images.githubusercontent.com/35821147/146428471-8355d801-492e-4aba-9c1f-71530c20097e.gif)

ShadelessFox:
1. The confirmation dialog pops up several times after pressing `Yes`, effectively performing `UPDATE` every time:
![R7aF1Ngqr7](https://user-images.githubusercontent.com/35821147/147105603-6eb6fa2f-d9d4-451a-a36a-74b4ba6515b3.gif)

2. The confirmation dialog pops up for the second time after pressing either `No` or `Cancel`:
![hekEjf85kZ](https://user-images.githubusercontent.com/35821147/147105707-709bd3a3-6795-47ed-ba5e-b4d53b1039e3.gif)

3. After pressing `Cancel` no more data could be loaded, refresh doesn't do anything:
![zKpzb7yiYs](https://user-images.githubusercontent.com/35821147/147106207-26cb9c5e-c5db-4e00-a50d-d2a3dbcd0db9.gif)

Matvey16:
1. Pressing cancel in a prompt leads to errors in fetch buttons
```
org.eclipse.core.commands.NotEnabledException: Trying to execute the disabled command org.jkiss.dbeaver.core.resultset.fetch.page
[...]

PR #14940, opened 2021-12-17
#14187 add confirmation dialog

Destrolaric:


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -221,7 +221,15 @@
...
     private long lastThemeUpdateTime;
+    private boolean isCancelled = false;
 
+    public boolean isCancelled() {
+        return isCancelled;
+    }
+
+    public void setCancelled(boolean cancelled) {
+        isCancelled = cancelled;
+    }
```
ShadelessFox: This property can be avoided by using `dataReceiver.isHasMoreData` and `dataReceiver.setHasMoreData`.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -3702,6 +3703,10 @@ public void readNextSegment()
         if (!dataReceiver.isHasMoreData()) {
             return;
         }
+        if (!checkForChanges()) {
+            return;
+        }
+        dataReceiver.setHasMoreData(true);
```
ShadelessFox, line 3709: Wrong assumption. Probably you need to refactor your logic one more time.

ShadelessFox (changes requested):

Issue #13924, opened 2021-09-16
Filter result set on aliased column


igorzhilin:
When filtering the result set on a calculated column (with alias), I receive an error: 
`Column "calculated_column_name" does not exist. Hint: Perhaps you meant to reference the column "source_column_name".`

I would like to be able to filter on calculated column without referencing the source column.

To be able to filter on a calculated column, I have to provide the source column name in the filter. Then it works.

Example:
Query with calculated columns in PostgreSQL
```
select
	datname				as db_name,
	pg_database_size(datname)	as db_size_bytes
from
	pg_catalog.pg_database
```
![image](https://user-images.githubusercontent.com/9438635/133609815-1a10eba1-9ed1-4e84-9dfb-215139a0edba.png)

Filter
![image](https://user-images.githubusercontent.com/9438635/133609869-904d6934-dd0f-4c52-8913-438aae76e598.png)

Error
![image](https://user-images.githubusercontent.com/9438635/133609887-8e3e4d2b-18a0-4d33-842f-3556db6230af.png)

tati-kru:
This is a bug, thank you for the report.

Skydiveroid:
This can be fixed by using nested selects for query filtering

Matvey16:
Doesn't work for some columns

![j3kzu3pfyX](https://user-images.githubusercontent.com/82543000/146885089-4c6ff703-b97b-42f4-b1c1-8f31ab96a065.gif)

Matvey16:
Doesn't work if aliased column is array type

![lOsrj8iJzH](https://user-images.githubusercontent.com/82543000/147132445-b8384126-9222-46ad-9da9-5537be9c62a9.gif)

Matvey16:
New ticket #15011 created for array columns. Verified for other types

PR #14984, opened 2021-12-22
#13924 Fix filter result set on computed aliased columns

E1izabeth:
The [previously proposed solution](https://github.com/dbeaver/dbeaver/pull/14957) did not work for computed columns and was revised.

The current solution offers to patch query with aliased columns and replace alias with aliased expression. I didn't find any possibility in current model to add custom expression as constraint. So, a new attribute constraint was proposed for that.

### DIFF
--- a/bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
+++ b/bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
@@ -1028,4 +1028,9 @@ public class CommonUtils {
         matcher.appendTail(sb);
         return sb.toString();
     }
+
+    @NotNull
+    public static <T> Optional<T> optionalIfInstanceOf(@NotNull Class<T> type, Object obj) {
+        return obj != null && type.isAssignableFrom(obj.getClass()) ? Optional.of((T)obj) : Optional.empty();
+    }
 }
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/DBDAttributeAssociatedExpressionConstraint.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/DBDAttributeAssociatedExpressionConstraint.java
@@ -1,0 +1,32 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2021 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/DBDAttributeAssociatedExpressionConstraint.java
```diff
@@ -0,0 +1,17 @@
+package org.jkiss.dbeaver.model.sql;
```
flambchap, line 17: Please add our copyright header.

E1izabeth: Added

bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
```diff
@@ -1028,4 +1028,9 @@ public static String replaceAll(@NotNull String input, @NotNull String regex, @N
         matcher.appendTail(sb);
         return sb.toString();
     }
+
+    @NotNull
+    public static <T> Optional<T> optionalIfInstanceOf(@NotNull Class<T> type, Object obj) {
+        return obj != null && type.isAssignableFrom(obj.getClass()) ? Optional.of((T)obj) : Optional.empty();
+    }
```
flambchap, line 1035: It is funny because I've proposed a similar util function twice, and it was rejected with the reasoning like "You are the only person who would use it" :)

Overall, I propose doing in in the following way:


```suggestion
    @Nullable
    public static <T> T tryCast(@Nullable Object obj, @NotNull Class<T> clazz) {
        if (clazz.isInstance(obj)) {
            return clazz.cast(obj);
        }
        return null;
    }
```
It would be more flexible this way.
* It would allow not only strict type equality but subclasses as well;
* It would make it sound not only in the `Stream` context. Currently, using it outside of a Stream makes the code too verbose;
* It would still be possible to produce `Optional<T>` by using `Optional.ofNullable()`

And I need to confess that my proposal is pretty much a copy-paste from the IntelliJ Platform :)

E1izabeth: Well, both of these method have their scenarious to use. The proposed `optionalIfInstanceOf(..)` aims to minimize verbosity preserving the readability, while `tryCast(..)` is more useful in imperative context, because there is no point in constantly repeating one and the same construct :)
(P.S. just pointless muttering either way)

PR #14995, opened 2021-12-22
#14187 fix thread race condition

Destrolaric:


serge-rider:
What is `isCancelled` for? It is never used.
Why adding `isLoadingData` when we already have `dataPumpRunning`?

`isAwaitingDialogResult` must be reanamed to something meaningful. Generally there is no dialog and we don't wait for any result.

Issue #14817, opened 2021-12-07
How to clear Bind Parameter(Variable) ?


podoni:
The method I know is to delete the Bind Parameter panel one by one or manually edit the file(sql-variables-driver-oracle_oracle_thin.json) and restart the deaver.
Is there a way to delete all at once by selecting an area in the Bind Parameter panel?

Skydiveroid:
Thank you for the idea.

Matvey16:
Parameters are not actually deleted, they just disappear from the list until the new one is created or the editor is reopened

![explorer_OAvIDwQta0](https://user-images.githubusercontent.com/82543000/148911863-6e76848d-1b79-49e6-bf64-609a5db38302.gif)

Matvey16:
Parameters are still remaining in variable panel (even thogh they are actually deleted) if they were created before current editor was opened 

![dbeaver_jo5szXooRX](https://user-images.githubusercontent.com/82543000/150157282-8ae3c079-afa5-411b-a1e7-b4f233608765.gif)

Matvey16:
Verified

PR #15087, opened 2021-12-29
#14817 add multi deletion in bind parameters

Destrolaric:
Now we can delete multiple variable values in the Bind parameters. 

And delete parameters

https://user-images.githubusercontent.com/20579256/147753457-e3569ebd-1ce7-4764-b159-9bc7e0cf3400.mp4


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/dialogs/SQLQueryParameterBindDialog.java
```diff
@@ -142,7 +144,16 @@ protected Control createDialogArea(Composite parent)
...
             valueColumn.setWidth(200);
-
+            paramTable.addKeyListener(new KeyAdapter() {
+                @Override
+                public void keyPressed(KeyEvent e) {
+                    if (e.character == SWT.DEL){
+                        for (TableItem tableItem : paramTable.getSelection()) {
+                            tableItem.setText(2, "");
+                        }
+                    }
```
ShadelessFox: I do not understand the purpose behind this code.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/variables/RemoveVariablesAction.java
```diff
@@ -22,26 +22,30 @@
...
     }
 
     @Override
     public void run() {
         if (UIUtils.confirmAction(
             editor.getSite().getShell(),
-            "Delete variable '" + varName + "'?",
-            "Delete variable '" + varName + "'?"))
+            "Do you want to delete variables?",
+            "Delete variables ?"))
```
ShadelessFox: After this change, the user doesn't know which variable they are about to delete in case of single variable deletion.

Let's keep the old messages for a single variable.

Also, I'd like to see localization for these strings as well.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/variables/SQLVariablesPanel.java
```diff
@@ -362,11 +364,10 @@ public void run() {
                 @Override
                 public void run() {
                     if (!varsTable.getSelection().isEmpty()) {
-                        Object varElement = ((IStructuredSelection) varsTable.getSelection()).getFirstElement();
-                        if (varElement instanceof DBCScriptContext.VariableInfo) {
-                            RemoveVariableAction action = new RemoveVariableAction(mainEditor, ((DBCScriptContext.VariableInfo) varElement).name);
-                            action.run();
-                        }
+                        final StructuredSelection selection = (StructuredSelection) varsTable.getSelection();
+                        List<String> varsList = Arrays.stream(selection.toArray()).map(el -> ((DBCScriptContext.VariableInfo) el).name).collect(Collectors.toList());
```
ShadelessFox, line 367: Unnecessary usage of Stream API.

The `selection` variable could be inlined into a `for-loop` that will accumulate variable names into a list.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/variables/SQLVariablesPanel.java
```diff
@@ -362,11 +364,10 @@ public void run() {
...
+                        RemoveVariablesAction action = new RemoveVariablesAction(mainEditor, varsList);
+                        action.run();
```
ShadelessFox: Redundant variable, please call `run()` directly on a newly created object instance.

ShadelessFox (changes requested):

serge-rider:
We need to add support of parameters delete as well.
Confirmation message must contain variable/parameter names.

Issue #15071, opened 2021-12-28
[Data Editor] Update table is executed N times if you press the Save button N times in a row


flambchap:
#### System information: 
- Operating system (distribution) and version Win 11
- DBeaver version 21.3.2 devel

#### Connection specification:
- Database name and version Hive 2.3 (although I believe it's not important)

#### Steps to reproduce, if exist:
1. Create a table in Hive with varchar of a float column (I think that DBMS, data types, and the number of columns are not important, though)
2. In the data editor, add a new row.
3. Press the Save button once.
4. While the update table is still in progress, press the Save button again.
5. After the update is finished, press F5 and see that the row was inserted twice.

![hive](https://user-images.githubusercontent.com/26031301/147568286-b0440c87-cf84-428b-8be4-af2cba5ffe30.gif)

LonwoLonwo:
Fixed.

uslss:
verfied

PR #15091, opened 2021-12-29
#15071 Disable "Save" and "Cancel" buttons 

LonwoLonwo:
during the update data process progress

PljQijKSyh

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetPersister.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetPersister.java
@@ -627,6 +627,7 @@ class ResultSetPersister {
         @Override
         protected IStatus run(DBRProgressMonitor monitor) {
             model.setUpdateInProgress(this);
+            UIUtils.syncExec(viewer::fireResultSetChange); // Update "save" and "cancel" buttons
             updateStats = new DBCStatistics();
             insertStats = new DBCStatistics();
             deleteStats = new DBCStatistics();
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetPersister.java
```diff
@@ -627,6 +627,7 @@ public Throwable getError() {
         @Override
         protected IStatus run(DBRProgressMonitor monitor) {
             model.setUpdateInProgress(this);
+            UIUtils.syncExec(viewer::fireResultSetChange); // Update "save" and "cancel" buttons
```
ShadelessFox: Can we fire this event outside the `DataUpdaterJob` job, so we can avoid using `syncExec` here?

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/handler/ResultSetPropertyTester.java
```diff
@@ -49,6 +49,7 @@
     public static final String PROP_CAN_NAVIGATE_HISTORY = "canNavigateHistory";
     public static final String PROP_EDITABLE = "editable";
     public static final String PROP_CHANGED = "changed";
+    private static final String PROP_UPDATE_NOT_IN_PROGRESS = "updateNotInProgress";
```
ShadelessFox: Generally, I wouldn't say I like the idea of providing properties that return negative/inverted values.

It would be more convenient to return whether the update is **in** progress or not.

serge-rider:
- updateNotInProgress -> canPersistData
- Use asyncExec in persister job to avoid unneccessary blocks

Issue #15042, opened 2021-12-27
Disable `Append to the file` option for binary file formats


Destrolaric:
XLSX files are binary. Appending to them will cause files to be unreadable. We need to turn off this functionality when the XLSX file is selected.

uslss:
The option should be unchecked in case it is greyed out
![image](https://user-images.githubusercontent.com/31996417/147730767-1a7e33cb-4087-4973-8b86-65691ca66e42.png)

uslss:
verified

PR #15096, opened 2021-12-30
#15042 unselect append to file if already selected

Destrolaric:
unselects `append to file` if no longer valid to be used


plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/pages/stream/StreamConsumerPageOutput.java
```diff
@@ -276,7 +276,7 @@ public void activatePage() {
 
         clipboardCheck.setSelection(settings.isOutputClipboard());
         singleFileCheck.setSelection(settings.isUseSingleFile());
-        appendToEndOfFileCheck.setSelection(settings.isAppendToFileEnd());
+        appendToEndOfFileCheck.setSelection(!isBinary && settings.isAppendToFileEnd());
```
ShadelessFox: Other conditions are not met, please take look at:
https://github.com/dbeaver/dbeaver/blob/9667baa4d6070e25090a9513136c532a2010fef8/plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/pages/stream/StreamConsumerPageOutput.java#L259

ShadelessFox (changes requested):

PR #15104, opened 2022-01-01
Add some new IT translations

capitanfuturo:
Hi all,
I have read the localization wiki page and I tried to update some Italian i10n entries. I use DBeaver every day in my work life and I want to contribute. I hope to be helpful.


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/CoreResources_it.properties
```diff
@@ -1,69 +1,413 @@
...
+
+dialog_connection_edit_wizard_shell_cmd = Comandi Shell
+
+dialog_connection_edit_wizard_shell_cmd_directory_label = Cartella di lavoro
+
+dialog_connection_edit_wizard_shell_cmd_directory_title = Cartella di lavoro
+
+dialog_connection_edit_wizard_shell_cmd_pause_label = Metti l'esecuzione in pausa dopo (ms)
+
+dialog_connection_edit_wizard_shell_cmd_variables_hint_label = Non si possono usare variabili nei comandi. Clicca per vedere l'elenco.
```
E1izabeth: Hi, @capitanfuturo! Thank you for contribution.
I used translator and noticed that _"Non si possono usare variabili nei comandi"_ means something like _"You **can't** use variables in commands"_, while this property contains "_You **can** use variables in commands"_ in English properties file. Could you, please, check, is this a mistake?

capitanfuturo: Hi @E1izabeth, thank you for your review. Yes it's my mistake, I checked the English version and in Italian it should be "Puoi usare delle variabili nei comandi"

E1izabeth: Thank you for correction!

PR #15145, opened 2022-01-11
#14817 fix parameters not being deleted

Destrolaric:
The reason why they were not deleted is that dialog acted preserved previous values in its static variable and rewrote them every time OK was pressed


serge-rider:
Please double check with named parameters (:name), anonymous parameters (?) and variables (${name}).
Variables must be visible across all scripts and must be visible in dialog. Editor reopen or DBeaver restart mustn't break this behavior.

PR #15179, opened 2022-01-14
#14817 fix deletion of variables and parameters

Destrolaric:
Fix multiple issues with deletion after previous commits:
- Fixed variables not being deleted if all variables in the script are set to be removed
- Fixed parameter type recognition


Destrolaric:
Changes in behavior:
- Parameters are now bound with their prefix, if prefix is changed you will need to recreate parameter
- Variable first tries to find its name without prefix, if it will not find anything it will try to find parameter name with prefix

Issue #14296, opened 2021-10-19
Oracle Alternative Quoting Mechanism Not Properly Ended


ShinobiAlsedo:
#### System information: 
- Windows 10 64-bit
- DBeaver 21.2.3
- No additional extensions

#### Connection specification:
- Oracle Database 12c
- Oracle JDBC driver 12.2.0.1
- No tunnels or proxies

#### Describe the problem you're observing:

Alternative quoting mechanism ("Q") for string literals is not working correctly when the alternative quoting character is repeated at the end of a string.

#### Steps to reproduce:

I execute this SQL statement:

`SELECT q'!Test!!' FROM DUAL;`

And get this error:

> SQL Error [933] [42000]: ORA-00933: SQL command not properly ended

If I first highlight the whole SQL statement or remove the semi-colon before executing it, it runs without errors and outputs:

```
Q'!TEST!!'|
----------+
Test!     |
```

In both cases, the appearance is wrong with all of `' FROM DUAL;` being SQL string colour and `q'{{Test}}` not being so.

Other scenarios:

* If you add an odd number of quotes to the string (e.g. `q'!It's a test!!'`), it'll run fine but still appear wrong.

* If the string has the alternative quoting character followed by a quote within it (e.g. `q'!Test!'!'`), it'll correctly throw an error.

* If the character before the final alternative quoting character is anything else (e.g. `q'!Test! !'`), it'll both run and look fine.

* If the string has an odd number of the alternative quoting characters (e.g. `q'!Test!!!'`), it'll both run and look fine.

#### Include any warning/errors/backtraces from the logs
```
!ENTRY org.jkiss.dbeaver.model 4 0 2021-10-19 17:20:57.514
!MESSAGE SQL Error [933] [42000]: ORA-00933: SQL command not properly ended

!SUBENTRY 1 org.jkiss.dbeaver.model 4 0 2021-10-19 17:20:57.515
!MESSAGE SQL Error [933] [42000]: ORA-00933: SQL command not properly ended

!SUBENTRY 2 org.jkiss.dbeaver.model 4 0 2021-10-19 17:20:57.515
!MESSAGE Error : 933, Position : 27, Sql = SELECT q'!Test!!' FROM DUAL;, OriginalSql = SELECT q'!Test!!' FROM DUAL;, Error Msg = ORA-00933: SQL command not properly ended

!STACK 0
[...]

PR #15184, opened 2022-01-15
#14296 fix Oracle Q-strings recognition

E1izabeth:
Refactored to be clearly readable and fixed incorrect Oracle Q-strings recognition.


ShadelessFox (commented):
Could you please consider taking a look at [PostgreEscapeStringRule](https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/sql/PostgreEscapeStringRule.java) class?

It does nearly the same thing. Perhaps we should make it generic to avoid code duplication?

E1izabeth:
> Could you please consider taking a look at [PostgreEscapeStringRule](https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/sql/PostgreEscapeStringRule.java) class?
> 
> It does nearly the same thing. Perhaps we should make it generic to avoid code duplication?

These two mechanisms seem to be similar, but they do the opposite things: there is no escaping logic in Q-strings and there is no custom brackets mechanism in E-strings.

ShadelessFox:
I see, okay.

ShadelessFox (approved):

PR #15190, opened 2022-01-16
Update ModelResources Japanese translation

yoshinorin:
Hi.

I updated Japanese translation.

This PR just the `ModelResources_ja`.
I'm going to try to translate others step by step.

Thank you :)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/messages/ModelResources_ja.properties
```diff
@@ -16,13 +18,14 @@ model_navigator_Connection = \u63A5\u7D9A
 model_connection_events_event_after_connect = \u63A5\u7D9A\u5F8C
 model_connection_events_event_after_disconnect = \u5207\u65AD\u5F8C
 model_connection_events_event_before_connect = \u63A5\u7D9A\u524D
-model_connection_events_wevent_before_disconnect = \u5207\u65AD\u524D
```
yoshinorin, line 19: This change is the fix of the typo.

```diff
- model_connection_events_wevent_before_disconnect
+ model_connection_events_event_before_disconnect
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/messages/ModelResources_ja.properties
```diff
@@ -76,32 +79,66 @@ model_jdbc_None = \u306A\u3057
 model_jdbc_precision = \u7CBE\u5EA6
 model_jdbc_Procedure = \u30D7\u30ED\u30FC\u30B8\u30E3
 model_jdbc_rename_object = \u30AA\u30D6\u30B8\u30A7\u30AF\u30C8\u306E\u540D\u524D\u3092\u5909\u66F4
-model_jdbc_repeatable_read = \u7E70\u308A\u8FD4\u3057\u8AAD\u307F\u8FBC\u3080
+model_jdbc_read_committed = READ COMMITTED
+model_jdbc_read_uncommitted = READ UNCOMMITTED
+model_jdbc_repeatable_read = REPEATABLE READ
```
yoshinorin, line 84: IMHO, The transaction isolation levels no-need to translate.
I have written in English to make it clear that there is no need to translate.

Please see MySQL Japanese docs. These words were not translated.

> https://dev.mysql.com/doc/refman/8.0/ja/innodb-transaction-isolation-levels.html

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/messages/ModelResources_ja.properties
```diff
@@ -76,32 +79,66 @@ model_jdbc_None = \u306A\u3057
 model_jdbc_precision = \u7CBE\u5EA6
 model_jdbc_Procedure = \u30D7\u30ED\u30FC\u30B8\u30E3
 model_jdbc_rename_object = \u30AA\u30D6\u30B8\u30A7\u30AF\u30C8\u306E\u540D\u524D\u3092\u5909\u66F4
-model_jdbc_repeatable_read = \u7E70\u308A\u8FD4\u3057\u8AAD\u307F\u8FBC\u3080
```
yoshinorin, line 79: Current `model_jdbc_repeatable_read` is wrong.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/messages/ModelResources_ja.properties
```diff
@@ -76,32 +79,66 @@ model_jdbc_None = \u306A\u3057
...
 model_struct_Index = \u30A4\u30F3\u30C7\u30C3\u30AF\u30B9
 model_struct_Inheritance = \u7D99\u627F
+model_struct_No_Action = NO ACTION
+model_struct_Not_NULL = NOT NULL
 model_struct_Other = \u305D\u306E\u4ED6
 model_struct_Primary_Key = \u4E3B\u30AD\u30FC
-model_struct_Restrict = \u5236\u9650
+model_struct_Restrict = RESTRICT
+model_struct_Set_Default = SET DEFAULT
+model_struct_Set_NULL = SET NULL
```
yoshinorin, line 111: IMHO, some of foreign-keys no-need translate, same as isolation-level translation.
I have written in English to make it clear that there is no need to translate.

Please see MySQL Japanese docs. These words were not translated.

> https://dev.mysql.com/doc/refman/5.6/ja/create-table-foreign-keys.html

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/messages/ModelResources_ja.properties
```diff
@@ -76,32 +79,66 @@ model_jdbc_None = \u306A\u3057
...
 model_struct_Virtual_Foreign_Key = \u4EEE\u60F3\u5916\u90E8\u30AD\u30FC
 model_struct_Pseudo_Key = \u7591\u4F3C\u30AD\u30FC
 model_struct_Unknown = \u4E0D\u660E
 
+dbp_connection_type_table_development = Development
+dbp_connection_type_table_production = Production
+dbp_connection_type_table_production_database = \u672C\u756A\u30C7\u30FC\u30BF\u30D9\u30FC\u30B9
+dbp_connection_type_table_regular_development_database = \u901A\u5E38\u306E\u958B\u767A\u30C7\u30FC\u30BF\u30D9\u30FC\u30B9
+dbp_connection_type_table_test = Test
+dbp_connection_type_table_test_database = \u30C6\u30B9\u30C8 (QA) \u30C7\u30FC\u30BF\u30D9\u30FC\u30B9
```
yoshinorin, line 123: I check the connection setting UI.
Below labels are English is better than Japanese.

* `dbp_connection_type_table_development`
* `dbp_connection_type_table_production`
* `dbp_connection_type_table_test`

serge-rider:
Thanks for contribution 👍 
Note: if some terms shouldn't be translated it is better to delete them from localization completely (thus English version will be used).

yoshinorin:
@serge-rider 

> Note: if some terms shouldn't be translated it is better to delete them from localization completely (thus English version will be used).

Thank you for your comment :)
Yes, I know sometimes you point out it in other PR reviews. If I translate everything by myself, I think that's okay.

But, I have written in English these terms purposely as I wrote in my comment. Because, when other translators see it in the future, they probably can't understand whether "should not have translated" or "could not be translated". Actually, many of the terms have already been translated by other translators. But, I can't understand whether "should not have translated" or "could not be translated" about untranslated terms.

The Japanese translations are included some mistranslation (ex: `REPEATABLE READ`, `PROCEDURE`). I assume these mistranslations were caused by the translator did not confirm a context. IMHO, if we use English to these terms instead of Japanese, may reduce mistranslation when other translators contribute to translation ​in the future. Because, they may notice "this term no need to translate".

Finally, I'm going to follow the dbeaver team's opinion & policy, so please give me a review.

Thank you :)

P.S: My English isn't so good so feel free to ask me if there is anything unclear.

PR #15223, opened 2022-01-19
Fixed PostgreInsertReplaceMethod not quoting the column in CONFLICT

Seanld:
When using the import data wizard (in my case, a CSV) on a Postgres table that had capitalized column names (like `HelloWorld` rather than `helloworld`) this caused the insert/replace function to error out, because it did not properly quote the column name, and Postgres defaults capitalized column names to all-lowercase when they are unquoted. This fixes that.

I'm not really a Java programmer, so I just did the most straight-forward fix.


LonwoLonwo:
Hello @Seanld 

Thanks for your contribution!
But we can't always use quotes for attribute names. Usually, we use quotes only for attributes that need them.

However, thanks for the bug report. I'll create a new issue on our board and fix this bug.

LonwoLonwo:
https://github.com/dbeaver/dbeaver/pull/15231

Issue #14780, opened 2021-12-03
show comment in left element tree


LinWanCen:
**Is your feature request related to a problem? Please describe.**
We are always frustrated when I read the table and column comment need open a tab.

**Describe the solution you'd like**
We hope show comment in left element tree, like DataGrip open Viev -> Appearance -> Details in Tree View.

https://intellij-support.jetbrains.com/hc/requests/2827972

ShadelessFox:
Hello @LinWanCen,

Please post screenshots or text instead of attaching external links.

LinWanCen:
![image](https://user-images.githubusercontent.com/14926671/149669514-bb013bc5-4f3d-4a91-a276-48b114d4106b.png)

> Please post screenshots or text instead of attaching external links.

ShadelessFox:
I see, thanks. It would be handy.

Currently, the description is displayed in a tooltip.

ShadelessFox:
Implemented:

![java_Kfdf54HkIC](https://user-images.githubusercontent.com/35821147/150457442-1382bba2-95d0-4eb1-a0b8-c0b67e0b9597.png)

Can be toggled in several ways:
* In Preferences, <kbd>User Interface</kbd> &rArr; <kbd>Navigator</kbd> &rArr; <kbd>Show objects description</kbd>
* In connection's context menu, <kbd>Connection view</kbd> &rArr; <kbd>Show objects description</kbd>

LinWanCen:
Is the column comment also show?
Expectation!

ShadelessFox:
Yes.

uslss:
verified

LinWanCen:
It looks like it's loaded asynchronously,
and most of the time we don't see all the comments.
Can there be an option to load all of them and cache them.

PR #15240, opened 2022-01-21
#14780 Show objects description in the navigator

ShadelessFox:
I must admit, it looks really cool, and I like it. Do we want it to be enabled by default?

java_Kfdf54HkIC


serge-rider:
I think it should be disabled by default. Descriptions may be too long/redundant.
Also please test on databases with lazy description load.

Issue #13581, opened 2021-08-19
ORA-00900 error when create or replace package body with "table.colum%type" internal parameter


zeleznypa:
#### System information: 
- Operating system (distribution) and version: MacOS 11.5.1
- DBeaver version: 21.1.5
- Additional extensions: No

#### Connection specification:
- Driver name: Oracle ojdbc8:RELEASE [12.2.0.1]

#### Describe the problem you're observing:
Unable to compile the package body when there is a "table.column%type" as internal type.
 
#### Steps to reproduce, if exist:
When the table_name.column_name really exists the following code

```sql
CREATE OR REPLACE package BODY pck_foo is
  function foo(
  ) return table_name.column_name%type;
  as
    vParam table_name.column_name%type;
  begin
    return vParam;
  end;
end pck_foo;
```

fail on exception

```
SQL Error [900] [42000]: ORA-00900: neplatný příkaz SQL

  Error : 900, Position : 0, Sql = end pck_foo, OriginalSql = end pck_foo, Error Msg = ORA-00900: neplatný příkaz SQL

  ORA-00900: neplatný příkaz SQL
```

But the following code works well

```
CREATE OR REPLACE package BODY pck_foo is
  function foo(
  ) return table_name.column_name%type;
  as
    vParam varchar2(255);
  begin
    return vParam;
  end;
end pck_foo;
```

Even the `table_name.column_name%type` is `varchar2(255)`.

The same code works well in Oracle SQL Developer

LonwoLonwo:
Hello @zeleznypa 
Thanks for the bug report!

This is some kind of statement parsing issue, as I can see.

As a workaround, you can highlight your statement in SQL Editor and use Ctrl+Enter.

zeleznypa:
Hello @LonwoLonwo, thanks for your fast reply.

Unfortunately for me the workaround is not usable.
I have a lot of packages that I need to deploy in a batch.

So I have to use the [@include](https://dbeaver.com/docs/wiki/Client-Side-Scripting/) mechanism that execute them in separate window and this mechanism still cause the issue :(

Matvey16:
Verified

PR #15264, opened 2022-01-24
#13581 add ability to match dialect specific token predicates

E1izabeth:
We can describe prefix and suffix of the grammar rule that could contain delimiter or other literal which was recognized as the end of a statement by our parser.

Such dialect-specific rules were described for Oracle `CREATE OR REPLACE package BODY` and `WITH` statements.
Examples:
```
CREATE OR REPLACE package BODY pck_foo is
  function foo(
  ) return table_name.column_name%type;
  as
    vParam table_name.column_name%type;
  begin
    return vParam;
  end;
end pck_foo;
```

```
WITH
  FUNCTION with_function(p_id IN NUMBER) RETURN NUMBER IS
  BEGIN
    RETURN p_id + p_id;
  END;
SELECT with_function(rownum)
FROM   dual
WHERE  rownum = 1;
```


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/tokens/predicates/SQLTokenPredicateFabric.java
```diff
@@ -0,0 +1,90 @@
...
+            try {
+                scanner.reset();
+                TPToken token = fRule.evaluate(scanner);
+                if (!token.isUndefined()) {
+                    SQLTokenType tokenType = token instanceof TPTokenDefault ? (SQLTokenType) ((TPTokenDefault) token).getData() : SQLTokenType.T_OTHER;
+                    return new SQLTokenEntry(string, tokenType);
+                }
+            } catch (Throwable e) {
+                // some rules raise exceptions in a certain situations when the string does not correspond the rule
+                // e.printStackTrace();
```
LonwoLonwo: Maybe this is a good place for some log.debug() message?

E1izabeth: done

Issue #13157, opened 2021-07-09
Choose SQL Script for ALL Datasources


xunxky:
#### **Is your feature request related to a problem? Please describe.**
I am quite frustrated when i want to open a script in one datasource and can't because it was "associated" with another datasource. 
This feature is highly unintuitive and it only poses a disadvantage to me. 

#### **Describe the solution you'd like**
However to NOT have to interfere too much with the inner workings of  DBeaver it would be great if something like  "Open SQL Script" or "Recent SQL script" (there is no difference i can see!)  could be added there. 
For example _"Open SQL Script (ALL)"_.
**Meaning**: it would **not pre-select** available scripts based on the datasource BUT **allways** show the _"Choose SQL Script"-Dialoge_ with all available scripts. 
It would be best if the **currently selected datasource** would be used/associated when opening the script no matter which datasource was associated with the script before.

#### **Describe alternatives you've considered**
I currently select the required SQL-Script from the project pane. 
However this renders the convenience given by the sql button and the "Choose SQL Script"-Dialoge void(which includes date and sorting options) 

#### **Additional context**
As this is a quite common sense feature which is not available yet, I am fairly certain that there is some individual who opposes this. It would be great if YOU could please give reasoning  or why you do not think that wouldn't be a good _additional_ feature. 
Thank You

PS: 
**Alternative solution**
one script could have more associations ... (ultimately that's a special case from the previously given ALL case)

Matvey16:
@xunxky You suggest adding another button in this tab, correct? It will open a list of all saved scripts and ignore data source association?

![dbeaver_oLC7PzKSpy](https://user-images.githubusercontent.com/82543000/125739886-4f308133-a2a8-4cc5-9e20-7644fe62bf3e.png)

xunxky:
> @xunxky You suggest adding another button in this tab, correct? It will open a list of all saved scripts and ignore data source association?
> 
> ![dbeaver_oLC7PzKSpy](https://user-images.githubusercontent.com/82543000/125739886-4f308133-a2a8-4cc5-9e20-7644fe62bf3e.png)

exactly!

PS: 
and as pointed out I am well aware that the data source resolution will be tricky because some people might expect that scripts will use the current data source while others might expect the data source associated with the script to be true.

Matvey16:
@xunxky Why do you think that opening a script from Project explorer (where scripts can be saved without associated data source) is not enough?

xunxky:
[...]

PR #15273, opened 2022-01-24
#13157 open project scripts

Destrolaric:
We now can press SHIFT + F3 to open project scripts instead of only datasource one


serge-rider:
1. We don't need new icon icon, we need to add checkbox in script selector.
2. Do not read scripts by datasource (it is super-slow in case of big number of connections)

Issue #15069, opened 2021-12-28
[clickhouse] Use SQL to limit fetch size is not working


DoS174:
I select "Use SQL to limit fetch size" for DB in clickhouse and the query does not change. In the end, LIMIT does not appear.
Checked in postgresql. Everything works there

Matvey16:
@DoS174 Thank you for the report, we will work on that.

uslss:
verified

DoS174:
The "Use SQL to limit fetch size" item started adding a LIMIT 100 row to the query, but I still get an error

`_SQL Error [1002]: ClickHouse exception, code: 1002, host: ***, port: ***; Code: 164. DB::Exception: Cannot modify 'max_result_rows' setting in readonly mode. (READONLY) (version 21.10.5.3 (official build))_
`

If I turn off the LIMIT in the settings, and add the LIMIT to the request manually, then everything works

dbeaver version 21.3.4

uslss:
verified

PR #15275, opened 2022-01-24
#15069 Add QueryTransformerLimit for Clickhouse

E1izabeth:
QueryTransformerLimit was added, so now it's possible to use 'Use SQL to limit fetch size' option.
Also quotation was added for `DEFAULT` and `SYSTEM` keywords which is not keywords for Clickhouse, but leads to errors in JSQLParser when it's unquoted and meet, for example, in identifier names.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseSQLDialect.java
+++ b/plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseSQLDialect.java
@@ -70,6 +70,10 @@ public class ClickhouseSQLDialect extends GenericSQLDialect {
         "toLowCardinality",
         "formatRow"
     };
+    private static final String[] CLICKHOUSE_NONKEYWORDS = {
+            "DEFAULT",
+            "SYSTEM"
+    };
 
     public ClickhouseSQLDialect() {
         super("Clickhouse SQL", "clickhouse");
@@ -82,8 +86,9 @@ public class ClickhouseSQLDialect extends GenericSQLDialect {
 
     public void initDriverSettings(JDBCSession session, JDBCDataSource dataSource, JDBCDatabaseMetaData metaData) {
         super.initDriverSettings(session, dataSource, metaData);
-        removeSQLKeyword("DEFAULT");
-        removeSQLKeyword("SYSTEM");
+        for (String word : CLICKHOUSE_NONKEYWORDS) {
+            removeSQLKeyword(word);
+        }
         addFunctions(Arrays.asList(CLICKHOUSE_FUNCTIONS));
     }
 
@@ -103,5 +108,16 @@ public class ClickhouseSQLDialect extends GenericSQLDialect {
     @Override
     public boolean supportsNestedComments() {
         return true;
+    }
+
+    //We should quote keywords which is not keywords for clickhouse, otherwise JSQLParser can't parse statements
+    @Override
+    public boolean mustBeQuoted(@NotNull String str, boolean forceCaseSensitive) {
+        for (String word : CLICKHOUSE_NONKEYWORDS) {
+            if (word.equalsIgnoreCase(str)) {
+                return true;
+            }
+        }
+        return super.mustBeQuoted(str, forceCaseSensitive);
     }
 }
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseSQLDialect.java
```diff
@@ -104,4 +109,10 @@ public String getColumnTypeModifiers(@NotNull DBPDataSource dataSource, @NotNull
     public boolean supportsNestedComments() {
         return true;
     }
+
+    //We should quote keywords which is not keywords for clickhouse, otherwise JSQLParser can't parse statements
+    @Override
+    public boolean mustBeQuoted(@NotNull String str, boolean forceCaseSensitive) {
+        return Arrays.stream(CLICKHOUSE_NONKEYWORDS).anyMatch(k -> k.equalsIgnoreCase(str)) || super.mustBeQuoted(str, forceCaseSensitive);
```
ShadelessFox: I don't like the capturing lambda here. Let's unwrap the stream into a plain loop.

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseSQLDialect.java
```diff
@@ -82,8 +86,9 @@ public boolean supportsOrderByIndex() {
 
     public void initDriverSettings(JDBCSession session, JDBCDataSource dataSource, JDBCDatabaseMetaData metaData) {
         super.initDriverSettings(session, dataSource, metaData);
-        removeSQLKeyword("DEFAULT");
-        removeSQLKeyword("SYSTEM");
+        for (int i = 0; i < CLICKHOUSE_NONKEYWORDS.length; i++) {
+            removeSQLKeyword(CLICKHOUSE_NONKEYWORDS[i]);
+        }
```
ShadelessFox: Let's use enhances for-loop here.

```suggestion
        for (String word : CLICKHOUSE_NONKEYWORDS) {
            removeSQLKeyword(word);
        }
```

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseSQLDialect.java
```diff
@@ -104,4 +109,17 @@ public String getColumnTypeModifiers(@NotNull DBPDataSource dataSource, @NotNull
...
     }
+
+    //We should quote keywords which is not keywords for clickhouse, otherwise JSQLParser can't parse statements
+    @Override
+    public boolean mustBeQuoted(@NotNull String str, boolean forceCaseSensitive) {
+        boolean isSpecialWord = false;
+        for (String word : CLICKHOUSE_NONKEYWORDS) {
+            if (word.equalsIgnoreCase(str)) {
+                isSpecialWord = true;
+                break;
```
ShadelessFox: Can't it be a short-circuit operation? It's `OR`-ed afterward anyway, so it could be a simple `return` statement.

```suggestion
                return true;
```

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseSQLDialect.java
```diff
@@ -104,4 +109,17 @@ public String getColumnTypeModifiers(@NotNull DBPDataSource dataSource, @NotNull
...
+            }
+        }
+        return isSpecialWord || super.mustBeQuoted(str, forceCaseSensitive);
```
ShadelessFox: And there it would become a simple delegate to the super method:

```suggestion
        return super.mustBeQuoted(str, forceCaseSensitive);
```

ShadelessFox (changes requested):

ShadelessFox (approved):

PR #15281, opened 2022-01-25
#10149 Associate stream managers with commands (to provide keybindings)

ShadelessFox:


ShadelessFox:
Perhaps `ParameterizedCommand` could be used instead of creating commands for every stream manager.

PR #15325, opened 2022-01-28
CB-1555 refactor query manager api

alexander-skoblikov:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/qm/QMDisposable.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.qm;
+
+
+public interface QMDisposable {
+    void dispose();
+}
```
ShadelessFox, line 22: Can we use `java.lang.AutoCloseable` instead?

Also, I don't see any actual usage of this interface. Is it really needed?

serge-rider: AutoClosable has a different idea (resource management + auto-close). Disposable objet can be disposed at any moment.

PR #15331, opened 2022-01-30
Speedup Oracle metadata load

skybber:
Original select from all_constrains is slow on schema with a lot of tables.
Patch uses select with left join equivalent to original one. It is 6x faster.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleSchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleSchema.java
@@ -910,25 +910,27 @@ public class OracleSchema extends OracleGlobalObject implements DBSSchema, DBPRe
             StringBuilder sql = new StringBuilder(500);
             JDBCPreparedStatement dbStat;
              if (owner.getDataSource().isAtLeastV11() && forTable != null && !useSimpleConnection) {
-                 sql.append("SELECT \r\n" 
-                         + "    c.TABLE_NAME,\r\n" 
+                 sql.append("SELECT \r\n"
+                         + "    c.TABLE_NAME,\r\n"
                          + "    c.CONSTRAINT_NAME,\r\n"
-                         + "    c.CONSTRAINT_TYPE,\r\n" 
-                         + "    c.STATUS,\r\n" 
+                         + "    c.CONSTRAINT_TYPE,\r\n"
+                         + "    c.STATUS,\r\n"
                          + "    c.R_OWNER,\r\n"
-                         + "    c.R_CONSTRAINT_NAME,\r\n" 
-                         + "    (SELECT rc.TABLE_NAME FROM "
-                         + OracleUtils.getAdminAllViewPrefix(session.getProgressMonitor(), getDataSource(),"CONSTRAINTS")
-                         + " rc WHERE rc.OWNER = c.r_OWNER AND rc.CONSTRAINT_NAME = c.R_CONSTRAINT_NAME) AS R_TABLE_NAME,\r\n"
-                         + "    c.DELETE_RULE,\r\n" 
+                         + "    c.R_CONSTRAINT_NAME,\r\n"
+                         + "    rc.TABLE_NAME AS R_TABLE_NAME,\r\n"
+                         + "    c.DELETE_RULE,\r\n"
                          + "    (\r\n"
-                         + "      SELECT LISTAGG(COLUMN_NAME || ':' || POSITION,',') WITHIN GROUP (ORDER BY \"POSITION\") \r\n" 
-                         + "      FROM ALL_CONS_COLUMNS col\r\n" 
+                         + "      SELECT LISTAGG(COLUMN_NAME || ':' || POSITION,',') WITHIN GROUP (ORDER BY \"POSITION\") \r\n"
+                         + "      FROM ALL_CONS_COLUMNS col\r\n"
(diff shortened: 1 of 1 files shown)


serge-rider:

skybber:
Do you need Issue number?

LonwoLonwo (approved):

skybber:
I have another modification, that improves the previous one 4x

p91paul:
something I see here is that you're using ALL_CONS_COLUMNS irrespective of the "use dba views" preference value. When the user has dba access using ALL views is usually much slower (it was that way even before this pull request).

skybber:
original one was:
```
(SELECT rc.TABLE_NAME FROM OracleUtils.getAdminAllViewPrefix(session.getProgressMonitor(), getDataSource(),"CONSTRAINTS")
rc WHERE rc.OWNER = c.r_OWNER AND rc.CONSTRAINT_NAME = c.R_CONSTRAINT_NAME) AS R_TABLE_NAME

```
new one:
```
LEFT JOIN
OracleUtils.getAdminAllViewPrefix(session.getProgressMonitor(), getDataSource(),"CONSTRAINTS") rc
ON rc.OWNER = ? AND rc.CONSTRAINT_NAME = c.R_CONSTRAINT_NAME
```

in the result query I see `all_constrains` table, not `all_cons_columns`

ALL_CONS_COLUMNS in select clausule is preserved in current `Select` statement in the original form

LonwoLonwo:
Hello @skybber 
Thanks for your contribution.

As @p91paul said, you can also change the ALL_CONS_COLUMNS table name to this java method -  
OracleUtils.getAdminAllViewPrefix(session.getProgressMonitor(), getDataSource(), "CONS_COLUMNS")
As is done below in the code of the prepareObjectsStatement method for other Oracle versions.

skybber:
OK, but `ALL_CONS_COLUMNS` was used in original code before the patch. I will push another improvement in the evenening since I don't have GH authorization code now. According my tests it improves metadata load from 70s to 2s in my case

skybber:
Done now.

p91paul:
@skybber yeah, my comment was not about your patch but a general suggestion after looking at this code.

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleSchema.java
```diff
@@ -929,7 +929,7 @@ protected JDBCStatement prepareObjectsStatement(JDBCSession session, OracleSchem
                          + " c\r\n"
                          + "LEFT JOIN " 
                          + OracleUtils.getAdminAllViewPrefix(session.getProgressMonitor(), getDataSource(),"CONSTRAINTS") + " rc\r\n"
-                         + "ON rc.OWNER = c.r_OWNER AND rc.CONSTRAINT_NAME = c.R_CONSTRAINT_NAME\r\n"
+                         + "ON rc.OWNER = ? AND rc.CONSTRAINT_NAME = c.R_CONSTRAINT_NAME\r\n"
```
p91paul: Be careful, this change is a change in semantics. You are effectively saying `rc.OWNER = c.OWNER` instead of `rc.OWNER = c.r_OWNER`. That is because you write `rc.OWNER = ?` and `c.OWNER = ?` and for both you set `OracleSchema.this.getName()` as a bind parameter.
If this is a referential constraint, the results will not be the same. So unless the previous code had a bug, you are introducing one.

skybber: It seems OK for me. Code before patch:
```
                         + "LEFT JOIN " 
                         + OracleUtils.getAdminAllViewPrefix(session.getProgressMonitor(), getDataSource(),"CONSTRAINTS") + " rc\r\n"
                         + "ON **rc.OWNER = c.r_OWNER** AND rc.CONSTRAINT_NAME = c.R_CONSTRAINT_NAME\r\n"
                         + "WHERE\r\n" + "    c.CONSTRAINT_TYPE = 'R'\r\n" + "    AND **c.OWNER = ?**\r\n"
                         + "    AND c.TABLE_NAME = ?");
```

after patch:
```
                         + "LEFT JOIN " 
                         + OracleUtils.getAdminAllViewPrefix(session.getProgressMonitor(), getDataSource(),"CONSTRAINTS") + " rc\r\n"
                         + "ON **rc.OWNER = ?** AND rc.CONSTRAINT_NAME = c.R_CONSTRAINT_NAME\r\n"
                         + "WHERE\r\n" + "    c.CONSTRAINT_TYPE = 'R'\r\n" + "    AND c.OWNER = ?\r\n"

```
=>  `c.owner=?` in the WHERE statement can be propagated to left join and `*rc.OWNER = c.r_OWNER` can be finally replaced by `rc.OWNER = ?`

skybber: I see it now...

skybber: There could be cross schema constraint in the Oracle, but then the second commit cannot be used and the query is 6-10x slower, but definitely faster than original one. Thanks for feedback.

skybber:
I hope it's complete now. It looks that additional restriction on non cross schema constraints would bring another substantial improvement.

serge-rider:
@skybber Thank you.
We'll test it with different Oracle server versions

Issue #15339, opened 2022-01-31
Support of MariaDB with MariaDB Connector/J 3.0.3


mieseprem:
#### System information: 
- Operating system (distribution) and version: Windows 10
- DBeaver version: 21.3.3
- Additional extensions

#### Connection specification:
- Database name and version: MariaDB 10.6.5
- **Driver name: mariadb-java-client 3.0.3 (MariaDB Connector/J)**
- Do you use tunnels or proxies (SSH, SOCKS, etc)?: yes

#### Describe the problem you're observing:

Due to change of behaviour regarding [jdbc:mariadb scheme](https://mariadb.com/kb/en/mariadb-connector-j-303-release-notes/#jdbcmariadb-scheme) as of version 3.0 of MariaDB Connector DBeaver is not able to connect to MariaDB.

Connector >= 3.0 require either `jdbc:mariadb:` or `jdbc:mysql:....?permitMysqlScheme:`

I think, the issue can be found here: https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/MySQLDataSourceProvider.java#L108
As there is no other Driver Type as "MySQL" this prefix is always taken.

Workaround is to remain on version 2.7.5 of connector.

LonwoLonwo:
Hello @mieseprem 

Thanks for the report.

LonwoLonwo:
Fixed in DBeaver 21.3.4

It will work fine for the new connections. If connection problems appear for previously created connections - please try to use "Test connection" in the connection settings. 

![2022-02-01 16_48_38-DBeaver Headless Test Application 1 0 7 - _demo2_ Script-35](https://user-images.githubusercontent.com/45152336/152040364-8226c499-dc47-424c-836e-1367f5607790.png)

uslss:
verified

PR #15342, opened 2022-01-31
#15339 fix URL connection string for the newest MariaDB connectors

LonwoLonwo:


serge-rider:
Please fix:
- Log error exception
- Append jdbc:mysql string only when it it is needed

Issue #1175, opened 2017-01-23
Generate DDL from view creates incomplete command (PostgreSQL)


elring:
Generate DDL from view creates just the select command without DDL part.
For example instead of "CREATE OR REPLACE VIEW test AS select id from table" it creates SQL with just "select id from table".

serge-rider:
Will be fixed in the next version.

PR #15360, opened 2022-02-01
#1175 Sequences fixes

LonwoLonwo:
2022-02-02 01_03_17-DBeaver Ultimate 21 3 3 - _sample fdb_Script-116

2022-02-02 01_52_49-DBeaver Ultimate 21 3 4 - SEQTEST


mkgrgis:
DB without sqlite_sequence is normal. It means there is no sequences. I think it's not exception. I have seen SQLiteman only displays no sequences.

LonwoLonwo:
I've added a special message for this exception about the sqlite_seqence table. So now it's a kinda message-holder :)

PR #15368, opened 2022-02-02
Speed up Oracle metadata load

skybber:
Another improvement of metadata query for all_constraints. It limits the query to constraints of primary key type only.


serge-rider:
But what about unique keys?

skybber:
That's true, correct query should be `rc.CONSTRAINT_TYPE in ('P','U')` but it does not help at all. I'll play with the select little bit and try to find solution.

skybber:
It should cover both cases - primary + unique keys now. In my case the speadup  is ~5x. First query is executed in 4s in comparision to the previous version where it took 20s

Otherwise `all_constraits` has ~75.000 rows in my case

serge-rider:
In same cases this query modification leads to even worse performance.

In fact we already had several metadata optimization options in the connection settings.
I've added one more: `ALTER SESSION SET OPTIMIZER_FEATURES_ENABLE='10.2.0.5'`

See #15538 for details

Thanks for your assistance and hints.

Issue #13885, opened 2021-09-13
Change default foreign key dictionary settings


fjf2002:
**Is your feature request related to a problem? Please describe.**
From https://dbeaver.io/forum/viewtopic.php?f=2&t=668 I learnt that 
* The foreign dictionary uses by default "any column which name contains one of next strings: title, name, label, display, description, comment or remark. If no such column found then first varchar column is used.". 
* As far as I understand, I can override this behaviour on a per-table basis. This can be tedious: I have many tables with similar columns, but the column names do not match the default strings as described above.

**Describe the solution you'd like**
I would be more productive if I could change the default strings as mentioned above, for example in window->preferences (instead of configuring it for each table separately).

ShadelessFox:
Implemented:

![image](https://user-images.githubusercontent.com/35821147/153191029-4d24afb0-e7c7-4993-aad9-34d15b180f8d.png)

This preferences page can also be accessed via link in the <kbd>Define Description</kbd> dialog.

uslss:
verified

fjf2002:
Thank You.

PR #15419, opened 2022-02-08
#13885 Configure default description column keywords in the preferences

ShadelessFox:


plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/object/struct/EditDictionaryPage.java
```diff
@@ -78,8 +81,13 @@ public DBVEntity getDictionary()
     @Override
     protected void createContentsBeforeColumns(Composite panel)
     {
-        Label label = UIUtils.createControlLabel(panel,
-            "Choose dictionary description columns or set custom criteria");
+        Link label = UIUtils.createLink(panel, EditorsMessages.dialog_struct_edit_dictionary_tip, new SelectionAdapter() {
+            @Override
+            public void widgetSelected(SelectionEvent e) {
+                // FIXME: Cannot use constant due to circular dependency.
+                UIUtils.showPreferencesFor(null, null, /*PrefPageDataViewer.PAGE_ID*/ "org.jkiss.dbeaver.preferences.main.dataviewer");
```
ShadelessFox, line 88: Not sure what we can do about it.

The only thing is to consider moving `PrefPageDataViewer` along with `TargetPrefPage` to another module.

Issue #1397, opened 2017-03-06
[Question] How to reinstall an accidentally uninstalled driver?


EpocDotFr:
I accidentallly uninstalled the provided SQLite driver. Unfortunately, I'm unable to find information about how to reinstall a driver that was provided natively by BDeaver.

Of course, I already tried to uninstall/reinstall DBeaver completely but whithout success.

Any help appreciated 😃

serge-rider:
There are two options:

1. Open driver manager, find SQLite driver, click Edit, click "Reset to Defaults" (original driver artifact must appear) then "Download/Update".
2. Delete file `~/.dbeaver4/.metadata/.plugins/org.jkiss.dbeaver.core/drivers.xml` (not recommended - this drops all drivers cache).

serge-rider:
3. If you deleted SQLite driver then you can just create new one.
Add jar files as Maven artifact `org.xerial:sqlite-jdbc:RELEASE`.
URL pattern is `jdbc:sqlite:{file}`, driver class is `org.sqlite.JDBC`.

EpocDotFr:
Thank you for your response Serge.

1. Cannot find the SQLite driver at all because I really uninstalled it :/
2. Already done but has no effect
3. It worked! Thank you 👌

jussihirvi:
A note to new users like me: I had a new installation and nothing to lose. So I got my deleted driver back by uninstalling dbeaver AND manually deleting AppData\Roaming\DBeaverData. Warning: this is of course drastic – it deletes all data.

PR #15448, opened 2022-02-10
#1397 Add host key verification for ssh

E1izabeth:
With a new ssh connection DBeaver will ask user to confirm and write key to ".ssh/known-hosts" file. For further connections it assumed as already confirmed while key verification with remembered key in "known-hosts" file pass. If key is not matched with key in "known-hosts" file than user will be alarmed and connection will be refused due to possible man-in-the-middle attack.

It was done for both sshj and jsch.

Also "known-hosts" is observable in eclipse preferences, but Eclipse itself does not support some host key types (like ed25519, see https://bugs.eclipse.org/bugs/show_bug.cgi?id=520927 and https://bugs.eclipse.org/bugs/show_bug.cgi?id=567109) so they are not displayed in the SSH2 known hosts preferences page.

### DIFF
--- a/plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/KnownHostsVerifier.java
+++ b/plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/KnownHostsVerifier.java
@@ -1,0 +1,91 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2022 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.net.ssh;
+
+import net.schmizz.sshj.common.KeyType;
+import net.schmizz.sshj.common.SecurityUtils;
+import net.schmizz.sshj.transport.verification.OpenSSHKnownHosts;
+import org.jkiss.dbeaver.runtime.ui.DBPPlatformUI;
+
+import java.io.File;
+import java.io.IOException;
+import java.security.PublicKey;
+
+public class KnownHostsVerifier extends OpenSSHKnownHosts {
+
+    private DBPPlatformUI platformUI;
+
+    public KnownHostsVerifier(File khFile, DBPPlatformUI platformUI) throws IOException {
+        super(khFile);
+        this.platformUI = platformUI;
+    }
+
+    protected boolean hostKeyUnverifiableAction(String hostname, PublicKey key) {
+        KeyType type = KeyType.fromKey(key);
+
+        String confirmationMessage = new StringBuilder()
+                .append("The authenticity of host can't be established.")
+                .append("host: '").append(hostname).append("'\n")
+                .append(type).append("key fingerprint: ").append(SecurityUtils.getFingerprint(key)).append("\n")
+                .append("Are you sure you want to continue connecting?").toString();
+
+        boolean isConfirmed = platformUI.confirmAction("Connection confirmation", confirmationMessage);
+
+        if(isConfirmed) {
+            try {
+                this.entries().add(new HostEntry((Marker)null, hostname, KeyType.fromKey(key), key));
+                this.write();
+                String warnMessage = new StringBuilder()
+                        .append("'").append(hostname).append("' (").append(type).append(") ")
+                        .append("permanently added to the list of known hosts.\n").toString();
+                platformUI.showWarningMessageBox("Warning",  warnMessage);
+                return true;
+            } catch (IOException e) {
+                throw new RuntimeException(e);
+            }
+        } else {
+            return false;
+        }
+    }
+
+    protected boolean hostKeyChangedAction(String hostname, PublicKey key) {
+        KeyType type = KeyType.fromKey(key);
+        String fp = SecurityUtils.getFingerprint(key);
+        String path = this.getFile().getAbsolutePath();
+        String warnMessage = new StringBuilder()
+                .append("Remote host identification has changed. It could be man-in-the-middle attack or the host key has just been changed \n")
+                .append("The fingerprint for the ").append(type).append("key sent by the remote host is ").append(fp).append("\n")
+                .append("Please contact your system administrator or add correct host key in ").append(path).append("to get rid of this message.")
+                .append("\n").toString();
+        platformUI.showWarningMessageBox("Warning",warnMessage);
+        return false;
+    }
+
+    @Override
+    public void write(KnownHostEntry entry) throws IOException {
+        this.khFile.getParentFile().mkdirs();
+        super.write(entry);
+        SSHUtils.forcePlatformReloadKnownHostsPreferences();
+    }
+
+    @Override
+    public void write() throws IOException {
+        this.khFile.getParentFile().mkdirs();
+        super.write();
+        SSHUtils.forcePlatformReloadKnownHostsPreferences();
+    }
+}
--- a/plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/KnownHostsVerifier.java
+++ b/plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/KnownHostsVerifier.java
@@ -19,6 +19,7 @@ package org.jkiss.dbeaver.model.net.ssh;
 import net.schmizz.sshj.common.KeyType;
 import net.schmizz.sshj.common.SecurityUtils;
 import net.schmizz.sshj.transport.verification.OpenSSHKnownHosts;
+import org.jkiss.code.NotNull;
 import org.jkiss.dbeaver.runtime.ui.DBPPlatformUI;
 
 import java.io.File;
@@ -29,11 +30,12 @@ public class KnownHostsVerifier extends OpenSSHKnownHosts {
 
     private DBPPlatformUI platformUI;
 
-    public KnownHostsVerifier(File khFile, DBPPlatformUI platformUI) throws IOException {
+    public KnownHostsVerifier(@NotNull File khFile, @NotNull DBPPlatformUI platformUI) throws IOException {
         super(khFile);
         this.platformUI = platformUI;
     }
 
+    @Override
     protected boolean hostKeyUnverifiableAction(String hostname, PublicKey key) {
         KeyType type = KeyType.fromKey(key);
 
@@ -62,6 +64,7 @@ public class KnownHostsVerifier extends OpenSSHKnownHosts {
         }
     }
 
+    @Override
     protected boolean hostKeyChangedAction(String hostname, PublicKey key) {
         KeyType type = KeyType.fromKey(key);
         String fp = SecurityUtils.getFingerprint(key);
--- a/plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
+++ b/plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
@@ -29,7 +29,6 @@ import org.jkiss.dbeaver.model.preferences.DBPPreferenceStore;
 import org.jkiss.dbeaver.utils.RuntimeUtils;
 import org.jkiss.utils.CommonUtils;
 import org.jkiss.utils.IOUtils;
-import org.jkiss.utils.StandardConstants;
 
 import java.io.File;
 import java.nio.file.InvalidPathException;
@@ -43,6 +42,12 @@ import java.util.List;
 public class SSHUtils {
 
     private static final Log log = Log.getLog(SSHUtils.class);
+
+    private static final String PLATFORM_SSH_PREFERENCES_NODE = "org.eclipse.jsch.core"; //$NON-NLS-1$
+    private static final String PLATFORM_SSH_PREFERENCES_SSH2HOME_KEY = IConstants.KEY_SSH2HOME;
+    private static final String DEFAULT_SSH_HOME_DIR_NAME = IConstants.SSH_DEFAULT_HOME;
+    private static final String DEFAULT_SSH_HOME_DIR_NAME_WIN_OLD = IConstants.SSH_OLD_DEFAULT_WIN32_HOME;
+    private static final String KNOWN_SSH_HOSTS_FILE_NAME = "known_hosts";
 
     static int findFreePort(DBPPlatform platform)
     {
@@ -90,11 +95,6 @@ public class SSHUtils {
         return false;
     }
 
-    private static final String PLATFORM_SSH_PREFERENCES_NODE = "org.eclipse.jsch.core"; //$NON-NLS-1$
-    private static final String PLATFORM_SSH_PREFERENCES_SSH2HOME_KEY = IConstants.KEY_SSH2HOME;
-    private static final String DEFAULT_SSH_HOME_DIR_NAME = IConstants.SSH_DEFAULT_HOME;
-    private static final String DEFAULT_SSH_HOME_DIR_NAME_WIN_OLD = IConstants.SSH_OLD_DEFAULT_WIN32_HOME;
-    private static final String KNOWN_SSH_HOSTS_FILE_NAME = "known_hosts";
 
     @NotNull
     public static File getKnownSshHostsFileOrDefault() {
@@ -115,7 +115,7 @@ public class SSHUtils {
                 log.warn("Failed to resolve SSH known hosts file location at " + sshHomePathString, e);
                 if (forceFileObjectOnFail) {
                     return new File(sshHomePathString + File.pathSeparator + KNOWN_SSH_HOSTS_FILE_NAME);
-                } else{
+                } else {
                     return null;
                 }
             }
@@ -148,12 +148,11 @@ public class SSHUtils {
 
                         if (updatePreferences) {
                             Platform.getPreferencesService()
-                                    .getRootNode().node(PLATFORM_SSH_PREFERENCES_NODE)
-                                    .put(PLATFORM_SSH_PREFERENCES_SSH2HOME_KEY, sshHomeDir.getAbsolutePath());
+                                .getRootNode().node(PLATFORM_SSH_PREFERENCES_NODE)
+                                .put(PLATFORM_SSH_PREFERENCES_SSH2HOME_KEY, sshHomeDir.getAbsolutePath());
                         }
 
-                        Path knownSshHostsFilePath = sshHomeDirPath.resolve(KNOWN_SSH_HOSTS_FILE_NAME);
-                        return knownSshHostsFilePath.toFile();
+                        return sshHomeDirPath.resolve(KNOWN_SSH_HOSTS_FILE_NAME).toFile();
                     } else {
                         log.warn("Failed to resolve default SSH known hosts file location due to invalid SSH home directory " + sshHomeDirPath.toAbsolutePath());
                     }
@@ -174,6 +173,7 @@ public class SSHUtils {
             return null;
         }
     }
+
     public static void forcePlatformReloadKnownHostsPreferences() {
         JSchCorePlugin.getPlugin().setNeedToLoadKnownHosts(true);
         JSchCorePlugin.getPlugin().getJSch().setHostKeyRepository(null);
--- a/plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
+++ b/plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
@@ -26,6 +26,7 @@ import org.jkiss.dbeaver.Log;
 import org.jkiss.dbeaver.ModelPreferences;
 import org.jkiss.dbeaver.model.app.DBPPlatform;
 import org.jkiss.dbeaver.model.preferences.DBPPreferenceStore;
+import org.jkiss.dbeaver.utils.RuntimeUtils;
 import org.jkiss.utils.CommonUtils;
 import org.jkiss.utils.IOUtils;
 import org.jkiss.utils.StandardConstants;
@@ -124,11 +125,6 @@ public class SSHUtils {
         }
     }
 
-    private static boolean isOnWindows() {
-        String osName = System.getProperty(StandardConstants.ENV_OS_NAME);
-        return osName != null && osName.startsWith("Windows");
-    }
-
     private static File resolveDefaultKnownSshHostsFile(boolean forceFileObjectOnFail, boolean updatePreferences) {
         try {
             String userHomePathString = System.getProperty(IConstants.SYSTEM_PROPERTY_USER_HOME);
@@ -138,7 +134,7 @@ public class SSHUtils {
                     Path sshHomeDirPath = userHomeDirPath.resolve(DEFAULT_SSH_HOME_DIR_NAME);
                     File sshHomeDir = sshHomeDirPath.toFile();
 
-                    if (isOnWindows() && (!sshHomeDir.isDirectory() || !sshHomeDir.exists())) {
+                    if (RuntimeUtils.isWindows() && (!sshHomeDir.isDirectory() || !sshHomeDir.exists())) {
                         Path sshHomeOldDirPath = userHomeDirPath.resolve(DEFAULT_SSH_HOME_DIR_NAME_WIN_OLD);
                         File sshHomeOldDir = sshHomeDirPath.toFile();
                         if (sshHomeOldDir.isDirectory()) {
@@ -172,7 +168,7 @@ public class SSHUtils {
         }
 
         if (forceFileObjectOnFail) {
-            Path forcedUserProfilePath = Paths.get(isOnWindows() ? "%USERPROFILE%" : "~"); // let's pretend it'll resolve itself
+            Path forcedUserProfilePath = Paths.get(RuntimeUtils.isWindows() ? "%USERPROFILE%" : "~"); // let's pretend it'll resolve itself
             return forcedUserProfilePath.resolve(DEFAULT_SSH_HOME_DIR_NAME).resolve(KNOWN_SSH_HOSTS_FILE_NAME).toFile();
         } else {
             return null;
--- a/plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
+++ b/plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
@@ -17,18 +17,29 @@
 package org.jkiss.dbeaver.model.net.ssh;
 
 import com.jcraft.jsch.*;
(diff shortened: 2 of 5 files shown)


plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +89,98 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+                    return null;
+                }
+            }
+        } else {
+            // seems preference path not set at all, so try to resolve it and preserve
+            return resolveDefaultKnownSshHostsFile(forceFileObjectOnFail, true);
+        }
+    }
+
+    private static boolean isOnWindows() {
```
LonwoLonwo: Maybe `RuntimeUtils.isWindows()` can be beneficial?

E1izabeth: Done

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,93 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+                        }
+
+                        Path knownSshHostsFilePath = sshHomeDirPath.resolve(KNOWN_SSH_HOSTS_FILE_NAME);
+                        return knownSshHostsFilePath.toFile();
+                    } else {
+                        log.warn("Failed to resolve default SSH known hosts file location due to invalid SSH home directory " + sshHomeDirPath.toAbsolutePath());
+                    }
+                } else {
+                    log.warn("Failed to resolve default SSH known hosts file location due to missing user home directory " + userHomeDirPath.toAbsolutePath());
+                }
```
Destrolaric: I think this could be made with one conditional to avoid code duplication

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,93 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+    private static File getKnownSshHostsFileImpl(boolean forceFileObjectOnFail) {
+        String sshHomePathString = Platform.getPreferencesService().getString(PLATFORM_SSH_PREFERENCES_NODE, PLATFORM_SSH_PREFERENCES_SSH2HOME_KEY, null, null);
+        if (CommonUtils.isNotEmpty(sshHomePathString)) {
+            try {
+                return Paths.get(sshHomePathString, KNOWN_SSH_HOSTS_FILE_NAME).toFile();
+            } catch (InvalidPathException e) {
+                log.warn("Failed to resolve SSH known hosts file location at " + sshHomePathString, e);
+                if (forceFileObjectOnFail) {
+                    return new File(sshHomePathString + File.pathSeparator + KNOWN_SSH_HOSTS_FILE_NAME);
+                } else{
```
Destrolaric: code style

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,93 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+        }
+
+        if (forceFileObjectOnFail) {
+            Path forcedUserProfilePath = Paths.get(RuntimeUtils.isWindows() ? "%USERPROFILE%" : "~"); // let's pretend it'll resolve itself
+            return forcedUserProfilePath.resolve(DEFAULT_SSH_HOME_DIR_NAME).resolve(KNOWN_SSH_HOSTS_FILE_NAME).toFile();
+        } else {
+            return null;
+        }
+    }
+    public static void forcePlatformReloadKnownHostsPreferences() {
```
Destrolaric: code style

Destrolaric (changes requested):

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: It can't be made with one conditional.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: fixed

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: fixed

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,94 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+                        }
+                    }
+
+                    if (sshHomeDir.isDirectory() || !sshHomeDir.exists()) {
+                        // don't need to create it until we'll need to write known hosts file
+
+                        if (updatePreferences) {
+                            Platform.getPreferencesService()
+                                    .getRootNode().node(PLATFORM_SSH_PREFERENCES_NODE)
+                                    .put(PLATFORM_SSH_PREFERENCES_SSH2HOME_KEY, sshHomeDir.getAbsolutePath());
```
ShadelessFox: We use 4 spaces for continuation indentation.

```suggestion
                                .getRootNode().node(PLATFORM_SSH_PREFERENCES_NODE)
                                .put(PLATFORM_SSH_PREFERENCES_SSH2HOME_KEY, sshHomeDir.getAbsolutePath());
```

In case of Intellij IDEA, code style can be adjusted in <kbd>Editor</kbd> &rArr; <kbd>Code Style</kbd> &rArr; <kbd>Java</kbd> &rArr; <kbd>Continuation indent</kbd>.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,94 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+                        }
+
+                        Path knownSshHostsFilePath = sshHomeDirPath.resolve(KNOWN_SSH_HOSTS_FILE_NAME);
+                        return knownSshHostsFilePath.toFile();
```
ShadelessFox: This variable can be inlined.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,94 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+                }
+            } else {
+                log.warn("Failed to resolve default SSH known hosts file location due to missing user home system property.");
+            }
+        } catch (Throwable e) {
+            log.warn("Failed to resolve default SSH known hosts file location.", e);
+        }
+
+        if (forceFileObjectOnFail) {
+            Path forcedUserProfilePath = Paths.get(RuntimeUtils.isWindows() ? "%USERPROFILE%" : "~"); // let's pretend it'll resolve itself
```
ShadelessFox, line 170: Please use `System.getProperty(StandardConstants#ENV_USER_HOME)` instead of this dirty hack.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,94 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+            // seems preference path not set at all, so try to resolve it and preserve
+            return resolveDefaultKnownSshHostsFile(forceFileObjectOnFail, true);
+        }
+    }
+
+    private static File resolveDefaultKnownSshHostsFile(boolean forceFileObjectOnFail, boolean updatePreferences) {
+        try {
+            String userHomePathString = System.getProperty(IConstants.SYSTEM_PROPERTY_USER_HOME);
+            if (userHomePathString != null) {
+                Path userHomeDirPath = Paths.get(userHomePathString);
```
ShadelessFox, line 132: Can we use `java.io.File` in the first place to avoid countless occurrences of `toFile`?

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,94 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+                if (userHomeDirPath.toFile().isDirectory()) {
```
ShadelessFox, line 133: Just a tip: `Files#isDirectory`. Please see my previous comment.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,94 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+            return forcedUserProfilePath.resolve(DEFAULT_SSH_HOME_DIR_NAME).resolve(KNOWN_SSH_HOSTS_FILE_NAME).toFile();
+        } else {
+            return null;
+        }
+    }
+
+    public static void forcePlatformReloadKnownHostsPreferences() {
+        JSchCorePlugin.getPlugin().setNeedToLoadKnownHosts(true);
+        JSchCorePlugin.getPlugin().getJSch().setHostKeyRepository(null);
+    }
```
ShadelessFox, line 180: Why do we even touch the platform JSCH plugin?

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,94 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+    public static File getKnownSshHostsFileOrNull() {
+        return  getKnownSshHostsFileImpl(false);
+    }
+
+    private static File getKnownSshHostsFileImpl(boolean forceFileObjectOnFail) {
+        String sshHomePathString = Platform.getPreferencesService().getString(PLATFORM_SSH_PREFERENCES_NODE, PLATFORM_SSH_PREFERENCES_SSH2HOME_KEY, null, null);
+        if (CommonUtils.isNotEmpty(sshHomePathString)) {
+            try {
+                return Paths.get(sshHomePathString, KNOWN_SSH_HOSTS_FILE_NAME).toFile();
+            } catch (InvalidPathException e) {
```
ShadelessFox, line 114: In which cases may this exception may be thrown? I doubt the path can be malformed.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,94 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+    public static File getKnownSshHostsFileOrDefault() {
+        return  getKnownSshHostsFileImpl(true);
+    }
+
+    @Nullable
+    public static File getKnownSshHostsFileOrNull() {
+        return  getKnownSshHostsFileImpl(false);
+    }
+
+    private static File getKnownSshHostsFileImpl(boolean forceFileObjectOnFail) {
```
ShadelessFox, line 109: Please annotate even non-public API methods with `@NotNull`/`@Nullable` annotations.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,94 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+                            sshHomeDirPath = sshHomeOldDirPath;
+                            sshHomeDir = sshHomeOldDir;
+                        }
+                    }
+
+                    if (sshHomeDir.isDirectory() || !sshHomeDir.exists()) {
+                        // don't need to create it until we'll need to write known hosts file
+
+                        if (updatePreferences) {
+                            Platform.getPreferencesService()
```
ShadelessFox, line 150: Please use our preference store. See `DBPPlatform#getPreferenceStore`

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -78,4 +90,94 @@ public static boolean isKeyEncrypted(byte[] privKeyValue) {
...
+                        return knownSshHostsFilePath.toFile();
+                    } else {
+                        log.warn("Failed to resolve default SSH known hosts file location due to invalid SSH home directory " + sshHomeDirPath.toAbsolutePath());
+                    }
+                } else {
+                    log.warn("Failed to resolve default SSH known hosts file location due to missing user home directory " + userHomeDirPath.toAbsolutePath());
+                }
+            } else {
+                log.warn("Failed to resolve default SSH known hosts file location due to missing user home system property.");
+            }
```
ShadelessFox, line 164: I think these duplicated logging lines can be moved just above line 170. If we get so far, there will be an error anyways.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: The point here is not to get null and pass any string to library to let it resolve this situation itself in case of failure. For example, if user has no user home we don't want to have NPE, we want to get some error with meaningfull message from ssh library.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: Because we need to update preferences page to reload content of known-hosts file. It doesn't reload it on content changes automatically.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: I suppose it's important to have meaningful error message which is different for these situations.
Anyway the point that we can don't get an error here, but something goes wrong. The error can be further in usage, but It can be avoided if user reads logs. So, I'm not sure that it's good to replace different messages with one not straight pointing to the problem.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: ok, thanks

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
ShadelessFox: It's not possible to not have the user home directory.

E1izabeth: I'm not so sure about it, but okay, I'll believe you.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
ShadelessFox: I still don't get it. Is it somehow related to `PLATFORM_SSH_PREFERENCES_NODE`?

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
ShadelessFox: I haven't noticed that they aren't the same. My bad.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
ShadelessFox: It's okay to use this API to obtain some of the internal properties set by the JSch plugin in Eclipse.

I might have misunderstood this part.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: The nullability of these two methods depends on argument. Which of these annotations should I use in this case?

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: Okay, thanks

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
ShadelessFox: Maybe the caller should handle exceptional situations themselves?

Let's not pass the `forceFileObjectOnFail` parameter and let the caller decide how to handle this situation.

So it would become `@Nullable`.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: Yes, it's related. It's about eclipse jsch plugin. 
We take preferences about user ssh dir and content of known-hosts file from it and when known-hosts file changed we want to see updated list of records in preferences page. We need to force reload it due to eclipse doesn't reread file if there is changes only in it's content.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: The reason for this parameter is that we have two ssh libraries and one of them can handle this situation, but another one is not. So we might just get NPE here instead of meaningful error message, which is bad.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: For example if something happened with file system or user changed preference not according to paths in it's file system (accidentally or intentionally but it leads to wrong path).

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: Well, even if it's not possible not to have user directory at all, user can accidentally remove user home system property. Then we will fail on System.getProperty(StandardConstants.ENV_USER_HOME). It will return null, we will get NPE instead of error message from the library, which we want to reach.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: It's again about accidentally nonexistence of user home property. I'm not sure that it's okay to assume that everything is normal. In my point of view it's better to provide handling for such cases.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: done

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
ShadelessFox: It's a very exceptional situation. Let's not consider it.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
ShadelessFox: Makes sense.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
ShadelessFox: It's unlikely that the filesystem may cause the path to become malformed.

E1izabeth: Sure, but the user can easily write wrong path to text area.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
E1izabeth: I disagree that we should not consider it, because it can happened.

plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/KnownHostsVerifier.java
```diff
@@ -0,0 +1,94 @@
...
+
+        String confirmationMessage = new StringBuilder()
+                .append("The authenticity of host can't be established.")
+                .append("host: '").append(hostname).append("'\n")
+                .append(type).append("key fingerprint: ").append(SecurityUtils.getFingerprint(key)).append("\n")
+                .append("Are you sure you want to continue connecting?").toString();
+
+        boolean isConfirmed = platformUI.confirmAction("Connection confirmation", confirmationMessage);
+
+        if(isConfirmed) {
```
ShadelessFox, line 50: This condition could be flipped to perform an early return.

plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/KnownHostsVerifier.java
```diff
@@ -0,0 +1,94 @@
...
+    public KnownHostsVerifier(@NotNull File khFile, @NotNull DBPPlatformUI platformUI) throws IOException {
+        super(khFile);
+        this.platformUI = platformUI;
+    }
+
+    @Override
+    protected boolean hostKeyUnverifiableAction(String hostname, PublicKey key) {
+        KeyType type = KeyType.fromKey(key);
+
+        String confirmationMessage = new StringBuilder()
```
flambchap, line 42: i10n maybe not a bad idea

plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/KnownHostsVerifier.java
```diff
@@ -0,0 +1,94 @@
...
+        boolean isConfirmed = platformUI.confirmAction("Connection confirmation", confirmationMessage);
+
+        if(isConfirmed) {
+            try {
+                this.entries().add(new HostEntry((Marker)null, hostname, KeyType.fromKey(key), key));
+                this.write();
+                String warnMessage = new StringBuilder()
+                        .append("'").append(hostname).append("' (").append(type).append(") ")
+                        .append("permanently added to the list of known hosts.\n").toString();
+                platformUI.showWarningMessageBox("Warning",  warnMessage);
```
akilovich, line 57: i10n maybe not a bad idea

plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/KnownHostsVerifier.java
```diff
@@ -0,0 +1,94 @@
...
+
+        boolean isConfirmed = platformUI.confirmAction("Connection confirmation", confirmationMessage);
+
+        if(isConfirmed) {
+            try {
+                this.entries().add(new HostEntry((Marker)null, hostname, KeyType.fromKey(key), key));
+                this.write();
+                String warnMessage = new StringBuilder()
+                        .append("'").append(hostname).append("' (").append(type).append(") ")
+                        .append("permanently added to the list of known hosts.\n").toString();
```
flambchap, line 56: i10n maybe not a bad idea

plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/KnownHostsVerifier.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    protected boolean hostKeyChangedAction(String hostname, PublicKey key) {
+        KeyType type = KeyType.fromKey(key);
+        String fp = SecurityUtils.getFingerprint(key);
+        String path = this.getFile().getAbsolutePath();
+        String warnMessage = new StringBuilder()
+                .append("Remote host identification has changed. It could be man-in-the-middle attack or the host key has just been changed \n")
+                .append("The fingerprint for the ").append(type).append("key sent by the remote host is ").append(fp).append("\n")
+                .append("Please contact your system administrator or add correct host key in ").append(path).append("to get rid of this message.")
+                .append("\n").toString();
```
flambchap, line 76: i10n maybe not a bad idea

PR #15462, opened 2022-02-10
#12339 fix Azure Synapse data type query

LonwoLonwo:
No test environment. So this is the blind fix. I hope it helps.


LonwoLonwo:
https://docs.microsoft.com/en-us/sql/t-sql/functions/serverproperty-transact-sql?view=sql-server-ver15

Issue #14433, opened 2021-10-29
[Clickhouse] Table size is lost in navigator after refresh


Matvey16:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 21.2.4

#### Connection specification:
- Database name and version Clickhouse 20.4.4.18
- Driver name ru.yandex.clickhouse-jdbc 0.2.6
- Do you use tunnels or proxies (SSH, SOCKS, etc)? SSH tunnel

#### Steps to reproduce, if exist:
1. Connect to clickhouse database
2. Expand tables folder in navigator and observe tables' sizes
3. Click refresh on schema or tables folder

**Expected:** Size statistic is refreshed
**Actual:** Size statistic disappear

![Hc2Pss6qx9](https://user-images.githubusercontent.com/82543000/139390006-6edf4834-a2ad-4dcc-9d77-94bb18071c78.gif)

Destrolaric:
Reproducible in Oracle and Dynamo.

https://user-images.githubusercontent.com/20579256/153021257-d8c4f817-2965-4f4e-b601-582d5ef98d58.mp4

uslss:
still reproducible in case table is refreshed
![dbeaver_4YHaFxrFqP](https://user-images.githubusercontent.com/31996417/153564430-49cbcfdb-f16f-428b-9264-5ecf014b0db3.gif)

Destrolaric:
Implemented fix for Redshift, Informix, Oracle, Clickhouse, Dynamo, Vertica.

uslss:
Vertica, Informix:
![dbeaver_JjvOAwZbCl](https://user-images.githubusercontent.com/31996417/154225206-a99491eb-73bf-4e73-9cd2-d752b8129db4.gif)
DynamoDB:
![dbeaver_i1pwa0SDGC](https://user-images.githubusercontent.com/31996417/154225297-45b1336b-2e5c-418d-9097-b2f8e71f5c61.gif)

PR #15465, opened 2022-02-11
#14433 fix tables refresh

Destrolaric:


plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseSchema.java
```diff
@@ -65,10 +65,14 @@ public boolean isStatisticsCollected() {
...
     @Override
     public synchronized DBSObject refreshObject(@NotNull DBRProgressMonitor monitor) throws DBException {
-        hasStatistics = false;
+        resetStatistics();
         return super.refreshObject(monitor);
     }
 
+    void resetStatistics() {
+        hasStatistics = false;
+    }
```
ShadelessFox, line 74: Can we move this method to the `DBPObjectStatisticsCollector` interface?

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseTable.java
```diff
@@ -147,6 +150,12 @@ public String generateTableUpdateBegin(String tableName) {
         return "ALTER TABLE " + tableName + " UPDATE ";
     }
 
+    @Override
+    public DBSObject refreshObject(@NotNull DBRProgressMonitor monitor) throws DBException {
+        ((ClickhouseSchema) getSchema()).resetStatistics();
```
ShadelessFox: Please see the previous note first. By doing so, we can move this logic to the very base class.

ShadelessFox (changes requested):

Issue #15467, opened 2022-02-11
regression introduced with: d0740bad01441fe6d0ceb7079bfa0e48fac06d42 (#13916 Add big query catalog change limitation (#15362))


kai-morich:
while trying to add a HANA feature, based on latest devel branch, I got this error while executing statements:

      class org.jkiss.dbeaver.ext.hana.model.HANADataSourceInfo cannot be cast to class org.jkiss.dbeaver.ext.generic.model.GenericDataSourceInfo (org.jkiss.dbeaver.ext.hana.model.HANADataSourceInfo is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @23986b98; org.jkiss.dbeaver.ext.generic.model.GenericDataSourceInfo is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @761334e0)
      
with d0740bad01441fe6d0ceb7079bfa0e48fac06d42 a cast was added to GenericExecutionContext.java:

    public boolean supportsSchemaChange() {
        GenericDataSource dataSource = getDataSource();
        final GenericDataSourceInfo info = (GenericDataSourceInfo) dataSource.getInfo(); <<<

but not all plugins derive from GenericDataSourceInfo, some derive from JDBCDataSourceInfo

serge-rider:
Thanks for the bug report and for the investigation 👍

uslss:
verified

PR #15478, opened 2022-02-13
#15467 fix unchecked GenericDataSourceInfo cast

Destrolaric:


ShadelessFox (changes requested):
Please use the early return instead of surrounding the entire code block with the `if` check.

plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericExecutionContext.java
```diff
@@ -215,30 +215,34 @@ public GenericSchema getDefaultSchema() {
...
     public boolean supportsSchemaChange() {
         GenericDataSource dataSource = getDataSource();
         final GenericDataSourceInfo info = (GenericDataSourceInfo) dataSource.getInfo();
-        if (dataSource.isSelectedEntityFromAPI() || !CommonUtils.isEmpty(dataSource.getQuerySetActiveDB())) {
-            if (CommonUtils.isEmpty(dataSource.getSelectedEntityType())) {
-                return !dataSource.hasCatalogs() && dataSource.hasSchemas() && info.supportsSchemaSelection();
-            }
-            if (dataSource.hasSchemas()) {
-                return (GenericConstants.ENTITY_TYPE_SCHEMA.equals(dataSource.getSelectedEntityType()) || !dataSource.hasCatalogs()) && info.supportsSchemaSelection();
+        if (dataSource.getInfo() instanceof GenericDataSourceInfo) {
```
LonwoLonwo: Wrong Line! The cast is on the line above.

LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericExecutionContext.java
Destrolaric: already fixed, in the last commit

Issue #12148, opened 2021-04-19
Display array elements in the spreadsheet


abolognino:
#### System information: 
- Operating system (distribution) and version: Ubuntu 20.04.2 LTS
- DBeaver version: 21.0.3
- Additional extensions

#### Connection specification:
- Database name and version: Oracle 19
- Driver name: ojdbc8-12.2.0.1
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No

#### Describe the problem you're observing:
when selecting a row with a collection type inside, only the first element of
the collection is shown

#### Steps to reproduce, if exist:
    create type tst_kv as object (k varchar2(100), v varchar2(100));

    create type tst_kv_coll is table of tst_kv;

    create table tst_tab (id number, coll tst_kv_coll)
    NESTED TABLE COLL STORE AS "TST_TAB_NST1";

    insert into tst_tab(id, coll)
    select 1, tst_kv_coll(tst_kv('0', 'a')) x from dual;

    commit;

    /* REFRESH THE SCHEMA DEFINITION TO AVOID ANOTHER UNRELATED BUG */

    declare
        v_row tst_tab%rowtype;
    begin
        select * into v_row
        from tst_tab
        where id = 1;
        v_row.coll.extend();
        v_row.coll(2) := tst_kv('2', 'b');
        update tst_tab set row = v_row where id = 1;
        commit;
    end;

    select * from tst_tab;

    /* NOTICE ONLY the ('0', 'a') record is shown, sqlplus shows both ('0', 'a') and ('2', 'b') */

    -- cleanup
    drop type tst_kv_coll;
    drop type tst_kv;
    drop table tst_tab;

![image](https://user-images.githubusercontent.com/14804609/115208000-6b758a00-a0fc-11eb-9e85-cd3de9057d1c.png)
![image](https://user-images.githubusercontent.com/14804609/115208059-7a5c3c80-a0fc-11eb-9fc5-18e8fcccac15.png)

#### Include any warning/errors/backtraces from the logs
no errors

LonwoLonwo:
Hello @abolognino 
Thanks for the bug report.

As some workaround - you can see right values in "Text" mode.
![2021-04-20 10_26_58-DBeaver Enterprise 21 0 3 - _orcl_ Script-42](https://user-images.githubusercontent.com/45152336/115355528-6d565080-a1c3-11eb-95c7-c8e572604f51.png)

emironovDB:
Hi, @abolognino! 

[...]

PR #15486, opened 2022-02-14
#12148 Show collection elements in the RSV

ShadelessFox:
- [ ] Fix highlighting. The presentation does not know anything about nested rows. Need to introduce the `depth` property. Still needs to be integrated with `LightGrid` which we want to avoid.
- [ ] Fix value panel. We don't extract the proper attribute from the selected cell.
- [ ] Fix value editing. Doesn't work for the same reason as described above
- [ ] Cells that don't represent collection elements must not be editable
- [ ] Make sure data transfer doesn't care about nested rows
- [ ] Make sure copy/paste of cell values still works with nested rows

WIP:
image


mkgrgis:
RSV will really useful for arrays, especially in PostgreSQL.

PR #15492, opened 2022-02-14
#15069 Force always patch query with limit for clickhouse

E1izabeth:


zhicwu:
Hi @E1izabeth, could you elaborate why patching query with limit? Let me know if you think `setMaxRow` did not work and I can help to fix it in the driver.

As to index, actually projection will show up as "index". I was not able to find index metadata from system tables so I only added projections. Let me know if you have better ideas and I can make changes accordingly as well.

E1izabeth:
Hi, @zhicwu! We encountered such error
`_SQL Error [1002]: ClickHouse exception, code: 1002, host: ***, port: ***; Code: 164. DB::Exception: Cannot modify 'max_result_rows' setting in readonly mode. (READONLY)`
using `setMaxRow` for user having readonly permissions for queries. In general case `setMaxRow` works fine, thanks.

LonwoLonwo:
Hello @zhicwu 

We removed "Indexes" tab some times ago. Does Clickhouse supports indexes now? I the new driver?

zhicwu:
Thanks @E1izabeth for the details. I think we probably should set readonly to `2` instead of `1` according to doc at [here](https://clickhouse.com/docs/en/operations/settings/permissions-for-queries/#settings_readonly):

```
Possible values:

    0 — All queries are allowed.
    1 — Only read data queries are allowed.
    2 — Read data and change settings queries are allowed.
```

I'm going to release another "patch" with ClickHouse/clickhouse-jdbc#859 fixed in the coming weekend. Will let you know how it goes. In worst case, it's probably better to patch the query with limit in the driver.


> We removed "Indexes" tab some times ago. Does Clickhouse supports indexes now? I the new driver?

Hi @LonwoLonwo, ClickHouse does support data skip index, for example:

```sql
DROP TABLE IF EXISTS temperature_sensors;
CREATE TABLE temperature_sensors(
    sensor_id Int32,
    datetime DateTime,
    value Decimal(5,2)
) Engine = MergeTree PARTITION BY toYYYYMM(datetime) ORDER BY (sensor_id, datetime);
ALTER TABLE test.temperature_sensors ADD INDEX temp_idx value TYPE minmax GRANULARITY 1;
```

Besides, [projections](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/mergetree/#projections) are supported as well and the feature is production ready since 22.2.

I didn't find a way to retrieve table index(besides parsing table_ddl), so I only considered projections in the driver. Let me know if you have better idea and we can discuss perhaps in a separate thread.

zhicwu:
@E1izabeth, FYI, just released v0.3.2-patch7 which supports read only mode. As long as `readonly` is set to 2, which is the default behavior of Connection.setReadOnly(true), setting max_result_rows will work.

E1izabeth:
@zhicwu, thank you!

PR #15506, opened 2022-02-16
dbeaver/dbeaver-ee#1389 add support of parent local clients

LonwoLonwo:


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceProviderDescriptor.java
```diff
@@ -174,9 +174,18 @@ public DataSourceProviderDescriptor(DataSourceProviderRegistry registry, IConfig
 
         // Load native clients
         {
-            for (IConfigurationElement nativeClientsElement : config.getChildren("nativeClients")) {
-                for (IConfigurationElement clientElement : nativeClientsElement.getChildren("client")) {
-                    this.nativeClients.add(new NativeClientDescriptor(clientElement));
+            IConfigurationElement[] nativeClients = config.getChildren("nativeClients");
+            if (ArrayUtils.isEmpty(nativeClients)) {
+                String inheritClients = config.getAttribute("inheritClients");
+                if (CommonUtils.isNotEmpty(inheritClients) && inheritClients.equals("true") && parentProvider != null) {
```
ShadelessFox: Instead of an equality check for a boolean, you can use `CommonUtils#toBoolean`.

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceProviderDescriptor.java
```diff
@@ -174,9 +174,18 @@ public DataSourceProviderDescriptor(DataSourceProviderRegistry registry, IConfig
 
         // Load native clients
         {
-            for (IConfigurationElement nativeClientsElement : config.getChildren("nativeClients")) {
-                for (IConfigurationElement clientElement : nativeClientsElement.getChildren("client")) {
-                    this.nativeClients.add(new NativeClientDescriptor(clientElement));
+            IConfigurationElement[] nativeClients = config.getChildren("nativeClients");
+            if (ArrayUtils.isEmpty(nativeClients)) {
```
ShadelessFox: Why can't we have both `nativeClients` and `inheritClients` at the same time? IMO it makes sense.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceProviderDescriptor.java
```diff
@@ -367,6 +370,9 @@ public boolean removeDriver(DriverDescriptor driver)
     // Native clients
 
     public List<NativeClientDescriptor> getNativeClients() {
+        if (inheritClients && parentProvider != null) {
+            nativeClients.addAll(parentProvider.getNativeClients());
+        }
```
ShadelessFox, line 375: It will bloat `nativeClients` every time this method is called.

Can we populate it in the constructor, just after reading the configuration?

ShadelessFox (changes requested):

Issue #13237, opened 2021-07-16
Create table: DDL is not updated after viewing it once


maffe:
#### System information: 
- Windows 10 64 Bit
- DBeaver 21.1.2

#### Connection specification:
- PostgreSQL server 10.13
- PostgreSQL driver 42.2.8
- Not using tunnels or proxies

#### Describe the problem you're observing:
When creating a table via the GUI, the DDL is not updated after viewing it once.

#### Steps to reproduce, if exist:
- Right-click on tables, choose Create table
- Add column1
- View DDL, see table definition with column1
- Add column2
- View DDL, column2 is not included

uslss:
thanks for the report

uslss:
verified

PR #15515, opened 2022-02-16
#13237 Refresh sql source viewer editor part if source object is not persistent

E1izabeth:


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
```diff
@@ -124,6 +124,15 @@ protected boolean isReadOnly()
...
     {
     }
+    
+    @Override
+    public void activatePart() {
+    	super.activatePart();
+    	
+    	if (!getSourceObject().isPersisted()) {
+    		refreshPart(this, true);
+    	}
```
ShadelessFox:
```suggestion
        super.activatePart();
        
        if (!getSourceObject().isPersisted()) {
            refreshPart(this, true);
        }
```
Usage of tabs instead of spaces

ShadelessFox (changes requested):

Issue #14032, opened 2021-09-28
Session manager displays Active Query column gigantic in Linux. In Windows is fine.


danielbgto:
HI all!

Active Query column in DBeaver Linux 21.2.1 (Linux Min 20.2) is showing the whole query code and there is no real viability of the session. The tooltip is taking practically the whole screen....
For example in Windows the tooltip and the Active Query column content size are well controlled.
In the screenshots is visible what I wish to explain.

is this a bug or can be fixed somehow? 

Windows:

![image](https://user-images.githubusercontent.com/5525957/135104163-20db3dd9-ff0a-4ed5-9ecf-b484671cca66.png)

Linux:

![image](https://user-images.githubusercontent.com/5525957/135104691-e2b1eede-dbf5-41f0-b5ef-94f7b597ed90.png)

LonwoLonwo:
Hello @danielbgto 
Thanks for the bug report.

This issue may be fixed here https://github.com/dbeaver/dbeaver/issues/13142
So you can wait for the official release 21.2.2 to test it. Or you can try our Early Access version: https://dbeaver.io/files/ea/

danielbgto:
WoW! Thank you for the good news!!

danielbgto:
I have updated from Early Access but  I found that behavior still active  on one place - on the task edit/Data Transfer window:

Windows:
![image](https://user-images.githubusercontent.com/5525957/135148576-f001a7ac-2352-4384-9f19-effdb5d208b6.png)

Linux:
![image](https://user-images.githubusercontent.com/5525957/135148705-997d24cf-50bd-4ca4-83e4-b5e1682c0e5b.png)

Matvey16:
@danielbgto Thank you for the report

ShadelessFox:
Another case can be seen here: #15582

Matvey16:
Verified

PR #15590, opened 2022-02-22
#14032 Replace line separators with paragraph symbol in container names

ShadelessFox:


serge-rider:
- Let's use single line conversion in UI classes only
- We need to leave FQN in the model where they are

PR #15597, opened 2022-02-22
CB-1668 limit drivers access to file system

alexander-skoblikov:


ShadelessFox:
Are you aware of https://openjdk.java.net/jeps/411? It may cause problems in the future.

serge-rider:
No sure about these checks: 
`container.isManageable()
&& JDBCDataSourceProvider.class.isAssignableFrom(driver.getDataSourceProvider().getClass())`

It is better to add special interface to data source providers (if needed). 
Also `isManageable` not related to just user datasources.
We probably should add property requiresAccessCheck to datasource or to the entire data source registry

alexander-skoblikov:
> Are you aware of https://openjdk.java.net/jeps/411? 

@ShadelessFox Yes, we know that SecurityManager has been marked as deprecated, but now we have no alternative way to restrict the rights of third-party code.

>It may cause problems in the future.

in the next few years, we do not want to update the java version to 17, and after that it will be used for several more years, so problems won't be in the near future

Issue #13287, opened 2021-07-21
Constraint generation during DDL as **separate** "ALTER...." statements


YuriDude:
**Is your feature request related to a problem? Please describe.**
I would like to have an option that is set as a configuration setting or other global that during the DDL generation of a table **all** constraints are generated as _**separate**_ ALTER statement(s)  and **not** in-line (meaning within "create table ..." statement) as it is current behavior now!

**Rationale**
I think that separate ALTER statement(s) for constraint(s) should in-fact be default behavior and "in-line" should be an option, but I am ok either way.
"THE WHY": during the DB cloning/installation for another environment (test, 2nd/3rd DEV, etc...) having separate DDLs for CREATE & ALTER provide more **flexibility**: tables can be "installed" in **any** order and then constraints can be added/enabled separately. Having constraint(s) in-line , as it is currently, implies that "some" **order must be followed**; e.g. parent table must be "installed"/be present before a child to be implemented. 
 

**Describe alternatives you've considered**
None

**Additional context**
Please also follow this pattern when DLL generated for a **schema**: CREATE TABLE's first , and ALTER on the "bottom".

Matvey16:
@YuriDude DDL generation depends on the database, which one do you have in mind?

YuriDude:
This behavior is the same on Oracle & MS SQL Server currently. 
Suggestion (feature request) is also relevant to these RDBMS'.
I did not check Postgres , MySQL DB2 and other "classical" RDBMS' due to the lack of access, but to my mind - feature should be applicable "across the board": 
- just like you are developing Universal DB Tool -> all shops/enterprises would have almost the same activities related to the Data Sources/Data Stores.

Also, just to clarify:
In original posting I neglected to mention that it related to FK  (foreign key)  constraints - they are the "culprits" ;-) of connections and thus somewhat elevated complexity. All others constraints (PK, not null, check, unique) don't need "special treatment" .

Matvey16:
@YuriDude We will look if it can be implemented

gabrielrra:
![image](https://user-images.githubusercontent.com/39421688/144238943-ae249eab-0d21-4415-915b-1a61306565c3.png)
When generating DLL now the Unique constraint is generated twice.
I can confirm that other constraints such as Foreign Keys, Primary Keys and other non clustered indexes are not duplicated, only Unique Constraints.

E1izabeth:
@gabrielrra, please, specify, which database do you use?

[...]

PR #15604, opened 2022-02-23
#13287 Add separate foreign keys generation in ddl for Oracle

E1izabeth:
Foreign keys are already generating separately from create table statement in DDL for common case.

Such behavior was added for Oracle.
It's not available to do for MySQL and DB2, because we get DDL from database and there is no such option for it.

While generating ddl for multiple tables, the order is taken into account and foreign keys can be inside create table statement, but this create statement is in the end of ddl.


plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleUtils.java
```diff
@@ -171,6 +176,33 @@ public static String getDDL(
...
 
+    private enum DBMSMetaDependentObjectType {
+        INDEX,
+        CONSTRAINT,
+        REF_CONSTRAINT,
+        TRIGGER
+    }
+
+    private static String invokeDBMSMetadataGetDependentDDL(JDBCSession session, OracleSchema schema, OracleTableBase object, DBMSMetaDependentObjectType dependentObjectType) {
+        StringBuilder sb = new StringBuilder();
```
ShadelessFox: Not sure about the usage of `StringBuilder` here. It looks redundant to me.

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleUtils.java
```diff
@@ -171,6 +176,33 @@ public static String getDDL(
...
+                dbStat.setString(2, schema.getName());
+            }
+            try (JDBCResultSet dbResult = dbStat.executeQuery()) {
+                if (dbResult.next()) {
+                    sb.append("\n\n").append(dbResult.getString(1).trim());
+                }
+            }
+        } catch (Exception e) {
+            // No dependent index DDL or something went wrong
+            log.debug("Error reading dependent index DDL", e);
```
ShadelessFox, line 203: Perhaps the caller should handle the exceptional situation themselves? It will allow providing human-readable error messages. Otherwise, at least dump some of the method's parameters.

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleTable.java
```diff
@@ -59,7 +60,12 @@
         + FOREIGN_KEY_NAME_PLACEHOLDER + " DISABLE";
     private static final String ENABLE_REFERENTIAL_INTEGRITY_STATEMENT = "ALTER TABLE " + TABLE_NAME_PLACEHOLDER + " MODIFY CONSTRAINT "
         + FOREIGN_KEY_NAME_PLACEHOLDER + " ENABLE";
-
+    
+    private static final String[] supportedOptions = new String[] {
```
ShadelessFox:
```suggestion
    private static final String[] supportedOptions = new String[]{
```

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleUtils.java
E1izabeth: It used to work this way before. I suppose it was done for some purpose. Not sure that it's good to change this behavior.

E1izabeth:
Checkboxes for DDL tab 
image
and Generate DDL
image
 window were added.
 There is no such option if it's not supported for database.

Issue #14843, opened 2021-12-08
Wrong path is opened when clicking browse in embedded database connection


Matvey16:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 21.3.1 EA

#### Steps to reproduce, if exist:
1. Open connection settings for embedded database (e.g. SQLite)
2. Click on **Browse..**

**Expected:** Path to the database location is opened
**Actual:** Last opened path is opened

![Yi9J43CKXd](https://user-images.githubusercontent.com/82543000/145221363-36db0b59-d084-4f4a-9b11-b1395741b772.gif)

LonwoLonwo:
Fixed for generic Databases (like SQLite, FireBird)

Matvey16:
Still doesn't work

![dbeaver_Gp0ZzI9aOJ](https://user-images.githubusercontent.com/82543000/155680011-8ec8d1e1-80d5-4e68-b282-8e46a6550e86.gif)

LonwoLonwo:
Suddenly it turned out to be an Eclipse issue https://bugs.eclipse.org/bugs/show_bug.cgi?id=577190

Hopefully, the bug will be corrected in the new version of Eclipse.

Matvey16:
Fixed in the latest eclipse version

PR #15609, opened 2022-02-24
#14843 set path to the directory (if we can parse it) for generic emb…

LonwoLonwo:
…edded databases


ShadelessFox (commented):
We already have some functionality to deal with such behavior.

Please take a look at:

https://github.com/dbeaver/dbeaver/blob/f4a5ad3e0860f3d3c2239d902e3bfb6086d311a3/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/TextWithOpenFile.java#L82-L99

Can we reuse it?

Issue #14810, opened 2021-12-06
[Oracle] Include filters for databases are case sensitive, but exclude filters are case-insensitive.


uslss:
#### System information: 
- Operating system (distribution) and version win10
- DBeaver version 21.3.0

#### Connection specification:
- Database name and version Oracle

#### Describe the problem you're observing:
Include filters for databases are case sensitive, but exclude filters are case-insensitive.
Existing databases:
![image](https://user-images.githubusercontent.com/31996417/144844386-6b7d07ad-0e4a-4269-ae30-4898852a6390.png)
Exclude filter:
![image](https://user-images.githubusercontent.com/31996417/144844884-c74b2f45-0a27-4015-bfea-645d6809bfc6.png)

Include filter:
![image](https://user-images.githubusercontent.com/31996417/144844971-ab5f4eb7-de02-4183-bac7-ec8d2f3e325b.png)
![image](https://user-images.githubusercontent.com/31996417/144845040-533f6ddd-b060-4933-9148-145ca8e661d6.png)

DariaMarkaryan:
Verified

PR #15613, opened 2022-02-24
#14810 case sensitive filtration for oracle

Destrolaric:


ShadelessFox (changes requested):
Instead of introducing another method, you can use the existing `SQLDialect#getCaseInsensitiveExpressionFormatter` and check it against `null`.

Issue #14694, opened 2021-11-26
Cannot modify the language using Window->Preferences->User Interface


wangqian-49:
#### System information: 
- Operating system (distribution) and version : Centos7
- DBeaver version : 21.2.2
- Additional extensions : nothing

#### Connection specification:
- Database name and version : postgresql 12.3
- Driver name : PostgreSQL JDBC Driver
- Do you use tunnels or proxies (SSH, SOCKS, etc)? : no

#### Describe the problem you're observing:

Cannot modify the language using Window->Preferences->User Interface

#### Steps to reproduce, if exist:

start dbeaver, click Window->Preferences, left click User Interface, and In the right window to modify the language, restart according to the prompt, the interface language has not changed. If closed and restarted, the language is modified. The -nl parameter in the dbeaver.ini is successfully modified.

Matvey16:
@wangqian-49 Hello, can you add how did install dbeaver?

wangqian-49:
> @wangqian-49 Hello, can you add how did install dbeaver?

@Matvey16 

1. Download the installation package "dbeaver-ce-21.2.2-stable.x86_64.rpm" from the website "https://dbeaver.io/download/".
2. Install using command "rpm -ivh dbeaver-ce-21.2.2-stable.x86_64.rpm".
3. cd /usr/share/dbeaver-ce, and ./dbeaver to start.
4. click Window->Preferences, left click User Interface,  and In the right window to modify the language,restart according to the prompt, the interface language has not changed.

The installation package of “rpm -ivh dbeaver-ce-21.3.0-stable.x86_64.rpm ” also has the same problem in the website "https://dbeaver.io/download/".

Matvey16:
@wangqian-49 I was not able to reproduce your issue because I don't have privileges to modify files in /usr/share directory and therefore can't change language in UI. I assume you modified privileges to be able to edit that directory?

Matvey16:
@wangqian-49 Did you by any chance start DBeaver from terminal with `sudo`?

wangqian-49:
@Matvey16 Yes, I created an ordinary user named test and changed the permission of the directory " /usr/share/dbeaver-ce/" to test with the command "chown -R test:test /usr/share/dbeaver-ce/", and then start DBeaver with command "cd  /usr/share/dbeaver-ce/;./dbeaver".

Matvey16:
[...]

PR #15615, opened 2022-02-24
#14694 Platform language change

ShadelessFox:


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageDatabaseUserInterface.java
```diff
@@ -203,16 +184,12 @@ public boolean performOk()
...
-                            getShell(),
-                            "Restart " + GeneralUtils.getProductName(),
-                            "You need to restart " + GeneralUtils.getProductName() + " to perform actual language change.\nDo you want to restart?")) {
-                            UIUtils.asyncExec(() -> PlatformUI.getWorkbench().restart());
-                        }
+                    ((DBPPlatformLanguageManager) DBWorkbench.getPlatform()).setPlatformLanguage(language);
+                    if (UIUtils.confirmAction(
+                        getShell(),
+                        "Restart " + GeneralUtils.getProductName(),
+                        "You need to restart " + GeneralUtils.getProductName() + " to perform actual language change.\nDo you want to restart?")) {
```
flambchap, line 191: I think messages in the context of changing language have to be i18nzed.

Issue #15097, opened 2021-12-30
data lost when copying result as html with some column hidden


dafuzzz:
#### System information: 
- Operating system (distribution) and version: ubuntu 20.04
- DBeaver version: 21.3.2.202112271055
- Additional extensions

#### Connection specification:
- Database name and version: postgres (PostgreSQL) 10.19 (Debian 10.19-1.pgdg90+1)
- Driver name: PostgreSQL
- Do you use tunnels or proxies (SSH, SOCKS, etc)? no

#### Describe the problem you're observing:

Lost data when advanced copying the whole line but hidding some column.

#### Steps to reproduce, if exist:

0. I have a table with structure and value shown below:
![image](https://user-images.githubusercontent.com/23740013/147739302-b43df968-320c-409b-8432-42e74f32aa2c.png)

1. filter some column(here, `column3`) with filter setting
![image](https://user-images.githubusercontent.com/23740013/147739392-0c15f94c-bd3c-40e3-bff7-7a60457a5d63.png)
![image](https://user-images.githubusercontent.com/23740013/147740191-65102e1d-45f6-4894-bef1-b595e78789d9.png)

2. right click the line num and select `Advanced Copy -> Copy ad HTML`

3. got html, but lost data of column4
![image](https://user-images.githubusercontent.com/23740013/147740780-c5a64976-c68c-49d8-877b-bec3fa6013e2.png)

4. Also, try to copy as JSON, got error
![image](https://user-images.githubusercontent.com/23740013/147740955-24d92432-c7b8-4293-a5a8-c311783db57c.png)

#### Include any warning/errors/backtraces from the logs

here is the log file tail output

```
2021-12-30 17:41:37.521 - Error fetching 'column4' value: Index 3 out of bounds for length 3
2021-12-30 17:41:37.521 - Bad attribute 'column4' index: 3 is out of row values' bounds (3)
2021-12-30 17:48:30.730 - Error fetching 'column4' value: Index 3 out of bounds for length 3
2021-12-30 17:48:30.730 - Bad attribute 'column4' index: 3 is out of row values' bounds (3)
2021-12-30 17:53:09.860 - Error fetching 'column4' value: Index 3 out of bounds for length 3
2021-12-30 17:53:09.865 - Index 3 out of bounds for length 3
java.lang.ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3
[...]

PR #15623, opened 2022-02-24
#15097 fix json and html format

Destrolaric:
The problem was related to the incorrect indexation in multiple places. 
We have two lists for attribute:

- Containing all attributes
- Containing visible attributes

Previously this happened:
We got index of first list and tried to get element from secong by it. This caused an incorrect behaviour when lists weren't same.

Fixed by directly getting it from list.

There is a probability that this will need a tweaking, if there is some cases when it isn't working.


serge-rider:
Incorrect fix. We must use standard attribute fetch functions

Issue #15572, opened 2022-02-21
Changes to Bind Parameter Window Broke Value Tabbing Capabilities


Xylandor:
#### System information: 
- OS:  MacOS Big Sur v11.4
- DBeaver version:  21.3.5.202202200327
- Additional extensions:  None

#### Connection specification:
- Local installation of MariaDb
- Driver name: MariaDB
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No

#### Describe the problem you're observing:
The latest release (21.3.5.202202200327) added the multi-line value editor to the Bind Parameter(s) window, but by doing so, it broke the ability to tab to the next parameter's value for editing. In previous versions, I could enter a value, press tab and it would set focus on the next parameter's value field (highlighting all text). Now I have to manually use the mouse to set focus on the next value. I use bind parameters all the time and many of my queries have several of them, so this can become very time consuming when having to switch back and forth between keyboard and mouse. 

#### Steps to reproduce, if exist:
1. Create a query with more than one bind parameter.
2. Enter the first parameter's value and try using TAB to set focus on the next parameter's value field. You can't. 

#### Include any warning/errors/backtraces from the logs
No errors produced.

Matvey16:
Verified

PR #15668, opened 2022-02-28
#15572 fix tabbing

Destrolaric:


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/dialogs/SQLQueryParameterBindDialog.java
```diff
@@ -183,21 +183,34 @@ public void widgetSelected(SelectionEvent e) {
...
-                    String newValue = ((Text) control).getText();
+                    String newValue = null;
+                    if (!(control instanceof Text)) {
+                        Control[] children = ((Composite) control).getChildren();
+                        for (Control child : children) {
+                            if (child instanceof Text) {
+                                newValue = ((Text) child).getText();
+                            }
+                        }
+                    } else {
```
ShadelessFox: Why is it even a case, by the way?

Destrolaric: When we tab out of edited cell we send composite which contains itself, button, and text. In this cases we need to find the Text child

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/dialogs/SQLQueryParameterBindDialog.java
```diff
@@ -183,21 +183,34 @@ public void widgetSelected(SelectionEvent e) {
...
+                    editor.addModifyListener(e -> saveEditorValue(editor, index, item));
                     editor.addTraverseListener(e -> {
                         if (e.detail == SWT.TRAVERSE_RETURN && (e.stateMask & SWT.CTRL) == SWT.CTRL) {
                             UIUtils.asyncExec(SQLQueryParameterBindDialog.this::okPressed);
                         }
                     });
-                    editor.addModifyListener(e -> saveEditorValue(editor, index, item));
-
+                    button.addTraverseListener(e -> {
+                        if (e.detail == SWT.TRAVERSE_TAB_NEXT) {
```
ShadelessFox: Could we handle `SWT.TRAVERSE_TAB_PREVIOUS` as well?

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/dialogs/SQLQueryParameterBindDialog.java
ShadelessFox: I don't see the place where we "send" our composite. The invocation of `keyTraversed` should not cause the issue IMO.

Destrolaric: Look towards the `createEditor` method.

ShadelessFox: I don't see anything suspicious in particular in that method.

Destrolaric: This method is overridden inside SQLParameterBindDialog, this implementation returns not the text editor but previously mentioned composite

ShadelessFox: Perhaps we shouldn't try to find the control of interest that way. Instead, we can bookkeep the `Text` instance as a class field and refer to it later when saving value.

Issue #15682, opened 2022-03-01
DBeaver tries to rollback transactions on disconnect even if the connection is in autocommit mode, causing the driver to emit errors


flambchap:
#### System information: 
- Operating system (distribution) and version: Win11
- DBeaver version 22.0.0 devel branch
- Additional extensions

#### Connection specification:
- Database name and version ~~Vertica 11~~
- Driver name
- Do you use tunnels or proxies (SSH, SOCKS, etc)? 

#### Steps to reproduce, if exist:
1. Open Vertica connection and wait for it to establish
2. Disconnect.
3. Examine logs and see the message below.

#### Include any warning/errors/backtraces from the logs
```Java
java.sql.SQLException: [Vertica][JDBC](11680) Cannot use rollback while Connection is in auto-commit mode.
	at com.vertica.exceptions.ExceptionConverter.toSQLException(Unknown Source)
	at com.vertica.jdbc.common.SConnection.rollback(Unknown Source)
	at org.jkiss.dbeaver.model.impl.jdbc.JDBCDataSource.lambda$closeConnection$3(JDBCDataSource.java:307)
	at org.jkiss.dbeaver.utils.RuntimeUtils$MonitoringTask.run(RuntimeUtils.java:423)
	at org.jkiss.dbeaver.utils.RuntimeUtils$1.run(RuntimeUtils.java:220)
	at org.jkiss.dbeaver.model.runtime.AbstractJob.run(AbstractJob.java:105)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

flambchap:
As I suspected, this issue has nothing to do with Verica. Foe example, here is what happens on MySQL:

```Java
com.mysql.jdbc.exceptions.jdbc4.MySQLNonTransientConnectionException: Can't call rollback when autocommit=true
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at com.mysql.jdbc.Util.handleNewInstance(Util.java:425)
[...]

PR #15683, opened 2022-03-01
dbeaver/dbeaver#15682 do not rollback on disconnect if connection is not in autocommit mode

flambchap:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCExecutionContext.java
```diff
@@ -352,6 +351,15 @@ public boolean isAutoCommit()
...
     }
 
+    private boolean isAutoCommit(boolean fallbackValue) {
+        try {
+            return isAutoCommit();
+        } catch (DBCException e) {
+            log.debug("Unable to determine if connection is in autocommit mode", e);
+            return fallbackValue;
+        }
+    }
```
ShadelessFox: IMO it would be more eye-friendly if this method was inlined, so no obscure `defaultValue` is passed back and forth.

LonwoLonwo (approved):

ShadelessFox (approved):

Issue #14739, opened 2021-11-30
PostgreSQL backup/restore task can't be executed if the wizard is opened from a tasks window


Matvey16:
#### System information: 
- DBeaver version 21.3.0

#### Connection specification:
- Database name and version PostgreSQL 12/14

#### Steps to reproduce, if exist:
1. Create a backup task
2. Double click on that task in **Database tasks** window
3. Immidiately click `Proceed`

**Expected:** Backup task is executed
**Actual:** Nothing changes and NPE is thrown

```
java.lang.NullPointerException
	at org.jkiss.dbeaver.ext.postgresql.tools.PostgreBackupWizardPageObjects.saveState(PostgreBackupWizardPageObjects.java:313)
	at org.jkiss.dbeaver.ext.postgresql.tools.PostgreBackupWizard.saveTaskState(PostgreBackupWizard.java:61)
	at org.jkiss.dbeaver.tasks.ui.wizard.TaskConfigurationWizard.saveConfigurationToTask(TaskConfigurationWizard.java:307)
	at org.jkiss.dbeaver.tasks.ui.wizard.TaskConfigurationWizard.saveTask(TaskConfigurationWizard.java:302)
	at org.jkiss.dbeaver.tasks.ui.wizard.TaskConfigurationWizard.performFinish(TaskConfigurationWizard.java:234)
	at org.jkiss.dbeaver.tasks.ui.nativetool.AbstractNativeToolWizard.performFinish(AbstractNativeToolWizard.java:210)
	at org.jkiss.dbeaver.tasks.ui.nativetool.AbstractNativeExportWizard.performFinish(AbstractNativeExportWizard.java:75)
	at org.jkiss.dbeaver.ui.dialogs.MultiPageWizardDialog.buttonPressed(MultiPageWizardDialog.java:353)
	at org.jkiss.dbeaver.tasks.ui.wizard.TaskConfigurationWizardDialog.buttonPressed(TaskConfigurationWizardDialog.java:225)
	at org.eclipse.jface.dialogs.Dialog.lambda$0(Dialog.java:619)
	at org.eclipse.swt.events.SelectionListener$1.widgetSelected(SelectionListener.java:84)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:252)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4237)
[...]

PR #15684, opened 2022-03-01
#14739 disable proceed if any page is not loaded

Destrolaric:
Disable Proceed button if any page is not loaded


serge-rider:
This is definitely not correct. Task may be saved in PG wizards too. There are too many different workflows linked to this function.

PR #15731, opened 2022-03-04
dbeaver/dbeaver-ee#1302 Support 2fa with pam for mariadb

E1izabeth:
For each connection user will get such dialog 
image

Unfortunately, there is no callback for custom dialog plugin in connector/j library as it is in connector/c. Driver overrides custom dialog plugin class, so there are two options:
- make dirty hacks with class loading
- retry on specific error during connection

Class loading hacks are hardly combined with dynamic loading, so the second way is presented. Connector/J throws ordinary SQL Exception while trying to connect using two-factor authentication without secondary password being specified, so we are intercepting this exception and checking its message for a reason then displaying password prompt if needed.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql.ui/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.mysql.ui/plugin.xml
@@ -224,4 +224,8 @@
         <configurator type="mysqlScriptExecute" class="org.jkiss.dbeaver.ext.mysql.tools.MySQLTaskConfigurator"/>
     </extension>
 
+	<extension point="org.jkiss.dbeaver.ext.mysql.auth.secondaryStepDialogProvider">
+        <provider class="org.jkiss.dbeaver.ext.mysql.ui.SecondaryStepAuthDialogProvider"/>
+    </extension>
+	
 </plugin>
--- a/plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/SecondaryStepAuthDialogProvider.java
+++ b/plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/SecondaryStepAuthDialogProvider.java
@@ -1,0 +1,78 @@
+package org.jkiss.dbeaver.ext.mysql.ui;
+
+import org.eclipse.jface.dialogs.IDialogConstants;
+import org.eclipse.osgi.util.NLS;
+import org.eclipse.swt.SWT;
+import org.eclipse.swt.layout.GridData;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Text;
+import org.jkiss.dbeaver.ext.mysql.SecondaryStepAuthProvider;
+import org.jkiss.dbeaver.ext.mysql.ui.internal.MySQLUIMessages;
+import org.jkiss.dbeaver.model.runtime.RunnableWithResult;
+import org.jkiss.dbeaver.ui.UIUtils;
+import org.jkiss.dbeaver.ui.dialogs.BaseDialog;
+
+public class SecondaryStepAuthDialogProvider implements SecondaryStepAuthProvider {
+
+    @Override
+    public String obtainSecondaryPassword(String reason) {
+        return UIUtils.syncExec(new RunnableWithResult<String>() {
+            @Override
+            public String runWithResult()  {
+                InputPasswordDialog  dialog = new InputPasswordDialog(UIUtils.getActiveWorkbenchShell(), reason);
+                if (dialog.open() == IDialogConstants.OK_ID) {
+                    return dialog.getPasswordString();
+                } else {
+                    return null;
+                }
+            }
+        });
+    }
+
+    private class InputPasswordDialog extends BaseDialog {
+        private String passwordText;
+        private Text passwordTextControl;
+        private String reason;
+        
+        public InputPasswordDialog(Shell parentShell, String reason) {
+            super(parentShell, MySQLUIMessages.two_factor_auth_secondary_password_dialog_title, null);
+            this.reason = reason;
+        }
+        
+        @Override
+        protected boolean isResizable() {
+            return false;
+        }
+
+        public String getPasswordString() {
+            return passwordText;
+        }
+
+        @Override
+        protected Composite createDialogArea(Composite parent) {
+            Composite composite = super.createDialogArea(parent);
+            
+            GridData gd = new GridData(GridData.FILL_BOTH);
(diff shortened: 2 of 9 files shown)


product/debug/DBeaver.product.launch
```diff
@@ -34,6 +34,7 @@
         <setEntry value="com.sun.jna*5.8.0.v20210503-0343@default:default"/>
         <setEntry value="com.sun.jna.platform*4.5.1.v20190425-1842@default:default"/>
         <setEntry value="com.sun.jna.platform*5.8.0.v20210406-1004@default:default"/>
+        <setEntry value="jakarta.servlet-api@default:default"/>
```
ShadelessFox: Probably unintended change.

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/SecondaryStepAuthProvider.java
```diff
@@ -0,0 +1,33 @@
...
+                    try {
+                        return (SecondaryStepAuthProvider)elts[0].createExecutableExtension("class");
+                    } catch (CoreException e) {
+                        log.error("Failed to load SecondaryStepDialogProvider", e);
+                    }
+                }
+            }
+        }
+        return null;
+    }
```
ShadelessFox, line 32: Please don't do that.

Create a separate registry that will hold a set of defined descriptors via an associated extension point.

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/SecondaryStepAuthProvider.java
```diff
@@ -0,0 +1,33 @@
...
+import org.eclipse.core.runtime.IExtension;
+import org.eclipse.core.runtime.IExtensionPoint;
+import org.eclipse.core.runtime.Platform;
+import org.jkiss.dbeaver.Log;
+
+public interface SecondaryStepAuthProvider {
+    
+    static final Log log = Log.getLog(SecondaryStepAuthProvider.class);
+    
+    String obtainSecondaryPassword(String reason);
```
ShadelessFox, line 14: Please specify `@NotNull`/`@Nullable` annotations for the method.

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/SecondaryStepAuthProvider.java
```diff
@@ -0,0 +1,33 @@
...
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IConfigurationElement;
+import org.eclipse.core.runtime.IExtension;
+import org.eclipse.core.runtime.IExtensionPoint;
+import org.eclipse.core.runtime.Platform;
+import org.jkiss.dbeaver.Log;
+
+public interface SecondaryStepAuthProvider {
+    
+    static final Log log = Log.getLog(SecondaryStepAuthProvider.class);
```
ShadelessFox, line 12: See the previous comment about the registry.

plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
```diff
@@ -266,6 +265,9 @@
                     <property name="@dbeaver-default-resultset.maxrows.sql" value="true"/>
                     <!--<property name="@dbeaver-default-resultset.format.datetime.native" value="true"/>-->
                     <parameter name="krb5.enabled" value="true"/>
+                    
+                    <overrideClass originalName="org.mariadb.jdbc.internal.com.send.authentication.SendPamAuthPacket"
+                                   implementationName="org.jkiss.dbeaver.ext.mysql.auth.DialogAuthPlugin2" />
```
ShadelessFox: I couldn't find the definition for the `overrideClass` attribute along with the implementation class.

Is this PR incomplete?

plugins/org.jkiss.dbeaver.ext.mysql.ui/plugin.xml
```diff
@@ -224,4 +224,8 @@
         <configurator type="mysqlScriptExecute" class="org.jkiss.dbeaver.ext.mysql.tools.MySQLTaskConfigurator"/>
     </extension>
 
+	<extension point="org.jkiss.dbeaver.ext.mysql.auth.secondaryStepDialogProvider">
+		<provider class="org.jkiss.dbeaver.ext.mysql.ui.SecondaryStepAuthDialogProvider" />
+    </extension>
```
ShadelessFox: Please fix the formatting.

plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/SecondaryStepAuthDialogProvider.java
```diff
@@ -0,0 +1,78 @@
...
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Shell;
+import org.eclipse.swt.widgets.Text;
+import org.jkiss.dbeaver.ext.mysql.SecondaryStepAuthProvider;
+import org.jkiss.dbeaver.ext.mysql.ui.internal.MySQLUIMessages;
+import org.jkiss.dbeaver.model.runtime.RunnableWithResult;
+import org.jkiss.dbeaver.ui.UIUtils;
+import org.jkiss.dbeaver.ui.dialogs.BaseDialog;
+
+public class SecondaryStepAuthDialogProvider implements SecondaryStepAuthProvider {
```
ShadelessFox, line 16: We already have a set of methods for retrieving user credentials.

Please see `DBPPlatformUI#promptUserCredentials`

ShadelessFox (changes requested):
Generally speaking, I wouldn't say I like that this model is tightly bound to MariaDB only.

Some other databases might also support 2FA authentication, and we may want to provide support for them in DBeaver by reusing this model.

plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/SecondaryStepAuthDialogProvider.java
E1izabeth: Unfortunately, it's not fully suitable for this case. The option `showSavePassword` there only enables/disables checkbox instead of removing it at all. In this case we can't have Save password option at all. 

image

Besides that, the label and the checkbox are in a one line while dialog window is not resizable which doesn't work for this case when user opens SQL Script with a long name.

image

Issue #15663, opened 2022-02-28
Text view in data editor has clickable button that should be greyed


Matvey16:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 22.0.0

#### Describe the problem you're observing:
Button to edit cell value in separate dialog is not greyed out in text view

#### Steps to reproduce, if exist:
1. Open any table with data in data editor
2. Change grid view to text
3. Click on `edit cell value in separate dialog` button

**Expected:** The button either works or is greyed and unclickable

**Actual:** The button does nothing

![dbeaver_21uG3140mX](https://user-images.githubusercontent.com/82543000/155986634-c15973ab-58aa-4599-af06-80c82506a714.gif)

Matvey16:
Verified

PR #15770, opened 2022-03-09
#15663 disable button for read-only fields

Destrolaric:


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/handler/ResultSetPropertyTester.java
```diff
@@ -106,7 +107,7 @@ private boolean checkResultSetProperty(ResultSetViewer rsv, String property, Obj
                 }
                 if ("edit".equals(expectedValue) || "inline".equals(expectedValue)) {
                     DBDAttributeBinding attr = rsv.getActivePresentation().getCurrentAttribute();
-                    if (attr == null) {
+                    if (attr == null || !(rsv.getActivePresentation() instanceof IResultSetEditor)) {
```
ShadelessFox, line 110: That's a really dirty hack.

Use `DBSDataContainer#getSupportedFeatures` and exclude `FEATURE_DATA_FILTER` for the <kbd>Statistics</kbd> container.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/handler/ResultSetPropertyTester.java
ShadelessFox, line 110: That's a dirty hack.

Please use `ResultSetViewer#getReadOnlyStatus` instead.

ShadelessFox (changes requested):

Issue #14589, opened 2021-11-17
Line Numbers for Text Results


jetzerb:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
In the results pane, the Grid-based presentation includes line numbers.  I'd like the option to include line numbers in the Text-based presentation as well:
![image](https://user-images.githubusercontent.com/36859020/142208502-940eb7a8-8950-4526-a56c-3d924648bb1e.png)

The text-presentation results would then look something like this:
![image](https://user-images.githubusercontent.com/36859020/142207122-0991d318-6a1a-4a1f-9f37-46675281d667.png)

**Describe alternatives you've considered**
Switching back and forth between grid & text presentation via ``CTRL+` `` to see the line number for a given row works for the first few rows in the result set, but not for later results since switching presentations jumps back to the top of the results.
Adding a `row_number()` column to each query is tedious and requires me to keep the `order by`s sync'd between the overall query and the window function.

**Additional context**
Grid presentation shows line numbers:
![image](https://user-images.githubusercontent.com/36859020/142206739-26aa5406-aca5-460e-a5e0-a8c1557e76b3.png)

Text presentation currently does not:
![image](https://user-images.githubusercontent.com/36859020/142206843-6a352cdb-b6ca-4984-8f73-cb8e18f0b356.png)

LonwoLonwo:
@jetzerb 

Thanks for the idea!

uslss:
line number values could be shown in the Value panel and incorrect values are shown in this case
![dbeaver_lhwiaoBXJo](https://user-images.githubusercontent.com/31996417/158559901-3c2d70e4-3f55-4438-b88e-dbf04b846efd.gif)

uslss:
verified

PR #15771, opened 2022-03-09
#14589 add line numeration for text view

Destrolaric:
image
image


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/plaintext/PlainTextPresentation.java
```diff
@@ -294,9 +298,21 @@ private void printGrid(boolean append) {
         }
         // Print header
         if (delimLeading) grid.append("|");
-        for (int i = 0; i < attrs.size(); i++) {
+        if (lineNumbers && attrs.size() != 0) {
+            if (extraSpaces) {
+                grid.append(" ");
+            }
+            grid.append("№");
```
ShadelessFox, line 305: Please use the `#` sign instead.

ShadelessFox (changes requested):

Issue #13431, opened 2021-08-03
Where filter can be modifyed in 'Statistics' tab (SQL Script execution result)


Wroud:
#### Describe the problem you're observing:
Where filter can be edited in statistics tab, result set can be refreshed (only once)

#### Steps to reproduce, if exist:
execute any SQL Script (Alt + X) in SQL Editor

![image](https://user-images.githubusercontent.com/811729/128060897-8649bc18-6f0a-4805-9d1c-9da057c413bf.png)
![image](https://user-images.githubusercontent.com/811729/128060941-5cb35c87-53a9-465e-9cc3-cb2d2b811012.png)

uslss:
![image](https://user-images.githubusercontent.com/31996417/158546461-e613756f-bd21-40e0-ae2c-0fdc69e13e69.png)
filter panel is still available if the result is an error

uslss:
verified

ShadelessFox:
Editing the filter on a valid query to become invalid hides the filter panel, and I believe this is NOT the expected behavior:

https://user-images.githubusercontent.com/35821147/158721952-4486b911-bda8-4024-8fbc-4cfa9e7ae060.mp4

There's no way to stop the automatic refresh if a database error occurs during refresh (simulated by dropping the selected table):

https://user-images.githubusercontent.com/35821147/158722829-4ba097cb-5cf7-4559-bb57-839a2f46f2bb.mp4

Can we gray-out the filter panel instead of hiding it? This blank space drives me nuts.

serge-rider:
Last fix was reverted. We need filters in error panel.

uslss:
verified

PR #15773, opened 2022-03-09
#13431 filter panel removed if resultset is statistic

Destrolaric:
image


ShadelessFox (changes requested):
Without the filters panel, it looks weird.

I think it would be better to just gray it out.

To achieve this, `DBSDataContainer#getSupportedFeatures` should exclude `FEATURE_DATA_FILTER` for the <kbd>Statistics</kbd> tab.

serge-rider:
@Destrolaric please attach screenshot of statistics panel

Issue #13570, opened 2021-08-17
One-click to turn on/off READ-ONLY setting


achangs:
**Is your feature request related to a problem? Please describe.**

I often need to turn this setting on/off, to prevent my clobbering a database by mistake.   Currently, to do this, I need to click several times to "Edit Connection", which is inconvenient.

**Describe the solution you'd like**

Make turning on/off of READ-ONLY into a simple one-click from the right-button pull-down menu.

Admittedly, this is only a minor issue, but I thought I'd ask to make this great tool even BETTER!    Thank you!

**Describe alternatives you've considered**

**Additional context**

uslss:
in case the connection read-only setting is switched on/off when data viewer is opened and another connection is chosen at db navigator, focus on connection is lost.
![dbeaver_Xf6STGeJRR](https://user-images.githubusercontent.com/31996417/158589596-33042ce9-3319-4669-b73a-45e1a35f1a87.gif)

uslss:
verified

PR #15832, opened 2022-03-14
#13570 Add readonly setting on toolbar

E1izabeth:
image


ShadelessFox (commented):
I'm unsure about the manual approach of refreshing the action enablement state.

Have you looked at custom expression definitions (`org.eclipse.core.expressions.definitions`)?

E1izabeth:
I followed the same approach as used for the other connection-related toolbar handlers (connect/disconnect commands and data source selection).
If you know, how to use expressions in this particular case, including correct refresh on connection parameters change, then advice me please.
Also there is no way to specify expression for "checkedWhen" according to handlers extension point schema.

PR #15851, opened 2022-03-15
fix: pt-br metadata translation on properties of sql console

TiagoValdrich:
Fixing metadata translation on SQL Console settings. It should be "Metadados"in PT-BR instead of "Metadatdoa".

image


E1izabeth:
Hi, @TiagoValdrich! Thank you for contribution!

PR #15894, opened 2022-03-19
#15839 Fix locating current sql script element in the editor

### LINKED ISSUE #15839: "Next Query", "Previous Query" shortcut keys work error
#### System information: 
- Windows 10 Pro
- DBeaver 22.0.0
- Additional extensions

When executing "Next Query", "Previous Query" shortcut keys, general SQL works normally.
However, in the "@set" syntax, it moves only to the upper syntax.

Thank for reading.

arvillion:
Fix [iss#15839](https://github.com/dbeaver/dbeaver/issues/15839) and add two test cases related.


serge-rider:
Looks good, thank you for the contribution.
However we need to test it for possible side effects (control commands were handled like delimiters for a reason).

Issue #13067, opened 2021-07-01
Metadata search: search Exasol metadata in definitions


flambchap:
Currently, we only support PostgreSQL, MySQL, Oracle, MSSQL, and DB2.

flambchap:
Exasol's system tables contain definitions for views and scripts, so we can teach DBeaver to look for these objects by definitions.

Matvey16:
Verified

PR #15896, opened 2022-03-19
dbeaver/dbeaver#13067 add search in definitions for Exasol

flambchap:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCUtils.java
```diff
@@ -1,857 +1,898 @@
...
+    }
+
+    public static long getGeneratedKey(PreparedStatement dbStat) throws SQLException {
+        try (final ResultSet keysRS = dbStat.getGeneratedKeys()) {
+            keysRS.next();
+            return keysRS.getLong(1);
+        }
+    }
+
+    public static <T> void fetchObjects(@NotNull Collection<? super T> objects, int limit, @NotNull JDBCSession session, @NotNull CharSequence sql,
```
LonwoLonwo, line 865: Maybe some comments for this method can be beneficial.

flambchap: Agreed.

Issue #14591, opened 2021-11-17
Add Python to Copy SQL as source code


rrousselot:
**Is your feature request related to a problem? Please describe.**
Python is currently the most popular programming language on the planet.  Give it some love.  

**Describe the solution you'd like**
Add an option for Python multi line formatting.

**Describe alternatives you've considered**
Do it manually now.

**Additional context**
output example..

"CREATE TABLE Example AS " \
        "SELECT BlockFIPS, HU, HH, Pop, " \
        "FROM Source" \
        "GROUP BY BlockFIPS;"

ShadelessFox:
Hello @rrousselot,

Thanks for feature request.

Are you certain that provided snippet is a valid Python code? Looks more like a C-way of string concatenation.

Did you mean something like this?

```py
"""
CREATE TABLE Example AS
SELECT BlockFIPS, HU, HH, Pop,
FROM Source
GROUP BY BlockFIPS;
"""
```

rrousselot:
I think the \ are optional so it would only need to encase each line with "

Or use """ (triple quotes) at beginning (first line) and end (last line) as you a have shown.

Six of one, half dozen the other; I have no preference.

python - Pythonic way to create a long multi-line string - Stack Overflow<https://stackoverflow.com/questions/10660435/pythonic-way-to-create-a-long-multi-line-string>

I tried to find an accepted PEP addressing multi line strings but could not.

Matvey16:
Verified

PR #15902, opened 2022-03-21
#14591 add python for copy as source code

Destrolaric:
Example input :
```
CREATE EVENT TRIGGER log_ddl_drop_info ON sql_drop
	EXECUTE FUNCTION testas.log_ddl_drop()
```
StringBuilder example output:
```
list = []
list.append('CREATE EVENT TRIGGER log_ddl_drop_info ON sql_drop\n')
list.append('EXECUTE FUNCTION testas.log_ddl_drop()\n')
query = ''.join(list)
```
No StringBuilder example output:
```
"""
CREATE EVENT TRIGGER log_ddl_drop_info ON sql_drop
EXECUTE FUNCTION testas.log_ddl_drop()
"""
```


ShadelessFox (changes requested):
I doubt that this "string builder" pattern is useful in the case of Python and here's why:

1. It uses two local variables (`type` and `query`), one of them is temporary
2. The variable `list` shadows the global function `list`, which is terrible according to PEP.
3. It performs extra operations instead of simply creating a single multiline string constant.

Let's remove it.

ShadelessFox (approved):

Issue #15376, opened 2022-02-03
Update H2GIS to 2.0.0


ebocher:
H2GIS 2.0.0 is out on Maven Central.
We have updated it on your plugin repo https://github.com/orbisgis/orbisrcp/tree/master/plugins/org.orbisgis.h2gis
But we have an issue to detect a Geometry data type.
DBeaver set the type name from a plugin.xml file, here https://github.com/orbisgis/orbisrcp/tree/master/plugins/org.orbisgis.h2gis
Since H2 2.X version the string type representation for a geometry has changed. See my issue to the H2 community https://github.com/h2database/h2database/issues/3419
Is there a way to set the type name to DBeaver from code and not from a static value in a file ?
Best regards

ShadelessFox:
Hello @ebocher,

If I understand you correctly, you can use `<type name="*"/>` instead of specifying a concrete type name.

PR #15972, opened 2022-03-25
Fix H2GIS geometry type to support the new H2 type signature

ebocher:
#15376

This PR introduces a fix on the new geometry type signature returned by H2 since 2.0.X version.

e.g : GEOMETRY(POINT, 4326)

when a user creates the following table

```
CREATE TABLE mygeotable (ID INTEGER, GEOM GEOMETRY(POINT, 4326));
```


serge-rider:

PR #16021, opened 2022-03-31
CB-1808 auth model adminOnly property

yagudin10:


serge-rider:
We need to populate this property in graphql schema

Issue #16046, opened 2022-04-01
German translations in Table Editor are miss-leading


KLIS-IT:
some of the German translations in Table Editor are miss-leading

E1izabeth:
Does this issue still actual?

PR #16048, opened 2022-04-01
#16046 correct some German translations

KLIS-IT:


plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/internal/EditorsResources_de.properties
```diff
@@ -44,9 +44,9 @@ dialog_morph_delimited_target_group_delim_result = Ergebnis Trennzeichen
 dialog_morph_delimited_target_group_leading_text = F\u00FChrender Text
 dialog_morph_delimited_target_group_trailing_text = Nachlaufender Text
 
-database_editor_command_save_name = Speichern ...
-database_editor_command_save_tip = Vorschau\u00E4nderungen und speichern
+database_editor_command_save_name = Speichern
```
E1izabeth, line 47: I don't see the reason to remove `...` from `Save...` command because `...` indicates continuation of dialog.

plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/internal/EditorsResources_de.properties
```diff
@@ -44,9 +44,9 @@ dialog_morph_delimited_target_group_delim_result = Ergebnis Trennzeichen
...
+database_editor_command_save_tip = Alle\u00C4nderungen speichern
 database_editor_command_revert_name = Zur\u00FCcksetzen
-database_editor_command_revert_tip = Alle \u00C4nderungen im aktuellen Editor zur\u00FCcksetzen
+database_editor_command_revert_tip = Alle \u00C4nderungen zur\u00FCcksetzen
```
E1izabeth, line 50: Not sure that we need to shorten tips like:
`Reset all changes in the current editor` -> `Reset all changes`
`Preview changes and save` -> `Save all changes`

E1izabeth (commented):
But maybe someone has another point of view?

plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/internal/EditorsResources_de.properties
KLIS-IT: Why are the missleading translations better than the new ones?

E1izabeth: Could you explain, please, why the previous translation is misleading? Maybe we could come up with something better.

KLIS-IT: okay - let's check line by line:

**database_editor_command_save_name** --> I'm totally fine with "Speichern ..." although it is not really common to use it in this way
**database_editor_command_save_tip** --> "Vorschauänderungen und speichern" is not German speech. Better is "Alle\u00C4nderungen speichern" which is "save all changes" which definitely better describes what pgm will do
**database_editor_command_revert_tip** --> why we need to state that we're talking about changes in current editor. But if really needed we could leave it as "Alle Änderungen im aktuellen Editor zurücksetzen"

what do you think?

E1izabeth: I think it's important to state that changes will be reset only for current editor in `database_editor_command_revert_tip`.
About `database_editor_command_save_tip`: I'm not sure how to express in German thought about not only save, but also preview changes. If there is no good way for that, then let's change it as you propose. 

Thank you so much for your work and explanation.

Issue #15244, opened 2022-01-21
No error is thrown when executing erroneous postgres script task


DariaMarkaryan:
#### System information: 
- Win10
- DBeaver CE 21.3.3

#### Connection specification:
- Database name and version: PostgreSQL(14, 12...)

#### Describe the problem you're observing:
The script runs without problems, but we 

#### Steps to reproduce, if exist:
1. Create SQL script with mistakes and save it to your computer
2. Check that it doesn't works when executing from SQL editor
3. Create a task with type "PostgreSQL Script task" with the created file
4. Start this task
**Expected**:we expect an error to be thrown
**Actual**: no data was inserted, but we see "success message"
![image](https://user-images.githubusercontent.com/43252954/150507655-34637be4-f611-4c09-bc9b-3c1a25a73436.png)

Matvey16:
Verified

PR #16118, opened 2022-04-07
#15244 fix error notifications for script tasks

Destrolaric:
The only way to determine if that type of task had any error is to look for stderr.

### DIFF
--- a/plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/AbstractNativeToolHandler.java
+++ b/plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/AbstractNativeToolHandler.java
@@ -52,6 +52,9 @@ import java.util.*;
 
 public abstract class AbstractNativeToolHandler<SETTINGS extends AbstractNativeToolSettings<BASE_OBJECT>, BASE_OBJECT extends DBSObject, PROCESS_ARG> implements DBTTaskHandler {
 
+    private String taskErrorMessage;
+
+
     @Override
     @NotNull
     public DBTTaskRunStatus executeTask(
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/AbstractNativeToolHandler.java
```diff
@@ -476,24 +483,16 @@ public void run() {
...
-                            buf.setLength(0);
-                        }
-                        //int avail = input.available();
+
+                if (isLogInputStream) {
+                    Object errorMessage = readStream(input.getErrorStream());
+                    readStream(input.getInputStream());
+                    if (!CommonUtils.isEmpty(String.valueOf(errorMessage))) {
+                        task.getProperties().put("Error Message", errorMessage);
                     }
```
ShadelessFox: Store the instance of the exception in this job instead.

plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/AbstractNativeToolHandler.java
```diff
@@ -75,7 +75,10 @@ public DBTTaskRunStatus executeTask(
                 listener.taskStarted(task);
                 Throwable error = null;
                 try {
-                    doExecute(monitor, task, settings, log);
+                    final boolean executionResult = doExecute(monitor, task, settings, log);
+                    if (!executionResult) {
+                        error = new DBCException("Task failed, reason: " + task.getProperties().get("Error Message"));
```
ShadelessFox: See previous note.

ShadelessFox (changes requested):

serge-rider:
- Use constants but weird hardcoded "Error Message" strings.
- You can't put error message in properties. It has to be placed in a separate variable in tool handler.
- Why `Object errorMessage`? And then cast it to string 👎

Issue #15067, opened 2021-12-28
[XLSX] Allow to append new data to existing file in Data-transfer


Destrolaric:
Currently, we cannot append any information towards the XLSX format. If something is appended file becomes unreadable.

**Describe the solution you'd like**
If the XLSX format is selected, we need to append the required information correctly.

ShadelessFox:
Duplicates #15042

ShadelessFox:
Implemented for XLSX, CSV, and TXT exporters.

Regression: it's now impossible to enable append for file formats that are not listed above.

Matvey16:
Verified

PR #16151, opened 2022-04-11
#15067 Append exported data to an existing file

ShadelessFox:


ShadelessFox:
Implemented for XLSX, CSV, and TXT ezporters.

Regression: it's now impossible to enable append for file formats that are not listed above.

PR #16160, opened 2022-04-12
dbeaver/dbeaver-ee#1529 fix eula dialog not being shown after license

Destrolaric:


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/eula/EULAUtils.java
```diff
@@ -51,4 +56,36 @@ public static String getPackageEula() {
         return eula;
     }
 
+    /*
+     *  License manager overrides all other workbench advisors if shown. We need to call EULA manually from it.
+     */
+    public static void summonEulaIfNeeded(@NotNull Shell shell) {
```
ShadelessFox: We don't summon demons here. Let's use the word `show` instead of `summon`.

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/eula/EULAUtils.java
```diff
@@ -51,4 +56,36 @@ public static String getPackageEula() {
         return eula;
     }
 
+    /*
+     *  License manager overrides all other workbench advisors if shown. We need to call EULA manually from it.
+     */
```
ShadelessFox: Let's not include this comment here. Instead, you can move it to this method's callsite.

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/eula/EULAUtils.java
```diff
@@ -51,4 +56,36 @@ public static String getPackageEula() {
...
+        EULAConfirmationDialog eulaDialog = new EULAConfirmationDialog(shell, eula);
+        eulaDialog.open();
+    }
+
+    private static void showEulaInfoDialog(@NotNull Shell shell, @Nullable String eula) {
+        EULAInfoDialog eulaDialog = new EULAInfoDialog(shell, eula);
+        eulaDialog.open();
+    }
+
+
```
ShadelessFox: Extra lines

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/eula/EULAUtils.java
```diff
@@ -51,4 +56,36 @@ public static String getPackageEula() {
         return eula;
     }
 
+    /*
+     *  License manager overrides all other workbench advisors if shown. We need to call EULA manually from it.
+     */
+    public static void summonEulaIfNeeded(@NotNull Shell shell) {
+        Preferences preferences = Preferences.userNodeForPackage(DBWorkbench.getPlatform().getApplication().getClass());
+        if (preferences.get(DBEAVER_EULA, null) == null || !preferences.get(DBEAVER_EULA, null).equals(EULAUtils.getEulaVersion())) {
```
ShadelessFox:
```suggestion
        if (!EULAUtils.getEulaVersion().equals(preferences.get(DBEAVER_EULA, null))) {
```

ShadelessFox (changes requested):

ShadelessFox:
Commit's message does not match the changes. It primarily refactors the code.

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/eula/EULAConfirmationDialog.java
```diff
@@ -56,7 +55,7 @@ protected void buttonPressed(int buttonId) {
                 break;
             case IDialogConstants.YES_ID:
                 Preferences preferences = Preferences.userNodeForPackage(DBWorkbench.getPlatform().getApplication().getClass());
-                preferences.put(DBEAVER_EULA, EULAUtils.getEulaVersion());
+                preferences.put(EULAUtils.DBEAVER_EULA, EULAUtils.getEulaVersion());
```
ShadelessFox, line 58: Let's move this logic into the `EULAUtils` class, too.

ShadelessFox (changes requested):
I don't really see the reason for having the `EULAUtils` class. All its methods can be moved into the initializer or handler classes.

PR #16182, opened 2022-04-13
Fix typo for importing connections from sqldeveloper

turakamou:
Connection was imported using the Host as a user.
Correction made: use the correct getter.


serge-rider:

turakamou:
@serge-rider Happy to contribute.
FYI, I'm preparing a commit that will be able to import also the passwords from sqldeveloper.
I hope you will be interested.

Issue #16163, opened 2022-04-12
Can't change database in connection dialog


OlegKustol:
Hi Team,

2 days ago I faced with a problem and it still drive me mad. I can't change postgres database name in "choose catalog/schema dialog" window. When I try to change it I can see next window:
![image](https://user-images.githubusercontent.com/103504435/162950104-0aea21e9-cb93-4d77-adef-3bfc963f61e1.png)

LonwoLonwo:
Hello @OlegKustol 

What DBeaver version do you use?
What Postgres server and driver versions do you use?

Also, please provide an [error log](https://github.com/dbeaver/dbeaver/wiki/Log-files). I can't reproduce your issue for now.

OlegKustol:
Dbeaver version is Version 21.2.0.202109061154
JDBC driver  for postgresql version 13
PostgreSQL server: AWS RDS PostgreSQL server version 13.

[dbeaver-debug.log](https://github.com/dbeaver/dbeaver/files/8472520/dbeaver-debug.log)
[ - Copy.log](https://github.com/dbeaver/dbeaver/files/8472522/-.Copy.log)

LonwoLonwo:
I can offer you to update the DBeaver version to check the condition of your issue. We do not create patches for previous versions.

OlegKustol:
Got it. But I updated DBeaver today. I don't know why the version is old. I'll update app and let you know if some issues with change database name are there.

LonwoLonwo:
Ok, thnaks for the answer @OlegKustol

OlegKustol:
I updated DBeaver, but I still have the same issues.
![image](https://user-images.githubusercontent.com/103504435/162967686-ad9f51c5-b0ad-4f8f-b9e3-6b0c668c28b9.png)
![image](https://user-images.githubusercontent.com/103504435/162967731-a8449850-cd27-43d5-90d5-afae4adddbde.png)

LonwoLonwo:
Maybe your log changed?

OlegKustol:
Hello,
please find my logs in attachment. Today I still have this issues.

[dbeaver-debug.log](https://github.com/dbeaver/dbeaver/files/8482564/dbeaver-debug.log)
[ - Copy.log](https://github.com/dbeaver/dbeaver/files/8482567/-.Copy.log)

LonwoLonwo:
Ok. what do you mean about "can't change postgres database name"?

[...]

PR #16198, opened 2022-04-15
#16163 show error message instead of NPE

LonwoLonwo:
2022-04-15 12_42_27-Window


E1izabeth:
Maybe it's more correct to say: "Database object was not found"

PR #16221, opened 2022-04-18
#2407 Introduce `@export` command for running data transfer from SQL

### LINKED ISSUE #2407: Set default 'Export From Query' path using comment
I use a lot the *Export From Query* functionality. However, sometimes I accidentally overwrite a previous export because the path for the CSV file is kept from the previous run.

It would be great if a comment could be added on top of the query with a default path for that export. Something like:

```
-- @ExportPath /tmp/my_csv_file.csv.gz
SELECT * FROM ...
```

I have been checking the documentation but right now I think this is not possible.

ShadelessFox:
https://user-images.githubusercontent.com/35821147/164252883-76c4a938-5c95-498e-8bad-42806c4e73cd.mp4


Matvey16:
Verified

PR #16222, opened 2022-04-18
#15126 Close extra result tabs for single query

### LINKED ISSUE #15126: "Window -  Preferences - User Interface - Confirmations - Close extra result tabs"     not working
#### System information: 
- Operating system (distribution) and version: Win 10
- DBeaver version: 21.3.2.202112271055
- Additional extensions

- Database name and version: MySQL 8
- Driver name: mysql-connector-java-8.0.17.jar
- Do you use tunnels or proxies (SSH, SOCKS, etc)? no

When there are multiple result tabs in the view, then i use shortcut "Ctrl+Enter" to execute sql
the extra result tabs not closed

E1izabeth:
Previously opened tabs are prompted to close when executing script with Ctrl+Enter too (but not when executing with Ctrl+\). The same as it works for Alt+X (prompted) and Ctrl+Alt+Shift+X (not prompted) execution.


Matvey16:
Verified

PR #16223, opened 2022-04-18
#14731 Set minimum digits count based on type's precision

### LINKED ISSUE #14731: Dbeaver will show wrong data with column datatype "decimal(4,0) unsigned zerofill" (MariaDB)
Dbeaver (21.3.0) will show wrong data with column DataType ; "decimal(4,0) unsigned zerofill" with default value of  "0101"

It should show the value 

0101 

DBbeaver shows value as:
00000000000000000000000000000000000000000000101

PHP Myadmin and other tools are showing the value corectly

ShadelessFox:
image

DDL:
```sql
CREATE TABLE `issue_14731` (
  `id` int(11) NOT NULL,
  `value` decimal(5,2) unsigned zerofill NOT NULL,
  `value2` decimal(5,0) unsigned zerofill NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1;
```


Matvey16:
Verified

PR #16254, opened 2022-04-20
#15656 Show count of distinct values in the filter panel

### LINKED ISSUE #15656: In the filter-per-column pane, show the total number of distinct values
When I open the filters pane on a column from the table view page, I want to see how many distinct values there are in that column. That is to say, I want to see how many rows appear in the aforementioned filter pane.

At the bottom of the aforementioned filter pane, I want to see a comment like "5 distinct values".

Could try to match Power BI's distribution-per-column charts. They're really good.

**Additional context**
Example screenshot: 
image

ShadelessFox:
image


Matvey16:
Verified

PR #16265, opened 2022-04-21
dbeaver/dbeaver-ee#1594 make sizes of files and folders human-readable

flambchap:


bundles/org.jkiss.utils/src/org/jkiss/utils/ByteNumberFormat.java
```diff
@@ -163,4 +163,9 @@ private enum Unit {
             this.fullName = fullName;
         }
     }
-}
\ No newline at end of file
+
+    public enum UnitNames {
+        SHORT,
+        FULL,
```
ShadelessFox: Please add javadoc comments describing differences between them.

bundles/org.jkiss.utils/src/org/jkiss/utils/ByteNumberFormat.java
```diff
@@ -163,4 +163,9 @@ private enum Unit {
             this.fullName = fullName;
         }
     }
-}
\ No newline at end of file
+
+    public enum UnitNames {
```
ShadelessFox: Perhaps `UnitNaming` would be more convenient?

bundles/org.jkiss.utils/src/org/jkiss/utils/ByteNumberFormat.java
```diff
@@ -31,13 +33,21 @@ public class ByteNumberFormat extends NumberFormat {
 
     private static final DecimalFormat fpFormat = new DecimalFormat("#.#");
 
-    private boolean useLongUnitNames = false;
+    /**
+     * Determines whether this formatter should use long names for units.
+     * Note that this formatter uses ISO 80000 compliant names.
+     */
```
ShadelessFox: This can be avoided by having javadoc on the enum itself.

bundles/org.jkiss.utils/src/org/jkiss/utils/ByteNumberFormat.java
flambchap: I don´t see the point tbqh

ShadelessFox: Or `UnitStyle` to be consistent with the `java.time.format` API naming scheme because I like it.

Reference:
https://docs.oracle.com/javase/8/docs/api/java/time/format/FormatStyle.html

flambchap: I think I'll change the name to match the language used in ISO 80000-13 (I'm going to use such language in Javadoc anyway)

Issue #6536, opened 2019-08-16
Smooth Scrolling in Query Results Pane


jim256:
When browsing the results of a query (whether a query I wrote or the default that's run when I open a table and go to the "Data" tab), the fact that the scrolling is jumpy makes it difficult to keep track of where I'm looking.

The horizontal scrolling is definitely the most frustrating for me personally. When trying to scroll sideways using the Apple Magic Mouse or the trackpad, it will usually jump 2-4 columns over, which can be frustrating.

I'd like to see smooth scrolling (horizontal and vertical) implemented. For an example, see how Postico scrolling works.

serge-rider:
Agreed in general. Column-based scrolling makes much sense in some cases but having it as default (and the only) scrolling mode isn't good at all.

I've added smooth scrolling mode, And the preference option for it: 

![image](https://user-images.githubusercontent.com/6398845/67958952-0b89db80-fc09-11e9-9d1c-eaa5e006b8da.png)

It is enabled by default so smooth scrolling will be the default mode since 6.2.4.

jim256:
Awesome news! Really excited about that. Thank you!

kseniiaguzeeva:
Verified

Pengchenghua:
> 已验证

WIN10 Function is invalid

kseniiaguzeeva:
Thank you for the buf report. Investigating.

kseniiaguzeeva:
Found out that the function is disabled by default in PostgreSQL (but it can be activated). Please tell which database do you use?
![7KwZTWxDD6](https://user-images.githubusercontent.com/51405061/68181368-15fcfa00-ffa8-11e9-9168-927d1d5f2833.png)

serge-rider:
@kseniiaguzeeva it is not disabled for PostgreSQL. Perhaps it was changed for a particular connection?

kseniiaguzeeva:
My mistake. Now I see, it is not set to default if Restore defaults.

kseniiaguzeeva:
@Pengchenghua please reopen if you still reproduce it. In that case I kindly ask you to right more details.

RecycleBin-X:
This function is not working for Mac version somehow

kseniiaguzeeva:
I cannot reproduce. Please be sure that the function is enabled (Properties>Editors>Data Editor>Presentation):
[...]

PR #16275, opened 2022-04-21
#6536 #10184 macOS Horizontal Scroll Issue

### LINKED ISSUE #10184: MacOS: Results Grid View - Smooth (Pixel-by-Pixel) Scrolling?
Hello! Wonderful project, thank you in advance for all of the great work that's been put in to make DBeaver so awesome!

As a MacBook Pro user, scrolling through the results grid with the trackpad accurately is EXTREMELY difficult to do. Frequently do I find myself 20 columns to the right/left of where my scroll position originally began (when only wanting to scroll vertically), and my trackpad sensitivities are not set above default levels for the MacOS operating system. Microscopic gestures in the wrong direction of your desired scroll will send you *wildly* off track in large data sets.

If the option to disable scroll stops on column and row endings existed, it would be much easier to control my scrolling.

I think it would be a great idea to have the option to enable pixel-by-pixel (precise) scrolling in the results grid view to both the vertical (row) and horizontal (column) axis. This feature should have the ability to be toggled on/off, since use cases exist for scroll stops being locked to row & column endings.

There are not any alternatives to this issue, as the ability to control scroll behavior exists only within DBeaver's backend.

**Additional context**
Trackpads that support advanced gestures are extremely handy, and as time goes on they are becoming more and more precise with the input they provide to the operating system. Utilizing these features would go a long way to the user experience within DBeaver.

If i'm alone in feeling this way, or a solution already exists that I missed in my original research before posting this, please let me know and I'll do my best with your advice :smile:

Thanks again!

Andrew0002:
The method scrollHorizontally() is called on event SWT.MouseHorizontalWheel. ~~In my testing, it seems like this event is being raised on macOS, but not Windows.~~ The method overrides normal horizontal scrolling and forces either the leftmost or rightmost visible column to be fully visible - depending on whether the user is scrolling to the left or right. This causes scrolling to feel "jumpy". If the user has columnScrolling disabled (i.e. Smooth Scrolling enabled), this method probably shouldn't be called.

The value for HORZ_SCROLL_INCREMENT is only used if columnScrolling is disabled (Smooth Scrolling enabled) and ~~only~~ appears impacts the UI when responding to a horizontal mouse wheel or if the user clicks the arrows on the scrollbar. The reason for increasing this value is that previously horizontal scroll was moving a column at a time. With this change, the scroll wheel is moving pixel by pixel, and that made scrolling extremely slow. I tested a range of values on a 2019 MacBook Pro and a Microsoft Surface Laptop 4 (in order to test with a Windows Precision Touchpad enabled device), and a value of 12 gave the closest approximation to scrolling in applications like Excel. Behavior was unchanged when using a basic wired mouse.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/LightGrid.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/LightGrid.java
@@ -62,7 +62,7 @@ public abstract class LightGrid extends Canvas {
     /**
      * Horizontal scrolling increment, in pixels.
      */
-    private static final int HORZ_SCROLL_INCREMENT = 5;
+    private static final int HORZ_SCROLL_INCREMENT = 12;
 
     /**
      * The area to the left and right of the column boundary/resizer that is
@@ -3023,7 +3023,7 @@ public abstract class LightGrid extends Canvas {
     }
 
     private void onMouseHorizontalWheel(Event e) {
-        if (hScroll.getVisible()) {
+        if (hScroll.getVisible() & columnScrolling) {
             scrollHorizontally(e.count);
             e.doit = false;
         }


Andrew0002:
Updated my original post after some additional testing. SWT.MouseHorizontalWheel does fire on Windows, my mistake.  Also noticed that the change to HORZ_SCROLL_INCREMENT also impacts if a user clicks the arrows on the scrollbar. Neither of these are problems, I just wanted to have my facts straight 😁.

serge-rider:
Closing as it doesn't resolve the issue

We keep inbestigating it.

Andrew0002:
@serge-rider I've been using it on my Mac for over a week and it resolves the issue for me. I'm curious what configuration you tested with where this didn't fix the issue? Would it be helpful if I uploaded a video?

serge-rider:
Well, if it fixes the problem in your env then we'll give it another chance.

We'll try to test it on different machines.
Thanks.

Andrew0002:
@serge-rider Here's a short video that shows the current build vs. when I built it with my changes. This is on my 2019 MacBook Pro. I also tested on a 2021 MacBook Pro. Both running macOS 12.3.1. 

[Screen Recording on Vimeo](https://vimeo.com/705833553/fb644bdc69)

serge-rider:
The fix will be included in 22.0.5.

PR #16278, opened 2022-04-21
#7576 Don't populate parameters which have already been set

### LINKED ISSUE #7576: Bugs in script > export using @set with multiple tables
Running a file with @set and multiple tables/queries is still buggy in version 6.3.2 (Windows)

It still seems to run all of the @set commands at once, before running any of the queries, instead of running them in the given sequence.

My scenario:  I have one Orders table each day.  Orders_20191230, Orders_20191231, Orders_20200103, etc.

My script (simplified):

@set t=_20191230
select ${t} as orderdate, count(*) as ordercount from Orders${t};

@set t=_20191231
select ${t} as orderdate, count(*) as ordercount from Orders${t};

@set t=_20200103
select ${t} as orderdate, count(*) as ordercount from Orders${t};

I highlight all of this and click Export From Query to run it.

Problem 1:
If I use a filename format that has ${timestamp}, it only outputs one file, with one query, and the last timestamp.  
(The export dialog shows me 3 target filenames with different timestamps before I execute the script -- where does it get the timestamps??  But it only outputs one file.)

Problem 2:
I added ${index} to the filename.  Now I do get three separate files.
But all three files have the data for the last query, last date -- the three files are identical except for the filename.  I get no data for the first two dates.

#### System information: 
Windows 10
DBeaver 6.3.2

Postgres

E1izabeth:


Matvey16:
Verified

PR #16287, opened 2022-04-22
#15100 Fix 'create' connection parameter

### LINKED ISSUE #15100: Unable to save connections made from CLI using connection parameters
When creating a connection using the cli as per https://github.com/dbeaver/dbeaver/wiki/Command-Line#connection-parameters. 
The connections do no persist once the dbeaver client is closed. Can see manual creations from ui in the data-sources.json are persistent 

tried with create = true even though its default but connections never save in ui unless manually created

`dbeaver -con "name=${NAME}|driver=Test|host=${HOST}|port=${PORT}|database=${DATABASE}|user=${MSID}|${SSL}|${SSL_PASSW}|folder=Test|connect=false|create=true"`

option to save the cli connection string in the data-spurces.json file with the manual created in ui

reference alternative saved file which opens for each client open

**Additional context**

E1izabeth:


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceUtils.java
```diff
@@ -171,6 +172,12 @@ public static DBPDataSourceContainer getDataSourceBySpec(
                 case PARAM_AUTO_COMMIT:
                     autoCommit = CommonUtils.toBoolean(paramValue);
                     break;
+                case PARAM_CREATE:
+                    createNewDataSource = CommonUtils.toBoolean(paramValue);
+                    if (parameterHandler != null) {
+                        parameterHandler.setParameter(paramName, paramValue);
+                    }
```
ShadelessFox, line 179: What does it do? I don't see such logic for the parameters above.

E1izabeth: It's one of three parameters which is stored in `InstanceConnectionParameters implements GeneralUtils.IParameterHandler`. And this parameter is not the same as others in non-default cases. It used to be handled in default switch branch and set to `parameterHandler` in the last `else` branch before fix.

PR #16290, opened 2022-04-24
#16132 Fix "Words that start with underscore are treated as keywords"

### LINKED ISSUE #16132: Words that start with underscore are treated as keywords
#### System information: 
- Operating system (distribution) and version: Fedora 35
- DBeaver version: Version 22.0.2.202204041542
- Additional extensions: N/A

- Database name and version: Postgresql 12, 13
- Driver name: Postgres
- Do you use tunnels or proxies (SSH, SOCKS, etc)? NO

Keywords, when prefixed with non-literal characters are still highlighted as keywords. 
While a lot of those non-literals are not valid as starting characters for variables, underscore is an accepted one.
This results in highlighting words like _sql, _default as to be keywords. Which is obviously not true.
See the image below.
it's just a screenshot of a fragment of the code. Note % as variable name prefix will not work likely. Just to show behavior pattern.
image

See above

N/A

PILINING:


PILINING:
#16132

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -316,7 +316,7 @@ public String[] getInnerBlockPrefixes() {
 
     @Override
     public boolean isWordStart(int ch) {
-        return Character.isUnicodeIdentifierStart(ch);
+    	return Character.isUnicodeIdentifierStart(ch) || ch == 95;
```
E1izabeth:
```suggestion
        return Character.isUnicodeIdentifierStart(ch) || ch == '_';
```

E1izabeth (changes requested):
HI, @PILINING! Thank you for contribution! Could you, please, make mentioned changes to make it clearer?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
PILINING: Hi, @E1izabeth, I have made the change, thanks for the suggestion!

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -316,7 +316,7 @@ public String[] getInnerBlockPrefixes() {
 
     @Override
     public boolean isWordStart(int ch) {
-        return Character.isUnicodeIdentifierStart(ch);
+    	return Character.isUnicodeIdentifierStart(ch) || ch == '_';
```
ShadelessFox: Please use four spaces instead of tabulation

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
PILINING: Thanks for the reminder, it's been modified.

E1izabeth (approved):

ShadelessFox (approved):
👍🏻

PR #16292, opened 2022-04-24
Adding the feature of Align numeric values on decimal

### LINKED ISSUE #16152: Align numeric values on decimal in grid-view
Related to #4512, I'd like for data in the results pane to align on the decimal place rather than being right aligned. 
```
  have    |     want  
 -------  |   -------
  1.234   |     1.234
  12.34   |    12.34
  123.4   |   123.4
   1234   |  1234.0
```
Enabling "Use data type scale for fraction digits" seems to mostly handle this but it doesn't always work. 
It seems the alignment becomes out of sync when maximum fraction digits is set > 5.

image

And the text view
image

Tazhio:
Fix https://github.com/dbeaver/dbeaver/issues/16152 by adding the feature of Align numeric values on decimal in n grid-view, as well as text-view.
This is implemented by padding white space for the data is user checked "Align with decimal" in perferences. The whitespace will be removed if user choose to unchecked it.
There may be better way to achieve this.

Attention to be paid: Current method may not work well with long numbers, and there are excessive whitespaces. This happens because I haven't figured out a way of getting all the data of a specific column then figure out length of a grid.

This can be fixed by adding UI in the preferences and let users tell us the estimted length of the numbers, though the best approach will be figuring out the needed white space automatically.
I am willing to keep working on those feaures. But that will be many, many changes I guess.

I am sorry that I accidentally git added files that shouldn't be commited. I haven't found good ways of removing them myself, is it possible that you only accept part of this branch?


E1izabeth:
Hi, @Tazhio! There is no way to accept only part of changes from pull request. You can revert files, which you don't want to change, to the revision where they are in initial state.

ShadelessFox:
Hello @Tazhio,

Can you please attach a screenshot showing how the introduced option affects the output?

Tazhio:
> Hi, @Tazhio! There is no way to accept only part of changes from pull request. You can revert files, which you don't want to change, to the revision where they are in initial state.

Sorry for the late reply. I think may work on the latest branch then make my changes again

Tazhio:
> Hello @Tazhio,
> 
> Can you please attach a screenshot showing how the introduced option affects the output?

Sorry for the late reply. Here are the screenshots for the output:
164975298-ac12fa37-bca1-4661-8cf1-e911b89f570a
164975329-290c3f3b-6edc-446c-a2f7-cb778810b5eb
164975337-7b7f5a1a-16b2-445f-9736-30d8d6f7f786

E1izabeth:
Hi, @Tazhio! Please, remove all unintended changes from this PR unless we can't accept it. Also you need to merge changes from devel to your branch.

ShadelessFox:
Hello @Tazhio,

This PR looks stale.

Please fix all conflicts and remove unnecessary changes. Otherwise, it will be closed.

PR #16298, opened 2022-04-25
#15117 fix drivers for folder without parents

### LINKED ISSUE #15117: Dbeaver 21.3.2 not loading
Application not loading at startup itself. BELOW IS THE DATA FROM LOG. 

!MESSAGE Problems occurred when invoking code from plug-in: "org.eclipse.equinox.preferences".

Destrolaric:
Closes #15117


Matvey16:
Drivers need to be downloaded into `C:\Users\usr_name\AppData\Roaming\DBeaverData\drivers` but now they are downloaded into `C:\Users\usr_name\AppData\Roaming\DBeaverData\workspace6\drivers` which can lead to excessive storage use as well as other hidden issued.

Matvey16:
Verified

PR #16303, opened 2022-04-25
#13501 Select spreadsheet cells according to selected features on a map

### LINKED ISSUE #13501: select records using spatial viewer
When one of more cells with spatial data are selected the spatial data are shown on a map. This is a very useful feature. It would be even more useful if it could also be used the other way around: selecting records with spatial data graphically.

When I have a table data frame opened (with or without a filtered selection) and click the spatial column, all records are selected and shown on a map. I can zoom and pan the map and click on items, but it is not possible to select a number of features by using for example Ctrl-mouse. Maybe a bit like Shift-mouse (used for zooming), but then add the selected area to the filtered selection.
An optional empty map could opened from the context menu after right-clicking the spatial column.

ShadelessFox:
https://user-images.githubusercontent.com/35821147/165108424-f43956d0-ecbd-42cf-a01a-7227c4445b15.mp4

Features:
- Map features can be selected by clicking on them
- By holding CTRL, the selection will be extended
- Pressing ESC or clicking outside features will clear the selection


uslss:
verified

Issue #16009, opened 2022-03-30
Write to the single file, different table SQL should start at a new line


ny5667:
#### System information: 
- Operating system (distribution) and version:ubuntu
- DBeaver version: 22.0.1.202203181646
- Additional extensions

#### Connection specification:
- Database name and version

- Driver name
- Do you use tunnels or proxies (SSH, SOCKS, etc)? 

#### Describe the problem you're observing:
Write to the single file, different table SQL should start at a new line
![IMG20220330190633](https://user-images.githubusercontent.com/82657677/160821225-a6a0eb0f-423e-4103-9891-8978ac1eb0d2.jpg)
![屏幕截图_1](https://user-images.githubusercontent.com/82657677/160819496-20908aff-9420-4174-8077-d6e065e53017.png)
![屏幕截图](https://user-images.githubusercontent.com/82657677/160817649-e72a34f8-20e3-4ce4-8716-8fb4454fc302.png)

LonwoLonwo:
Unfortunately, can't reproduce.

![2022-03-31 11_05_24-export_202203311104 sql — H__asya_special_dbeaver_backup — Atom](https://user-images.githubusercontent.com/45152336/161007681-7bf06b70-44d2-4ff2-ad87-80371f572a60.png)

Can you provide please screenshots of setting from these two setting pages?

![2022-03-31 11_06_35-Data Transfer](https://user-images.githubusercontent.com/45152336/161007934-41f34490-66b0-4209-9d21-88d8b94eea94.png)

ny5667:
Ok, here is the screenshots of settings.
Extraction settings:
![Snipaste_2022-04-01_18-23-44](https://user-images.githubusercontent.com/82657677/161245707-a3b76022-5d52-4ed1-8edb-3397f6b6f922.png)
Format settings:
![Snipaste_2022-04-01_18-23-54](https://user-images.githubusercontent.com/82657677/161245732-95aaa8a4-8a8c-4a0b-8a17-4bf90f12b94f.png)

LonwoLonwo:
Ok, I can reproduce with the "Data rows per statement" = 1 setting.
Thanks for the bug report.

![2022-04-03 21_31_49-export_202204032130 sql — H__asya_special_dbeaver_backup — Atom](https://user-images.githubusercontent.com/45152336/161442874-a37796b9-c7ed-4ba9-bcc3-644357bdbf1a.png)

CalmaKarma:
[...]

PR #16310, opened 2022-04-25
#16204 Append exported data to existing .sql files 

### LINKED ISSUE #16204: Append exported data to existing .sql files
We have a feature to append exported data to csv, txt, xlsx files. We need to expand this feature for SQL files.

CalmaKarma:
Implement a feature enhancement [#16204](https://github.com/dbeaver/dbeaver/issues/16204), make appending to existing files and exporting to a single file available for SQL file exporting to SQL files.
Fixed a bug [#16009](https://github.com/dbeaver/dbeaver/issues/16009), so that appending files are correctly starting new lines instead of appended to the last SQL statement of the previous files.

### DIFF
--- a/plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterSQL.java
+++ b/plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterSQL.java
@@ -16,6 +16,7 @@
  */
 package org.jkiss.dbeaver.tools.transfer.stream.exporter;
 
+import org.jkiss.code.NotNull;
 import org.jkiss.dbeaver.DBException;
 import org.jkiss.dbeaver.Log;
 import org.jkiss.dbeaver.model.*;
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterSQL.java
```diff
@@ -374,12 +376,17 @@ public void exportFooter(DBRProgressMonitor monitor) {
...
             }
-        } else if (!oneLineEntry){
+        } else if (!oneLineEntry) {
             if (rowCount > 0) {
                 addOnConflictExpression(out);
                 out.write(";");
+                out.write(rowDelimiter);
+            }
+        } else {
+        	if (rowCount > 0) {
```
ShadelessFox: Let's merge nested `if` into a single `else if` :^)

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterSQL.java
```diff
@@ -45,7 +47,7 @@
 /**
  * SQL Exporter
  */
-public class DataExporterSQL extends StreamExporterAbstract {
+public class DataExporterSQL extends StreamExporterAbstract implements IAppendableDataExporter{
```
ShadelessFox: Gotha, please include space before `{`.

ShadelessFox (changes requested):

CalmaKarma:
I changed the whole if-else structure since there is a common `if` statement in all previous `if` branches.
That changed line of comment is by mistake. I will change it back with a commit.

ShadelessFox (approved):
👍🏻

Matvey16:
Verified

PR #16315, opened 2022-04-26
#15322 Fix saved custom where filter reading

### LINKED ISSUE #15322: date=CURRENT_DATE filter lost on restart despite saving
#### System information: 
- Windows 10 Professional x64 - CCW1064V64
- DBeaver version 21.3.1.2021122130652
- Additional extensions: none

- Database name and version: sqlite3

I use dBeaver to work on a sqlite3 database with a single table called bailPositions.
I want the bailPositions table to always open with a filter on the date field, date=CURRENT_DATE.
Other column settings persist across sessions, but not the filter on the date field.

I'm aware of issue #4166, but this is different because all my column settings (order, formatting, pinning, etc.) are being preserved, just not the filter.

I input that filter in the filter box like so:

image

It filters the table properly. Then I click "Save filter settings", and a notification says "data filter was saved".

But when I close out then open dBeaver again, the table is not filtered by date=CURRENT_DATE (and no filter in the filter box).

My saved-data-filter.xml _does_ show the filter:
`<filter objectId="positions.sqlite3@@/@@bailPositions" anyConstraint="false" where="date=CURRENT_DATE">`

This is the only message in the logs that reads like maybe an error:
`!MESSAGE Column '"pkNumber" AUTOINCREMENT' not found in table 'bailPositions' for PK 'bailPositions.BAILPOSITIONS_PK'`

pkNumber is a field in my table that autoincrements. Everything in the table works fine so I don't know what this means.

E1izabeth:


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/DataFilterRegistry.java
```diff
@@ -264,6 +264,7 @@ public void saxStartElement(SAXReader reader, String namespaceURI, String localN
                 case "filter": {
                     curSavedDataFilter = new SavedDataFilter();
                     curSavedDataFilter.anyConstraint = CommonUtils.toBoolean(atts.getValue("anyConstraint"));
+                    curSavedDataFilter.where = atts.getValue("where");
```
ShadelessFox, line 267:
```suggestion
                    curSavedDataFilter.order = atts.getValue("order");
                    curSavedDataFilter.where = atts.getValue("where");
```

ShadelessFox (changes requested):

ShadelessFox (approved):

uslss:
verified

Issue #16087, opened 2022-04-05
Postgres database not visible


brunoalreis:
#### System information: 
- Windows 11
- DBeaver version 22.0.2.202204021640

#### Connection specification:
- Postgresql 14
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No

#### Describe the problem you're observing:

Not possible to see the following database: Test.Api.Local
When I create a new database with this name, it is visible. If I Refresh databases, it disappears.

#### Steps to reproduce, if exist:
1 - Create database with name: Test.Api.Local
2 - Refresh databases

LonwoLonwo:
Hello @brunoalreis 

By default, DBeaver shows only one database.

So maybe enabling the "Show all databases" in Connection Settings can help you?

![2022-04-05 17_21_08-Window](https://user-images.githubusercontent.com/45152336/161775950-c9e5a186-91a8-47bb-8864-7a32208f5939.png)

brunoalreis:
Yes, it worked.

Thanks.

LonwoLonwo:
Thanks for the answer @brunoalreis

PR #16330, opened 2022-04-27
#16087 recreate editors on mode switch instead of hiding

### LINKED ISSUE #16078: Timestamp values are automatically modified in data editor
#### System information: 
- Linux Mint 20.3
- DBeaver 22.0.2

- Postgresql 12.9-0ubuntu0.20.04.1
- PostgreSQL JDBC Driver 42.2.16
- SSH tunnel

In data editor when only "browsing" with cursor over timestamp values some values are automatically marked as modified - background color changes and when I close the table I got the confirmation window for Persist changes.

See example - SQL with random timestamps and only KeyDown pressed.

test

Create table with any timestamp values or execute SQL:
```
SELECT current_timestamp + random() * interval '1 day' 
FROM generate_series(1, 1000);
```
Browse data with mouse click or keys...

This behaviour is since last few versions - something 22.0.0+ or version just before. I can find last version which worked ok if desired.

Destrolaric:
DateTime editors on Linux have an issue with sending selection events when they are invisible, we will change it by disposing of not needed editors on mode switch. Also we don't want to trigger Modify event when setting date to text editor
Closes #16078


Matvey16:
Now timestamp values with calendar view just don't show actual data

VirtualBoxVM_Bl01NYWcka

Matvey16:
Now calendar stopped working in external editor <kbd>Ctrl + Enter</kbd>

VirtualBoxVM_hIkn7YaFyj

uslss:
verified

PR #16334, opened 2022-04-27
#13305 Sync script file association with script active datasource on a toolbar

### LINKED ISSUE #13305: Default data source association
version: dBeaver community, Mac OS version 21.0.2.202104042040 5-april-2021

I have dozens of datasources and hundreds of scripts. All scripts are organized in folders.
When I create a new script this is my current workflow:

- Right mouse click on a folder in the 'scripts' tree, choose create, choose new file.
- this file is named 'script-XX' and a tab is opened with "none" association
- Rename the file. This can be done in the tab or in the folder tree.
- Now I need to associate the data source. This can't be done in the tab, but if I do this in the projects scripts map, the tab itself stays in association [none]
- Also the closing of the tab is mandatory: Associating in the tree view and opening the same tab does not work: The tab stays in [none] association.
- So I need to re-open the file. However if I close the tab, for some magical reason, dbeaver deletes the file. So I insert some garbage text into the new script to avoid this
- Now I can associate the datasource by right-clicking the file in the project tree.
- Now I need to re-open the file to get an associated tab
- Now I need to delete the garbage and I can start with the new file.

So it takes 9 steps just to create a new file. This is silly. 

So I tried alternatives:
- Create a new script by right-clicking on the datasource, sql editor, new script
- Now I get a new script on the top of my scripts folder
- I can rename the file but not drag it to the right folder as they do not open automatically
- I need to open the correct destination folder first
- But the number of objects in the project tree is too large so I can't drag drop the file to the right location (there is no drag-scolling implemented in this window)
- So I can copy-paste the file. There is no cut-paste option
- But now the association is gone again. So I end up again with a non-associated file, and need to follow the previous 9 steps again.

There can be multiple solutions, and maybe there is already a way to do this that I did not find in the hundreds of dbeaver options.
- Best solution: When creating a new file in the scripts folder of the project tree, ask the name of the file and the association. This would reduce 9 steps to 1
- 2nd option: Immediately ask for association. Never associate with [none]
- 3rd option: Allow for a default association
- 4rd option: Allow for association on the tab right-click menu 
- 5th option: sync the script tree file association with the tab-association 
- 6th option: Allow cut/paste in the project scripts and/or implement drag-scrolling and auto-folder-opening
- 7th option: Allow 'open file location' with finder in the project tree so you can use finder to do movements 

Here are some 

https://user-images.githubusercontent.com/13032529/126783549-ee4dcee2-2e8a-405e-a59f-aed1541c2994.mp4

E1izabeth:


uslss:
verified

PR #16336, opened 2022-04-27
#6012 add active project pattern resolving for script title

### LINKED ISSUE #6012: [FR] Tab Title Pattern - projectName variable
#### System information: 
- Operating system (distribution) and version - Arch Linux
- DBeaver version - 6.0.3

- Database name and version 10.3
- Driver name mariadb
- Do you use tunnels or proxies (SSH, SOCKS, etc)? yes

In the tab title pattern would it be possible to have a projectName variable?

https://github.com/dbeaver/dbeaver/issues/1502

My scripts are split into projects and can be run against different connections would be nice to be able to see which tabs belong to which projects

Destrolaric:
Closes #6012


uslss:
New variable isn't mentioned in the tooltip
image

Probably, it would be more expectedly if the variable would be named _projectName_ (already existing variables - connectionName, fileName)

uslss:
verified

Issue #14858, opened 2021-12-09
Enabling date/time formatting doubles values in the column filter for dates


kseniiaguzeeva:
#### System information: 
- Operating system (distribution) and version Mac
- DBeaver version 21.3.1 EA

#### Steps to reproduce, if exist:
1. Enable date/time formatting in the connections settings
2. Open the filter in the column header for a date column (e.g. DB2 `DB2INST1.PURCHASEORDER` column `ORDERDATE`)

**Expected**: the number of filtered values ​​is equal to the number of unique values ​​in the column
**Actual result**: each unique column value is doubled by the same quoted value

![Screenshot 2021-12-09 at 17 26 10](https://user-images.githubusercontent.com/51405061/145417493-65141a56-a5ee-4dbd-bd44-ee654c9700af.png)

Matvey16:
Still reproducible in Oracle

![image](https://user-images.githubusercontent.com/82543000/165732650-734e999b-70b7-4718-9913-f2fd08cec8a9.png)

PR #16339, opened 2022-04-27
 #14858 add more checks for date/time data types for filters

LonwoLonwo:
2022-04-27 17_51_45-Filter by 'TimestampTz6_Column'


Matvey16:
Verified

Issue #15334, opened 2022-01-31
Setting to show JSON View in collapsed state


Mikescher:
Hello,

currently the JSON view (eg for MongoDB) always shows all items fully expanded.  

This is annoying when you have big objects in the database or big arrays in some objects.  
On one hand it makes the JSON view really unwieldy to use, and when the objects get really big it even starts to lag 

Other MongoDB clients (eg MongoDB Compass) collapse all subnodes by default and you have to manually expand them

In my opinion an option to collapse all nodes in the JSON view by default (or even just doing it without a setting) would be really useful.

&nbsp;

Regards
 ~ Mike

ShadelessFox:
Hello @Mikescher,

It sounds like a reasonable improvement.

Thank you.

ShadelessFox:
Associated PRs: https://github.com/dbeaver/dbeaver/pull/16379, https://github.com/dbeaver/dbeaver-ee/pull/1615

ShadelessFox:
Implemented:

![image](https://user-images.githubusercontent.com/35821147/165987225-f5c27a8a-dafd-461c-875b-08870568a3bb.png)

PR #16379, opened 2022-04-30
#15334 Load preferences before disabling page controls

ShadelessFox:


Matvey16:
Can't start DBeaver with current fix

PR #16381, opened 2022-05-01
dbeaver/dbeaver-ee#1563 Shortcut for exporting the ERD diagram

ShadelessFox:
1. A separate command was introduces for this action, so it now can be bound to a shortcut.
2. An `Export diagram` (as ERD) action was replaced with `Export diagram in external format` action.


uslss:
verified

PR #16421, opened 2022-05-05
dbeaver/dbeaver-ee#1562 Propose most recently used filename during diagram export

E1izabeth:
Also fix export format selection for filepath containing dot character in the middle of the path (not in the filename itself)


uslss:
Custom ERD:
Name for saving in external formats is provided with extension (e.g. test.erd)

image

uslss:
verified

PR #16426, opened 2022-05-06
dbeaver/dbeaver-ee#1584 keep comments label color in case background …

LonwoLonwo:
…color changing

2022-05-06 10_58_16-DBeaver Ultimate 22 0 4 - Employee


Matvey16:
Verified

PR #16450, opened 2022-05-11
#16441 Add save argument to cli -con parameter

### LINKED ISSUE #16441: 'temporary connection' feature lost after upgrade to 22.0.4
#### System information: 
- Windows 11 x64
- DBeaver 22.0.4

- SQLite database file
- sqlite

I added file association of *.db file to open with DBeaver sqlite driver by editing windows registry.

The behavior of opening sqlite db by double click on file has changed after upgrade to 22.0.4.

Before DBeaver 22.0.4, the connection is opened like a temporary connection, the text of connection name is italic.
After I restart dbeaver, the connections are removed from dbeaver. This is very helpful because I usually open a lot of sqlite db file.

But after I upgrade to 22.0.4, the connection is opened as a normal connection. I must clean the connection by deleting them manually.

Windows registry value:
sqlite3_db\shell\Open with DBeaver\command
C:\Program Files\DBeaver\dbeaver.exe -con "driver=sqlite|database=%1"

#### Expect

Can i have this feature again via settings or command line? Thanks.

E1izabeth:
Example of usage:
`-con "name=DS|folder=Con|driver=mysql8|database=mysql8|connect=false|create=true|save=true"`

`save=true`   - makes temporary connection
`save=false`  - connection will be saved between DBeaver launches

### DIFF
--- a/plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceUtils.java
+++ b/plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceUtils.java
@@ -65,6 +65,7 @@ public class DataSourceUtils {
     private static final String PARAM_FOLDER = "folder";
     private static final String PARAM_AUTO_COMMIT = "autoCommit";
     private static final String PARAM_CREATE = "create";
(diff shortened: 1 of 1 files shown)


uslss:
verified

PR #16470, opened 2022-05-12
#16129 Generic JDBC URL template for matching non-genuine URLs

### LINKED ISSUE #16129: Import Connection String in "Connect to a database" screen
I have to add databases to my DBeaver a lot, multiple times a day. I get the connection details of these databases as a connection string of the form `postgresql://user:password@hostname:port/database`. Every time I paste that into the `Host` input field, then copy the database to the `Database` field, the port to `Port` and finally `Username` and `Password` as well. Then I remove the protocol and leftover `:@/` from `Host` and can test the connection - and then continue.
 
It would be super neat if there was an "Import connection string" button where I get a windows with a text field where I can paste the connection string, then click "Parse and Import" (or similar) and the window closes and the parse parts of the connection string are filled in.

(Even more advanced (and less UI clutter) would be if one could just paste a connection string into the `Host` input field, it would detect that this is a connection string and not just a hostname and parse it and move the data into the appropriate fields automatically.)

Any other alternative I could think of would not actually be much faster than my current workflow (E.g. external tool that splits the connection string into its segments to copy paste or similar) unfortunately.

ShadelessFox:
image

A small lamp-like overlay is shown on drivers matched by a non-genuine URL. Somewhat useless, but looks cool.


uslss:
verified

PR #16489, opened 2022-05-16
#16402 Add option to replace current result set tab on single query execution

### LINKED ISSUE #16402: Execute SQL queries always open new result tabs after executing a query in a new tab
#### System information: 
- Operating system (distribution) and version: Ubuntu 20.04
- DBeaver version: 22.0.4

- Database name and version: PostgreSQL 13.6
- Driver name: PostgreSQL JDBC driver (version 42.3.4)
- Do you use tunnels or proxies (SSH, SOCKS, etc)?  No

When I execute a query in a new result tab, all query executions after that always open an new result tab, even if I don't choose the "Execute SQL in a new tab".

1. Execute a SQL query => the result appears in a single tab.
2. Execute a new query in a new tab => a new result tab appears with the new query
3. Execute a SQL query (not in a new tab) => the result systematically displayed in a new tab.

What I expect is that the last query result appears in the last focused result tab.

I did not find any preference to custom this behavior (but perhaps didn't I find it...).

Kind regards,
Cedric

E1izabeth:
image


uslss:
verified

PR #16502, opened 2022-05-17
#16403 search project node by project UUID

### LINKED ISSUE #16403: No active project is chosen in case of enabling Reset UI settings in the installer
#### System information: 
- Operating system (distribution) and version win10
- DBeaver version 22.0.4

DB Navigator Tab is empty and no project is bolded at the project tab in case Reset UI settings option was chosen in the installer.
Project General isn't shown as bold, but it is impossible to set It as active. DB Navigator is empty.
image
dbeaver_v16pD8Wve3

LonwoLonwo:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNRoot.java
```diff
@@ -180,8 +177,10 @@ public DBNProject getProjectNode(IProject project) {
     }
 
     public DBNProject getProjectNode(DBPProject project) {
+        UUID projectID = project.getProjectID();
         for (DBNProject node : projects) {
-            if (node.getProject() == project) {
+            DBPProject nodeProject = node.getProject();
+            if (nodeProject != null && (nodeProject == project || nodeProject.getProjectID().compareTo(projectID) == 0)) {
```
ShadelessFox, line 183:
```suggestion
            if (nodeProject == project || nodeProject.getProjectID().equals(projectID)) {
```

ShadelessFox (changes requested):

uslss:
verified

PR #16507, opened 2022-05-17
dbeaver/dbeaver-ee#1589 Keep temporary files after DND is finished

ShadelessFox:


serge-rider:
That's odd. We explicitly copy stream before delete. Maybe some flush/sync is needed? Or a 1-second delay?

serge-rider:
Having many big temporary files on disk is not good..

Matvey16:
Verified on Linux, needs further testing on MacOS

PR #16511, opened 2022-05-17
Added missing italian translations for dbeaver.model

gualask:


E1izabeth:
Thank you for contribution!

PR #16513, opened 2022-05-17
#16479 Allow to specify grouping size for the number formatter

### LINKED ISSUE #16479: Add "Digit group size" option for numbers data formatting
None, but it can provide better internationalization support.

Add an input component "`Digit group size`" when "Use Grouping" is checked:
A drop down list box, which contains at least two options:
 * 3
 * 4

Or, a combo box, which contains least two options: 3 and 4, and allow user input a custom `java.text.DecimalFormat` pattern string in it.

**Additional context**
https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping

ShadelessFox:
image


uslss:
verified

PR #16531, opened 2022-05-19
#13726 Add coloring by range for text

### LINKED ISSUE #13726: Coloring by range doesn't work for the text
#### System information: 
- Win10, DBeaver 21.2.0

Expected: Text color can be set as gradient in data editor
Actual: Text color set as the low bound color

https://user-images.githubusercontent.com/43252954/131300687-8ff963b4-47a5-4fd0-b402-1f9fd7c3504a.mp4

E1izabeth:
If both foreground colors are set:
image
image

If no foreground color is set:
image

If only one foreground color is set:
image


Matvey16:
Foreground color doesn't change if no background color is set

Matvey16:
Verified

PR #16536, opened 2022-05-19
#16346 QM search text hint fix

### LINKED ISSUE #16346: Search field in transaction log on MacOS has two tooltips on top of each other
#### System information: 
- Operating system (distribution) and version MacOS
- DBeaver version 22.0.3

1. Open transaction log in the top toolbar

**Expected:**Search field has only one tooltip
**Actual:** Two tooltips overlap each other

image

ShadelessFox:


Matvey16:
This removed the hint on Windows. It should be visible if nothing is typed in the search field. It works as expected on MacOS, but removes the hint on Windows if caret is placed in the search field (initial condition when transaction log is opened)
dbeaver_vkFE5Ee0gC

Matvey16:
This removed the hint on Windows. It should be visible if nothing is typed in the search field. It works as expected on MacOS, but removes the hint on Windows if caret is placed in the search field (initial condition when transaction log is opened)
dbeaver_vkFE5Ee0gC

Matvey16:
Seems to be working as expected

PR #16541, opened 2022-05-19
#16493 multithreading & fix bug with obstacles

### LINKED ISSUE #7257: Diagram relationship notation connect to columns
Dbeaver actualy uses only IDEF1X notation in DB diagram view and I am really missing the "connect to columns" available in MySQL Workbench. Even if Workbench is slow and especially buggy and not fully compatible with other DB, not even mariadb, I am still forced to use it because of this only feature. And I didn't find any better replacement yet. Dbeaver is fast, reliable and to me as a eclipse using developer is very intuitive and comfortable, but still I need some other tool to work with DB, if this feature is not implemented 
 
I am working with databases and especially with MySQL and later its fork mariadb more than 15 years and (re)designed couple of complex relationship dbmodels. By designing new DB model, or reversed engineering of existing one it is very important to see exact column relationships directly from the picture. Especially when foreign keys have not the same or similar names as the keys to which they are connected (not everybody is following good and readable naming conventions). The IDEF1X has not the ability to show it directly and get the right view. 

I would like to have the possibility to switch between the relationship notation views, like it is possible in MySQL workbench. 

If it is possible, I would like to contribute as developer or leader to implement this feature, because it will save me a lot of time switching between different tools in future development. So if is already someone working on this feature just contact me for any help that I can provide. 

Thanks

### LINKED ISSUE #16493: New routing method polishing
In the scope of the #15755 new method of erd path creation was added it still contains some moments that need to be improved:
1. Path created sometimes changes after a refresh
2. Line created on the border of the diagram sometimes glitches and start to repeatedly switch positions
3. Performance slows significantly if the entity  is dragged outside of the diagram boundary
4. Path manipulation may need some optimization

Destrolaric:


serge-rider:
Please describe what exactly was improved

Destrolaric:
Signifficant improvement to performance, especially path manipulation, now it should be good for large diagrams.
Now zoom and refresh action should not cause different paths to be shown
Fixed bug with some paths not being resolved due to them already having bendpoints.
Fixed bug with obstacles not being properly deleted.
Also little changes relating to pathbuilding, I think they should look a little more readable now

Issue #1551, opened 2017-04-08
Simultaneous queries


fayca121:
Hi,

I use Dbeaver 4.0.6 with oracle 11g database on windows 7, my problem is when I execute a query which takes a long time simultaneously with another one in different tabs (using the same connection), they take the same time of execution, even if the second query is too small (usually it takes less than 1s when it is executed alone).

Is it possible to use multiple-threads execution for queries (one connection per tab) ?

single query:

![alone](https://cloud.githubusercontent.com/assets/547548/24831672/2f791fba-1c96-11e7-824b-668280793425.png)

simultaneous queries: 

![sim2](https://cloud.githubusercontent.com/assets/547548/24831679/534cdc88-1c96-11e7-879f-a8c562f55a25.png)

![sim-1](https://cloud.githubusercontent.com/assets/547548/24831678/4cd814c6-1c96-11e7-8580-44b6a46a357e.png)

Thank you.

serge-rider:
You can execute multiple queries simultaneously using ctrl+alt+x (execute queries in separate tabs).

PR #16543, opened 2022-05-19
dbeaver/dbeaver#1551 check current exit value for the process

LonwoLonwo:


LonwoLonwo:
EE pr https://github.com/dbeaver/dbeaver-ee/pull/1638

PR #16554, opened 2022-05-20
#14326 fix connection for Netezza users with lower cased or specific …

### LINKED ISSUE #14326: Netezza username in lowercase
#### System information: 
- Operating system (distribution) and version Windows 10, 1909
- DBeaver version Version 21.2.3.202110151816

- Database name and version Netezza
- Driver name nzjdbc3.jar

trying to login with user name with dot example "a.user" causes
```
Error setting default catalog/schema for new connection

Reason:
SQL Error [1114] [08001]: LDAP authentication failed for user 'A.USER'
```
(DBeaver tries to use a.user (default names on netezza are uppercase), "a.user" should by used to force original spelling)

LonwoLonwo:
…names

2022-05-20 16_26_40-Window


Matvey16:
Verified

Issue #16556, opened 2022-05-20
Action icons can be unclickable on machines that use gtk


flambchap:
#### System information: 
- Operating system (distribution) and version: Pop!_OS 22.04 LTS
- DBeaver version 22.0.5 UE devel

#### Describe the issue you're observing:
If your machine uses gtk and you have a lot of connections in the navigator, you cannot click action icons.

#### Steps to reproduce, if exist:
1. Enable the action icons (Right-click on the navigator area -> Connection view -> Show action icons)
2. Make sure that you cannot see all your connections all at once, e.g., you can scroll them up and down 
3. Create or alter a connection, so it has an actions icon (make a connection read-only or create a cloud or over-SSH one)
4. Hover over the button. A vertical scrollbar will overlay the button preventing from clicking it.

PR #16557, opened 2022-05-20
dbeaver/dbeaver#16556 fix action icons' clickability on gtk

flambchap:


flambchap:
I kindly ask the QA engineers to check this fix on gtk, KDE, and cocoa (macOS).

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/StatisticsNavigatorNodeRenderer.java
```diff
@@ -601,4 +619,10 @@ private static boolean isHorizontalScrollbarEnabled(Tree tree) {
         }
         return horizontalBar.isVisible();
     }
+
+    private static boolean isNodeActionsEnabled() {
+        DBPPlatform platform = DBWorkbench.getPlatform();
+        DBPPreferenceStore preferenceStore = platform.getPreferenceStore();
+        return preferenceStore.getBoolean(NavigatorPreferences.NAVIGATOR_SHOW_NODE_ACTIONS);
```
ShadelessFox, line 626: A bit mouthful, I'd get rid of both local variables.

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/StatisticsNavigatorNodeRenderer.java
flambchap: I thought about it for a second. and I've settled on the current variant because, without these variables, the method doesn't fit on small screens.

LonwoLonwo (approved):

serge-rider:
Must be tested on Windows, MacOS and Linux

flambchap:
This approach leads to visual glitches. We need a better solution.

PR #16560, opened 2022-05-20
#16477 Connection settings + Readonly toolbar button

### LINKED ISSUE #16477: Connection Properties in inconsistent state after cancelling an operation from within the "edit connection" dialog
I found a bug:
 
1.	Create a read-only-connection
2.	Open SQL-Editor and type something.
3.	Edit the connection, and uncheck the read-only checkbox
4.	When the question pops up, click “Cancel”
image
5.	The connection is still read-only. However, if you open the connection tab using “edit connection” again, the checkbox is unchecked (but we are still using the read/only connection)

In general this happens whenever you cancel the pending “reconnect with changed settings” process.

E1izabeth:
- Change readonly and non-readonly connection icons
image
image
- Add confirmation to reconnect if there are active transactions
image
- Save changed settings for active connection only if reconnect is accepted to maintain consistent state


uslss:
connection type doesn't change if the connection isn't connected
dbeaver_sPP2TSlsnM

uslss:
commit transaction is asked even if there are no active transactions. And sometimes confirmation dialog for the one script is shown twice
dbeaver_lHqPRvCLbn

E1izabeth:
I tried to reproduce the issue about confirmation with no active transactions and found that there is active transaction on the video. The asterisk near the editor tab name points to it. The behavior in this case is the same as on editor close.

E1izabeth:
Issue about showing twice confirmation dialog exists for a long time. I guess we need to create a separate issue for this.
It happens on reconnect after connection settings changing when active transaction exists due to editor is dirty (dirty due to active transaction).

uslss:
verified

PR #16573, opened 2022-05-23
#14163 Copy object name from the context menu of the active editor tab

### LINKED ISSUE #14163: copy table name from context menu
In the database navigator is is possible to right click e.g. a column or table name and select "Copy" to select its name. The same can be done in the table data grid. In would be convenient if the table name can be selected from the context menu from the table tabsheet.
afbeelding

ShadelessFox:
java_9ql3Ca9sCc


uslss:
verified

PR #16574, opened 2022-05-23
#14397 Move shell if it can't fit the screen

### LINKED ISSUE #14397: Task wizard window goes out of bounds on narrow monitors
#### System information: 
- DBeaver version 21.2.3

Task wizard appears in the center of the screen, but when it gets initially resized the right side moves further than screen border if the monitor is not wide enough.

1. Open Dbeaver on a machine with narrow screen
2. Start creating a task

ShadelessFox:


Matvey16:
Verified

PR #16575, opened 2022-05-23
#16172 Set the document cursor to the mouse position after the RMB click

### LINKED ISSUE #16172: Context menu in SQL Editor doesn't respect cursor position on which it is called
#### System information: 
- DBeaver version 22.0.3

1. Type two different queries in SQL Editor (e.g. two selects from different table)
2. Left-click on one of the queries to put a text cursor on it
3. Right-click on another query and pick Export from query in the context menu
4. Perform export

**Expected:** Exported data is fetched from the query on which the context menu was called
**Actual:** Data is fetched from the query that had the caret prior to the right-click

dbeaver_tqH2Y3T4m9

ShadelessFox:
https://user-images.githubusercontent.com/35821147/169871592-4569f020-224c-477b-863b-bc6209588bb8.mp4


Matvey16:
Verified

PR #16578, opened 2022-05-24
#14471 apply checkbox change on activation on filter dialog

### LINKED ISSUE #14471: Selection of columns to display not working correctly
#### System information: 
- OSX Mojave
- DBeaver version 21.2.4
- Java version 8
- Additional extensions None

- Database name and version MariaDB 10.5
- Driver name
- Do you use tunnels or proxies (SSH, SOCKS, etc)? SSH

When I open a table and click on the 
"Resultset/Order filter settings" to select the columns to view, some of the fields, usually close to the last columns, refuse to be shown. When I try again, the selection is still unticked, even though I definitely ticked it. I can repeat this many times but it refuses to stick and the column does not show.

Also, the new way of ticking is incredibly frustrating. Why was it changed? The old way worker very well. The new way needs at least 3 clicks before the tick registers - sometimes a lot more. I wish I never upgarded DBeaver and I am beginnign to seriously look at alternatives.

Destrolaric:
Closes #14471

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/CustomCheckboxCellEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/CustomCheckboxCellEditor.java
@@ -43,7 +43,7 @@ import org.jkiss.utils.CommonUtils;
  */
 public class CustomCheckboxCellEditor extends CellEditor implements BooleanStyleDecorator {
 
-    private static final boolean CHANGE_ON_ACTIVATE = false;
+    private final boolean changeOnActivate;
 
     private Label checkBox;
     private boolean initialValue;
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/CustomCheckboxCellEditor.java
```diff
@@ -52,8 +52,16 @@ public class CustomCheckboxCellEditor extends CellEditor implements BooleanStyle
     private UIElementAlignment alignment;
 
     public CustomCheckboxCellEditor(Composite parent) {
+        this(parent, false);
+    }
+    public CustomCheckboxCellEditor(Composite parent, boolean changeOnActivate) {
```
ShadelessFox, line 58: Missing a new line

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/CustomCheckboxCellEditor.java
```diff
@@ -52,8 +52,16 @@ public class CustomCheckboxCellEditor extends CellEditor implements BooleanStyle
...
         super(parent);
-
+        this.changeOnActivate = changeOnActivate;
+        //added to prevent checkbox flicking when changeOnActivate is true
+        if (!changeOnActivate) {
+            checkBox.setBackground(parent.getBackground());
```
ShadelessFox:
```suggestion
        if (!changeOnActivate) {
            // added to prevent checkbox flickering when the changeOnActivate is set
            checkBox.setBackground(parent.getBackground());
```

ShadelessFox (changes requested):

serge-rider:
Let's add setter setChangeOnActivate instead of passing it in constructor

Matvey16:
Verified

PR #16581, opened 2022-05-24
#16023 Restrict restore task execution for readonly connection

### LINKED ISSUE #16023: Readonly connection still allows restore
#### System information: 
- Operating system: Windows 10 Pro 20H2
- DBeaver version: 22.0.1.202203181646

- Database name and version
- Driver name: MariaDB JDBC driver
- Do you use tunnels or proxies (SSH, SOCKS, etc)?  No

Setting a connection as read-only (connection -> edit connection -> general) prevents sql scripts from changing data, but does not prevent restore.

- Set connection to read-only
- Execute Tools -> Restore database

Expected behaviour:
A read-only connection should not be able to overwrite the whole database by performing a restore.

E1izabeth:
image


uslss:
verified

uslss:
it is possible to create and launch a restore task on read-only connection from the main menu
dbeaver_TuK2KAM8Nn

E1izabeth:
It is not executed if to confirm dangerous action execution.

PR #16595, opened 2022-05-25
#16172 Respect block selection when updating cursor position

### LINKED ISSUE #16172: Context menu in SQL Editor doesn't respect cursor position on which it is called

ShadelessFox:
Same as in https://github.com/dbeaver/dbeaver/pull/16575 but with support for block text selection.


Matvey16:
Verified

PR #16601, opened 2022-05-25
#13127 implement variable resolving for default connection name

### LINKED ISSUE #13127: change default connection name to hostname
When we create a new connection to the database server, the default connection name is database name
So, let's say in Oracle it is like SID it doesn't mean anything, doesn't do understanding on which server i connect. Or in postgresql i usually connect to postgres database.
Having many database server i connect on each server once, then I choose a specific database from top menu
it would be more convenient to set default connection name to hostname of database server

Destrolaric:
Closes #13127 
image


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ConnectionNameResolver.java
```diff
@@ -0,0 +1,96 @@
...
+            }
+            if (CommonUtils.isEmpty(newName)) {
+                newName = getConfiguration().getServerName();
+            }
+            if (CommonUtils.isEmpty(newName)) {
+                newName = getDataSourceContainer().getDriver().getName();
+            }
+            if (CommonUtils.isEmpty(newName)) {
+                newName = CoreMessages.dialog_connection_wizard_final_default_new_connection_name;
+            }
```
E1izabeth: Maybe it's better to assign variable only if the value is not empty instead of checking the same condition and assigning the value several times?

Destrolaric: Legacy behaviour

LonwoLonwo: And you can improve it.

Matvey16:
There should be a default naming pattern if this value is empty

Matvey16:
Verified

PR #16603, opened 2022-05-25
#15798 make deletion confirmation more noticing

### LINKED ISSUE #15798: Confirm 'Delete this script' in the the SQL Editor tab context menu
#### System information: 
- Operating system (distribution) and version: Archlinux newest
- DBeaver version: 21.3.4.202202142347
- Additional extensions: none

N/A

It is possible to permanently delete a SQL file (or any file) opened in the SQL editor from the tab context menu.

In the open SQL Editor tab click RMB to see context menu.

N/A

This is a bad UI design practice. Even with a confirmation modal window, it is still inappropriate place for such a profound action. It is too similar to other actions that require confirmation too. Permanent deletion of a file calls for at least familiar file management window, if not external file manager to handle the issue.

Destrolaric:
image
Closes #15798


E1izabeth:
Maybe it should be warning instead of error?

E1izabeth:
Maybe it should be warning instead of error?

Matvey16:
Verified

PR #16611, opened 2022-05-26
#9875 escape wild cards for the Snowflake metadata search properly

### LINKED ISSUE #9875: DBeaver Performance Issue Due to Meta Data calls
Hi Team

We are trying to setup DBeaver client SQL tool, we have setup the connection, but while testing we’ve run into performance issues around the client trying to read the system metadata where each table being read submits a show all columns statement for a given database and it takes quite some time to load those. Can you help us in setting this up, is there any property/setting within the tool which can help us avoid this.

Regards

Vivek

LonwoLonwo:
2022-05-26 15_34_19-Window

vs

2022-05-26 15_36_32-Window


uslss:
verified

PR #16613, opened 2022-05-26
#13456 Improve filtering in the `Quick Switch Editor` dialog

### LINKED ISSUE #13456: Quick Switch Editor search should find anywhere in tab name
The Quick Switch Editor search dialog (default Ctrl+E) has a largely useless search functionality, since the search is anchored at the beginning of the tab name, which includes the DB name.  For example, if the tab name is `<mydb> scratch.sql`, and I search for `scratch`, no search result pops up -- I have to search for a string anchored at the beginning `<mydb> scr`.

Text entered into the Quick Switch Editor search dialog should search anywhere in the tab name.  Ideally it would be a space-separated list of search strings that are AND-ed together.  For example, if I search for `gamma beta`, it should find `<mydb> alpha-beta-gamma.sql`.

ShadelessFox:
https://user-images.githubusercontent.com/35821147/170519569-3c20e711-1011-4c1d-866f-14017bfa624e.mp4

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ui/plugin.xml
@@ -73,4 +73,21 @@
                 description="DBeaver Perspective Context"/>
     </extension>
 
+    <extension point="org.eclipse.ui.handlers">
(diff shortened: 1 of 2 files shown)


Matvey16:
Verified

PR #16615, opened 2022-05-26
dbeaver/dbeaver-ee#1655 DROP TABLE to dangerous query execution confirmation

E1izabeth:
image


uslss:
ctrl+enter doesn't work, if value is set to Never (Alt+X works)

uslss:
verified

PR #16622, opened 2022-05-27
#16477 Fix npe in connection wizard

### LINKED ISSUE #16477: Connection Properties in inconsistent state after cancelling an operation from within the "edit connection" dialog

E1izabeth:


uslss:
verified

PR #16627, opened 2022-05-29
dbeaver/dbeaver-ee#696 Fix result set column mapping for document databases

E1izabeth:
Please, test `select column1, column2 from table;` for complex column types in smth like PostgreSQL or Oracle


uslss:
verified

Issue #16350, opened 2022-04-28
SQL Editor does not honor the "resultset fetch size" for Teradata


ahhmedbilal:
#### System information: 
- Operating system (distribution) and version: Windows 10
- DBeaver version: 22.0.3
- Additional extensions

#### Connection specification:
- Database name and version: Teradata 17.10
- Driver name: Teradata JDBC 17.10
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No

#### Describe the problem you're observing:

In Global setting for "Data Editor" I have following settings. 
"Resultset Fetch Size" is set to 6000. 
"auto fetch next segment" and "refresh data on next page read" and "use fetch size" is ticked. 

when i run following query is "SQL Editor"
select * from table; 
where table is having million of rows e.g 30 million rows. 
then resultset panel shows first 6000 rows, that is ok but the bottom right corner of the DBeaver window keep on showing the query progress, the "Background Tasks" also shows that query is running, where as it should have stopped after fetching 6000 rows. 

Now if i run any other small query e.g row count query from a small table, then it keeps on running as well and is added to "Background Tasks" but dont return any result unless the first query is ended. And the first query takes a long time to complete and first query keeps fetching the data and DBeaver memory size in Windows task manager grows to more than 1.2 GB and keeps growing. 

Above scenario is also true even after setting OFF following settings. 
"auto fetch next segment" and "refresh data on next page read" 

Above scenario is related to SQL Editor. 

Now below scenario is related to "View Data" on right click context menu of a table or view. 

When i right click a table or view and click "view Data", then the table/view window shows 6000 rows only and query stops after it i.e there is no background query running and the bottom right corner of DBeaver window does not show any progress, so every thing is OK in this case. 

Can you please fix the "Resultset Fetch Size" issue for SQL Editor, i.e query should end after fetching 6000 rows and there should not be any background task running after fetching 6000 rows. 

Thanks....

ahhmedbilal:
This driver i am using: Teradata JDBC Driver 17.10.00.27
from : https://downloads.teradata.com/download/connectivity/jdbc-driver

ahhmedbilal:
![image](https://user-images.githubusercontent.com/46369037/165725349-6875755d-dcbc-489a-9e07-d53976e844fe.png)
This data editor setting (global setting) i am using.

ahhmedbilal:
it was closed mistakenly.

Matvey16:
[...]

PR #16634, opened 2022-05-30
#16350 ability to use SQL limit fetch size for Teradata added

LonwoLonwo:
2022-05-30 16_47_36-DBeaver Ultimate 22 1 0 - _DBC_ Script-29


serge-rider:
Why not reuse existing one for SQL Server?

PR #16637, opened 2022-05-30
#16634 query transformer top limit moved from sql server to the model

### LINKED ISSUE #16350: SQL Editor does not honor the "resultset fetch size" for Teradata

LonwoLonwo:
now used for teradata too

2022-05-30 16_47_36-DBeaver Ultimate 22 1 0 - _DBC_ Script-29

So please test Teradata and SQL Server with this setting:

2022-05-30 17_56_51-DBeaver Ultimate 22 1 0 - _DBC_ Script-29


uslss:
verified

ahhmedbilal:
Can you plz tell in which build this will be fixed?

LonwoLonwo:
It will be fixed in DBeaver 22.1.0

PR #16639, opened 2022-05-30
#14926 Use 1-based numeration for collection elements

### LINKED ISSUE #14926: Array elements position differ in record mode and value panel
#### System information: 
- DBeaver version 21.3.1

The numbers indicating the position of array elements are different in record mode and value panel. In value panel they start from 0 and in record mode they start form 1.

1. Open a table with array in record mode
2. Compare numbers that indicate element position with value panel

MCGFiqveLn

ShadelessFox:
image


Matvey16:
Verified

PR #16701, opened 2022-06-03
dbeaver/dbeaver#16556 add a margin to the navigator's right if the vertical scrollbar is in overlay mode

### LINKED ISSUE #16556: Action icons can be unclickable on machines that use gtk

flambchap:
image
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/StatisticsNavigatorNodeRenderer.java
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/StatisticsNavigatorNodeRenderer.java
@@ -22,10 +22,7 @@ import org.eclipse.core.runtime.Status;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.custom.CTabFolder;
 import org.eclipse.swt.graphics.*;
-import org.eclipse.swt.widgets.Event;
-import org.eclipse.swt.widgets.ScrollBar;
-import org.eclipse.swt.widgets.Tree;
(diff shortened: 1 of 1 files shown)


serge-rider:
Lets use real scrollbar width instead of magic number 15

flambchap:
> Lets use real scrollbar width instead of magic number 15

Scrollbars on gtk3 are dynamic, meaning their size varies based on the pointer's position:

https://user-images.githubusercontent.com/26031301/172330715-3b3f41a2-ea06-4d41-a6bb-fcda17ce70c9.mp4

Moreover, the `Scrollbar` class does not report its maximum width. Sure, we can dynamically shift the icons. However, in the simplest form, it leads to the following behavior:

https://user-images.githubusercontent.com/26031301/172333222-c1cc8699-1e97-43c2-a39a-bc38af478014.mp4

I find it terrible.

We can try finding a solution that moves all the icons at once at precisely the time scrollbar changes its width, but:
* It's complicated and impactful,
* Will lead to performance degradation, potentially an unacceptable one,
* And it might be impossible.

All that makes it not even worth trying,

I agree that this magic number solution doesn't look great, but I think it's better than the alternatives.

PR #16714, opened 2022-06-06
#14360 add cell move after enter & refactor grid preferences

### LINKED ISSUE #14360: behaviour of enter-key when inline editing a table cell
When editing a table cell inline
afbeelding
pressing the enter-key ends editing, and highlights the current cell
afbeelding
Personally I rarely want to stay in the current cell after a cell-edit. Most of the time I want to go down 1 cell or go 1 cell to the right. Please consider pressing "enter" with a cell in edit mode moves a cell down after closing edit (e.g. like in Excel) or even better: pressing enter on the typewriter-pad and on the numeric pad show different behaviour (one moves down and other moves right) e.g. like in the old pgadmin III. Perhaps adding a few "commands" to the user interfase keyboard preferences like "close edit and down" and "close edit and right".

Destrolaric:
Now user can tick "Select next cell after inline edit", which automatically switches the cell to one on the next row or, in case of shift+enter, the next column.
Closes #14360
image


serge-rider:
LEt's try to do it without `additionalActions` and handle everything in traverse listener

uslss:
verified

PR #16723, opened 2022-06-07
#16340 allow null for calendar

### LINKED ISSUE #16340: Can't use calendar on fields with null value
#### System information: 
- DBeaver version 22.0.3

1. Open open a table with a date/time column
2. Click on the field with null value
3. Make sure that value panel has text presentation
4. Change value panel to calendar

**Expected:** Calendar shows default value or at least opens with blank fields
**Actual:** Error appears

dbeaver_5xFe4OEhPj

Destrolaric:
Closes #16340


serge-rider:
Let's use zero time (1970?) instead of current time for nulls

Matvey16:
Broke value panel for text presentation of date/time

Matvey16:
Still the same

Matvey16:
Hint about current time doesn't appear when switching from text value

dbeaver_JKQFpyOHiD

Matvey16:
Verified

Issue #10247, opened 2020-11-05
Netezza JDBC error


tikala:
#### System information: 
- Operating system (distribution) and version: Windows 10
- DBeaver version: 7.1.1, 7.1.5, 7.2.3 and 7.2.4 ce (all this tried)
- Additional extensions

#### Connection specification:
- Database name and version: Netezza 7.2.1
- Driver name: JDBC from IBM
- Do you use tunnels or proxies (SSH, SOCKS, etc)?: no

#### Describe the problem you're observing:
When testing connection, getting this error:

Unexpected driver error occurred while connecting to database
java.lang.NullPointerException

Driver and connection works fine from SQuirreL.

Same probleam as this: https://github.com/dbeaver/dbeaver/issues/5392

#### Include any warning/errors/backtraces from the logs

!ENTRY org.jkiss.dbeaver.model 4 0 2020-11-05 11:08:24.188
!MESSAGE 
!SUBENTRY 1 org.jkiss.dbeaver.model 4 0 2020-11-05 11:08:24.188
!MESSAGE 
!STACK 0
java.lang.NullPointerException
	at org.netezza.sql.NzConnection$Initializer.handshake(NzConnection.java:2163)
	at org.netezza.sql.NzConnection$Initializer.init(NzConnection.java:2091)
	at org.netezza.sql.NzConnection.open(NzConnection.java:298)
	at org.netezza.datasource.NzDatasource.getConnection(NzDatasource.java:719)
	at org.netezza.datasource.NzDatasource.getConnection(NzDatasource.java:706)
	at org.netezza.Driver.connect(Driver.java:155)
	at org.jkiss.dbeaver.model.impl.jdbc.JDBCDataSource.lambda$0(JDBCDataSource.java:173)
	at org.jkiss.dbeaver.model.impl.jdbc.JDBCDataSource.openConnection(JDBCDataSource.java:192)
	at org.jkiss.dbeaver.ext.generic.model.GenericDataSource.openConnection(GenericDataSource.java:124)
	at org.jkiss.dbeaver.model.impl.jdbc.JDBCExecutionContext.connect(JDBCExecutionContext.java:97)
	at org.jkiss.dbeaver.model.impl.jdbc.JDBCRemoteInstance.initializeMainContext(JDBCRemoteInstance.java:91)
	at org.jkiss.dbeaver.model.impl.jdbc.JDBCRemoteInstance.<init>(JDBCRemoteInstance.java:57)
[...]

Issue #5392, opened 2019-03-04
Netezza JDBC error


lavalex:
#### System information: 
- Operating system (distribution) and version Windows 10
- DBeaver version 5.3.5 Community Edition
- Java version Whatever comes with DBeaver
- Additional extensions

#### Connection specification:
- Database name and version Netezza 7.2.1
- Driver name JDBC from IBM
- Do you use tunnels or proxies (SSH, SOCKS, etc)? no

#### Steps to reproduce, if exist:
When testing connection, getting this error:

Unexpected driver error occurred while connecting to database
  
  java.lang.NullPointerException

Driver works fine from a Java program, but not from DBeaver.

serge-rider:
Sorry for late reply.
Could you attach error/debug logs with this error? (https://github.com/dbeaver/dbeaver/wiki/Log-files)

uslss:
there is no update more than a month. Ticket closed.  
If issue is still actual feel free to reopen the ticket or create a new one.

PR #16736, opened 2022-06-08
Workaround for NullPointerException with older Netezza Drivers

rosspb3:
Older Netezza Drivers look for system property "sun.java.command" and fail when value is null.  This code sets the value to an empty string, if the value is null prior to initializing the driver.  See issues [5932](https://github.com/dbeaver/dbeaver/issues/5392) and [10247](https://github.com/dbeaver/dbeaver/issues/10247).


LonwoLonwo:
Hello @rosspb3 

Maybe you have any link for the old Netezza driver? To test your PR.

LonwoLonwo (approved):

Matvey16:
Verified

PR #16746, opened 2022-06-08
#13127 fix name if not changed

### LINKED ISSUE #13127: change default connection name to hostname

Destrolaric:
Closes #13127

### DIFF
--- a/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageGeneral.java
+++ b/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageGeneral.java
@@ -157,8 +157,7 @@ public class ConnectionPageGeneral extends ConnectionWizardPage implements Navig
             ConnectionPageSettings settings = wizard.getPageSettings();
 
             if (dataSourceDescriptor != null && !CommonUtils.isEmpty(dataSourceDescriptor.getName())) {
-                connectionNameText.setText(generateConnectionName(settings, ModelPreferences.getPreferences().getString(ModelPreferences.DEFAULT_CONNECTION_NAME_PATTERN)));
-                connectionNameChanged = true;
+                connectionNameText.setText(dataSourceDescriptor.getName());
             } else {
                 if (CommonUtils.isEmpty(connectionNameText.getText()) || !connectionNameChanged) {
                     String newName = generateConnectionName(settings, ModelPreferences.getPreferences().getString(ModelPreferences.DEFAULT_CONNECTION_NAME_PATTERN));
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageGeneral.java
```diff
@@ -291,8 +290,10 @@ public void createControl(Composite parent)
             String connectionName = dataSourceDescriptor == null ? "" : dataSourceDescriptor.getName(); //$NON-NLS-1$
             connectionNameText = UIUtils.createLabelText(miscGroup, CoreMessages.dialog_connection_wizard_final_label_connection_name, CommonUtils.toString(connectionName));
             connectionNameText.addModifyListener(e -> {
-                connectionNameChanged = true;
-                ConnectionPageGeneral.this.getContainer().updateButtons();
+                if (dataSourceDescriptor == null || !connectionNameText.getText().equals(connectionName)) {
+                    connectionNameChanged = true;
+                    ConnectionPageGeneral.this.getContainer().updateButtons();
```
ShadelessFox:
```suggestion
                    getContainer().updateButtons();
```

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageGeneral.java
```diff
@@ -562,8 +563,16 @@ public void saveSettings(DBPDataSourceContainer dataSource) {
...
+        String name;
+        if (dataSourceDescriptor == null) {
+            name = connectionNameChanged
+                    ? generateConnectionName(getWizard().getPageSettings(), connectionNameText.getText()) :
+                    generateConnectionName(getWizard().getPageSettings(), ModelPreferences.getPreferences()
+                            .getString(ModelPreferences.DEFAULT_CONNECTION_NAME_PATTERN));
+        } else {
+            name = connectionNameChanged ? generateConnectionName(getWizard().getPageSettings(),
+                    connectionNameText.getText()) : dataSourceDescriptor.getName();
+        }
```
ShadelessFox: 1. Please don't mix `if` statements with ternary operator expressions.
2. Continuation indentation must be four spaces, not 8.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageGeneral.java
```diff
@@ -562,8 +563,17 @@ public void saveSettings(DBPDataSourceContainer dataSource) {
...
+        String defaultValue;
+        if (dataSourceDescriptor == null) {
+            defaultValue = ModelPreferences.getPreferences()
+            .getString(ModelPreferences.DEFAULT_CONNECTION_NAME_PATTERN);
+        } else {
+            defaultValue = dataSourceDescriptor.getName();
+        }
+        String name = connectionNameChanged ? generateConnectionName(getWizard().getPageSettings(),
+            connectionNameText.getText()) :
+            generateConnectionName(getWizard().getPageSettings(), defaultValue);
```
ShadelessFox: 🤦🏻‍♂️

```suggestion
    final String name;

    if (connectionNameChanged) {
        name = connectionNameText.getText();
    } else if (dataSourceDescriptor != null) {
        name = dataSourceDescriptor.getName();
    } else {
        name = ModelPreferences.getPreferences().getString(ModelPreferences.DEFAULT_CONNECTION_NAME_PATTERN);
    }

    dataSource.setName(generateConnectionName(getWizard().getPageSettings(), name));
```

ShadelessFox (changes requested):

ShadelessFox (approved):

DariaMarkaryan:
1. project.path doesn't work
2. where does the {server} value comes from
3. we can't use {day}, {month}, {year} variables, but they should be available

Matvey16:
Verified

PR #16754, opened 2022-06-09
#15698 Improve support for structs

### LINKED ISSUE #15698: Empty strings in PostgreSQL arrays are recognized as NULL
#### System information: 
- DBeaver version 22.0.0 devel branch
- Additional extensions

- Database name and version: PostgreSQL 11,6
- Driver name

1. Execute the following:
```sql
select
  array[''],
  array[]::text[],
  row('', 'NULL'),
  row(null, 'NULL'), 
  array[null, 'NULL'], 
  row(array[null]),
  row(array[null, 'NULL']), 
  array[row(null), row('NULL')];
```
The results are displayed like this:
image

But this is not correct:
image

Credit to @lukaseder for discovering this bug.

ShadelessFox:
Now structs are also compatible with the output of `psql`.

Several test cases:

```sql
create type pair as (a int[], b int[]);
select array[row(array[1,2,3], array[4,5,6])::pair, row(array[1,2,3], NULL)::pair]::pair[];
```

```sql
create type a as (a text);
select row(null)::a, row('')::a, row('"')::a, row('()')::a;
```

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreValueParser.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreValueParser.java
@@ -49,9 +49,7 @@ public class PostgreValueParser {
 
     public static Object convertStringToValue(DBCSession session, DBSTypedObject arrayType, String string) throws DBCException {
         if (arrayType.getDataKind() == DBPDataKind.ARRAY) {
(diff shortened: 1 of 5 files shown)


Matvey16:
Pretty edge case, but this for example will give incorrect result
```sql
create type aa as (a text, b text);

select array[row(null, null)::aa, null];
```
dbeaver_63engc4laX
dbeaver_bzlkNRwzZy
cmd_Wf7a8RPqnJ

@serge-rider Do you think if this should be fixed, or works as expected?

Matvey16:
Verified
Opened a different ticket for the edge case #16822

PR #16761, opened 2022-06-09
#16733 Fix postgresql enum attribute binding

### LINKED ISSUE #16733: Cannot Edit Postgres Enum Columns: Column "ColumnName" is read-only: No corresponding table column
#### System information: 
- DBeaver version 22.1.0 CE ... regression from 22.0.5 CE

When you try to edit any enum cell from either a View Data or Read Data in SQL Editor, DBeaver prevents the edit with a Column "ColumnName" is read-only: No corresponding table column message

and there is not autocomplete list shown

btw. DBeaver 22.0.5 works fine... and even better than I'm used to because, for a long while, I wasn't able to edit with Read Data in SQL Editor only with View Data

E1izabeth:


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -520,6 +520,11 @@ public Collection<PostgreDataType> getLocalDataTypes()
     {
         return getDefaultInstance().getLocalDataTypes();
     }
+    
+    @Override
+    public DBSDataType resolveDataType(DBRProgressMonitor monitor, String typeFullName) throws DBException {
```
ShadelessFox: Please add missing `@NotNull`/`@Nullable` annotations from the base method.

Also, what's the point of overriding this method? It's not used anywhere among other changes.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
```diff
@@ -853,6 +857,18 @@ public static void bindAttributes(
         }
     }
 
+    private static boolean isEnumType(DBSEntityAttribute tableColumn) {
+        if (tableColumn instanceof DBSTypedObjectEx) {
+            DBSDataType columnDataType = ((DBSTypedObjectEx) tableColumn).getDataType();
+            Object typeExt = columnDataType.geTypeExtension();
+            if (typeExt instanceof DBPNamedObject) {
+                Predicate<String> hasEnumSubstring = Pattern.compile("\\benum\\b", Pattern.CASE_INSENSITIVE).asPredicate();
```
ShadelessFox: Kinda strange way of checking for a type, but whatever.

Please consider making this pattern a `static final` member of this class to avoid compiling it every time the method is called.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
```diff
@@ -853,6 +857,18 @@ public static void bindAttributes(
...
+                return hasEnumSubstring.test(((DBPNamedObject)typeExt).getName());
```
ShadelessFox:
```suggestion
                return hasEnumSubstring.test(((DBPNamedObject) typeExt).getName());
```

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
E1izabeth: I know. Unfortunately, there is no other way to check it.

ShadelessFox (approved):

serge-rider:
Please move `enum` type check into PG extension

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
```diff
@@ -862,10 +859,7 @@ public static void bindAttributes(
     private static boolean isEnumType(DBSEntityAttribute tableColumn) {
         if (tableColumn instanceof DBSTypedObjectEx) {
             DBSDataType columnDataType = ((DBSTypedObjectEx) tableColumn).getDataType();
-            Object typeExt = columnDataType.geTypeExtension();
-            if (typeExt instanceof DBPNamedObject) {
-                return HAS_ENUM_SUBSTRING.test(((DBPNamedObject) typeExt).getName());
-            }
+            return columnDataType == null ? false : columnDataType.isEnum();
```
ShadelessFox:
```suggestion
            return columnDataType != null && columnDataType.isEnum();
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
```diff
@@ -84,8 +84,6 @@ public class DBExecUtils {
     private static final List<DBPDataSourceContainer> ACTIVE_CONTEXTS = new ArrayList<>();
     public static final boolean BROWSE_LAZY_ASSOCIATIONS = false;
 
-    private static final Predicate<String> HAS_ENUM_SUBSTRING = Pattern.compile("\\benum\\b", Pattern.CASE_INSENSITIVE).asPredicate();
-
```
ShadelessFox: Please remove redundant imports:

```java
import java.util.function.Predicate;
import java.util.regex.Pattern;
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -523,9 +523,8 @@ public Collection<PostgreDataType> getLocalDataTypes()
     
     @Override
     public DBSDataType resolveDataType(@NotNull DBRProgressMonitor monitor, @NotNull String typeFullName) throws DBException {
-        // TODO: implement full type name resolution
-        // now typeFullName is assumed as local which may leads to some bugs
-        return super.resolveDataType(monitor, typeFullName);
+        // TODO: implement full type name resolution due to typeFullName being treated as local type name, which is inconsistent
+	return super.resolveDataType(monitor, typeFullName);
```
ShadelessFox: No tabs pls

```suggestion
        return super.resolveDataType(monitor, typeFullName);
```

ShadelessFox (changes requested):

ShadelessFox (approved):

uslss:
verified

PR #16784, opened 2022-06-14
#16443 Add DDL from result set generator

### LINKED ISSUE #16443: Generate DDL from a result set
Hi, 

My suggestion is add an option to generate da simple DDL (create table) based on the data types of the result set. 
This can be very useful when we need to work with data movement and data manipulation where we need to include some result into a new table.
Where the DDL generated already match with the result set in use.

E1izabeth:
image

Doesn't work for PostgreSQL due to bug https://github.com/dbeaver/pro/issues/1053


uslss:
incorrect context menu
image

uslss:
verified

PR #16786, opened 2022-06-14
dbeaver/dbeaver-ee#1707 Add data type by full type name resolution in PostgreSQL

E1izabeth:
Please, test different use cases of PostgreSQL usage


uslss:
verified

PR #16787, opened 2022-06-14
CB-2123 qm server plugin

yagudin10:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/qm/meta/QMMConnectionInfo.java
```diff
@@ -109,6 +138,62 @@ public QMMConnectionInfo getConnection() {
         return this;
     }
 
+    @Override
+    public Map<String, Object> toMap() throws DBException {
+        Map<String, Object> serializedConnectionInfo = new LinkedHashMap<>();
+        serializedConnectionInfo.put("containerId", getContainerId());
```
LonwoLonwo, line 158: You can use constants for "containerId" and other strings because they have duplicates below in your code.

PR #16797, opened 2022-06-15
#15584 double click listener added to Full-Text and Metadata search pages

### LINKED ISSUE #15584: DB Full-Text Search limited at 5000 tables, unable to search/load more in the tree
#### System information: 
- Windows 10 Pro 64bit
- DBeaver version 21.3.4.202202051448

- PostgreSQL 11.9
- PostgreSQL JDBC Driver [42.2.5]

When attempting to conduct a full-text search in a DB that has ~10000 tables, the tree in the databases field truncates at 5000 tables. When clicking the checkbox for the Tables node, it will not return with a full check-mark, but rather a square indicating that only partial of all the child nodes are selected. 

image

This is despite that when scrolling down the tree, all nodes are selected, including the overflow/roll-up "More" node representing the remaining 5000 tables.

image

Clicking/Double-clicking on the "More" node will not load the remaining tables into the tree (in contrast to the DB Navigator, which will do that).  Nothing occurs, the tree remains static.

Completing a Full-Text Search in this state will result in only the 5000 visible nodes being searched, the remaining 5000 tables are not searched.

Deselecting the initial 5000 nodes, and selecting only the "More" node results in the Search button remaining inactive, so searching only those tables is not possible.

image

Inspected all logs, no relevant entries present in them.  (None of the timestamps correlate to when attempting this)

LonwoLonwo:
013i8UxcWS


Matvey16:
Verified

PR #16805, opened 2022-06-15
fix(org.jkiss.dbeaver.data.office): CB-1543 remove background from th…

devnaumov:
…e excel icons


ShadelessFox:
Why are they so thin and blurry now?

PR #16810, opened 2022-06-16
#14544 ability to read columns with constraints implemented 

### LINKED ISSUE #14544: [PostgreSQL] Read constraints with the columns.
As continue of this ticket: https://github.com/dbeaver/dbeaver/issues/13410

We need to find a solution to the expensive constraints reading on the step table columns reading. 
It will help to place the unique key icon properly.

Maybe we can add a special setting in the PG connection settings for that.

LonwoLonwo:
with the new connection setting

2022-06-16 11_58_28-DBeaver Ultimate 22 1 0 - Employee_Audit


serge-rider:
Try to use isFullyCached

serge-rider:
Let's read constraints along with columns but not read them on demand.

Matvey16:
Verified

PR #16812, opened 2022-06-16
dbeaver/dbeaver-ee#1134 Execute script from the position, Re-execute query from Execution Log 

E1izabeth:
https://user-images.githubusercontent.com/28875055/174158783-671f1219-2b1c-495a-8276-f3e58dd233a6.mp4

Execute script from the position:

image

Re-execute query:
image

We don't have enough information to get the corresponding to query editor. So:
- Re-execute query works only in Execution Log, because we have `SQLEditor` as `filter` there.
- Re-execute query doesn't work when option `Use separate connection for each editor` is not set, because we can't split queries from different editors in this case. All queries for different editors for this db connection mess up in Execution Log and it becomes possible to execute query from script corresponding to another editor in active editor.

To have opportunity to reexecute query in the editor context from Query Manager or jump to query we need to add information about the corresponding to query editor `IWorkbenchPartReference` and query position in script.

Despite that, it's possible to jump to the failed query from Results tab
image
or from Problems view.


Matvey16:
Veried

PR #16813, opened 2022-06-16
#13907 Add paging for dictionary panel

### LINKED ISSUE #13907: Redesign dictionary panel UI
When the dictionary tab is opened, not all rows of the referenced table are loaded. There is no ability to load the next page of dictionary results. 

When the referenced table is of a certain size (say 1000 rows), not all rows will be loaded in the dictionary tab, and the missing rows are not available for selection.

Enable pagination in dictionary tab, so that a user is able to browse for the desired row.

Search bar that lookup certain columns in the referenced table. However, pagination may still be required as there may be a large number of results.

Destrolaric:
image
Closes #13907


serge-rider:
Please fix codestyle of function definition.
Instead of load direction use simple paging (offset + size)

Matvey16:
Seems like this broke the dictionary

Before the PR:
dbeaver_K875K1Opi8

After the PR:
dbeaver_ZMkABZivag


@Destrolaric I think your screenshot was taken without this PR

Matvey16:
Verified

PR #16816, opened 2022-06-16
#16079 add varchar conversion for Oracle

### LINKED ISSUE #16079: Mapping "VARCHAR" during export from MySQL to Oracle
Oracle VARCHAR datatype is a synonym to VARCHAR2 but should not be used --> [https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-DF7E10FC-A461-4325-A295-3FD4D150809E](url)

So during export/import from MySQL to Oracle, VARCHAR columns should be mapped to VARCHAR2.

Destrolaric:
Closes #16079

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleSQLDialect.java
+++ b/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleSQLDialect.java
@@ -535,6 +535,11 @@ public class OracleSQLDialect extends JDBCSQLDialect implements SQLDataTypeConve
         String localDataType = null, dataTypeModifies = null;
 
         switch (externalTypeName) {
+            case "VARCHAR":
+                //We don't want to use a VARCHAR it's not recommended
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleSQLDialect.java
```diff
@@ -535,6 +535,11 @@ public String convertExternalDataType(@NotNull SQLDialect sourceDialect, @NotNul
         String localDataType = null, dataTypeModifies = null;
 
         switch (externalTypeName) {
+            case "VARCHAR":
+                //We don't want to use a VARCHAR it's not recommended
+                //See https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-DF7E10FC-A461-4325-A295-3FD4D150809E
```
ShadelessFox, line 540:
```suggestion
                // We don't want to use the VARCHAR type as it's not recommended
                // See https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-DF7E10FC-A461-4325-A295-3FD4D150809E
```

ShadelessFox (changes requested):

uslss:
verified

PR #16819, opened 2022-06-17
#16295 ability to use or not to use column names as header 

### LINKED ISSUE #16295: Rows misplaced when exporting to .txt file
#### System information: 
- DBeaver version: 22.0.4

The export wizard is not performing correctly when exporting to TXT files with "write to the single file" enabled.
For the latter files exported(or appended) to the single file, their column names are replaced with the first row.
image
Exporting to seperate files works perfectly fine:
image

Export two or more tables or views to TXT format, with "write to the single file" option enabled and other settings as default.

LonwoLonwo:
in TXT export added

2022-06-16 22_08_54-Playlist_202206162208 txt — H__asya_special_dbeaver_backup — Atom

2022-06-16 21_47_09-Data Transfer


uslss:
verified

PR #16828, opened 2022-06-18
#14537 fix selected columns export to another database

### LINKED ISSUE #14537: Export in table "Only selected columns" has wrong mapping
#### System information: 
- Win10
- DBeaver CE 21.2.5

- Not specific

Because of all columns are created if you check "only selected columns" 

1. Create table with two columns: the first is varchar and the second is Date type.
2. insert data in row
image
3. select only date value cell
image
4. activate "only selected columns" 
image
5. Start export. Then you can see that value from the first column try to insert in the exporting column (so we get an error that says 'abc' is wrong format for date column )
image

LonwoLonwo:
Please test it attentively

- Tables with hidden columns
- Tables with changed order columns
- Document-Based data bases
- Export from SQL editor - different type of queries
- All with and without "selected columns only" setting


uslss:
for now, all columns are created, but not selected column got null values
image

uslss:
verified

PR #16831, opened 2022-06-18
#16695 Grid array model and UI

### LINKED ISSUE #16695: Spreadsheet: Refactor model to support nested rows
Currently, our presentation model is unaware of collection elements and other stuff. That's why we have complex value editors.

There's not a lot of information regarding this issue because it's about development.

This ticket is a part of #12148.

serge-rider:


uslss:
1. Composite Types cannot be expanded
image

2. Empty value panel is shown for the nested elements
image

3. Delete button is available for the nested element, but it proposes to drop the whole row
image

4. In case the focus is on the nested object on Create button click - the row with a random number of nested rows is created. Also, the focus jumps to the random object above 
dbeaver_oVYWiIzRrA

serge-rider:
1. Fixed
2. Fixed
3. Works as expeted. Currently "delete" action deletes entire rows (as before)
4. It is not a "random" but next real row (not array element). "Create row" creates a row, not array element. New row positioning was fixed.

uslss:
empty nested elements rows in case of expand when Show complex structure is enabled 
dbeaver_Pewhoe7iiC

uslss:
Second element isn't shown. First element is shown twice
image
image

uslss:
verified

PR #16832, opened 2022-06-18
#16669 Do original column name resolution for condition clauses

### LINKED ISSUE #16669: SQL Server: ResultSet is read-only when column aliases are used
#### System information: 
- Operating system (distribution) and version Win
- DBeaver version 22.1.0.202205310836

- Database name and version SQL Server

Write a select statement with column aliases in the SQL Editor and execute it

**Expected**: the table can be edited in the ResultSet
**Actual result**: the table is read-only

dbeaver_1Xon8HvI3O

serge-rider:


uslss:
still reproducible, the alias is used instead of the column name.

also resultset becomes readonly for SQLite
image

serge-rider:
Fixed

uslss:
verified

PR #16833, opened 2022-06-19
#15921 dt improvements

### LINKED ISSUE #15921: Import wizard ignores existing data types while DDL is generated
#### System information: 
- DBeaver version 22.0.1

Import wizard automaps table to be created even if another relation (e.g. data type) with that name already exists

1. Create a CSV file with a name of one of custom data types in your schema
2. In the context menu of your schema start data import
3. Leave all settings by default

**Expected:** Data is imported
**Actual:** Table can't be created

dbeaver_s9r8zPIMD8

dbeaver_nHZqJXY6cw

serge-rider:


Matvey16:
The fix doesn't work if the custom type is `enum`

serge-rider:
Enums supprot was added

Matvey16:
Verified

PR #16881, opened 2022-06-24
#667 Console View for the SQL Editor

### LINKED ISSUE #667: Feature Request: Display Results From Scripted Queries in a "Console" Window
This is a follow-up to issue 644, which has been closed.  It would be nice to have the entire result set from a script populate a single "console" window.  The results could then be copied and pasted into a report or other application.

For example, I have created a script that contains 30 queries.  When I run the script, 30 tabs get opened.  I then have to separately select each of these tabs,  export the result sets to 30 separate HTML files, and then import the HTML files into my report document.  

It would be nice to have an option whereby the results set could either populate a text-based window, or create one HTML document containing a set of tables, where the results could then be copied and pasted into another application in one action.

Submitted as per Serge's suggestion.  Comments welcome.

Kevin

E1izabeth:
debug launch update for ee: https://github.com/dbeaver/dbeaver-ee/pull/1760

Console View

image

Default value setting. Works until user enable or disable console view manually. After that the last state is saved and can be restored.

image

Also Execute script in a new tabs were removed from sql editor left toolbar.

Possible further improvements:
- errors output
- include queries which doesn't produce any result set ?
- cursor ?
- selection ?


uslss:
verified

PR #16898, opened 2022-06-27
#15805 Clarify the meaning of the "Custom criteria"

### LINKED ISSUE #15805: Question: Define description columns using Custom Criteria
In the Data tab of a MySQL table I opened the Value viewer panel. In this panel it shows me a dictionary if I select a foreign key value in the Data tab. By clicking Define Description I open the Dictionary structure dialog.

I see there is a Custom criteria text box available. What can I type in there? for example if I type in `select 'foo'` then the output in the Value viewer becomes completely blank. I can type in two or more columns from the existing table like `col1, col2` and that seems to work - but that's just the same as selecting those columns using the checkboxes provided above.

So the question is : what can I use the custom criteria for? if there's already a section in the user guide - I'd be happy to get pointed to that as well

image

image

#### System information: 
- DBeaver version
- Java version 
- Additional extensions

- Database name and version
- Driver name

ShadelessFox:


uslss:
verified

PR #16918, opened 2022-06-29
#16030 add case override for generic drivers

### LINKED ISSUE #16030: [Kylin] Objects with a mixed case name are quoted but mustn't be
#### System information: 
- Win11
- DBeaver CE 22.0.2

- Apache Kylin

The SQL Assistant offers to insert a database name with quotas, and this  query falls:
image
image
But without quotes the query is executed without problems
image

Destrolaric:
added 2 new parameters for generics they allow to explicitly designate driver case for quoted/unquoted strings  
Closes #16030

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
@@ -123,6 +123,8 @@
                     <property id="supports-indexes" label="Driver supports indexes" description="Driver supports table indexes" type="boolean" required="false" defaultValue="true"/>
                     <property id="supports-table-constraints" label="Driver supports constraints" description="Driver supports table constraints (primary keys)" type="boolean" required="false" defaultValue="true"/>
                     <property id="supports-stored-code" label="Driver supports stored code" description="Whether this driver supports stored code (procedures, functions, packages, etc)" type="boolean" required="false" defaultValue="true"/>
+                    <property id="stores_quoted_case" label="Quoted string case" description="Which letter capitalisation is supported for a quoted strings(mixed/upper/lower)" type="string" required="false" defaultValue=""/>
+                    <property id="stores_unquoted_case" label="Unquoted string case" description="Which letter capitalisation is supported for a unquoted strings(mixed/upper/lower)" type="string" required="false" defaultValue=""/>
                     <property id="supports-references" label="Driver supports references" description="Driver supports table references (foreign keys)" type="boolean" required="false" defaultValue="true"/>
                     <property id="supports-select-count" label="Driver supports SELECT count(*) clause" description="Driver supports SELECT count(*) clause" type="boolean" required="false" defaultValue="true"/>
                     <property id="supports-views" label="Driver supports views" description="Driver supports table views" type="boolean" required="false" defaultValue="true"/>
@@ -1220,7 +1222,7 @@
                     categories="sql">
                     <file type="jar" path="maven:/org.apache.kylin:kylin-jdbc:RELEASE"/>
                     <file type="jar" path="maven:/javax.xml.bind:jaxb-api:RELEASE"/>
-
+                    <parameter name="stores_unquoted_case" value="mixed"/>
                     <parameter name="omit-catalog-name" value="true"/>
                     <parameter name="supports-indexes" value="false"/>
                     <parameter name="supports-references" value="false"/>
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericSQLDialect.java
```diff
@@ -96,6 +108,24 @@ public void initDriverSettings(JDBCSession session, JDBCDataSource dataSource, J
...
+            switch (DBPIdentifierCase.valueOf(driverCase.toUpperCase())) {
+                case MIXED:
+                    return DBPIdentifierCase.MIXED;
+                case UPPER:
+                    return DBPIdentifierCase.UPPER;
+                case LOWER:
+                    return DBPIdentifierCase.LOWER;
+                default:
+                    return null;
+            }
```
ShadelessFox: wtf

plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericSQLDialect.java
```diff
@@ -96,6 +108,24 @@ public void initDriverSettings(JDBCSession session, JDBCDataSource dataSource, J
         this.supportsMultiInsert = CommonUtils.toBoolean(driver.getDriverParameter(GenericConstants.PARAM_SUPPORTS_MULTI_INSERT));
     }
 
+    @Nullable
+    private DBPIdentifierCase determineCase(@Nullable String driverCase) {
```
ShadelessFox: Please use `CommonUtils.valueOf` instead.

plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/GenericConstants.java
```diff
@@ -45,6 +45,8 @@ public class GenericConstants {
     public static final String PARAM_OMIT_SINGLE_CATALOG = "omit-single-catalog";
     public static final String PARAM_OMIT_SINGLE_SCHEMA = "omit-single-schema";
     public static final String PARAM_OMIT_CATALOG_NAME = "omit-catalog-name";
+    public static final String PARAM_STORES_UNQUOTED_CASE = "stores_unquoted_case";
+    public static final String PARAM_STORES_QUOTED_CASE = "stores_quoted_case";
```
ShadelessFox: Use `-` instead of `_` to be consistent with other parameters.

```suggestion
    public static final String PARAM_STORES_UNQUOTED_CASE = "stores-unquoted-case";
    public static final String PARAM_STORES_QUOTED_CASE = "stores-quoted-case";

```

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -123,6 +123,8 @@
                     <property id="supports-indexes" label="Driver supports indexes" description="Driver supports table indexes" type="boolean" required="false" defaultValue="true"/>
                     <property id="supports-table-constraints" label="Driver supports constraints" description="Driver supports table constraints (primary keys)" type="boolean" required="false" defaultValue="true"/>
                     <property id="supports-stored-code" label="Driver supports stored code" description="Whether this driver supports stored code (procedures, functions, packages, etc)" type="boolean" required="false" defaultValue="true"/>
+                    <property id="stores-quoted-case" label="Quoted string case" description="Which letter capitalisation is supported for a quoted strings(mixed/upper/lower)" type="string" required="false" defaultValue=""/>
+                    <property id="stores-unquoted-case" label="Unquoted string case" description="Which letter capitalisation is supported for a unquoted strings(mixed/upper/lower)" type="string" required="false" defaultValue=""/>
```
flambchap:
```suggestion
                    <property id="stores-quoted-case" label="Quoted string case" description="Letter capitalization of quoted strings (mixed/upper/lower)" type="string" required="false" defaultValue=""/>
                    <property id="stores-unquoted-case" label="Unquoted string case" description="Letter capitalization of unquoted strings (mixed/upper/lower)" type="string" required="false" defaultValue=""/>
```

Reasoning: 

1. We use NA spelling,
2. We cannot use `a` here,
3. Stylistically, it's better to avoid passive voice here.
4. The proposed version is a bit less verbose

plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/GenericConstants.java
```diff
@@ -45,6 +45,8 @@ public class GenericConstants {
     public static final String PARAM_OMIT_SINGLE_CATALOG = "omit-single-catalog";
     public static final String PARAM_OMIT_SINGLE_SCHEMA = "omit-single-schema";
     public static final String PARAM_OMIT_CATALOG_NAME = "omit-catalog-name";
+    public static final String PARAM_STORES_UNQUOTED_CASE = "stores-unquoted-case";
+    public static final String PARAM_STORES_QUOTED_CASE = "stores-quoted-case";
```
ShadelessFox: They should be called `stored-xxx` instead of `stores-xxx`, actually. Despite the same naming in the `SQLDialect` class, it's clearly a typo.

ShadelessFox (changes requested):

ShadelessFox (approved):

uslss:
verified

PR #16927, opened 2022-06-30
#16427 Set default save path to `Scripts` when saving an SQL file

### LINKED ISSUE #16427: Save SQL console as script
I changed default 'SQL' command to 'SQL console' instead of 'SQL script', to avoid creating a lot of useless scripts.
If I now have a statement worth saving, I have to create a script and copy from console to script. 
Would be good if the 'SQL console' window header would have a 'Save as Script' to save in 'Project/Scripts' folder. 
Save as real file is already possible.

ShadelessFox:


Matvey16:
Verified

PR #16934, opened 2022-06-30
#15555 check for a JavaIdentifier instead of only a isLetterOrDigit

### LINKED ISSUE #15555: Sybase jConnect connections: Queries specifying columns that start with go_ are treating the 'go' as a script delimiter.
Version 21.3.4.202202051448
OS: Windows 10 Enterprise, Version: 1901
Database: Sybase jConnect, Driver Type: SQL Server (Generic)
Tunnels/Proxies: None

When creating a query on a table that has a column that starts with go_ the go is interpreted as a script delimiter string and the query returns an SQL Error after stripping away the 'go'.
This occurs even when changing the 'script delimiter' setting in the Advanced parameters of the driver settings from ';,go' to ';'.

To reproduce, create a table with a column name 'go_foo',
Connect to the database, start an SQL script with a simple select:

select * from _Tablename_ where go_foo = 'bar'

Run the sql script.

Using double quotes on the column name "go_foo" seems to work but is less desirable and also doesn't explain/fix why the removal of 'go' as a script delimiter doesn't work.

Note: Naming a column with characters other than '_' after 'go' seems fine. So naming the column 'good_foo' does not generate the error when running a script that uses the good_foo column name.

Destrolaric:
Closes #15555


uslss:
verified

PR #16940, opened 2022-07-01
#16007 Highlight word part instead of entire line if no length is known

### LINKED ISSUE #16007: SQL Errror underlining/highlighting/... exceeds actual error
#### System information: 
- Operating system Linux, Debian/SID
- DBeaver Version 22.0.1.202203181646
- Additional extensions none

- Database name and version Postgres 12
- Driver name org.postgresql.Driver

SQL Error underlining/highlighting/... goes from the error position until the end of the query. Severely limits the usefulness of the marking, since the default red and underlining rather blurs instead of highlighting the actual issue.

```
SELECT foo, bar, MAX(baz) FROM xyz ORDER BY bar
```
will cause an error like 
`SQL Error [42803]: ERROR: column "foo" must appear in the GROUP BY clause or be used in an aggregate function. Position: ...`
and the underlining/highlighting/... will start at that position and reach until the end.
Since the position and the offending field (in this case) is known, the marking should really only be on "foo".

ShadelessFox:


uslss:
verified

PR #16943, opened 2022-07-01
#15963 Allow the `Code Editor` preference page to be datasource-specific

### LINKED ISSUE #15963: Add a preference to disable new marker-based issue highlighting on per datasource basis
I think we need to be able to make changes introduced in the https://github.com/dbeaver/dbeaver/issues/15961 on a per-datasource basis.

ShadelessFox:


uslss:
verified

PR #16946, opened 2022-07-01
#16886 Add Copy button for executed statements in Query Log

### LINKED ISSUE #16886: In query log add button Copy near of "Re-execute"
- [x] Add ability to copy executed query in Query Log
- [x] Use DDL label instead of Text icon for Generate DDL from Result Set action

E1izabeth:
Label instead of image for Generate DDL from result set

image

Copy button

image


ShadelessFox (commented):
I would keep the icon along with the label, it looks goofy IMO

E1izabeth:
Agreed

image

uslss:
re-execute button isn't shown for now
image

E1izabeth:
Re-execute button doesn't appear for embedded databases or when Open separate connection for each editor is disabled, because we can't split queries from different editors in this case. All queries for different editors for this db connection mess up in Execution Log and it becomes possible to execute query from script corresponding to another editor in active editor.

uslss:
verified

PR #16960, opened 2022-07-04
#15635 Fix duplicated ddl statements for unique constraint creation

### LINKED ISSUE #15635: [SQL Server] DDL has duplicate statements for unique constraint creation
#### System information: 
- DBeaver version 22.0.0

- Database name and version Microsoft SQL Server 2019

Generated DDL includes the creation of the index twice

1. Create a table with a unique constraint
2. Open DDL tab in table properties

**Expected:** Generated DDL can be copied and executed in a script

**Actual:** DDL includes a separate statement to create an index as well as a parameter inside the create statement

```sql
-- BikeStores.dbo.sysdiagrams definition

-- Drop table

-- DROP TABLE BikeStores.dbo.sysdiagrams;

CREATE TABLE BikeStores.dbo.sysdiagrams (
	name sysname COLLATE Cyrillic_General_CI_AS NOT NULL,
	principal_id int NOT NULL,
	diagram_id int IDENTITY(1,1) NOT NULL,
	version int NULL,
	definition varbinary(MAX) NULL,
	CONSTRAINT PK__sysdiagr__C2B05B619C2E2514 PRIMARY KEY (diagram_id),
	CONSTRAINT UK_principal_name UNIQUE (principal_id,name)
);
CREATE UNIQUE NONCLUSTERED INDEX UK_principal_name ON BikeStores.dbo.sysdiagrams (principal_id, name);
```

E1izabeth:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableManager.java
```diff
@@ -49,6 +49,8 @@
     public static final String BASE_VIEW_NAME = "NewView"; //$NON-NLS-1$
     public static final String BASE_MATERIALIZED_VIEW_NAME = "NewMView"; //$NON-NLS-1$
 
+    private final List<String> skipIndexNames = new ArrayList<>();
```
ShadelessFox: Instances of managers are singletons; they are not created on a per-object basis. You definitely shouldn't be doing that. Instead, store this skip list in the `options` map passed to every method of a manager.

Additionally, you definitely should use `Set` here instead.

Moreover, I believe that instances of `DBSEntityConstraint` in both cases are the same (in constraints and indexes). Can you store their references rather than their names?

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableManager.java
E1izabeth: Thank you for noticing. Actually, this collection should not be a field, it's a method-level variable.

The instances are not the same. They are connected through index field of `SQLServerTableUniqueKey`. In other way each constraint is connected with table index having the same name in constraint.

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/edit/SQLServerTableManager.java
```diff
@@ -118,7 +120,15 @@ public void renameObject(@NotNull DBECommandContext commandContext, @NotNull SQL
...
-    protected boolean isIncludeIndexInDDL(DBRProgressMonitor monitor, DBSTableIndex index) throws DBException {
+    protected boolean isIncludeIndexInDDL(@NotNull DBRProgressMonitor monitor, @NotNull DBSTableIndex index) throws DBException {
+        Collection<? extends DBSEntityConstraint> constraints = index.getTable().getConstraints(monitor);
+        if (constraints.size() > 0 && index.isUnique() && constraints.stream().anyMatch(
+            c -> c instanceof SQLServerTableUniqueKey
+                && c.getConstraintType() == DBSEntityConstraintType.UNIQUE_KEY 
+                && ((SQLServerTableUniqueKey) c).getIndex() == index
+        )) {
+            return false;
+        }
```
ShadelessFox: It's hard to read. Consider rewriting using for-each loop.

E1izabeth: Ok

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/edit/SQLServerTableManager.java
```diff
@@ -122,13 +122,16 @@ public void renameObject(@NotNull DBECommandContext commandContext, @NotNull SQL
...
-        if (constraints.size() > 0 && index.isUnique() && constraints.stream().anyMatch(
-            c -> c instanceof SQLServerTableUniqueKey
-                && c.getConstraintType() == DBSEntityConstraintType.UNIQUE_KEY 
-                && ((SQLServerTableUniqueKey) c).getIndex() == index
-        )) {
-            return false;
+        if (constraints.size() > 0 && index.isUnique()) {
+            for (DBSEntityConstraint constraint : constraints) {
+                if (constraint instanceof SQLServerTableUniqueKey && constraint.getConstraintType() == DBSEntityConstraintType.UNIQUE_KEY
+                    && ((SQLServerTableUniqueKey) constraint).getIndex() == index
```
ShadelessFox:
```suggestion
                if (constraint instanceof SQLServerTableUniqueKey
                    && constraint.getConstraintType() == DBSEntityConstraintType.UNIQUE_KEY
                    && ((SQLServerTableUniqueKey) constraint).getIndex() == index
```

ShadelessFox (approved):

uslss:
verified

PR #16964, opened 2022-07-04
#14285 Add database proposals for data sources with schema active objects

### LINKED ISSUE #14285: Database name not proposed by autocompletion
Hello DBeaver teams,

I got trouble with autocompletion, I juste upgraded from 21.0.1 to 21.2.3.

Before I was able to specify database name prefix with autocompletion, in 3-4 characters :
image

Since 21.2.3 I have to write it manually, autocompletion seems to not complete words using database name :/

E1izabeth:


uslss:
for now, the suggestion isn't shown until the first non-numeric symbol of the DB name is typed
dbeaver_hfGT27PWGN

E1izabeth:
We don't want to auto activate proposals on digits.

uslss:
verified

PR #16968, opened 2022-07-05
#15605 Fix formatter tokenizer for identifiers start with digits for MySQL

### LINKED ISSUE #15605: Issue on to Format SQL option
At Edit menu, Format SQL option (CTRL + Shift + F):
When the table names that contains digits + letters, the formatter SQL converts them to table name with space.

`SELECT * FROM 2021data;`
converts to

```
select
	*
from
	2021 data;
``` 
<<<< MISTAKE (space)

thanks.

#### System information: 
- DBeaver version
- Additional extensions

- Database name and version
- Driver name

E1izabeth:


uslss:
verified

PR #16990, opened 2022-07-06
#16893 Use `export{n}` instead of the table name if it can't be resolved

### LINKED ISSUE #16893: DBeaver v22.1.1 // Export Tasks does not respect ${table}
#### System information: 
- Operating system (distribution) and version: Windows 10 Enterprise 
- DBeaver version: 22.1.1 x64 Portable
- Additional extensions: Microsoft Office Integration & Darkest Theme

- Database name and version: DB2
- Driver name: DB2 JCC (Type 4) 11.5.6.0 FP000
- Do you use tunnels or proxies (SSH, SOCKS, etc)? Yes

I have an export task with multiple queries
As export type I have Microsoft Excel (Microsoft Office Integration)
As an output, I have specified ${tables}
On execution, I expect to get get 4 different xlsx files.
However, I get a single file with the name "export.xlsx", consisting on the export of the last query. 

Reverted to v22.1 and was able to to 4 different xlsx files named each one based on ${table}

ShadelessFox:


uslss:
verified

PR #17022, opened 2022-07-08
#16427 Show "Save As" for SQL console only

### LINKED ISSUE #16427: Save SQL console as script

ShadelessFox:
|SQL Console|SQL Script|
|---|---|
|java_xFoa39REv2|java_8XOuXZUfbH|


DariaMarkaryan:
Verified

Issue #17021, opened 2022-07-08
Add padding for buttons in Event view from Query Log


E1izabeth:
![image](https://user-images.githubusercontent.com/28875055/178005095-b84b27ad-0468-42d0-ad34-c9e84ed3c648.png)

PR #17035, opened 2022-07-08
#17021 Add padding for buttons in Event view of Query Log

E1izabeth:
image


serge-rider:
Do not use explicit paddings in dialogs. We want to use standard ones everywhere.
There is a function createDialogButton for that, I believe.

E1izabeth:
This paddings is to make the size of place between left corner of the window and the left corner of the button and between the right corner of the window and the right corner of the button equal as @ShadelessFox asked. This padding is for Composite type, not for the button itself. The default padding leads to unequal size.

E1izabeth:
image

PR #17045, opened 2022-07-11
#17041 add only data containers for the full-text search results list

### LINKED ISSUE #17041: DB Full-Text search bug in new version
#### System information: 
- Operating system (distribution) and version: MS Win 11 Pro
- DBeaver version: 22.1.2.202207091909

- Database name and version: Oracle 11g
- Driver name: Oracle
- Do you use tunnels or proxies (SSH, SOCKS, etc)? : no

I noticed two problems after installing the latest version of DBeaver:
1. When trying to mark the objects in which want to search, there is no checkBox and cannot see which objects is (un)marked.
2. If I don't mark the objects and search through the entire database, I get an error (picture attached)
checkBox
full_db_search

LonwoLonwo:


uslss:
verified

PR #17046, opened 2022-07-11
#17021 Fix button bar padding in the QM event viewer

### LINKED ISSUE #17021: Add padding for buttons in Event view from Query Log

ShadelessFox:


DariaMarkaryan:
Verified

PR #17049, opened 2022-07-11
#15387 Fix hippie autocompletion for words with prefixes

### LINKED ISSUE #15387: SQL server variable name autocompletion removes @
#### System information: 
- DBeaver version: 21.3.3
- Additional extensions: none

- Database name and version: SQL Server 2016

When I have a declared variable in the script text, and then I type @ and start typing variable name, it appears in autocompletion. Then I select the suggested value from the autocompletion and press Enter. The suggested value is inserted into the editor but the @ at the beginning of the variable is removed.
ezgif-6-dd11933442

```
declare @variable nvarchar(255)
-- then type @ and start typing variable name
-- then select it from autocompletion and hit Enter
```

E1izabeth:


uslss:
verified

PR #17055, opened 2022-07-12
#15777 Throw exception on invalid quote sequence for csv import

### LINKED ISSUE #15777: Import process stopped if last symbol of text field has backslash
original string = ABC "Z" MTK
export string is csv-file as (for example 3 field)
123,"ABC ""Z"" MTK",456
and cannot import second field, write cannot format "Z" to digit.

ver.22.0.0

E1izabeth:
Please, test for different cases.

Should throw with this data when `\` is an escape char in import preferences
```
1,"ABC ""Z"" MTK",4
2,"AR ""Z"" MTK\",5
3,"AB ""Z"" MTK",6
```
image


DariaMarkaryan:
Verified

PR #17058, opened 2022-07-12
#5499 Add support for RLS

### LINKED ISSUE #5499: PostgreSQL Row Security Policies (RLS)
Hello, first, thanks for this awesome software, you guys are doing a great work making something really usable to manage databases!

There is a feature that allow to manage rules in postgres ( https://github.com/dbeaver/dbeaver/issues/2284 ).

I think it could be good to know something similar for Policies: https://www.postgresql.org/docs/current/ddl-rowsecurity.html

```sql
CREATE TABLE public."user" (
	id serial NOT NULL,
	"name" text NULL
);

CREATE POLICY user_read_all_policy ON public.user
  FOR SELECT
  USING (true);

CREATE POLICY user_insert_any_policy ON public.user
  FOR INSERT
  WITH CHECK(true);

ALTER TABLE public.user ENABLE ROW LEVEL SECURITY;
```

Actually, the way to see what policies have a table, we can use:
```sql
select * from pg_policies;
```
imagen

----------------
I hope you can consider this for future releases.

ShadelessFox:
image


DariaMarkaryan:
Verified

PR #17074, opened 2022-07-13
#17070 fix Oracle null/not null column modifier

### LINKED ISSUE #17070: "Not Null" preference is getting lost while creating a new column
#### System information: 
- Operating system (distribution) and version - Windows 10
- DBeaver version - 22.1.1.202206261508

- Database name and version - Oracle 19c

When we try to create a new column using GUI as shown below,

image

and if you select/tick **Not Null** checkbox to make column as not null

image

It shows that you have selected **Not Null** option but when you try to save it you lose that preference as below,

image
 
The generated sql has no Not null in it so when you persist it you see that the column is nullable

LonwoLonwo:
2022-07-13 15_05_26-NEWTABLE_2 - Persist Changes

The creation of new tables and editing of existing ones should be tested.


uslss:
verified

Issue #16411, opened 2022-05-04
Rearrange new table's columns


plotn:
Hi! Could it be possible to realize fields rearrange while they still not stored into DB? see pi
![dbeaver_fields_rearrange](https://user-images.githubusercontent.com/11477081/166720932-7236f7a0-3eeb-4c7e-a9fc-444629e67120.png)
c

ShadelessFox:
Hello @plotn,

Thanks for the feature request.

We can totally do that since the metadata is yet to be persisted, so rearranging these attributes is possible.

E1izabeth:
Connected with #17096

PR #17095, opened 2022-07-15
#16411 Rearrange new table's columns in PostgreSQL

E1izabeth:
Another issue was created for the found bug #17096


serge-rider:
The idea is to make it possible to rearrange new columns for any database, not just PostgreSQL.

PR #17107, opened 2022-07-18
#14959 ability to exclude views from the DDL added to DB2 tables

### LINKED ISSUE #14959: [DB2] Exclude views from generated table DDL
### Discussed in https://github.com/dbeaver/dbeaver/discussions/14955

<sup>Originally posted by **twielens** December 20, 2021</sup>
Hello,

We've been using DBeaver for quite a while and now we finally obtained the rights to use the generate DDL function (DB2LOOK) which is working but the problem is that when I try to get the DDL for a table I also get a lot of views, even views from different schema's. Is there a way to add the -noview option to the db2look command that is used by dbeaver?

I can't put my finger on which specific views it generates the DDL, but it's a whole lot.

Kind regards,

Tim</div>

We can add DDL options for DB2 tables. Like in the Oracle table tab.

LonwoLonwo:
2022-07-18 08_42_33-DBeaver Ultimate 22 1 2 - ACT


uslss:
verified

PR #17137, opened 2022-07-21
#15620 Advanced (multiline) text editor focus handle fix

### LINKED ISSUE #15620: Unaccurate value saving
When editing (adding) column - there is such dialog. 
Type text to column comment and do not press enter, but click on another grid element (e.g. datatype). Text will disappear. This is ugly behaviour, value must be saved not only by enter but by focus loss too.

OS: Any
DBeaver version: 21.3.5

Снимок экрана от 2022-02-24 19-21-32
g

serge-rider:


serge-rider:
Please test the fix in different editors: table column create, data formatting settings, tasks configuration

Matvey16:
Seems like this has broken multi-line edit (it can't be opened)

dbeaver_MWX2prYFgr

serge-rider:
Fixed

Matvey16:
Verified

PR #17150, opened 2022-07-22
#17135 save container mappings changes for tasks

### LINKED ISSUE #17135: Table properties in import task are deleted if the task is edited after the first run
#### System information: 
- Operating system (distribution) and version Win11
- DBeaver version 22.1.2

1. Create an import task with create table mapping and custom table properties in DDL
2. Execute that task
3. Double click the task name and change something insignificant in data load tab
4. Save the task
5. Delete the table created in step 2
6. Execute the task again

**Expected:** Table is created with correct properties
**Actual:** Table is created with default properties and custom properties are not saved in the task

LonwoLonwo:


Matvey16:
Verified

PR #17157, opened 2022-07-22
#17154 Rename root element in the value panel (`<root>` -> `Collection`)

### LINKED ISSUE #17154: Rename root elements of collections in the value panel
Currently, it just shows `<root>`, which is kinda geeky for regular users:

image

We need to name it more friendly.

ShadelessFox:
java_nt7LpDLMg5

Suggestions are welcome.


uslss:
verified

PR #17181, opened 2022-07-27
dbeaver/dbeaver-ee#1507 fix popup selection for cloud files

Destrolaric:
For remote files, we need to create a temp file before we try to open it. This was done when file handling was using system default, but was not when using popup selection. 
P.s Somehow original file new lines used whitespace + CRLF, strange.


Matvey16:
Verified

PR #17186, opened 2022-07-27
CB-2366 web connection configuration types

yagudin10:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/connection/DBPDriver.java
```diff
@@ -185,6 +186,8 @@ public interface DBPDriver extends DBPNamedObject
      */
     DBPDriver createOriginalCopy();
 
+    Set<DBPDriverConfigurationType> getConfigurationTypes();
```
alexander-skoblikov:
```suggestion
    Set<DBPDriverConfigurationType> getSupportedConfigurationTypes();
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/connection/DBPConnectionConfiguration.java
```diff
@@ -107,6 +107,7 @@ public class DBPConnectionConfiguration implements DBPObject {
     private final List<DBWHandlerConfiguration> handlers;
     private final DBPConnectionBootstrap bootstrap;
     private DBPConnectionType connectionType;
+    private DBPDriverConfigurationType configurationType;
```
alexander-skoblikov, line 110: It looks like this field is mostly `null`, it's better to add a default value and annotate get/set methods as `NotNull`

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/driver/DriverDescriptor.java
```diff
@@ -1463,6 +1475,10 @@ public List<DBPDriverLibrary> getOrigFiles() {
         return origFiles;
     }
 
+    public Set<DBPDriverConfigurationType> supportedConfigurationTypes() {
```
alexander-skoblikov:
```suggestion
    public Set<DBPDriverConfigurationType> getSupportedConfigurationTypes() {
```

PR #17214, opened 2022-08-01
#13658 remove search focus to output panel

### LINKED ISSUE #13658: Text search goes on "Output" tab
Run a whatever query, and try to search a text in SQL window...
In this case, search for "THISISABUG"; as you can see, "String not found" in the find window, even if Sql Script window is selected. Keep selecting the Sql Script window, focus keeps going to Output window, and "String not found".

DBeaver Community Edition	21.1.5.202108160727
MacOs BigSur 11.4 (20F71)

image

Destrolaric:
Closes #13658


Matvey16:
Verified

PR #17221, opened 2022-08-01
#15471 add shortened version as tooltip & fix no dialog

### LINKED ISSUE #15471: Grid - Status Bar - Information of fetched rows
Hello,

in the Status Bar it is possible to get the following information since Version 21.3.3:
**1 row(s) fetched - 9ms, on 2022-02-12 at 11:25:07**

With Version 21.3.4 the format has changed to a not persistent storable format (Year is missing, Date is not sortable):
1 row(s) fetched - 9ms, on Feb 02 at 11:25:07

I hope you change the date format again to YYYY-MM-DD, or it is possible to set it in the preferences. I use this in my analyse code.

Furthermore, in the dialog "Status details" there should be a button with "Copy" like in the "Persist Changes" dialog.
And it would be very helpful, if there exists a shortcut to insert this information as comment in the "SQL editor".
(Alternatively, there exists a shortcut to copy this information to the clipboard - but the solution above is more helpful).

Thanks to implement this important feature.

BR
Günter
2022-02-12_DBeaver - Grid Bar

### LINKED ISSUE #16895: Status bar with fetched rows and execution time
#### System information: 
- Windows 10 21H1
- DBeaver version 22.1.1

- Database name and version All

In the previous versions, if you click on the grafik icon, you get the dialog Status Details. After then, it is possible to copy this information.
In 22.1.1 it is not possible to open the Status Details.

grafik

Thanks for resolving this bug.
Günter

See also #15471

Destrolaric:
Closes #15471, Closes #16895
Fix dialog not being shown & show the short date version as a tooltip
java_2noxg4RKx6


uslss:
verified

PR #17230, opened 2022-08-02
#16605 fix focus for statistics tabs

### LINKED ISSUE #16605: Statistics tab not focused after SQL Update
#### System information: 
- Operating system (distribution) and version: MacOS 12.3.1
- DBeaver version: 22.0.5
- Additional extensions: none

- Database name and version: AS/400 and MariaDB
- Driver name: DB2 iSeries/AS 400   and MariaDB
- Do you use tunnels or proxies (SSH, SOCKS, etc)? Issue happens without using tunnel or proxy

In the last few months, I've noticed that after performing an SQL update, the Statistics tab that appears at the bottom of the SQL editor indicating how many rows were affected  is not being focused and instead I see one of the previous results tabs (e.g. the results of a previous select query).  The Statistics tab is being created and I can switch to it but this used to happen automatically and still should IMHO.  

FYI, I have DBeaver set to create a new result tab every time (not sure where this setting is but I remember setting it).  

- Open an SQL editor and connect to DB.
- Execute a select statement several times until there are more result tabs at the bottom that can be displayed and the ">>" icon appears for the overflow
- Select one of the first result tabs
- Execute an UPDATE statement

Expected behavior:  a Statistics tab should appear saying how many rows were affected.  This tab should be visible without having to manually switch to it

Actual behavior: a previous Results tab is displayed showing rows from one of the SELECTS.  The Statistics tab is created but not selected/focused.

Destrolaric:
Closes #16605

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
@@ -4067,6 +4067,11 @@ public class SQLEditor extends SQLEditorBase implements
                     int resultsIndex = 0;
                     for (QueryResultsContainer results : queryProcessor.resultContainers) {
                         if (results.query != query) {
+                            // This happens when query results is statistics tab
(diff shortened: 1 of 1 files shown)


Matvey16:
Verified

PR #17233, opened 2022-08-02
#17192 fix keys metadata reading

### LINKED ISSUE #17192: Updating while the table is sorted corrupts data
#### System information: 
- Operating system (distribution) and version: Ubuntu 20.04
- DBeaver version 22.1.3 CE
- Additional extensions N/A

- Database name and version H2DB
- Driver name JDBC

While the table is sorted, updating cells can corrupt the table:

I ordered by the first column and corrected the shop references(no keys) after the previous corruption and this is what I got:
https://ibb.co/KqsVB4D

But if I use a select from the console then I get this:
https://ibb.co/chP4kQt

Note that when I got all the rows with "shop2" the data in the first view has been saved and no error got thrown!

It just simply inserted "shop2" in every row. When I first loaded the data from a csv, sorted by the middle column and updated the last columns then it renamed all references of "shop2" to "shop1".

Have a simple table with 3 columns. Sort it and updating multiple cells at a time will corrupt the table sooner or later.

LonwoLonwo:
2022-08-02 15_06_39-DBeaver Ultimate 22 1 4 - ALBUM


uslss:
verified

PR #17236, opened 2022-08-02
#14700 ability to store empty properties implemented

### LINKED ISSUE #14700: Driver property cannot be set to null (causes Redshift IAM failure)
#### System information: 
- Operating system (distribution) and version: Mint Linux 20.1
- DBeaver version: 21.2.5.202111141807
- Additional extensions: None

- Database name and version: Any, but can try with postgres
- Driver name: : Any, but can try with postgres
- Do you use tunnels or proxies (SSH, SOCKS, etc)? : No

1. It is not possible to connect to Redshift using IAM *repeatedly*. It is possible to connect once.
2. This is because when saving a connection with no `ssl` value it does "work". But then restarting DBeaver (or reopening the Edit Connection window) the value is set back to `true`.

1. Create any postgres Driver.
2. Set `ssl` to nothing (blank value)
3. Save
4. Reopen, it will have been set back to `true`.

This is a problem as when connecting to Redshift, then `ssl` must be `null`. Note, this is not a User defined property so it is always present. Setting it to `false` / 0 etc is not enough. As the screenshot shows, it must be `null`.

The fix would be that "no value" is stored as `null` persistently.

The workaround is to open DBeaver, then navigate to the connection and to blank out the `ssl` value every time the application is launched.

NB, This doesn't disable SSL when connecting to Redshift, as the `sslmode` can be set.

image

image

Immediate message when restarting DBeaver:

image

LonwoLonwo:
2022-08-02 20_08_00-data-sources json — C__Users_anvolka_AppData_Roaming_DBeaverData_workspace6_Gene

2022-08-03 11_23_27-Connection _dev 2_ configuration

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/json/JSONUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/json/JSONUtils.java
@@ -193,18 +193,18 @@ public class JSONUtils {
     public static void serializeProperties(@NotNull JsonWriter json, @NotNull String tagName, @Nullable Map<String, ?> properties) throws IOException {
         if (!CommonUtils.isEmpty(properties)) {
             json.name(tagName);
-            serializeMap(json, properties, false);
+            serializeMap(json, properties);
         }
     }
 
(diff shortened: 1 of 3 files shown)


Matvey16:
Verified

PR #17239, opened 2022-08-02
#16674 Fix selected rows export with client-side ordering

### LINKED ISSUE #16674: copy as sql return bad result if the result is sorted
#### System information: 
- Operating system (distribution) and version  Windows 11 21H2
- DBeaver version     22.0.5.202205220912
- Additional extensions

- Database name and version   5.6.28-cdb2016-log
- Driver name   com.mysql.cj.jdbc.Driver  Driver for MySQL 8 and later

copy as sql return bad result if the result is sorted

```sql
CREATE TABLE `foo` (
  `a` int(10) unsigned NOT NULL DEFAULT '0' ,
  `b` int(10) unsigned NOT NULL DEFAULT '0' 
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;


insert into foo(a,b)value(1,1001);
insert into foo(a,b)value(2,2001);
insert into foo(a,b)value(3,3001);
insert into foo(a,b)value(4,4001);

# select result
select * from foo;

```
prepare data for table `foo`
image
we click line 3 which value is `(3, 3001)`and `copy as sql` will get result `INSERT INTO foo (a,b) VALUES (3,3001);`
that's ok~
click the button and let `a` col desc like this
image
then clien line 3 which value is `(2,2001)` and `copy as sql` will get result `INSERT INTO foo (a,b) VALUES (3,3001);` , actually expect result is `INSERT INTO foo (a,b) VALUES (2,2001);`
image

[log.log](https://github.com/dbeaver/dbeaver/files/8819807/log.log)
[dbeaver-debug.log](https://github.com/dbeaver/dbeaver/files/8819810/dbeaver-debug.log)

E1izabeth:
Should fix the issue in export and copy as..

### DIFF
--- a/plugins/org.jkiss.dbeaver.data.office.ui/src/org/jkiss/dbeaver/data/office/ui/handlers/OpenSpreadsheetHandler.java
+++ b/plugins/org.jkiss.dbeaver.data.office.ui/src/org/jkiss/dbeaver/data/office/ui/handlers/OpenSpreadsheetHandler.java
@@ -63,9 +63,9 @@ public class OpenSpreadsheetHandler extends AbstractHandler
         IResultSetSelection rsSelection = resultSet.getSelection();
         List<ResultSetRow> rsSelectedRows = rsSelection.getSelectedRows();
         if (rsSelectedRows.size() > 1) {
-            List<Long> selectedRows = new ArrayList<>();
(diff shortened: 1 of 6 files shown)


uslss:
verified

PR #17247, opened 2022-08-03
#16408 Reload syntax rules for editor on format settings change

### LINKED ISSUE #16408: SQL Formatting settings are not applied correctly
Go to Preferences -> Editors -> SQL Editor -> Formatting
1. Uncheck "Use bold font for keywords": does not apply to an existing query (and new queries too)
2. Change a keyword case: does not apply to an existing query, but works with a new one
3. Check "Insert line feed before commas": does not apply to an existing query, but if you select a query and go to Format -> Format SQL in the context menu, it will apply. Now if you uncheck this checkbox the comma will remain in the same place.
---
It make sense to check the behavior of other formatting settings.

E1izabeth:


LonwoLonwo (approved):

Matvey16:
Verified

PR #17249, opened 2022-08-03
#14945 Composite types filtering and sorting

### LINKED ISSUE #14945: Missing parentheses on generated query when filtering/sorting by composite type subfields (Postgres)
#### System information: 
- Operating system (distribution) and version: Windows 10 Pro 21H1
- DBeaver version: 21.3.1
- Additional extensions: Office support, git support, SSHJ, Debug and SVG

- Database name and version: Postgresql 14.1
- Driver name: PostgreSQL JDBC Driver

When using the data editor to filter or sort by a field from a composite type, DBeaver doesn't enclose the name of the composite column in parentheses, causing Postgres to interpret it as the name of a table.

For example, it generates this query:
```pgsql
SELECT x.* FROM public.on_hand x
WHERE x.item."name" IN ('Item 10')
```

Instead of this:
```pgsql
SELECT x.* FROM public.on_hand x
WHERE (x.item)."name" IN ('Item 10')
```

When sorting, the generated query looks like this:
```pgsql
SELECT * FROM public.on_hand
ORDER BY item.price
```

But it also needs the parentheses to work:
```pgsql
SELECT * FROM public.on_hand
ORDER BY (item).price
```

In both cases, Postgres gives the following error:
```
SQL Error [42P01]: ERROR: missing FROM-clause entry for table "item"
```

1. Create an example set on Postgres:

```pgsql
CREATE TYPE inventory_item AS (
	name text,
	supplier_id integer,
	price numeric(5,2)
);

CREATE TABLE on_hand (
	item inventory_item,
	count integer
);

INSERT INTO on_hand
	SELECT
		ROW('Item ' || n, random() * 100, random() * 100)::inventory_item,
		random() * 10000
	FROM generate_series(1, 1000) as n;
```

2. Open table `on_hand` in the data editor and try to filter it by `name`, using the filter button from the column header:

image

Alternatively, just try to sort by any of the 3 fields of composite column `item`.

3. Enjoy your `SQL Error [42P01]: ERROR: missing FROM-clause entry for table "item"` :)

```
!ENTRY org.jkiss.dbeaver.model 4 0 2021-12-17 20:31:18.578
!MESSAGE SQL Error [42P01]: ERROR: missing FROM-clause entry for table "item"
  Position: 40
!SUBENTRY 1 org.jkiss.dbeaver.model 4 0 2021-12-17 20:31:18.578
!MESSAGE SQL Error [42P01]: ERROR: missing FROM-clause entry for table "item"
  Position: 40
!SUBENTRY 2 org.jkiss.dbeaver.model 4 0 2021-12-17 20:31:18.578
!MESSAGE ERROR: missing FROM-clause entry for table "item"
  Position: 40
!STACK 0
org.postgresql.util.PSQLException: ERROR: missing FROM-clause entry for table "item"
  Position: 40
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2440)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2183)
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:308)
	at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:441)
	at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:365)
	at org.postgresql.jdbc.PgStatement.executeWithFlags(PgStatement.java:307)
	at org.postgresql.jdbc.PgStatement.executeCachedSql(PgStatement.java:293)
	at org.postgresql.jdbc.PgStatement.executeWithFlags(PgStatement.java:270)
	at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:266)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCStatementImpl.execute(JDBCStatementImpl.java:330)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCStatementImpl.executeStatement(JDBCStatementImpl.java:130)
	at org.jkiss.dbeaver.model.impl.jdbc.struct.JDBCTable.readData(JDBCTable.java:190)
	at org.jkiss.dbeaver.ui.controls.resultset.ResultSetJobDataRead.lambda$0(ResultSetJobDataRead.java:118)
	at org.jkiss.dbeaver.model.exec.DBExecUtils.tryExecuteRecover(DBExecUtils.java:171)
	at org.jkiss.dbeaver.ui.controls.resultset.ResultSetJobDataRead.run(ResultSetJobDataRead.java:116)
	at org.jkiss.dbeaver.ui.controls.resultset.ResultSetViewer$ResultSetDataPumpJob.run(ResultSetViewer.java:4810)
```

### LINKED ISSUE #14998: Filter dialog doesn't show all options for Postgres composite type subfields, even with "Read from server" checked
#### System information: 
- Operating system (distribution) and version: Windows 10 Pro 21H1
- DBeaver version: 21.3.1
- Additional extensions: Office support, git support, SSHJ, Debug and SVG

- Database name and version: Postgresql 14.1
- Driver name: PostgreSQL JDBC Driver

When I click on the funnel icon at the header of a subfield of Postgres composite column, the filter window that appears for me to select the values only shows those that are already loaded in the data editor. Keeping "Read from server" checked doesn't seem to make a difference.

1. Create testing set:
```pgsql
CREATE TYPE inventory_item AS (
	name text,
	supplier_id integer,
	price numeric(5,2)
);

CREATE TABLE test (
	id integer PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
	item inventory_item
);

INSERT INTO test (item)
	SELECT ROW('Item ' || n, n / 10 + 1, random() * 100)::inventory_item
	FROM generate_series(1, 1000) as n;
```

2. Open table `test` in the data editor with the result-set fetch size set to 200. Make sure it didn't load more than the first 200 rows.

3. Click on the filter icon of column `item.supplier_id`. Make sure "Read from server" is checked. You should see only values 1 to 21, like this:
image

4. If you scroll to the bottom of the currently loaded rows in order to load the next 200, repeating step 3 will show values 1 to 41. And so on.

```
!ENTRY org.jkiss.dbeaver.model 4 0 2021-12-22 14:05:24.613
!MESSAGE SQL Error [42809]: ERROR: "inventory_item" is a composite type
  Position: 35
!SUBENTRY 1 org.jkiss.dbeaver.model 4 0 2021-12-22 14:05:24.613
!MESSAGE ERROR: "inventory_item" is a composite type
  Position: 35
!STACK 0
org.postgresql.util.PSQLException: ERROR: "inventory_item" is a composite type
  Position: 35
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2440)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2183)
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:308)
	at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:441)
	at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:365)
	at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:143)
	at org.postgresql.jdbc.PgPreparedStatement.execute(PgPreparedStatement.java:132)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCPreparedStatementImpl.execute(JDBCPreparedStatementImpl.java:261)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCPreparedStatementImpl.executeStatement(JDBCPreparedStatementImpl.java:205)
	at org.jkiss.dbeaver.model.impl.jdbc.struct.JDBCTableColumn.getValueEnumeration(JDBCTableColumn.java:240)
	at org.jkiss.dbeaver.ui.controls.resultset.valuefilter.GenericFilterValueEdit$5.lambda$0(GenericFilterValueEdit.java:344)
	at org.jkiss.dbeaver.model.exec.DBExecUtils.tryExecuteRecover(DBExecUtils.java:171)
	at org.jkiss.dbeaver.ui.controls.resultset.valuefilter.GenericFilterValueEdit$5.readEnumeration(GenericFilterValueEdit.java:342)
	at org.jkiss.dbeaver.ui.controls.resultset.valuefilter.GenericFilterValueEdit$KeyLoadJob.run(GenericFilterValueEdit.java:633)
```

E1izabeth:
- [x] Use parenthesis only for PostgreSQL
- [x] Add parenthesis for server-side sorting
- [x] Add parenthesis for filtering
- [x] Fix query for getting distinct column values (needs model changes) fixes #14998

About the last todo point.
Let's say that we have such table
image

The problem is that we have wrong query for composite types:
```
SELECT DISTINCT "name"
FROM public.zzz
ORDER BY 1
```
instead of
```
SELECT DISTINCT (composite)."name"
FROM aaa
ORDER BY 1
```
It happens because we have type definition (`public.zzz`) in PostgreDataTypeAttribute instead of the table (`aaa`) where this type is used. And no information about parent (`composite`) of the composite column (`name`) is provided. That's why we have a bunch of bugs (not only with filtering and sorting) connected with composite types. I'm sure that this model changes is the way to start fixing those bugs too.

_I'm not sure how to merge it to existing model (I guess there is no way to do it) without creating new interfaces. Ready for suggestions._
_P.S. I found a bunch of bugs connected with composite types and nested arrays, so, I'm sure that provided model changes is useful_

To be tested:
- composite type filtering
- composite type sorting

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataType.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataType.java
@@ -48,8 +48,9 @@ import java.util.*;
 /**
  * PostgreTypeType
  */
-public class PostgreDataType extends JDBCDataType<PostgreSchema> implements PostgreClass, PostgreScriptObject, DBPQualifiedObject, DBPImageProvider
-{
+public class PostgreDataType extends JDBCDataType<PostgreSchema> 
+    implements PostgreClass, PostgreScriptObject, DBPQualifiedObject, DBPImageProvider, DBSBindableDataType {
+
     private static final Log log = Log.getLog(PostgreDataType.class);
 
     //private static final String CAT_MAIN = "Main";
@@ -530,6 +531,25 @@ public class PostgreDataType extends JDBCDataType<PostgreSchema> implements Post
     @Override
     public DBSEntityType getEntityType() {
         return DBSEntityType.TYPE;
+    }
+
+    @Nullable
+    @Override
+    public List<? extends DBSContextBoundAttribute> bindAttributesToContext(
+        @NotNull DBRProgressMonitor monitor,
+        @NotNull DBSEntity dataContainer,
+        @NotNull DBSEntityAttribute memberContext
+    ) throws DBException {
+        List<PostgreDataTypeAttribute> attrs = this.getAttributes(monitor);
+        if (attrs == null) {
+            return null;
+        }
+    
+        List<PostgreDataBoundTypeAttribute> boundAttrs = new ArrayList<>(attrs.size());
+        for (PostgreDataTypeAttribute attr : attrs) {
+            boundAttrs.add(new PostgreDataBoundTypeAttribute(monitor, (PostgreTableBase) dataContainer, memberContext, attr));
+        }
+        return boundAttrs;
     }
 
     @Override
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
@@ -72,13 +72,43 @@ public final class DBUtils {
     @NotNull
     public static String getQuotedIdentifier(@NotNull DBPNamedObject object)
     {
-        return object instanceof DBSObject ? getQuotedIdentifier(((DBSObject) object).getDataSource(), object.getName()) : object.getName();
+        if (object instanceof DBSContextBoundAttribute) {
+            return ((DBSContextBoundAttribute) object).formatMemberReference(false, null, DBPAttributeReferencePurpose.UNSPECIFIED);
+        } else {
+            return object instanceof DBSObject 
+                ? getQuotedIdentifier(((DBSObject) object).getDataSource(), object.getName()) 
+                : object.getName();
+        }
+    }
+    /**
+     * Get object name in quotes if they are needed.
+
+     * @param object to get identifier of
+     * @return object identifier
+     */
+    @NotNull
+    public static String getQuotedIdentifier(@NotNull DBSObject object) {
+        if (object instanceof DBSContextBoundAttribute) {
+            return ((DBSContextBoundAttribute) object).formatMemberReference(false, null, DBPAttributeReferencePurpose.UNSPECIFIED);
+        } else {
+            return getQuotedIdentifier(object.getDataSource(), object.getName());
+        }
     }
 
+    /**
+     * Get object name in quotes if they are needed.
+
+     * @param object to get identifier of
+     * @param purpose of identifier usage
+     * @return object identifier
+     */
     @NotNull
-    public static String getQuotedIdentifier(@NotNull DBSObject object)
-    {
-        return getQuotedIdentifier(object.getDataSource(), object.getName());
+    public static String getQuotedIdentifier(@NotNull DBSObject object, @NotNull DBPAttributeReferencePurpose purpose) {
+        if (object instanceof DBSContextBoundAttribute) {
+            return ((DBSContextBoundAttribute) object).formatMemberReference(false, null, purpose);
+        } else {
+            return getQuotedIdentifier(object.getDataSource(), object.getName());
+        }
     }
 
     public static boolean isQuotedIdentifier(@NotNull DBPDataSource dataSource, @NotNull String str) {
@@ -1610,16 +1640,47 @@ public final class DBUtils {
         if (object instanceof DBPQualifiedObject) {
             return ((DBPQualifiedObject) object).getFullyQualifiedName(context);
         } else if (object instanceof DBSObject && ((DBSObject) object).getDataSource() != null) {
-            return getObjectFullName(((DBSObject) object).getDataSource(), object, context);
+            return getObjectFullName(((DBSObject) object).getDataSource(), object, context, DBPAttributeReferencePurpose.UNSPECIFIED);
         } else {
             return object.getName();
         }
     }
+    
+    /**
+     * Get the full name of the object.
 
+     * @param dataSource container
+     * @param object object to get name of
+     * @param context evaluation context
+     * @return full name of the object
+     */
     @NotNull
-    public static String getObjectFullName(@NotNull DBPDataSource dataSource, @NotNull DBPNamedObject object, DBPEvaluationContext context)
+    public static String getObjectFullName(
+        @NotNull DBPDataSource dataSource,
+        @NotNull DBPNamedObject object, 
+        @NotNull DBPEvaluationContext context) {
+        return getObjectFullName(dataSource, object, context, DBPAttributeReferencePurpose.UNSPECIFIED);
+    }
+
+    /**
+     * Get the full name of the object.
+
+     * @param dataSource container
+     * @param object object to get name of
+     * @param context evaluation context
+     * @param purpose to use object name to
+     * @return full name of the object
+     */
+    @NotNull
+    public static String getObjectFullName(
+        @NotNull DBPDataSource dataSource,
+        @NotNull DBPNamedObject object, 
+        @NotNull DBPEvaluationContext context,
+        @NotNull DBPAttributeReferencePurpose purpose)
     {
-        if (object instanceof DBPQualifiedObject) {
+        if (object instanceof DBDAttributeBinding) {
+            return ((DBDAttributeBinding) object).getFullyQualifiedName(context, purpose);
+        } else if (object instanceof DBPQualifiedObject) {
             return ((DBPQualifiedObject) object).getFullyQualifiedName(context);
         } else {
             return getQuotedIdentifier(dataSource, object.getName());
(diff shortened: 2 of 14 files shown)


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataType.java
```diff
@@ -532,6 +533,25 @@ public DBSEntityType getEntityType() {
...
+    @NotNull
+    @Override
+    public List<? extends DBSContextBoundAttribute> bindAttributesToContext(
+        @NotNull DBRProgressMonitor monitor,
+        @NotNull DBSEntity dataContainer,
+        @NotNull DBSEntityAttribute memberContext
+    ) throws DBException {
+        List<PostgreDataTypeAttribute> attrs = this.getAttributes(monitor);
+        if (attrs == null) {
+            return null;
```
LonwoLonwo, line 545: But you have @NotNull annotation on this method. Chose one something (:

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBSDataType.java
```diff
@@ -17,6 +17,8 @@
 
 package org.jkiss.dbeaver.model.struct;
 
+import java.util.List;
```
LonwoLonwo: Is it needed here?

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataType.java
E1izabeth: yeah, thanks

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/transformers/ArrayAttributeTransformer.java
```diff
@@ -48,8 +49,8 @@ public void transformAttribute(@NotNull DBCSession session, @NotNull DBDAttribut
         }
         if (collectionType != null) {
             DBSDataType componentType = collectionType.getComponentType(session.getProgressMonitor());
-            if (componentType instanceof DBSEntity) {
-                ComplexTypeAttributeTransformer.createNestedTypeBindings(session, attribute, rows, (DBSEntity) componentType);
+            if (componentType instanceof DBSEntity || componentType instanceof DBSBindableDataType) {
```
alexander-skoblikov: I'm embarrassed that only "Entity" classes was used before, but we also use "DataType" now.
Maybe DBSBindableDataType must also be entity.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -72,13 +72,23 @@ public final class DBUtils {
     @NotNull
     public static String getQuotedIdentifier(@NotNull DBPNamedObject object)
     {
-        return object instanceof DBSObject ? getQuotedIdentifier(((DBSObject) object).getDataSource(), object.getName()) : object.getName();
+        if (object instanceof DBSContextBoundAttribute) {
+            return ((DBSContextBoundAttribute) object).formatMemberReference("");
```
alexander-skoblikov: Can we use `null` instead of empty string?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/transformers/ArrayAttributeTransformer.java
alexander-skoblikov: also looks strange that for "normal" columns we use PostgreDataType (entity), but for composite PostgreDataTypeAttribute (not entity)

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
E1izabeth: We use `null` here to add default container (table) name before attribute name (PostgreDataBoundTypeAttribute:L65 `DBUtils.getQuotedIdentifier(context.getParentObject()`), but we could add another method parameter to specify whether we need to add container name. Then it will look like `return ((DBSContextBoundAttribute) object).formatMemberReference(false, null);` . Or `formatMemberReference(true, containerName)` in another place of usage. How do you think: it will be better?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/transformers/ArrayAttributeTransformer.java
E1izabeth: This check is not used here now as PostgreDataType is DBSEntity and we work only in PostgreSQL now. So, I removed it.

alexander-skoblikov (approved):

Matvey16:
Now composite fields can't be updated because of the excess parenthesis

Generated query before (it works)

```sql

UPDATE "Matvey".aaa
	SET composite."name"=NULL
	WHERE id=8
```
Generated query now (it leads to an error in SQL)

```sql

UPDATE "Matvey".aaa
	SET (composite)."name"=NULL
	WHERE id=8;

```

Matvey16:
Verified

Issue #17276, opened 2022-08-05
Clickhouse columns have incorrect names in special case.


LonwoLonwo:
### Description

DBeaver trims columns and tables names when reading.
Because of that, we can't read table data properly, do filtering via UI, do grouping, etc.

![2022-08-05 18_27_14-Window](https://user-images.githubusercontent.com/45152336/183110069-d6c660cb-688a-4a6f-8194-aa804d0ea481.png)

### DBeaver Version

22.1.3

### Database and driver

Clickhouse any version

### Steps to reproduce

Create table

```sql
CREATE TABLE ` test_table_name_with_spaces ` (
    `col1` String,
    ` col2` String,
    ` col3. col4` String
) ENGINE = Log;
```

or

```sql
CREATE TABLE test (
    `col1` String,
    ` col2` String,
    ` col3. col4` String
) ENGINE = Log;
```

PR #17277, opened 2022-08-05
#17276 ability to read columns/tables names with spaces around

LonwoLonwo:
2022-08-05 19_43_52-Greenshot capture form

2022-08-05 19_43_31-Window

I do not want to not trim catalog/schema names.

And I think we should trim names by default in the case of Generic driver.

Please check:
Filters, data reading, data editing, table DDL, table name with spaces, columns names with spaces. Some others tables/columns manipulations.


plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/jdbc/ClickhouseResultSetMetaDataImpl.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package org.jkiss.dbeaver.ext.clickhouse.model.jdbc;
+
+import org.jkiss.dbeaver.model.exec.jdbc.JDBCResultSet;
+import org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCResultSetMetaDataImpl;
+
+import java.sql.SQLException;
+
+/**
+ * JDBCUtils#normalizeIdentifier method usually use dor columns/tables name reading.
```
E1izabeth: Did you mean this?
```suggestion
 * JDBCUtils#normalizeIdentifier method is usually used for columns/tables name reading.
```

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/jdbc/ClickhouseResultSetMetaDataImpl.java
```diff
@@ -0,0 +1,44 @@
...
+ * normalizeIdentifier trim names, but we can create tables/columns with spaces around their names
+ * Let's just read names from the driver directly
```
E1izabeth, line 27: As I understand extra spaces around table/column names is quite specific for several databases. Why do you create code without trimming only for clickhouse? Could we make trimming database-specific and remove trimming from common code?

LonwoLonwo: Because I'm not sure, that only Firebird adds extra spaces around tables/columns names in the ResultSet. And we have banch of genric databases. I think, that tsis is very risky - do not trim result by default. 
Also I thinks, that spaces around tables/columns names - this is side case (and not the best idea).

E1izabeth: Okay, thanks

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/jdbc/ClickhouseResultSetMetaDataImpl.java
LonwoLonwo: Yeah, thanks!

uslss:
verified

PR #17278, opened 2022-08-05
#17156 change metadata structure tables properties dialog UI

### LINKED ISSUE #17156: Columns in table properties during data import are not properly resized
#### System information: 
- DBeaver version 22.1.3
- 
1. Open import wizard and select a file that will create a new table
2. Open table properties configuration

**Expected:** The values are readable
**Actual:** Column can be to narrow so the value is cropped

dbeaver_nSwk3WNle7

LonwoLonwo:
2022-08-05 20_32_42-Window

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/properties/PropertyTreeViewer.java
+++ b/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/properties/PropertyTreeViewer.java
@@ -59,6 +59,7 @@ import org.jkiss.dbeaver.ui.controls.bool.BooleanStyleDecorator;
 import org.jkiss.dbeaver.ui.internal.UIMessages;
 import org.jkiss.dbeaver.utils.GeneralUtils;
 import org.jkiss.dbeaver.utils.RuntimeUtils;
+import org.jkiss.utils.ArrayUtils;
 import org.jkiss.utils.BeanUtils;
 import org.jkiss.utils.CommonUtils;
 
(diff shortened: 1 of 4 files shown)


plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/properties/PropertyTreeViewer.java
```diff
@@ -275,6 +278,32 @@ public void repackColumns() {
...
+                tree.setRedraw(false);
+                TreeColumn[] columns = tree.getColumns();
+                if (!ArrayUtils.isEmpty(columns) && columns.length > 1) {
+                    for (int i = 0; i < columns.length; i++) {
+                        if (i == 0) {
+                            if (columns[0].getWidth() < MIN_COLUMN_WIDTH) {
+                                columns[0].setWidth(MIN_COLUMN_WIDTH);
+                            }
+                        } else if (columns[i].getWidth() < MAX_COLUMN_WIDTH) {
+                            columns[i].setWidth(MAX_COLUMN_WIDTH);
```
E1izabeth: This constants more look like `FIRST_COLUMN_WIDTH` and `SECOND_COLUMN_WIDTH` or `NAME_COLUMN_WIDTH` and `VALUE_COLUMN_WIDTH`.

LonwoLonwo: Good point

E1izabeth (approved):

Matvey16:
Verified

PR #17290, opened 2022-08-08
dbeaver/dbeaver-ee#1696 fix incorrect selection provider & fix mikami for VQB

Destrolaric:
Linked pr: https://github.com/dbeaver/dbeaver-ee/pull/1843

### DIFF
--- a/plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/part/EntityPart.java
+++ b/plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/part/EntityPart.java
@@ -97,6 +97,17 @@ public class EntityPart extends NodePart {
         }
     }
 
+    /**
+     * VQB isn't supposed to have column relations, so
+     * getModelSourceConnections and getModelTargetConnections should return
+     * only entity connections
+     *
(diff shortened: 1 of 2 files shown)


LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/part/EntityPart.java
```diff
@@ -97,6 +97,17 @@ public void performRequest(Request request) {
...
     }
 
+    /**
+     * VQB isn't supposed to have column relations, so
+     * getModelSourceConnections and getModelTargetConnections should return
+     * only entity connections
+     *
+     * @return is this part VQB related
+     */
+    protected boolean isVQB() {
```
ShadelessFox: Perhaps it should be called `isHideAttributeAssociations` and the default implementation will perform checks for `ERDUIConstants.PREF_ROUTING_TYPE` and `ERDAttributeVisibility.isHideAttributeAssociations`?

I don't think the ERD should know about the VQB in that particular context.

ShadelessFox (changes requested):

ShadelessFox (approved):

Issue #17293, opened 2022-08-08
Global preference links don't work in non-global preferences


E1izabeth:
### Description

https://user-images.githubusercontent.com/28875055/183445261-ee2cc2a7-6e5f-42ff-8c79-355e33b6f56e.mp4

```
java.lang.NullPointerException
       java.lang.NullPointerException: Cannot invoke "org.eclipse.ui.preferences.IWorkbenchPreferenceContainer.openPage(String, Object)" because "<parameter1>" is null
	at org.eclipse.ui.dialogs.PreferenceLinkArea.lambda$0(PreferenceLinkArea.java:65)
	at org.eclipse.swt.events.SelectionListener$1.widgetSelected(SelectionListener.java:84)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:252)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4251)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1066)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1090)
	at org.eclipse.swt.widgets.Widget.sendSelectionEvent(Widget.java:1107)
	at org.eclipse.swt.widgets.Link.wmNotifyChild(Link.java:713)
	at org.eclipse.swt.widgets.Control.wmNotify(Control.java:5800)
	at org.eclipse.swt.widgets.Composite.wmNotify(Composite.java:1961)
	at org.eclipse.swt.widgets.Control.WM_NOTIFY(Control.java:5368)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4802)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5034)
	at org.eclipse.swt.internal.win32.OS.CallWindowProc(Native Method)
	at org.eclipse.swt.widgets.Link.callWindowProc(Link.java:161)
	at org.eclipse.swt.widgets.Widget.wmLButtonUp(Widget.java:1993)
	at org.eclipse.swt.widgets.Control.WM_LBUTTONUP(Control.java:5181)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4783)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5034)
	at org.eclipse.swt.internal.win32.OS.DispatchMessage(Native Method)
[...]

Issue #1793, opened 2017-06-19
DB Compare with DDL generation


lrweck:
Hello!

I would like to suggest a new feature/improvement to the database compare feature.

In the window where you choose the compare settings for the objects there would be an option to create DDL for the missing objects.

For example, in the origin db you have an index that doesn't exist in the target db, so DBeaver generates a "CREATE INDEX" with the same rules as the origin db.

Thanks!

djonatanfav:
**up**

serge-rider:
Duplicate of #957

PR #17294, opened 2022-08-08
#1793 Add more links in preferences to Colors and Fonts pref page

E1izabeth:
image
image

Also the bug was found: https://github.com/dbeaver/dbeaver/issues/17293


LonwoLonwo (approved):

Matvey16:
Those links work strange in some places

If preferences are opened for a specific connection > clicking a link leads to NPE

dbeaver_QsPvPhg4YJ

If they are opened from the icon in SQL Editor > nothing happens

dbeaver_6kwM9bYUDC

E1izabeth:
I created a separate bug for this behavior #17293 when I created this PR. It used to (not) work this way before with other links (for example, Text Editor).

Matvey16:
Verified

PR #17308, opened 2022-08-09
#17028 Restrict editing for BinaryPanelEditor if datasource is readonly

### LINKED ISSUE #17028: Hex editor ignores Read-only status
If the table is read-only, the hex editor should disable the ability to edit table values. But in its current state, this doesn't happen and you can edit those values.
Example query:
`select 'test::bytea`

E1izabeth:
To be tested:
- editing hex in hex editor in Value panel with readonly connection


LonwoLonwo (approved):

Matvey16:
Verified

PR #17318, opened 2022-08-09
#17316 fix incorrect find resources

### LINKED ISSUE #17316: DBeaver no longer associates SQL Editor files with connection
### Description

When you right-clicked a connection and selected "SQL Editor", when you wanted to select a file, it used to consider only the scripts used by that connection. Right now it considers ALL of them. 
If you unclick "Show scripts for all connections" the only change is the icon at the left of any row as shown in these two images
image
image

### DBeaver Version

Community Edition 22.1.4.202208051447

### Operating System

Windows 11

### Database and driver

SQL Server and Oracle

### Steps to reproduce

1) Have two connections for different database servers. Both of them with previously created SQL script files to be opened via SQL Editor
2) Connect to one of the servers, open a SQL file. Change anything.
3) Close the connection to that server, so the file will be saved.
4) Open the other connection, and then choose from SQL Editor "Open recent file"
5) It will open the file of the connection of steps 2 and 3.

Expected behaviour: The second connection should not open the file associated with the first one.

### Additional context

_No response_

Destrolaric:


E1izabeth (approved):

ShadelessFox (changes requested):
I'd use labels here to `continue` the outer loop. It took me a solid minute to figure what this logic does 🤔

ShadelessFox (approved):

Matvey16:
Verified

Issue #1492, opened 2017-03-28
Please add the copy SQL to programming language string


Andyfoo:
Please add the copy SQL to programming language string ：

**SQL Editor:  Right-click menu:** 
**copy SQL string to:** 
                    Java/C# string
                    PHP string
                    C/C++ string
                     ![1](https://cloud.githubusercontent.com/assets/864512/24399221/a64f8b76-13de-11e7-9050-04900bdf349d.png)

Andyfoo:
thank you @serge-rider

mapes610:
Hi, do you can add Delphi support as well? Thank you.

serge-rider:
This could be a good feature.
And it makes sense to make target format configurable.

serge-rider:
4.0.6 will contain copying into Java, C/C++, Delphi, HTML and plain (single line) formats.
Further improvements (including additional formats) should be tracked in separate tickets.

seveneleven:
Where is this feature now?

PR #17322, opened 2022-08-09
#1492 Don't set pinned value to null if column can't be pinned

E1izabeth:
To be tested that all checkboxes in "pinned" column are not checked and are not editable for nosql databases and composite types in postgresql.

**_Ticket for pinned columns improvement https://github.com/dbeaver/dbeaver-ee/issues/1859_**


LonwoLonwo (approved):

Matvey16:
Verified

PR #17345, opened 2022-08-12
#16684 explain parameters added

### LINKED ISSUE #16684: Improving the implementation of the EXPLAIN command
Good afternoon,
I am using the Enterprise edition of DBeaver

Now Dbeaver has a function to get the plan of the query being executed.
However, at the moment:
- Version 22.0.0.202203131528
no support for all possible options for the EXPLAIN command
For Postgres 14 (latest current version) these options are: `ANALYZE`, `VERBOSE`, `COSTS`, `SETTINGS`, `BUFFERS`, `WAL`, `TIMING`, `SUMMARY`.

Currently only supported:
`ANALYZE` (can be controlled) + parameters that have `True` by default: `COSTS` and `TIMING`.

Please add other options to enable them on request.

Why is it important to use `BUFFERS` in your work:
- Nikolai Samokhvalov wrote very well about this:
https://postgres.ai/blog/20220106-explain-analyze-needs-buffers-to-improve-the-postgres-query-optimization-process#recommendations-to-content-creators (en)
https://www.youtube.com/watch?v=qCLRyF4qChI (ru)
https://www.youtube.com/watch?v=TEIXS3qOci0 (ru) (at 53 minutes, a conversation with a developer from DBeaver)

Other options are also important  :) 

I also ask you to give the opportunity to see the full query plan in the form of text, the `FORMAT TEXT` option

Thank you very much for your product, I really hope for your understanding and implementation of this in the next releases.

LonwoLonwo:
2022-08-12 20_06_34-DBeaver Ultimate 22 1 4 - _postgres_ Script-34


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/plan/PostgreExecutionPlan.java
```diff
@@ -92,8 +92,30 @@ public String getPlanQueryString() {
...
+                if (PostgreQueryPlaner.PARAM_TIMING.equals(key)) {
+                    if (!CommonUtils.toBoolean(entry.getValue())) {
+                        explainStat.append(",TIMING FALSE");
+                    }
+                    continue;
+                }
+                // In other cases we can just add parameter keyword
+                if (CommonUtils.toBoolean(entry.getValue())) {
+                    explainStat.append(",").append(key);
+                }
```
E1izabeth: IMHO it can be written easier and more straightforward

```suggestion
                if (PostgreQueryPlaner.PARAM_COSTS.equals(key) && !CommonUtils.toBoolean(entry.getValue())) {
                    explainStat.append(",COSTS FALSE");
                } else if (PostgreQueryPlaner.PARAM_TIMING.equals(key) && !CommonUtils.toBoolean(entry.getValue())) {
                    explainStat.append(",TIMING FALSE");
                } else if (CommonUtils.toBoolean(entry.getValue())) {
                    // In other cases we can just add parameter keyword
                    explainStat.append(",").append(key);
                }
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/plan/PostgreExecutionPlan.java
```diff
@@ -92,8 +92,30 @@ public String getPlanQueryString() {
         if (oldQuery) {
             return "EXPLAIN " + (verbose ? "VERBOSE " : "") + query;
         } else {
-            boolean doAnalyze = CommonUtils.toBoolean(this.configuration.getParameters().get(PostgreQueryPlaner.PARAM_ANALYSE));
-            return "EXPLAIN (FORMAT XML" + (doAnalyze ? ", ANALYSE" : "") + ") " + query;
+            Map<String, Object> parameters = this.configuration.getParameters();
+            StringBuilder explainStat = new StringBuilder(64);
+            explainStat.append("EXPLAIN (FORMAT XML");
```
E1izabeth, line 97: In the original issue user asked:
> I also ask you to give the opportunity to see the full query plan in the form of text, the FORMAT TEXT option

Maybe we should make FORMAT parameter too with options TEXT | XML | JSON | YAML

E1izabeth: Another issue could be created for this.

ShadelessFox (changes requested):
I think it would be better to introduce an enumeration containing all possible options for the `EXPLAIN` command. This way we can construct UI elements in a loop rather than manually creating each individual one.

ShadelessFox (approved):

LonwoLonwo:
2022-08-31 18_54_12-DBeaver Ultimate 22 2 0 - _postgres 9 6_Script-175

Version dependencies added - this is the window for PostgreSQL 9.6

uslss:
options from higher pg versions are trying to applied
dbeaver_FO8yEXRXc8

ShadelessFox (approved):

uslss:
verified

Issue #17248, opened 2022-08-03
Array expand items doesn't work for NoSQL databases


E1izabeth:
### Description

https://user-images.githubusercontent.com/28875055/182627551-1e25807c-8ba7-405f-909a-0c23f744bdae.mp4

### DBeaver Version

DBeaver 22.1.4

PR #17351, opened 2022-08-15
#17248 Fix collection elements expand in documents

ShadelessFox:
java_nC997PLAtF

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
@@ -761,84 +761,64 @@ public final class DBUtils {
     @Nullable
     public static Object getAttributeValue(
         @NotNull DBDAttributeBinding attribute,
-        DBDAttributeBinding[] allAttributes,
-        Object[] row,
-        int[] nestedIndexes)
-    {
+        @NotNull DBDAttributeBinding[] allAttributes,
+        @NotNull Object[] row,
+        @Nullable int[] nestedIndexes
+    ) {
         if (attribute.isCustom()) {
-            return DBVUtils.executeExpression(((DBDAttributeBindingCustom)attribute).getEntityAttribute(), allAttributes, row);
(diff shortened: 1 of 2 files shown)


E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -761,84 +761,64 @@ public static Object getAttributeValue(
...
                 break;
             }
-            DBDAttributeBinding attr = attribute.getParent(depth - i - 1);
-            assert attr != null;
-            try {
-                int nestedIndex = 0;
-                if (curValue instanceof DBDCollection) {
-                    nestedIndex = nestedIndexes == null ? 0 : nestedIndexes[indexNumber++];
+
+            final int nestedIndex;
```
E1izabeth, line 793: Maybe it would be better to create a variable outside the loop, and reset it to 0 here.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -761,84 +761,64 @@ public static Object getAttributeValue(
...
-                } else {
-                    if (i == depth - 1) {
-                        // Sibling non-collection attribute
-                        return DBDVoid.INSTANCE;
-                    }
-                }
+        }
+
+        while (nestedIndexes != null && curNestedIndex < nestedIndexes.length && curValue instanceof List<?>) {
+            final int nestedIndex = nestedIndexes[curNestedIndex++];
```
E1izabeth, line 814: And here

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
ShadelessFox: I want to keep it `final` to minimize the total amount of assignments to it. I believe it improves visual code flow.

Matvey16:
Verified

E1izabeth (approved):

ShadelessFox:
Consumed by #17418

PR #17373, opened 2022-08-17
#16996 Fix column length calculation for unicode types

### LINKED ISSUE #16996: Azure SQL nvarchar columns show incorrect lengths
#### System information: 
- DBeaver version: 22.1.1

- Database name and version: Azure SQL
- Driver name: com.microsoft.sqlserver.jdbc.SQLServerDriver

When viewing tables with nvarchar columns in Azure SQL The column lengths are divided by 2, which reports an incorrect column length.

I believe it was due to this work around which should be reverted: https://github.com/dbeaver/dbeaver/commit/7d2a5551a27d801a60b62eb5e71c230c5f4f56c9

Which stemmed from this: https://github.com/dbeaver/dbeaver/issues/5981

Here is an example of what SSMS (SQL Server Management Studio) shows:

image

As shown in DBeaver (notice FileName shows 127 instead of the correct value of 255):

image

Table example script:

```sql
CREATE TABLE [dbo].[Downloads](
	[DownloadId] [int] IDENTITY(1,1) NOT NULL,
	[CenterId] [int] NOT NULL,
	[FileName] [nvarchar](255) NOT NULL,
	[Description] [nvarchar](max) NOT NULL,
	[FileDataId] [int] NULL,
	[CreateDate] [datetime2](7) NOT NULL,
	[UpdateDate] [datetime2](7) NULL,
 CONSTRAINT [PK_Downloads] PRIMARY KEY CLUSTERED 
(
	[DownloadId] ASC
)WITH (STATISTICS_NORECOMPUTE = OFF, IGNORE_DUP_KEY = OFF, OPTIMIZE_FOR_SEQUENTIAL_KEY = OFF) ON [PRIMARY]
) ON [PRIMARY] TEXTIMAGE_ON [PRIMARY]
GO
```

ShadelessFox:


E1izabeth (approved):

uslss:
image
image
Still reproducible

ShadelessFox:
Should be tested in both SQL Server and Azure SQL.

uslss:
fixed for Azure, still reproducible for SQL Server

uslss:
verified

PR #17374, opened 2022-08-17
#16431 resolve driver properties

### LINKED ISSUE #16431: use variables in connection's driver properties won't work
#### System information: 
- Ubuntu20.04
- DBeaver version 22.04

- Mariadb
- mariadb jdbc
- socks 

If I use full path in the driver properties it works, but neither ${home} nor ${HOME} work as path.

Destrolaric:
Closes #16431 
Has limitations:
1. Resolve will happen once, after that resolved value will be used
2. Resolving happens only in connection configuration stage

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/DriverPropertiesDialogPage.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/DriverPropertiesDialogPage.java
@@ -27,9 +27,13 @@ import org.eclipse.swt.widgets.Link;
 import org.jkiss.dbeaver.model.DBPDataSourceContainer;
 import org.jkiss.dbeaver.model.connection.DBPConnectionConfiguration;
 import org.jkiss.dbeaver.registry.DataSourceDescriptor;
+import org.jkiss.dbeaver.registry.driver.DriverDescriptorSerializer;
 import org.jkiss.dbeaver.runtime.properties.PropertySourceCustom;
 import org.jkiss.dbeaver.ui.UIUtils;
+import org.jkiss.dbeaver.ui.controls.VariablesHintLabel;
 import org.jkiss.dbeaver.ui.internal.UIConnectionMessages;
+import org.jkiss.dbeaver.utils.GeneralUtils;
+import org.jkiss.dbeaver.utils.SystemVariablesResolver;
 import org.jkiss.utils.CommonUtils;
 
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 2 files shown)


E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/properties/PropertySourceCustom.java
```diff
@@ -98,6 +99,8 @@ public void addDefaultValues(Map<String, ?> defaultValues)
     public Map<String, Object> getPropertyValues() {
         Map<String, Object> allValues = new HashMap<>(originalValues);
         allValues.putAll(propValues);
+        allValues.values().forEach(it -> GeneralUtils.replaceVariables(CommonUtils.toString(it),
+            SystemVariablesResolver.INSTANCE));
```
ShadelessFox: I'm suspicious about converting property values from * something * into string just to expand variables.

If variable values can only be strings, then it's better to return `Map<String, String>` instead.

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/DriverPropertiesDialogPage.java
```diff
@@ -123,7 +127,8 @@ public void saveSettings(DBPDataSourceContainer dataSource) {
             for (Map.Entry<String, Object> entry : propertySource.getPropertyValues().entrySet()) {
                 String propName = CommonUtils.toString(entry.getKey());
                 if (!propName.isEmpty()) {
-                    properties.put(propName, CommonUtils.toString(entry.getValue()));
+                    properties.put(propName, GeneralUtils.replaceVariables(CommonUtils.toString(entry.getValue()),
+                        new SystemVariablesResolver()));
```
ShadelessFox, line 131: See previous comment

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/DriverPropertiesDialogPage.java
```diff
@@ -174,7 +179,11 @@ public void widgetSelected(SelectionEvent e) {
             }
             netConfigLink.setLayoutData(new GridData(GridData.HORIZONTAL_ALIGN_END));
         }
-
+        VariablesHintLabel variablesHintLabel = new VariablesHintLabel(ph,
```
ShadelessFox, line 182: Redundant variable declaration

ShadelessFox (changes requested):

Matvey16:
Verified

PR #17382, opened 2022-08-18
#16749 add complex values inline editing

### LINKED ISSUE #16749: Need an option to turn off array editor in data grid (PostgreSQL)
In DBeaver 22.1 I can't find an option to edit PostgresSQL array values like a normal text field. When I try to edit an array value a separate panel opens with an array editor described in issue #403. I don't want this as it makes editing much more complicated:

- it is difficult to use it with keyboard (if possible at all - I couldn't find a way to add a new element with keyboard) so I need to switch to mouse trying to aim at the small buttons
- the value panel opens up to the right and takes space unnecessarily and immediately after editing I need to close it

I would like to edit arrays like I could before this feature was implemented - so that an array value is editable like a text value: pressing Enter allows me to edit it inline and pressing Shift+Enter opens a larger text editor where I can freely edit the value in plain text. When I have values like `{"F","S"}` it much easier for me to edit them like text instead of playing with the special editor.

Please add an option to disable the array editor and display the array values in the native format - I can see that currently arrays of string don't display quotes - `{F,S}` - for inline editing I suppose we would need to have it in PostgreSQL format like `{"F","S"}`.

Destrolaric:
Limitations:
1. As most functionality of array editing works strangely in MongoDB.
2. Editing multidimensional arrays produces an error because we need to have the ability to store internal arrays inside of JDBCCollection, and for now we don't have that.


E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Changes in one of the subfields sets the entire array to NULL

dbeaver_YeCDzrmsN9

PR #17406, opened 2022-08-21
#17288 Fix typo in single connection setter of driver descriptor and add single connection for Snowflake

### LINKED ISSUE #17288: Azure SQL Server MFA Authentication Causes Repeated Browser Windows (not yet fixed!)
### Description

For most databases I authenticate using "Azure SQL Server MFA Authentication". Since version 22.1 a new browser tab opens every time I connect or reconnect to a database, this is also true for automatic reconnects after some idle time. The browser tab points to https://localhost:<someport> and only says "Authentication complete. You can close the browser and return to the application."

A similar issue was raised [here](https://github.com/dbeaver/dbeaver/issues/16873) before that was fixed in 22.1.3, but I still get tons of browser tabs. Running 22.1.4 now and the problem still persists.

Expected behavior: No browser tabs are created when connecting/reconnecting to a database.

I'd also be happy if there's some kind of keepalive for the connection that reduces the amount of reconnects.

### DBeaver Version

Community edition 22.1.4

### Operating System

Windows 10 x64

### Database and driver

Azure SQL / Azure SQL Server MFA Authentication

### Steps to reproduce

- Connect to Azure SQL Server using "Active Directory - MFA"
- A browser window will open that only says "Authentication complete. You can close the browser and return to the application."
- Leave the session idle for some time (until it disconnects, this usually is 15 minutes or so)
- Send a SQL query => automatic reconnect => another tab opens.
- ...

### Additional context

_No response_

E1izabeth:


ShadelessFox (approved):

ShadelessFox (approved):

uslss:
verified

PR #17417, opened 2022-08-22
#16208 UUID notification

### LINKED ISSUE #16208: Generate UUID command needs to grayed out if it cannot be worked
when nothing in main window, uuid command(ctrl+alt+shift+u) in edit panel is activated looks like it can be used. but if I command it, there are noting action after command. So I think I need to be grayed out when it cannot be worked.

Destrolaric:
Closes #16208

### DIFF
--- a/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/CoreMessages.java
+++ b/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/CoreMessages.java
@@ -608,6 +608,9 @@ public class CoreMessages extends NLS {
 	public static String monitor_panel_transaction_monitor_tip;
 	public static String variable_host_or_database;
 
+    public static String notification_org_jkiss_dbeaver_ui_actions_common_uuid_copy;
+    public static String notification_org_jkiss_dbeaver_ui_actions_common_uuid_copy_text;
+
 
     static {
 		// initialize resource bundle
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/CoreMessages.java
```diff
@@ -608,6 +608,9 @@ public class CoreMessages extends NLS {
 	public static String monitor_panel_transaction_monitor_tip;
 	public static String variable_host_or_database;
 
+	public static String notification_org_jkiss_dbeaver_ui_actions_common_uuid_copy;
+	public static String notification_org_jkiss_dbeaver_ui_actions_common_uuid_copy_text;
+
```
ShadelessFox: Please replace tabs with spaces at least within new additions

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/CoreResources.properties
```diff
@@ -580,5 +580,8 @@ variable_host_or_database = host or database
 toolbar_checkbox_connection_readonly_tooltip = Make {0} read-only
 toolbar_checkbox_connection_not_readonly_tooltip = Make {0} not read-only
 
+notification_org_jkiss_dbeaver_ui_actions_common_uuid_copy = UUID value copied
+notification_org_jkiss_dbeaver_ui_actions_common_uuid_copy_text = Generated UUID value copied to the clipboard
```
ShadelessFox:
```suggestion
notification_org_jkiss_dbeaver_ui_actions_common_uuid_copy_text = Generated UUID value was copied to the clipboard
```

ShadelessFox (changes requested):

ShadelessFox (approved):

E1izabeth (approved):

Matvey16:
Verified

PR #17424, opened 2022-08-22
#16526 Fix query without from transform for SQL Server

### LINKED ISSUE #16526: [SQL Server] select sequence error
#### System information: 
- Operating system (distribution) and version: windows 10 (19044.1645)
- DBeaver version: 22.0.4
- Additional extensions

- Database name and version: sql server
- Driver name: Microsoft JDBC Driver for SQL Server (MSSQL)

select sequence with this clause 'select next value for seq_name' will produce this error:
SQL Error [11739] [S0001]: NEXT VALUE FOR function cannot be used if ROWCOUNT option has been set, or the query contains TOP or OFFSET.

But if use sequence in a insert clause 'insert into table (col1) values (next value for seq_name)', it is normal.

E1izabeth:
Please, test that limit works for other databases and other queries.


LonwoLonwo (approved):

uslss:
verified

serge-rider:
I'm afraid that it may fail in queries like
```sql
select xx.nextval, a,b,c from thetable where a>0
```

Nextval itself doesn't break limit.

uslss:
verified

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDataSource.java
```diff
@@ -533,6 +536,25 @@ public boolean isForceTransform(DBCSession session, SQLQuery sqlQuery) {
...
+        boolean hasNextValExpr = false;
+        try {
+            Statement statement = SQLSemanticProcessor.parseQuery(this.sqlDialect, query.getText());
+            if (statement instanceof Select) {
+                SelectBody selectBody = ((Select) statement).getSelectBody();
+                if (selectBody instanceof PlainSelect) {
+                    hasNextValExpr = ((PlainSelect) selectBody).getSelectItems().stream().anyMatch(
+                        item -> (item instanceof SelectExpressionItem) 
+                            && (((SelectExpressionItem) item).getExpression() instanceof NextValExpression)
+                    );
```
serge-rider: Does ANY nextval expression in SELECT ... restricts setLimit?
Is it possible to run 
`select xxx.nextval, t.col1 FROM table t`

I think query must not operate with tables.

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDataSource.java
```diff
@@ -533,6 +536,25 @@ public boolean isForceTransform(DBCSession session, SQLQuery sqlQuery) {
         return false;
     }
 
+    public boolean isLimitApplicableTo(SQLQuery query) {
```
serge-rider, line 533: Javadocs for public API

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -1390,10 +1391,17 @@ public static DBCStatement makeStatement(
             makeStatement(session, queryText, doScrollable);
         dbStat.setStatementSource(executionSource);
 
+        boolean isShouldSetLimit = true;
```
serge-rider: Please move it to line 1351 or around. Also leave transformProvider var declaration where it was before. It is not needed in other code blocks.

E1izabeth: I can't have transformProvider var declaration inside that if block as I need transformProvider here on the line 1395.

uslss:
verified

PR #17431, opened 2022-08-23
#16425 Respect `#` comments when extracting select items from the query

### LINKED ISSUE #16425: Dbeaver is throwing error virtual_pk is missing for commented comma
First of all thanks for great software. It helps a lot!

#### System information: 
- Operating system (distribution) and version - Mac Big Sur 
- DBeaver version - 22.0.3.202204170718
- Additional extensions - None

- Database - AWS Aurora
- Driver name - MySQL
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No proxies/tunnels

Having comment with "," breaks functionality while deleting rows and it shows below error "Attributes of unique key 'VIRTUAL_PK' are missing in result set" 
image
 
- Use below script to create required test table
- `CREATE TABLE employee_test (
  employee_id int(11) NOT NULL AUTO_INCREMENT,
  employee_name varchar(255) NOT NULL,
  PRIMARY KEY (employee_id)
) ENGINE=InnoDB AUTO_INCREMENT=1 DEFAULT CHARSET=utf8;
insert into employee_test(employee_name) values ('test1');
insert into employee_test(employee_name) values ('test2');
`
- Important step - Invalidate/Reconnect connection to your database.
- Run below query as exactly typed with "#," appended on first line 
- `select #,
o.* from employee_test o;`
- Try to delete the rows by selecting all rows in the result panel, clicking on delete rows from status bar and then save.
- 

https://user-images.githubusercontent.com/31636584/167017613-5b1c788b-07bb-4730-9deb-2d8aa9cb593f.mov

Note that this is happening only with Mac version and Windows version works correctly. Although I mentioned latest dbeaver version, it was happening with earlier versions as well.

ShadelessFox:
java_JD2g9LmEbW

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLQuery.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLQuery.java
@@ -17,6 +17,7 @@
 
 package org.jkiss.dbeaver.model.sql;
 
+import net.sf.jsqlparser.expression.Expression;
 import net.sf.jsqlparser.schema.Column;
 import net.sf.jsqlparser.schema.Database;
 import net.sf.jsqlparser.schema.Table;
@@ -47,6 +48,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * SQLQuery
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLQuery.java
```diff
@@ -209,6 +211,25 @@ private void parseQuery() {
         }
     }
 
+    private boolean isValidSelectItem(@NotNull SelectItem item) {
+        // Workaround for JSQLParser not respecting the `#` comment in MySQL and treating them as valid values
```
E1izabeth, line 215: Maybe we could create an issue for them.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLQuery.java
ShadelessFox: There's an issue and they won't fix it. https://github.com/JSQLParser/JSqlParser/issues/454

uslss:
veriied

PR #17446, opened 2022-08-24
Update zh translation.

fengchao:
Update zh translation.


E1izabeth (commented):
There is no need to copy English translation if some resources should not be translated.
```
Bundle-Vendor=JKISS
Bundle-Name=DBeaver IDE
productName=DBeaver
productWebsite=https://dbeaver.io/
```
 If there is no some resource in _zh.properties bundle, then the default English translation is used.

fengchao:
Directly copied translations are removed. Please check again.

E1izabeth:
Thank you for contribution.

PR #17447, opened 2022-08-24
#16995 Add Bool type and value handler for Clickhouse

### LINKED ISSUE #16995: Object parameter [true] binding not supported(DBeaver 21.3.2, ClickHouse 22.6.1.1985)
When I manually insert/modify data, the field of type Bool is set to true and an error is reported, But I have no problem using sql to insert

image
image

image

E1izabeth:


ShadelessFox (approved):

uslss:
verified

Issue #17065, opened 2022-07-12
Cannot view Snowflake procedures


StefanoCocciaThryv:
#### System information: 
- Windows 10 Enterprise  -  21H2
- DBeaver version 22.1.2.202207091909, Rel. July 11, 2022
- Additional extensions

#### Connection specification:
- Snowflake
- 3.13.6-jar
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No

#### Describe the problem you're observing:

Getting this whenever opening any procedure. Started getting this error right after last Dbeaver update.
/* ERROR WHILE READING SOURCE:

SQL Error [1003] [42000]: SQL compilation error:
parse error line 1 at position 32 near '95'.
*/

LonwoLonwo:
Hello @StefanoCocciaThryv 

Can you please provide a procedure DDL example?

E1izabeth:
This issue is about opening procedures via Navigator. 

This options are broken for Snowflake:
- open procedure
- rename procedure (also that for some reason rename isn't applied if Procedure view is opened, last seconds of the video)
- delete procedure

https://user-images.githubusercontent.com/28875055/178765773-bf62909f-9d4a-4352-801c-d09b1c0c51f0.mp4

LonwoLonwo:
> These options are broken for Snowflake:

In the case of the "ILIKE" schema name. This is the side case with schema name as a keyword.

LonwoLonwo:
@StefanoCocciaThryv logs will be also very helpful for us.
You can find instructions here: https://github.com/dbeaver/dbeaver/wiki/Log-files

LonwoLonwo:
Now I can not see procedure parameters because of extra quotes.

`-- Load procedure columns ['TEST_DB', PUBLIC, PROC1, %] `

It does not work.

We should quote only keywords or names with special characters.

![2022-08-29 16_39_18-DBeaver Ultimate 22 2 0 - PROC1(ID FLOAT)](https://user-images.githubusercontent.com/45152336/187214579-dbc00c79-38bc-4076-b232-15bbbddaa7e0.png)

ShadelessFox:
Unfortunately, this bug can't be fixed from our side. Our attempt to fix it resulted in other bugs mentioned in https://github.com/dbeaver/dbeaver/issues/17065#issuecomment-1230322892.

We need to update the driver after they fix this issue: https://github.com/snowflakedb/snowflake-jdbc/issues/550

[...]

PR #17462, opened 2022-08-25
#17065 Escape `catalog` when calling `getProcedureColumns`

ShadelessFox:
I don't like this solution. Any suggestions are welcome.


E1izabeth (approved):

uslss:
still reproducible
image
image

uslss:
verified

PR #17481, opened 2022-08-26
Merge generator fixed #7853

### LINKED ISSUE #7853: Add PostgreSQL-specific generate SQL
I am talking about:

Specific-merge: insert ... on conflict do update ... excluded
Specific delete: DELETE FROM table USING another_table
Specific update: update .. from ...

gensql

PS: It may require additional user's input, e.g. "specific delete" needs to request "another_table" real name, thus field list should be generated from primary key columns...

plotn:
Hi! This is my first MR so be indulent if I did somesthing wrong.

The details is here: https://github.com/dbeaver/dbeaver/issues/7853

Additionally:
- Merge generator fixed (removed special fields);
- Alter type add value added.

I have no idea how to change visibility of these items on non-PostgreSQL objects, so they are visible everywhere (I think this is not a problem) - but if you know, I'd be grateful.


ShadelessFox:
Thanks for the contribution.

You can specify `org.jkiss.dbeaver.ext.postgresql.model.PostgreTableBase` instead of `org.jkiss.dbeaver.model.struct.DBSDataManipulator` for these generators, it should work for PostgreSQL tables and views only after that.

Destrolaric:
@plotn Thank you for your contribution! It will be good if commits and title of pr will contain ticket numbers like this:
`#7853 commit description`
And as I see by check style you have a tab configured for new files. Please change it to 4 spaces instead.

plotn:
> You can specify

No, it doesnt work

ShadelessFox:
You need to define these generators inside the PostgreSQL plugin.

plotn:
I'll rework

PR #17483, opened 2022-08-26
#7853 PostgreSQL specific templates added

### LINKED ISSUE #7853: Add PostgreSQL-specific generate SQL

plotn:
delete using, insert on conflict, add value to type, update from, check procedure


plotn:
god damn ))) I passed thru all checks!

ShadelessFox:
I can still see tabs used along with spaces. Can you please replace all tabs with 4 spaces?

plotn:
I added one new file. And i cant open "details" to see what's wrong - seems the service is down

plotn:
> I can still see tabs used along with spaces. Can you please replace all tabs with 4 spaces?

I removed all the tabs but some checks fails. The "details" still could not be opened, what should I do, @ShadelessFox ?

Destrolaric:
@plotn Hi, you have a failed compilation check, see the error below. I think it should be checkGenerator.  also one of your commits doesn't contain the proper ticket number and the ticket number in the title should be before the text description.

```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:2.7.4:compile (default-compile) on project org.jkiss.dbeaver.ext.postgresql.ui: Compilation failure: Compilation failure: 

[ERROR] /var/jenkins_home/jobs/DBeaverDesktop/jobs/dbeaver-ce-github-checks/branches/PR-17483/workspace/dbeaver/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/PostgreSourceViewEditor.java:[111] 

[ERROR] 	DBRRunnableWithResult<String> generator = SQLEditorHandlerCheckProcedureConsole.CHECK_GENERATOR(entities);

[ERROR] 	                                                                                ^^^^^^^^^^^^^^^

[ERROR] The method CHECK_GENERATOR(List<DBSProcedure>) is undefined for the type SQLEditorHandlerCheckProcedureConsole

[ERROR] 1 problem (1 error)

[ERROR] -> [Help 1]

[ERROR] 

```

Destrolaric:
@plotn Hi, as I see by the checkstyle, you need to swap your internal java imports and other imports. This will resolve the issue.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/PostgreSourceViewEditor.java
```diff
@@ -82,6 +97,31 @@ public void run() {
                         refreshPart(PostgreSourceViewEditor.this, true);
                     }
                 }, true));
+            contributionManager.add(ActionUtils.makeActionContribution(
+                new Action("Check", Action.AS_PUSH_BUTTON) {
+                    {
+                        setToolTipText("Check (via plpgsql_check)");
+                    }
+
```
Destrolaric: I think it should be made via localization files.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/sql/handlers/SQLEditorHandlerCheckProcedureConsole.java
```diff
@@ -0,0 +1,105 @@
...
+            }
+        }
+        if (navContext == null || navContext.getDataSourceContainer() == null) {
+            log.debug("No active datasource");
+            return null;
+        }
+
+        DBRRunnableWithResult<String> generator = checkGenerator(entities);
+
+        String title = "Stored procedures check";
```
Destrolaric: I think it should be made via localization files.

plotn:
@Destrolaric , help please - what does it mean "Import statement for org.eclipse.core.runtime.CoreException is in the wrong order. Should be in the THIRD_PARTY_PACKAGE group, expecting not assigned imports on this line."

???

How should I do imports? Just sort alphabetically?

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/sql/handlers/SQLEditorHandlerCheckProcedureConsole.java
```diff
@@ -0,0 +1,105 @@
...
+        if (entities.size() == 1 && !CommonUtils.isEmpty(procName)) {
+            title = procName + " check";
+        }
+
+        try {
+            openConsoleCheck(workbenchWindow, generator, navContext, title, true, currentSelection);
+        } catch (Exception e) {
+            DBWorkbench.getPlatformUI().showError("Open console", "Can open SQL editor", e);
```
Destrolaric: I think it should be made via localization files.

plotn:
"need to swap your internal java imports and other imports" - what is internal? import org.jkiss.dbeaver.ext.postgresql ?
Others - external? Internal first, external last? Or vice versa? Both lists alphabetically?

plugins/org.jkiss.dbeaver.ext.postgresql/plugin.xml
```diff
@@ -743,5 +743,21 @@ SELECT
...
+        </generator>
+        <generator id="tableUpdateFrom" class="org.jkiss.dbeaver.ext.postgresql.model.sql.generator.SQLGeneratorUpdateFrom" label="UPDATE FROM" description="UPDATE FROM (UPDATE with JOIN)" order="8">
+            <objectType name="org.jkiss.dbeaver.ext.postgresql.model.PostgreTableBase"/>
+        </generator>
+        <generator id="tableDeleteUsing" class="org.jkiss.dbeaver.ext.postgresql.model.sql.generator.SQLGeneratorDeleteUsing" label="DELETE USING" description="DELETE USING (DELETE with JOIN)" order="9">
+            <objectType name="org.jkiss.dbeaver.ext.postgresql.model.PostgreTableBase"/>
+        </generator>
+        <generator id="alterTypeAddValue" class="org.jkiss.dbeaver.ext.postgresql.model.sql.generator.SQLGeneratorTypeAddValue" label="ADD VALUE" description="ALTER TYPE ADD VALUE" order="1002">
+            <objectType name="org.jkiss.dbeaver.ext.postgresql.model.PostgreDataType"/>
+        </generator>
```
Destrolaric, line 760: I think label and description should be done via localization.

Destrolaric (changes requested):
The contribution looks good. But it's preferable to have the ability to localize those strings, documentation of how it may be done can be found here https://github.com/dbeaver/dbeaver/wiki/Localization. For other strings which should not be localized, please mark them as `//$NON-NLS-1$`.
P.S this is optional

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/PostgreSourceViewEditor.java
```diff
@@ -17,20 +17,35 @@
 
 package org.jkiss.dbeaver.ext.postgresql.ui.editors;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import org.eclipse.core.runtime.CoreException;
```
Destrolaric: The error appears due to internal java imports being first in PostgreSourceViewEditor

plugins/org.jkiss.dbeaver.ext.postgresql/plugin.xml
plotn: I guess this case should not be translated - they are SQL expressions. "Basic" templates are the same - INSERT, UPDATE, MERGE. I see only one word, that could be translated - "with". 
Are you sure we need to translate it?

Destrolaric: Yep, my bad, I agree.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/sql/handlers/SQLEditorHandlerCheckProcedureConsole.java
```diff
@@ -0,0 +1,105 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2022 DBeaver Corp and others
+ * Copyright (C) 2010-2017 Eugene Fradkin (eugene.fradkin@gmail.com)
```
LonwoLonwo: 4 line is redundant here

plotn:
Collegues. is everything ok now?

E1izabeth:
When adding localization to resource files, I guess you used some plugin which reordered all strings in this files. It would be nice to revert these unintended changes and add only a couple of lines which you really need.
Thank you for your corrections. We appreciate that.

plotn:
@E1izabeth , I followed by @Destrolartic recommendation - https://github.com/dbeaver/dbeaver/wiki/Localization
using ResourceBundle Editor. The reordering result - its work. If I shoud not use this way - then this documentation isnt correct, am I right?

E1izabeth (approved):

Destrolaric:
@plotn It's correct, but I think we need to add info about using IntelliJ IDEA for localization. Most of the devs use it, which is why we have different formatting.

plotn:
Ok, I'll change.
Add some notes into those document for further juniors.

ShadelessFox:
> The reordering result - its work. If I shoud not use this way - then this documentation isnt correct, am I right?

In general, it's ok to use it as long as you reformat it manually.

We primarily use IntelliJ IDEA which has its own version of editor for properties files and it does not result in such formatting. Personally, I like to edit localization files manually using text editor.

Destrolaric (approved):

LonwoLonwo:
I will approuve this PR after changes in localization files (PostgresResources.properties)

plotn:
> I will approuve this PR after changes in localization files (PostgresResources.properties)

done

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/PostgreSourceViewEditor.java
```diff
@@ -82,6 +98,32 @@ public void run() {
...
+                    public void run() {
+                        IWorkbenchWindow workbenchWindow = UIUtils.getActiveWorkbenchWindow();
+                        List<DBSProcedure> entities = new ArrayList<>();
+                        entities.add((DBSProcedure) sourceObject);
+                        DBRRunnableWithResult<String> generator = SQLEditorHandlerCheckProcedureConsole.checkGenerator(entities);
+                        UIUtils.runInUI(workbenchWindow, generator);
+                        String sql = CommonUtils.notEmpty(generator.getResult());
+                        SQLNavigatorContext navContext = new SQLNavigatorContext(sourceObject);
+                        String procName = ((DBSProcedure) sourceObject).getName();
+                        String title = procName + " " + PostgreMessages.procedure_check_label2;
```
ShadelessFox: I'm not a fan of concatenating localized strings. Please use `Check {0}` and `NLS.bind`.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/handlers/SQLEditorHandlerOpenObjectConsole.java
```diff
@@ -109,15 +109,35 @@ void openConsole(IWorkbenchWindow workbenchWindow, DBRRunnableWithResult<String>
...
+            doRun,
+            currentSelection,
+            sql, 
+            false);
+    }
+
+    /**
+     * openAndExecuteSQLScript with force sqlScript executing
+     */
+    public static void openAndExecuteSQLScriptExt(
```
ShadelessFox: Please just make it an overload, don't add `Ext` suffix

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/handlers/SQLEditorHandlerOpenObjectConsole.java
```diff
@@ -109,15 +109,35 @@ void openConsole(IWorkbenchWindow workbenchWindow, DBRRunnableWithResult<String>
...
         IWorkbenchWindow workbenchWindow,
         SQLNavigatorContext navigatorContext,
         String title,
         boolean doRun,
         ISelection currentSelection,
-        String sql) throws CoreException
-    {
+        String sql,
+        boolean forceScript) throws CoreException {
```
ShadelessFox: Clarify the parameter name and fix line wraps

```suggestion
        boolean forceProcessAsScript
    ) throws CoreException {
```

plugins/org.jkiss.dbeaver.ext.postgresql.ui/plugin.xml
```diff
@@ -310,6 +310,51 @@
...
+    <extension point="org.eclipse.ui.menus">       
+        <menuContribution allPopups="false" locationURI="popup:org.eclipse.ui.popup.any?after=generate">
+            <command commandId="org.jkiss.dbeaver.ext.postgresql.ui.editors.sql.handlers.plpgsql_check">
+                <visibleWhen checkEnabled="true"/>
+            </command>
+        </menuContribution>
+    </extension>
+    
+    <extension point="org.eclipse.ui.commands">
+        <command id="org.jkiss.dbeaver.ext.postgresql.ui.editors.sql.handlers.plpgsql_check" name="Check (via plpgsql_check)" description="Check function (via plpgsql_check)" categoryId="org.jkiss.dbeaver.ext.postgresql"/>
```
ShadelessFox: Having localization for `name` and `description` would be nice

```suggestion
        <command
            id="org.jkiss.dbeaver.ext.postgresql.ui.editors.sql.handlers.plpgsql_check"
            name="Check (via plpgsql_check)"
            description="Check function (via plpgsql_check)"
            categoryId="org.jkiss.dbeaver.ext.postgresql"/>
```

plugins/org.jkiss.dbeaver.ext.postgresql.ui/plugin.xml
```diff
@@ -310,6 +310,51 @@
...
+    </extension>
+    
+    <extension point="org.eclipse.ui.handlers">
+        <handler commandId="org.jkiss.dbeaver.ext.postgresql.ui.editors.sql.handlers.plpgsql_check" class="org.jkiss.dbeaver.ext.postgresql.ui.editors.sql.handlers.SQLEditorHandlerCheckProcedureConsole">
```
ShadelessFox:
```suggestion
        <handler
            commandId="org.jkiss.dbeaver.ext.postgresql.ui.editors.sql.handlers.plpgsql_check" 
            class="org.jkiss.dbeaver.ext.postgresql.ui.editors.sql.handlers.SQLEditorHandlerCheckProcedureConsole">
```

plugins/org.jkiss.dbeaver.ext.postgresql.ui/plugin.xml
```diff
@@ -310,6 +310,51 @@
...
+                        </adapt>
+                    </with>
+                </or>
+            </enabledWhen>            
+        </handler>
+    </extension>
+    
+    <!-- SQL Generators -->
+    <extension point="org.jkiss.dbeaver.sqlGenerator">
+        <generator id="procedureCheck" class="org.jkiss.dbeaver.ext.postgresql.model.sql.generator.SQLGeneratorProcedureCheck" label="CHECK (via plpgsql_check)" description="Check (via plpgsql_check)" order="2000">
```
ShadelessFox, line 316: Having localization for `label` and `description` would be nice

```suggestion
        <generator
        id="procedureCheck"
        class="org.jkiss.dbeaver.ext.postgresql.model.sql.generator.SQLGeneratorProcedureCheck"
        label="CHECK (via plpgsql_check)"
        description="Check (via plpgsql_check)" order="2000">
```

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/PostgreMessages.java
```diff
@@ -190,7 +190,18 @@ public class PostgreMessages extends NLS {
...
+    /* Procedure check, etc */
+    
+    public static String procedure_check_label;
+    public static String procedure_check_label2;
+    public static String procedure_check_label_ext;
+    public static String procedure_check_description;
+    public static String source_view_show_header_label;
+    public static String source_view_show_header_description;
+    public static String message_open_console;
+    public static String error_cant_open_sql_editor;
```
ShadelessFox, line 203: Please group and prefix localization strings according to their usage, e.g.:

```
...
action_procedure_check_label
action_procedure_check_description
...
dialog_procedure_check_title
dialog_procedure_check_title_with_object
...
```

See my another note regarding localization

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/model/sql/generator/SQLGeneratorProcedureCheck.java
```diff
@@ -0,0 +1,77 @@
...
+     * Generate PostgreSQL procedure check SQL - via https://github.com/okbob/plpgsql_check
+     */
+    public void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, 
+            Collection<? extends DBSProcedureParameter> parameters) {
+        List<DBSProcedureParameter> inParameters = new ArrayList<>();
+        if (parameters != null) {
+            inParameters.addAll(parameters);
+        }
+        sql.append("select * from plpgsql_check_function('" + proc.getFullyQualifiedName(DBPEvaluationContext.DML) + "(");
+        if (!inParameters.isEmpty()) {
```
ShadelessFox: Redundant check

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/model/sql/generator/SQLGeneratorProcedureCheck.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Generate PostgreSQL procedure check SQL - via https://github.com/okbob/plpgsql_check
+     */
+    public void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, 
+            Collection<? extends DBSProcedureParameter> parameters) {
+        List<DBSProcedureParameter> inParameters = new ArrayList<>();
+        if (parameters != null) {
+            inParameters.addAll(parameters);
+        }
```
ShadelessFox: Can be avoided by using `CommonUtils.safeList`

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/model/sql/generator/SQLGeneratorProcedureCheck.java
```diff
@@ -0,0 +1,77 @@
...
+        sql.append("select * from plpgsql_check_function('" + proc.getFullyQualifiedName(DBPEvaluationContext.DML) + "(");
+        if (!inParameters.isEmpty()) {
+            boolean first = true;
+            for (DBSProcedureParameter parameter : inParameters) {
+                switch (parameter.getParameterKind()) {
```
ShadelessFox: Redundant `switch` block, just use `if` with a check for `DBSProcedureParameter.IN`

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/model/sql/generator/SQLGeneratorProcedureCheck.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+public class SQLGeneratorProcedureCheck extends SQLGeneratorProcedure {
+
+    /**
+     * Generate PostgreSQL procedure check SQL - via https://github.com/okbob/plpgsql_check
+     */
+    public void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc,
```
ShadelessFox: This method is not called outside of this class. Consider making it `private` and moving it to the very bottom.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/sql/handlers/SQLEditorHandlerCheckProcedureConsole.java
```diff
@@ -0,0 +1,106 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class SQLEditorHandlerCheckProcedureConsole extends SQLEditorHandlerOpenObjectConsole {
+
+    private static final Log log = Log.getLog(SQLEditorHandlerCheckProcedureConsole.class);
+
+    public SQLEditorHandlerCheckProcedureConsole() {
+    }
```
ShadelessFox: Redundant constructor

ShadelessFox (changes requested):
Your PR has inconsistent use of `\n` and `getLineSeparator()`.

Consider using the latter to allow the `Compact SQL` option to work correctly.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/model/sql/generator/SQLGeneratorProcedureCheck.java
plotn: Can you write down an example? parameters arent List, but Collection

ShadelessFox: Then you can use `CommonUtils.safeCollection`.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/PostgresResources.properties
```diff
@@ -172,4 +172,13 @@ dialog_trigger_label_combo_event_type_tip = Choose trigger event type
 dialog_trigger_label_title = Create new Event Trigger
 
 dialog_trigger_edit_page_label_trigger_function = Trigger function
-dialog_trigger_edit_page_select_function_title = Select function for trigger
\ No newline at end of file
+dialog_trigger_edit_page_select_function_title = Select function for trigger
+
+procedure_check_label = Check
+procedure_check_label2 = "{0}" check
```
ShadelessFox: The `"Something" check` label looks weird to me. Wouldn't it be more natural if spelled `Check "Something"` instead?

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/model/sql/generator/SQLGeneratorProcedureCheck.java
```diff
@@ -0,0 +1,94 @@
...
+    @Override
+    protected void generateSQL(DBRProgressMonitor monitor, StringBuilder sql, DBSProcedure proc) throws DBException {
+        Collection<? extends DBSProcedureParameter> parameters = proc.getParameters(monitor);
+        generateStoredProcedureCall(sql, proc, CommonUtils.safeCollection(parameters));
+    }
+    
+    /**
+     * Generate PostgreSQL procedure check SQL - via https://github.com/okbob/plpgsql_check
+     */
+    private void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc,
```
ShadelessFox: What is even the purpose of this method? You can inline it to avoid unnecessary indirection in the `generateSQL`.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/sql/handlers/SQLEditorHandlerCheckProcedureConsole.java
```diff
@@ -0,0 +1,103 @@
...
+import org.jkiss.utils.CommonUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class SQLEditorHandlerCheckProcedureConsole extends SQLEditorHandlerOpenObjectConsole {
+
+    private static final Log log = Log.getLog(SQLEditorHandlerCheckProcedureConsole.class);
+    
+    void openConsoleCheck(IWorkbenchWindow workbenchWindow, DBRRunnableWithResult<String> generator,
```
ShadelessFox, line 47: This method can be inlined too.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/sql/handlers/SQLEditorHandlerCheckProcedureConsole.java
```diff
@@ -0,0 +1,103 @@
...
+        if (navContext == null || navContext.getDataSourceContainer() == null) {
+            log.debug("No active datasource");
+            return null;
+        }
+
+        DBRRunnableWithResult<String> generator = checkGenerator(entities);
+
+        String title = PostgreMessages.procedure_check_label_ext;
+        if (entities.size() == 1 && !CommonUtils.isEmpty(procName)) {
+            title = procName + " " + PostgreMessages.procedure_check_label2;
```
ShadelessFox: Please use `NLS.bind`

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/handlers/SQLEditorHandlerOpenObjectConsole.java
```diff
@@ -109,15 +109,40 @@ void openConsole(IWorkbenchWindow workbenchWindow, DBRRunnableWithResult<String>
...
+    /**
+     * openAndExecuteSQLScript in a SQL window
+     */
+    public static void openAndExecuteSQLScript(
+            IWorkbenchWindow workbenchWindow,
+            SQLNavigatorContext navigatorContext,
+            String title,
+            boolean doRun,
+            ISelection currentSelection,
+            String sql
```
ShadelessFox:
```suggestion
        IWorkbenchWindow workbenchWindow,
        SQLNavigatorContext navigatorContext,
        String title,
        boolean doRun,
        ISelection currentSelection,
        String sql
```

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/PostgresResources.properties
```diff
@@ -175,10 +175,10 @@ dialog_trigger_edit_page_label_trigger_function = Trigger function
 dialog_trigger_edit_page_select_function_title = Select function for trigger
 
 procedure_check_label = Check
-procedure_check_label2 = "{0}" check
+procedure_check_label2 = check "{0}"
```
ShadelessFox: Is there any particular reasons to say `check` instead of `Check`?

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/model/sql/generator/SQLGeneratorProcedureCheck.java
```diff
@@ -32,19 +32,11 @@
...
-    }
-    
-    /**
-     * Generate PostgreSQL procedure check SQL - via https://github.com/okbob/plpgsql_check
-     */
-    private void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, 
-            Collection<? extends DBSProcedureParameter> parameters) {
+        Collection<? extends DBSProcedureParameter> parameters = CommonUtils.safeCollection(proc.getParameters(monitor));
         List<DBSProcedureParameter> inParameters = new ArrayList<>();
         inParameters.addAll(CommonUtils.safeCollection(parameters));
```
ShadelessFox: You don't need these. Just inline `CommonUtils.safeCollection(proc.getParameters(monitor))` in the loop, like so:

```java
for (DBSProcedureParameter parameter : CommonUtils.safeCollection(proc.getParameters(monitor))) {
    ...
}
```

ShadelessFox (changes requested):

ShadelessFox (approved):

E1izabeth (approved):

LonwoLonwo (approved):

serge-rider:
@plotn I see that you added a new command, new command handler and menu contributions for plpgsql_check.
Generally it looks very similar to what we call Tools. For example take a look at PostgreToolAnalyze or PostgreToolVacuum classes.
There is no need in new commands etc, it will automatically appear in Tools menu (context and main).

WDYT?
Thanks for the contribution :)

plotn:
Hi! Firstly - i googled what does it mean - WDYT. Ok, I found :)
But I still cannot catch the idea with "Tools" - I called popup menu on the function - I see tools (see screenshot), all the items are not assigned with this function - they are only "general" (e.g. Backup, Session Manager, etc ...).
dbeaver_tools

P.S.: It was a very difficult challenge for me to do that :) I was to make code style conformity even in that modules, which I did not wrote - only added some lines, it is very hard to "external" developer, I mean - the code clearance (with current requirements) are performing by the developer, who did not write this code before

plotn:
See two user scenarios, please
dbeaver_check_way1
dbeaver_check_way2

serge-rider:
A tool can be assigned to several object types: tbles, procedures, schemas, etc.
Open menu on a table to see available table tools there.
We don't have PG procedure tools yet.

So you could add a first PG procedure-specific tool.
Use `<objectType name="org.jkiss.dbeaver.ext.postgresql.model.PostgreProcedure"/>`

plotn:
Lets confirm your suggestion:
1. We remove (pg_check) item from "Function submenu", "Generate SQL submenu";
2. We put them under the "tools" menu;
3. But they do directly the same as they do now (open SQL window and executes check script).

Did I understand correct?

PS: I think we should let "UPDATE FROM", "INSERT ON CONFLICT", "DELETE USING" under the "Generate SQL" as they are now, yes?

serge-rider:
The idea is to delete command, command handler and direct menu contributions. And add tool instead.
From UI perspective it will look like you described.

plotn:
Let us do the following:

1. I will delete command, command handler and direct menu contributions for the "Check (via plpgsql_check)" item;
2. "UPDATE FROM", "INSERT ON CONFLICT", "DELETE USING" will remain unchanged - they are at their right place I think;
3. Button "Check" in the function's sql edit window also stays as where it is now (near "Show header").

This will be current MR.

Further: I will add tools for other plpgsql_check functionality (check, profiler, etc.)

This will be future MRs. Is that ok?

serge-rider:
Yes, sounds liek a good plan 👍

plotn:
@serge-rider, see please

serge-rider:
Thank you 👍 
We will test it first. The plan is to include it in 22.2.2 release.

plotn:
Thank you! I plan to continue contribution if this case will be successful (as we discussed and more). Additionally - look at my "merge generator fixed: - it is quite simple and clear. There is another one generator broken with same problem (maybe insert - I dont remember, but it is - I will fix it further).

Matvey16:
Verified

PR #17507, opened 2022-08-30
#17399 Temporary falling back ConsoleView switch from checkbox to simple command contribution

### LINKED ISSUE #17399: Console view icon is missing on Linux
### Description

It's impossible to toggle on the console view since the icon is missing.

image

### DBeaver Version

Community 22.1.4

### Operating System

Pop!_OS 22.04

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:


Matvey16:
I don't think it works as expected now. Because the button is still togglable but it's unknown if it is toggled or not which is very confusing

dbeaver_YgKF2EYlCx

Matvey16:
Verified

Issue #15427, opened 2022-02-08
errors when reading large HANA lobs


kai-morich:
with default settings, lob behavior changes on 10k and 1MB
![image](https://user-images.githubusercontent.com/25850372/153037302-c4eb14af-5089-4265-9828-0a11529bd5f3.png)

this testcase 
```
DROP TABLE l1;
CREATE TABLE l1(i int, c nclob);
INSERT INTO l1 values (1, 'x');

UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1); 
SELECT LENGTH(c), * FROM l1;  -- 8kB, shown in result table cell
UPDATE l1 SET c = (SELECT c||c FROM l1);
SELECT LENGTH(c), * FROM l1;  -- 16kB, shown in value window
UPDATE l1 SET c = (SELECT c||c FROM l1);
SELECT LENGTH(c), * FROM l1;  -- 32kB, empty value window, no error
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
UPDATE l1 SET c = (SELECT c||c FROM l1);
SELECT LENGTH(c), * FROM l1;  -- 1MB, fails when opening value window: Error loading text value, IO error while copying content, SAP DBTech JDBC: [1033]: error while parsing protocol: invalid lob locator id (piecewise lob reading)
```
shows no value starting with 32KB and errors starting with 1MB
[...]

PR #17512, opened 2022-08-31
#15427 show warning message for exception on lob reading

LonwoLonwo:
if the context is in auto-commit mode

2022-08-31 14_25_01-DBeaver Ultimate 22 2 0 - MAPS


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/data/JDBCContentLOB.java
```diff
@@ -109,4 +114,15 @@ public DBDValueCloneable cloneValue(DBRProgressMonitor monitor)
 
     protected abstract JDBCContentLOB createNewContent();
 
+    void handleContentReadingException(DBCException e) throws DBCException {
+        DBCTransactionManager transactionManager = DBUtils.getTransactionManager(executionContext);
+        if (!errorMessageIsShown && transactionManager != null && transactionManager.isAutoCommit()) {
+            DBWorkbench.getPlatformUI().showWarningMessageBox(
+                ModelMessages.jdbc_content_view_error_message_title,
+                ModelMessages.jdbc_content_view_error_message_body);
+            errorMessageIsShown = true;
```
ShadelessFox: I don't like that model communicates with UI. This doesn't seem right.

It would be best if you handled exceptions in the UI. See the `ContentPanelEditor` class.

ShadelessFox (changes requested):

PR #17575, opened 2022-09-06
#16626 Fix remote host identifier changed dialog for jsch

### LINKED ISSUE #16626: SSH2 Message about remote host identification chage hides
#### System information: 
- archlinux current
- DBeaver version -  22.0.4.202205022238
- Additional extensions - NO

- Do you use tunnels or proxies (SSH, SOCKS, etc)? - YES

On dBeaver startup if script with connection using ssh is active dBever asks to confirm changed remote host identification. On clicking outside of the SSH message dialog or if other modal dialog (Tip of the day, message about new release etc.) is shown on startup SSH message hides beneath other windows but keeps holding input focus. In that case all GUI becomes locked and there is no way to get back to SSH message dialog to confirm or deny new remote host identification. I think if SSH dialog keeps holding input it must be shown on top and do not hide beneath other windows.

Open dBeaver with script using ssh tunneled connection with changed remote host identification. When SSH message is shown click any dBeaver GUI element outside this message.

E1izabeth:
_**How to check:**_

0. Run DBeaver
1. Choose the database and switch to JSCH in Advanced SSH settings
2. Open some script for the chosen database and enable setting to connect to the database on script opening
3. Turn on Tip of the day
4. Shutdown DBeaver
3. Change a couple of symbols of the key for the host of the chosen database in .ssh/known_hosts file
5. Run DBeaver
6. Wait for the Tip of the day and ssh message about changed host identifier. SSH Message will grab focus.
7. After clicking on the tip of the day modal window the SSH Message window should not be hidden.

### DIFF
--- a/plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/SSHUIMessages.properties
+++ b/plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/SSHUIMessages.properties
@@ -64,4 +64,4 @@ model_ssh_configurator_variables_hint_label = You can use variables in SSH param
 model_ssh_configurator_ssh_documentation_link = <a>SSH Documentation</a>
 
 jsch_remote_host_identifier_changed_warning_title = SSH Message
-}
+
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/SSHUIMessages.properties
```diff
@@ -62,3 +62,6 @@ model_ssh_configurator_group_jump_server_checkbox_label = Use jump server
 model_ssh_configurator_variables_hint_label = You can use variables in SSH parameters.
 
 model_ssh_configurator_ssh_documentation_link = <a>SSH Documentation</a>
+
+jsch_remote_host_identifier_changed_warning_title = SSH Message
+}
```
ShadelessFox:
```suggestion
```

ShadelessFox (changes requested):

ShadelessFox (approved):

Matvey16:
Verified

PR #17581, opened 2022-09-07
#15427 show warning hint message for exceptions on lob reading

### LINKED ISSUE #13367: Derby: Click on the BLOB in data grid throws an error
#### System information: 
- Win10, DBeaver CE 21.1.4

- Derby server
- Driver version 10.15.1.3

Any attempt to open the blob results in error. I tried changing the commit modes, but the problem persists
image

```
ERROR XJ215: You cannot invoke other java.sql.Clob/java.sql.Blob methods after calling the free() method or after the Blob/Clob's transaction has been committed or rolled back.
	at org.apache.derby.client.am.Lob.checkValidity(Lob.java:392)
	at org.apache.derby.client.am.ClientBlob.length(ClientBlob.java:119)
	at org.jkiss.dbeaver.model.impl.jdbc.data.JDBCContentBLOB.getLOBLength(JDBCContentBLOB.java:65)
	at org.jkiss.dbeaver.model.impl.jdbc.data.JDBCContentLOB.getContentLength(JDBCContentLOB.java:48)
	at org.jkiss.dbeaver.model.impl.jdbc.data.JDBCContentBLOB.getContents(JDBCContentBLOB.java:85)
	at org.jkiss.dbeaver.ui.data.BinaryPanelEditor.primeEditorValue(BinaryPanelEditor.java:80)
	at org.jkiss.dbeaver.ui.data.BinaryPanelEditor.primeEditorValue(BinaryPanelEditor.java:1)
	at org.jkiss.dbeaver.ui.data.editors.ContentPanelEditor$StreamValueLoadService.evaluate(ContentPanelEditor.java:497)
	at org.jkiss.dbeaver.ui.data.editors.ContentPanelEditor$StreamValueLoadService.evaluate(ContentPanelEditor.java:1)
	at org.jkiss.dbeaver.ui.LoadingJob.run(LoadingJob.java:88)
	at org.jkiss.dbeaver.ui.LoadingJob.run(LoadingJob.java:72)
```

### LINKED ISSUE #15427: errors when reading large HANA lobs

LonwoLonwo:
2022-09-07 10_50_13-DBeaver Ultimate 22 2 0 - _HANA_Script-173

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/data/JDBCContentBLOB.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/data/JDBCContentBLOB.java
@@ -83,42 +83,46 @@ public class JDBCContentBLOB extends JDBCContentLOB {
         throws DBCException
     {
         if (storage == null && blob != null) {
-            long contentLength = getContentLength();
-            DBPPlatform platform = DBWorkbench.getPlatform();
-            if (contentLength < platform.getPreferenceStore().getInt(ModelPreferences.MEMORY_CONTENT_MAX_SIZE)) {
-                try {
-                    try (InputStream bs = blob.getBinaryStream()) {
-                        storage = BytesContentStorage.createFromStream(
-                            bs,
-                            contentLength,
-                            getDefaultEncoding());
+            try {
+                long contentLength = getContentLength();
+                DBPPlatform platform = DBWorkbench.getPlatform();
+                if (contentLength < platform.getPreferenceStore().getInt(ModelPreferences.MEMORY_CONTENT_MAX_SIZE)) {
+                    try {
+                        try (InputStream bs = blob.getBinaryStream()) {
+                            storage = BytesContentStorage.createFromStream(
+                                bs,
+                                contentLength,
+                                getDefaultEncoding());
+                        }
+                    } catch (IOException e) {
+                        throw new DBCException("IO error while reading content", e);
+                    } catch (Throwable e) {
+                        throw new DBCException(e, executionContext);
                     }
-                } catch (IOException e) {
-                    throw new DBCException("IO error while reading content", e);
-                } catch (Throwable e) {
-                    throw new DBCException(e, executionContext);
-                }
-            } else {
-                // Create new local storage
-                File tempFile;
-                try {
-                    tempFile = ContentUtils.createTempContentFile(monitor, platform, "blob" + blob.hashCode());
-                }
-                catch (IOException e) {
-                    throw new DBCException("Can't create temporary file", e);
-                }
-                try (OutputStream os = new FileOutputStream(tempFile)) {
-                    try (InputStream bs = blob.getBinaryStream()) {
-                        ContentUtils.copyStreams(bs, contentLength, os, monitor);
+                } else {
+                    // Create new local storage
+                    File tempFile;
+                    try {
+                        tempFile = ContentUtils.createTempContentFile(monitor, platform, "blob" + blob.hashCode());
                     }
-                } catch (IOException e) {
-                    ContentUtils.deleteTempFile(tempFile);
-                    throw new DBCException("IO error while copying stream", e);
-                } catch (Throwable e) {
-                    ContentUtils.deleteTempFile(tempFile);
-                    throw new DBCException(e, executionContext);
+                    catch (IOException e) {
+                        throw new DBCException("Can't create temporary file", e);
+                    }
+                    try (OutputStream os = new FileOutputStream(tempFile)) {
+                        try (InputStream bs = blob.getBinaryStream()) {
+                            ContentUtils.copyStreams(bs, contentLength, os, monitor);
+                        }
+                    } catch (IOException e) {
+                        ContentUtils.deleteTempFile(tempFile);
+                        throw new DBCException("IO error while copying stream", e);
+                    } catch (Throwable e) {
+                        ContentUtils.deleteTempFile(tempFile);
+                        throw new DBCException(e, executionContext);
+                    }
+                    this.storage = new TemporaryContentStorage(platform, tempFile, getDefaultEncoding(), true);
                 }
-                this.storage = new TemporaryContentStorage(platform, tempFile, getDefaultEncoding(), true);
+            } catch (DBCException e) {
+                handleContentReadingException(e);
             }
             // Free blob - we don't need it anymore
             releaseBlob();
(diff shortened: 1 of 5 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/data/JDBCContentLOB.java
```diff
@@ -32,6 +37,8 @@
  */
 public abstract class JDBCContentLOB extends JDBCContentAbstract implements DBDContent {
 
+    private static final String ERROR_MESSAGE_PARAM_IS_SHOWN = "error-message-is-shown";
```
E1izabeth: As it's a common execution context, I assume this parameter should be called like "lob-reading-error-message-is-shown".

By the way, I'm not sure, but is it possible to store this property like a field? Maybe there is no need to store it in execution context.

ShadelessFox: Do we really want to bloat lob class with such a specific field? Execution context seems to be the most appropriate place to store such an information. Moreover, by doing so, the message will be shown only once per session rather than every time you open a lob.

E1izabeth: I think that maybe execution context is a too common place. But I just asked for a reason. I thought that JDBCContentLOB is created not every time you open a lob.

E1izabeth: Anyway the parameter should not be called just "error-message-is-shown", because it's not clear what message this parameter about.

ShadelessFox: I believe they have discussed it before prior to making this decision.

I agree that the parameter should be renamed.

LonwoLonwo: Yep-yep, just forgotten

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/data/JDBCContentBLOB.java
```diff
@@ -83,42 +83,46 @@ public DBDContentStorage getContents(DBRProgressMonitor monitor)
...
-                try (OutputStream os = new FileOutputStream(tempFile)) {
-                    try (InputStream bs = blob.getBinaryStream()) {
-                        ContentUtils.copyStreams(bs, contentLength, os, monitor);
+                } else {
+                    // Create new local storage
+                    File tempFile;
+                    try {
+                        tempFile = ContentUtils.createTempContentFile(monitor, platform, "blob" + blob.hashCode());
+                    }
+                    catch (IOException e) {
```
ShadelessFox:
```suggestion
                    } catch (IOException e) {

```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/data/JDBCContentCLOB.java
```diff
@@ -84,41 +84,45 @@ public DBDContentStorage getContents(DBRProgressMonitor monitor)
...
+                            throw new DBCException(e, executionContext);
+                        }
+                    }
+                } else {
+                    // Create new local storage
+                    File tempFile;
+                    try {
+                        tempFile = ContentUtils.createTempContentFile(monitor, platform, "clob" + clob.hashCode());
+                    }
+                    catch (IOException e) {
```
ShadelessFox:
```suggestion
                    } catch (IOException e) {

```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/data/JDBCContentBLOB.java
LonwoLonwo: :face_with_rolling_eyes:

ShadelessFox (approved):

E1izabeth (approved):

Matvey16:
The message appears only once for the current editor wich will be confusing if the user accidentaly closes it before reading.

dbeaver_Sr7kXun2aI

Matvey16:
Verified

PR #17618, opened 2022-09-12
#16545 Connect by custom URL for Generic, PostgreSQL, SQL Server

### LINKED ISSUE #16545: Multi-protocol support for ClickHouse
Below is the dialog popped up when creating a connection to ClickHouse. It's all good except I cannot change URL to something like `jdbc:ch:tcp://localhost` directly, so I have to create the connection first, and then manually change URL by editing driver settings.
image

1. add a drop-down list for user to choose `Protocol`(one of grpc, http and tcp, defaults to http)
2. make `Port` optional because each protocol has its default port, for examples: 8123 for http, 9000 for tcp and 9100 for grpc
3. update URL template to something like `jdbc:(ch|clickhouse)[:{protocol}]://{host}[:{port}][/{database}]`

Make `JDBC URL` field editable in the popup.

**Additional context**
This request has nothing to do with ClickHouse(legacy) driver as it only supports http.

### LINKED ISSUE #17380: DBeaver does not allow anything after "jdbc:sqlserver://" in MS SQL Server URL template, breaking sqlexpress and localdb connections
### Description

In DBeaver 22.1.1, the SQL Server URL template is something like

`jdbc:sqlserver://;serverName=localhost;databaseName=testdb`

For a lot of people, this isn't a problem, because they can just use the `serverName` parameter and port to connect via TCP/IP.
However, this breaks support for **sqlexpress** and **localdb** connections, which **cannot** be connected to with the `serverName` parameter. For these connections, the host must be inserted right after the `jdbc:sqlserver://` bit. For example, a **sqlexpress** connection string looks like this:

`jdbc:sqlserver://localhost\sqlexpress;databaseName=testdb;integratedSecurity=true;`

Because DBeaver does not allow inserting anything after the `jdbc:sqlserver://` bit, **sqlexpress** and **localdb** connections don't work.

Please see this StackOverflow answer for a little more info: https://stackoverflow.com/a/40830281

This is also related to (but different than) https://github.com/dbeaver/dbeaver/issues/4243

This should be a very easy fix anyway.

### DBeaver Version

Community Edition 22.1.1

### Operating System

Windows 10 Enterprise

### Database and driver

MS SQL Express & localdb (e.g. `localhost\sqlexpress` or `localdb\testdb`)
com.microsoft.sqlserver:mssql-jdbc:RELEASE [9.2.0.jre8]
com.microsoft.sqlserver:mssql-jdbc_auth:RELEASE [9.2.0.x64]

E1izabeth:
Also closes #17380

image

image

You need to explicitly configure ssh port mapping if ssh tunnel required (for example, like in the following pic).

image


plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageAbstract.java
```diff
@@ -283,5 +290,23 @@ protected Image createImage(String imageFilePath) {
...
+            control.setEnabled(!useUrl);
+            if (control instanceof Text) {
+                ((Text) control).setEditable(!useUrl);
+            }
+        }
+    }
+
+    protected void addControlToGroup(String group, Control control) {
+        List<Control> controlList = propGroupMap.computeIfAbsent(group, k -> new ArrayList<>());
+        controlList.add(control);
```
ShadelessFox:
```suggestion
        propGroupMap
            .computeIfAbsent(group, k -> new ArrayList<>())
            .add(control);
```

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageAbstract.java
```diff
@@ -283,5 +290,23 @@ protected Image createImage(String imageFilePath) {
...
+
+        for (Control control : propGroupMap.get(GROUP_CONNECTION)) {
+            control.setEnabled(!useUrl);
+            if (control instanceof Text) {
+                ((Text) control).setEditable(!useUrl);
+            }
+        }
+    }
+
+    protected void addControlToGroup(String group, Control control) {
```
ShadelessFox: Please add nullability annotations

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageAbstract.java
```diff
@@ -283,5 +290,23 @@ protected Image createImage(String imageFilePath) {
         ImageDescriptor imageDescriptor = ResourceLocator.imageDescriptorFromBundle(getClass(), imageFilePath).orElse(null);
         return imageDescriptor == null ? null : imageDescriptor.createImage();
     }
+    
+    protected void setupConnectionModeSelection(Button typeURLRadio, Button typeManualRadio, Text urlText, boolean useUrl) {
```
ShadelessFox: Please add nullability annotations

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageAbstract.java
```diff
@@ -39,12 +39,19 @@
...
+import java.util.List;
+import java.util.Map;
+
 /**
  * ConnectionPageAbstract
  */
+public abstract class ConnectionPageAbstract extends DialogPage implements IDataSourceConnectionEditor {
+
+    protected static final String GROUP_CONNECTION = "connection"; //$NON-NLS-1$
+    protected Map<String, List<Control>> propGroupMap = new HashMap<>();
```
ShadelessFox: Can be `final`

ShadelessFox (dismissed):

uslss:
verified

PR #17637, opened 2022-09-13
fix german translation string

bulldog98:


E1izabeth (approved):

E1izabeth:
Thank you for contribution.

PR #17645, opened 2022-09-14
#17479 add datetime64 support

### LINKED ISSUE #17479: [Clickhouse] Datetime64/Datetime support
Currently, we have one datatype for handling timestamp values. As we found out in the scope of https://github.com/dbeaver/dbeaver/issues/17362 Clickhouse has two datatypes for timestamps. Datetime doesn't support milliseconds and has a lesser range of dates but Datetime64 supports them. We need to be able to handle them. We need to be able to handle date formatting issues that arise when we use Clickhouse.

Destrolaric:
Closes #17479


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/formatters/DateTimeDataFormatter.java
```diff
@@ -79,11 +81,22 @@ public DateFormat getDateFormat() {
...
+    @Override
+    public void setPattern(@NotNull String pattern) {
+        this.pattern = pattern;
+        String sdfPattern = pattern.replace("n", "f");
+        dateFormat = new ExtendedDateFormat(sdfPattern, locale);
+        dateFormat.setLenient(false);
+        buffer = new StringBuffer();
+        position = new FieldPosition(0);
+        String java8DatePattern = pattern.replaceAll("f+", "n");
+        dateTimeFormatter = DateTimeFormatter.ofPattern(java8DatePattern);
```
E1izabeth: It looks like a copy paste from `init` method. It's better to use `init` instead of creating such a strange "setter".

E1izabeth (changes requested):

E1izabeth:
I don't think that we need `setPattern` at all.

ShadelessFox (changes requested):
By manually setting the pattern, you're effectively ignoring one set by the user. This is bad and we don't want to do that.

Just let the user configure the pattern of their choice.

ShadelessFox (approved):
This is better.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.clickhouse/plugin.xml
```diff
@@ -35,7 +35,7 @@
                     <replace provider="generic" driver="yandex_clickhouse"/>
                     <file type="jar" path="maven:/ru.yandex.clickhouse:clickhouse-jdbc:RELEASE[0.2.6]" bundle="!drivers.clickhouse"/>
                     <file type="jar" path="drivers/clickhouse" bundle="drivers.clickhouse"/>
-
+                    <property name="@dbeaver-default-dataformat.type.timestamp.pattern" value="yyyy-MM-dd HH:mm:ss[.fffffffff]"/>
```
ShadelessFox, line 38: Please bring back line separator

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.clickhouse/plugin.xml
ShadelessFox: Whatever

ShadelessFox (approved):

Matvey16:
Any timestamp column is now set to NULL if you try to add or change milliseconds

dbeaver_wHs6lb79CJ

bundles/org.jkiss.utils/src/org/jkiss/utils/time/ExtendedDateFormat.java
```diff
@@ -146,6 +146,7 @@ public Date parse(@NotNull String text, @NotNull ParsePosition pos)
         int index = pos.getIndex();
         if (index < text.length() && nanoStart > 0) {
             long nanos = 0;
+            index += nanoPrefix.length();
```
E1izabeth: `nanoPrefix` can be null as I see

Destrolaric: fixed

bundles/org.jkiss.utils/src/org/jkiss/utils/time/ExtendedDateFormat.java
```diff
@@ -159,8 +162,7 @@ public Date parse(@NotNull String text, @NotNull ParsePosition pos)
                     return null;
                 }
                 long digit = ((int)c - (int)'0');
-                for (int k = MAX_NANO_LENGTH - i; k > 0; k--) {
-                    digit *= 10;
+                for (int k = MAX_NANO_LENGTH - i; k > 0; k--) {digit *= 10;
```
ShadelessFox: ?

ShadelessFox (dismissed):

bundles/org.jkiss.utils/src/org/jkiss/utils/time/ExtendedDateFormat.java
Destrolaric: accidentally formatted, fixed

E1izabeth (approved):

Matvey16:
Verified

PR #17681, opened 2022-09-19
dbeaver/dbeaver-ee#1907 supports shared tasks configurations; 

LonwoLonwo:
RMProjects.Type enum moved to the single class


E1izabeth:
What's about hiding commands in the UI if user doesn't have the proper permissions?

LonwoLonwo:
@E1izabeth Yeah, you are right. 
Still in progress for this reason.

PR #17683, opened 2022-09-19
dbeaver/dbeaver-ee#1933 Speed up generic container refresh

ShadelessFox:
This is achieved by caching its structure afterward.

Please ensure schema refresh works nicely in generic databases, such as Vertica.


Matvey16:
Verified

PR #17685, opened 2022-09-19
#15860 search by attributes case insensitive for tasks columns names

### LINKED ISSUE #15860: Using a task to import a csv file fails if columns are case sensitive
#### System information: 
- Linux Mint 20.3 Cinnamon
- DBeaver 22.0.0.202203060510 Community Edition

- PostgreSQL 13
- Drivers: PostgreSQL JDBC Driver

I have an existing table with those columns:
`id: text`
`organizationKey: text`

I can't import a CSV file **using a task** because columns are case sensitive. `organizationKey` has a non null clause, and since DBeaver ignores case-sensitive columns, it creates a new `organizationkey` column instead of mapping to the right one.

**csv file:**
```csv
"id","organizationKey"
edlund,"680456df-8330-4c09-8cfe-33daf0bd3f0f"
```
**Note that doing the exact same steps manually work**. If I save this as a task and run it, it fails for the reason mentioned above.

- Create a db with 1 case-sensitive column, eg `Key`
- Create a csv file
- Start the steps to import it, save a task
- Run the task: fails.

```
SQL Error [23502]: Batch entry 0 INSERT INTO public."processModel" (id,organizationkey)
	VALUES ('edlund','680456df-8330-4c09-8cfe-33daf0bd3f0f') was aborted: ERROR: null value in column "organizationKey" of relation "processModel" violates not-null constraint
```

LonwoLonwo:
2022-09-20 00_47_49-Data Transfer -  import_qualified_table_column


uslss:
verified

PR #17710, opened 2022-09-23
#6528 Add dialect customization 

### LINKED ISSUE #6528: How to add datatype to keyword list?
#### System information: 
- Operating system (distribution) and version Windows 10 1903
- DBeaver version 6.1.1

- Database name and version

#### Describe your question:
When creating a new column of type UUID it isn't highlighted.  Is this something I can add to so that it highlights properly?

Destrolaric:
Now dialects keywords are set via configuration and by UI. Now we can add additional keywords or remove existing ones from the dialect. All keywords are inherited by default.
javaw_0hwKOlnUDL

Closes #6528


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/dialects/SQLDialectDescriptorSerializer.java
```diff
@@ -0,0 +1,148 @@
...
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.StringReader;
+import java.util.*;
+
+public class SQLDialectDescriptorSerializer {
+    private static final Log log = Log.getLog(SQLDialectDescriptorSerializer.class);
+
+
+    public static final String ATTR_DIALECTS_VALUE_KEYWORDS = "keywords"; //$NON-NLS-1$
```
LonwoLonwo: Please create attributes as constants in the commonplace to use in the SQLDialectDescriptor

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/dialects/SQLDialectDescriptor.java
```diff
@@ -0,0 +1,465 @@
...
+import org.jkiss.utils.BeanUtils;
+import org.jkiss.utils.CommonUtils;
+
+import javax.management.ReflectionException;
+import java.lang.reflect.Field;
+import java.util.*;
+
+/**
+ * SQLDialectDescriptor
+ */
```
E1izabeth: There is no point in having such java docs

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/dialects/SQLDialectDescriptor.java
```diff
@@ -0,0 +1,465 @@
...
+                break;
+            case "types":
+                this.types = values;
+                break;
+            case "functions":
+                this.functions = values;
+                break;
+            case "insertMethods":
+                insertMethodNames = new ArrayList<>(values);
+                break;
```
E1izabeth: Why don't you use created constants here?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialectDescriptor.java
```diff
@@ -84,25 +90,25 @@ public class SQLDialectDescriptor extends AbstractContextDescriptor implements S
...
                 case "types":
-                    this.types = loadList(propValue);
+                    this.types = loadSet(propValue);
                     break;
                 case "functions":
-                    this.functions = loadList(propValue);
+                    this.functions = loadSet(propValue);
                     break;
                 case "insertMethods":
                     insertMethodNames = loadList(propValue);
```
E1izabeth: You can use constants here too

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/dialects/SQLDialectDescriptor.java
```diff
@@ -0,0 +1,465 @@
...
+import org.jkiss.utils.CommonUtils;
+
+import javax.management.ReflectionException;
+import java.lang.reflect.Field;
+import java.util.*;
+
+/**
+ * SQLDialectDescriptor
+ */
+public class SQLDialectDescriptor extends AbstractContextDescriptor implements SQLDialectMetadata {
```
E1izabeth: Now we have two classes with the same name and similar methods. What is the purpose of having both of them?

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/preferences/PrefPageSQLDialects.java
```diff
@@ -164,44 +164,90 @@ private void createDialectItem(Tree dialectTable, TreeItem parentItem, SQLDialec
...
+            reservedWordsText.getText().isEmpty()
+            && dataTypesText.getText().isEmpty()
+            && functionNamesText.getText().isEmpty()
+            && ddlKeywordsText.getText().isEmpty()
+            && dmlKeywordsText.getText().isEmpty()
+            && executeKeywordsText.getText().isEmpty()
+            && transactionKeywordsText.getText().isEmpty()
+            && statementDelimiterText.getText().isEmpty()
+             );
+    }
```
E1izabeth: Does it mean that user can't see any existing keywords in the settings?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/dialects/SQLDialectDescriptor.java
Destrolaric: Mistake after refactoring, old one now removed

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialectDescriptor.java
Destrolaric: done

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/dialects/SQLDialectDescriptor.java
```diff
@@ -0,0 +1,518 @@
...
+    private String label;
+    private String description;
+    private ObjectType implClass;
+    private DBPImage icon;
+    private boolean isAbstract;
+    private boolean isHidden;
+    private SQLDialectDescriptor parentDialect;
+    private Set<SQLDialectDescriptor> subDialects = null;
+
+    public enum WordType {
```
E1izabeth: Maybe KeywordType?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/dialects/SQLDialectDescriptor.java
```diff
@@ -0,0 +1,518 @@
...
+        ATTR_DIALECTS_VALUE_SCRIPT_SEPARATOR("scriptSeparator"),
+        ATTR_DIALECTS_VALUE_FUNCTIONS("functions"), //$NON-NLS-1$
+        ATTR_DIALECTS_VALUE_TYPES("types"); //$NON-NLS-1$
+
+        @Nullable
+        public static WordType getByTypeName(@NotNull String typename) {
+            Optional<WordType> first = Arrays.stream(values())
+                .filter(it -> it.getTypeName().equals(typename))
+                .findFirst();
+            return first.orElse(null);
```
E1izabeth:
```suggestion
            return Arrays.stream(values())
                .filter(it -> it.getTypeName().equals(typename))
                .findFirst();
                .orElse(null);
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/dialects/SQLDialectDescriptor.java
```diff
@@ -0,0 +1,518 @@
...
+                log.debug("Can't get SQL insert replace methods");
+            }
+        }
+
+        return insertMethodDescriptors;
+    }
+
+    public void setScriptDelimiter(String scriptDelimiter) {
+        this.scriptDelimiter = scriptDelimiter;
+    }
```
E1izabeth: It is not used.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/dialects/SQLDialectDescriptor.java
```diff
@@ -0,0 +1,518 @@
...
+        if (parentDialect != null) {
+            nestedTXNKeywords.addAll(parentDialect.getTransactionKeywords(true));
+        }
+        return getTransformedValues(WordType.ATTR_DIALECTS_TXN_KEYWORDS, nestedTXNKeywords);
+    }
+
+    @Override
+    @NotNull
+    public String getScriptDelimiter() {
+        return CommonUtils.isEmpty(scriptDelimiter) ? ";" : scriptDelimiter;
```
E1izabeth: It doesn't seem to be used somewhere. Also I don't like the idea of duplicating the setting about statements delimiter.

plugins/org.jkiss.dbeaver.ext.postgresql/plugin.xml
```diff
@@ -638,11 +638,34 @@
...
+            POSTGRE_FUNCTIONS_MATH,
+            POSTGRE_FUNCTIONS_NETWROK,
+            POSTGRE_FUNCTIONS_RANGE,
+            POSTGRE_FUNCTIONS_SEQUENCE,
+            POSTGRE_FUNCTIONS_SRF,
+            POSTGRE_FUNCTIONS_STRING,
+            POSTGRE_FUNCTIONS_TEXT_SEARCH,
+            POSTGRE_FUNCTIONS_TRIGGER,
+            POSTGRE_FUNCTIONS_WINDOW,
+            POSTGRE_FUNCTIONS_XML"/>
```
ShadelessFox: It's not valid XML.

Destrolaric: Well, it parsed correctly, any idea how it could be made better?

ShadelessFox: You could use `CDATA`: https://en.wikipedia.org/wiki/CDATA, or keep it in a long single line to avoid making the `plugin.xml` REALLY long.

Destrolaric: Well, I will look how it can be improved, maybe we can use annotations in the class itself, so we can avoid defining a lot of keywords in XML.

PR #17735, opened 2022-09-26
#17591 Add preference for Windows to use or not Windows trust store

### LINKED ISSUE #17591: Can't download driver files due to certificate issues
### Description

I try to connect to database, but I don't have downloaded driver files yet.
Download driver files window appears, then I get the following exception.

image

After some investigation we found out that removing -Djavax.net.ssl.trustStoreType=WINDOWS-ROOT from dbeaver.ini helps. 
https://repo1.maven.org/ is successfully opening with web browser. Windows update didn't help.

### Proposed solution

Maybe the parameter javax.net.ssl.trustStoreType should be dynamic. Also we could create preference for this.

### DBeaver Version

DBeaver 22.2.0

### Operating System

Windows 10

E1izabeth:
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
+++ b/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
@@ -37,6 +37,7 @@ import org.jkiss.code.NotNull;
 import org.jkiss.code.Nullable;
 import org.jkiss.dbeaver.DBeaverPreferences;
 import org.jkiss.dbeaver.Log;
+import org.jkiss.dbeaver.ModelPreferences;
 import org.jkiss.dbeaver.core.DBeaverActivator;
 import org.jkiss.dbeaver.model.DBConstants;
 import org.jkiss.dbeaver.model.app.DBPApplicationController;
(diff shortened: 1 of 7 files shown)


plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -274,6 +278,12 @@ public Object start(IApplicationContext context) {
         }
         TimezoneRegistry.overrideTimezone();
 
+        if (RuntimeUtils.isWindows() &&
+            ModelPreferences.getPreferences().getBoolean(ModelPreferences.PROP_USE_WIN_TRUST_STORE_TYPE)
+        ) {
```
ShadelessFox: It fits in a single line, let's not wrap it

```suggestion
        if (RuntimeUtils.isWindows() && ModelPreferences.getPreferences().getBoolean(ModelPreferences.PROP_USE_WIN_TRUST_STORE_TYPE)) {
```

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -97,6 +98,9 @@ public class DBeaverApplication extends DesktopApplicationImpl implements DBPApp
     private static final String PROP_EXIT_DATA = IApplicationContext.EXIT_DATA_PROPERTY; //$NON-NLS-1$
     private static final String PROP_EXIT_CODE = "eclipse.exitcode"; //$NON-NLS-1$
 
+    private static final String PROP_TRUST_STORE_TYPE = "javax.net.ssl.trustStoreType"; //$NON-NLS-1$
+    private static final String WIN_TRUST_STORE = "WINDOWS-ROOT"; //$NON-NLS-1$
```
ShadelessFox: Perhaps we need to introduce a naming convention for such key-value pairs.

Taking the convention from the `java.awt.RenderingHints`:

```suggestion
    private static final String VALUE_TRUST_STRORE_TYPE_WINDOWS = "WINDOWS-ROOT"; //$NON-NLS-1$
```

ShadelessFox (approved):

Matvey16:
Verified

PR #17741, opened 2022-09-27
dbeaver/dbeaver-ee#1956 HSQL plain execution plan

ShadelessFox:
The proposed solution just executes the `EXPLAN PLAN FOR <query>` statement and shows the result set:

image

Unfortunately, the output is not consistent, making it really hard for parsing and further processing:

 - It contains extra closing brackets in several places
 - Object names are quoted in several places only


uslss:
verified

PR #17742, opened 2022-09-27
#17061 Allow hiding columns whose values within loaded rows are NULL

### LINKED ISSUE #17061: Hiding a column when all the rows matching this columns are empty or NULL
Hi,

I would like to know if there is a way to hide **dynamically** empty columns (all rows are empty on this column).
It is very time consuming searching for empty columns every query and display only usefull infos.

image

Thanks in advance for your answer.

ShadelessFox:
image


LonwoLonwo (approved):

E1izabeth (approved):

uslss:
Column is hide even if there is data in a row that is not fetched for now
dbeaver_sJDGdxlFTt

ShadelessFox:
This is an expected behavior. We don't want to reset visibility after fetching new data

uslss:
verified

deft01:
Its amazing thank you so much. Great job

PR #17745, opened 2022-09-27
#5529 Show column description in the record mode

### LINKED ISSUE #5529: Add ability to show metadata column information in records view
Currently using DBeaver 6. 

It would be a nice feature enhancement to have the ability to show column metadata in record view.

Here is what it looks like in grid view:

image

Would like ability to turn on/off metadata in records view (separate option from grid view column metadata view since some people will probably like the condensed view better).

image

I am thinking that it could simply be that the metadata is in another column to the right, or appended onto the Column name (such as [column name] - [column metadata])

ShadelessFox:
java_CPuHXuZDPI


uslss:
verified

PR #17765, opened 2022-09-28
#17757 Driver config: save deleted libraries info

### LINKED ISSUE #17757: Download driver files dialog when I reopen DBeaver appears everytime
### Description

When I reopen DBeaver it always opens "Download driver files" error. After I manually delete offered driver in Driver Manager then connection works fine. But reopening DBeaver and problem is back.

This problem first appeared in 22.2.1. Up till 22.2.0 this was not a problem. It looks like regression problem.

### DBeaver Version

Community Edition 22.2.1

### Operating System

Windows 10

### Database and driver

1. DATABASE name AND version: Db2 v11.5.7.0
2. Driver name AND version: Db2 JDBC Driver 4.31.10

### Steps to reproduce

1. Open DBeaver and click on connection from Database Navigator window and double click connection to connect to database. Error appears: "Download driver files". I expect to get no error. This error is most probably related to the JDBC driver I manually change. To solve this problem I do:
2. Database | Driver Manager.
3. DB2 | DB2 LUW and Edit button.
4. On Libraries tab there are two drivers. First line is default JDBC driver which is few years old. And second is the latest JDBC driver I have manually added few months ago.
image
5. I mark first line and click on Delete button and confirm with Yes button. Driver is deleted.
6. I click on OK button and then on Close button, to close down dialogs.
7. Double click on connection and connection is successful.

Now to problem reappear, close down DBeaver and start it again and you are at step 1 again. Reappearing problem...

This looks like a regression in DBeaver 22.2.1, beucase in DBeaver 22.2.0 and older this problem was not happening.

### Additional context

By the way I have already reported problem that there is old Db2 JDBC driver link in DBeaver at [issue 17427]( https://github.com/dbeaver/dbeaver/issues/17427) and like I see it, this problem is already fixed in 22.2.2 currently in development.

When new version DBeaver 22.2.2 is released, this problem may be hidden, because I will just download the latest driver and probably new driver Db2 will not be available for few months/years. Problem will go away, but hidden until new driver will manually be selected.

serge-rider:


uslss:
Cannot re-download default driver after it was manually removed
dbeaver_tn2s8bT4Po

serge-rider:
Fixed. also zip files download fixed (e.g. Athena driver)

uslss:
1. Driver isn't proposed to download automatically in case it was manually removed before
dbeaver_fetQQRZ6rD
2. Reset do default button isn't working. The last chosen but not the default driver version is proposed
dbeaver_Bg45xf9R4I

PR #17773, opened 2022-09-29
#16049 Allow opening target table in a new tab

### LINKED ISSUE #16049: No shortcut to open a table from the References panel
When following references between multiple tables, the _References_ panel comes useful to go from table A to table B. However, if table C has a foreign key refering to table B, it is then inconvenient to get to C.

I see no way to do that with DBeaver 22.0.1.

A simple way would be to allow opening the References panel in its own full tab, so that we can then use References again.

Perhaps a more elaborate solution like Toad's Master-Detail Browser would be even better.

ShadelessFox:
image


E1izabeth (approved):

Matvey16:
Verified

Issue #17763, opened 2022-09-28
Statistics tab always becomes active when running a SQL script


jyinpci:
### Description

This is related to a previously closed ticket  https://github.com/dbeaver/dbeaver/issues/7704.

The issue had resurfaced in the past 3 releases, still happening in the current 22.2.1 release.  Whenever one executes a script having a variable (or more than SQL statements), the Statistics tab will be displayed first.  One has to click on the Result tab to actually see the result.  And to make it worse, if one has to scroll down in the data editor and fetch more rows, it would jump back to the Statistics tab every time more rows are retrieved, very annoying.  Would appreciate it greatly if this can be fixed soon.  Thanks!

### DBeaver Version

Community edition 22.2.1 (and previous 2 releases)

### Operating System

Windows 10 Pro.

### Database and driver

Oracle ojdbc8 and orai18n.

### Steps to reproduce

@set myvar='foo'
select * from my_table
where name= :myvar
order by id;

bickgr:
Adding to follow this issue, also still happening for me.

**DBeaver Version**
CE Version 22.2.1.202209251341

**OS**
macOS 12.6

Seeing the status tab shown as default when running a script with variables or multiple selects.  Also when multiple results windows are open, when refreshing one view jumps back to the stats tab.

LonwoLonwo:
Hello.

Can you provide some video of the issue?

bickgr:
Sure thing, here's a video in action

https://user-images.githubusercontent.com/46910221/192897153-0c93b05c-e224-4df3-9c2b-be17c2d0c08f.mp4

I would expect that after running the script the default view would be the last query run, the tab labeled Demographics.  After clicking to the Demographics tab and clicking the refresh I would expect the view to remain on the Demographics tab.

ShadelessFox:
Thanks for the bug report

William-Robert-Robertson:
Hi,

I've just noticed this problem as well - at first I thought I must have made a mistake while I was tired but I realized that it seemed to have started just after upgrading.

When I run a select query or hit the Save button to send changes to the results back to the database the query runs then the Statistics tab immediately gets focus so that I have to click on the Results tab each time to get back to the results exactly as shown in the video above except that my SQL only has a single Select statement and a few comments.

Help -> About
Version 22.2.1.202209251341

Here's the database back end (MySQL):
SHOW VARIABLES LIKE 'version';
version	8.0.30

Here's the OS version for DBeaver:
[...]

PR #17782, opened 2022-09-30
#17654 #17064 SQL Terminal

### LINKED ISSUE #17064: Add option to include server output in console view
It would be useful to include the "database server output" in the console view, so that those server messages are collated with the rowset data.

This could be toggled via a config setting:
image

Consider a (MSSQL) script like this (from [comment on #667](https://github.com/dbeaver/dbeaver/issues/667#issuecomment-565182905)):
```sql
print 'Explanation for data set 1';
select ... from ...;

print `Explanation for data set 2`;
select ... from ...;

. . . etc . . .
```

Behavior as of 22.1.2.202207091909 puts printed messages only in the server output window.
image

Desired output (query text removed for clarity):
```txt
Explanation for data set 1

| # | v |
+---+---+
| 1 | 1 |
| 2 | 2 |
| 3 | 3 |

3 row(s) fetched.

Explanation for data set 2

| # | v |
+---+---+
| 1 | a |
| 2 | b |
| 3 | c |

3 row(s) fetched.
```
> See #17014 for discussion on including query text in the console view

### LINKED ISSUE #17654: Refactor SQL Terminal
We need to remove early startup from data.console plugin and create new extension point for sql terminal to let sql editor somehow  handle such contributions.

E1izabeth:
- Refactoring
- Preference to add query text
- Preference to add server output


uslss:
1. Looks like Terminal opens more than once
dbeaver_nEFRXjBMyK

uslss:
2. Output data is shown after the sql query result even if it was executed before query
dbeaver_YoA5UQxlNY

uslss:
3. Focus is set on the statistics tab instead of resultset (probably, another issue is needed)

dbeaver_AFigEClXxF

E1izabeth:
2. That's how the server output is fetched. We discussed and don't want to change it for now.
3. I guess it's this issue #17763

uslss:
verified

PR #17792, opened 2022-10-03
#16663 update Athena driver to 2.0.32 version

### LINKED ISSUE #16663: Update the Athena JDBC Driver to version 2.0.30
I have tried updating the Athena JDBC Driver in DBeaver, to version 2.0.30 however the connection could not be established as it does not accept the template settings of region in the JDBC String.

The present JDBC driver for Athena that comes with the default installation of DBeaver is SimbaAthena version 2.0.25, which was released in Nov' 2021. There has been class name changes in the JDBC driver since then.

It will be great if we can have the default version of Simba Athena JDBC Driver updated to the latest 2.0.30 which has been released in May' 2022. The links to the Official version of JDBC driver with SDK and without SDK are available on the webpage - [Using Athena with JDBC Driver](https://docs.aws.amazon.com/athena/latest/ug/connect-with-jdbc.html)

LonwoLonwo:


Matvey16:
Verified

PR #17814, opened 2022-10-05
#13238 do not set inherited status for the new created constraints

### LINKED ISSUE #13238: Create table: DDL does not include primary key
#### System information: 
- Windows 10 64 Bit
- DBeaver 21.1.2

- PostgreSQL server 10.13
- PostgreSQL driver 42.2.8
- Not using tunnels or proxies

Primary key is not included in the DDL

- Right-click on tables, choose Create table
- Add column1
- Switch to constraints, add primary key over column1
- Switch back to columns, add column2
- DDL contains both columns, but not the key

LonwoLonwo:
2022-10-05 12_58_57-DBeaver Ultimate 22 2 1 - NewTable


uslss:
verified

PR #17816, opened 2022-10-05
dbeaver/dbeaver-ee#1911 Store configs on RM for TE

E1izabeth:


plugins/org.jkiss.dbeaver.ui.dashboard/src/org/jkiss/dbeaver/ui/dashboard/registry/DashboardRegistry.java
```diff
@@ -85,26 +89,33 @@ private DashboardRegistry(IExtensionRegistry registry) {
         }
 
         // Load dashboards from config
-        File configFile = getDashboardsConfigFile();
-        if (configFile.exists()) {
+        if (DBWorkbench.getPlatform().getWorkspace().hasRealmPermission(RMConstants.PUBLIC)) {
+            log.warn("The user has no permission to load dashboard configuration");
```
LonwoLonwo, line 93: So... User do not have permissions, but we just logging this event?

E1izabeth: It's not finished yet

LonwoLonwo: A. Ok. It is not converted to draft :)

PR #17843, opened 2022-10-07
#16445 support of UUID MariaDB data type added

### LINKED ISSUE #16445: MariaDB: Foreign key "cross reference" of UUID needs to be quoted
#### System information: 
- OS: MacOS 12.3.1
- DBeaver version: 22.0.0.202203131528
- Additional extensions: none

- Database name and version: MariaDB 10.7.3
- Driver name: mariadb-java-client-2.7.1.jar / local client MySQL 8.0.28
- Do you use tunnels or proxies (SSH, SOCKS, etc)? SSH tunnel

Clicking the cross reference button image opens up a new view, which appears as a SELECT statement. However, the UUID is unquoted which causes an error. Ensuring the UUID is quoted would fix the issue.

image

LonwoLonwo:
2022-10-07 10_35_07-334 825 971 - AnyDesk

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
@@ -316,6 +316,7 @@
             <type standard="BIT"/>
             <type name="JSON"/>
             <type name="GEOMETRY"/>
+            <type name="UUID"/>
 
         </provider>
     </extension>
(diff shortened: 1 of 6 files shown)


E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #17844, opened 2022-10-07
CB-2077 save pattern variables

yagudin10:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -471,6 +471,14 @@ public static String generateVariablesLegend(@NotNull String[][] vars) {
         return text.toString();
     }
 
+    public static String extractVariableName(String string) {
```
ShadelessFox: Please add nullability annotations

ShadelessFox (approved):

PR #17850, opened 2022-10-08
1989 update eclipse version

serge-rider:


ShadelessFox:
Something is wrong with commits

Matvey16:
Verified

PR #17866, opened 2022-10-10
dbeaver/dbeaver-ee#1819 Schema variable in export wizards + default o…

serge-rider:
…utput folder + PG/MySQL refactoring


Matvey16:
Looks kind of inconistent for MySQL where the viriable ${schema} in standart export is resolved as a database name, but in dump it would be resolved as `null`

serge-rider:
Fixed

Matvey16:
Verified

PR #17869, opened 2022-10-10
15650 the value viewer panel does not remember by which column is data sorted

### LINKED ISSUE #15650: The Value viewer panel does not remember by which column is data sorted
#### System information: 
- Windows 10
- DBeaver version community 21.3.5
- Additional extensions

- sqlite
- odbc sqlite driver

The Value viewer panel does not remember by which column is data sorted

-in Database navigator, select table 
-in Data view, click on 'modelid' column (a column that has foreign keys, )
-in Value viewer panel, data from models table is displayed

Value viewer has two columns, 'values' and 'description. Sorted by 'values' by default
I click on 'description' and data in value viewer is sorted by description. Good.

back In data view, I click on another column, then again on the 'modelid'
data in value viewer is again sorted by values.

expected situation is that data in value viewer is sort by 'description', just as it was last sorted.

suggested way to fix:
remember last sorting  etting of 'value viewer', or,
add another column in define criteria tab, where users can set sorting of 'value viewer'
('define criteria' setting is saved between clicking on different columns in data view.)

Why is this needed:
Sometimes you add another model, and when viewing/editing data in the main table, from the linked table, it is easier to have data in the linked table sorted alphabetically.

Thanks.

serge-rider:


uslss:
verified

PR #17873, opened 2022-10-11
dbeaver/dbeaver-ee#2034 check existing of fmterrtbl column in Greenplum

LonwoLonwo:
to avoid errors in tables metadata reading

2022-10-11 04_08_20-Event

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumSchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumSchema.java
@@ -78,10 +78,12 @@ public class GreenplumSchema extends PostgreSchema {
                                                     @NotNull PostgreTableContainer container,
                                                     @Nullable PostgreTableBase object,
                                                     @Nullable String objectName) throws SQLException {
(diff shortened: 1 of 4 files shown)


plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumSchema.java
```diff
@@ -91,9 +93,8 @@ public JDBCStatement prepareLookupStatement(@NotNull JDBCSession session,
                     "array_to_string(x." + execLocationColumn + ", ',') AS execlocation,\n" +
                     "pg_encoding_to_char(x.encoding) AS encoding,\n" +
                     "x.writable,\n")
-                    .append(container.getDataSource().isServerVersionAtLeast(9,4) ?
-                            "" :
-                            "case when x.fmterrtbl is not NULL then true else false end as \"is_logging_errors\",\n")
+                    .append(serverSupportFmterrtblColumn ?
+                            "case when x.fmterrtbl is not NULL then true else false end as \"is_logging_errors\",\n" : "")
```
E1izabeth, line 97: Can we use the same case as in previous parts of this query here? It looks inconsistent now.

E1izabeth (approved):

Matvey16:
Verified

PR #17894, opened 2022-10-12
#15644 Remove precision and scale for unbound numeric types

### LINKED ISSUE #15644: Numeric field inside custom type displayed as numeric(1000,999)
#### System information: 
- Windows 10
- DBeaver 21.3.5.202202200327

- Postgres 10
- Driver name: PostgreSQL
- no SSH or tunnels

A custom type created with SQL and that contains a numeric field (without precision or scale) it is displayed in the source part of the type as numeric(1000,999) 

Open SQL editor and run:

```
create type core.xxx_test_type as (
	numix numeric
);
```

Then open the newly created type, go to the source tab and look at the numeric field:

image_2022-02-25_175757

ShadelessFox:


Matvey16:
Verified

PR #17905, opened 2022-10-12
#17904 Tasks view shows only active project. Task create/edit command…

### LINKED ISSUE #17904: Can't create the new task in some cases
### Description

If I do not select something and just want to create a new task by the context menu pressing on an empty place in the tasks area, I can't.
image

Also we have toolbar with "create new task", "create tasks folder" and "delete". This actions appear when I focus a navigator object, but do nothing and just disappear when I click on this buttons.

https://user-images.githubusercontent.com/28875055/195463847-788f6f23-2904-4573-bf35-e83596f5fe73.mp4

Also I don't think that we need delete toolbar button to be appeared on navigator object click. It's not clear what are we going to delete. Delete should appear only when I click on the task in my opinion.

### DBeaver Version

DBeaver 22.2.2

### Operating System

Windows 10

serge-rider:
…s enablement fix.


uslss:
verifed

PR #17926, opened 2022-10-14
#16570 Show a warning when no task types are available; introduce `standalone` task type flag

### LINKED ISSUE #16570: Can't create tasks without any connections in the project
Tasks can't be created on an empty project without any connections:

image

This is a partially expected behavior since all<sup>1</sup> tasks operate over database objects, but it's not clear what's wrong in this case.

If there are no applicable task types we should display a message describing what's happening.

_<sup>1</sup> The `Shell command` task does not require a connection._

1. Switch to an empty project
2. Attempt to create a task

ShadelessFox:
java_Mm11UopLgD

### DIFF
--- a/plugins/org.jkiss.dbeaver.tasks.ui/src/org/jkiss/dbeaver/tasks/ui/internal/TaskUIMessages.properties
+++ b/plugins/org.jkiss.dbeaver.tasks.ui/src/org/jkiss/dbeaver/tasks/ui/internal/TaskUIMessages.properties
@@ -31,6 +31,7 @@ task_config_wizard_page_task_text_label_name = Name
 task_config_wizard_page_task_control_label_descr = Description
 task_config_wizard_page_task_text_label_task_id = Task ID
 task_config_wizard_page_task_create_folder_label = Task folder
+task_config_wizard_page_task_no_task_types_available = No suitable task types found for the current project.\nTo continue, please create a new connection.
 
 task_config_wizard_stub_title_create_task = Create a task
 task_config_wizard_stub_page_name_void = Void page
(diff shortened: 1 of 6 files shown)


plugins/org.jkiss.dbeaver.tasks.ui/src/org/jkiss/dbeaver/tasks/ui/internal/TaskUIMessages.properties
```diff
@@ -31,6 +31,7 @@ task_config_wizard_page_task_text_label_name = Name
 task_config_wizard_page_task_control_label_descr = Description
 task_config_wizard_page_task_text_label_task_id = Task ID
 task_config_wizard_page_task_create_folder_label = Task folder
+task_config_wizard_page_task_no_task_types_available = No suitable task types found for the current project.\nTo continue, please create a new connection.
```
E1izabeth:
```suggestion
task_config_wizard_page_task_no_task_types_available = No suitable task types found for the current project.\nTo continue please create a new connection.
```

LonwoLonwo (approved):

E1izabeth (approved):

Matvey16:
Create task button disappeared
Expected:
dbeaver_34tVdMjwY0
Actual:
dbeaver_GLsGucsHfK

LonwoLonwo:
@Matvey16 your case is here: https://github.com/dbeaver/dbeaver/issues/17904

Matvey16:
@LonwoLonwo 
@ShadelessFox 
No, I compared it to other builds and this bug is specific to this branch, where those buttons don't appear no matter what. And closing DBeaver and installing any other build immidiately fixes it

dbeaver_5ZR2Ezwc3d

ShadelessFox:
Perhaps this branch is just outdated. I have pulled changes from `devel`, please try again

Matvey16:
Still the same

ShadelessFox:
It's weird. I'm 100% sure that it's not caused by this PR as the changes only affect the task wizard. Can we just assume that everything is ok?

Matvey16:
Verified

PR #17936, opened 2022-10-17
#16205 Append data to existing sheets

### LINKED ISSUE #16205: Improve xlsx files export
The feature added in #15067 allows the user to append exported data to existing file by adding a new sheet into file. We need to add a setting that would allow appending data to the existing sheet instead.
Another improvement is to allow changing default sheet names with templates/variables like we have for file names.

ShadelessFox:
image


E1izabeth (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #17937, opened 2022-10-17
dbeaver/dbeaver#16972 fix sqlite recreate ddl

### LINKED ISSUE #16972: SQLite drop column generates bad SQL
Operating system and version: Win Server 2019
DBeaver version: Community 22.1.1.202206261508
Driver class name: org.sqlite.JDBC
Database type: SQLite
SQLite version: 3.36.0

Describe the problem you're observing:

I believe I found some problems with the SQL that DBeaver generates when dropping a column on a table.  The generated scripts do not work in some cases.

These issues are not super-critical as I can edit the scripts DBeaver generates.

----- #1:
SQLite syntax requires that column defaults be delimited by parenthesis.

DBeaver generates this DDL:
```
CREATE TABLE sql_services (
	id_sql_service VARCHAR(16) DEFAULT lower(hex(randomblob(16))) NOT NULL,
	id_credential VARCHAR(16),
	host VARCHAR(256) NOT NULL,
	service_type VARCHAR(5) DEFAULT 'de' NOT NULL,
	instance_name VARCHAR(256) DEFAULT '' NOT NULL,
	port VARCHAR(5) DEFAULT '' NOT NULL,
	added TIMESTAMP DEFAULT strftime('%Y-%m-%d %H:%M:%f', 'now'),
	updated TIMESTAMP DEFAULT strftime('%Y-%m-%d %H:%M:%f', 'now'),
	CONSTRAINT SQL_SERVICES_PK PRIMARY KEY (id_sql_service)
);
```

It should generate (note the parenthesis around the `DEFAULT` expressions):
```
CREATE TABLE sql_services (
	id_sql_service VARCHAR(16) DEFAULT (lower(hex(randomblob(16)))) NOT NULL,
	id_credential VARCHAR(16),
	host VARCHAR(256) NOT NULL,
	service_type VARCHAR(5) DEFAULT ('de') NOT NULL,
	instance_name VARCHAR(256) DEFAULT ('') NOT NULL,
	port VARCHAR(5) DEFAULT ('') NOT NULL,
	added TIMESTAMP DEFAULT (strftime('%Y-%m-%d %H:%M:%f', 'now')),
	updated TIMESTAMP DEFAULT (strftime('%Y-%m-%d %H:%M:%f', 'now')),
	CONSTRAINT SQL_SERVICES_PK PRIMARY KEY (id_sql_service)
);
```

----- #2:
If a PK is defined, DBeaver generates a `create unique index` command.  This command is not needed and generates an error.

`CREATE UNIQUE INDEX sqlite_autoindex_sql_services_1 ON sql_services (id_sql_service);`
==> SQL Error [1]: [SQLITE_ERROR] SQL error or missing database (object name reserved for internal use: sqlite_autoindex_sql_services_1)

----- #3:
If a unique constraint is defined, DBeaver does not generate the code correctly.

SQLite cannot create unique constraints as a separate command.  They must be created when the table is created.

For example:
DBeaver generates this command:
`CREATE UNIQUE INDEX sqlite_autoindex_sql_services_2 ON sql_services (host, service_type, instance_name, id_credential);`
==> SQL Error [1]: [SQLITE_ERROR] SQL error or missing database (object name reserved for internal use: sqlite_autoindex_sql_services_2)

Instead, DBeaver should generate this DDL:
```
create table sql_services (
id_sql_service varchar(16) not null primary key default (lower(hex(randomblob(16))))
,id_credential varchar(16)
...
,latest_check timestamp
,added timestamp default (strftime('%Y-%m-%d %H:%M:%f', 'now'))
,updated timestamp default (strftime('%Y-%m-%d %H:%M:%f', 'now'))
	,unique (host, service_type, instance_name, id_credential) -- <----- right here
);
```

Destrolaric:
Closes #16972 
Fixes issue with incorrect DDL. In the process was found that we have a pretty annoying dependency on the class position in the `CHILD_TYPES` enum, it may require large refactoring in the future.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableColumnManager.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableColumnManager.java
@@ -90,29 +90,7 @@ public abstract class SQLTableColumnManager<OBJECT_TYPE extends DBSEntityAttribu
         NullNotNullModifier.appendModifier(monitor, column, sql, command);
     };
 
-    protected final ColumnModifier<OBJECT_TYPE> DefaultModifier = (monitor, column, sql, command) -> {
-        String defaultValue = CommonUtils.toString(column.getDefaultValue());
-        if (!CommonUtils.isEmpty(defaultValue)) {
-            DBPDataKind dataKind = column.getDataKind();
-            boolean useQuotes = false;//dataKind == DBPDataKind.STRING;
-            if (!defaultValue.startsWith(QUOTE) && !defaultValue.endsWith(QUOTE)) {
-                if (useQuotes && defaultValue.trim().startsWith(QUOTE)) {
-                    useQuotes = false;
-                }
-                if (dataKind == DBPDataKind.DATETIME) {
-                    final char firstChar = defaultValue.trim().charAt(0);
-                    if (!Character.isLetter(firstChar) && firstChar != '(' && firstChar != '[') {
-                        useQuotes = true;
-                    }
-                }
-            }
-
-            sql.append(" DEFAULT "); //$NON-NLS-1$
-            if (useQuotes) sql.append(QUOTE);
-            sql.append(defaultValue);
-            if (useQuotes) sql.append(QUOTE);
-        }
-    };
+    protected ColumnModifier<OBJECT_TYPE> DefaultModifier = new BaseDefaultModifier();
 
     protected ColumnModifier[] getSupportedModifiers(OBJECT_TYPE column, Map<String, Object> options)
     {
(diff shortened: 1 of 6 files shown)


E1izabeth (approved):

E1izabeth:
Maybe it would be better to override this method for `SQLiteTableColumn` manager, but then we need to almost copy paste the code from `SQLTableColumn` manager. So, I'm not sure.

Destrolaric:
> Maybe it would be better to override this method for `SQLiteTableColumn` manager, but then we need to almost copy paste the code from `SQLTableColumn` manager. So, I'm not sure.

Not sure, we only need to add extremely minor part to the query, it will be just excessive

ShadelessFox:
**These** modifications are excessive. Unnecessary abstractions just to avoid copy-pasting only required parts on existing code.

ShadelessFox:
Please create an SQLite-specific implementation of the `ColumnModifier` for the `DEFAULT` modifier with forced parentheses.

LonwoLonwo (approved):

Matvey16:
Verified

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableColumnManager.java
```diff
@@ -307,5 +285,47 @@ public static void addColumnCommentAction(List<DBEPersistAction> actionList, DBS
             "COMMENT ON COLUMN " + DBUtils.getObjectFullName(table, DBPEvaluationContext.DDL) + "." + DBUtils.getQuotedIdentifier(column) +
                 " IS " + SQLUtils.quoteString(column.getDataSource(), CommonUtils.notEmpty(column.getDescription()))));
     }
-}
 
+    protected class GenericDefaultModifier implements ColumnModifier<OBJECT_TYPE> {
```
serge-rider: It is not a Generic driver, it is a base model. Let's use another term.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableManager.java
```diff
@@ -185,6 +185,12 @@ public DBEPersistAction[] getTableDDL(DBRProgressMonitor monitor, OBJECT_TYPE ta
 
         final DBERegistry editorsRegistry = DBWorkbench.getPlatform().getEditorsRegistry();
         SQLObjectEditor<DBSEntityAttribute, OBJECT_TYPE> tcm = getObjectEditor(editorsRegistry, DBSEntityAttribute.class);
+        /*
```
serge-rider, line 188: It is not just for primary keys, it is for all constraints (primary, unique).
Maybe we should use more precise constraint types here consequently.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableColumnManager.java
```diff
@@ -90,29 +90,7 @@
...
-                }
-            }
-
-            sql.append(" DEFAULT "); //$NON-NLS-1$
-            if (useQuotes) sql.append(QUOTE);
-            sql.append(defaultValue);
-            if (useQuotes) sql.append(QUOTE);
-        }
-    };
+    protected ColumnModifier<OBJECT_TYPE> DefaultModifier = new BaseDefaultModifier();
```
E1izabeth, line 93:
```suggestion
    protected ColumnModifier<OBJECT_TYPE> defaultModifier = new BaseDefaultModifier();
```

E1izabeth (changes requested):

E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableColumnManager.java
ShadelessFox: It also should be `final`. Maybe we'll keep it's original name?

E1izabeth: I'm not insist.

PR #17952, opened 2022-10-18
dbeaver/dbeaver#17363 fix ClickhouseSSL

### LINKED ISSUE #17363: Add a Use SSL option for ClickHouse
It would be great if we had a `Use SSL` button on the ClickHouse Connection Page, just like in InfluxDB:

image

Destrolaric:
Closes #17363


E1izabeth:
Does it work both in Team Edition, Community and PRO?
I'm not sure about isDistributed() check

E1izabeth (approved):

Matvey16:
Verified

PR #17956, opened 2022-10-18
dbeaver/dbeaver-ee#2023 use eclipse resource menu to open blobs

Destrolaric:
Closes dbeaver/pro#1202
Now the Eclipse editor selection will pop up instead, where the user can select the required editor. Also, now the user can change associations via Eclipse file associations preference.


E1izabeth (approved):

LonwoLonwo (approved):

Matvey16:
Verified

Issue #8394, opened 2020-04-09
Copy field with type


plotn:
![dbeaver_field_and_type](https://user-images.githubusercontent.com/11477081/78879877-32242c00-7a5d-11ea-96b7-538b02f7c553.png)

There are 2 kinds of copy - "copy" and "full information".
Could you add "copy field + type" - in sql manner:

field1 varchar(X),
field2 int, 
....

?

kseniiaguzeeva:
Thank you for your suggestion.

plotn:
@E1izabeth, could you help me? I am trying to add breakpoint to QueryLogCommandHandler at first line (QueryLogViewer logViewer = getActiveQueryLog(event);) of:

public Object execute(ExecutionEvent event) throws ExecutionException

But this doent work. I guess I was right with this line of code? Other places are caught by debugger as well...

E1izabeth:
Similar request #15659

PR #17959, opened 2022-10-18
#8394 Copy field with type and nn

plotn:
image

image


E1izabeth:
Closes #8394

ShadelessFox (changes requested):
Please reuse code instead of creating two nearly identical generators.

You can create a `protected` method that will tell whether the `NOT NULL` part is required or not and simply override it.

plotn:
> 

done

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/generator/SQLGeneratorColumnNameTypeNN.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.sql.generator;
+
+import org.jkiss.dbeaver.DBException;
+import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
+import org.jkiss.dbeaver.model.struct.DBSEntityAttribute;
+
+public class SQLGeneratorColumnNameTypeNN extends SQLGeneratorColumnNameType {
```
ShadelessFox: And it would be better if this class was called `SQLGeneratorColumnNameTypeNotNull` instead.

Now I have a question, do we really need a separate generator for `NOT NULL`? Can't we do that always when the attribute is required?

plotn: The question is hard enough, but let us face it up. When we need "Name + Type" or something else? Simply - it is useful. When we want to create scripts more quickly. Sometimes we need strict rules, as "not null", other times - not, when we need to load temp data, that looks like existing table, but we not sure about nulls and other checks. 
Dont forget about "default" - i missed it, but not forget )))
What is the purpose of "copy" or "copy full data" in the fieldlist? both are not so convient.

Should I remove one of these? 

PS: Should we use "field type null / field type not null" or "field type / field type not null" ? First looks weird for me, but is more strict by standard. Anyway, we try avoid "field type null" in our database git....

ShadelessFox: I dunno.

It seems like in the future we might end implementing another generator that will have the `DEFAULT` modifier. And then they will multiply endlessly because we will want to have `DEFAULT` with `NOT NULL`, or without, etc.

I think the best solution would be just to include `NOT NULL` by default (if the attribute is indeed "required").

plotn: So we can call a table ddl call - then copy columns, that we need, then remove null / not null modifiers when we need / dont need them, remove default modifier in similar circumstances, remove commas if we need / or not to insert data into table.
We can do all of above.
There is 0 (zero) generators exists for DBSEntityAttribute. 
Is the growth by a two looks too much for them for a 2.5 years of initial ticket life? 
Is there a chance for them to "multiply endlessly" ? 

I dunno )))))

ShadelessFox: Okay, then it seems fine.

ShadelessFox (approved):

E1izabeth (commented):
I'd prefer to choose one of the two provided generators or add NOT NULL if the column has such attribute.

E1izabeth:
Also I'm not sure that we should generate NOT NULL if there is no NOT NULL attribute for the selected column of the source table.

plotn:
> 

I'm adding NN only for real NN fields, sure. Maybe label "name + type + not null" is not quite clear, but I did not invented something better. Anyway, one shoud catch the difference after first use, I think

E1izabeth:
But the both actions are always visible. Am I correctly understand?

plotn:
> But the both actions are always visible. Am I correctly understand?

yes. The idea is:
1. When I need to copy only "field + type", I use first action;
2. When I need to copy "field + type" and modifier "not null" for a required field - I use the second.

E1izabeth:
Well, then if object is not required, these both actions do the same.

So, we could just add to the SQLGeneratorColumnNameType::generateSQL and remove the second action.
```
        if (object.isRequired()) {
            sql.append(" NOT NULL"); //$NON-NLS-1$
        }
```
This way we will get `NOT NULL` for the cases when object is required and will not get it when it's not.

@ShadelessFox what do you think?

plotn:
> So, we could just add ...
> This way we will get `NOT NULL` for the cases when object is required and will not get it when it's not.
> @ShadelessFox what do you think?

Collegues, I understand this, but the design of these generators is slightly different, because of situations, let's see the examples:

1. I need to create function and to add some parameters like some fileds in the existing table. This is often case. I need not NOT NULL modifier whether field is required or not, I need just "p_patam1 type1, p_param2 type2" etc. NB: It could be perfect to specify additional prefix (p_), but maybe in the future.
2. Case 2. I need to copy some fields from existing table to a script (when I construct new table). Then I probably prefer not to loose NN modifier (then I probably will want DEFAULT modifier too, why not).
3. Case 3. I need to create table with similar fields, but without strict rules (e.g. for ETL purposes, etc.). In this case I do not want NN even it exists.

So I offer several combinations to these generators - with or without some parts of them.

E1izabeth (approved):

DariaMarkaryan:
Can't see any new context menu items in columns metadata, also behaviour of old menu items (Copy/Advanced copy) still the same

plotn:
> Can't see any new context menu items in columns metadata, also behaviour of old menu items (Copy/Advanced copy) still the same

I tested with Postgres Database, which database do you use? 
It works only on DBSEntityAttribute (Postgres columns are all DBSEntityAttribute)

DariaMarkaryan:
oh, 'Generate SQL' slipped past me. Now I see your changes. I've been looking for something like 'Copy ...' .-.

Matvey16:
Verified

serge-rider:
I can't say that I like this. It pollutes Generate SQL menu and it is not really convenient. Different people need different DDL in different situations. We can't add all possible cases in the context menu.

Perhaps a better solution would be to extend our Advanced Copy command (ctrl+shift+c). Currently it copies all columns of all selected rows. We can customize it, e.g. by showing popup with a list of columns with checkboxes.

WDYT?

plotn:
Hi!

I think, this work was not started more than two years (from the moment , when issue was approved), and, as a result - i did it myself. I did it the way that I could do. 
Your idea is better, sure, but is more complicated for me and I dont know - will I be able to perform it myself or not.
I think that the fear of polluting "Generate SQL" is exaggregated - there are very few elements, and they do not grow rapidly (they do not grow at all, I'd say).

It is up for you to decide...

plotn:
Column names are so often needed - with comma after each column - and there are two variants - compact (in one line) and common (with \n), so I added this generator

E1izabeth:
Maybe think it's better to do in Advanced Copy -> Copy column names. 
image
image

It has already got a menu where we could add options for columns copy (with type, with not null) along with others which are already exists in the following dialog
image

@serge-rider @ShadelessFox @LonwoLonwo what do you think?

E1izabeth:
@plotn could you move the introduced functionality to this command with additional option in the Options dialog about adding not null modifier?

plotn:
> @plotn could you move the introduced functionality to this command with additional option in the Options dialog about adding not null modifier?

I think I can, but I interpret this only as an addition to my solution, because it can only be done when there is grid with data, mine - is more flexible:
1. From the object tree;
2. From the table definition;
image
image

Please, make a desision - can we implement both - then I will plan further work at this stream

E1izabeth:
Thank you for your reply. We will discuss it soon.

E1izabeth:
@plotn we don't want to add this functionality under `Genetate DDL` command, because it duplicates `Copy column name(s)` functionality.

plotn:
Unfortunately,  this reduses the degree of my involvement. I will try to continue to make PRs and so on, but what I see - my ideas firstly approve, then marinate for an year, then I do them myself, making PR, then new coutntdown starts and lasts...
I think, I'm not enough matured, maybe.

E1izabeth:
Hello, @plotn!

Thanks for your contribution.
Unfortunately, the issue you worked on turns out to be a controversial from the UX side. We have specific life terms for open PR-s and it's been a while since no update here, so I'm closing it. 

If you want to rewrite your PR using the information proposed above, our doors are always open to you.

Best Regards,
DBeaver Team.

PR #17960, opened 2022-10-18
dbeaver/dbeaver-ee#2043 NIO filesystem directory creation/deletion

ShadelessFox:


Matvey16:
@ShadelessFox 
Is the fact that it's impossible to create a folder in root directory intentional?

dbeaver_y3VlByTDvO

dbeaver_y5VrIZ6Txq

ShadelessFox:
Yes

PR #17986, opened 2022-10-20
#11072 Show task variables before running it

### LINKED ISSUE #11072: Request variable values before execution
When a task has defined variables, these variables has to be set in the task definition. It would be great, when launching a tasks with variables, to have a screen prompting to validate or change the values of these variables.

So when defining a variable, I would expect to have more parameter to a variable:
- readonly 
   - true: variable is set in the task definition
   - false: the variable is prompted in the screen before execution
- pattern: regex to validate the input. This can be improved with custom input field (date, number, ...) but that can also be a pain to convert later...
- help message: a brief description of the parameter to help the user to write the variable value

ShadelessFox:
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.tasks.ui.view/src/org/jkiss/dbeaver/tasks/ui/view/TaskHandlerRun.java
+++ b/plugins/org.jkiss.dbeaver.tasks.ui.view/src/org/jkiss/dbeaver/tasks/ui/view/TaskHandlerRun.java
@@ -19,15 +19,19 @@ package org.jkiss.dbeaver.tasks.ui.view;
 import org.eclipse.core.commands.AbstractHandler;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
+import org.eclipse.jface.dialogs.IDialogConstants;
 import org.eclipse.jface.viewers.ISelection;
 import org.eclipse.jface.viewers.IStructuredSelection;
 import org.eclipse.ui.commands.IElementUpdater;
 import org.eclipse.ui.handlers.HandlerUtil;
 import org.eclipse.ui.menus.UIElement;
+import org.jkiss.code.NotNull;
 import org.jkiss.dbeaver.model.DBIcon;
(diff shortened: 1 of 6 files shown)


plugins/org.jkiss.dbeaver.tasks.ui.view/src/org/jkiss/dbeaver/tasks/ui/view/TaskHandlerRun.java
```diff
@@ -85,4 +93,25 @@ public void updateElement(UIElement element, Map parameters) {
         }
 
     }
+
+    private static boolean promptTaskVariables(@NotNull DBTTask task) {
```
E1izabeth: I'd prefer to call it like `confirmTaskVariables` as we stop task execution if user press Cancel or close the dialog.

plugins/org.jkiss.dbeaver.tasks.ui.view/src/org/jkiss/dbeaver/tasks/ui/view/TaskHandlerRun.java
```diff
@@ -85,4 +93,25 @@ public void updateElement(UIElement element, Map parameters) {
...
+        if (task.getType().supportsVariables()) {
```
E1izabeth: IMHO it's better to check it before calling this method. Like 
```
if (task.getType().supportsVariables() && !confirmTaskVariables(task)) {
    return;
}
```

plugins/org.jkiss.dbeaver.tasks.ui.view/src/org/jkiss/dbeaver/tasks/ui/view/TaskHandlerRun.java
ShadelessFox: It's sounds a bit odd to me but okay.

E1izabeth (approved):

LonwoLonwo (approved):

DariaMarkaryan:
Verified

PR #17987, opened 2022-10-20
dbeaver/dbeaver-ee#1981 Add remote call service and UI progress showing

E1izabeth:
For a short waiting we have a busy indicator for the cursor. If the operation is long we have a modal window showing a progress.

https://user-images.githubusercontent.com/28875055/197014260-c0b561bf-8ed8-4729-989d-fed7c259bb0b.mp4


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ui/DBPPlatformUI.java
```diff
@@ -132,6 +132,15 @@ UserChoiceResponse showUserChoice(
...
     void executeWithProgress(@NotNull DBRRunnableWithProgress runnable) throws InvocationTargetException, InterruptedException;
 
+    /**
+     * Execute runnable task synchronously while displaying job indicator if needed
+     */
+    @NotNull
+    <RESULT> RESULT executeWithProgressBlocking(
+        @NotNull String operationDescription,
+        @NotNull RunnableWithResult<RESULT> runnable
+    )throws InvocationTargetException, InterruptedException;
```
ShadelessFox: I think `RESULT` is too mouthful. In Java, you often see `T` because of conventions. Longer names may be used in case of ambiguity, but there's none.

Additionally, are you sure that this method should be marked as `@NotNull`? I believe the `runnable` can result in `null`, too. At least I was assuming so whilst scrolling through the code of it.

```suggestion
    <T> T executeWithProgressBlocking(
        @NotNull String operationDescription,
        @NotNull RunnableWithResult<T> runnable
    ) throws InvocationTargetException, InterruptedException;
```

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -1803,6 +1807,94 @@ public static <T> T syncExec(RunnableWithResult<T> runnable) {
...
+                    @Override
+                    protected Boolean runWithResultImpl() throws Exception {
+                        return !job.join(getLongOperationTime(), new NullProgressMonitor());
+                    }
+
+                    @Override
+                    protected void onCancelled() { }
+                };
+                Collection<Shell> shells = Arrays.asList(display.getShells());
+                shells.forEach(s -> s.setEnabled(false));
```
ShadelessFox: Uh, let's use plain `for each` loop here along with removing unnecessary conversion into a collection

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -1803,6 +1807,94 @@ public static <T> T syncExec(RunnableWithResult<T> runnable) {
...
+                BusyIndicator.showWhile(display, new RunnableWithResultEx<>() {
+                    protected Object runWithResultImpl() throws Exception {
+                        ModalContext.run((IProgressMonitor monitor) -> {
+                            shortWait.run();
+                        }, true, new NullProgressMonitor(), display);
```
ShadelessFox:
```suggestion
                        ModalContext.run(monitor -> shortWait.run(), true, new NullProgressMonitor(), display);
```

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -1803,6 +1807,94 @@ public static <T> T syncExec(RunnableWithResult<T> runnable) {
...
+                        return null;
+                    }
+                });
+                shells.forEach(s -> s.setEnabled(true));
```
ShadelessFox: Same as previous comment

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -1803,6 +1807,94 @@ public static <T> T syncExec(RunnableWithResult<T> runnable) {
...
         }
     }
+    
+    private static long getLongOperationTime() {
+        try {
+            return PlatformUI.getWorkbench().getProgressService().getLongOperationTime();
+        } catch (Exception ex) { // when workbench is not initialized yet during startup
+            return 800; // see org.eclipse.ui.internal.progress.ProgressManager.getLongOperationTime()
+        }
+    }
```
ShadelessFox: Let's put it somewhere at the bottom

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/runtime/RunnableWithResultEx.java
```diff
@@ -0,0 +1,80 @@
...
+     * Cancel execution
+     */
+    public final void cancel() { 
+        if (!isCancelled) {
+            isCancelled = true;
+            onCancelled();
+        }
+    }
+
+    protected void onCancelled() { }
```
ShadelessFox:
```suggestion
    protected void onCancelled() {
        // do nothing by default
    }
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/runtime/RunnableWithResultEx.java
```diff
@@ -0,0 +1,80 @@
...
+ * Runnable which stores some result
+ */
+public abstract class RunnableWithResultEx<RESULT_TYPE>
+    extends RunnableWithResult<RunnableWithResultEx.ResultOrError<RESULT_TYPE>> {
+
+    public static class ResultOrError<RESULT_TYPE> { 
+        private final RESULT_TYPE result;
+        private final Exception exception;
+        
+        public ResultOrError(RESULT_TYPE result, Exception exception) {
```
ShadelessFox: Please add nullability annotation to the `exception` parameter.

Earlier I explained why we would not want to have the `result` parameter any nullability annotations at all

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/runtime/RunnableWithResultEx.java
```diff
@@ -0,0 +1,80 @@
...
+    @Override
+    public final ResultOrError<RESULT_TYPE> runWithResult() {
+        try {
+            return new ResultOrError<>(this.runWithResultImpl(), null);
+        } catch (Exception ex) {
+            return new ResultOrError<>(null, ex);
+        }
+    }
+
+    protected abstract RESULT_TYPE runWithResultImpl() throws Exception;
```
ShadelessFox: I'd move it to be near the `onCancelled` method

ShadelessFox (changes requested):

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/runtime/DBRCancellable.java
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jkiss.dbeaver.model.runtime;
+
+public interface DBRCancellable {
```
ShadelessFox: Do we need this interface?

E1izabeth: We use it to have cancel method for RunnableWithResult

ShadelessFox: I mean, do we really need the interface? Can we keep just the method itself?

E1izabeth: yeah, you're right, we can add this method to RunnableWithResult directly

ShadelessFox (approved):

uslss:
verified

PR #17992, opened 2022-10-21
dbeaver/dbeaver-ee#2074 Prefer column name case from the metadata

ShadelessFox:
It was happening when the `rowid` was spelled in lowercase in the query.


E1izabeth (approved):

DariaMarkaryan:
Verified

PR #18008, opened 2022-10-24
#17975 Add tooltip for separate connection combo

### LINKED ISSUE #17975: Update wiki about Separate connection options
We changed UI representation of this setting here https://github.com/dbeaver/dbeaver/pull/17948
Maybe we also should describe `Default` option.

E1izabeth:
image


ShadelessFox:
I don't like that you've hardcoded database names in the tooltip.

Instead, we can overcomplicate it a bit:
1. Dynamically collect all drivers with this property set
2. Replace commas with newlines to improve visibility

ShadelessFox (approved):

flambchap:
What do you think about renaming the "Default" option to "Auto"? @E1izabeth @ShadelessFox @LonwoLonwo 

The idea behind the naming is that DBeaver decides for itself if it opens separate connections, hence "auto".

E1izabeth:
I'm not sure, that it's better. If you insist there is no problem to rename it, but "Auto" is something which is dynamically calculated for me, which is not our case.

flambchap:
Well, I'm in no position to insist. From my standpoint, it's calculated on the fly, indeed, since it depends on the data source.

PR #18009, opened 2022-10-24
dbeaver/dbeaver-ee#2088 confirmation if statement opens a lot of results

Destrolaric:
Closes dbeaver/pro#1233
Adds a new preference to warn users about many result sets being opened. If `No` is pressed, we don't load other result sets. Otherwise, we continue loading. 20 is default.
image
image


ShadelessFox (approved):

E1izabeth (approved):

DariaMarkaryan:
Verified

PR #18018, opened 2022-10-24
Add setting for SQL editor variable pattern

### LINKED ISSUE #15978: Ability to define variables using flexible syntax
DBeaver is a great tool and I often use it copying SQL from other tools into DBeaver to process, review and to develop it further.
After that I copy it back to the other tool - in my case ETL tools.
As ETL also use variables in their individual style it is always a process of replaceing them when copying the SQL to DBeaver and then replaceing them back before copying it back to the individual tool. This is very time consuming and tidious so I would prefer an easy way by supporting all kind of variable styles.

It would be great to get DBeaver supporting all kinds of variable styles.
I could imagine a command upfront the SQL like 

> SET VARIABLE "#PS_SAMPLE.DBSchema# = BASE

Which would dynamically replace the parameterized schema name in the following SQL with the the schema string BASE.

> SELECT col1, col2
> FROM #PS_SAMPLE.DBSchema#.EMPLOYEE
> WHERE col1 > #Date#

To replace #Date# as well an additional SET could be used like this:

> SET VARIABLE "#Date# = '2022-03-26'

The parameter/variable style of #variable# is just an example - it should be flexible to support this for as many other tools as possible.
Also the SET VARIABLE is just an example and could be any style and syntax. The only important thing is that the original SQL does not need to be changed.

Alternatively there could be a central configuration to specify the parameter style.

This would enable us to copy SQL to DBeaver and running it unchanged and then copying it back with no addtional changes.
This realy would improve the usability and would be a great help. 

This could also solve the following feature request as well:
_Custom syntax for variable placeholders_ #12590 

Thanks in advance

mcpiroman:
Adds a setting to specify the exact regex used for variables.

todo:
- Less hacky implementation of `ScriptVariableRule`?
- Tip text?
- A list of predefined formats?
- Tests?

Please share your feedback.

Fixes #15978,

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRuleManager.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRuleManager.java
@@ -144,7 +144,7 @@ public class SQLRuleManager {
         {
             if (!minimalRules && syntaxManager.isVariablesEnabled()) {
                 // Variable rule
-                rules.add(new ScriptVariableRule(parameterToken));
+                rules.add(new ScriptVariableRule(parameterToken, syntaxManager.getVariablePattern()));
             }
         }
 
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLScriptParser.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLScriptParser.java
@@ -714,10 +714,19 @@ public class SQLScriptParser {
                         parameters = new ArrayList<>();
                     }
 
+                    String varName = paramName;
+                    for (String prefix : syntaxManager.getNamedParameterPrefixes()) {
+                        if (varName.startsWith(prefix)) {
+                            varName = varName.substring(prefix.length());
+                            break;
+                        }
+                    }
+
                     SQLQueryParameter parameter = new SQLQueryParameter(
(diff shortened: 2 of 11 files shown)


E1izabeth:
Any update? Will you finish it or should I close this PR?

mcpiroman:
I'd like to first hear some feedback about which changes are welcome and which of my todos to include.

E1izabeth:
Ok, we will look at it and give you some feedback

E1izabeth:
@mcpiroman
Sorry for so long waiting for review. We missed your PR as it was draft.

We made some changes recently, so you need to pull changes from `devel` to continue.
We strip prefix on the parse stage now.

You could briefly review incoming changes in this PRs (some of these changes are not affecting what are you doing, but some new methods could be useful):
https://github.com/dbeaver/dbeaver/pull/18613
https://github.com/dbeaver/dbeaver/pull/18711
https://github.com/dbeaver/dbeaver/pull/18761

Pay attention at `SQLScriptParser:723`. It will need correction.

In general your solution is ok. I'd like the idea of adding a tip for a new setting. Of course, it's better to write unit tests.
Not sure about predefined formats, I suppose it's not necessary.

E1izabeth:
@mcpiroman do you need any help?

mcpiroman:
No, I just have not time ATM.

LonwoLonwo:
Hello @mcpiroman 

Thanks for your contribution.
But for now, we have many unanswered questions about the code. And we have specific life terms for open PR-s.

I'm closing it now because of it and current conflicts.

If you want to rewrite your PR using the information proposed above, our doors are always open to you.
Best Regards,
DBeaver Team.

PR #18035, opened 2022-10-25
#15352 Add hint for database field on the connection page for PostgreSQL

### LINKED ISSUE #15352: Authentication username is used as database name
If authentication username is different than `postgres` while the `Database` field is empty then DBeaver fails to establish a connection to the DB because it uses authentication username as a database name.

I'm using DBEaver Community, Version 21.3.3.202201221033, Release 24th January 2022.

I'm using `pavel` as my authentication user name:
Screen Shot 2022-02-01 at 17 02 43

When I click on `Test Connection` button I get fatal error that database with name `pavel` does not exists
Screen Shot 2022-02-01 at 17 04 01

Seems like the only reason it worked previously is because I used `postgres` for authentication name.
This is very annoying bug as this prevents from using users names to separate permissions and track users operations in audit log.

E1izabeth:
image

By default we set database and user name to postgres, so user will see this hint only if they intentionally deletes the text from this field.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/PostgresResources.properties
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/PostgresResources.properties
@@ -137,7 +137,7 @@ dialog_setting_connection_password = Password
 dialog_setting_connection_port = Port
 dialog_setting_connection_settings = Settings
 dialog_setting_connection_user = User
-dialog_database_name_hint = equals to username if not specified
+dialog_database_name_hint = Username is used if not specified
 
 # PostgreCreateRoleDialog
 dialog_create_role_title = Create role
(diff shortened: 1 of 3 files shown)


Destrolaric:
If possible, it's better to show the current value of the username text box instead.

E1izabeth:
1) username field could be empty too
2) it require some refactoring: at least I need access to the username field of auth model configurator and the method to add TextListener to this field from the postgresql connection page. I find this solution cumbersome for such a task. In addition, it loses the useless if the username field is empty too.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/PostgresResources.properties
```diff
@@ -137,6 +137,7 @@ dialog_setting_connection_password = Password
 dialog_setting_connection_port = Port
 dialog_setting_connection_settings = Settings
 dialog_setting_connection_user = User
+dialog_database_name_hint = equals to username if not specified
```
Destrolaric: Probably 
`Uses username if not specified`
is better

Destrolaric (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/PostgresResources.properties
E1izabeth: yeah, maybe

DariaMarkaryan:
Verified

Destrolaric (approved):

PR #18047, opened 2022-10-26
#17266 Reveal hidden attribute in its original position

### LINKED ISSUE #17266: Hidden columns shows in the end of grid
### Description

When you hide columns in the grid and then show them again, this column expands at the end of the grid

Screenshot 2022-08-04 at 18 25 48

### Steps to reproduce

_No response_

### DBeaver Version

22.1.0 UE

### Operating System

Windows 11

ShadelessFox:


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -276,7 +276,8 @@ public void setAttributeVisibility(@NotNull DBDAttributeBinding attribute, boole
             constraint.setVisible(visible);
             if (attribute.getParentObject() == null || attribute.getParentObject() == documentAttribute) {
                 if (visible) {
-                    visibleAttributes.add(attribute);
+                    final int position = Math.min(constraint.getVisualPosition(), visibleAttributes.size());
```
E1izabeth, line 279: What does constraint.getVisualPosition() return if constraint.isVisible() == false? Will not it turn out that visualPosition is less than zero or something like that?

ShadelessFox: No. It returns the current visual position. You can see it in the filter panel even when the attribute is hidden.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -276,7 +276,8 @@ public void setAttributeVisibility(@NotNull DBDAttributeBinding attribute, boole
...
+                    visibleAttributes.add(position, attribute);
```
Destrolaric, line 280: Looks like it will have IndexOutOfBounds for the `visibleAttributes.size()`
EDIT: nvm

Destrolaric (changes requested):

Destrolaric (approved):

DariaMarkaryan:
Verified

PR #18048, opened 2022-10-26
CB-2638: Oracle wallet authentication support. Read features

oksana-cherniavskaia:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/properties/ObjectPropertyDescriptor.java
```diff
@@ -222,10 +220,21 @@ public boolean isEditable(Object object)
         return getEditableValue(object);
     }
 
+    private List<String> parseFeatures() {
+        String features = propInfo.features();
+        if (features == null || features.isBlank()) {
```
LonwoLonwo: Please look at the CommonUtils#isNotEmpty method in the future

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/properties/ObjectPropertyDescriptor.java
```diff
@@ -243,11 +252,19 @@ public String[] getFeatures() {
...
         if (this.isViewable()) features.add(DBConstants.PROP_FEATURE_VIEWABLE);
         if (this.isPassword()) features.add(DBConstants.PROP_FEATURE_PASSWORD);
+
         return features.toArray(new String[0]);
     }
 
     @Override
     public boolean hasFeature(@NotNull String feature) {
+
+        List<String> features = parseFeatures();
```
serge-rider: Function hasFeature may be called 1000 times per second. It is not good to parse, create array, etc for every call.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/properties/ObjectPropertyDescriptor.java
serge-rider: Or maybe we should make it String[] features() default {};

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/properties/ObjectPropertyDescriptor.java
oksana-cherniavskaia: done

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/properties/ObjectPropertyDescriptor.java
```diff
@@ -280,7 +283,10 @@ public boolean hasFeature(@NotNull String feature) {
             case DBConstants.PROP_FEATURE_PASSWORD:
                 return this.isPassword();
         }
-        return false;
+
+        String[] features = propInfo.features();
```
serge-rider: Again, this function may be called 10000 times per second. Do not use expensive operations including streams.
Use check for empty array (99% cases) + ArrayUtils.contains.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/properties/ObjectPropertyDescriptor.java
```diff
@@ -280,7 +284,12 @@ public boolean hasFeature(@NotNull String feature) {
             case DBConstants.PROP_FEATURE_PASSWORD:
                 return this.isPassword();
         }
-        return false;
+
+        if (ArrayUtils.isEmpty(propInfo.features())) {
```
alexander-skoblikov: such  check is already exist in ArrayUtils.contains

plugins/org.jkiss.dbeaver.model/OSGI-INF/l10n/bundle.properties
```diff
@@ -156,6 +156,8 @@ meta.org.jkiss.dbeaver.model.navigator.DBNResource.resourceSize.name=Size
 meta.org.jkiss.dbeaver.model.impl.auth.AuthModelDatabaseNativeCredentials.userName.name = User name
 meta.org.jkiss.dbeaver.model.impl.auth.AuthModelDatabaseNativeCredentials.userPassword.name = User password
 
+meta.com.dbeaver.db.oracle.net.auth.wallet.OracleAuthModelWalletCredentials.walletArchive.name = Oracle wallet archive
```
serge-rider: This must be in Oracle EE bundle

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/OSGI-INF/l10n/bundle.properties
oksana-cherniavskaia: done

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/properties/ObjectPropertyDescriptor.java
oksana-cherniavskaia: addressed

serge-rider (approved):

PR #18059, opened 2022-10-27
#17043 Add parameters rule before word rule

### LINKED ISSUE #17043: Named parameter prefix doesn't work anymore with underscores
#### System information: 
- Operating system (distribution) and version: Windows 11
- DBeaver version: 22.1.2.202207091909
- Additional extensions

- Database name and version: PostgreSQL 14
- Driver name

In Preferences / Editors / SQL Editor / SQL processing, the "Named parameter prefix" doesn't work anymore with the underscore character (it used to work in 22.0.4).
It works with :, &, +, ... 
We have tons of scripts with underscore as parameter, so it would be preferable to keep this setting, otherwise I'll stick with the old version.

Set "Named parameter prefix" to _
>select _parameter
SQL Error [42703]: ERROR: column "_parameter" does not exist
  Position: 8

E1izabeth:
Was broken by https://github.com/dbeaver/dbeaver/pull/16290

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRuleManager.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRuleManager.java
@@ -217,6 +217,11 @@ public class SQLRuleManager {
         }
 
         if (!minimalRules) {
+            // Parameter rule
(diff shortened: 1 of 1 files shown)


ShadelessFox (changes requested):
Please fix spelling in the commit's message

ShadelessFox (approved):

DariaMarkaryan:
Verified

PR #18061, opened 2022-10-27
#16520 Normalize resource path on properties read/write

### LINKED ISSUE #16520: Active schema changes without notice after disconnection
#### System information: 
Windows 10 Enterprise 21H2
DBeaver 22.0.4.202205011839
No additional extensions

10.1.48-MariaDB-0ubuntu0.18.04.1
MariaDB
SSH Tunnel

Active catalog/schema dropdown not updated after disconnection while in reality the SQL executed after disconnection are executed on the default DB. This could lead to data loss because the query is not executed in the intended database.

1. Open an SQL editor and select a not default database
2. Write a query
3. Execute the query on the not default database
4. Disconnect and reconnect
5. Execute the query again, this time will be executed on the default database

serge-rider:


DariaMarkaryan:
Verified

PR #18099, opened 2022-11-01
#15268 Pass parameters to dependent methods during completion resolution

### LINKED ISSUE #15268: [Autocompletion] Autocompletion inside packages
Part of #15103
2. Autocompletion with functions inside packages is not working properly as before, now it autocompletes parameters with their types instead of parameter names as before:
image

It should autocomplete the parameter names in the form :param_value, or param_name => :parameter_value (better)

E1izabeth:


LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
```diff
@@ -775,7 +775,7 @@ private boolean tableHaveJoins(DBSEntity table1, DBSEntity table2) {
         }
     }
 
-    private void makeDataSourceProposals() throws DBException {
+    private void makeDataSourceProposals(Map<String, Object> parameters) throws DBException {
```
ShadelessFox: Just a nitpick: please add nullability annotations

ShadelessFox (approved):

uslss:
verified

ShadelessFox (approved):

PR #18100, opened 2022-11-01
#17246 Search for multiple aliases or table names while completion

### LINKED ISSUE #17246: Auto complete only offers one of two table aliases
### Description

I have to work with especially long table names for some queries. (DB is PostgreSQL 11)
The table aliases are generated by DBeaver and for this expample below, DBeaver only offers `sadpoootbapt` for autocomplete when typing `sad`. `sadpoootbtwt` is missing.

```sql
select *
from hub_aggr_daily_production_order_operation_occupied_time hadpooot
join sat_aggr_daily_prod_order_op_occupied_time_by_adj_prod_time sadpoootbapt
  on hadpooot.id = sadpoootbapt.dpooot
join sat_aggr_daily_prod_order_op_occupied_time_by_total_work_time sadpoootbtwt
  on hadpooot.id =   <auto complete only offers `sadpoootbapt` here>
```

### Steps to reproduce

_No response_

### DBeaver Version

Community Edition Version 22.1.3.202207240747

### Operating System

macOs 10.15.7

E1izabeth:


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
```diff
@@ -1317,7 +1317,7 @@ public void visit(@Nullable CreateView createView) {
                             if (aliases.contains(s) || sqlDialect.getKeywordType(s) != null) {
                                 return true;
                             }
-                            return extractTableName(s, false) != null;
+                            return extractTableNames(s, false).size() > 0;
```
ShadelessFox:
```suggestion
                            return !extractTableNames(s, false).isEmpty();
```

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
```diff
@@ -664,8 +664,8 @@ private void makeProposalsFromQueryParts() {
             if (wordPart.indexOf(request.getContext().getSyntaxManager().getStructSeparator()) != -1 || wordPart.equals(ALL_COLUMNS_PATTERN)) {
                 return;
             }
-            final Pair<String, String> name = extractTableName(wordPart, true);
-            if (name != null) {
+            final List<Pair<String, String>> names = extractTableNames(wordPart, true);
+            for (Pair<String, String> name: names) {
```
ShadelessFox:
```suggestion
            for (Pair<String, String> name : extractTableNames(wordPart, true)) {
```

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
```diff
@@ -918,11 +918,13 @@ private DBSObject getTableFromAlias(DBSObjectContainer sc, @Nullable String toke
             token = token.substring(0, token.length() - 1);
         }
 
-        final Pair<String, String> name = extractTableName(token, false);
-        if (name != null && CommonUtils.isNotEmpty(name.getFirst())) {
-            final String[][] quoteStrings = sqlDialect.getIdentifierQuoteStrings();
-            final String[] allNames = SQLUtils.splitFullIdentifier(name.getFirst(), catalogSeparator, quoteStrings, false);
-            return SQLSearchUtils.findObjectByFQN(monitor, sc, request.getContext().getExecutionContext(), Arrays.asList(allNames), !request.isSimpleMode(), request.getWordDetector());
+        final List<Pair<String, String>> names = extractTableNames(token, false);
+        for (Pair<String, String> name: names) {
```
ShadelessFox:
```suggestion
        for (Pair<String, String> name : extractTableNames(token, false)) {
```

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
E1izabeth: I'd prefer to have this variable for debug. It used to be a separate variable and I don't like to change it, because it's useful.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
E1izabeth: The same is here.

LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
```diff
@@ -1317,7 +1317,7 @@ public void visit(@Nullable CreateView createView) {
                             if (aliases.contains(s) || sqlDialect.getKeywordType(s) != null) {
                                 return true;
                             }
-                            return extractTableName(s, false) != null;
+                            return extractTableNames(s, false).isEmpty();
```
ShadelessFox: We want to make sure that there are results, don't we?

```suggestion
                            return !extractTableNames(s, false).isEmpty();
```

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
```diff
@@ -938,7 +940,7 @@ private enum InlineState {
     };
 
     @Nullable
-    private Pair<String, String> extractTableName(@Nullable String tableAlias, boolean allowPartialMatch) {
+    private List<Pair<String, String>> extractTableNames(@Nullable String tableAlias, boolean allowPartialMatch) {
```
ShadelessFox, line 943: This method is still marked as nullable but its return type is treated as not-null without any checks in several places. I beg you to use IntelliJ IDEA for development, at least it has nullability inspections that would have prevented you from doing this.

But now all its `null` paths can now return empty lists instead.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
E1izabeth: Thank you, I forgot to change nullable annotation. Now It's changed to NotNull.

Actually, IDEA says nothing to me. And I don't quite understand which "null paths" are you talking about, because all usages of this method was fixed.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
```diff
@@ -937,8 +939,8 @@ private enum InlineState {
         MATCHED
     };
 
-    @Nullable
-    private Pair<String, String> extractTableName(@Nullable String tableAlias, boolean allowPartialMatch) {
+    @NotNull
+    private List<Pair<String, String>> extractTableNames(@Nullable String tableAlias, boolean allowPartialMatch) {
         final SQLScriptElement activeQuery = request.getActiveQuery();
         if (activeQuery == null) {
             return null;
```
ShadelessFox: Here's the remaining `null` path

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
ShadelessFox: I guess that's because you haven't configured annotation processing yet.

image

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
E1izabeth: Yeah, thanks.

uslss:
verified

PR #18106, opened 2022-11-02
#15266 Fix changing case for block beginning keywords

### LINKED ISSUE #15266: [Autocompletion] Begin end autocompletion issues
Part of #15103
3. Autocompletion only works for select statements in sql editor, if you write a piece of code like the following example, it doesn't autocomplete anything:
```
BEGIN
  package_name.*;
  current_schema.package_name.*;
  other_schema.package_name.*;
END;
```

4. In sql editor when you just write "begin" in lowercase and press enter, what it does is to autocomplete with following:
```
BEGIN

end
```

Even when you have capitalization set up in lowercase,  it ignores it and puts the being in upper case, and the later end in lowercase. Also it should add a ; and a / at the end if it is placed in the sqleditor, or just the ; if it is in the package editor. The correct autocompletion should be:

```
begin

end;
/
```

5. The "END IF" statement is not properly managed, having this code as an example:

```
BEGIN
	IF(1=2) THEN
	  test_code();
	END IF;
END;
/
```

E1izabeth:


uslss:
verified

PR #18107, opened 2022-11-02
CB-2797: TE. Support deleting connection folders

oksana-cherniavskaia:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/rm/RMController.java
```diff
@@ -85,6 +85,19 @@ void saveProjectDataSources(
...
+    /**
+     * Delete project connection folders.
+     *
+     * @param projectId    the project id
+     * @param folderPaths  the folder paths
+     * @param dropContents the drop contents
+     * @throws DBException the db exception
+     */
+    void deleteProjectConnectionFolders(
+        @NotNull String projectId, @NotNull String[] folderPaths,
```
ShadelessFox:
```suggestion
        @NotNull String projectId,
        @NotNull String[] folderPaths,
```

PR #18127, opened 2022-11-04
#18121 Fix data export on file conflict

### LINKED ISSUE #18121: Export to a file always prompts to take actions against existing file
### Description

Export to a file always prompts to take actions against existing file, even if there were no any files before starting the export. An empty file is created before the dialog is opened, and if there was a file, it will be truncated, making the `Append` choice useless.

### DBeaver Version

DBeaver 22.2.4

E1izabeth:


Matvey16:
Verified

PR #18132, opened 2022-11-07
#16189 Don't force limit for fetch size for clickhouse

### LINKED ISSUE #16189: Patch query with limit causing problem on ClickHouse
#### System information: 
- macOS 12.3.1
- DBeaver Version 22.0.2.202204021640

- ClickHouse 22.3
- clickhouse-jdbc 0.3.2-patch7
- No tunnel or proxy used

`select * from numbers(5) format JSONEachRow` does not work because DBeaver patched the query with limit, resulting below query sent to ClickHouse. This was discussed in PR #15492.

```sql
select * from numbers(5) format JSONEachRow
LIMIT 0, 200
```

Issue above query against ClickHouse.

E1izabeth:


serge-rider:
Changes are in devel

PR #18135, opened 2022-11-07
Postgres better function calling#16352

### LINKED ISSUE #16352: PostgreSQL: better function calling
Hi! There is well known PostgreSQL feature (when calling functions) - we must strictly use specified types.
E.g. function:

CREATE OR REPLACE FUNCTION test_function(
	p_task_id		int8,
	p_item_title 	varchar(2000),
	p_item_author	int8,
	p_create_ts		timestamp,
	p_update_ts		timestamp
)

cannot be called via:
select test_function(
	p_task_id => 1,
	p_item_title => 'тестовый элемент 3',
	p_item_author => 406,
	p_create_ts => current_timestamp,
	p_update_ts => null
)

The error is - "function does not exist. No function matches the given name and argument types."

When we "right click" onto function we see "Generate SQL (CALL)" and "Execute procedure", The SQL generated is:
SELECT test_function(:p_task_id,:p_item_title,:p_item_author,:p_create_ts,:p_update_ts,:p_closed_by,:p_close_ts,:p_close_comment,:p_before_id);

And this does not solve our problem, the only way to call function is:
select test_function(
	p_task_id => 1::int8,
	p_item_title => 'тестовый элемент 3'::varchar,
	p_item_author => 406::int8,
	p_create_ts => current_timestamp::timestamp,
	p_update_ts => null::timestamp
)

The idea is: 
1. Generate "fully qualified call sql" - with type casting via :: or (maybe better) cast(x as ...)
2. Add column "data type" to params binding dialog and use it in an appropriate way (add casts "underneath"). 

I think both ideas should be realized.

plotn:
dbeaver_casted_params

as shown...


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -822,25 +823,35 @@ public void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, Co
...
             boolean first = true;
             for (DBSProcedureParameter parameter : inParameters) {
+                String typeName = parameter.getParameterType().getFullTypeName();
                 switch (parameter.getParameterKind()) {
                     case IN:
                         if (!first) {
-                            sql.append(",");
+                            sql.append(", ");
+                        }
+                        if (castParams) {
```
serge-rider, line 846: You can put PostgreSQL-specific cast syntax to abstract SQL dialect. It has to be in PostgreDialect

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
```diff
@@ -251,6 +255,23 @@ public void run() {
                         }
                     }, true));
             }
+            if (sourceObject.supportsObjectDefinitionOption(DBPScriptObject.OPTION_CAST_PARAMS)) {
+                toolBarManager.add(ActionUtils.makeActionContribution(
+                        new Action(SQLEditorMessages.source_viewer_cast_params_text, Action.AS_CHECK_BOX) {
```
serge-rider, line 260: We probably shouldn't add two identical localizations for "Cast params" string. Let's use the same SQLEditorMessages constants everywhere.

Also I'd use "Use explicit type cast" instead.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
plotn: Sorry, I dont agree. This is common syntax, not only for postgres. It brings more help to postgres, but applicable as 'common sql syntax'

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
plotn: But you use them too:

sql_generator_dialog_button_show_comments = Show comments
sql_generator_dialog_button_show_permissions = Show permissions

source_viewer_show_comments_text = Show comments
source_viewer_show_permissions_text = Show permissions

How should I do correct?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
E1izabeth: According to SQL-2003 grammar cast is defined as `CAST (<cast operand> AS <cast target>)`.
So, it's not something PostgreSQL-specific unless we want to add this feature only for PostgreSQL for now.

E1izabeth: The thing I'm not sure about is `:` before parameter name.

plotn: Why? This works fine.

Yes, you are right. The PG specific cast is "<cast operand>::<cast target>" and I intentionally did not use it (also :: conflicts with :param).

E1izabeth: I'm not sure, that `:param-name` is supported in all databases

plotn: I left this design as it was originally. There are 2 variations in code - ":param" and question mark.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
ShadelessFox: The `:param` is DBeaver-specific syntax.

DBeaver will prompt the user to enter values for these parameters before executing the query.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
ShadelessFox: They meant you should call it differently.

plotn: But I already changed to "Use explicit type cast"

ShadelessFox: Oh, sorry, I didn't notice.

ShadelessFox (approved):

Matvey16:
Verified

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
serge-rider: Then we should use preference SQL_NAMED_PARAMETERS_PREFIX instead of hardcoded `:`.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBPScriptObject.java
```diff
@@ -66,6 +66,8 @@ public interface DBPScriptObject extends DBPObject {
     String OPTION_INCLUDE_PARTITIONS = "ddl.includePartitions"; //$NON-NLS-1$
 
     String OPTION_USE_SPECIAL_NAME = "ddl.useSpecialName"; //$NON-NLS-1$
+    
+    String OPTION_CAST_PARAMS = "ddl.castParams"; //$NON-NLS-1$
```
serge-rider: It is not DDL-specific option. It is generic SQL option so lets name it "sql.castParameter".

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
```diff
@@ -251,6 +255,23 @@ public void run() {
                         }
                     }, true));
             }
+            if (sourceObject.supportsObjectDefinitionOption(DBPScriptObject.OPTION_CAST_PARAMS)) {
```
serge-rider, line 258: I don't see any DBPScriptObjectExt2 implementation which supportsObjectDefinitionOption for the new parameter. 
It should be supported by procedures I suppose? Or I miss something..

serge-rider (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
plotn: @serge-rider , there is harcoded : (colon) in your existing code, so I left this logic unchanged. If it shoud be updated, couuld you give me an example how to use SQL_NAMED_PARAMETERS_PREFIX ?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBPScriptObject.java
plotn: Ok, agreed

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
plotn: @serge-rider , help me please - should I remove here

if (sourceObject.supportsObjectDefinitionOption(DBPScriptObject.OPTION_CAST_PARAMS)) {
                options.put(DBPScriptObject.OPTION_CAST_PARAMS, getShowCastParams());
            }

?
(I did it to be like other options in code)

plotn:
@E1izabeth , @serge-rider help me please finish this work - there are minor questions above

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
E1izabeth: As I see the condition `sourceObject.supportsObjectDefinitionOption(DBPScriptObject.OPTION_CAST_PARAMS)` will never be true, because implementation of this method is like in PostgreProcedure:supportsObjectDefinitionOption
```
    public boolean supportsObjectDefinitionOption(String option) {
        return DBPScriptObject.OPTION_INCLUDE_COMMENTS.equals(option) || DBPScriptObject.OPTION_INCLUDE_PERMISSIONS.equals(option);
    }
```
I doubt how it works for you. I guess you need to modify `PostgreProcedure:supportsObjectDefinitionOption` like
```
    public boolean supportsObjectDefinitionOption(String option) {
        return DBPScriptObject.OPTION_INCLUDE_COMMENTS.equals(option) || DBPScriptObject.OPTION_INCLUDE_PERMISSIONS.equals(option) || DBPScriptObject.OPTION_CAST_PARAMS.equals(option);
    }
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
E1izabeth: @plotn like this
```
DBPPreferenceStore prefStore;
DBPDataSource dataSource = proc.getDataSource();
if (dataSource != null) {
    prefStore = dataSource.getContainer().getPreferenceStore();
} else {
    prefStore = DBWorkbench.getPlatform().getPreferenceStore();
}
String namedParameterPrefix = prefStore.getString(ModelPreferences.SQL_NAMED_PARAMETERS_PREFIX);
```

plotn: Okay! it works

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
plotn: Ahhh, thank you, @E1izabeth , I understand now

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDialect.java
```diff
@@ -262,7 +262,8 @@ public String getColumnTypeModifiers(
     }
 
     @Override
-    public void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, Collection<? extends DBSProcedureParameter> parameters) {
+    public void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, Collection<? extends DBSProcedureParameter> parameters,
+                                            boolean castParams) {
```
ShadelessFox:
```suggestion
    public void generateStoredProcedureCall(
        StringBuilder sql,
        DBSProcedure proc,
        Collection<? extends DBSProcedureParameter> parameters,
        boolean castParams
    ) {
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
```diff
@@ -828,7 +828,8 @@ public List<PostgreDependency> getDependencies(DBRProgressMonitor monitor) throw
 
     @Override
     public boolean supportsObjectDefinitionOption(String option) {
-        return DBPScriptObject.OPTION_INCLUDE_COMMENTS.equals(option) || DBPScriptObject.OPTION_INCLUDE_PERMISSIONS.equals(option);
+        return DBPScriptObject.OPTION_INCLUDE_COMMENTS.equals(option) || DBPScriptObject.OPTION_INCLUDE_PERMISSIONS.equals(option) 
+                || DBPScriptObject.OPTION_CAST_PARAMS.equals(option);
```
ShadelessFox: We use 4 spaces for continuation indent:

```suggestion
        return DBPScriptObject.OPTION_INCLUDE_COMMENTS.equals(option)
            || DBPScriptObject.OPTION_INCLUDE_PERMISSIONS.equals(option) 
            || DBPScriptObject.OPTION_CAST_PARAMS.equals(option);
```

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/generator/SQLGenerator.java
```diff
@@ -122,10 +123,18 @@ public boolean isShowPartitionsDDL() {
...
+    }
+
     public void setShowPartitionsDDL(boolean showPartitionsDDL) {
         this.showPartitionsDDL = showPartitionsDDL;
     }
 
+    public void setShowCastParams(boolean showCastParams) {
+        this.showCastParams = showCastParams;
+    }
+
```
ShadelessFox: Please keep get-set pairs together

```suggestion
    public void setShowPartitionsDDL(boolean showPartitionsDDL) {
        this.showPartitionsDDL = showPartitionsDDL;
    }

    public boolean isShowCastParams() {
        return showCastParams;
    }

    public void setShowCastParams(boolean showCastParams) {
        this.showCastParams = showCastParams;
    }
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -809,37 +812,56 @@ protected String getProcedureCallEndClause(DBSProcedure procedure) {
     }
 
     @Override
-    public void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, Collection<? extends DBSProcedureParameter> parameters) {
+    public void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, Collection<? extends DBSProcedureParameter> parameters, 
+                                            boolean castParams) {
```
ShadelessFox:
```suggestion
    public void generateStoredProcedureCall(
        StringBuilder sql,
        DBSProcedure proc,
        Collection<? extends DBSProcedureParameter> parameters,
        boolean castParams
    ) {
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialect.java
```diff
@@ -441,7 +441,11 @@ default boolean supportsAliasInConditions() {
      */
     String formatStoredProcedureCall(DBPDataSource dataSource, String sqlText);
 
-    void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, Collection<? extends DBSProcedureParameter> parameters);
+    /**
+     * Generates stored procedure call. Parameters (optionally) can be surrounded by cast(:param as paramType).
+     */
+    void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, Collection<? extends DBSProcedureParameter> parameters,
+                                     boolean castParams);
```
ShadelessFox: Please specify nullability annotations for the method's parameters. Don't forget about its implementations.

```suggestion
    void generateStoredProcedureCall(
        StringBuilder sql,
        DBSProcedure proc,
        Collection<? extends DBSProcedureParameter> parameters,
        boolean castParams);
```

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
```diff
@@ -288,4 +309,11 @@ protected Boolean getShowPartitionsDDL() {
         }
         return showPartitionsDDL;
     }
+
+    protected Boolean getShowCastParams() {
```
ShadelessFox:
```suggestion
    protected boolean getShowCastParams() {
```

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDialect.java
```diff
@@ -262,8 +262,12 @@ public String getColumnTypeModifiers(
     }
 
     @Override
-    public void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, Collection<? extends DBSProcedureParameter> parameters,
-                                            boolean castParams) {
+    public void generateStoredProcedureCall(
+	    StringBuilder sql, 
+		DBSProcedure proc, 
+		Collection<? extends DBSProcedureParameter> parameters,
+        boolean castParams
```
ShadelessFox: Don't mix tabs and spaces (also missing annotations, see https://github.com/dbeaver/dbeaver/pull/18135#discussion_r1085280631)

```suggestion
        StringBuilder sql, 
        DBSProcedure proc, 
        Collection<? extends DBSProcedureParameter> parameters,
        boolean castParams
```

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialect.java
```diff
@@ -441,7 +441,15 @@ default boolean supportsAliasInConditions() {
...
-    void generateStoredProcedureCall(StringBuilder sql, DBSProcedure proc, Collection<? extends DBSProcedureParameter> parameters);
+    /**
+     * Generates stored procedure call. Parameters (optionally) can be surrounded by cast(:param as paramType).
+     */
+    void generateStoredProcedureCall(
+        StringBuilder sql, 
+        DBSProcedure proc, 
+        Collection<? extends DBSProcedureParameter> parameters,
+        boolean castParams
+    );
```
ShadelessFox, line 452: Please specify nullability annotations for the method's parameters. Don't forget about its implementations.

plotn: Please,  give me an example. I just edited your file - so it (maybe) do not meet your new requirement?

E1izabeth: We use annotations to specify function arguments and return type nullability: `@NotNull` and `@Nullable`
It can be done like the following in this case
```
    void generateStoredProcedureCall(
        @NotNull StringBuilder sql, 
        @NotNull DBSProcedure proc, 
        @NotNull Collection<? extends DBSProcedureParameter> parameters,
        boolean castParams
    );
```

plotn: I edited EXISTING method and i see that you dont used it while developing it. Give an advise for my further activities ? I can only say about parameter i added (wheather null or not, but example still needed).
Why do you require code cleanup for the codebase you (or precessors) was created?

ShadelessFox: It's always nice to clean up an older code, especially when editing it anyways.

plotn: > We use annotations to specify function arguments and return type nullability: @NotNull and @nullable It can be done like the following in this case
> 
> ```
>     void generateStoredProcedureCall(
>         @NotNull StringBuilder sql, 
>         @NotNull DBSProcedure proc, 
>         @NotNull Collection<? extends DBSProcedureParameter> parameters,
>         boolean castParams
>     );
> ```

Please confirm (fix) the method declaration and add (missing) code examples of checking for 'nullability'.  thank you

E1izabeth: @ShadelessFox I don't see the reason to strongly require it. Can we pass it to QA please?

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialect.java
ShadelessFox: It's not a strict requirement.

E1izabeth:
@plotn please, pull latest changes from devel

E1izabeth:
Oh, sorry, it seems pulling latest changes from devel is not needed right now.

Matvey16:
Verified

PR #18146, opened 2022-11-08
#2316 Tab autocompletion

### LINKED ISSUE #2316: DBeaver - Press Tab to accept the autocomplete choice.
Hi,

I'm not sure if there is a settings we can set the Tab key as the keyword confirmation in the SQL Explorer?  I'm used to pressing Tab in Visual Studio IDE.  Right now, we need to press Enter, which is fine, but it would be nice to also use the Tab key.  Thanks!

E1izabeth:
It's the only working hack I found.
With the current dirty hack autocompletion choice will work by both tab and enter.


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorSourceViewer.java
```diff
@@ -50,6 +60,25 @@ void refreshTextSelection(){
     @Override
     protected StyledText createTextWidget(Composite parent, int styles) {
         StyledText textWidget = super.createTextWidget(parent, styles);
+        textWidget.addListener(ST.VerifyKey, new Listener() {
+            @Override
+            public void handleEvent(Event event) {
+                switch (event.type) {
+                    case ST.VerifyKey: {
+                        if (event.character == '\t') {
```
serge-rider: Lets make it configurable

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorSourceViewer.java
E1izabeth: image

uslss:
verified

serge-rider (approved):

PR #18158, opened 2022-11-08
dbeaver/dbeaver#18156 use confirmation dialog when deleting navigator nodes

### LINKED ISSUE #18156: Use the standard message box when deleting navigator nodes
Our confirmation dialog that appears when you try to detele navigator nodes looks nothing like our other confirmation dialogs. We need to use our standard confirmation dialog.

flambchap:


bundles/org.jkiss.utils/src/org/jkiss/utils/MutableBox.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * A simple mutable container that holds only one value.
+ *
+ * <p>
+ * Among other situations, it's helpful when you need to generify a method
+ * with an effectively final variable without suppressing unchecked cast warning.
+ *
+ * @param <T> type of value
+ */
+public class MutableBox<T> {
```
ShadelessFox: Why is this better than creating an array of a single element? Despite the fact that this class itself introduces an indirection, it will create additional overhead when using primitives by boxing them.

Additionally, it's only used in a single place. I don't think it worth it.

If you really want to use some kind of box type, then consider using `Reference<T>`

ShadelessFox (changes requested):

bundles/org.jkiss.utils/src/org/jkiss/utils/MutableBox.java
flambchap: You can use an `Object[1]` indeed, but it requires warning suppression in case of generic code like the newly introduced `syncExec(Supplier<T>)`. I'm not sure what you mean by indirection in this context (arrays themselves introduce indirection, don't they?). I'm also unsure what exact `Reference<T>` you mean.

E1izabeth: I'd prefer making this class to be private and nested if it's possible or avoid creating such utility class at all.

flambchap: OK then

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/dialogs/NavigatorNodesDeletionConfirmations.java
```diff
@@ -19,77 +19,99 @@
...
-    public static ConfirmNavigatorNodesDeleteDialog of(Shell shell, String title, String message, List<?> selectedObjects, NavigatorObjectsDeleter deleter) {
-        return new ConfirmNavigatorNodesDeleteDialog(shell, title, message, selectedObjects, deleter);
-    }
-
-    public static ConfirmNavigatorNodesDeleteDialog of(Shell shell, List<?> selectedObjects, NavigatorObjectsDeleter deleter) {
+    public static Reply confirm(
+        Shell shell,
+        Collection<?> selectedObjects,
+        @Nullable NavigatorObjectsDeleter deleter
+    ) {
```
ShadelessFox: Please specify nullability for this method and its parameters

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/dialogs/NavigatorNodesDeletionConfirmations.java
```diff
@@ -19,77 +19,99 @@
...
         String message = NLS.bind(node instanceof DBNLocalFolder ? UINavigatorMessages.confirm_local_folder_delete_message : UINavigatorMessages.confirm_entity_delete_message, node.getNodeType(), node.getNodeName());
-        return new ConfirmNavigatorNodesDeleteDialog(shell, title, message, selectedObjects, deleter);
+        return confirm(shell, title, message, selectedObjects, deleter);
+    }
+
+    public static Reply confirm(
+        Shell shell,
+        String title,
+        String message,
+        Collection<?> selectedObjects,
```
ShadelessFox: Same here

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/dialogs/NavigatorNodesDeletionConfirmations.java
```diff
@@ -19,77 +19,99 @@
...
+            .setPrimaryImage(DBIcon.STATUS_WARNING)
+            .setReplies(replies)
+            .setDefaultReply(Reply.CANCEL)
+            .setCustomArea(parent -> createCustomArea(parent, selectedObjects, deleter));
+        return UIUtils.syncExec(messageBoxBuilder::showMessageBox);
     }
 
-    @Override
-    protected Control createCustomArea(Composite parent) {
+    private static void createCustomArea(Composite parent, Collection<?> selectedObjects, @Nullable NavigatorObjectsDeleter deleter) {
```
ShadelessFox: Same here

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/dialogs/NavigatorNodesDeletionConfirmations.java
```diff
@@ -19,77 +19,99 @@
...
             for (NavigatorObjectsDeleter.Option option : deleter.getSupportedOptions()) {
-                createCheckbox(parent, option);
+                createCheckbox(parent, option, deleter);
             }
         }
-        return super.createCustomArea(parent);
     }
 
-    private void createObjectsTable(Composite parent) {
+    private static void createObjectsTable(Composite parent, Collection<?> selectedObjects) {
```
ShadelessFox: Same here

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/dialogs/NavigatorNodesDeletionConfirmations.java
```diff
@@ -125,7 +147,7 @@ private void createObjectsTable(Composite parent) {
         UIUtils.asyncExec(() -> UIUtils.packColumns(objectsTable, true));
     }
 
-    private void createDeleteContents(Composite parent) {
+    private static void createDeleteContents(Composite parent, @NotNull NavigatorObjectsDeleter deleter) {
```
ShadelessFox: Same here

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/dialogs/NavigatorNodesDeletionConfirmations.java
```diff
@@ -154,7 +176,11 @@ public void widgetSelected(SelectionEvent e) {
         );
     }
 
-    private void createCheckbox(Composite checkboxesComposite, NavigatorObjectsDeleter.Option option) {
+    private static void createCheckbox(
```
ShadelessFox, line 204: Same here

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -489,17 +490,14 @@ public static boolean confirmAction(@Nullable Shell shell, final String title, f
...
             .setMessage(message)
             .setReplies(Reply.YES, Reply.NO)
             .setDefaultReply(Reply.NO)
-            .setPrimaryImage(image)
-            .showMessageBox()
-        );
-
-        return reply[0] == Reply.YES;
+            .setPrimaryImage(image);
+        Reply reply = syncExec(builder::showMessageBox);
```
ShadelessFox, line 499: This variable looks redundant

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -1804,6 +1802,22 @@ public static <T> T syncExec(RunnableWithResult<T> runnable) {
...
+     * @param supplier code to execute
+     * @return result of computation
+     * @param <T> result type
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> T syncExec(Supplier<T> supplier) {
+        final Object[] obj = {null};
+        syncExec(() -> {
+            obj[0] = supplier.get();
+        });
```
ShadelessFox, line 1815: This block seems redundant

```suggestion
        syncExec(() -> obj[0] = supplier.get());
```

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -1804,6 +1802,22 @@ public static <T> T syncExec(RunnableWithResult<T> runnable) {
...
 
+    /**
+     * Executes a piece of code in the UI thread and retrieves the result of its calculation.
+     *
+     * @param supplier code to execute
+     * @return result of computation
+     * @param <T> result type
+     */
+    @SuppressWarnings("unchecked")
+    public static <T> T syncExec(Supplier<T> supplier) {
```
ShadelessFox: Please specify nullability

ShadelessFox (changes requested):

ShadelessFox:
A screenshot showing the difference would be nice to have

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
flambchap: Actually, no.  `() -> obj[0] = supplier.get()` returns a value, thus making this method fall into infinite recursion. What we need to do is to pass a `Runnable`. Therefore, we need braces.

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
flambchap: I believe it improves readability.

ShadelessFox: Dunno, maybe.

ShadelessFox (approved):

flambchap:
Screenshot 2022-11-08 at 02 30 32

LonwoLonwo (approved):

uslss:
verified

PR #18186, opened 2022-11-11
#17823 Add null count function to the calculation pane

### LINKED ISSUE #17823: Add null count function to the calculation pane
When exploring a column in a table, I make heavy use of the calculation pane to see basic stats (count, count distinct, sum, etc.). I love this quick view, but I really wish there was a function to count the number of nulls. 

Screen Shot 2022-10-05 at 3 09 19 PM

I would love a function to count the number of nulls in a column.

I can always query for the value.

**Additional context**

tchibatron:


ShadelessFox (approved):
Welcome and thanks for the contribution!

This PR looks really clean :)

E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/aggregate/FunctionCountNulls.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.data.aggregate;
+
+public class FunctionCountNulls implements IAggregateFunction {
+    private long count = 0;
+
+    @Override
+    public boolean accumulate(Object value, boolean aggregateAsStrings) {
+        if (value == null) {
```
ShadelessFox: I was thinking a lot and believe you would benefit from using `DBUtils#isNullValue` here.

The reason is that complex values, such as composites and arrays, aren't `null` but their contents is `null`, and that's what the proposed method is looking for.

You don't need to alter tests or anything, it will be a drop-in replacement.

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/aggregate/FunctionCountNulls.java
tchibatron: Done.

uslss:
verified

Camden-Lee:
Awesome! I really appreciate this feature enhancement. Thank you @tchibatron @ShadelessFox

Issue #5527, opened 2019-03-19
MSSQL Clustered ColumnStore indices can not be queried


roa-nyx:
I have not found any open issue that seems relevant to this.

#### System information: 
Windows 10 (latest updates as of March 18, 2019)
DBeaver 6.0.0.201903101945
Java (no idea)
Additional extensions: mssql driver

#### Connection specification:
Microsoft SQL Server 2017 (RTM-CU5) (KB4092643) - 14.0.3023.8 (X64)   Mar  2 2018 18:24:44   Copyright (C) 2017 Microsoft Corporation  Web Edition (64-bit) on Windows Server 2016 Datacenter 10.0 <X64> (Build 14393: ) (Hypervisor) 
Driver name: Microsoft JDBC Driver for SQL Server
No tunnels or proxies

#### Describe the problem you're observing:
I can't query tables that have a clustered columnstore index.

```
SELECT DISTINCT [FieldA] FROM [ColStoreTable]
/* results in:
Query execution failed

Reason:
SQL Error [35370] [S0001]: Cursors are not supported on a table which has a clustered columnstore index.
*/
```

#### Steps to reproduce, if exist:
Create a table in an MS-SQL server with a clustered columnstore index, and try to query it.

It seems to me that DBeaver uses a cursor to iterate through results, but this isn't allowed for clustered columnstore indexes. Not sure how this can be addressed.

#### Include any warning/errors/backtraces from the logs
I'm skipping the logs here as I feel there is enough information here and I don't want to try to anonymize/sanitize any potential leakages from the log files however if you feel you do end up needing the log files please let me know.

wgrusznis:
I have the same problem.
I tried the latest Microsoft JDBC driver without any success. 
This makes DBeaver unusable with clustered columnstore indexed tables.

as a workaround I use 
SELECT DISTINCT top 10 [FieldA] FROM [ColStoreTable]

I always thought 'Use SQL to limit fetch size' does exactly this but it seems I was wrong.

marcushenningsen:
I can confirm this problem. My workaround is to use SSMS, but I'd much prefer to consolidate my SQL work in DBeaver

marcoracer:
I can confirm this problem. We use Azure SQL Database and when we create a clustered columnstore index and query it we get:

Query execution failed

Reason:
SQL Error [35370] [S0001]: Cursors are not supported on a table which has a clustered columnstore index.

higorhi:
[...]

PR #18210, opened 2022-11-14
#18071 add support of COLUMNSTORE indexes for the better statement li…

### LINKED ISSUE #18071: SQL Server: max rows works incorrectly
### Description

We use setMaxRows for all SELECT statements.
SQL Server driver (old and new ones) uses  SET ROWCOUNT. Which affects all SELECT-like queries. Including INSERT FROM SELECT which can be executed in referenced stored procedures.
If we not use setMaxRows then we use TOP X to limit fetch size. This leads to the same result.

Maybe we should disable setMaxRows execution for SQL Server (and Sybase).

Possible soltion: disable query limit default.

### DBeaver Version

22.2

### Operating System

Any

### Database and driver

- SQ LServer
- Azure
- Sybase

LonwoLonwo:
…mits handling

This all is a continuation of these tickets: #5527 and ##9787

2022-11-14 17_43_52-DBeaver Ultimate 22 2 4 - SimpleTable4

2022-11-14 17_53_32-DBeaver Ultimate 22 2 4 - SimpleTable6

2022-11-14 17_53_48-DBeaver Ultimate 22 2 4 - SimpleTable4

Also please check with/without this setting

2022-11-14 17_54_47-DBeaver Ultimate 22 2 4 - SimpleTable3

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
@@ -535,11 +535,13 @@ public class SQLServerSchema implements DBSSchema, DBPSaveableObject, DBPQualifi
                     indexType = DBSIndexType.OTHER;
                     break;
             }
+            boolean isColumnstoreIndex = indexTypeNum == 5 || indexTypeNum == 6;
             return new SQLServerTableIndex(
                 parent,
                 indexName,
                 indexType,
-                dbResult);
+                dbResult,
+                isColumnstoreIndex);
         }
 
         @Nullable
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
@@ -484,6 +484,15 @@ public class SQLServerSchema implements DBSSchema, DBPSaveableObject, DBPQualifi
      * Index cache implementation
      */
     static class IndexCache extends JDBCCompositeCache<SQLServerSchema, SQLServerTableBase, SQLServerTableIndex, SQLServerTableIndexColumn> {
+        private static final int INDEX_TYPE_HEAP = 0;
+        private static final int INDEX_TYPE_CLUSTERED_ROWSTORE = 1;
+        private static final int INDEX_TYPE_NONCLUSTERED_ROWSTORE = 2;
+        private static final int INDEX_TYPE_XML = 3;
+        private static final int INDEX_TYPE_SPATIAL = 4;
+        private static final int INDEX_TYPE_CLUSTERED_COLUMNSTORE = 5;
+        private static final int INDEX_TYPE_NONCLUSTERED_COLUMNSTORE = 6;
+        private static final int INDEX_TYPE_NONCLUSTERED_HASH = 7;
+
         IndexCache(TableCache tableCache)
         {
             super(tableCache, SQLServerTableBase.class, "table_name", "name");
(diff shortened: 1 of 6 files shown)


ShadelessFox (approved):

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
```diff
@@ -535,11 +535,13 @@ protected SQLServerTableIndex fetchObject(JDBCSession session, SQLServerSchema o
                     indexType = DBSIndexType.OTHER;
                     break;
             }
+            boolean isColumnstoreIndex = indexTypeNum == 5 || indexTypeNum == 6;
```
E1izabeth: Please don't add magic values. Let's define a constants for them.
I see that we have already got a switch with such magic values, but let's stop it.

LonwoLonwo: I added it here. But I'm not sure that we need to move all documentations stuff in our code :(

E1izabeth: I'd like to have enum.

LonwoLonwo: What for?

ShadelessFox: Introducing an enum would be overkill since these values are only used in a single place.

E1izabeth: It's hard to support. I should always return back to the comment and read it again to know what's the meaning for each number.
Besides that, I don't understand what is this code for when I read `indexTypeNum  == 5`, it's hard to understand.

E1izabeth: These values have been already used in two places.

E1izabeth: okay, at least just a constants...

LonwoLonwo: We can create SQL Server extra index types, but in this case, my condition will be not
`index.isColumnStore` 
but 
`index.getIndexType == SQLServerIndexType.COLUMNSTOREClustered || index.getIndexType == SQLServerIndexType.COLUMNSTORENonClustered`

And these extra indexes need to be also like clustered/nonclustered children... I think it is overcomplicated.

indexTypeNum == 5 - Now in the comment. But I still think from my side that documentation is always our friend. And this is not the best idea to keep this sometimes outdated info in DBeaver.

ShadelessFox: Perhaps she meant introducing private constants. Let me do that for you:

```java
private static final int INDEX_TYPE_HEAP = 0; 
private static final int INDEX_TYPE_CLUSTERED_ROWSTORE = 1;
private static final int INDEX_TYPE_NONCLUSTERED_ROWSTORE = 2;
private static final int INDEX_TYPE_XML = 3;
private static final int INDEX_TYPE_SPATIAL = 4;
private static final int INDEX_TYPE_CLUSTERED_COLUMNSTORE = 5;
private static final int INDEX_TYPE_NONCLUSTERED_COLUMNSTORE = 6;
private static final int INDEX_TYPE_NONCLUSTERED_HASH = 7;
```

It's not overkill since these would be inlined after compilation.

E1izabeth: Yes, I meant what @ShadelessFox proposed.

E1izabeth: Thank you.

LonwoLonwo: Still, I think that this is overcomplicated.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
```diff
@@ -523,23 +529,35 @@ protected SQLServerTableIndex fetchObject(JDBCSession session, SQLServerSchema o
...
             DBSIndexType indexType;
+            // Type of index:
+            // 0 = Heap
+            // 1 = Clustered rowstore (B-tree)
+            // 2 = Nonclustered rowstore (B-tree)
+            // 3 = XML
+            // 4 = Spatial
+            // 5 = Clustered columnstore index. Applies to: SQL Server 2014 (12.x) and later.
+            // 6 = Nonclustered columnstore index. Applies to: SQL Server 2012 (11.x) and later.
+            // 7 = Nonclustered hash index. Applies to: SQL Server 2014 (12.x) and later.
```
ShadelessFox: This comment is now redundant, please remove it

LonwoLonwo: No. It is not.

ShadelessFox: It is, it replicates constants

E1izabeth: This comment is useful if we assign each piece of comment to the corresponding constant.

uslss:
verified

PR #18222, opened 2022-11-15
#16238 Preselect profile, allow copying existing profiles

### LINKED ISSUE #16238: Pre-select the correct SSH tunnel profile when entering the edit window
Have at least two SSH profiles.
Have a connection using a profile that is not listed as the first one.
Click the "edit" button on profile selection in a Database connection setting window.

Expected (would be nice to have): the profile to be edited gets also pre-selected in the profile edit window.
Actual bahviour: the first profile in the list is selected.

(It happened to me that I changed the wrong profile because I wasn't aware that not the profile was pre-selected for which I opened the profile edit window)

ShadelessFox:


E1izabeth (approved):

LonwoLonwo (approved):

uslss:
verified

Issue #17554, opened 2022-09-05
MySQL username isn't with proper back quotes when dragging/copying to SQL editor


yoni-yad2:
### Description

When copying a MySQL user from DB navigator to SQL editor it shows as:
`username@%`

or when dragging as:

`` `username@%` ``

instead of the correct way of:
`` `username`@`%` ``

### DBeaver Version

Community Edition Version 22.2.0.2022090406

### Operating System

Windows 10 64 bit

### Database and driver

MySQL / MariaDB

LonwoLonwo:
Thanks for the bug report @yoni-yad2

LonwoLonwo:
Fixed.

PR #18228, opened 2022-11-16
#17554-Fix-for-MySQL-username-with-proper-back-quotes

amargulaganji:
Fix for MySQL username isn't with proper back quotes when dragging/copying to SQL editor #17554

[#17554](https://github.com/dbeaver/dbeaver/issues/17554)

17554-Test

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLUser.java
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLUser.java
@@ -95,7 +95,7 @@ public class MySQLUser implements DBAUser, DBARole, DBPRefreshableObject, DBPSav
     @Override
     @Property(viewable = true, order = 1)
     public String getName() {
-        return userName + "@" + host;
+        return "`" + userName + "`@`" + host + "`";
     }
 
     public String getUserName() {
--- a/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLUser.java
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLUser.java
@@ -95,7 +95,7 @@ public class MySQLUser implements DBAUser, DBARole, DBPRefreshableObject, DBPSav
     @Override
     @Property(viewable = true, order = 1)
     public String getName() {
-        return "`" + userName + "`@`" + host + "`";
+        return userName + "@" + host;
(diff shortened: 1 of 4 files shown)


LonwoLonwo:
Hello @amargulaganji 
Thanks for the contribution.
Yes, your fix helps with user's names.
But `quoteIdentifier` is a base method. What if `str` will be a table name? Like `R@se`? Or column name? Or catalog name? With symbol @ inside.

amargulaganji:
Hi @LonwoLonwo,
Thank you for your observation.
As you noticed, changes to `quoteIdentifier` method will affect the table and column names.
So removed changes from `quoteIdentifier`. Made changes that will make difference to the user's name only.

Test Result:
Test-user-table-column-name

LonwoLonwo:
That's better.

2022-11-17 14_04_37-DBeaver 22 2 5 - _localhost 3_ Script-22

But now the user's list looks something like this. Strange that I can't see it on your video.

amargulaganji:
@LonwoLonwo Made changes that will not affect the user's list.

Test Result :
Test-username-ui

LonwoLonwo:
Oh, no :( You can not fix the local problem on the general level

ShadelessFox (changes requested):
We don't want to handle such a weird case on the top level.

Instead, just implement the `DBPQualifiedObject` interface for the `MySQLUser` class.

amargulaganji:
Thank you @ShadelessFox  for your suggestion. It helped me to fix the issue locally.

Please review my changes.

Test Result:
17554-Test-Result

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLUser.java
```diff
@@ -309,4 +311,9 @@ public DBSObject refreshObject(@NotNull DBRProgressMonitor monitor) throws DBExc
         grants = null;
         return this;
     }
+
+    @Override
+    public String getFullyQualifiedName(DBPEvaluationContext context) {
+        return "`" + userName + "`@`" + host + "`";
```
ShadelessFox: The `userName` might contain backquotes, please use `DBUtils#getQuotedIdentifier` with `quoteAlways` set to `true`

ShadelessFox (changes requested):

ShadelessFox (approved):

E1izabeth (approved):

amargulaganji:
Hi @LonwoLonwo , Could you please review and merge?

E1izabeth:
@amargulaganji your PR is waiting for testing.

Matvey16:
Verified

PR #18230, opened 2022-11-16
dbeaver/pro#139 Update licenses for CE and EE products drivers

LonwoLonwo:
list of drivers: https://github.com/dbeaver/pro/issues/139#issuecomment-1315607841


plugins/org.jkiss.dbeaver.ext.h2/plugin.xml
```diff
@@ -87,7 +91,8 @@
                         dialect="h2">
 
                     <file type="jar" path="maven:/com.h2database:h2:RELEASE[2.1.210]" bundle="!drivers.h2_v2"/>
-                    <file type="jar" path="drivers/h2_v2" bundle="drivers.h2_v2"/>
```
E1izabeth, line 90: Was it removed intentionally?

LonwoLonwo: Yep. It was my previous mistake. We do not have this folder in drivers.

E1izabeth (approved):

ShadelessFox (approved):

uslss:
CE verified

PR #18250, opened 2022-11-17
#17040 check temp file existence

### LINKED ISSUE #17040: Out Of Memory
#### System information: 
- Operating system (distribution) and version : Centos 7.4
- DBeaver version : 21.2.2
- Additional extensions

- Database name and version : Oracle
- Driver name 

When previewing GIF files, it takes up too much memory, resulting in out of memory

1. Create a new Oracle connection and create a table in which the column type is blob.

2. On the data tab page, add a row, right click -->edit-->load from file, Select a GIF file with a size of 14MB.

3. Use free -h to check that the machine memory is 6.2GB.

3. Click the value box on the right to view the preview content.

4. Use free -h to check that the machine memory is 4.9GB.

The screenshot is as follows:

1657507113911_BD8F9CFC-0E4D-4b57-A1A4-F59E461583AA

1657506936787_708ABDAE-45BD-462b-801B-470615E6F0E9

1657506976254_10939A02-5538-491d-8CBB-5D48803C810C

Destrolaric:
Closes #17040


uslss:
Text data isn't shown in value panel, if there are text data and images in the same column
dbeaver_lewtzgZwiF
dbeaver_CTHmGv4S2H

PR #18268, opened 2022-11-20
Tree filter enchancement fix#9444

plotn:
Unfortunally there were NPE, appeared during refactoring. Now it is Ok


plotn:
Sorry, wrong files added

Issue #9444, opened 2020-08-04
Tree filter enchancement


plotn:
... as shown
![filter_ext](https://user-images.githubusercontent.com/11477081/89275647-364b0980-d64b-11ea-9362-0f033b8075a9.png)

kseniiaguzeeva:
Thank you for trhe suggestion.

goxr3plus:
Very good idea!

PR #18269, opened 2022-11-20
#9444 NPE fixed

plotn:
Unfortunally, there appeared NPE during refactoring. Now it is ok.


plotn:
@serge-rider , merge please, because devel is unstable when objects in tree searched

E1izabeth:
Please provide NPE stack trace, which you are trying to fix.

serge-rider:
It happens every time you type a dot after a schema name. Somehow we missed this issue.
@plotn thanks for the fix

plotn:
thank you collegues, and I am sorry for this

PR #18277, opened 2022-11-21
#16756 Big scripts run performance

### LINKED ISSUE #16756: Dbeaver is very slow to run big SQL scripts
#### System information: 
- Operating system (distribution) and version: Microsoft Windows [Version 10.0.22000.675]
- DBeaver version: Version 21.3.2.202112271055

- Database name and version: Any, they're all the same

I'm often loading big piles of data into a database. For some databases, I don't have any tools to expedite this process, so I'm using Dbeaver. The loading takes forever in some cases, e.g. 20 minutes.

Use any of these sakila database scripts, e.g. the SQL Server one:
https://github.com/jOOQ/sakila

- https://github.com/jOOQ/sakila/blob/main/sql-server-sakila-db/sql-server-sakila-schema.sql (reasonably fast)
- https://github.com/jOOQ/sakila/blob/main/sql-server-sakila-db/sql-server-sakila-insert-data.sql (very slow)

A VisualVM session shows that a large amount of time is spent in resizing some control:

image

This seems to be a completely avoidable thing, at least for large scripts. I suspect that this is for this scrolling animation? I think I could do without that, even for smaller scripts

The following ZIP file contains a VisualVM snapshot, if you want to look at that without running the above SQL script: [snapshot-1654761133115.zip](https://github.com/dbeaver/dbeaver/files/8868167/snapshot-1654761133115.zip)

Even opening the large file takes a long time (~20s), though I could live with that if the execution was faster. Here's what Dbeaver is doing when opening the large file:

image

And another VisualVM snapshot file: [snapshot-1654761535592.zip](https://github.com/dbeaver/dbeaver/files/8868195/snapshot-1654761535592.zip)

Again, I think that effort could be avoided for this use-case? I don't care for editing the file much (though it should still be possible in case it has errors, but perhaps with less tooling). I just want to run it.

Anyway, I don't know how to best fix this:

- By offering simpler additional editors with less features, so they're faster
- By optimising slow stuff
- By removing unnecessary features, such as the auto-scrolling while executing

I trust that you know this better than me

E1izabeth:
- new preference for big script size definition
image

- automatically disable sql syntax parser for big scripts
image
image

- propose to disable fetch result set for big scripts
image

(`Remember my choice for this editor` is not saved between launches)

When user opens big file, the red lamp should appear, `Disable sql syntax parser` option should be disabled so user can't turn on/off it which shows that it has been already applied because of big script, word occurrences, most syntax highlighting rules, folding autocompletion should not work for big scripts.
To apply preference Big file size preference changes script should be reopened.
When user run big script the confirmation asking to disable result set fetching should appear.


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/internal/SQLEditorMessages.properties
```diff
@@ -346,5 +347,15 @@ sql_generator_title_text = Generated SQL ( {0} )
...
+sql_editor_prefs_disable_services_text = Disable SQL syntax parser
+sql_editor_prefs_disable_services_tip = Overrides settings for code completion, foldings and word occurrences highlighting\n(Automatically applied for large script files)
+
+sql_editor_prefs_script_advanced_settings = Advanced settings
+sql_editor_prefs_script_disable_sql_syntax_parsing_for_scripts_bigger_than = Disable SQL syntax parser for files bigger than (KB)
+sql_editor_confirm_no_fetch_result_for_big_script_title = Preparing to execute the script
+sql_editor_confirm_no_fetch_result_for_big_script_question = Large script execution may be dramatically slow while display in results of each statement.\n\nDo you want to disable result set fetching for this script execution?
+sql_editor_confirm_no_fetch_result_for_big_script_yes = Yes
+sql_editor_confirm_no_fetch_result_for_big_script_no = No
+sql_editor_confirm_no_fetch_result_for_big_script_remember = Remember my choice for this editor
```
ShadelessFox: If it saves that preference just for the current session, can we express it in the message?

E1izabeth: I'm able to think out something like `Remember my choice for this editor in the current session`, but it seems cumbersome for me. What do you think?

ShadelessFox: I don't think its cumbersomness is too bad, since it explicitly tells that it does.

Maybe just rephrase it a bit:
> Don't show this message again for this editor till the end the session

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/internal/SQLEditorMessages.properties
```diff
@@ -347,8 +347,7 @@ sql_generator_title_text = Generated SQL ( {0} )
 sql_generator_no_obj_container_text = Unable to get container
 sql_generator_no_ddl_text = No DDL is available for the result set
 sql_generator_nonsql_text = The container does not support the standard DDL structure.
-sql_editor_separate_connection_no_editor_or_ds_selected = No editor or data source selected
-
+sql_editor_separate_connection_no_editor_or_ds_selected = No editor or data source selecteds
```
ShadelessFox:
```suggestion
sql_editor_separate_connection_no_editor_or_ds_selected = No editor or data source selected
```

ShadelessFox (changes requested):

ShadelessFox (approved):

uslss:
verified

Issue #12070, opened 2021-04-11
[mysql] incorrect handling of stored procedure and function with the same name


buttha:
- Operating system (distribution) and version: Debian 10
- DBeaver version: 21.0.2
- Database name and version: mysql 8.0.23
- Driver name: MySQL

If there are one stored procedure and one function with the same name, you could loose function while editing the procedure.

1) create "test" PROCEDURE

```
DELIMITER $$
$$
CREATE PROCEDURE testdb.test()
BEGIN
END$$
DELIMITER ;
```

2) create "test" FUNCTION

```
DELIMITER $$
$$
CREATE FUNCTION testdb.test()
RETURNS INT
DETERMINISTIC
BEGIN
	RETURN 0;
END$$
DELIMITER ;
```

3) open the procedure, edit and save. CTRL+S generates the correct code:

```
DROP PROCEDURE IF EXISTS testdb.test;

DELIMITER $$
$$
CREATE DEFINER=`root`@`localhost` PROCEDURE `testdb`.`test`()
BEGIN
	# test
END$$
DELIMITER ;
```

4) edit the procedure again and press CTRL+S.... boom:

DROP **FUNCTION** IF EXISTS testdb.test;

DELIMITER $$
$$
CREATE DEFINER=`root`@`localhost` PROCEDURE `testdb`.`test`()
BEGIN
	# test 1
END$$
DELIMITER ;

As you can see the second time you save, there is a DROP FUNCTION and CREATE PROCEDURE, so "test" function is deleted.

uslss:
thanks for the report

Matvey16:
Verified

PR #18282, opened 2022-11-21
#17740 redundant method removed for the better cache search work

### LINKED ISSUE #17740: MySQL DB object "not found in catalog" error when trying to navigate it using Ctrl+click
### Description

When navigating the DB object(in my case, stored procedure) using the Ctrl+click shortcut, the "not found in catalog" error occurs.
Please note that this error occurs only when this action is performed the second and all subsequent times, while the first time, the error does not occur.

### DBeaver Version

DBeaver Community 22.2.1.202209251341

### Operating System

Windows 10
Windows Server 2019 10.0

### Database and driver

1. MySQL 8.0.30, MySQL 5.7.38 enterprise
2. MySQL Connector/J mysql-connector-java-8.0.30 

### Steps to reproduce

1. Connect to MySQL
2. Try to navigate to some procedure using the Ctrl+Click on its name on the SQL Editor window
2. Repeat the previous action
3. See an error:
Untitled
Untitled2
Untitled3

### Additional context

[.log](https://github.com/dbeaver/dbeaver/files/9654385/default.log)

LonwoLonwo:
`getResultType()` is always `null` for Procedures: 
https://dev.mysql.com/doc/refman/5.7/en/information-schema-routines-table.html

If `getProcedureType()` will be `DBSProcedureType.FUNCTION` - search by procedures by procedure name (in MySQLCatalog) will not work.

So, I do not see a reason to keep this method here.

Qriteries:

- [x] https://github.com/dbeaver/dbeaver/issues/12070 - is steel not broken
- [x] Procedures can be founded by ctrl+click from SQL Editor
- [x] Functions without parameters can be launched by ctrl+click from SQL Editor
- [x] Functions with parameters can be launched by ctrl+click from SQL Editor


E1izabeth (approved):

serge-rider:
This method is probably incorrect.
But what about functions?

This function was added to resolve https://github.com/dbeaver/dbeaver/issues/12070 . Does it still work with your fix?

LonwoLonwo:
> But what about functions?

Unique routine names are always helpful for a better search. But in the case of MySQL - I do not see any reason to use it. Because MySQL Procedures and functions can not be created with equal names but with different parameters. Can not be overloaded.

2022-11-22 12_13_19-DBeaver Ultimate 22 2 5 - _MySQL 5 5_ Script-83

So in the case of MySQL database - procedures/functions names are already unique without parameters.

LonwoLonwo:
> This function was added to resolve https://github.com/dbeaver/dbeaver/issues/12070 . Does it still work with your fix?

Yes, I check it.

ShadelessFox (approved):

uslss:
#12070 - error appears on procedure opening
```
eclipse.buildId=unknown
java.version=11.0.12
java.vendor=Eclipse Foundation
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en
Command-line arguments:  -os win32 -ws win32 -arch x86_64

org.jkiss.dbeaver.model
Error
Fri Nov 25 11:42:59 CET 2022
ArrayIndexOutOfBoundsException

java.lang.ArrayIndexOutOfBoundsException
```

uslss:
error and empty SQL Editor on the attempt to execute a function or procedure using UI
dbeaver_KEbvzRMvTQ
[exec.log](https://github.com/dbeaver/dbeaver/files/10091786/exec.log)

uslss:
#18334

uslss:
verified

PR #18285, opened 2022-11-21
dbeaver/pro#1249 Fix zip slip vulnerability

E1izabeth:
Please, check that
- project Import from zip works
- Oracle Wallet authentication works
- load zip driver libraries works


uslss:
verified

Issue #18148, opened 2022-11-08
Error when import csv file with columns tables containing single-quote 


Maxdeneb:
### Description

From version 22.2.3 it is impossible to import CSV files with columns containing single quotes (Es.: "Modalita' di Pagamento"). Below is the error log

.....
!ENTRY org.jkiss.dbeaver.model 4 0 2022-11-08 12:38:32.332
!MESSAGE Error transfering data from file.csv to schema.table [Existing]
!SUBENTRY 1 org.jkiss.dbeaver.model 4 0 2022-11-08 12:38:32.332
!MESSAGE Can't insert row
!SUBENTRY 2 org.jkiss.dbeaver.model 4 0 2022-11-08 12:38:32.332
!MESSAGE Errore SQL [1054] [42S22]: (conn=294915) Unknown column 'Modalita'''' di Pagamento' in 'field list'
!STACK 0
org.jkiss.dbeaver.model.sql.DBSQLException: Errore SQL [1054] [42S22]: (conn=294915) Unknown column 'Modalita'''' di Pagamento' in 'field list'
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCStatementImpl.executeStatementBatch(JDBCStatementImpl.java:154)
	at org.jkiss.dbeaver.model.impl.data.ExecuteBatchImpl.flushBatch(ExecuteBatchImpl.java:290)
	at org.jkiss.dbeaver.model.impl.data.ExecuteBatchImpl.processBatch(ExecuteBatchImpl.java:212)
	at org.jkiss.dbeaver.model.impl.data.ExecuteBatchImpl.execute(ExecuteBatchImpl.java:81)
	at org.jkiss.dbeaver.tools.transfer.database.DatabaseTransferConsumer.lambda$0(DatabaseTransferConsumer.java:429)
	at org.jkiss.dbeaver.model.exec.DBExecUtils.tryExecuteRecover(DBExecUtils.java:173)
	at org.jkiss.dbeaver.tools.transfer.database.DatabaseTransferConsumer.insertBatch(DatabaseTransferConsumer.java:427)
	at org.jkiss.dbeaver.tools.transfer.database.DatabaseTransferConsumer.fetchEnd(DatabaseTransferConsumer.java:483)
	at org.jkiss.dbeaver.tools.transfer.stream.importer.DataImporterCSV.runImport(DataImporterCSV.java:288)
[...]

PR #18287, opened 2022-11-22
#17914 Fix excess quoting when renaming target entity/attribute name

### LINKED ISSUE #17914: Some symbols( ` , ") in columns name are repeated twice during data import to new table
### Description

Some symbols( ` , ") in columns name are repeated twice during data import to a new table if the column name was changed during the import.

was founded during the check of #17619 

### DBeaver Version

22.2.2

### Operating System

win10

### Database and driver

mysql 5.7

### Steps to reproduce

1. Start import from csv to new table
2. Change some column name by adding, e.g. grave accent - We can see that grave accent is doubled in the mapping dialog
dbeaver_T1KRfjwi11

3. Check table DDL - there are four grave accent in DDL
imageimporting

### Additional context

_No response_

ShadelessFox:


LonwoLonwo:
And it still not working for this case #18148

2022-11-22 12_36_43-Configure metadata structure

[one_column_with_header.csv](https://github.com/dbeaver/dbeaver/files/10066043/one_column_with_header.csv)

ShadelessFox:
Turns out, we don't need to ever escape identifiers in the data transfer wizard. They will be escaped by the processor.

E1izabeth:
Maybe we want to write unit tests also?

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
```diff
@@ -206,13 +206,24 @@ public boolean mustBeQuoted(String str, boolean forceCaseSensitive) {
...
     @Override
     public String escapeString(String string) {
-        return string.replace("'", "''").replace("`", "``").replaceAll("\\\\(?![_%?])", "\\\\\\\\");
+        return escapeString(string, null);
+    }
+
+    @NotNull
+    protected String escapeString(@NotNull String string, @Nullable String[] quotes) {
+        if (quotes != null) {
+            string = string.replace(quotes[0], quotes[0] + quotes[0]);
```
E1izabeth, line 221: I'm not sure about the idea behind this code, but maybe you wanted something like this?
```suggestion
            string = string.replace(quotes[0], quotes[0] + quotes[1]);
```

ShadelessFox:
All changes were made within UI bundles. In any case, I'm not sure how to test it.

LonwoLonwo:
> All changes were made within UI bundles.

This is not true.

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
```diff
@@ -206,13 +206,24 @@ public boolean mustBeQuoted(String str, boolean forceCaseSensitive) {
...
+
     @NotNull
     @Override
     public String escapeString(String string) {
-        return string.replace("'", "''").replace("`", "``").replaceAll("\\\\(?![_%?])", "\\\\\\\\");
+        return escapeString(string, null);
+    }
+
+    @NotNull
+    protected String escapeString(@NotNull String string, @Nullable String[] quotes) {
```
LonwoLonwo, line 219: You can at least create a dialect test if you do not know how to make a datatransfer test. We already have PostgreDialectTest and OracleDialectTest. Please look at them.

LonwoLonwo (changes requested):

serge-rider:
We need unit tests for that

test/org.jkiss.dbeaver.ext.mysql.test/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialectTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class MySQLDialectTest {
+    @Test
+    public void quoteStringTest() {
+        final MySQLDialect dialect = new MySQLDialect();
+
+        Assert.assertEquals("`a ' b`", dialect.getQuotedIdentifier("a ' b", false, true));
+        Assert.assertEquals("`a `` b`", dialect.getQuotedIdentifier("a ` b", false, true));
```
E1izabeth, line 31:
```suggestion
        Assert.assertEquals("`a ' b`", dialect.getQuotedIdentifier("a ' b", false, true));
        Assert.assertEquals("`a `` b`", dialect.getQuotedIdentifier("a ` b", false, true));
        Assert.assertEquals("a b", dialect.getQuotedIdentifier("a b", false, true));
```

ShadelessFox: 👍🏻

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
ShadelessFox: Done

E1izabeth (approved):

LonwoLonwo (approved):

uslss:
verified

PR #18298, opened 2022-11-22
dbeaver/pro#1272 use "int"/"integer" synonyms for the better data typ…

LonwoLonwo:
…es mapping

2022-11-22 20_13_57-143 453 106 - AnyDesk

Qriteries:

- [ ] The XLSX file can be imported in the new table with the "INT"/"INTEGER" data type for the new numeric column
- [ ] Tables with INT columns can be imported into other databases. INT data type will be not converted or converted to INTEGER.
- [ ] Tables with INTEGER columns can be imported into other databases. The INTEGER data type will be not converted or converted to INT.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
@@ -48,6 +48,13 @@ public final class DBStructUtils {
 
     private static final Log log = Log.getLog(DBStructUtils.class);
 
+    private static final String INT_DATA_TYPE = "int";
+    private static final String INTEGER_DATA_TYPE = "integer";
+    private static final String FLOAT_DATA_TYPE = "float";
+    private static final String REAL_DATA_TYPE = "real";
+    private static final String DOUBLE_DATA_TYPE = "double";
+    private static final String TEXT_DATA_TYPE = "text";
+
     @Nullable
     public static DBSEntityReferrer getEnumerableConstraint(@NotNull DBRProgressMonitor monitor, @NotNull DBDAttributeBinding attribute) throws DBException {
         DBSEntityAttribute entityAttribute = attribute.getEntityAttribute();
@@ -317,7 +324,7 @@ public final class DBStructUtils {
                 // It seems this data type has modifiers. Try to find without modifiers
                 dataType = dataTypeProvider.getLocalDataType(SQLUtils.stripColumnTypeModifiers(typeName));
             }
-            if (dataType == null && typeNameLower.equals("double")) {
+            if (dataType == null && typeNameLower.equals(DOUBLE_DATA_TYPE)) {
                 dataType = dataTypeProvider.getLocalDataType("DOUBLE PRECISION");
                 if (dataType != null) {
                     typeName = dataType.getTypeName();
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
```diff
@@ -351,41 +353,53 @@ public static String mapTargetDataType(DBSObject objectContainer, DBSTypedObject
                             }
                         }
                         if (targetType == null) {
-                            if (typeNameLower.contains("float") ||
-                                typeNameLower.contains("real") ||
+                            String intDataType = "int";
+                            String integerDataType = "integer";
+                            String floatDataType = "float";
+                            String realDataType = "real";
```
ShadelessFox: These are not being reassigned anywhere. Please make them `static final` constants.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
```diff
@@ -317,7 +317,8 @@ public static String mapTargetDataType(DBSObject objectContainer, DBSTypedObject
                 // It seems this data type has modifiers. Try to find without modifiers
                 dataType = dataTypeProvider.getLocalDataType(SQLUtils.stripColumnTypeModifiers(typeName));
             }
-            if (dataType == null && typeNameLower.equals("double")) {
+            String doubleDataType = "double";
```
ShadelessFox: Same here

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
```diff
@@ -340,6 +341,7 @@ public static String mapTargetDataType(DBSObject objectContainer, DBSTypedObject
                 if (!possibleTypes.isEmpty()) {
                     // Try to get any partial match
                     targetType = possibleTypes.get(typeNameLower);
+                    String textDataType = "text";
```
ShadelessFox: Same here

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
LonwoLonwo: 2022-11-23 11_37_54-143 453 106 - AnyDesk

And I do not see any reason to move as Util class variables.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
ShadelessFox: It's about code smell. No one does that, everyone use class constants.

And the second reason is that local variable access is not optimized. Turning these into constants will make the compiler inline them without any runtime cost. Of course if wouldn't affect the performance that much, but anyways.

LonwoLonwo: I will repeat it - Util class is not an appropriate place for the class constants. It is a commonplace. It has many not bounded methods.

ShadelessFox (approved):

Matvey16:
Verified

PR #18303, opened 2022-11-22
#18274 Bring resultset fetch size back

### LINKED ISSUE #18274: Unable to set result fetch size in 22.2.5
### Description

Unable to set result fetch size in 22.2.5
image
Before it was like:
image

Maybe there is a setting somewhere but I can't find it - anyway this possibility to set fetch size on the fly is very important and handy to me.

### DBeaver Version

CE 22.2.5

### Operating System

win 10

### Steps to reproduce

Just select on table with more than 200 rows

serge-rider:


uslss:
The fetch size works quite strangely for now.
It is empty on table/result set loading and the value that was entered for the current table, applies to the next open table

https://user-images.githubusercontent.com/31996417/203540160-d9282410-cac2-4dd6-b343-825b484fb013.mp4

serge-rider:
Fixed

uslss:
verified

PR #18304, opened 2022-11-22
#17784 Replace path with virtualized to show resource in file explorer for win store app

### LINKED ISSUE #17784: Windows App "open file location" doesnt work
### Description

I am at a loss where to find my script files from the windows app. When you are editing a script and right-click on the tab header, you get the option to "got to file location" but that results in an error. I think this is a problem with the windows app exclusively.

The folder that is referred to in the documentation "%APPDATA%\DBeaverData\"  does not exist on my machine. A file search for "workspace6" also yields nothing.

### DBeaver Version

Windows App Comm 22.1.0

### Operating System

Windows 10

### Database and driver

Postgres

### Steps to reproduce

Download windows app
Create an sql script
save the script
now right click on the tab header, choose "go to file location"
Error is displayed

### Additional context

_No response_

E1izabeth:
Please, check that `Show in file explorer` command works correct.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.app.winstore/src/org/jkiss/dbeaver/winstore/WinStoreEnvPathMapper.java
+++ b/plugins/org.jkiss.dbeaver.ui.app.winstore/src/org/jkiss/dbeaver/winstore/WinStoreEnvPathMapper.java
@@ -16,43 +16,44 @@
  */
 package org.jkiss.dbeaver.winstore;
 
-import java.nio.file.Path;
-
 import org.jkiss.code.NotNull;
 import org.jkiss.dbeaver.Log;
 import org.jkiss.dbeaver.runtime.IEnvironmentPathMapper;
 import org.jkiss.dbeaver.utils.RuntimeUtils;
 
+import java.nio.file.Path;
+
 public class WinStoreEnvPathMapper implements IEnvironmentPathMapper {
     private static final Log log = Log.getLog(WinStoreEnvPathMapper.class);
 
-    private static final String windowsAppLocalDataPackage = "DBeaverCorp.DBeaverCE_1b7tdvn0p0f9y";
-    private static final String appDataRoamingPathString = System.getenv("AppData"); 
-    private static final String localAppDataPathString = System.getenv("LOCALAPPDATA");
-    private static final String userHomePathString = System.getProperty("user.home");
+    private static final String WINDOWS_APP_LOCAL_DATA_PACKAGE = "DBeaverCorp.DBeaverCE_1b7tdvn0p0f9y";
+    private static final String APP_DATA_ROAMING_PATH_STRING = System.getenv("AppData");
+    private static final String LOCAL_APP_DATA_PATH_STRING = System.getenv("LOCALAPPDATA");
+    private static final String USER_HOME_PATH_STRING = System.getProperty("user.home");
     
     private final Path realVirtualizedRoot;
     
     public WinStoreEnvPathMapper() {
-        Path localAppDataPath = localAppDataPathString != null
-            ? Path.of(localAppDataPathString) 
-            : Path.of(userHomePathString, "AppData", "Local");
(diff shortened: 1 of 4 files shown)


plugins/org.jkiss.dbeaver.ui.app.winstore/src/org/jkiss/dbeaver/winstore/WinStoreEnvPathMapper.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.jkiss.code.NotNull;
+import org.jkiss.dbeaver.Log;
+import org.jkiss.dbeaver.runtime.IEnvironmentPathMapper;
+import org.jkiss.dbeaver.utils.RuntimeUtils;
+
+public class WinStoreEnvPathMapper implements IEnvironmentPathMapper {
+    private static final Log log = Log.getLog(WinStoreEnvPathMapper.class);
+
+    private static final String windowsAppLocalDataPackage = "DBeaverCorp.DBeaverCE_1b7tdvn0p0f9y";
```
Destrolaric: Are you sure that will not change after the new version? I am not sure that it is hardcoded.

Destrolaric (changes requested):

plugins/org.jkiss.dbeaver.ui.app.winstore/src/org/jkiss/dbeaver/winstore/WinStoreEnvPathMapper.java
E1izabeth: It's generated from the package and publisher identify information in the application package manifest.
As far as I know, it won't be changed while we change the information in the application package manifest.

Destrolaric: Looks like a thing we will forgot about when we will try to change the package identity :(

E1izabeth: I don't think we will change it. Anyway, at the moment I haven't found a good way to get it from somewhere.

Destrolaric (approved):

plugins/org.jkiss.dbeaver.ui.app.winstore/src/org/jkiss/dbeaver/winstore/WinStoreEnvPathMapper.java
```diff
@@ -0,0 +1,59 @@
...
+    private static final String appDataRoamingPathString = System.getenv("AppData"); 
+    private static final String localAppDataPathString = System.getenv("LOCALAPPDATA");
+    private static final String userHomePathString = System.getProperty("user.home");
```
ShadelessFox: I believe these should be `UPPER_CASE`.

plugins/org.jkiss.dbeaver.ui.app.winstore/src/org/jkiss/dbeaver/winstore/WinStoreEnvPathMapper.java
```diff
@@ -0,0 +1,59 @@
...
+        return RuntimeUtils.isWindowsStoreApplication() && localEnvPath.startsWith(appDataRoamingPathString);
+    }
+
+    @NotNull
+    @Override
+    public String map(@NotNull String localEnvPath) {
+        Path remappedPath = realVirtualizedRoot.resolve(Path.of(appDataRoamingPathString).relativize(Path.of(localEnvPath)));
+        String resultPath = remappedPath.toString();
+        
+        log.warn("Remapping file path [" + localEnvPath + "] to [" + resultPath + "]");
```
ShadelessFox: Maybe we want the level to be either `debug` or `trace`? It's not a warning.

ShadelessFox (approved):

ShadelessFox (approved):

PR #18305, opened 2022-11-23
dbeaver/pro#1263 Use drop-down menu with ordering and filters on colu…

serge-rider:
…mn click


ShadelessFox (changes requested):
This icon looks terrible due to having huge margins. Can we use a smaller one?

Before:
d8Qq53uGKA

After:
image

Notice how stretched it looks 😟

uslss:
The sorting icon isn't visible until it was used
dbeaver_6vcwfHlQ3s

serge-rider:
We removed sorting arrow on purpose (to reduce amount of icons on the screen).

Maybe we shouldn't leave empty space as well (it is supposed to use drop-down icon + ordering selection)). However it may look not very good..

serge-rider:
We disabled order by single click (now you need to click menu, then order asc/desc).
Menu and sort icons were swaped.

ShadelessFox (approved):

Issue #17162, opened 2022-07-24
Found some hard-coded cases where the localization properties file was not found (part 3）


bianyun1981:
#### System information: 
- Operating system (distribution) and version: Windows 10 Enterprise Edition (Simplified Chinese), 21H1
- DBeaver version: 22.1.3-SNAPSHOT (The latest development version at the time of posting)

#### Describe the problem you're observing:
After I updated the code of the develop branch to the latest version (last commit time: 2022-07-22 16:27 (GMT+8), commit by [yagudin10](https://github.com/dbeaver/dbeaver/commits?author=yagudin10),  commit hash: 75c25c1cc2a761b37272fa61a78c76b51f58cad0), I still found the following hard-coded cases without corresponding localization properties file. After these properties files are built, I can complete the Simplified Chinese localization of this part.

<details>

```
plugins/org.jkiss.dbeaver.ui.editors.data/plugin.xml
  - line 875:
        name="Database data flow"

plugins/org.jkiss.dbeaver.core/plugin.xml
  - line 33:
        <content-type file-extensions="lnk" id="org.jkiss.dbeaver.shortcut" name="Windows Shortcut" priority="normal">

plugins/org.jkiss.dbeaver.ext.oracle.ui/plugin.xml
  - line 268-269:
        <toolGroup id="org.jkiss.dbeaver.ext.oracle.ui.tools.statistics" label="Statistics"/>
        <toolGroup id="org.jkiss.dbeaver.ext.oracle.ui.tools.maintenance" label="Maintenance"/>
  - line 274:
        label="Execute in SQL*Plus">
  - line 284:
        label="Validate Structure"
  - line 292:
        label="Refresh materialized view"
  - line 301:
        label="Gather Statistics"
  - line 312:
        label="Truncate"
  - line 321:
        label="Disable trigger"
  - line 330:
        label="Enable trigger"

plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
  - line 402-436:
        name and description attribute in multiple lines 

plugins/org.jkiss.dbeaver.ext.oracle/plugin.xml
  - line 495-529:
        name and description attribute in multiple lines

plugins/org.jkiss.dbeaver.ext.postgresql/plugin.xml
  - line 659-707:
        name and description attribute in multiple lines


org.jkiss.dbeaver.model.exec.DBExecUtils
  - line 821:
        bindingMeta.setRowIdentifierStatus("No corresponding table column");
  - line 835
[...]

PR #18311, opened 2022-11-23
#17162 Complete some Simplified Chinese localization work

bianyun1981:
Most of the cases in issue #17162 already have corresponding property files, and the corresponding Simplified Chinese localization work has also been completed. 

There are only a few places where the corresponding property files are still not available. I will record these places in a new issue. 

If these places are not needed or cannot be localized, please comment them as '//$NON-NLS -1$' to avoid double-checking these places in the future.


plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
```diff
@@ -412,7 +412,7 @@
     </extension>
 
     <extension point="org.jkiss.dbeaver.task">
-        <category id="mysql" name="MySQL" description="%category.mysql.description" icon="icons/mysql_icon.png"/>
+        <category id="mysql" name="MySQL" description="%org.jkiss.dbeaver.task.category.mysql.description" icon="icons/mysql_icon.png"/>
```
LonwoLonwo, line 415: :+1:

E1izabeth (approved):

E1izabeth:
Thank you for contribution.

Matvey16:
Verified

PR #18324, opened 2022-11-24
dbeaver/pro#1284 Block UI queue during remote client invocations

serge-rider:


serge-rider:
Additional UI queue fixes have been committed.

Matvey16:
Verified

PR #18343, opened 2022-11-25
dbeaver/pro#135 Fix port forwarding not cancelling task correctly

ShadelessFox:


Matvey16:
Still reproducible.
STR
1. Open a script for SSHJ jump server connection
2. Run a query
3. Disconnect and reconnect VPN
4. Run a query again

It will get stuck
image

PR #18360, opened 2022-11-28
dbeaver/pro#1280 decrease consumers number if we lost producer(s) in …

LonwoLonwo:
…a task to avoid a non-editable task

2022-11-28 19_07_37-Error loading configuration

2022-11-28 19_07_43-Data Transfer -  export_case2

### DIFF
--- a/plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/DataTransferSettings.java
+++ b/plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/DataTransferSettings.java
@@ -185,9 +185,7 @@ public class DataTransferSettings implements DBTTaskSettings<DBPObject> {
                     // In this case, we had data transfer from table(s) to a file or another container.
                     // But, probably, table(s) were deleted, and we lost our producer(s).
                     // Usually, consumers do not have any special info, so we can delete extra items.
-                    IDataTransferConsumer<?,?>[] initConsumersCopy = new IDataTransferConsumer<?,?>[initProducers.length];
-                    System.arraycopy(initConsumers, 0, initConsumersCopy, 0, initProducers.length);
-                    initConsumers = initConsumersCopy;
+                    initConsumers = Arrays.copyOf(initConsumers, initProducers.length);
                 } else {
                     throw new IllegalArgumentException("Producers number must match consumers number");
                 }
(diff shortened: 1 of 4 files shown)


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/DataTransferSettings.java
```diff
@@ -179,7 +180,17 @@ private void initializePipes(
             // Both producers and consumers specified
             // Processor belongs to non-database nodes anyway
             if (initProducers.length != initConsumers.length) {
-                throw new IllegalArgumentException("Producers number must match consumers number");
+                // Something went wrong
+                if (initProducers.length < initConsumers.length && initProducers[0] instanceof DatabaseTransferProducer) {
```
ShadelessFox, line 184: Is it possible, by any chance, to have an array of producers (`initProducers`) that has zero elements?

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/DataTransferSettings.java
```diff
@@ -179,7 +180,17 @@ private void initializePipes(
...
+                    // In this case, we had data transfer from table(s) to a file or another container.
+                    // But, probably, table(s) were deleted, and we lost our producer(s).
+                    // Usually, consumers do not have any special info, so we can delete extra items.
+                    IDataTransferConsumer<?,?>[] initConsumersCopy = new IDataTransferConsumer<?,?>[initProducers.length];
+                    System.arraycopy(initConsumers, 0, initConsumersCopy, 0, initProducers.length);
+                    initConsumers = initConsumersCopy;
```
ShadelessFox: You can use the `Arrays#copyOf` here.

```suggestion
                    initConsumers = Arrays.copyOf(initConsumers, initProducers.length);
```

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/DataTransferSettings.java
ShadelessFox: Is it possible, by any chance, to have an array of producers (`initProducers`) that has zero elements?

LonwoLonwo: You can look in the code higher to know it.

ShadelessFox (approved):

LonwoLonwo:
Cases:

- [x] Export to CSV task with one table. Delete table. Try to open "Edit task" on the task in the Task Manager.
- [x] Export to CSV task with two (or more) tables. Delete one table. Try to open "Edit task" on the task in the Task Manager. The Task Manager will open after the error message.
- [x] Export to CSV task with two (or more) tables. Delete one table. Try to run the task. The task will fail with the message.
- [x] Export to another schema task with two (or more) new tables. Delete one table. Try to open "Edit task" on the task in the Task Manager. The Task Manager will open after the error message.
- [x] Export to another schema task with two (or more) new tables. Delete one table. Try to run the task. The task will fail with the message.
- [x] Export to another schema task with two (or more) existing tables. Delete one table. Try to open "Edit task" on the task in the Task Manager. The Task Manager will open after the error message.
- [x] Export to another schema task with two (or more) existing tables. Delete one table. Try to run the task. The task will fail with the message.

uslss:
verified

PR #18368, opened 2022-11-28
#18367 connection edit dialog is slow

### LINKED ISSUE #18367: Connection edit dialog is slow
### Description

When we open connection edit dialog it takes about 2 seconds to render UI.
This is very annoying.

It happens because of many tab folder and especially SSH and SSL configurators.
We must initialize them in a lazy mode.

Also we should improve SSH config page.

### DBeaver Version

22.2.5

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

serge-rider:


uslss:
verified

PR #18373, opened 2022-11-29
dbeaver/pro#1202 don't use Eclipse external selector for Linux and Mac

Destrolaric:
It was found out that this is not working for Linux and MacOS
Closes dbeaver/pro#1202


Matvey16:
Verified

PR #18374, opened 2022-11-29
Update DataVirtualityMetaModel.java

drozdse1:
This fixes a DV upper/lower case notation of objects bug.


drozdse1:
This fixes a DV upper/lower case notation of objects bug.

PR #18384, opened 2022-11-29
CB-2927 ssh bugs and improvements

alexander-skoblikov:


bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
```diff
@@ -186,6 +186,10 @@ public static boolean isEmpty(@Nullable Collection<?> value) {
         return value == null || value.isEmpty();
     }
 
+    public static boolean isNotEmpty(@Nullable Map<?, ?> value) {
+        return !isNotEmpty(value);
```
ShadelessFox: Is this meant to be `!isEmpty`?

alexander-skoblikov: Yep)

PR #18385, opened 2022-11-29
dbeaver/dbeaver#18383 Native script execute config UI fixes

### LINKED ISSUE #18383: Make field for path to postgres script editable
This field isn't editable, unlike all of the other places where we should choose path to file
image

serge-rider:


E1izabeth:
Shouldn't we validate that the path is correct and not empty?

ShadelessFox:
It would be a nice addition, indeed. Now it's working just as before, but now you can enter whatever you want in that field.

Matvey16:
Clicking on input field opens OS explorer to pick a file, which makes editing impossible

dbeaver_KeUCSYVJ9w

serge-rider:
Fixed

Matvey16:
Verified

PR #18395, opened 2022-12-01
dbeaver/pro#1268 Convert SSL keys (PEM -> DER) after connection failure

ShadelessFox:
https://github.com/dbeaver/pro/wiki/Sample-databases#postgresql-14-with-ssl


E1izabeth (approved):

LonwoLonwo (approved):

uslss:
verified

serge-rider:
Why is it for PostgreSQL only?
We have the SSL in MySQL, SQL Server and NoSQL databases

PR #18408, opened 2022-12-02
Fix french translation

despatates:
Fix the window title when exporting data.

In french, the verb is `Transférer` and the noun is `Transfert`.

image


E1izabeth (approved):

E1izabeth:
Thank you for contribution.

LonwoLonwo (approved):

PR #18421, opened 2022-12-02
import using directly the url

arons:
I tried to import some complex jdbc connection string from a csv file I created.

Example: jdbc:oracle:thin:@(DESCRIPTION=(FAILOVER=ON)(LOAD_BALANCE=OFF)(ADDRESS=(PROTOCOL=TCP)(HOST=10.11.12.13)(PORT=1521))(ADDRESS=(PROTOCOL=TCP)(HOST=10.11.12.14)(PORT=1521))(CONNECT_DATA=(SERVICE_NAME=DB_RW)))

Or sometime with additional parameter like: useUnicode=true&useJDBCCompliantTimezoneShift=true&useLegacyDatetimeCode=false&serverTimezone=UTC

Those kind of connection were resulting in a empty import url.

I think that the url should be in generally not changed from what the user specified.
That is why I commented out the part in which the url is parsed "adaptConnectionUrl".
The other fields like host, port and database are not necessary in case of a connection string url.

I tested for postgres and oracle jdbc connection string.

I there are other kind of connections that required the code to be present I would help to write code to parse in a proper way.

Thanks
Renzo


ShadelessFox (commented):
Thanks for the contribution!

I'm not sure how deleting a chunk of code somehow fixed your case. Mind explaining magic behind it?

Additionally, are you sure that this doesn't cover just your case and it still performs well against other variation of input strings?

arons:
The idea behind is that the jdbc connection string (url) it the final piece of information used to connect to the db, beside username and password.
So, if provided, information like host, port and db are not necessary (those 3 where parsed into the deleted code).
Instead, if the url is not provided, the url should be forged using host port and db in the way the driver will accept it, that is done the the part of the code that is left, probably not really according to the driver spec....that I did not check, but I leave it untouched.

The piece of deletec code was trying to parse the url according to driver spec in order to extract host, port and db, which as I said are not necessary a proper url is provided.

If you have some general requirement or test cases about the deleted part of code, I can help to code according to the required use cases.
If not, I keep my mind, that the url is the only necessary information for the jdbc driver and I don't see why we need to extract host, port and db from it. Especially I do not like that the import do not fail or report any error, just the url result is something like that: "jdbc:oracle:thin:@///" for a entry where I provided a fully working url.

serge-rider:
I don't like the ide of removing URL parser.
It still makes sense for many driver. Especially it makes sense for SSH connections (which are not supported for URL-based configurations).

arons:
Can you please provide me an example of ssh connection that are imported and for which the url need to be validated and why?
Just to understand in which case is necessary to parse and prove this url, maybe in this case make really sense to check it and can be implemented an ad hoc strategy, also with more information/feedback to the user. 
As I said, now very normal jdbc urls are not accepted. 
If a parser need to be implemented, should allow at least this normal cases, I can provide more examples.
If this ssh connections are exotic cases, does not sound logic to me to enforce parsing for those and create problem for more general cases.

E1izabeth (changes requested):
Please, don't remove working functionality, it may break some use cases. You could retry after unsuccessful parsing and create connection info by `conConfig.setConnectionType(DBPDriverConfigurationType.URL)`.

E1izabeth:
Please do not hesitate to ask if you need any assistance.

arons:
I added the parser of the driver sampleURL back in place. 
In case of error we not fail but instead proceed with the import with passed url.

E1izabeth:
Please, add this part of code after filling connection configuration `ConfigImportWizard:importConnection:L187-195`. It allows to specify whether connection url should be used directly or not after connection creation.
```
        if (CommonUtils.isEmpty(connectionInfo.getHost())) {
            config.setConfigurationType(DBPDriverConfigurationType.URL);
        } else {
            config.setConfigurationType(DBPDriverConfigurationType.MANUAL);
        }
```

arons:
Done and tested with the use cases I have. Looks nice.

E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Verified

Issue #5960, opened 2019-05-21
Support SpatiaLite in Spatial Viewer


ar-jan:
I would like to suggest support for [SpatiaLite](https://www.gaia-gis.it/fossil/libspatialite/index) in the Spatial Viewer, similar to #4222 for PostGIS.

neylsongularte:
+1

fegyi001:
+1

serge-rider:
We will use standard SQLite (Xerial) driver and load mod_spatialite extension dynamically.
No sure - whether we need to make a separate driver or just provide extensions management UI.

There is a set of issues with spatialite extension:
- https://github.com/sqlitebrowser/sqlitebrowser/issues/267
- https://github.com/sqlitebrowser/sqlitebrowser/wiki/SpatiaLite-on-Windows#workaround-for-windows-10-failure

I'm not sure that we should handle these problems on dbeaver side. Having extensions management UI this can be resolved by users. Spatialite isn't the only interesting extension for sqlite after all.

serge-rider:
Depends on #6830

fegyi001:
+1

zoot-inge:
+1

wojtech:
+1

Pixelfahnder:
+1

PR #18431, opened 2022-12-05
#6830 SQLite extension management

### LINKED ISSUE #6830: SQLite: extension management UI
We need some UI where we can manager (list/install/uninstall) SQLite extensions.
Most straight solution is to add extra connection settings page. Thus it will be available for all SQLite clones.

SpatiaLite and other similar "drivers" will depend on this feature.

ShadelessFox:
image


E1izabeth (approved):

LonwoLonwo (approved):

DariaMarkaryan:
Verified

ar-jan:
What's the status of this feature? It was scheduled for 22.3.3 in #5960 but it looks like it wasn't included yet?

ShadelessFox:
It's included in 22.3.3.

This PR brings support for extension management UI. You can install the SpatiaLite extension via the new UI.

ar-jan:
Ah good to hear. I do not see the `SQLite extensions`  tab shown in the screenshot above, and there was no mention in the release notes, hence my confusion.

 I did now find **enable_load_extension** under Driver properties. The tooltip says `Enable SQLite load_extension() function, native driver only`. What does "native driver" mean here? I'm using the default xerial driver (`SQLite JDBC 3.40.0.0`).

I've set `enable_load_extension`: `true`, but that doesn't result in a `SQLite extensions` tab. It seems I can do `SELECT load_extension('mod_spatialite');` successfully however.

Is the best way to use it then to add `SELECT load_extension('mod_spatialite');` as a bootstrap query per connection? And is there a way to get the extension tab? (I'm on Linux, if it matters).

ShadelessFox:
Sorry for the created confusion. The extension management UI is part of PRO products.

You can download the EA version here: https://dbeaver.com/files/ea/.

malcook:
> Is the best way to use it then to add `SELECT load_extension('mod_spatialite');` as a bootstrap query per connection?

FWIW: I have found this approach to work in my use of the [pivot_vtab](https://github.com/jakethaw/pivot_vtab) sqlite3 extension from within DBeaver.

nyurik:
i understand the need to differentiate paid/ce product lines (which allows further development of this awesome tool!), but I think this feature may need to be CE because it is required for the feature parity with PostgreSQL maps support. Otherwise, PostgreSQL + PostGIS maps view works in CE, but Sqlite + Spatialite maps requires EA.  Thanks!!

bazmap:
+1, @nyurik is right, without this feature DBeaver doesn't work with spatialite nor geopackage (which is a spatialite db in fact).
I think this feature is really important for GIS users

ShadelessFox:
It's still possible to load extensions manually via SQL commands. https://github.com/dbeaver/dbeaver/issues/34475#issuecomment-2191376521

bazmap:
Hi, I've seen it but it makes using spatialite databases and geopackage not userfriendly. A lot's of people working in GIS uses DBeaver and being forced to do some magic trick to read really common GIS databases is really a shame.
Maybe it would be nice to autoload the spatial extension when it comes to use DBeaver with spatialite or geopackage db ?

PR #18432, opened 2022-12-05
#18258 Variables resolution

### LINKED ISSUE #18258: Data Transfer output parameter variables malfunction when used more than once
### Description

The first variable that is mentioned, when used the second time, will apply the variable name as a string.

(Examples below were tested on 2022-11-17)

Example (in the output directory parameter)
/some/path/${year}/${year}-${month}/${year}-${month}-${day}

Will result in:
/some/path/2022/${year}-11/${year}-11-17
(Note that year is the first variable, and it works for the first instance. Note that month is the second variable, and it works for both instances.)

Or second example:
/some/path/2022/2022-${month}/${year}-${month}-${day}

Will result in:
/some/path/2022/2022-11/2022-${month}-17

The first instance of the first variable renders correctly. The subsequent instances of the first variable do not.
The subsequent variables work fine no matter how many times they are called.

### DBeaver Version

Community Edition 22.2.4

### Operating System

macOS Monterey 12.6

### Database and driver

_No response_

### Steps to reproduce

Export a query, on the Data Transfer / Output page, in the output directory, place

/some/path/${year}/${year}-${month}/${year}-${month}-${day}

Expected output: 
/some/path/2022/2022-11/2022-11-17

Actual output:
/some/path/2022/${year}-11/${year}-11-17

### Additional context

_No response_

E1izabeth:
Variables in data export should be resolved correctly. If the variable encounters more than one time, it's value should be substituted in all occurrences.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
@@ -533,11 +533,10 @@ public class GeneralUtils {
     @NotNull
     private static String substituteVariable(@NotNull String string, @NotNull Matcher matcher, @NotNull String varValue) {
         if (matcher.start() == 0 && matcher.end() >= string.length() - 1) {
-            string = varValue;
+            return varValue;
         } else {
-            string = string.substring(0, matcher.start()) + varValue + string.substring(matcher.end());
+            return string.substring(0, matcher.start()) + varValue + string.substring(matcher.end());
         }
-        return string;
     }
 
     public static IStatus makeExceptionStatus(Throwable ex) {
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
@@ -489,17 +489,24 @@ public class GeneralUtils {
             return string;
         }
         // We save resolved vars here to avoid resolve recursive cycles
-        List<String> resolvedVars = null;
+        Map<String, String> resolvedVars = null;
         try {
             Matcher matcher = VAR_PATTERN.matcher(string);
             int pos = 0;
             while (matcher.find(pos)) {
                 pos = matcher.end();
                 String varName = matcher.group(2);
-                if (resolvedVars != null && resolvedVars.contains(varName)) {
-                    continue;
+                String varValue = null;
+                if (resolvedVars != null) {
+                    varValue = resolvedVars.get(varName); 
+                    if (varValue != null) {
+                        string = substituteVariable(string, matcher, varValue);
+                        matcher = VAR_PATTERN.matcher(string);
+                        pos = 0;
+                        continue;
+                    }
                 }
-                String varValue = resolver.get(varName);
+                varValue = resolver.get(varName);
                 if (varValue == null) {
                     varValue = matcher.group(3);
                     if (varValue != null && varValue.startsWith(":")) {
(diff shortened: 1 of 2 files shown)


test/org.jkiss.dbeaver.test.platform/src/org/jkiss/utils/GeneralUtilsTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Map;
+
+import org.jkiss.code.NotNull;
+import org.jkiss.dbeaver.utils.GeneralUtils;
+import org.junit.Test;
```
ShadelessFox:
```suggestion
import org.jkiss.code.NotNull;
import org.jkiss.dbeaver.utils.GeneralUtils;
import org.junit.Test;

import java.text.SimpleDateFormat;
import java.util.Date;
import java.util.Map;

import static org.junit.Assert.assertEquals;
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -527,6 +530,16 @@ public static String replaceVariables(@NotNull String string, IVariableResolver
...
     }
 
+    @NotNull
+    private static String substituteVariable(@NotNull String string, @NotNull Matcher matcher, @NotNull String varValue) {
+        if (matcher.start() == 0 && matcher.end() >= string.length() - 1) {
+            string = varValue;
+        } else {
+            string = string.substring(0, matcher.start()) + varValue + string.substring(matcher.end());
+        }
+        return string;
```
ShadelessFox:
```suggestion
        if (matcher.start() == 0 && matcher.end() >= string.length() - 1) {
            return varValue;
        } else {
            return string.substring(0, matcher.start()) + varValue + string.substring(matcher.end());
        }
```

E1izabeth: I don't think that multiple exit points is a good practice in a common case. There is no reason to do it in this method.

E1izabeth: I don't think that multiple exit points is a good practice in a common case. There is no reason to do it in this method.

ShadelessFox: Reassigning an argument is neither a good practice.

E1izabeth: OK

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -532,12 +532,13 @@ public static String replaceVariables(@NotNull String string, IVariableResolver
...
+        String result;
         if (matcher.start() == 0 && matcher.end() >= string.length() - 1) {
-            string = varValue;
+            result = varValue;
         } else {
-            string = string.substring(0, matcher.start()) + varValue + string.substring(matcher.end());
+            result = string.substring(0, matcher.start()) + varValue + string.substring(matcher.end());
         }
-        return string;
+        return result;
```
ShadelessFox: Do you think that this is prettier than two `return` statements?

ShadelessFox (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #18448, opened 2022-12-06
#15912 Allow disabling editing if no primary key is defined

### LINKED ISSUE #15912: feature request: data editor: option to prevent using "all columns" as key in data editor
When you're editing the data in the result set from a query, and the result set does not contain the primary key, dbeaver alerts you that data beyond what is shown maybe modified.

image

However, it still allows you to press the button "Use All Columns" !

This has caused a big problem in the main database, and it would be nice to disable this option via Settings, or making this possibility less accesible.

Alternatively, the same WHERE that was applied to the original query that produced the data visible on the editor, could be applied to the modification script.

Nevertheless I understand the problem was in the guy that pressed the button, but would be nice to prevent this behaviour.

Best regards

ShadelessFox:
image

Acceptance criteria:
- [x] The `Use all table columns as key upon saving` works as before
- [x] The `Disable data editing` doesn't restring data editing if the primary or virtual key is defined
- [x] These two options can't be used together


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/preferences/PrefPageResultSetMain.java
```diff
@@ -314,6 +324,7 @@ protected void clearPreferences(DBPPreferenceStore store) {
 
         store.setToDefault(ResultSetPreferences.KEEP_STATEMENT_OPEN);
         store.setToDefault(ResultSetPreferences.RS_EDIT_USE_ALL_COLUMNS);
+        store.setToDefault(ResultSetPreferences.RS_EDIT_KEY_VALIDATION_STRATEGY);
```
Destrolaric: Looks like we don't have a default value defined anywhere

Destrolaric (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/preferences/PrefPageResultSetMain.java
ShadelessFox: The default value is `null`.

E1izabeth (approved):

LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/preferences/PrefPageResultSetMain.java
```diff
@@ -214,11 +215,39 @@ public void widgetSelected(SelectionEvent e) {
 
         }
 
+        {
+            final Group group = UIUtils.createControlGroup(leftPane, "Editing", 1, GridData.VERTICAL_ALIGN_BEGINNING, 0);
```
E1izabeth: I'd like to have a localization for "Editing"

ShadelessFox: Yes, thank you.

E1izabeth (approved):

E1izabeth (approved):

LonwoLonwo (approved):

Matvey16:
I found some strange behavior that happens because virtual key is created.

STR
1. Check `Use all columns as key upon saving`
2. Change something in a table and save changes
3. Open properties and change missing key behavior to `Disable editing`
4. Change something in a table and click save.

**Expected:** The table doesn't have a key so it can't be edited
**Actual:** The table actually has virtual key that was created prior so changes are saved

Maybe the actual behavior is expected, but it seems strange to me

Matvey16:
Verified

PR #18456, opened 2022-12-06
#18451 Don't disable Use Windows trust store option if it was set by DBeaver

### LINKED ISSUE #18451: Use Windows trust store option is disabled when shouldn't
### Description

If I haven't set trust store manually, DBeaver did it for me and option is disabled. This option shouldn't be disabled if this property was set by DBeaver.

### DBeaver Version

DBeaver 22.3.0

### Operating System

Windows

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Should be disabled ONLY if user set `javax.net.ssl.trustStore` to any value and/or `javax.net.ssl.trustStoreType` not equals to `WINDOWS_TRUSTSTORE`


ShadelessFox (approved):

uslss:
verified

PR #18512, opened 2022-12-14
CB-2462 sync tokens naming

yagudin10:


ShadelessFox (commented):
Do you really need to prefix these fields with the `sm` even though they are fields of classes with the `SM` prefix too?

PR #18542, opened 2022-12-20
dbeaver/pro#1256 use complex target name in export query if query has…

LonwoLonwo:
… joins; add the ability to change table name for the SQL export

2022-12-20 14_24_21-Window

2022-12-20 18_09_52-Window

Qriteries:

- [x] The table name does not change in the export statement if the "Target table name" property is empty
- [x] The table name changes in the export statement if the "Target table name" property has any string
- [x] The "Target table name" property has a description
- [x] The target file name with {table} parameter will have a merged name (ask me)


plugins/org.jkiss.dbeaver.data.transfer/OSGI-INF/l10n/bundle.properties
```diff
@@ -75,6 +75,8 @@ dataTransfer.processor.sql.property.includeAutoGenerated.description=Include aut
 dataTransfer.processor.sql.property.escape.name=Escape character
 dataTransfer.processor.sql.property.escape.description=Character which escapes single quote character
 dataTransfer.processor.sql.property.extension.label = File extension
+dataTransfer.processor.sql.property.userTableName.name = Target table name
+dataTransfer.processor.sql.property.userTableName.description = You can specify the name of the target table in this field. It will be used to generate an INSERT statement.\nBy default, the name of the original table/tables is used.\nThe table name introduced into this field will be used as is.\nIf you want the table name also contains the scheme/catalog identifier, then you can prescribe it in this field by yourself.
```
E1izabeth: Maybe we could write it shorter.
For example
```suggestion
dataTransfer.processor.sql.property.userTableName.description = You can specify the name of the target table. It will be used as is to generate an INSERT statement.\nIf you want the table name also contains the scheme or/and catalog identifier, you can prescribe it in this field by yourself.\nBy default, the name of the original table is used.
```

E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #18566, opened 2022-12-22
#18207 Show url in navigator instead of host if connection mode is by url

### LINKED ISSUE #18207: Database navigator connection host info is wrong when using connect by url
### Description

When using connect by Url instead of connect by Host, the host info in the database navigator is displayed as "localhost" instead of the actual targeted host. In my case i connected with an external ms sql server. See attached screenshot of the database navigator.

My expectation was that when "connect by url" was chosen, the host would get parsed from the url and not show "localhost" which is the field in the connection window that is now grayed out.

database navigator connection info

### DBeaver Version

22.2.4

### Operating System

Windows Server 2019

### Database and driver

Database: MS SQL Server 2019
Driver: mssql-jdbc 9.2.0.jre8

### Steps to reproduce

1. Make a new connection to a MS SQL Server
2. choose connect by url
3. observer in database navigator

### Additional context

_No response_

E1izabeth:
Navigator should show actual information about connection host/url depends on the connection mode.


ShadelessFox (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #18586, opened 2022-12-25
Postgres track function changed#18412

plotn:
See original issue's screenshot


plotn:
sorry, incorrect branch ((

PR #18587, opened 2022-12-25
#17209 Print indentation of nested elements

### LINKED ISSUE #17209: Improved "Plan" results window
Currently, there are many cases of comparing "PLAN" for sql tuning, so I hope that some improvements will be made in the PLAN execution result window.

1. Add a function to select all in the Oracle PLAN execution result window (ex : ctrl+a)
  image

3. When advanced copy is executed, I want the "Operation" item to be indented as shown.
   Or, I would like depth to be displayed.
 image

Thank you for reading. :)

ShadelessFox:
Only the second request was implemented. The first one wasn't implemented because it's very specific.

Note: works with the `Copy` action only.

image


LonwoLonwo (approved):

E1izabeth (approved):

uslss:
verified

PR #18588, opened 2022-12-25
#18412 Postgres function editing enchancement

### LINKED ISSUE #18412: Postgres function editing enchancement
Hi!
There is function editing dialog (or, maybe, more widely - object editing dialog).
And we can change function text, sure, then pressing Ctrl-S and persist our changes. And there are 3 ways:
1. Object is saved and everything is ok;
2. We _intentionally_ changed the name of the object and it was saved under different name leaving old object unchanged;
3. We changed parameters or types and (even we did not want) the new object is created (overloaded) - this is (at least) PG feature.

So the 2 and 3 scenarios look not consistent - we stay at the old object, but created (sometimes unintentionally) new object.

Could we throw some warning, e.g. "object signature (name and/or parameter list) was changed, so new object will be created instead of editing of current, proceed" ?
But, I guess it could be diffucult to determine - will it be replaced or created - I dont know. Maybe we can try "create" firstly, then, if fail, "create or replace" and, if success - we are sure, that we replaced, otherwise - created. Anyway, the message can be "post-action" - "object signature (name and/or parameter list) was changed, so new object was created instead of editing of current"

plotn:
see issue's screenshot  #18412


plotn:
@E1izabeth , the link 
https://build.dev.dbeaver.com/job/DBeaverDesktop/job/dbeaver-ce-github-checks/job/PR-18588/2/display/redirect
is down for some days - will it be available later?

E1izabeth:
Hi! It's not down, it may be behind the firewall. I can share a log for you [log.txt](https://github.com/dbeaver/dbeaver/files/10308239/log.txt)

plotn:
@E1izabeth, thank you.
But I have no idea what is wrong (and it is not my code, i guess):

[ERROR] Failed to execute goal org.eclipse.tycho:tycho-p2-repository-plugin:3.0.0:assemble-repository (default-assemble-repository) on project org.jkiss.dbeaver.ce.product.updateSite: Execution default-assemble-repository of goal org.eclipse.tycho:tycho-p2-repository-plugin:3.0.0:assemble-repository failed: An API incompatibility was encountered while executing org.eclipse.tycho:tycho-p2-repository-plugin:3.0.0:assemble-repository: java.lang.NoSuchMethodError: 'void org.eclipse.equinox.internal.p2.repository.helpers.ChecksumProducer.<init>(java.lang.String, java.lang.String, java.lang.String)'

E1izabeth:
@plotn I believe so too, because I don't see any weird import or something unusual in your code. Have you tried to pull last changes from devel branch?

plotn:
Good idea, i'll try

E1izabeth (approved):

Destrolaric (approved):

E1izabeth:
Close #18412

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreMessages.java
```diff
@@ -0,0 +1,36 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.ext.postgresql;
+
+import org.eclipse.osgi.util.NLS;
+
+public class PostgreMessages extends NLS {
```
LonwoLonwo: We already have PostgreMessages https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/PostgreMessages.java

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreMessages.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2022 DBeaver Corp and others
+ * Copyright (C) 2017 Liu, Yuanyuan (liuyuanyuan@highgo.com)
```
LonwoLonwo: redundant srting

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgresResources.properties
```diff
@@ -0,0 +1,2 @@
+procedure_warning_name=Warning
+procedure_warning_text=The initial object seems to be unchanged. This may happen when object signature was changed (name and/or parameters), so the new object was created (e.g. overloaded), leaving initial object unchanged. Check this if it was uninentionally.
```
LonwoLonwo: procedure_warning_text=The initial object seems to be unchanged. This may happen when the object signature is changed (name and/or parameters), so the new object is created (e.g., overloaded), leaving the initial object unchanged. Check this if it was unintentional.
How about this?

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreMessages.java
plotn: Yes, but they are in UI plugin. Is there a way to use them from another plugin? I could not do this, so created new file

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreMessages.java
plotn: fixed

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgresResources.properties
plotn: is... was... guess you're right, changed

uslss:
Procedure tab is incorrectly shown


https://user-images.githubusercontent.com/31996417/213144534-4edc03c6-2284-471e-8ca2-08919ed9e02b.mp4

plotn:
@uslss , I guess this is irrelevant with this PR - the reason maybe related with something else

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
```diff
@@ -311,6 +319,22 @@ public void setKind(PostgreProcedureKind kind) {
     public long getObjectId() {
         return oid;
     }
+    
+    public long getXmin() {
+        return xmin;
+    }
```
Destrolaric, line 343: I think we can use @Property annotation to show it in procedure metadata.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
```diff
@@ -311,6 +319,22 @@ public void setKind(PostgreProcedureKind kind) {
...
+        return procWasCreatedOrReplaced;
+    }
+    
+    public void setProcWasCreatedOrReplaced(boolean value) {
+        this.procWasCreatedOrReplaced = value;
+    }
+    
+    public boolean getProcWasNotChanged() {
+        return procWasNotChanged;
+    }
```
Destrolaric: Not entirely sure that we need two variables to mark that proc was changed. At least I don't see `procWasNotChanged` checked anywhere.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreMessages.java
plotn: @LonwoLonwo , finally what should I do?

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
plotn: Can you show me where I can see thap procedure metadata?
I added annotation you said

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
plotn: removed

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
E1izabeth: When you open some function, its metadata is in the top panel of properties view. Probably xmin will be somewhere here
image

plotn: thank you !

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreMessages.java
E1izabeth: I tested your code and found that the reason for this https://github.com/dbeaver/dbeaver/pull/18588#issuecomment-1386803374 is the following error. Actually such errors happens not only in this case, but even when trying to create new connection for postgresql. It occurred because `PostgreMessages` class with localization fields and bundle name `"org.jkiss.dbeaver.ext.postgresql.PostgresResources"` are absolutely the same for ui bundle and non-ui bundle where you added them. Probably `NLS.initializeMessages(BUNDLE_NAME, PostgreMessages.class);` should have unique arguments otherwise previous initialization will be overwritten.

So I suggest to rename the class or/and resource bundle. Maybe renaming .properties files would be enough.
```
java.lang.NoSuchFieldError: source_view_show_header_label
	at org.jkiss.dbeaver.ext.postgresql.ui.editors.PostgreSourceViewEditor.contributeEditorCommands(PostgreSourceViewEditor.java:90)
	at org.jkiss.dbeaver.ui.editors.sql.SQLEditorNested$EditorPageControl.fillCustomActions(SQLEditorNested.java:415)
	at org.jkiss.dbeaver.ui.controls.ProgressPageControl.fillCustomActions(ProgressPageControl.java:259)
	at org.jkiss.dbeaver.ui.controls.ObjectEditorPageControl.fillCustomActions(ObjectEditorPageControl.java:121)
	at org.jkiss.dbeaver.ui.editors.entity.properties.ObjectPropertiesEditor$1.fillCustomActions(ObjectPropertiesEditor.java:116)
	at org.jkiss.dbeaver.ui.controls.ProgressPageControl.hideControls(ProgressPageControl.java:304)
	at org.jkiss.dbeaver.ui.controls.ProgressPageControl.setChildControl(ProgressPageControl.java:173)
	at org.jkiss.dbeaver.ui.controls.ProgressPageControl.substituteProgressPanel(ProgressPageControl.java:139)
	at org.jkiss.dbeaver.ui.controls.ProgressPageControl.createOrSubstituteProgressPanel(ProgressPageControl.java:146)
	at org.jkiss.dbeaver.ui.editors.sql.SQLEditorNested.createPartControl(SQLEditorNested.java:143)
	at org.jkiss.dbeaver.ui.editors.entity.properties.TabbedFolderPageEditor.createControl(TabbedFolderPageEditor.java:98)
	at org.jkiss.dbeaver.ui.controls.folders.TabbedFolderComposite$FolderPane.onFolderSwitch(TabbedFolderComposite.java:153)
	at org.jkiss.dbeaver.ui.controls.folders.TabbedFolderComposite$FolderPane$2.widgetSelected(TabbedFolderComposite.java:136)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:252)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:89)
```

plotn: @E1izabeth , I agree with renaming, can you suggest proper class name? PGMessages?

E1izabeth: I'd call it PostgreSQLMessages

plotn:
@E1izabeth , hi! What is "sp1" label ? and there is "wait for response" - is it "waiting from me"?

E1izabeth:
It broke some functionality. I investigated why and suggested how to fix it.
https://github.com/dbeaver/dbeaver/pull/18588#discussion_r1084658026

sp is a story point

E1izabeth:
I see that you commited changes recently, but I haven't reviewed it yet, so haven't changed the label

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreSQLMessages.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.ext.postgresql;
+
+import org.eclipse.osgi.util.NLS;
+
+public class PostgreSQLMessages extends NLS {
+    static final String BUNDLE_NAME = "org.jkiss.dbeaver.ext.postgresql.PostgresResources"; //$NON-NLS-1$
```
E1izabeth: This name should be equal to the name of .properties file
```suggestion
    static final String BUNDLE_NAME = "org.jkiss.dbeaver.ext.postgresql.PostgreSQLMessages"; //$NON-NLS-1$

```

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
```diff
@@ -329,7 +335,20 @@ public void setKind(PostgreProcedureKind kind) {
     public long getObjectId() {
         return oid;
     }
-
+    
+    @Property(order = 6)    
+    public long getXmin() {
```
ShadelessFox, line 341: This property lacks localization in the `plugins/org.jkiss.dbeaver.ext.postgresql/OSGI-INF/l10n/bundle.properties`

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/PostgreDatabasePersistAction.java
```diff
@@ -0,0 +1,44 @@
...
+        super(title, procedure.getBody(), complex);
+        this.procedure = procedure;
+        this.title = title;
+    }
+       
+
+    @Override
+    public void afterExecute(DBCSession session, Throwable error)
+            throws DBCException {
+        if ((procedure != null) && ("Create function".equals(this.title))) {
```
ShadelessFox: This looks hacky. Can we pass a `boolean` parameter denoting this action's purpose to create a new function? Maybe it's not even required since this action is created in one place?

Also, the `procedure` parameter can't be null because it's dereferenced before even creating an instance of this class.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/PostgreDatabasePersistAction.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.jkiss.dbeaver.ext.postgresql.model.PostgreProcedure;
+import org.jkiss.dbeaver.model.exec.DBCException;
+import org.jkiss.dbeaver.model.exec.DBCSession;
+import org.jkiss.dbeaver.model.impl.edit.SQLDatabasePersistAction;
+
+public class PostgreDatabasePersistAction extends SQLDatabasePersistAction {
+    
+    private PostgreProcedure procedure;
+    private String title;
```
ShadelessFox:
```suggestion
    private final PostgreProcedure procedure;
    private final String title;
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/PostgreDatabasePersistAction.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public PostgreDatabasePersistAction(String title, boolean complex, PostgreProcedure procedure) {
```
ShadelessFox:
```suggestion
    public PostgreDatabasePersistAction(@NotNull String title, boolean complex, @NotNull PostgreProcedure procedure) {
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
```diff
@@ -813,7 +832,19 @@ public Collection<PostgrePrivilege> getPrivileges(DBRProgressMonitor monitor, bo
...
+        long oldXmin = this.getXmin();
+        PostgreProcedure pgProc = getContainer().getProceduresCache().refreshObject(monitor, getContainer(), this);
+        if (oldProcWasCreatedOrReplaced) {
+            if ((oldOid != 0) && (oldXmin != 0) && (pgProc.oid == oldOid) && (pgProc.xmin == oldXmin)) {
+                DBWorkbench.getPlatformUI().showWarningMessageBox(PostgreSQLMessages.procedure_warning_name, 
+                    PostgreSQLMessages.procedure_warning_text);
+                this.setProcWasCreatedOrReplaced(false); // let throw warning once
+                pgProc.setProcWasCreatedOrReplaced(false);
+            }
+        }
```
ShadelessFox: Too many redundant parentheses; nested `if`-s can be merged

```suggestion
        if (oldProcWasCreatedOrReplaced && oldOid != 0 && oldXmin != 0 && pgProc.oid == oldOid && pgProc.xmin == oldXmin) {
            DBWorkbench.getPlatformUI().showWarningMessageBox(PostgreSQLMessages.procedure_warning_name, 
                PostgreSQLMessages.procedure_warning_text);
            this.setProcWasCreatedOrReplaced(false); // let throw warning once
            pgProc.setProcWasCreatedOrReplaced(false);
        }
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
```diff
@@ -118,6 +120,8 @@ enum TransitionModifies {
     }
 
     private long oid;
+    private long xmin;
+    private boolean procWasCreatedOrReplaced;
```
ShadelessFox: What's the point of the `proc` prefix if we're already inside the class called `PostgreProcedure`? Please remove it

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
plotn: May I remove "property" from this field? I think it is unnessesary

ShadelessFox: You can.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/PostgreDatabasePersistAction.java
plotn: Okay, I'd better intruduce enum

LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/PostgreDatabasePersistAction.java
ShadelessFox: How is enum applicable here?

ShadelessFox (approved):

Destrolaric (approved):

E1izabeth (approved):

Matvey16:
Verified

serge-rider:
Generally PR is ok. But I don't like to do it for procedures only. We should use the same for other system objects (tables, columns, etc).

Also, there are many PostgreSQL forks. Some of them do not support xmin at all (e.g. CockroachDB). I wouldn't rely on it, or at least use some workaround when xmin is not supported.

plotn:
> But I don't like to do it for procedures only. We should use the same for other system objects (tables, columns, etc).

Procedures is the general pain, because they can be "imperceptibly" copied (with changed parameters), why not to make life a little bit better right now?

> Also, there are many PostgreSQL forks. Some of them do not support xmin at all (e.g. CockroachDB). I wouldn't rely on it, or at least use some workaround when xmin is not supported.

So I can add xmin checking if we decide to merge this PR. Can you advise the right way? (I can make it via try/except, but if there is a more intelligent solution).

But, finally - what we'll do now?

plotn:
@serge-rider, I did some investigation and found the following:

we need to add to:
public interface PostgreServerExtension

something like:
 boolean supportsXmin(); (like supportsOids())

And maybe:
String getProceduresXminColumn(); (like String getProceduresOidColumn())

but I do not feel the power to analyze all of the PG forks, so what if?:
- you help me to get this information (if you have experience among all variations and versions of them), or:
- I make this change only for "classic" PG (edb and PostgreSQL) and mark others as "not support xmin" leaving old behaviour.

plotn:
This code works, should I use it this way?:

```
if (xmin_supported == null) { // Check if xmin supported
        	xmin_supported = false;
        	try (JDBCPreparedStatement dbXminCheck = session.prepareStatement(
                        "SELECT xmin FROM pg_catalog." + serverType.getProceduresSystemTable() + " limit 1")) {
                        try (JDBCResultSet dbResult = dbXminCheck.executeQuery()) {
                            while (dbResult.next()) {
                        	xmin_supported = true;
                            }
                        }
        	}
            }
            JDBCPreparedStatement dbStat = session.prepareStatement(
                "SELECT p." + oidColumn + " as poid,p.*," +
                    (session.getDataSource().isServerVersionAtLeast(8, 4) ? "pg_catalog.pg_get_expr(p.proargdefaults, 0)" : "NULL") +
                    " as arg_defaults,d.description, " + (xmin_supported ? "p.xmin" : "0") + " as pxmin\n" +
                    "FROM pg_catalog." + serverType.getProceduresSystemTable() + " p\n" +
                    "LEFT OUTER JOIN pg_catalog.pg_description d ON d.objoid=p." + oidColumn + "\n" +
                    "WHERE p.pronamespace=?" +
                    (object == null ? "" : " AND p." + oidColumn + "=?") +
                    "\nORDER BY p.proname"
            );
```

ShadelessFox (approved):

Matvey16:
@plotn Hi, I am a little bit lost in all of the changes. Can you list acceptance criteria for this PR. Like what behaviour was changed and for which type of objects and databases?

plotn:
> @plotn Hi, I am a little bit lost in all of the changes. Can you list acceptance criteria for this PR. Like what behaviour was changed and for which type of objects and databases?

Read the very first comment please.
This is for Postgres fuction only

plotn:
added xmin check

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
```diff
@@ -1099,15 +1099,28 @@ public ProceduresCache() {
...
+            if (xmin_supported == null) { // Check if xmin supported
+                xmin_supported = false;
+                try (JDBCPreparedStatement dbXminCheck = session.prepareStatement(
+                        "SELECT xmin FROM pg_catalog." + serverType.getProceduresSystemTable() + " limit 1")) {
+                    try (JDBCResultSet dbResult = dbXminCheck.executeQuery()) {
+                        while (dbResult.next()) {
+                            xmin_supported = true;
+                        }
+                    }
+                }
```
ShadelessFox, line 1129: You're missing a `catch` block here. An exception will be thrown if no such column is present in the result set.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
ShadelessFox: You can verify that by changing the `xmin` column to something that is not present in the result set.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
```diff
@@ -1099,28 +1099,30 @@ public ProceduresCache() {
...
                 try (JDBCPreparedStatement dbXminCheck = session.prepareStatement(
                         "SELECT xmin FROM pg_catalog." + serverType.getProceduresSystemTable() + " limit 1")) {
                     try (JDBCResultSet dbResult = dbXminCheck.executeQuery()) {
                         while (dbResult.next()) {
-                            xmin_supported = true;
+                            xminSupported = true;
                         }
                     }
+                } catch(Exception e) {
+                    // do nothing
```
ShadelessFox:
```suggestion
                } catch (Exception ignored) {
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
```diff
@@ -1099,28 +1099,30 @@ public ProceduresCache() {
...
             PostgreServerExtension serverType = owner.getDataSource().getServerType();
             String oidColumn = serverType.getProceduresOidColumn(); // Hack for Redshift SP support
-            if (xmin_supported == null) { // Check if xmin supported
-                xmin_supported = false;
+            if (xminSupported == null) { // Check if xmin supported
+                xminSupported = false;
                 try (JDBCPreparedStatement dbXminCheck = session.prepareStatement(
                         "SELECT xmin FROM pg_catalog." + serverType.getProceduresSystemTable() + " limit 1")) {
                     try (JDBCResultSet dbResult = dbXminCheck.executeQuery()) {
                         while (dbResult.next()) {
```
ShadelessFox:
```suggestion
                        if (dbResult.next()) {
```

ShadelessFox (approved):

Matvey16:
Seems like this feature stopped working after all the changes

https://user-images.githubusercontent.com/82543000/234557730-b85f42ce-e91d-47b4-80ae-48ce52455325.mp4

plotn:
@E1izabeth, @ShadelessFox , @LonwoLonwo could you help me?
I did clean repo (compiled under 202303), clean branch, cherry-picked needed commits - to do the clean tests.
Everything works fine under debugger under eclipse, but built application - doesnt work.
I cannot understand (

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/PostgreDatabasePersistAction.java
```diff
@@ -0,0 +1,55 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2022 DBeaver Corp and others
```
LonwoLonwo, line 3:
```suggestion
 * Copyright (C) 2010-2023 DBeaver Corp and others
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreSQLMessages.java
```diff
@@ -0,0 +1,32 @@
+/*
```
LonwoLonwo, line 1:
```suggestion
/*
 * DBeaver - Universal Database Manager
 * Copyright (C) 2010-2023 DBeaver Corp and others
 *
```

LonwoLonwo:
Ok, maybe it will help.

E1izabeth:
Does merging the devel branch by @LonwoLonwo helped you to built the project? 

Please let us know if you're going to fix the mentioned [issue](https://github.com/dbeaver/dbeaver/pull/18588#issuecomment-1523238865).

E1izabeth:
It's been a while since no activity here, so I'm closing this PR. Please feel free to let me know if you're working on it. We can reopen it or you can create a new PR if you fix the [mentioned bug](https://github.com/dbeaver/dbeaver/pull/18588#issuecomment-1523238865).

PR #18590, opened 2022-12-26
dbeaver/pro#1297 Verify real hostname when connecting through jump server

ShadelessFox:


LonwoLonwo (approved):

E1izabeth (approved):

Matvey16:
It is still reproducible to me for SSHJ connections

dbeaver_KglcDRe6Fk

E1izabeth (changes requested):
May be it is not a good idea actually, because it will be unrecognizable if local ports would be unintentionally mapped to some 3rd-party host by other applications, thus guiding our connection to the unwanted destination. The correct solution seems to be the substitution of the localhost with the target host name during the fingerprint validation, so that the fingerprint of the intended remote host would be correctly matched by the existing validation algorithm.

E1izabeth:
Also currently we verify jump server host identity, but doesn't verify ssh target host identity - we are trying to verify it's fingerprint with localhost which is not correct. Skip host key verification when the host is localhost is bad, because then we're not checking target host at all.

The correct solution is still substitution of the localhost with the target host name during the fingerprint validation. It can be done by overriding verifyKey method for sshj and setting hostAlias for jsch.

plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/KnownHostsVerifier.java
```diff
@@ -21,35 +21,45 @@
...
+        this.actualHostConfiguration = actualHostConfiguration;
+    }
+
+    @Override
+    public boolean verify(String hostname, int port, PublicKey key) {
+        if (hostname.equals(DBConstants.HOST_LOCALHOST)) {
+            return super.verify(actualHostConfiguration.getHostname(), actualHostConfiguration.getPort(), key);
+        } else {
+            return super.verify(hostname, port, key);
+        }
```
E1izabeth, line 47:
```suggestion
    @Override
    public boolean verify(String hostname, int port, PublicKey key) {
        if (hostname.equals(DBConstants.HOST_LOCALHOST)) {
            return super.verify(actualHostConfiguration.getHostname(), actualHostConfiguration.getPort(), key);
        } else {
            return super.verify(hostname, port, key);
        }
    }
    
    @Override
    public List<String> findExistingAlgorithms(String hostname, int port) {
        if (localEndPointTargetHostName != null && hostname.equalsIgnoreCase(DBConstants.HOST_LOCALHOST)) {
            return super.findExistingAlgorithms(actualHostConfiguration.getHostname(), actualHostConfiguration.getPort());
        } else {
            return super.findExistingAlgorithms(hostname, port);
        }
    }
        
        
```

plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/jsch/JSCHUIPromptProvider.java
```diff
@@ -73,12 +75,26 @@ public void showMessage(String message) {
...
         private boolean shouldUsePassword() {
-            return configuration.getType().usesPassword() && (configuration.isSavePassword() || CommonUtils.isNotEmpty(configuration.getPassword()));
+            final SSHAuthConfiguration auth = configuration.getAuthConfiguration();
+            return auth.getType().usesPassword() && (auth.isSavePassword() || CommonUtils.isNotEmpty(auth.getPassword()));
         }
 
         @Override
         public boolean promptYesNo(String question) {
-            return DBWorkbench.getPlatformUI().confirmAction(SSHUIMessages.jsch_remote_host_identifier_changed_warning_title, question);
+            // HACK: There's a bug in JSCH: in the message, it includes the original hostname instead of its alias
```
E1izabeth, line 84: Also `showMessage(String message)` should be corrected to fix the message.

E1izabeth (changes requested):

E1izabeth:
Also I've found alive jsch repository https://github.com/mwiede/jsch. Maybe we could migrate to it. (For discussion)

plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/jsch/JSCHUIPromptProvider.java
ShadelessFox: The `showMessage` is used for debug logging. Do we really care about it? I haven't seen it popping up in my console.

plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/KnownHostsVerifier.java
ShadelessFox: And also in literally every other method of the `OpenSSHKnownHosts` class 😞

plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/jsch/JSCHUIPromptProvider.java
E1izabeth: I guess it could write to debug log incorrect message too.

E1izabeth (approved):

ShadelessFox:
The migration is possible, yet the fork still has the same issue regarding confirmation messages: https://github.com/mwiede/jsch/pull/257

E1izabeth:
Yeah, I see. But we could contribute here and it's supported.

ShadelessFox:
I did, in fact. It's not a blocker by any means. We can migrate at any time and eliminate the hack on our side.

E1izabeth:
I'm not sure about the host alias semantics actually. Maybe the correct fix is not to replace `host `to `chost`, but show both of them or check host alas existence first. (Just my thoughts to think about).

LonwoLonwo (approved):

Matvey16:
Verified

PR #18593, opened 2022-12-26
#18556 Update Exasol driver to version 7.1.16

### LINKED ISSUE #18556: Exasol driver cannot be installed automatically (maven:/com.exasol:exasol-jdbc.RELEASE not found) - new base version
### Description

Try to install/configure an Exasol connection (any dummy host entry will do). If Dbeaver tries (or you try in the Driver Manager etc.) to download the driver, you might get the error message "maven:/com.exasol:exasol-jdbc.RELEASE not found" and no driver gets installed.

If you get that problem as well, then the problem is related to changes, where the driver is located since 7.1.16 (I think). Until 7.1.4 the driver was located within the Exasol repo, with 7.1.16 "central" is used:
https://mvnrepository.com/artifact/com.exasol/exasol-jdbc

The underlying issue seems to be, that the "Add connection" wizard uses an older driver version by default and then tries to find the driver. As Exasol has removed their own Maven repo completely during a website redesign, the download fails (e.g. 7.1.14 is given in Maven to be available from https://maven.exasol.com/artifactory/exasol-releases/com/exasol/exasol-jdbc/7.1.14/exasol-jdbc-7.1.14.jar, which does not return anything, now).

Immediate solution: Database -> Driver Manager -> Exasol -> Edit -> Libraries -> double click on com.exasol:exasol-jdbc:RELEASE and change the artifact to:
Artifact Id (unchanged): exasol-jdbc
Group Id (unchanged): com.exasol
Version (updated): 7.1.16
Fallback Version: RELEASE
Click on Ok -> Download/Update (which should work now) -> Close all windows and restart Dbeaver

General solution: The default driver version for the Exasol driver should be increased to at least 7.1.16.

### DBeaver Version

Community Edition 22.3.0

### Operating System

Windows

### Database and driver

Exasol

### Steps to reproduce

Try to install/configure an Exasol connection (any dummy host entry will do). If Dbeaver tries (or you try in the Driver Manager etc.) to download the driver, you might get the error message "maven:/com.exasol:exasol-jdbc.RELEASE not found" and no driver gets installed.

### Additional context

_No response_

LonwoLonwo:
- Update Exasol driver to version 7.1.16
- add new encryption connection setting

2022-12-26 13_40_59-Window


plugins/org.jkiss.dbeaver.ext.exasol.ui/src/org/jkiss/dbeaver/ext/exasol/ui/internal/ExasolResources.properties
```diff
@@ -54,6 +54,8 @@ exasol_partition_name = Partition info
 label_backup_host_list = Backup Host List
 label_database = Database
 label_encrypt = Encrypt Communication
+connection_page_checkbox_legacy_encrypt = Use Legacy Encryption
+connection_page_checkbox_legacy_encrypt_tip = 7.1 driver will always try to establish a TLS connection first, regardless of the database version.\nThis means using a 7.1 driver to connect to any database version requires you to take action to make a connection to the database (see options below).\nIf the database does not fully support TLS (versions < 6.2.15 and < 7.0.10), then the driver will reconnect with ChaCha20.
```
E1izabeth:
```suggestion
connection_page_checkbox_legacy_encrypt_tip = All connections created using Exasol-provided 7.1 drivers will try to establish TLS connections to Exasol per first, regardless of the database version.\nThis means using a 7.1 driver to connect to any database version requires you to take action to make a connection to the database. You can copy fingerprint from the error message to the `fingerprint` driver property and reconnect or use legacy encryption (this option).\nIf the database does not fully support TLS (versions < 6.2.15 and < 7.0.10), then the driver will reconnect with ChaCha20.
```

plugins/org.jkiss.dbeaver.ext.exasol.ui/src/org/jkiss/dbeaver/ext/exasol/ui/internal/ExasolResources.properties
```diff
@@ -54,6 +54,8 @@ exasol_partition_name = Partition info
 label_backup_host_list = Backup Host List
 label_database = Database
 label_encrypt = Encrypt Communication
+connection_page_checkbox_legacy_encrypt = Use Legacy Encryption
+connection_page_checkbox_legacy_encrypt_tip = All connections created using Exasol-provided 7.1 drivers will try to establish TLS connections to Exasol per first, regardless of the database version.\nThis means using a 7.1 driver to connect to any database version requires you to take action to make a connection to the database. You can copy fingerprint from the error message to the `fingerprint` driver property and reconnect or use legacy encryption (this option).\nIf the database does not fully support TLS (versions < 6.2.15 and < 7.0.10), then the driver will reconnect with ChaCha20.
```
E1izabeth:
```suggestion
connection_page_checkbox_legacy_encrypt_tip = All connections created using Exasol-provided 7.1 drivers will try to establish TLS connections to Exasol first, regardless of the database version.\nThis means using a 7.1 driver to connect to any database version requires you to take action to make a connection to the database, if you don't have properly configured certificates. You can copy fingerprint from the error message to the `fingerprint` driver property and reconnect or use legacy encryption (this option).\nIf the database does not fully support TLS (versions < 6.2.15 and < 7.0.10), then the driver will reconnect with ChaCha20.
```

E1izabeth (approved):

ShadelessFox (approved):

uslss:
verified

PR #18598, opened 2022-12-26
#18589 Column header render fix

### LINKED ISSUE #18589: Why change option filter column? Please revert same 22.3.0
Why change option filter column? Please revert same 22.3.0

serge-rider:


Matvey16:
Verified

PR #18605, opened 2022-12-27
#16121 Fix templates context initialization and filtering on insert

### LINKED ISSUE #16121: SQL Editor Templates don't filter based on context
#### System information: 
- Windows 10
- DBeaver 22.0.2.202204041542

In a SQL editor, when I hit `CTRL+ALT+SPACE` to generate the list of Templates to choose from, _all_ templates are shown, rather than just templates that apply to my current connection.

1. Create two templates, each with "context" set to a specific DB engine:
image

1. Open a new SQL Editor tab of _any_ connection

1. Hit `CTRL+ALT+SPACE` and note that all templates are shown, even if the context does not match the current connection.  Also note that the context is not listed next to the template, so if two templates have the same name, you must look through the (unformatted) preview to determine which is the correct one for your current connection.
image

E1izabeth:
- Templates should be correctly saved and restored, while changing its context.
- Only templates with the corresponding context should be proposed and auto inserted. For example, if you opened editor for SQLite connection, you should be able to use for auto insertion templates with SQLite and SQL context, but templates with other contexts, for example, PostgreSQL, should not be available. The same is for templates proposal (Ctrl + Alt + Space).

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorUtils.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorUtils.java
@@ -46,6 +46,8 @@ import org.jkiss.dbeaver.ui.editors.sql.handlers.SQLEditorVariablesResolver;
 import org.jkiss.dbeaver.ui.editors.sql.handlers.SQLNavigatorContext;
 import org.jkiss.dbeaver.ui.editors.sql.internal.SQLEditorActivator;
 import org.jkiss.dbeaver.ui.editors.sql.scripts.ScriptsHandlerImpl;
+import org.jkiss.dbeaver.ui.editors.sql.templates.SQLContextTypeBase;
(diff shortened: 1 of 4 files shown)


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorUtils.java
```diff
@@ -544,4 +546,28 @@ private static void notifyPrefs(@NotNull DBPPreferenceStore prefStore, boolean n
             newServicesEnabled && markWordForSelectionEnabled
         );
     }
+    
+    /**
+     * Returns type id of the driver of data source container, associated with SQLEditor
+     * Returns null if editor is not instance fo SQLEditor or data source container is null
```
ShadelessFox: This will be merged into a single line.

```suggestion
     * Returns type id of the driver of data source container, associated with SQLEditor, or {@code null} if editor is not instance fo SQLEditor or data source container is null
```

ShadelessFox (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #18613, opened 2022-12-27
#17222 Allow quotation of script parameter names

### LINKED ISSUE #17222: BUG - Explain plan bind variables dialog is missing quoted bind variable names
### Description

Database type: ORACLE
Query: [select :1 ,:"SYS_B_1" ,:"MYBVAR8" from dual;]

### Steps to reproduce

Explain plan bind variables dialog does not detect quoted("") bind variable names.

dbeaver-explain

### DBeaver Version

Community Edition Version 22.1.3.202207240747

### Operating System

Windows 10

E1izabeth:
Script parameters should be recognized even in quotes. The example from issue should work.
The same quotes as for identifiers could be used. For example it's both " and ` in mysql.


ShadelessFox (commented):
Judging by the code, this change will allow parameters such as `:""""""` to pass. Is this intentional?

E1izabeth:
You're right. Perhaps, it shouldn't be allowed. I'll try to propose another solution.

test/org.jkiss.dbeaver.test.platform/src/org/jkiss/dbeaver/model/sql/parser/SQLScriptParserGenericsTest.java
```diff
@@ -95,6 +98,23 @@ public void parseFromCursorPositionBeginTransaction() throws DBException {
         }
     }
     
+    @Test
+    public void parseNamedParameters() throws DBException {
+        var expectedParamNames = List.of(":1", ":\"SYS_B_1\"", ":\"MyVar8\"", ":ABC", ":\"#d2\"");
+        var invalidParamNames = List.of("&6^34", ":%#2", ":\"\"\"\"");
+        StringJoiner joiner = new StringJoiner(", ", "select ", " from dual");
+        expectedParamNames.stream().forEach(n -> joiner.add(n));
+        invalidParamNames.stream().forEach(n -> joiner.add(n));
```
ShadelessFox: Why use streams here? If you really want to do it functional, you can use the `forEach` method of the collection itself. But I'd stick with iterator-based loops because:
1. you have mixed code style with different approaches
2. streams/forEach are slower, although it's not a bottleneck in this particular case

test/org.jkiss.dbeaver.test.platform/src/org/jkiss/dbeaver/model/sql/parser/SQLScriptParserGenericsTest.java
```diff
@@ -95,6 +98,23 @@ public void parseFromCursorPositionBeginTransaction() throws DBException {
         }
     }
     
+    @Test
+    public void parseNamedParameters() throws DBException {
+        var expectedParamNames = List.of(":1", ":\"SYS_B_1\"", ":\"MyVar8\"", ":ABC", ":\"#d2\"");
+        var invalidParamNames = List.of("&6^34", ":%#2", ":\"\"\"\"");
+        StringJoiner joiner = new StringJoiner(", ", "select ", " from dual");
```
ShadelessFox: I don't like the way `var` and explicit types are mixed

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/rules/ScriptParameterRule.java
```diff
@@ -28,7 +28,7 @@
 */
 public class ScriptParameterRule implements TPRule {
     
-    private final static char QUOTE_CHAR = '"';
+    private static final char QUOTE_CHAR = '"';
```
ShadelessFox: I'm not sure if it will bring be beneficial rather than problematic, but you can use `SQLDialect#getIdentifierQuoteStrings` and `SQLDialect#getStringQuoteStrings` to build a set of quotes you might want to support.

E1izabeth: I added usage of `SQLDialect#getIdentifierQuoteStrings`.

ShadelessFox (approved):

LonwoLonwo (approved):

Matvey16:
Verified

Matvey16:
Case inside the quotes is ignored when DBeaver saves values
dbeaver_7Y8okZKfC9

E1izabeth:
Quoted parameters are case sensitive now.
Unquoted parameters remains to become upper cased. (Now user can see upper cased parameters in bind parameter dialog, this way it is clear that this parameter is case-insensitive).
Named parameter prefix is dropped to make parameters work between different connections with different prefixes.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLScriptParser.java
```diff
@@ -718,10 +718,22 @@ public static List<SQLQueryParameter> parseParametersAndVariables(SQLParserConte
                             parameters = new ArrayList<>();
                         }
 
+                        String preparedParamName;
+                        String paramMark = paramName.substring(0, 1);
```
ShadelessFox, line 722: Isn't the prefix can be of any length?

E1izabeth: No, it can be only of length 1 for named parameters.

ShadelessFox (approved):

Matvey16:
If similiar parameters with different case are used in the same script, the value is now saved only for the last one (values for previous parameters are still resolved, but not saved)

dbeaver_40wYTyNcw3

Matvey16:
Verified

LonwoLonwo (approved):

PR #18622, opened 2022-12-28
dbeaver/pro#1318 Fix incorrect data type for non-simple select items with aliases as table columns

E1izabeth:
Please, test different cases of queries with and without aliases, including:
- column and alias are the same
- column and alias are not the same, but alias is the same with another column name
- method call with or without some alias
- check select queries for different databases (mysql, oracle, postgresql, sqlite) and no-sql database (mongo or dynamodb)


ShadelessFox (approved):

LonwoLonwo (approved):

uslss:
verified

PR #18653, opened 2023-01-04
Attempt to fix #5450

### LINKED ISSUE #5450: Show start time on query execution results
After query execution (MySQL), it shows a tab with "Statistics" or results (updated rows, query, finish time). Would be great if we could also see the **start time** of the query. It would be also cool to see elapsed time.

bob27aggiustatutto:
Attempt to fix #5450


E1izabeth:
Thank you for contribution.
Maybe I didn't notice, please guide me where the start time is recorded?

bob27aggiustatutto:
into `org.jkiss.dbeaver.model.exec.DBCStatistics` in the `long startTime` field
https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBCStatistics.java#L29
@E1izabeth

E1izabeth:
Yes, but where is the value of this field set?

bob27aggiustatutto:
I don't think I understand your question, maybe here is the answer
https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBCStatistics.java#L42
@E1izabeth

E1izabeth:
Yes, that's what I missed. Thank you.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBCStatistics.java
```diff
@@ -104,6 +104,10 @@ public long getEndTime() {
         return startTime + getTotalTime();
     }
 
+    public long getStartTime (){
```
E1izabeth: Could you add a whitespace here?
```suggestion
    public long getStartTime () {

```

E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBCStatistics.java
```diff
@@ -104,6 +104,11 @@ public long getEndTime() {
         return startTime + getTotalTime();
     }
 
+    public long getStartTime () {
+
+        return this.startTime;
```
ShadelessFox:
```suggestion
    public long getStartTime() {
        return this.startTime;

```

ShadelessFox (approved):

ShadelessFox (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #18668, opened 2023-01-05
#17198 change case sensitivity condition for generic schema cache

### LINKED ISSUE #17198: SAP hana database schema case sensitive issue
#### System information: 
  Windows 11
- DBeaver version
  22.1.3
- Additional extensions

- Database name and version
  SAP HANA EXPRESS 2.0
- Driver name
  default sap hana jdbc driver
  no

  when connect SAP hana, there are two schemas, one is name "MES" (uppercase), another one is "mes" (lowercase), no matter which one I select, the "mes"( lowercase one) always be selected, and I try to hide the "mes", both are hide. so I have no way to  selected the uppercase one.

image

LonwoLonwo:
The main problem is here:
https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/cache/AbstractObjectCache.java#L310

If we have caseSensitive flag as false (like we have for the schema cache), we do not have "test" and "TEST" schemas in the objectMap, only one "TEST" as a key.

So IMHO it was a mistake condition.

Postgres has a Mixed case and has false as caseSensitive
Oracle has upper case and true as caseSensitive
And this: https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/cache/AbstractObjectCache.java#L264

Please check:
- That schemas list still can have "TEST" and "test"
- "TEST" can be chosen as default from SQL Editor
- "test" can be chosen as default from SQL Editor
- "TEST" can be chosen as default from the navigator tree
- "test" can be chosen as default from the navigator tree

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericDataSource.java
+++ b/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericDataSource.java
@@ -527,7 +527,7 @@ public class GenericDataSource extends JDBCDataSource implements DBPTermProvider
                     List<GenericSchema> tmpSchemas = metaModel.loadSchemas(session, this, null);
                     if (tmpSchemas != null) {
                         this.schemas = new SimpleObjectCache<>();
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

Destrolaric (approved):

uslss:
verified

PR #18669, opened 2023-01-05
#17739 don't suggest hippie after dot

### LINKED ISSUE #17739: Autocompletion doubles table alias
### Description

dbeaver_autocompletion
The alias of a table is doubled by autocompletion when the FROM clause already exists and when the column is already used in the SELECT list.

### DBeaver Version

Community Edition 22.2.1

### Operating System

Ubuntu 22.04

### Database and driver

PostgreSQL 14.5
PostgreSQL JDBC driver 42.3.4.jar

### Steps to reproduce

1. Write the FROM clause with a table and an alias
2. Write the SELECT list with the alias of the table and a column.
3. Write again the alias of the table as a new selected column : the preceding column is proposed by autocompletion. If we accept the autocompletion, the table alias is doubled. 

### Additional context

_No response_

Destrolaric:
I am not sure that there are cases when we need this. 
Close #17739


LonwoLonwo (approved):

E1izabeth (approved):

uslss:
verified

PR #18681, opened 2023-01-05
#8647 use case insensitive schemas search for MIXED case databases

### LINKED ISSUE #8647: Autocomplete does not work if the schema name in script has different case than in navigator
#### System information: 
- Operating system (distribution) and version: Win 10
- DBeaver version: 7.0.4

- Database name and version: Sybase IQ SAP IQ/16.1.030.1437
- Driver: jConnect (TM) for JDBC(TM)/16.0 SP03 PL02 (Build 27403)/P/EBF27518/JDK 1.6.0/jdbcmain/OPT/Mon Aug 28 18:41:14 PDT 2017

Autocomplete does not give suggestions of the tables if the schema name in script has different case than in navigator.

I.e. if in Navigator schema name is DB, then when typing in script db. then autocomplete does not give suggestions of the tables under the schema

LonwoLonwo:
```java
@Nullable
    public static <T extends DBPNamedObject> T findObject(@Nullable Collection<T> theList, String objectName, boolean caseInsensitive) {
        if (theList != null && !theList.isEmpty()) {
            for (T object : theList) {
                if (caseInsensitive ? object.getName().equalsIgnoreCase(objectName) : object.getName().equals(objectName)) {
                    return object;
                }
            }
        }
        return null;
    }
```

to not use caseInsensitive as false always for schema searching
IMHO, Mixed case databases usually do not have the ability to create objects with the same name (like PG or Sybase). If you have schema "test," you can not create "TEST".

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericCatalog.java
+++ b/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericCatalog.java
@@ -19,6 +19,7 @@ package org.jkiss.dbeaver.ext.generic.model;
 import org.jkiss.code.NotNull;
 import org.jkiss.code.Nullable;
 import org.jkiss.dbeaver.DBException;
+import org.jkiss.dbeaver.model.DBPIdentifierCase;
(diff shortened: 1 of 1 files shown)


ShadelessFox (approved):

E1izabeth (approved):

uslss:
verified

PR #18692, opened 2023-01-09
CB-3024-cb-aws-user-authentication-failed-error-reading-user-permissions

alexander-skoblikov:


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/schema/SQLSchemaManager.java
```diff
@@ -172,7 +171,16 @@ public void save() throws IOException {
...
-                dbStat.execute(line);
+                try {
+                    dbStat.execute(line);
+                } catch (SQLException e) {
+                    //TODO: find a better way to avoid migration errors
+                    // 11 migration version sometimes crashes in h2
+                    log.error("Error during sql migration", e);
+                    log.info("Trying to apply the migration again");
+                    dbStat.execute(line);
+                    log.error("The second attempt was successful");
```
LonwoLonwo, line 182: And if not?)

alexander-skoblikov: then we are sad, cry and go to repair the base using hands 🥲

LonwoLonwo (approved):

PR #18693, opened 2023-01-09
#17350 add triggers creation statements to view DDL for PostgreSQL

### LINKED ISSUE #17350: trigger not showed on a view ddl
### Description

open ddl tab on view. You cannot see create trigger code in creation script

### DBeaver Version

22.1.4

### Operating System

windows 10

### Database and driver

Postgresql 14.2

### Steps to reproduce

1. Create view
CREATE OR REPLACE VIEW v
AS SELECT...
2. create trigger on this view
create trigger trg_v_change instead of...
3. open ddl tab on view v. You cannot see create trigger code in creation script

### Additional context

_No response_

LonwoLonwo:
2023-01-09 14_08_25-DBeaver 22 3 2 - customer_projects_view

Statements to check: https://www.dbi-services.com/blog/can-i-do-it-with-postgresql-18-instead-of-triggers-on-views/

- [x] View with trigger has the correct DDL
- [x] View with triggers has the correct DDL
- [x] View without trigger has the correct DDL
- [x] Check some other random PG-like databases (EDB, Greenplum, etc.)

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreViewBase.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreViewBase.java
@@ -27,6 +27,7 @@ import org.jkiss.dbeaver.model.edit.DBEPersistAction;
 import org.jkiss.dbeaver.model.exec.jdbc.JDBCSession;
 import org.jkiss.dbeaver.model.impl.DBObjectNameCaseTransformer;
 import org.jkiss.dbeaver.model.impl.edit.SQLDatabasePersistAction;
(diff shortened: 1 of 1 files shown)


E1izabeth (commented):
Why create trigger is in lowercase while keywords in other statements are in upper case? Can we make it consistent?

ShadelessFox (approved):

LonwoLonwo:
@E1izabeth, we are reading triggers with the PG function and using formatting on it already:

https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreTrigger.java#L262

And the "Create trigger" statement is in the lower case in the end because of PostgreDialect case settings.

E1izabeth (approved):

uslss:
checked in cb also.
verified

PR #18695, opened 2023-01-09
pro#1318 Fix attribute binding resolution

E1izabeth:
At least following should be tested.

- [x] Data read.
Queries with and without aliases, including:
- [x] column and alias are the same
- [x]  column and alias are not the same, but alias is the same with another column name
- [x] method call with or without some alias
- [x] select queries for different databases (mysql, oracle, postgresql, sqlite) and no-sql database (mongo or dynamodb)
- [x] non-select queries
- [x] select * from x;


ShadelessFox (approved):

LonwoLonwo (approved):

uslss:
verified

PR #18699, opened 2023-01-09
dbeaver/pro#1201 Allow disabling values completion

ShadelessFox:
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/preferences/PrefPageSQLCompletion.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/preferences/PrefPageSQLCompletion.java
@@ -56,6 +56,7 @@ public class PrefPageSQLCompletion extends TargetPrefPage
     private Button csInsertSpace;
     private Button csSortAlphabetically;
     private Button csShowServerHelpTopics;
+    private Button csShowValues;
     private Combo csInsertTableAlias;
 
     private Button csMatchContains;
@@ -90,7 +91,8 @@ public class PrefPageSQLCompletion extends TargetPrefPage
             store.contains(SQLPreferenceConstants.PROPOSALS_MATCH_CONTAINS) ||
             store.contains(SQLPreferenceConstants.USE_GLOBAL_ASSISTANT) ||
             store.contains(SQLPreferenceConstants.SHOW_COLUMN_PROCEDURES) ||
-            store.contains(SQLPreferenceConstants.SHOW_SERVER_HELP_TOPICS)
+            store.contains(SQLPreferenceConstants.SHOW_SERVER_HELP_TOPICS) ||
+            store.contains(SQLPreferenceConstants.SHOW_VALUES)
         ;
     }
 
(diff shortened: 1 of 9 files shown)


E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/preferences/PrefPageSQLCompletion.java
```diff
@@ -150,6 +152,7 @@ protected Control createPreferenceContent(@NotNull Composite parent) {
             csInsertSpace = UIUtils.createCheckbox(assistGroup, SQLEditorMessages.pref_page_sql_completion_label_insert_space, null, false, 2);
             csSortAlphabetically = UIUtils.createCheckbox(assistGroup, SQLEditorMessages.pref_page_sql_completion_label_sort_alphabetically, null, false, 2);
             csShowServerHelpTopics = UIUtils.createCheckbox(assistGroup, SQLEditorMessages.pref_page_sql_completion_label_show_server_help_topics, SQLEditorMessages.pref_page_sql_completion_label_show_server_help_topics_tip, false, 2);
+            csShowValues = UIUtils.createCheckbox(assistGroup, SQLEditorMessages.pref_page_sql_completion_label_show_values, SQLEditorMessages.pref_page_sql_completion_label_show_values_tip, false, 2);
```
LonwoLonwo, line 155: It does not need to be false by default, IMHO.

E1izabeth: I agree with LonwoLonwo.

ShadelessFox: It's assigned later to the value from the preference store:

```java
csShowValues.setSelection(store.getBoolean(SQLPreferenceConstants.SHOW_VALUES));
```

LonwoLonwo: Ok, but wy it has an unchecked status here? What is the reason for this inconsistency?

ShadelessFox: Who cares about its initial value? I just copied the line above.

LonwoLonwo: I care. This is one of the places where I can see the default value. I need an extra click to find SQLPreferenceConstants and its default value.

ShadelessFox: Well, the initial state of a GUI control is not supposed to represent the initial value of the preference it modifies and you must not rely on it, so in the end, you will have to find where the preference's default value is initialized.

uslss:
verified

PR #18704, opened 2023-01-09
#18553 Get host and port from url for native tools if connection is by url

### LINKED ISSUE #18553: Dbeaver does not backup Postgresql if the connection details are in the Url.
### Description

Dbeaver does not backup Postgresql if the connection details are in the Url. 
And that is selected to connect with. 

It will try to get the connection details from the Host, Port and Database variables.

### DBeaver Version

Community Edition 22.3.0

### Operating System

Windows 10

### Database and driver

1. Driver : Postgresql JDBC driver
2. Database: Postgresql 11.15

### Steps to reproduce

1. Try to connect to a postgress database with a URL which is not the default. (A dns url, which is not localhost.)
2. Do not change the default Host, Database and Port variables. 
3. Connection succeeds
4. Try to backup the database with the tools. 
5. Its says it fails with connecting to the localhost, which is not in the Url, but defined in the Host.

### Additional context

_No response_

E1izabeth:
Dump and Restore should work for PostgreSQL and MySQL when connection was configured by specifying host and port and when connection was configured by url,

### DIFF
--- a/plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/NativeToolUtils.java
+++ b/plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/NativeToolUtils.java
@@ -102,7 +102,7 @@ public abstract class NativeToolUtils {
     }
     
     /**
-     * Appends <code>--host</code> and <code>--port</code> parameters from connection settings to the list of command parameters
+     * Appends {@code --host} and {@code --port} parameters from connection settings to the list of command parameters
      */
     public static void addHostAndPortParamsToCmd(@NotNull DBPDataSourceContainer dataSourceContainer, @NotNull List<String> cmd) {
         DBPConnectionConfiguration connectionInfo = dataSourceContainer.getActualConnectionConfiguration();
--- a/plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/NativeToolUtils.java
+++ b/plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/NativeToolUtils.java
@@ -17,14 +17,23 @@
 package org.jkiss.dbeaver.tasks.nativetool;
 
 import org.jkiss.code.NotNull;
+import org.jkiss.dbeaver.Log;
+import org.jkiss.dbeaver.model.DBPDataSourceContainer;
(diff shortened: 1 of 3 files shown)


LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/NativeToolUtils.java
```diff
@@ -91,4 +100,39 @@ public static String replaceVariables(@NotNull String name) {
...
+                try {
+                    URI uri = URI.create(url);
+                    hostname = uri.getHost();
+                    port = String.valueOf(uri.getPort());
+                } catch (Exception e) {
+                    log.error("Can't parse connection URL", e);
+                }
+            }
+        }
+        cmd.add("--host=" + hostname);
```
ShadelessFox, line 133: What if `hostname` is empty, too?

plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/NativeToolUtils.java
```diff
@@ -91,4 +100,39 @@ public static String replaceVariables(@NotNull String name) {
                 return System.getProperty(name);
         }
     }
+    
+    /**
+     * Appends <code>--host</code> and <code>--port</code> parameters from connection settings to the list of command parameters
```
ShadelessFox:
```suggestion
     * Appends {@code --host} and {@code --port} parameters from connection settings to the list of command parameters
```

plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/NativeToolUtils.java
E1izabeth: It was not supposed to be empty. I decided to let this behaviour be untouched.

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/NativeToolUtils.java
E1izabeth: Actually, I think it's not possible as it's not a valid URI anymore.

Matvey16:
Verified

PR #18711, opened 2023-01-10
#18706 Support quoted variables in set, fix case sensitivity

### LINKED ISSUE #18706: variables don't bind when using lowercase in the naming
### Description

When using the DBeaver @set command to set a variable, when executing the script the parameter binding is looking for the variable name in UPPER CASE.  The variables tab shows the variable created using same case as the @set command.  

### DBeaver Version

Version 22.3.2.202301060755

### Operating System

macOS 13.0.1

### Database and driver

PostgreSQL
PostgreSQL JDBC Driver

### Steps to reproduce

1. Open a new script
2. set a variable using @set, with name using some mix of uppper and lower case letters
3. use the variable in a SELECT statement and execute

Example of above
`@set userIdList = (SELECT "Id" FROM public."AspNetUsers" WHERE "UserName" IN ('greg@perf.labs', 'BickBick@perf.labs', 'bick2@perf.labs'))

-- title: AspNetUsers
SELECT
	*
FROM
	"AspNetUsers" anu
WHERE
	"Id" IN (:userIdList)
ORDER BY
	anu."UserName"
;`
parameter bind
Variables tab

**Expected results**
variables with any mix of upper and lower case letters will be recognized during script run so long as the set and the call match up

**Actual results**
Bind parameters during script execution is taking variable names and looking for the upper case version of the variable name.  Since variable was set using mix case the variable's value is not populated at run time and a pop up is presented for variable binding.  

If the variable is name with all upper case during the @set command and called using : later in the script with mixed case the variable replacement happens successfully.  

### Additional context

This was working properly on DBeaver version 22.3.1 and looks to have broken in 22.3.2

E1izabeth:
@set command should allow quotation depends on the dialect. IF variable is qouted then it's name is case sensitive. If variable is unquoted then it is case-insensitive.

The following script provides minimal set of cases to check.
```
@set aBc = 1
@set "aBc" = 2

select :aBc from Album a; -- 1
select :"ABC" from Album a; -- 1
select :ABC from Album a; -- 1
select :abc from Album a; -- 1
select :"aBc" from Album a; -- 2
select :"abc" from Album a; -- asks for binding
```

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/commands/SQLCommandSet.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/commands/SQLCommandSet.java
@@ -16,6 +16,7 @@
  */
 package org.jkiss.dbeaver.model.sql.commands;
 
+import org.jkiss.code.NotNull;
 import org.jkiss.dbeaver.DBException;
 import org.jkiss.dbeaver.model.exec.DBCException;
 import org.jkiss.dbeaver.model.sql.SQLControlCommand;
@@ -51,15 +52,15 @@ public class SQLCommandSet implements SQLControlCommandHandler {
         return true;
     }
     
-    private String tryParseVariableName(SQLScriptContext ctx, String text, int start) {
+    private String tryParseVariableName(@NotNull SQLScriptContext ctx, @NotNull String text, int start) {
         while (start < text.length() && Character.isWhitespace(text.charAt(start))) {
             start++;
         }
         SQLDialect sqlDialect = ctx.getExecutionContext().getDataSource().getSQLDialect();
         int end = ScriptParameterRule.tryConsumeParameterName(sqlDialect, text, start);
-        if (end > 0) {
+        if (end != -1) {
             String rawParamName = text.substring(start, end);
-            if (end > 0 && sqlDialect.isQuotedIdentifier(rawParamName)) {
+            if (sqlDialect.isQuotedIdentifier(rawParamName)) {
                 return sqlDialect.getUnquotedIdentifier(rawParamName);
             } else {
                 return rawParamName.toUpperCase(Locale.ENGLISH);
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/commands/SQLCommandSet.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/commands/SQLCommandSet.java
@@ -20,9 +20,13 @@ import org.jkiss.dbeaver.DBException;
 import org.jkiss.dbeaver.model.exec.DBCException;
 import org.jkiss.dbeaver.model.sql.SQLControlCommand;
 import org.jkiss.dbeaver.model.sql.SQLControlCommandHandler;
+import org.jkiss.dbeaver.model.sql.SQLDialect;
 import org.jkiss.dbeaver.model.sql.SQLScriptContext;
+import org.jkiss.dbeaver.model.sql.parser.rules.ScriptParameterRule;
 import org.jkiss.dbeaver.utils.GeneralUtils;
 import org.jkiss.utils.CommonUtils;
+
+import java.util.Locale;
 
 /**
  * Control command handler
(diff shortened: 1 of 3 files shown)


ShadelessFox (approved):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/commands/SQLCommandSet.java
```diff
@@ -32,18 +36,36 @@ public class SQLCommandSet implements SQLControlCommandHandler {
...
+    
+    private String tryParseVariableName(SQLScriptContext ctx, String text, int start) {
+        while (start < text.length() && Character.isWhitespace(text.charAt(start))) {
+            start++;
+        }
+        SQLDialect sqlDialect = ctx.getExecutionContext().getDataSource().getSQLDialect();
+        int end = ScriptParameterRule.tryConsumeParameterName(sqlDialect, text, start);
+        if (end > 0) {
+            String rawParamName = text.substring(start, end);
+            if (end > 0 && sqlDialect.isQuotedIdentifier(rawParamName)) {
```
ShadelessFox: The `end > 0` expression is always true

```suggestion
            if (sqlDialect.isQuotedIdentifier(rawParamName)) {
```

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/commands/SQLCommandSet.java
```diff
@@ -32,18 +36,36 @@ public class SQLCommandSet implements SQLControlCommandHandler {
...
-    ;
-
+    
+    private String tryParseVariableName(SQLScriptContext ctx, String text, int start) {
+        while (start < text.length() && Character.isWhitespace(text.charAt(start))) {
+            start++;
+        }
+        SQLDialect sqlDialect = ctx.getExecutionContext().getDataSource().getSQLDialect();
+        int end = ScriptParameterRule.tryConsumeParameterName(sqlDialect, text, start);
+        if (end > 0) {
```
ShadelessFox: Is it possible, and if so, valid, to have `end` be equal to `0`?

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/commands/SQLCommandSet.java
```diff
@@ -32,18 +36,36 @@ public class SQLCommandSet implements SQLControlCommandHandler {
...
-        scriptContext.setVariable(varName, varValue);
+        scriptContext.setVariable(varName.trim(), varValue);
 
         return true;
     }
-
-    ;
-
+    
+    private String tryParseVariableName(SQLScriptContext ctx, String text, int start) {
```
ShadelessFox: Please specify nullability annotations for the method's parameters and its return type.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/rules/ScriptParameterRule.java
```diff
@@ -117,7 +118,33 @@ && isIdentifierQuote(buffer.charAt(buffer.length() - 1), false, true)
...
+            while (position < buffer.length()) {
+                c = buffer.charAt(position);
+                position++;
+                if (isIdentifierQuote(c, false, true, quoteStrings)) {
+                    return position;
+                }
+            }
+            return -1;
+        } else {
+            while (position < buffer.length() && Character.isJavaIdentifierPart(c)) {
```
ShadelessFox, line 139: Can we rely on `SQLDialect#validIdentifierStart` and `SQLDialect#validIdentifierPart` instead?

E1izabeth: We parse unquoted variable name here, so it will not be correct to do so.

ShadelessFox (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #18724, opened 2023-01-10
#18716 create connection URL by template for non default databases

### LINKED ISSUE #18716: A bug with "Connect by URL" causes a schema exception
### Description

1 use PostgreSQL，and the server choose "connect by url" ，like jdbc:postgresql://localhost:5432/database_A
3. If you perform operations on databases other than database_A, all the data you see is from the database_A database. The same is true for delete operations. For example, if there is a database database_B, you will see the database_A data under the database_B menu If you delete the schema, the database_A data will be deleted. This is very misleading and can lead to serious bugs if applied to production environments

### DBeaver Version

CE 22.3.2

### Operating System

Win 10

### Database and driver

PostgreSQL 12.5

### Steps to reproduce

1. connect to the database using connect by mode
2. Enter the URL, for example, JDBC: postgresql: / / localhost: 5432 / database_A
3. Create a database, such as database_B
4. Open the database_B menu
5. You will see the data in database_A
6. Now perform the schema deletion
7. Open the database_A menu
8. You will see that database_A is missing that schema

### Additional context

_No response_

LonwoLonwo:
2023-01-10 19_58_46-DBeaver 22 3 2 - _database_a_ Script-95

Basically, I just copied logic from `getConnectionURL(conConfig)` method, yes. I do not see any important reason to create a new method in the driver class.

And probably we do not need

```java
if (CommonUtils.isEmpty(newURL)) {
                            newURL = driver.getDataSourceProvider().getConnectionURL(driver, conConfig);
}
```

but I'm not sure.

Important condition is higher:

!CommonUtils.equalObjects(instance.getName(), conConfig.getDatabaseName())

So, please check the following:

- [x] case from the ticket
- [x] and any other cases which come to your mind about opening a connection for PG when the new connection will have different from the connection settings database name

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
@@ -486,7 +486,12 @@ public class PostgreDataSource extends JDBCDataSource implements DBSInstanceCont
                     // Patch URL with new database name
                     if (CommonUtils.isEmpty(conConfig.getUrl()) || !CommonUtils.isEmpty(conConfig.getHostName())) {
                         conConfig.setDatabaseName(instance.getName());
-                        conConfig.setUrl(getContainer().getDriver().getConnectionURL(conConfig));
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

ShadelessFox (approved):

uslss:
verified

PR #18761, opened 2023-01-13
#18706 Support quotes in unset command, manual variables addition via UI, fix ${var} handling

### LINKED ISSUE #7938: Placeholders in PREPARE statement prompt for bind parameters
#### System information: 
- Elementary OS 5.1.2 (Ubuntu 18.04.3)
- DBeaver 6.3.5
- No additional extensions

- DB: Greenplum 5.22
- Driver: org.postgresql:postgresql:RELEASE 42.2.10
- No tunnels or proxies

Placeholders in `PREPARE` statement prompt for bind parameters.

```SQL
PREPARE query (INT) AS
  SELECT $1, :param
;
```
will prompt to bind both `1` and `param`, where only prompt for `param` is expected.

### LINKED ISSUE #18706: variables don't bind when using lowercase in the naming

E1izabeth:
- case sensitive variable name works only in `@set`, `@unset` and after `:`
You can do
```
@set "xXx" = 1 //interpreted as xXx
@set xxx = 2 // interpreted as XXX
select :"xXx", :xxx, :xXx, :XxX from dual; // interpreted as select 1, 2, 2, 2 from dual;
@unset "xXx"
@unset xxx
```
- "+" in Variables panel accepts quoted variable names and interpret them as case sensitive
- OK button in Add variable dialog is disabled while variable name is invalid - unquoted variable name can't contain symbols like @, quoted variable name can contain such symbols. Also `"te@st"n` is invalid variable name as it contains character after second quote.

- `${var}` variables should work correct and variable name interpreted as in upper case (case insensitive). Quoted variable names are not supported here.
```
@set xxx = 2 // interpreted as XXX
select ${xxx} from dual;
```

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/commands/SQLCommandSet.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/commands/SQLCommandSet.java
@@ -36,37 +36,37 @@ public class SQLCommandSet implements SQLControlCommandHandler {
 
     @Override
     public boolean handleCommand(SQLControlCommand command, SQLScriptContext scriptContext) throws DBException {
-        String parameter = command.getParameter();
-        String varName = tryParseVariableName(scriptContext, parameter, 0);
-        if (varName == null) {
+        SQLDialect sqlDialect = scriptContext.getExecutionContext().getDataSource().getSQLDialect();
+        String parameter = command.getParameter().stripLeading();
+        int varNameEnd = ScriptParameterRule.tryConsumeParameterName(sqlDialect, parameter, 0);
+        if (varNameEnd == -1) {
             throw new DBCException("Missing variable name. Expected syntax:\n@set varName = value or expression");
         }
-        int divPos = parameter.indexOf('=', varName.length());
+        String varName = prepareVarName(sqlDialect, parameter.substring(0, varNameEnd));
+        int divPos = parameter.indexOf('=', varNameEnd);
         if (divPos == -1) {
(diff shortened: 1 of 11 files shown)


ShadelessFox (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #18780, opened 2023-01-16
#17781 Fix focus on right mouse clicking by result set

### LINKED ISSUE #17781: SQL Editor : Context menu in result area shows disabled "Copy" function after first select
### Description

"Copy" function in the context menu of the SQL Editor result set is disabled on "first select". 

### DBeaver Version

Community  22.2.1.202209251341

### Operating System

Windows 10 64-Bit 

### Database and driver

Azure SQL Server 

Driver 
image

image

### Steps to reproduce

Perform a select in the SQL editor that gives back a result set. 
Right click on one of the cells in the result set...

image

=> In the context menu the "Copy" is disabled. 

If you now click "somewhere else" ( not sure what exactly is the trick) ...and then again do the right click on the same cell....

image

now the "Copy" is enabled...

=> You also have to "click" around to do the copy... Not a super big problem, but a little annoying

### Additional context

_No response_

E1izabeth:


ShadelessFox (commented):
I believe we can transfer focus upon any button press, not just the right button.

ShadelessFox (approved):

Matvey16:
Verified

PR #18781, opened 2023-01-16
#16397 ability to disable metadata objects comment reading 

### LINKED ISSUE #16397: Vertica - Please include option to disable comments altogether
This is related to issue #14625 . 
Reading from comments system table (even though it is not populated in our DB) takes a long time and is I/O intensive.
I am on version 22.0.4.202205011839. 
Please provide an option to disable comments altogether as it takes 14s in test env and over a min in a busy prod cluster.

LonwoLonwo:
- the ability to disable metadata objects comment reading for performance-increasing added in the connection settings
- Vertica connection page logo updated

2023-01-18 14_29_02-DBeaver 22 3 2 - _master_ Script-29

So. Many things to check.

- [x] The new setting is available on the Vertica Connection Page
- [x] The new setting saves after DBeaver restart
- [x] Table comment + columns comments are visible in the lazy mode if the setting is OFF
- [x] View comment + columns comments are visible in the lazy mode if the setting is OFF
- [x] Table comment + columns comments are NOT visible in the lazy mode if the setting is ON
- [x] View comment + columns comments are NOT visible in the lazy mode if the setting is ON
- [x] A constraint comment is visible if the setting is OFF
- [x] A constraint comment is NOT visible if the setting is ON
- [x] A sequence comment is visible if the setting is OFF
- [x] A sequence comment is NOT visible if the setting is ON
- [x] A list of projections can be read if the setting is OFF
- [x] A list of projections can be read if the setting is ON
- [x] Metadata search is working for all types of objects if the setting is OFF
- [x] Metadata search is working for all types of objects if the setting is ON
- [x] Metadata search IN COMMENTS is working for all types of objects if the setting is OFF (note: search in comments doesn't work by default in Vertica for table/views columns)
- [x] Metadata search IN COMMENTS is NOT working for all types of objects if the setting is ON - expected


E1izabeth:
I don't like it on the main page. Let's create a separate page for this option like we did for PostgreSQL or Snowflake.

plugins/org.jkiss.dbeaver.ext.vertica.ui/src/org/jkiss/dbeaver/ext/vertica/ui/internal/VerticaUIMessages.properties
```diff
@@ -0,0 +1,3 @@
+connection_page_group_performance = Performance
+connection_page_group_checkbox_disable_comments = Disable objects comments reading
+connection_page_group_checkbox_disable_comments_tip = Disable objects (tables/sequences/projections, etc.) comments reading.\nv_catalog.comments system table reading can dramatically reduce data loading speed.
```
ShadelessFox: I don't know if including such a piece of technical information has benefits.

plugins/org.jkiss.dbeaver.ext.vertica/src/org/jkiss/dbeaver/ext/vertica/model/VerticaDataSource.java
```diff
@@ -207,6 +212,15 @@ public boolean isChildCommentColumnAvailable(@NotNull DBRProgressMonitor monitor
         return childObjectColumnAvailable;
     }
 
+    /**
+     *
+     * v_catalog.comments dramatically reduces data loading speed. User can disable metadata objects comments reading to avoid it.
+     */
+    boolean avoidCommentsReading() {
```
ShadelessFox, line 219: Why this method uses the verb `avoid` but the property uses `disable`?

plugins/org.jkiss.dbeaver.ext.vertica/src/org/jkiss/dbeaver/ext/vertica/model/VerticaStructureAssistant.java
```diff
@@ -77,15 +78,21 @@ public DBSObjectType[] getAutoCompleteObjectTypes() {
     @Override
     protected void findObjectsByMask(@NotNull JDBCExecutionContext executionContext, @NotNull JDBCSession session, @NotNull DBSObjectType objectType, @NotNull ObjectsSearchParams params, @NotNull List<DBSObjectReference> references) throws DBException, SQLException {
         GenericSchema parentSchema = params.isGlobalSearch() ? null : params.getParentObject() instanceof GenericSchema ? (GenericSchema) params.getParentObject() : null;
+        boolean avoidCommentsReading = false;
+        JDBCDataSource dataSource = executionContext.getDataSource();
+        if (dataSource instanceof VerticaDataSource) {
```
ShadelessFox, line 83: Is it possible for `dataSource` not to be an instance of `VerticaDataSource`?

You don't really mind this possibility in other places.

LonwoLonwo: I'm not sure. The metadata search page can be unpredictable.

LonwoLonwo:
@E1izabeth Extra page for one setting? SQLServer, Sybase, Teradata, and Netezza already use this type of connection settings view.

plugins/org.jkiss.dbeaver.ext.vertica.ui/src/org/jkiss/dbeaver/ext/vertica/ui/internal/VerticaUIMessages.properties
LonwoLonwo: Yeah, maybe. I thought about it. Ok.

E1izabeth:
Snowflake has one setting on such page too, however it's ok

E1izabeth:
I still don't think that we should add such settings for advanced users on the main page of the connection settings. From my point of view, it makes ui more complex and scary users who is new in DBeaver.
However, for the case if other team members (@ShadelessFox @Destrolaric) think that it's ok to have this setting here, I don't block it.

LonwoLonwo:
More complex page VS users will not find it on the extra page
We don't have a perfect solution here
¯\_(ツ)_/¯

ShadelessFox:
I bet users will find it even on a separate page.

I don't like the fact that along with settings related strictly to the connection process, there is a checkbox that somehow improves metadata load performance.

I believe it should be moved into a separate page and contained within the "metadata" group rather than "performance", but with a brief description in its tooltip of what it's supposed to do.

LonwoLonwo:
🙄

LonwoLonwo:
2023-01-18 14_13_26-DBeaver 22 3 2 - _postgres_ Script-34

Oracle has a Performance group, the ticket is about performance, so it is still performance.

ShadelessFox (approved):

E1izabeth (approved):

Matvey16:
Verified

PR #18844, opened 2023-01-23
dbeaver/pro#1371 remove "typcategory" column recognition for Redshift

LonwoLonwo:
aux for https://github.com/dbeaver/dbeaver-ee/pull/2232


Matvey16:
Reproduced in CE on the latest Redshift driver

Matvey16:
Correction: Driver doesn't matter. It is reproducible in CE for Production connection after clicking `Test connection`
But you can still connect to a database, it's the test that is failing

PR #18845, opened 2023-01-23
#18707 Quote identifiers starts with numbers

### LINKED ISSUE #18707: Unable to switch to certain mysql databases with SQL syntax error
### Description

Hi there, 

I have a MySQL server with a number of databases, I'm able to access all but two of them without issue in DBeaver. However upon trying to switch, list or do anything with two of these database I'm presented with an SQL syntax error from DBeaver.

In the example below I'm trying to list the tables in this database.

image

This happens for one other database in this MySQL server also.

### DBeaver Version

Community 22.3.2.202301060755

### Operating System

Windows 10 Pro 19045.2364

### Database and driver

1. MySQL 8.0.18
2. MySQL Connector 8.0.29

### Steps to reproduce

1. Connect to the MySQL database server
2. Attempt to do anything in DBeaver with this specific database like list the tables, or make it the active database.
3. Get SQL syntax error

### Additional context

If I use the mysql cli I'm able to do everything with this database and the tables within it.

E1izabeth:
Check that DBeaver can correctly work (switch default database, for example) with
- identifiers like `1234`
- identifiers like `table1234`
- identifiers like `_fs%sfn`
- identifiers like `66e60ce`

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
@@ -33,6 +33,7 @@ import org.jkiss.utils.ArrayUtils;
 
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
```diff
@@ -125,6 +126,8 @@ public class MySQLDialect extends JDBCSQLDialect {
         "ST_POINTFROMTEXT",
         "ST_POLYFROMTEXT"
     };
+    
+    private static final Pattern QUOTED_IDENTIFIER_PATTERN = Pattern.compile("[0-9]+");
```
ShadelessFox: Kinda a weird name keeping in mind it matches digits 🤔

ShadelessFox (approved):

LonwoLonwo (approved):

uslss:
verified

PR #18858, opened 2023-01-24
Cb 3072 rm metadata for removed script still exists

EmYar:


ShadelessFox:
Bad merge commits, please drop unrelated commits and rebase from devel

ShadelessFox (dismissed):
Bad merge commits, please drop unrelated commits and rebase from devel

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/BaseProjectImpl.java
```diff
@@ -588,41 +588,36 @@ protected IStatus run(DBRProgressMonitor monitor) {
...
-                                    }
-                                    jsonWriter.endObject();
-                                }
-                                jsonWriter.endObject();
-
-                                jsonWriter.endObject();
-                                jsonWriter.flush();
+                try (Writer mdWriter = Files.newBufferedWriter(mdFile, StandardCharsets.UTF_8);
+                     JsonWriter jsonWriter = METADATA_GSON.newJsonWriter(mdWriter)
+                ) {
```
ShadelessFox, line 593: You don't use `mdWriter` anywhere, so it's redundant.

Additionally, you don't have to call it `jsonWriter`, it's too mouthful and specific in this case, so call it `writer` instead.

```suggestion
                try (JsonWriter writer = METADATA_GSON.newJsonWriter(Files.newBufferedWriter(mdFile, StandardCharsets.UTF_8))) {
```

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/BaseProjectImpl.java
```diff
@@ -588,41 +588,36 @@ protected IStatus run(DBRProgressMonitor monitor) {
...
+                            } else {
+                                jsonWriter.value(CommonUtils.toString(value));
                             }
                         }
+                        jsonWriter.endObject();
                     }
+                    jsonWriter.endObject();
 
+                    jsonWriter.endObject();
+                    jsonWriter.flush();
```
ShadelessFox, line 620: This `flush` call is redundant. The `close` call guarantees to flush all pending data before closing the writer.

ShadelessFox (approved):

PR #18862, opened 2023-01-24
#18354 Keep variable rule before parameter rule

### LINKED ISSUE #18354: Bind parameters feature on `$P{ABCD}` variables
### Description

- Previously parameters formated like `$P{ABCD}`  were individually indentified in the "Bind parameter(s)" window.
- Now only the `$P` is identified as a variable.

### DBeaver Version

22.2.5

### Operating System

PostgreSQL 12.7 (Ubuntu 12.7-0ubuntu0.20.10.1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 10.2.0-13ubuntu1) 10.2.0, 64-bit

### Database and driver

_No response_

### Steps to reproduce

1. Run a query containing `$P{ABCD}` parameters. 
2. Expect each parameter formated as describe to be individually in the "Bind parameter(s)" window.
3. In the last version of DBeaver this is not the case and only the `$P` is identified as a parameter.

image

### Additional context

- We use a tool that build queries with $P{ABCD} variables. This was extremely convenient to have DBeaver parse theses variables.
- I cannot say if the fact that it was working before is a feature, or a bug.
- I have also tried modifying the configuration without success.
image

E1izabeth:
Check unnamed (?) and named (:varname ${varname} $P{varname}) parameters and variables detection.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRuleManager.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRuleManager.java
@@ -119,11 +119,6 @@ public class SQLRuleManager {
         }
 
         if (!minimalRules) {
-            // Parameter rule
-            for (String npPrefix : syntaxManager.getNamedParameterPrefixes()) {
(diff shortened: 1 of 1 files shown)


LonwoLonwo (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #18864, opened 2023-01-24
#18522 Keep datasource's original provider and driver IDs

### LINKED ISSUE #18522: Keep original connection driver ID
Whenever datasource driver is replaced by some new driver (e.g. in EE version) we loose information about original driver.
When we return to original product version connection configuration is broken because new driver is not there.

We could keep original driver ID somewhere in connectin config and try to use it if driver not found.

ShadelessFox:
Acceptance criteria:
- [x] 1. In CE, existing data sources load fine
- [x] 2. In CE, after changing data sources and restarting the application, existing data sources load fine
- [x] 3. In EE, existing data sources load fine and were replaced by new drivers (if any)
- [x] 4. In EE, after changing data sources with replaced drivers and restarting the application, existing data sources load fine
- [x] 5. In CE, changed data sources load fine

### DIFF
--- a/plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceDescriptor.java
+++ b/plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceDescriptor.java
@@ -110,6 +110,8 @@ public class DataSourceDescriptor
     @NotNull
     private DBPDriver driver;
     @NotNull
+    private DBPDriver originalDriver;
+    @NotNull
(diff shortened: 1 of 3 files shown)


LonwoLonwo (approved):

E1izabeth (approved):

Matvey16:
DBeaver won't start

image

ShadelessFox:
I'm not sure it's related to this PR 🤔

Matvey16:
Still the same error, I don't know why but builds specifically from this PR (Both CE and EE) drop with the same error

Matvey16:
Verified

Issue #14113, opened 2021-10-03
PostgreSQL composite types - NULL arrays are displayed in a wrong way


JamesInform:
Error occurs at least in
DBeaver 21.2.0 / 21.2.1
PostgreSQL 13/14

If you run the following sql commands line by line

```
create type my_type as (a int8, b text, c float, d numeric, e int8[], f numrange, g timestamptz);
create table my_table (id int8, my my_type);
insert into my_table (id) values (1);
insert into my_table (id,my.a) values (2,100);
select id, (my).a, (my).b, (my).c,(my).d,(my).e,(my).f,(my).g from my_table;
select * from my_table;
```

you will see that the first select statement retrieves and displays the composite columns correctly as NULLs, while the second select incorrectly displays empty values for text, array and numrange.

Note that composite field `e` is displayed as "NULL" and not as "[NULL]" in both selects.

![image](https://user-images.githubusercontent.com/38714600/135745267-b279a08b-b977-498e-b411-0d735d41e7ee.png)

![image](https://user-images.githubusercontent.com/38714600/135745282-3732b989-21e6-4470-8fc3-10e3446864a9.png)

ShadelessFox:
Hello @JamesInform,

I believe that the first described problem was already fixed in #13349.

Can you try against an [early access](https://dbeaver.io/files/ea) version?

JamesInform:
I tested in early access. 
The Null handling of arrays is still not correct.

![image](https://user-images.githubusercontent.com/38714600/135750637-a21b6ae2-13d8-4165-9571-e833857d54a4.png)

ShadelessFox:
Yes, that's the second problem. Thank you for reply!

I will edit body of this issue to contain the only remaining problem.

JamesInform:
I checked with more field types and there are more types affected from wrong NULL handling:

```
create type mt as 
	(
		f_int int,f_int8 int8,f_float float,f_numeric numeric,f_date date,
		f_timestamp timestamp,f_timestamptz timestamptz,f_json json,f_jsonb jsonb,
		f_text text,f_varchar varchar(20),f_tsvector tsvector,f_bytea bytea,f_int8range int8range,
		f_daterange daterange,f_numrange numrange,f_tsrange tsrange,f_tstzrange tstzrange,
		f_datemultirange datemultirange, -- This is a new data type in PostgreSQL 14
		f_intarray int[], f_int8array int8[], f_textarray text[], f_daterangearray daterange[] 
	);
[...]

PR #18866, opened 2023-01-24
#16822 Treat `NULL` structs as so, don't check their values

### LINKED ISSUE #16822: Incorrect display of composite data type when all attributes are null
#### System information: 
- Operating system (distribution) and version Win11
- DBeaver version 22.1.0

1. Create a data type with two fields

```sql
create type aa as (a text, b text);
```

2. Execute a select with this data type so both fields are NULL
```sql
select array[row(null, null)::aa, null];
```

**Expected:** The result shows that there are two null fields
cmd_Wf7a8RPqnJ

**Actual:** nulls are merged into one null and are indistingushable from a single null field
dbeaver_63engc4laX
dbeaver_bzlkNRwzZy

ShadelessFox:
java_ILZAs6KyJT


E1izabeth:
Please, check that it doesn't break these cases #15698 #14113

E1izabeth (approved):

LonwoLonwo (approved):

uslss:
verified

PR #18880, opened 2023-01-26
"Typed" (enum) field filtering PostgreSQL fix dbeaver#14986

### LINKED ISSUE #14986: "Typed" (enum) field filtering
When I try to filter column of enum type - I cannot enter values by text (only by checking them), cannot filter the list when typing (it becomes empty).
And I cannot make partial filter - e.g. I want to filter: 
field_of_enum_type::text like 'something%'

But I cannot ((((
dbeaver_typed_column

plotn:
Tested under Postgres, works...


E1izabeth:
Please, link it with the issue

E1izabeth:
Also for the future it would be nice to see issue number at the beginning of commit message like `#num Add cast for enums filtering`.

E1izabeth:
Thank you

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTableColumn.java
```diff
@@ -200,6 +200,15 @@ public List<DBDLabelValuePair> getValueEnumeration(
         query.append("\nFROM ").append(DBUtils.getObjectFullName(getTable(), DBPEvaluationContext.DML));
         if (valuePattern instanceof String) {
             query.append("\nWHERE ");
+            boolean needCast = false;
+            String typeExt = "";
+            DBSDataType dataType = DBUtils.getDataType(this);
+            if (dataType != null) typeExt = CommonUtils.toString(dataType.geTypeExtension());
```
plotn: @E1izabeth , nota bene - possibly there is typo in existing code geTypeExtension ("ge", not "get") - but I left it unchanged, I dont know all the cases and dependencies

E1izabeth: Oh, I see. We will fix it later out of scope of this PR.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTableColumn.java
```diff
@@ -181,7 +181,7 @@ public List<DBDLabelValuePair> getValueEnumeration(
         boolean formatValues,
         boolean caseInsensitiveSearch) throws DBException
     {
-        final String identifier = DBUtils.getQuotedIdentifier(this, DBPAttributeReferencePurpose.DATA_SELECTION);
+        String identifier = DBUtils.getQuotedIdentifier(this, DBPAttributeReferencePurpose.DATA_SELECTION);
```
E1izabeth: I'd return `final` modifier back. I don't get why it was removed. If it was intentionally, could you elaborate please?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTableColumn.java
```diff
@@ -200,6 +200,15 @@ public List<DBDLabelValuePair> getValueEnumeration(
...
             query.append("\nWHERE ");
+            boolean needCast = false;
+            String typeExt = "";
+            DBSDataType dataType = DBUtils.getDataType(this);
+            if (dataType != null) typeExt = CommonUtils.toString(dataType.geTypeExtension());
+            needCast = "E".equals(typeExt); // for future - maybe not only Enum? But I have no such deep knowledge
+            if (needCast) {
+                identifier = "cast(" + identifier + " as varchar(1000))"; // it could be better for PostgreSQL to specify just "varchar" or "text"
+                                                                          // but this not conforms SQL standard, so I left varchar(1000) - it works  
+            }
```
E1izabeth: I suppose we don't need to check `"E".equals(typeExt)` if `dataType `is `null`, because it will always be `false`. The same is for cast - we will do nothing here if `dataType `is `null`, so let's add all this under that if statement. Variables declaration are not needed out of this `if`, so they go under `if` too.

Also I'd remove `needCast `variable as it's not so useful for debug purposes as `typeExt`, but makes code more verbose.
```suggestion
            DBSDataType dataType = DBUtils.getDataType(this);
            if (dataType != null) {
                String typeExt = CommonUtils.toString(dataType.geTypeExtension());
                if ("E".equals(typeExt)) { // for future - maybe not only Enum? But I have no such deep knowledge
                    // it could be better for PostgreSQL to specify just "varchar" or "text"
                    // but this not conforms SQL standard, so I left varchar(1000) - it works  
                    identifier = "cast(" + identifier + " as varchar(1000))"; 
            }
```

plotn: ok, agreed

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTableColumn.java
plotn: I modify indentifier later in code - thats why

E1izabeth: Yeah, I missed it. Thank you for explanations.

E1izabeth (approved):

E1izabeth:
**For QA: needs to be tested on PostgreSQL enums, MySQL enums and sets.**

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTableColumn.java
```diff
@@ -200,6 +200,15 @@ public List<DBDLabelValuePair> getValueEnumeration(
...
         if (valuePattern instanceof String) {
             query.append("\nWHERE ");
+            DBSDataType dataType = DBUtils.getDataType(this);
+            if (dataType != null) {
+                String typeExt = CommonUtils.toString(dataType.geTypeExtension());
+                if ("E".equals(typeExt)) { // for future - maybe not only Enum? But I have no such deep knowledge
+                    // it could be better for PostgreSQL to specify just "varchar" or "text"
+                    // but this not conforms SQL standard, so I left varchar(1000) - it works  
+                    identifier = "cast(" + identifier + " as varchar(1000))";
+                }
```
ShadelessFox: I don't like the idea of leaking database-specific hacks into the generic model. 

Maybe we can somehow delegate it to `SQLDialect`? Please take a look at `SQLDialect#getCastedAttributeName`.

E1izabeth: Do we have this bug only in PostgreSQL?

ShadelessFox: I was unable to reproduce it even against PostgreSQL, so I can't claim.

E1izabeth: Ok, I'll try a bit later with MySQL.

LonwoLonwo: Because this is PostgreSQL thing in the first hand - ability to have different data types in different schemas.

plotn: Not this reason! These are different bugs.
Reproduces clearly, i did it today

LonwoLonwo: Ok, give me an example from another database.

E1izabeth: Yes, seems the issue exists only for PotgreSQL, so it's better to do it dialect-specific.

plotn: done via dialect specific, please take a look

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataType.java
```diff
@@ -272,6 +272,13 @@ void resolveValueTypeFromBaseType(DBRProgressMonitor monitor) {
     @Nullable
     String getConditionTypeCasting(boolean isInCondition, boolean castColumnName) {
         final String typeName = getTypeName();
+        DBSDataType dataType = DBUtils.getDataType(this);
+        if ((dataType != null) && (isInCondition)) {
+            String typeExt = CommonUtils.toString(dataType.geTypeExtension());
+            if ("E".equals(typeExt)) { // for future - maybe not only Enum? But I have no such deep knowledge
+        	return "::text";
+            }    
+        }
```
ShadelessFox: - Unnecessary call to `getDataType` as it will yield the same reference as `this`
- Unnecessary string conversion, you can access its actual type since you're inside the PostgreSQL bundle
- Not useful comment

```suggestion
        if (isInCondition && typeCategory == PostgreTypeCategory.E) {
            return "::text";
        }
```

ShadelessFox (approved):

E1izabeth (approved):

uslss:
verified

PR #18887, opened 2023-01-26
#18885 Fix variables resolution in strings and comments

### LINKED ISSUE #18885: Get prompted every time a variable is referenced in a SQL script
### Description

I upgraded DBeaver Community Edition to version 22.3.3.202301211053 a few days ago. After this upgrade the behavior of the variables in DBeaver has changed., I use @set statements to setup values for the variable and was able to use them throughout my SQL scripts. The same scripts now kept prompting for the value of the variables. Looks like it does not use the values that I set in the script. Even after I entered the value in the "Bind parameters" dialog, it popped up each time the variable is used.

I have been searching to see if there is setting that I can control but to no avail... thinking of downgrading to 22.3.2,

Thanks

### DBeaver Version

Community Edition 22.3.3

### Operating System

Windows

### Database and driver

Snowflake

### Steps to reproduce

@set v_WorkFlow = Clone_Operations_Schema
@echo ${v_WorkFlow}, ${V_WORKFLOW}
select '${v_WorkFlow}';

### Additional context

_No response_

E1izabeth:
Variables ${varname} in strings and comments should be correctly resolved.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLScriptParser.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLScriptParser.java
@@ -781,7 +781,13 @@ public class SQLScriptParser {
                         }
 
                         if (param == null) {
-                            param = new SQLQueryParameter(syntaxManager, orderPos, matcher.group(0), start, matcher.end() - matcher.start());
+                            param = new SQLQueryParameter(
+                                syntaxManager,
+                                orderPos,
+                                matcher.group(SQLQueryParameter.VARIABLE_NAME_GROUP_NAME).toUpperCase(Locale.ENGLISH),
+                                start,
+                                matcher.end() - matcher.start()
+                            );
                             if (parameters == null) {
                                 parameters = new ArrayList<>();
                             }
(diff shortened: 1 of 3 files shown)


ShadelessFox (approved):

uslss:
verified

edmiller3:
Serge, I am not criticizing the fix at all, but I wanted to make you aware of something.

I downloaded and installed the preview version of 2.3.4 to test this fix myself.  While it does work, there is another (potentially unwanted) change related to this that happened in 2.3.3.

Previously, if I executed @set month_id=202301, when I looked down in the Variables window I would see it listed there as "month_id" with the value 202301.  Starting in 2.3.3, it is as if someone applied the UPPER function to all variable names as they were being parsed, because it began to show up in the Variables panel as "MONTH_ID".  ** That behavior still persists. **

It may not be significant for most people, as it still does appear to substitute this value (202301) anywhere it finds `${month_id} ` or `${MONTH_ID}`.  But if anyone happened to be using case-sensitive variables and setting month_id and MONTH_ID to different values, I suppose it would blow their script up.  I didn't really download 2.3.1 to verify how things used to behave, but I know that I did have lowercase variables before and now they all show as UPPERCASE in the Variables pane.

Just letting you know, in case this is important.
Screenshot_DBeaver

E1izabeth:
Thank you for feedback. It has been done intentionally. This variables has always been case-insensitive, so there is no difference between `@set xxx = 3` and `@set XxX = 3`. Only visual representation has been changed because of adding case-sensitivity for variables declared in quotes. Case-sensitive variables has not been supported until 22.3.2. Now they are supported only in parameters syntax. For example, you can do 
```
@set "yYy" = 1
@set "YYY" = 2
@set "yyy" = 3
select :"yYy" from dual; -- returns 1
select :yYy from dual; -- returns 2
select :yyy from dual; -- returns 2
select :"yyy" from dual; -- returns 3
```

edmiller3:
Elizabeth, I can't thank you enough for the reply.  It's helpful to know what the intentions are.  And you provided great examples for parameters.

PR #18902, opened 2023-01-30
dbeaver/pro#1350 Improve accessibility for entity properties

E1izabeth:
Works for NVDA:
- New shortcut for switching between left-side vertical tabs is Alt+Shift+Arrow_Up/Alt+Shift+Arrow_Down
- On table Properties opening selected left-side vertical tab name should be read
- On switching between left-side vertical tab selected tab name should be read
- Whole hierarchy of elements should be read on table Properties opening: Properties tab, left-side tab name, content element type

### DIFF
--- a/bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
+++ b/bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
@@ -174,6 +174,9 @@ public class ArrayUtils {
         }
     }
 
+    /**
+     * Returns index of the first found element satisfying a given predicate in the provided array 
+     */
     public static <T> int indexOf(T[] array, @NotNull Predicate<T> condition) {
         for (int i = 0; i < array.length; i++) {
             if (condition.test(array[i])) {
--- a/bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
+++ b/bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
@@ -25,6 +25,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
+import java.util.function.Predicate;
 
 /**
  * Common utils
@@ -171,6 +172,15 @@ public class ArrayUtils {
         } else {
             return Arrays.asList(array);
         }
+    }
+
+    public static <T> int indexOf(T[] array, @NotNull Predicate<T> condition) {
+        for (int i = 0; i < array.length; i++) {
+            if (condition.test(array[i])) {
+                return i;
+            }
+        }
+        return -1;
     }
 
     public static <T> int indexOf(T[] array, T element) {
--- a/bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
+++ b/bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
@@ -177,7 +177,7 @@ public class ArrayUtils {
     /**
      * Returns index of the first found element satisfying a given predicate in the provided array 
      */
-    public static <T> int indexOf(T[] array, @NotNull Predicate<T> condition) {
+    public static <T> int indexOf(@NotNull T[] array, @NotNull Predicate<T> condition) {
         for (int i = 0; i < array.length; i++) {
             if (condition.test(array[i])) {
                 return i;
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/properties/ObjectPropertiesTabFolderSwitchCommandHandler.java
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/properties/ObjectPropertiesTabFolderSwitchCommandHandler.java
@@ -48,9 +48,7 @@ public class ObjectPropertiesTabFolderSwitchCommandHandler extends AbstractHandl
                 String activeFolderId = propsEditor.getActiveFolderId();
                 int activeFolderIndex = ArrayUtils.indexOf(foldersInfo, f -> f.getId().equals(activeFolderId));
                 int targetFolderIndex = (activeFolderIndex + indexDelta + foldersInfo.length) % foldersInfo.length;
-                if (targetFolderIndex >= 0 && targetFolderIndex < foldersInfo.length) {
-                    propsEditor.switchFolder(foldersInfo[targetFolderIndex].getId());
-                }
+                propsEditor.switchFolder(foldersInfo[targetFolderIndex].getId());
             }
         }
         return null;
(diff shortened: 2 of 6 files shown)


bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
```diff
@@ -173,6 +174,15 @@ public static <T> List<T> safeArray(@Nullable T[] array)
         }
     }
 
+    public static <T> int indexOf(T[] array, @NotNull Predicate<T> condition) {
```
LonwoLonwo: Javadoc's comment can be helpful here, IMHO.

LonwoLonwo (approved):

bundles/org.jkiss.utils/src/org/jkiss/utils/ArrayUtils.java
```diff
@@ -173,6 +174,18 @@ public static <T> List<T> safeArray(@Nullable T[] array)
         }
     }
 
+    /**
+     * Returns index of the first found element satisfying a given predicate in the provided array 
+     */
+    public static <T> int indexOf(T[] array, @NotNull Predicate<T> condition) {
```
ShadelessFox:
```suggestion
    public static <T> int indexOf(@NotNull T[] array, @NotNull Predicate<? super T> condition) {
```

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/properties/ObjectPropertiesTabFolderSwitchCommandHandler.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class ObjectPropertiesTabFolderSwitchCommandHandler extends AbstractHandler {
+    
+    public static final String NEXT_PAGE_COMMAND_ID = "org.jkiss.dbeaver.entity.propsTab.nextPage";
+    public static final String PREV_PAGE_COMMAND_ID = "org.jkiss.dbeaver.entity.propsTab.prevPage";
+    
+    private static final Map<String, Integer> pageIndexDeltaByCommandId = Map.of(
+        NEXT_PAGE_COMMAND_ID, 1,
+        PREV_PAGE_COMMAND_ID, -1
+    );
```
ShadelessFox: It seems redundant. You can use a ternary operator:

```java
int delta = NEXT_PAGE_COMMAND_ID.equals(event.getCommand().getId()) ? 1 : -1;
```

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/properties/ObjectPropertiesTabFolderSwitchCommandHandler.java
```diff
@@ -0,0 +1,62 @@
...
+            Object currentPage = pagedEditor.getSelectedPage();
+            if (currentPage instanceof ObjectPropertiesEditor) {
+                ObjectPropertiesEditor propsEditor = (ObjectPropertiesEditor) currentPage;
+                TabbedFolderInfo[] foldersInfo = propsEditor.collectFolders(propsEditor);
+                String activeFolderId = propsEditor.getActiveFolderId();
+                int activeFolderIndex = ArrayUtils.indexOf(foldersInfo, f -> f.getId().equals(activeFolderId));
+                int targetFolderIndex = activeFolderIndex + indexDelta;
+                if (targetFolderIndex >= 0 && targetFolderIndex < foldersInfo.length) {
+                    propsEditor.switchFolder(foldersInfo[targetFolderIndex].getId());
+                }
```
ShadelessFox: WDYT of wrapping the index around? It would be useful IMO.

The "wrap-around" function can be expressed as so:
```java
public static int wrapAround(int index, int max) {
    return (index % max + max) % max;
}
```

It is proved to be working.

E1izabeth: I can do it, but it was not working like that before. Not sure, which is a good behaviour from the point of accessibility, but ok.

Why don't just `(index + max) % max` ?

E1izabeth: Never mind

E1izabeth: Done
I don't want to make it an utility method since here I know the the range of possible values, but in utility method I'll need to work with any value and match it to the function domain. I don't need it somewhere else, so it's unnecessary complexity.

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/properties/ObjectPropertiesTabFolderSwitchCommandHandler.java
```diff
@@ -50,7 +47,7 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
                 TabbedFolderInfo[] foldersInfo = propsEditor.collectFolders(propsEditor);
                 String activeFolderId = propsEditor.getActiveFolderId();
                 int activeFolderIndex = ArrayUtils.indexOf(foldersInfo, f -> f.getId().equals(activeFolderId));
-                int targetFolderIndex = activeFolderIndex + indexDelta;
+                int targetFolderIndex = (activeFolderIndex + indexDelta + foldersInfo.length) % foldersInfo.length;
                 if (targetFolderIndex >= 0 && targetFolderIndex < foldersInfo.length) {
```
ShadelessFox: I believe this check is redundant now

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/properties/ObjectPropertiesTabFolderSwitchCommandHandler.java
E1izabeth: Not sure, but ok

ShadelessFox (approved):

DariaMarkaryan:
1. the name of the first column is not pronounced (only the value if it's a list). if it is a tree, then after the value of the first column, pronounced the state "collapsed/expanded" and **without a pause, immediately the name of the column** (it turns out to be a hodgepodge, as for me).
2. does not read boolean values and values with a reference (which can be navigated with holding ctrl + left click).
3. when we expand the tree, let's say there is 1 element (I didn't find it with a large number, but it's still clear that it says the wrong thing) - it says "Expanded zero items" - but there is at least 1 child item ..
4. "One of one level zero" is what is said at the end of the voicing of any tree element if it's "collapsed" (I'm talking about the voicing "line - number of lines - lvl in the tree"). No matter if it first or last row on the current tab.
Only after listening to List I understood what it means (in the sheet it is normally voiced what line it is). At the same time, if the tree item is "expanded", then only "One of one" is pronounced on the parent and child, that is, the line number and number of lines are also incorrectly pronounced, and the lvl of the node in the tree disappears altogether.
More about this source text

E1izabeth:
We can't do something about 1 and 4.
2 and 3 looks like a bug, but we do not control this behaviour for now. I'll try to find out how can we fix it.

E1izabeth:
3 works this way for all such controls, even in Eclipse itself. It seems that it pronounce the number of added items to the current level, not the number of the child items.

E1izabeth:
2. ItemListControl uses SWT controls for trees and lists. Screen reader read text from the cells with getText(..) method. We set text to empty string for hyperlinks and booleans, so screen reader doesn't read actual values. dbeaver/pro#1395

PR #18915, opened 2023-01-31
dbeaver/pro#1308 Improve accessibility for data grid

E1izabeth:
Works for NVDA:
- On table opening the screen reader should say the table name and that we are in a grid view.
- On cell selection changing the screen reader should say the column name and the row number, then the value of the cell. If a cell contains specific data that cannot be read (image, blob, json, xml), the screen reader must specify the cell type.

Please, test for different databases (PostgreSQL, SQLite, Oracle, MySQL) and for different data types (blob, image, array, xml).

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/GridColumn.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/GridColumn.java
@@ -34,7 +34,7 @@ import java.util.List;
  *
  * @author serge@dbeaver.com
  */
-class GridColumn implements IGridColumn {
+public class GridColumn implements IGridColumn {
 
     /**
      * Default width of the column.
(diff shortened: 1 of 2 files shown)


LonwoLonwo (approved):

Destrolaric (approved):

Destrolaric:
Also, please test with `Select next cell after inline edit`.

E1izabeth:
I forgot to handle boolean values. Now it should work too.

Matvey16:
Verified

shadyar:
If you open Context Menu on a table > Data View, A grid will be shown.
I tested the new version. Now the grid has been really become accessible and cells read by the screen-reader but some things needed to improve.
On each cell The screen-reader reads:
"at row 7 column Titlestring value is Facelift"
As you can see, this state is too long complex and annoying. 
When the row changed by pressing Down / Up arrow key,, just the value plus row number should be read. eg. "Facelift At Row 6".
When the column changed by pressing Left / Right arrow key,, just the value plus column head  should be read. eg. "Facelift At column Title String".
About its order, I recommend the Type of data read at the end of the state also row or column read after the value because the  value is more important then other information.
I appreciate for your efforts to make accessible DB-Eaver for blind users.

PR #18917, opened 2023-01-31
#18319 fix filter when catalog parent is datasource

### LINKED ISSUE #18319: Hide not relevant schemas in projects tree
Suppose you have a long list of schemas in a database but you mainly work only with some of them and you don't want to scroll up and down looking for those. It would be cool to have the possibility to hide non relevant schemas, like for example "everything that has TEST% naming"

Destrolaric:
Closes #18319
Acceptance criteria:
Catalogs in generic databases which have datasource as a parent in the tree should now have filter actions, and it should work properly  
Test filtering for relations without folders `datasource -> catalog -> schema -> table`
Test Redis hierarchy keys
Important:
No reason to test with simple view, unfortunately it returns incorrect metadata so we should tune it in different ticket

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNDatabaseNode.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNDatabaseNode.java
@@ -746,39 +746,30 @@ public abstract class DBNDatabaseNode extends DBNNode implements DBNLazyNode, DB
 
     public DBSObjectFilter getNodeFilter(DBXTreeItem meta, boolean firstMatch) {
         DBPDataSourceContainer dataSource = getDataSourceContainer();
-        if (this instanceof DBNContainer) {
-            Class<?> childrenClass = this.getChildrenOrFolderClass(meta);
-            if (childrenClass != null) {
-                Object valueObject = getValueObject();
-                DBSObject parentObject = null;
-                if (valueObject instanceof DBSObject && !(valueObject instanceof DBPDataSource)) {
-                    parentObject = (DBSObject) valueObject;
-                }
-                return dataSource.getObjectFilter(childrenClass, parentObject, firstMatch);
+        Class<?> childrenClass = this.getChildrenOrFolderClass(meta);
+        if (childrenClass != null) {
+            Object valueObject = getValueObject();
+            DBSObject parentObject = null;
+            if (valueObject instanceof DBSObject && !(valueObject instanceof DBPDataSource)) {
+                parentObject = (DBSObject) valueObject;
             }
+            return dataSource.getObjectFilter(childrenClass, parentObject, firstMatch);
         }
         return null;
     }
 
     public void setNodeFilter(DBXTreeItem meta, DBSObjectFilter filter) {
         DBPDataSourceContainer dataSource = getDataSourceContainer();
-        if (this instanceof DBNContainer) {
-            Class<?> childrenClass = this.getChildrenOrFolderClass(meta);
-            if (childrenClass != null) {
-                Object parentObject = getValueObject();
-                if (parentObject instanceof DBPDataSource) {
-                    parentObject = null;
-                }
-                dataSource.setObjectFilter(
-                    childrenClass,
-                    (DBSObject) parentObject,
-                    filter);
-                dataSource.persistConfiguration();
-            } else {
-                log.error("Cannot detect child node type - can't save filter configuration");
+        Class<?> childrenClass = this.getChildrenOrFolderClass(meta);
+        if (childrenClass != null) {
+            Object parentObject = getValueObject();
+            if (parentObject instanceof DBPDataSource) {
+                parentObject = null;
             }
+            dataSource.setObjectFilter(childrenClass, (DBSObject) parentObject, filter);
+            dataSource.persistConfiguration();
         } else {
-            log.error("No active datasource - can't save filter configuration");
+            log.error("Cannot detect child node type - can't save filter configuration");
         }
     }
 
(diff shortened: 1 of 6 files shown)


plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorHandlerFilterConfig.java
```diff
@@ -49,26 +47,27 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
...
         }
+
         return null;
     }
 
     public static void configureFilters(Shell shell, DBNNode node)
     {
-        final DBNDatabaseFolder folder = (DBNDatabaseFolder) node;
-        DBXTreeItem itemsMeta = folder.getItemsMeta();
+        final DBNDatabaseNode dbnDatabaseNode = (DBNDatabaseNode) node;
```
ShadelessFox: Please remove this obscure prefix: `dbn`

Destrolaric: Done

ShadelessFox (approved):

E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNDatabaseNode.java
```diff
@@ -746,39 +746,30 @@ public DBPDataSource getDataSource() {
...
+                parentObject = (DBSObject) valueObject;
             }
+            return dataSource.getObjectFilter(childrenClass, parentObject, firstMatch);
         }
         return null;
     }
 
     public void setNodeFilter(DBXTreeItem meta, DBSObjectFilter filter) {
         DBPDataSourceContainer dataSource = getDataSourceContainer();
-        if (this instanceof DBNContainer) {
```
E1izabeth, line 765: Is there another class which we can use instead? It would be better then just delete this if. I believe it was here for some reason.

Destrolaric: Yes, because previously, our model assumed we only needed filtering for folders and sometimes datasources, because most of the other stuff was in folders. Now we have a lot of DBs which have direct `catalog -> schema -> table ` structure or non, like, Redis hierarchy.

Destrolaric: Now it allows you to set filter config for child element, because it can also have it's children. We don't need any other class for that because all functionality exists inside DBNDatabaseNode we just didn't use it previously for child objects.

E1izabeth (approved):

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorHandlerFilterClear.java
```diff
@@ -36,11 +35,8 @@ public class NavigatorHandlerFilterClear extends AbstractHandler {
     public Object execute(ExecutionEvent event) throws ExecutionException {
         final ISelection selection = HandlerUtil.getCurrentSelection(event);
         DBNNode node = NavigatorUtils.getSelectedNode(selection);
-        if (node instanceof DBNDatabaseItem) {
-            node = node.getParentNode();
-        }
-        if (node instanceof DBNDatabaseFolder) {
-            final DBNDatabaseFolder folder = (DBNDatabaseFolder) node;
+        if (node instanceof DBNDatabaseNode) {
+            final DBNDatabaseNode folder = (DBNDatabaseNode) node;
```
E1izabeth, line 39: Please check in advance if there is some other places where it should be fixed.

E1izabeth (approved):

LonwoLonwo:
Did you test this case? You remove the getParent code, but this is one of the cases for why we need a parent.

2023-02-02 15_19_28-DBeaver 22 3 4 - _orcl_ Script-39

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/actions/ObjectPropertyTester.java
```diff
@@ -201,26 +202,26 @@ public boolean test(Object receiver, String property, Object[] args, Object expe
                 break;
             }
             case PROP_CAN_FILTER: {
-                if (node instanceof DBNDatabaseItem) {
-                    node = node.getParentNode();
-                }
-                if (node instanceof DBNDatabaseFolder && ((DBNDatabaseFolder) node).getItemsMeta() != null) {
+                if ((node instanceof DBNDatabaseNode && ((DBNDatabaseNode) node).getItemsMeta() != null)) {
```
LonwoLonwo, line 205: extra brackets

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/actions/ObjectPropertyTester.java
```diff
@@ -201,26 +202,26 @@ public boolean test(Object receiver, String property, Object[] args, Object expe
...
                     return true;
                 }
                 break;
             }
             case PROP_CAN_FILTER_OBJECT: {
-                if (node.getParentNode() instanceof DBNDatabaseFolder && ((DBNDatabaseFolder) node.getParentNode()).getItemsMeta() != null) {
+                if ((node.getParentNode() instanceof DBNDatabaseNode && ((DBNDatabaseNode) node.getParentNode()).getItemsMeta() != null)) {
```
LonwoLonwo, line 211: extra brackets

serge-rider:
Too many changes in base model. It may break entire filters system.
Must be tested in all modes including simple/advanced, global filters in connection settings.
Filters in folders and filters of individual objects.

@Destrolaric please resolve issues above and do extra testing.

Destrolaric:
Agree. Too dangerous to change before the major version. Also, there is the case of databases with trees containing optional levels. In them, the actual navigation tree may differ from the meta tree. We probably need to do something with that before making mentioned changes.

PR #18919, opened 2023-02-01
#15282 Add setting for line ending auto fix on sql query execution

### LINKED ISSUE #15282: SQL execution with Inline double-dash comment
DBeaver 21.3.3, Windows 10

Working with SQLServer database, the execution of a query with a double-dash comment results in an execution error, while with MySQL, using the same query, the execution ends correclty
By opening the executed query tab, it looks like that for the SQLServer version the new lines are not kept

**MySQL**
image

**SQLServer**
image

E1izabeth:
We replaced \r to whitespace. SQL Server and probably some other databases don't like it.
Previously we replaced \r to \n which was weird and doesn't work correct for some users.
Let's have a setting for that
image

Query from the issue works with CR LF for SQL Server.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/ModelPreferences.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/ModelPreferences.java
@@ -22,6 +22,7 @@ import org.jkiss.dbeaver.model.DBConstants;
 import org.jkiss.dbeaver.model.connection.DBPConnectionConfiguration;
 import org.jkiss.dbeaver.model.exec.DBCExecutionPurpose;
 import org.jkiss.dbeaver.model.impl.preferences.BundlePreferenceStore;
+import org.jkiss.dbeaver.model.messages.ModelMessages;
 import org.jkiss.dbeaver.model.preferences.DBPPreferenceStore;
 import org.jkiss.dbeaver.model.qm.QMConstants;
 import org.jkiss.dbeaver.model.qm.QMObjectType;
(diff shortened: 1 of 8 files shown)


LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/preferences/PrefPageSQLExecute.java
```diff
@@ -47,12 +48,23 @@ public class PrefPageSQLExecute extends TargetPrefPage
     private static final Log log = Log.getLog(PrefPageSQLExecute.class);
 
     public static final String PAGE_ID = "org.jkiss.dbeaver.preferences.main.sqlexecute"; //$NON-NLS-1$
+    
+    private static final java.util.List<LineEndingNormalizationBehavior> LINE_ENDING_NORMALIZATION_VALUES = java.util.List.of(
```
ShadelessFox, line 52: Why is this a list? You can retrieve the values of an enum.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -733,9 +736,20 @@ public static String generateParamList(int paramCount) {
...
      */
-    public static String fixLineFeeds(String sql) {
+    public static String fixLineFeeds(String sql, DBPPreferenceStore prefStore) {
+        final String prefSetting = prefStore.getString(ModelPreferences.STATEMENT_LINE_ENDING_NORMALIZATION);
+        LineEndingNormalizationBehavior behavior = LineEndingNormalizationBehavior.parse(prefSetting);
+        switch(behavior) {
+            case AS_IS: return sql;
+            case DEFAULT: return fixLineFeedsDefault(sql);
+            default: return sql.replaceAll("\\R", behavior.getChars()); // replace Any Unicode linebreak sequence with given characters
+        }
```
ShadelessFox, line 749:
```suggestion
        switch(behavior) {
            case AS_IS:
                return sql;
            case DEFAULT:
                return fixLineFeedsDefault(sql);
            default:
                // replace Any Unicode linebreak sequence with given characters
                return sql.replaceAll("\\R", behavior.getChars());
        }
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/ModelPreferences.java
```diff
@@ -70,6 +71,39 @@ public static SeparateConnectionBehavior parse(String value) {
...
+        }
+        
+        public String getChars() {
+            return chars;
+        }
+
+        /**
+         * Convert value to LineEndingNormalizationBehavior option
+         */
+        public static LineEndingNormalizationBehavior parse(String value) {
```
ShadelessFox, line 102: I don't like that we have to pass the value itself. Can this method accept a `DBPPreferenceStore`? This way, we can rename it to `valueOf`.

ShadelessFox (commented):
Please add nullability annotations for method parameters and return types

PR #18923, opened 2023-02-01
#15282 Fix single \r to \r\n before query execution

### LINKED ISSUE #15282: SQL execution with Inline double-dash comment

E1izabeth:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -740,23 +740,25 @@ public static String fixLineFeeds(String sql) {
             return sql;
         }
         boolean hasFixes = false;
-        char[] fixed = sql.toCharArray();
-        for (int i = 0; i < fixed.length; i++) {
-            if (fixed[i] == '\r') {
-                if (i > 0 && fixed[i - 1] == '\n') {
+        char[] initial = sql.toCharArray();
+        StringBuilder fixed = new StringBuilder();
```
ShadelessFox:
```suggestion
        StringBuilder fixed = new StringBuilder(initial.length);
```

ShadelessFox (commented):
What do you think about adding a unit test for this method?

E1izabeth:
It's a good idea.

ShadelessFox (approved):

LonwoLonwo (approved):

uslss:
verified

PR #18924, opened 2023-02-01
dbeaver/pro#926 Add wiki link about exporting to external drive

ShadelessFox:
Please also check the wiki page: https://github.com/dbeaver/dbeaver/wiki/Data-transfer-external-storage

image

Note: such a placement of the `Global Settings` link is expected


E1izabeth (approved):

E1izabeth:
Maybe we should specify the concrete external storage (Google Sheets) in the link label as we tell only about it in our wiki.
Are you thinking about other instructions like exporting to yandex drive and baidu cloud?

ShadelessFox:
Yes, we want to extend our wiki to include other cloud storage providers in the future. For now, it's just Google Drive.

Matvey16:
Verified

PR #18934, opened 2023-02-02
dbeaver/pro#1387 ai enhancements

### LINKED ISSUE #18996: gpt.meta.transfer.confirmed by connection
Right now, it seems that the prompt about sending schema metadata to openai is a global confirmation. Would be good to make it by connection as I have connections to databases that I wouldn't want to share. I frequently change connections and it would be very easy to not notice which connection my script is connected to.

serge-rider:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/translator/DAIHistoryItem.java
```diff
@@ -0,0 +1,91 @@
...
+    public DAIHistoryItem(String naturalText, String completionText) {
+        this.naturalText = naturalText;
+        this.completionText = completionText;
+    }
+
+    public DAIHistoryItem(Map<String, Object> map) {
+        this.id = JSONUtils.getString(map, "id");
+        this.naturalText = JSONUtils.getString(map, "naturalText");
+        this.completionText = JSONUtils.getString(map, "completionText");
+        this.time = new Date(JSONUtils.getLong(map, "time", 0));
```
E1izabeth, line 48: These strings are used at least twice. I'd make them to be constants.

serge-rider: We want replace them with native json serializatiom. It is temporary

E1izabeth (approved):

Issue #16683, opened 2022-06-02
[MySQL] Can't use calendar in value panel for date\time\timestamp\datetime values 


DariaMarkaryan:
#### System information: 
- Win11
- DBeaver CE 22.1.0

#### Connection specification:
- MySQL

#### Describe the problem you're observing:
https://user-images.githubusercontent.com/43252954/171619109-e9027c7b-86a8-4c25-bb9d-73963b050d4a.mp4

#### Steps to reproduce, if exist:
Set focus to the datetime format and open Value panel. 
Click on this icon for the calendar -> you'll get the error
![image](https://user-images.githubusercontent.com/43252954/171619733-5e54d0c7-fee5-43ac-a16e-939c50887c8e.png)

E1izabeth:
Is the issue still actual for the latest DBeaver version? I can't reproduce it.

PR #18937, opened 2023-02-02
#16683 fix localdatetime cast

Destrolaric:


ShadelessFox:
Weird commit name that has nothing to do with the actual change

PR #18945, opened 2023-02-03
18944 Change defaults for Exasol connections parameters.

### LINKED ISSUE #18944: Parameter defaults for Exasol
Some JDBC driver parameter defaults for Exasol are rewritten by the plugin:

```xml
<property name="querytimeout" value="600"/>
<property name="connecttimeout" value="1000"/>
```

Query timeout of 10 minutes fits well for Exasol-only queries, but might be too small if external data sources are involved.

"connecttimeout" is deprecated since version 7.1.11: [Drivers: Unified login timeout parameters](https://exasol.my.site.com/s/article/Changelog-content-14618?language=en_US)

The suggestion is to change querytimeout default to 1 hour (3600) and remove connecttimeout from default rewrites.

allipatev:
Was discussed with @Sargul .

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.exasol/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.exasol/plugin.xml
@@ -427,7 +427,7 @@
                         useURL="false"
                         defaultPort="8563"
                         description="%driver.exasol.description"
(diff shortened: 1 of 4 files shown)


E1izabeth (approved):

E1izabeth:
Thank you for contribution!

LonwoLonwo (approved):

uslss:
verified

PR #18961, opened 2023-02-06
#17614 Fix underscores from variables in export file name

### LINKED ISSUE #17614: Multiple underscores in table names get truncated to one when exporting a table to a file
### Description

Multiple underscores in table names get truncated to one when exporting a table to a file

### DBeaver Version

Community edition 22.2.0

### Operating System

Win 10

### Database and driver

Postgres but have also reproduced under Clickhouse so I don't think it's DB related

### Steps to reproduce

1. Create a table that has multiple underscores in it's name:
```
CREATE TABLE test____2 (
	some_column varchar NULL
);
```
2. Try to export it to a CSV file
3. For simplicity, in the "File name pattern", enter "${table}"
4. The file will be created with only one underscore test_2.csv instead of test____2.csv

### Additional context

_No response_

E1izabeth:


ShadelessFox (approved):

LonwoLonwo (approved):

LonwoLonwo:
Maybe exists a case when we need to remove the extra "_" from the file name?

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamTransferConsumer.java
```diff
@@ -876,7 +876,7 @@ private static String stripObjectName(String name) {
             if (Character.isLetterOrDigit(c)) {
                 result.append(c);
                 lastUnd = false;
-            } else if (!lastUnd) {
+            } else if (c == '_' || !lastUnd) {
                 result.append('_');
                 lastUnd = true;
             }
```
ShadelessFox, line 882: Actually, it looks silly.

Now, if we look closely, it basically replaces every non-letter or digit character with an underscore. 

I believe we can remove that condition entirely. Please correct me if I'm wrong.

E1izabeth: I was trying to understand what is the purpose of this code for a long. It replaces non-letter or digit characters to underscore, but only once - if there is a sequence of them, then only one _ will be added for all of them.

E1izabeth:
@LonwoLonwo I don't know such cases. This code is aimed to transform names like `apple%&@1` to `apple_1`.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamTransferConsumer.java
ShadelessFox: So maybe this was supposed to be a feature then?

E1izabeth: It seems strange for me. Why do someone need to replace multiple underscores to one if he used pattern and wants to see the name of the table or datasource.

ShadelessFox: I believe this function sanitizes the filename so the OS won't complain.

Not many characters are forbidden in file paths: `\`, `/`, `:`, `*`, `?`, `"`, `<`, `>`, `|`, and Unix-specific `\0`.

Maybe we can replace only them with an underscore and keep the rest?

E1izabeth: We can't use non-letter non-digit characters in file names - so we replace them to underscore. But if the file name contains underscore why do we need to do something here as underscore is allowed character for filename

E1izabeth: So you propose to rewrite whole this method, aren't you?

ShadelessFox: Nothing, just keep it?

ShadelessFox: ¯\\\_(ツ)\_/¯

ShadelessFox (approved):

Matvey16:
Verified

PR #18986, opened 2023-02-07
#18890 Handle multi-results with errors in SQL Server as warnings and…

### LINKED ISSUE #18890: Missing results/messages/errors for SQL Server
### Description

Using DBeaver on SQL Server, I have an error about missing messages and results.
Note that:
1. The error happens both with Execute SQL Statement and Execute SQL Script
2. The error happens both with or without separators.

When I run multiple queries, I get only the first result.
image

When I run multiple queries and I got any error, I get only the first result and the error message is not shown.
image

The error is not about missing execution of following statements.
I tested a temp table creation both before and after raiserror, and both tables are created.
The scripts are:
```
select 1 
select 1 as A into #t
raiserror('Test', 16, 1)

select 1 
raiserror('Test', 16, 1)
select 2 as B into #v
```
Table #t has been created as expected.
Table #v has been created as expected.

Further error:
I get a null pointer exception trying to test multiple select in a more complex script:
image

Foot Notes:
1. The issue doesn't happen when DBeaver is just started. But after some event (I have not understood which one) the issue happens everytime until I reboot DBeaver;
2. The error has already been posted ( #18816 ) but his solution didn't work for me.
3. SQL Server version is 2019

Thanks

### DBeaver Version

Community edition 22.3.3

### Operating System

Windows 10

### Database and driver

Database: MS SQL 2019
Driver: Microsoft JDBC Driver for SQL Server (MSSQL)

### Steps to reproduce

1. Connect to SQL Server
2. Run a Statement with 2 select queries and / or a raiserror. Examples:
```
select 1 as A;
select 2 as A;

select 1 as A;
raiserror('test error', 16, 1);
```
3. Check that only first select result is shown.

### Additional context

Note that the error does't happens when the application is just started.
After first time, it happens everytime.

When I understand new information I will update this issue report.

serge-rider:
… write them in output log

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDataSourceInfo.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDataSourceInfo.java
@@ -48,7 +48,7 @@ class SQLServerDataSourceInfo extends JDBCDataSourceInfo {
 
     @Override
     public boolean isMultipleResultsFetchBroken() {
-        return isSybase;
+        return true;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


Matvey16:
This creates a bug that stops the entire script without asking to skip on error if it encounters it.


https://user-images.githubusercontent.com/82543000/217840951-aa5065e3-4eb7-4aeb-9a10-855efa5f6bb1.mp4

Matvey16:
Verified

PR #19005, opened 2023-02-08
#17674 use database datatype when case not equivalent

### LINKED ISSUE #17674: Invalid type int8 instead of Int8 when export from PostgreSQL and import to ClickHouse
### Description

Can not transfer data from PostgreSQL to ClickHouse, because generated DDL for create new table has invalid type 'int8' instead of 'Int8'

### DBeaver Version

CE Version 22.2.0.202209051344

### Operating System

macOS Monterey 12.6

### Database and driver

Connection Drivers:
PostgreSQL standard driver
Yandex ClickHouse legacy driver (server version < 21.3)

### Steps to reproduce

1. In Postgres create table t1 with a column typed int8
2. Insert data to table t1 (insert into t1 values (1)) 
3. Export data by context menu from PostgreSQL to ClickHouse

### Additional context

_No response_

Destrolaric:
java_3hpkuVV9aS

Closes #17674

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
@@ -323,6 +323,10 @@ public final class DBStructUtils {
             if (dataType == null && typeName.contains("(")) {
                 // It seems this data type has modifiers. Try to find without modifiers
                 dataType = dataTypeProvider.getLocalDataType(SQLUtils.stripColumnTypeModifiers(typeName));
+                if (dataType != null) {
+                    int startPos = typeName.indexOf("(");
+                    typeName = dataType + typeName.substring(startPos);
+                }
             }
             if (dataType == null && typeNameLower.equals(DOUBLE_DATA_TYPE)) {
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
```diff
@@ -425,6 +425,8 @@ public static String mapTargetDataType(DBSObject objectContainer, DBSTypedObject
             }
             if (dataType != null) {
                 dataKind = dataType.getDataKind();
+                // Datatype caches ignore case, but we probably should use it with the original case
```
LonwoLonwo, line 432: But usually original <> target, but source. And in this case target type name is used.

LonwoLonwo: (About the comment)

serge-rider:
Strange fix. A couple of lines before the change we set typeName to targetType name. In this fix we use source type name. 

It will break all data transfers in all databases because idea is to use target database type.

uslss:
verified

PR #19006, opened 2023-02-08
#18906 use job instead of running in UI

### LINKED ISSUE #18906: Add progress bar to 'Show full DDL' receiving process
If you open Generate DDL for example for pretty big schema and click on 'Show full DDL', the DBeaver will thinking a lot of time, or even you'll get system message that application isn't responding. 

image

*solution?*
We should add a progress bar with a Cancel button to be able to stop the operation

Destrolaric:
Now it uses Jobs that could be canceled by just closing the dialog or changing the tick value so the job will restart for different parameters. Also UI should not freeze.

https://user-images.githubusercontent.com/20579256/217806670-c7fc9f39-dd48-4a32-a82a-9d92b78aaa22.mp4

Closes #18906

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
@@ -268,14 +268,14 @@ class SQLGeneratorDialog extends ViewSQLDialog {
     }
 
     private void startGenerateJob() {
-        generateDDLJob.cancel();
-        generateDDLJob.schedule();
         setSQLText("Loading DDL...");
         updateSQL();
         Button button = getButton(IDialogConstants.DETAILS_ID);
         if (button != null) {
             button.setEnabled(false);
         }
+        generateDDLJob.cancel();
+        generateDDLJob.schedule();
     }
 
     @Override
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
@@ -16,6 +16,13 @@
  */
 package org.jkiss.dbeaver.ui.editors.sql.generator;
 
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.jobs.IJobChangeEvent;
+import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.core.runtime.jobs.JobChangeAdapter;
+import org.eclipse.jface.dialogs.IDialogConstants;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.SelectionAdapter;
 import org.eclipse.swt.events.SelectionEvent;
@@ -25,16 +32,21 @@ import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Group;
 import org.eclipse.ui.IWorkbenchPartSite;
+import org.jkiss.dbeaver.Log;
 import org.jkiss.dbeaver.model.DBPScriptObject;
 import org.jkiss.dbeaver.model.DBPScriptObjectExt2;
 import org.jkiss.dbeaver.model.exec.DBCExecutionContext;
+import org.jkiss.dbeaver.model.runtime.DefaultProgressMonitor;
 import org.jkiss.dbeaver.model.sql.generator.SQLGenerator;
 import org.jkiss.dbeaver.ui.UIUtils;
 import org.jkiss.dbeaver.ui.editors.sql.dialogs.ViewSQLDialog;
 import org.jkiss.dbeaver.ui.editors.sql.internal.SQLEditorMessages;
 import org.jkiss.utils.CommonUtils;
 
+import java.lang.reflect.InvocationTargetException;
+
 class SQLGeneratorDialog extends ViewSQLDialog {
+    private static final Log log = Log.getLog(SQLGeneratorDialog.class);
 
     private static final String PROP_USE_FQ_NAMES = "GenerateSQL.useFQNames";
     private static final String PROP_USE_COMPACT_SQL = "GenerateSQL.compactSQL";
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
```diff
@@ -278,16 +281,21 @@ public void widgetSelected(SelectionEvent e) {
...
-                        updateSQL();
-                    }
+                    startGenerateJob();
                 }
             });
         }
         return composite;
     }
+
+    private void startGenerateJob() {
```
E1izabeth, line 270: I'd prefer to call it like this
```suggestion
    private void startGenerateDDLJob() {
```

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
```diff
@@ -99,11 +112,55 @@ protected Composite createDialogArea(Composite parent) {
...
-        }
+        generateDDLJob = new Job("Generating DDL") {
+            @Override
+            protected IStatus run(IProgressMonitor monitor) {
+                try {
+                    sqlGenerator.run(new DefaultProgressMonitor(monitor));
+                    return Status.OK_STATUS;
+                } catch (InvocationTargetException e) {
+                    log.error(e);
+                    return Status.error("Error running while DDL generation", e);
```
LonwoLonwo: Error running during DDL generation?

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
```diff
@@ -99,11 +112,55 @@ protected Composite createDialogArea(Composite parent) {
...
+        };
+        generateDDLJob.addJobChangeListener(new JobChangeAdapter() {
+            @Override
+            public void running(IJobChangeEvent event) {
+                if (SQLGeneratorDialog.this.getShell() != null && !SQLGeneratorDialog.this.getShell().isDisposed()) {
+                    UIUtils.syncExec(() -> {
+                        setSQLText("Loading DDL...");
+                        updateSQL();
+                        Button button = getButton(IDialogConstants.DETAILS_ID);
+                        if (button != null && !button.isDisposed()) {
```
LonwoLonwo: What for? Where is this button?

Destrolaric: Copy button

Destrolaric: No reason to copy if DDL is not yet ready

LonwoLonwo:
Please add images for UI tickets PR-s

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
LonwoLonwo: ok.

serge-rider:
@Destrolaric 
- Do not use plain Job, use our AbstractJob and our monitors.
- Use synExec right in jub.run, no need to use job listener.

Destrolaric:
@serge-rider fixed

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
```diff
@@ -99,11 +110,48 @@ protected Composite createDialogArea(Composite parent) {
...
-        UIUtils.runInUI(sqlGenerator);
-        Object sql = sqlGenerator.getResult();
-        if (sql != null) {
-            setSQLText(CommonUtils.toString(sql));
-        }
+        generateDDLJob = new AbstractJob("Generating DDL") {
+            @Override
+            protected IStatus run(DBRProgressMonitor monitor) {
+                try {
+                    if (SQLGeneratorDialog.this.getShell() != null && !SQLGeneratorDialog.this.getShell().isDisposed()) {
```
serge-rider: Move it outside of job (before job schedulke). No need to check shell at all.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
```diff
@@ -99,11 +110,48 @@ protected Composite createDialogArea(Composite parent) {
...
+                            updateSQL();
+                            Button button = getButton(IDialogConstants.DETAILS_ID);
+                            if (button != null && !button.isDisposed()) {
+                                button.setEnabled(false);
+                            }
+                        });
+                    }
+                    sqlGenerator.run(monitor);
+                    Object sql = sqlGenerator.getResult();
+                    if (
```
serge-rider: Move all checks inside syncExec (never operate with shell outside of UI thread. Even getShell function may fail).
No need to check isDisposed for button. Use button variable from block above instead of getButton.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
```diff
@@ -278,16 +260,27 @@ public void widgetSelected(SelectionEvent e) {
...
+                    startGenerateJob();
                 }
             });
         }
         return composite;
     }
+
+    private void startGenerateJob() {
+        generateDDLJob.cancel();
+        generateDDLJob.schedule();
```
serge-rider: Why doi you schedule job before disabling button and updating SQL?
PrepareUI first, then start generating then enableUI.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
Destrolaric: Fixed

uslss:
verified

PR #19011, opened 2023-02-08
#17244 set maxLength by default to -1 in case of the number format error

### LINKED ISSUE #17244: Datatable export doesn't create varchar(MAX) but varchar(1) instead
### Description

in Export Data -> Database
When selecting [configure] you can see that it correctly determined column should be _varchar(MAX)_ due to the large string. But it creates _varchar(1)_ instead and causes errors due to truncation of the string during export.

Before I updated DBeaver to see if perhaps it's fixed already there was an option to preview DLL (option now gone), and you could see that the table creation script said 'columnName varchar NULL', instead of 'columnName varchar(MAX) NULL' only for MAX. varchar columns of size that were a number are correct, like 'columnName varchar(255) NULL'. 

image

Note that the longest string is not actually that long (16777216), it shows that number for any column that contains a long string. 

### Steps to reproduce

As a source, use a datatable that has a column containing a very long string, longer than 255. Try to export it to a different schema where it will have to be created by DBeaver. 

### DBeaver Version

22.1.3

### Operating System

Windows 10

LonwoLonwo:
https://github.com/dbeaver/dbeaver/blob/71f0d3b8f33774c3b1304a18a8203b804ff09c6b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/struct/AbstractAttribute.java#L264

As in another method, I think that "-1" for the maxLenght parameter is better for recognition than "0" in side cases like varchar(max).

2023-02-08 18_17_39-Configure metadata structure

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/struct/AbstractAttribute.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/struct/AbstractAttribute.java
@@ -142,7 +142,7 @@ public abstract class AbstractAttribute implements DBSAttributeBase, DBSTypedObj
                 String modifiers = fullTypeName.substring(divPos + 1, divPos2);
                 int divPos3 = modifiers.indexOf(',');
                 if (divPos3 == -1) {
-                    maxLength = precision = CommonUtils.toInt(modifiers);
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

uslss:
verified

PR #19012, opened 2023-02-08
#18998 Fix variable names extraction for tasks

### LINKED ISSUE #18998: Variables in Database Tasks Are No Longer Being Passed to Scripts
### Description

This worked consistently until quite recently. When using the ${variable} notation within .sql scripts I was able to use the 'Variables' button in the Database Task and pass the value. After updating (however I'm not aware of which version this was introduced) the scripts only see NULL in place of the variables.

### DBeaver Version

Community Edition Version 22.3.4.202302060909

### Operating System

_No response_

### Database and driver

Tested with Postgres: 
image

Tested with MSSQL:
image

Result is the Same

### Steps to reproduce

Create a Database Task with a Script using a variable

Set Variable in the Task Options 
Database Tasks > Variables >
image

Error Occurs because NULL is being passed rather than Variable

Query Manager Shows
select * from NULL limit 10

### Additional context

_No response_

E1izabeth:
In tasks, variables are dialect-agnostic. So we can use BasicSQLDialect there.


LonwoLonwo (approved):

Matvey16:
Verified

PR #19016, opened 2023-02-09
#19015 Update data viewer presentation in sync mode

### LINKED ISSUE #19015: Errors in data viewer during SQL re-execute
### Description

When you a query in SQL editor and then execute anther one with less column numbers we can see a lot of errors in log.
Like
```
2023-02-09 08:00:39.917 - Bad attribute 'release_year' index: 3 is out of row values' bounds (3)
2023-02-09 08:00:39.917 - Bad attribute 'language_id' index: 4 is out of row values' bounds (3)
```

It happens because data viewer refresh its state in async mode after data model is refreshed.
WE need to fix this.

### DBeaver Version

22

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

serge-rider:


Matvey16:
Verified

PR #18916, opened 2023-01-31
Updated InterSystems IRIS JDBC driver location

isc-rsingh:
Updated InterSystems IRIS JDBC driver location due to a reorganization of files in the repo, plus an update to the driver.

PR #19020, opened 2023-02-09
InterSystems IRIS driver: fix broken link, update to 3.6.1

eduard93:
Old link is not working anymore, replaced jar link. Driver version 3.6.1 is compatible with old servers, so users who are using version 3.1.0 would not be affected.


E1izabeth:
Thank you for contribution. We appreciate it.
Unfortunatelly, we have already got PR with the same changes https://github.com/dbeaver/dbeaver/pull/18916
I have to close this PR as a duplicate. Sorry.

PR #19083, opened 2023-02-16
dbeaver/pro#1397 Add localization to spreadsheet accessibility

E1izabeth:
Localization for accessibility was added. Also I changed the order of reading according to the feedback from our users.
Also I added translation into Russian, so you can switch DBeaver into Russian and check that NVDA reads correctly on Russian while navigating through Results Grid.


Destrolaric (approved):

LonwoLonwo (approved):

uslss:
verified

Issue #19089, opened 2023-02-16
"Cancel running queries" popup alert message is malformed


brankoradovanovic-mcom:
### Description

When attempting to close an SQL script tab with a running query, a malformed alert message is displayed ('There are `"<missing argument>"` running SQL queries in this editor.')

### DBeaver Version

CE 22.3.4

### Operating System

Windows 10 x64

### Database and driver

Postgres 13.4, JDBC driver 42.5.0

### Steps to reproduce

1. Open a new SQL tab
2. In the new tab, start a long-running query
3. Try to close the tab by clicking on the "x"
4. The following popup window is displayed:

![image](https://user-images.githubusercontent.com/101108585/219430925-0fadff0c-6700-4995-8828-e0d8976d39e0.png)

`"<missing argument>"` is apparently a placeholder for the actual figure, expected to be 1 in this particular case.

E1izabeth:
Thank you for bug report.

PR #19092, opened 2023-02-17
#19089 Fix variables resolution in commands

E1izabeth:
Example script to check. What was wrong - usage of variables ${var} in commands `@set`, `@echo`, `@include`.
```
@set t = _20230215
@set o = Order${t}

select ${o};

@echo ${t}

@include x${o}y
```

I hope we discuss the possibility of covering such functionality by tests here https://github.com/dbeaver/pro/issues/1419

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/SQLScriptContext.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/SQLScriptContext.java
@@ -17,10 +17,12 @@
 
 package org.jkiss.dbeaver.model.sql;
 
+import org.eclipse.jface.text.Document;
 import org.jkiss.code.NotNull;
 import org.jkiss.code.Nullable;
 import org.jkiss.dbeaver.DBException;
 import org.jkiss.dbeaver.model.DBPContextProvider;
+import org.jkiss.dbeaver.model.DBPDataSource;
 import org.jkiss.dbeaver.model.DBPDataSourceContainer;
 import org.jkiss.dbeaver.model.connection.DBPDriver;
 import org.jkiss.dbeaver.model.exec.DBCExecutionContext;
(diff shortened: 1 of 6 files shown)


serge-rider:
I still don't like that we make so many changes to support variables to fix just case sensitivity. Something goes wrong.
I think we need more simple soluition.

E1izabeth:
Supporting quotes for variable names leads to much more changes and affects not only sql editor variables. Also this solution uses already implemented functionality and doesn't bring new changes. I don't think that there is so many changes. If we don't want to use existing functionality, then we need to write the same thing again which is weird.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -1089,19 +1089,28 @@ public static String stripColumnTypeModifiers(String type) {
     }
 
     public static void fillQueryParameters(SQLQuery sqlStatement, List<SQLQueryParameter> parameters) {
+        String rawQuery = sqlStatement.getText();
+        String query = fillParameters(rawQuery, parameters);
+        sqlStatement.setText(query);
+        //sqlStatement.setOriginalText(query);
```
ShadelessFox, line 1095:
```suggestion
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -1089,19 +1089,28 @@ public static String stripColumnTypeModifiers(String type) {
...
+    }
+
+    /**
+     * Substitute variable values to the provided text
+     */
+    public static String fillParameters(String text, List<SQLQueryParameter> parameters) {
```
ShadelessFox, line 1101:
```suggestion
    @NotNull
    public static String fillParameters(@NotNull String text, @NotNull List<SQLQueryParameter> parameters) {
```

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/SQLScriptContext.java
```diff
@@ -286,27 +316,29 @@ public boolean fillQueryParameters(SQLQuery query, boolean useDefaults) {
...
-            }
+            resolveParameters(parameters);
         }
 
         SQLUtils.fillQueryParameters(query, parameters);
 
         return true;
     }
 
+    private void resolveParameters(List<SQLQueryParameter> parameters) {
```
ShadelessFox, line 327:
```suggestion
    private void resolveParameters(@NotNull List<SQLQueryParameter> parameters) {
```

PR #19103, opened 2023-02-19
#19089 FIx variables replacement in echo, set, include commands

### LINKED ISSUE #19076: @set commands don't resolve variables in v22.3.4
### Description

Starting (I think) with version 22.3.4, @set commands don't properly resolve and apply variables. This worked as late as version 22.2.3. It did not work in 22.**3**.3, but was even "more broken" in that version than in 22.3.4, so clearly something relevant changed in the latest release.

Example: If I run this script:
**@set t = _20230215
@set o = Order${t}
select ${o};**

I expect the output to be:
**Order_20230215**

But in v22.3.4, the output is:
**Order**

(I regressed to version 22.3.3, and the output for the same script is
**NULL**
(It doesn't run the @set commands at all?)

--

This is important functionality to me because I load order data into a new table each day (due to large data sets). For my queries to work dynamically across all of those tables, I need to set the date dynamically using the "t" variable.

### DBeaver Version

22.3.4

### Operating System

windows 64 bit

### Database and driver

any db

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Fixed usage of variables ${var} in commands `@set`, `@echo`, `@include`.

Example script to check.
```
@set t = _20230215
@set o = Order${t}

select ${o};

@echo ${t}

@include x${o}y
```

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/commands/SQLCommandEcho.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/commands/SQLCommandEcho.java
@@ -32,7 +32,7 @@ public class SQLCommandEcho implements SQLControlCommandHandler {
     public boolean handleCommand(SQLControlCommand command, SQLScriptContext scriptContext) throws DBException {
         String parameter = command.getParameter();
         if (parameter != null) {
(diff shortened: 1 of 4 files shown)


LonwoLonwo (approved):

Destrolaric (approved):

Matvey16:
Verified

PR #19114, opened 2023-02-20
dbeaver/pro#1408 convert "string" data type to text during data transfer

LonwoLonwo:
2023-02-20 14_33_01-Configure metadata structure

A better data type for the `string` - is a `string`, but not all databases support it. Second choice - is `"text" `or a data type with the `"text"` name inside, like `"longtext"`. The last option - `varchar(100)` - default string value.

In the case of datatransfer from Mongo to MariaDB/MySQL - "text" data type without parameters as a target type - the best choice, imho.


ShadelessFox (approved):

E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -741,6 +741,9 @@ public String getColumnTypeModifiers(@NotNull DBPDataSource dataSource, @NotNull
                         return null;
                     }
                     return "(" + maxLength + ")";
+                } if ("VARCHAR".equals(typeName)) {
+                    // Default max length value for varchar column, because many databases do not support varchar without modifiers
+                    return "(100)";
```
ShadelessFox: Maybe but it higher so the `FEATURE_MAX_STRING_LENGTH` feature is respected?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
```diff
@@ -399,7 +400,10 @@ public static String mapTargetDataType(DBSObject objectContainer, DBSTypedObject
                             }
                         }
                     } else if (targetType == null && dataKind == DBPDataKind.STRING) {
-                        if (typeNameLower.contains(TEXT_DATA_TYPE)) {
+                        if (typeNameLower.contains(TEXT_DATA_TYPE) || STRING_DATA_TYPE.equals(typeNameLower)) {
```
serge-rider: We need something smarter

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -741,6 +741,9 @@ public String getColumnTypeModifiers(@NotNull DBPDataSource dataSource, @NotNull
                         return null;
                     }
                     return "(" + maxLength + ")";
+                } if ("VARCHAR".equals(typeName)) {
```
serge-rider: Please remove it, it breaks entire logic.
+ No constant
+ No `else`

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
LonwoLonwo: I checked. This max values look to scary. 65535 for MySQL. 10485760 for PostgreSQL. Not sure.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBStructUtils.java
LonwoLonwo: Ok, I removed it. Let's just not use varchar without modifiers by default.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
LonwoLonwo: yeah, forgot "else"

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleSQLDialect.java
```diff
@@ -567,6 +567,11 @@ public String convertExternalDataType(@NotNull SQLDialect sourceDialect, @NotNul
                     dataTypeModifies = String.valueOf(sourceTypedObject.getMaxLength());
                 }
                 break;
+            case "STRING":
+                // Default varchar type with default length for the string columns
+                localDataType = OracleConstants.TYPE_NAME_VARCHAR2;
+                dataTypeModifies = String.valueOf(256);
```
ShadelessFox: Redundant string conversion, just use "256"

uslss:
verified

serge-rider (approved):

Issue #18839, opened 2023-01-23
Connect to selected database


michal-bartak:
### Description

Hello
Please correct me if I'm wrong, but I can bet that in the past, newly opened scripts were connecting to a database selected in the database navigator. 

Considering I have `Show all databases` option enabled, and more databases available in a cluster, like in the picture below hitting `CTRL+]` shortcut should open a new script connected to `clients_and_terminals` database. 

![image](https://user-images.githubusercontent.com/7868445/214061340-d8fd8853-b050-4b63-bba0-001f156edb55.png)

Currently, it's not the case. It always connects to a database entered in the "database" field in connection settings. I found the current behavior counter-productive in conjunction with an allowance of showing more databases.

I'm rising it as an issue (see the first sentence). If it had never worked like this, feel free to turn it into feature request

with regards

### DBeaver Version

22.3.3

### Operating System

Fedora 37

### Database and driver

jdbc:postgres

### Steps to reproduce

1. create postgresql database cluster with more databases. ie: `db1`, `db2`
2. create new connection in DBeaver setting database name to `db1`. Select `show all databases` in connection properties
3. Select `db2` database in db explorer
4. press `CTRL+]`
5. newly opened script will be connected to `db1` db (instead of expected `db2`

the 

### Additional context

I use a separate connection for meta information.

LonwoLonwo:
hell @michalk-k 

Did you use connection by the host or by URL?

![2023-01-05 12_12_38-Connect to a database](https://user-images.githubusercontent.com/45152336/214092537-cf7f5330-b06c-4333-9639-a92894c034a0.png)

michal-bartak:
I do use `Host` option

E1izabeth:
Thank you for your issue. Not sure if it worked so earlier or not, but sounds like a good idea.

michal-bartak:
If I'm not mistaken, it was enough to have any item selected (not only the database root) in order to open a script connected to this database.

michal-bartak:
Another indirect proof that this feature was present in the past

![image](https://user-images.githubusercontent.com/7868445/214523649-8d298c69-d5e7-49f3-bdb1-5e818cb77b1e.png)

Context menu intetion is to work in the context of selected item. It's strange it opens connection to database other than the selected one.

E1izabeth:
I can't reproduce it on DBeaver 23.0.5.
@michalk-k could you check if the bug is still here for you in the latest DBeaver version?

[...]

PR #19130, opened 2023-02-21
#19124 PG URL patch revert

### LINKED ISSUE #19124: URL-configured Postgres connections stopped working in 22.3.5
### Description

I have two Postgres DB connections that use URLs and after upgrading to 22.3.5. They both stopped working and it appears that they're trying to connect to the Host settings, so I get error messages saying it can't connect to localhost:5432 because that's the default, even though I have the URL radio button selected.

The other connection has an older database that I no longer use configured in the host settings and for that one I get an password authentication error, because the old username/password no longer work.

### DBeaver Version

Community Edition 22.3.5

### Operating System

Windows 10

### Database and driver

PostgreSQL JDBC Driver

### Steps to reproduce

1. Open version 22.3.5
2. Attempt to create a new PostgreSQL connection
3. Click on the URL radio button
4. Enter a valid Postgres URL
5. Enter the Username and Password
6. Click on Test Connection ...

Expected:
Connection should succeed

Actual:
Get an error saying connection to localhost:5432 failed

Downgrading to 22.3.4 allows this flow, and the same connections that are working in 22.3.4 no longer work after upgrading.

### Additional context

_No response_

serge-rider:


E1izabeth:
Please check if it solves the following issues too 
- [ ] #19098
- [ ] #18839
- [ ] #19163

Matvey16:
Verified

Issue #12559, opened 2021-05-21
Lost connection inconvience - postgres


plotn:
Hi! I'm using dbeaver for year or two and some bug constantly remains - if the connection is lost (even if "for now" it was refsreshed and "good") and we have some (many) windows open, the behaviour is as follows:

1. I try to run script in the old windows;
2. A number of "reconnecting" windows appears (and reconnect finished happily);
3. But whe windows says: "connection was closed".

The only thing helps - copy script to clipboard, then instantly open new windows (ctrl+]), then paste and run. It works perfectly.

Matvey16:
Please clarify some things so we could look deeper into it.
1. What Database are you using? Postgres? 
2. What version of DBeaver are you using (You can click `Help -> About` to look for it)
3. What is disconnecting from the internet? PC with DBeaver or Database server?
4. What script are you executing before encountering this problem?

plotn:
1. Postges :) as written in title (version 12, 13)
2. Always maximally fresh
3. Database (vpn swithces off, so db become unreachable)
4. It doesnt matter (select).

Matvey16:
Please attach logs that are presented after you encounter this problem.
I tried to reproduce it and while the initial run of the script was not possible after loosing connection. Clicking invalidate/reconnect button fixed it, and I was able to run this script in the same window.

plotn:
[copy.log](https://github.com/dbeaver/dbeaver/files/6661990/copy.log)
here the log

uslss:
there is no update on the issue for a long time. if it is still actual - feel free to reopen it

plotn:
@uslss, why did you close the issue? I gave all that you asked from me, problem is still exists, whats wrong?

Matvey16:
@plotn Hello, I was reviewing this issue, and the error `This connection has been closed` can be constantly reproduced, but I need to know if you have tried to use Invalidate/reconnect button. Also useful would be to know your connection specifications. Do you use SSH tunnel, proxy, SSL?

plotn:
@Matvey16 , got this again. Version: 21.2.2.202110030714
Steps:
1. Run query in sql editor. Result: SQL Error [08003]: Соединение уже было закрыто
2. Refresh connection (right click in tree, select "Refresh"
3. Run query again. Same result.
4. Open new tab (Ctrl + ]). Run query inside it. All ok.

No tunnels, no ssh, direct connection (via enterprise VPN).

Matvey16:
[...]

Issue #6903, opened 2019-10-02
DBeaver takes too long to recover from network disconnects.


davidhooey:
#### System information: 
- MacOS Mojave 10.14.6
- DBeaver CE 6.2.0

#### Connection specification:
- Oracle and SQL Server
- VPN

#### Describe the problem you're observing:

1. The initial connection is great.
2. Run some queries.
3. VPN happens to disconnect.
4. Reconnect to VPN.
5. Disconnecting/Reconnecting or invalidating connection takes way too long. I Force Quit DBeaver and restart to avoid long wait period to reconnect.

This is my only annoyance with DBeaver. Everything else is awesome!

ryanvanderpol:
I'm also experiencing this issue.  It often takes several minutes to reconnect once the connection has been broken.  The initial connection takes less than one second, but reconnecting takes FOREVER.  Here are some screenshots.  It just seems to cycle back and forth between "Connecting main..." and "Set connection defaults..."

<img width="261" alt="Screen Shot 2019-11-27 at 12 23 59 PM" src="https://user-images.githubusercontent.com/1936460/69746017-6253d980-1111-11ea-834a-1d22598bb254.png">
<img width="257" alt="Screen Shot 2019-11-27 at 12 24 12 PM" src="https://user-images.githubusercontent.com/1936460/69746019-6253d980-1111-11ea-81ad-0fffbc2835bb.png">

kwmaharaj:
I have same issue. I end up closing dbeaver.  Opening, reconnecting and all is good. If i leave it for a bit, it goes to shit

p91paul:
I see this issue expecially if metadata connection gets stuck. Nothing works, not even manual invalidate or disconnect, and sometimes even after closing dbeaver it stays running in the background and I have to force kill it.
I have timeouts set for query cancel, connect and disconnect, could it be that the metadata connection ignores those timeouts?

ugurbor:
Having similar issues on DBeaver 7.0.1 on Linux. Connection is Oracle. It cannot recover from a disconnected VPN connection at all.

ismailyavuz:
Having similar issues on DBeaver 7.0.4 on MacOS. My database is PostgreSQL 10.12, with and without VPN same problem. Reconnection takes 1-2 minutes.

kpconnell:
[...]

PR #19131, opened 2023-02-21
#17326 try execute recover for generators

### LINKED ISSUE #17326: Automatic reconnect does not work for menu functionality (such as Generate DDL)
### Description

When a script window loses connectivity, it will automatically reconnect when I execute a new statement. However, when a connection has timed out and I select Generate SQL => DDL from a database object's context menu, an error message will pop up saying that the connection is invalid:
image
When I close it with clicking ok, an empty window pops up:
image
There is no point in showing an empty window when the connection fails.

I'd expect the connection to automatically be restored or at least the empty window not to pop up.

### DBeaver Version

22.1.4

### Operating System

Windows 10 x64

### Database and driver

Azure SQL

### Steps to reproduce

Connect to a DB
Expand the table tree view
let the connection time out
right click a table => Generate SQL => DDL
error message will pop up
empty window will pop up

### Additional context

_No response_

Destrolaric:
Closes #17326 
Acceptance:
When connection lost generator dialog will try to revalidate datasource until try limit or success reached

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorDialog.java
@@ -29,10 +29,12 @@ import org.eclipse.swt.widgets.Button;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Group;
 import org.eclipse.ui.IWorkbenchPartSite;
+import org.jkiss.dbeaver.DBException;
 import org.jkiss.dbeaver.Log;
 import org.jkiss.dbeaver.model.DBPScriptObject;
 import org.jkiss.dbeaver.model.DBPScriptObjectExt2;
 import org.jkiss.dbeaver.model.exec.DBCExecutionContext;
+import org.jkiss.dbeaver.model.exec.DBExecUtils;
 import org.jkiss.dbeaver.model.runtime.AbstractJob;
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

Matvey16:
DDL is generated but error log still continues to be populated with errors saying "This connection has been closed"

Destrolaric:
The issue with connection invalidation happens for other parts of DBeaver. Unfortunately, several still unresolved tickets describe the issue @Matvey16 mentioned. For additional information, https://github.com/dbeaver/dbeaver/issues/6903 and https://github.com/dbeaver/dbeaver/issues/12559 could be looked

serge-rider:
Why do you think that #6903 has the same nature? Any ideas how to fix it?

Destrolaric:
> Why do you think that #6903 has the same nature? Any ideas how to fix it?

The issue was reproduced for Oracle by @Matvey16, and it has the same problematic behaviour as Postgre, the issue, for example, is not reproducable for Maria and MySQL. I previously have tried to fix Postgre issues, but unfortunately have no idea where to look.

Matvey16:
I'm not sure that it is connected with #6903 too.
My idea is that the issue in #12559 is causing the current bug in this fix for postgres. And we were not able to consistetly reproduce that issue previously, but now we can, because I see a problem in the fact that DBeaver somehow continues to try reusing already lost connection even after reconnect.
I think that even if we can fix current issue in this PR with some crutches it is better to fix the main problem in #12559 while it is possible to reproduce it

Destrolaric:
@Matvey16 As discussed, I am returning this for QA, the bugs related to sockets are unfortunately DBeaver wide and this ticket unfortunately also affected by them.

E1izabeth:
IMHO we can't accept this PR in the current state. That bug should be fixed first then.

Matvey16:
Verified

PR #19134, opened 2023-02-21
#15519 Use separate connection for sql console if it was set in settings

### LINKED ISSUE #15519: SQL Consoles do not honor "Open separate connection for each editor"
#### System information: 
- DBeaver version: 21.3.4
- Java version: bundled
- Additional extensions: n/a

- Database name and version: Oracle 11
- Driver name: odjbc21
- Do you use tunnels or proxies (SSH, SOCKS, etc)? n/a

I started using SQL Consoles for quick queries in place of SQL Scripts a while ago. However I just recently noticed that they do not honor the "Open separate connection for each editor", all consoles run on the "main" connection. Which means that opening multiple consoles to launch long-running query results in them running sequentially instead of parallely.
I cannot say for sure if this is a regression from some previous version or they always worked this way.

E1izabeth:
SQL Console should be opened in a separate connection when separate connection setting was set and focus is in navigator.


LonwoLonwo (approved):

uslss:
verified

PR #19135, opened 2023-02-21
dbeaver/pro#1293 TE scripts copy fix

serge-rider:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNResource.java
```diff
@@ -416,6 +414,11 @@ public void dropNodes(Collection<DBNNode> nodes) throws DBException {
             protected IStatus run(DBRProgressMonitor monitor) {
                 monitor.beginTask("Copy files", nodes.size());
                 try {
+                    if (!resource.exists()) {
+                        if (resource instanceof IFolder) {
+                            ((IFolder) resource).create(true, true, new NullProgressMonitor());
+                        }
+                    }
```
ShadelessFox, line 421:
```suggestion
                    if (!resource.exists() && resource instanceof IFolder) {
                        ((IFolder) resource).create(true, true, new NullProgressMonitor());
                    }
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNResource.java
```diff
@@ -426,6 +429,9 @@ protected IStatus run(DBRProgressMonitor monitor) {
                                         true,
                                         monitor.getNestedMonitor());
                                 } else {
+                                    if (DBWorkbench.isDistributed() && !CommonUtils.equalObjects(otherResource.getProject(), resource.getProject())) {
```
ShadelessFox, line 432: Projects are global, so their references must match. Is that correct?

```suggestion
                                    if (DBWorkbench.isDistributed() && otherResource.getProject() != resource.getProject()) {
```

Matvey16:
Verified

PR #19136, opened 2023-02-21
CB-3147 get file extension from filename

yagudin10:


alexander-skoblikov (approved):

bundles/org.jkiss.utils/src/org/jkiss/utils/IOUtils.java
```diff
@@ -378,6 +378,11 @@ public static String getFileNameWithoutExtension(Path file) {
     @Nullable
     public static String getFileExtension(Path file) {
         String fileName = file.getFileName().toString();
+        return getFileExtension(fileName);
```
ShadelessFox, line 381:
```suggestion
        return getFileExtension(file.getFileName().toString());
```

PR #19154, opened 2023-02-22
#17724 add special case for object search: search schemas with equal …

### LINKED ISSUE #17724: Columns are read-only then schema and DB having the same name
### Description

This problem seems to be related to the use of schemas.

Run a simple query on a table which is not in the dbo schema, then try to edit the data in the grid. The following error is shown:

Column is read-only: No corresponding table column

### DBeaver Version

Community Edition 22.2.1.202209251341

### Operating System

Windows 11

### Database and driver

SQL Server

### Steps to reproduce

Open query editor
Run a simple query on a table which is in a non-dbo schema. E.g.: select * from MySchema.MyTable.
In the result grid, click in a cell to edit the data.
Error shows.

### Additional context

_No response_
image

LonwoLonwo:
…as database name

2023-02-22 16_53_21-DBeaver Ultimate 22 3 5 - _master_ Script-101


serge-rider:
It looks like a risky fix. Not sure that we shoiuld merge it before 23.0

serge-rider:
Can we add some unit tests for it?

E1izabeth (approved):

ShadelessFox (approved):

MashaKorax:
Verified

PR #19170, opened 2023-02-23
#18999 Treat composite as null if its data is null

### LINKED ISSUE #18999: Regression in 22.3.4: can't edit composite type fields in data editor when initial value is NULL
### Description

When using the data editor to edit a field of a composite type, the new values are only registered if the field is NOT NULL before the edition. Otherwise, the cell doesn't retain the input value. It does work when the initial value of the field is a record containing only NULL values.

Possibly introduced in #18866?

Log entries from trying to edit a NULL composite field (see steps to reproduce below):
```
!ENTRY org.jkiss.dbeaver.model 2 0 2023-02-07 17:37:29.405
!MESSAGE Null owner value for 'supplier_id', row 8

!ENTRY org.jkiss.dbeaver.model 2 0 2023-02-07 17:37:53.614
!MESSAGE Null owner value for 'price', row 8

!ENTRY org.jkiss.dbeaver.model 2 0 2023-02-07 17:38:22.942
!MESSAGE Null owner value for 'name', row 8
```

### DBeaver Version

Community Edition 22.3.4

### Operating System

Windows 10

### Database and driver

PostgreSQL 12.10
PostgreSQL JDBC Driver 42.5.0

### Steps to reproduce

1. Create testing set:
```pgsql
CREATE TYPE inventory_item AS (
	name text,
	supplier_id integer,
	price numeric(5,2)
);

CREATE TABLE test (
	id integer PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
	item inventory_item
);

INSERT INTO test (item)
VALUES
	(('Item 1', 1, 1.99)),
	(('Item 2', 2, NULL)),
	(('Item 3', NULL, 3.5)),
	(('Item 4', NULL, NULL)),
	((NULL, 3, 0.87)),
	((NULL, 4, NULL)),
	((NULL, NULL, 2.34)),
	((NULL, NULL, NULL)),
	(NULL);
```

2. Open table `test` in the data editor. Note the last 2 rows appear the same way, but in row 8 `item` is a record of NULLs, while in row 9 it's just a NULL.

3. Try to update any of the subfields from column `item` in the last row. The cells shouldn't retain the inserted values.

4. Try to update any of the subfields from column `item` in any of the other rows, including the second-to-last. It should work as expected.

### Additional context

No tunnels or proxy
Extensions installed: Office support, git support, Debug and SVG

ShadelessFox:
Also, please check #16822, #15698, and #14113.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreArrayValueHandler.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreArrayValueHandler.java
@@ -165,7 +165,7 @@ public class PostgreArrayValueHandler extends JDBCArrayValueHandler {
         @Nullable Object value,
         @NotNull DBDDisplayFormat format
     ) {
-        if (value == null) {
+        if (DBUtils.isNullValue(value)) {
             return SQLConstants.NULL_VALUE;
         }
 
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreStructValueHandler.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreStructValueHandler.java
@@ -94,7 +94,7 @@ public class PostgreStructValueHandler extends JDBCStructValueHandler {
         }
         try {
             if (object == null) {
-                return null;
+                return new JDBCCompositeStatic(session, structType, null);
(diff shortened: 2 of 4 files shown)


serge-rider:
I'm afraid it may cause regression as it changes model behavior. 
Probably we need unit tests here.

ShadelessFox:
Unit tests for what exactly?

serge-rider:
For NULL composite types. How we hanle them and how we handle composites with all fields set to NULL.
It may be not easy to implement though..

DariaMarkaryan:
can't save added data to NULL-record if 'Show compex columns structure' is off.
checked another tickets - they're OK

https://user-images.githubusercontent.com/43252954/221949539-ac01020c-9bd9-4b75-b298-1e79c62eb125.mp4

DariaMarkaryan:
Verified

PR #19171, opened 2023-02-23
#17411 Recognize WKT stored as binary data

### LINKED ISSUE #17411: Can't change SRID for BLOB columns with geometry data
### Description

If the blob system contains geometry data it is possible to change view format to geometry and select the SRID, but on the actual map the SRID will be selected as simple and can't be changed.

### DBeaver Version

22.1.5

### Operating System

Win 10

### Database and driver

SQLite

### Steps to reproduce

1. Enter geometry data into Blob column e.g. `POINT (10 10)`
2. Change the column format to geometry
3. Cnange the data presentation from grid to spatial
4. Try to cange the SRID

**Expected:** The SRID is changed
**Actual:** The SRID can't be changed form Simple

dbeaver_l6Q2ypxDo5

### Additional context

NULL

ShadelessFox:
Reproducible against PostgreSQL:

```sql
select 'POINT(30 60)'::bytea;
```


Matvey16:
Verified

PR #19183, opened 2023-02-24
#14710 Reduce the amount of spelled text when moving between rows/cols

### LINKED ISSUE #14710: Table viewer is not accessible for blind and visually impaired that are using screen readers
When I execute a query in the script editor and select execute from the context menu, after the table viewer appears I press ctrl+tab to move focus to table viewer, but I can not move through cells and columns by pressing arrow keys or tab and shift+tab.
Also, I can not edit cell values.
I'm running dbeaver in Windows 10, and using the latest version of "Nvda" screen reading software.
I recommend trying to reproduce this issue by using free and open source Nvda software.
I should be able to move through columns and rows by arrow keys or holding some keys +arrow keys (in html / web pages navigation through tables done by holding alt+ctrl+ arrow keys.)

I found other parts of Dbeaver that are not accessible for screen readers, and I highly recommend testing the accessibility by screen reading software.
If you want, I'm here to help and we could have a session if needed.
I'm a java developer and familiar with SWT and willing to contribute if needed.
Thanks.

ShadelessFox:
If just the column was changed:
```
[value] at column [column]
```

If just the row was changed:
```
[value] at row [row]
```

If both row and column were either unchanged or changed:
```
[value] at column [column] row [row] (readonly) [type] (foreign key to [table])
```

Explanation: `[xxx]` denotes placeholders, `(xxx)` denotes optional parts.


uslss:
verified

PR #19192, opened 2023-02-26
#18907 Fix focus on node loading for data source selector dialog

### LINKED ISSUE #18907: "Choose catalog/schema" dialog works badly with keyboard
### Description

Selecting from DB in left side of "Choose catalog/schema" dialog is pretty much impossible using keyboard (while can get into dialog via keyboard ctrl+0), since every key press steals focus and moves it into right side of schema selection (which likely has only single "public" entry anyhow...)

Thus have to use mouse to for selection which makes dialog really akward.

image

### DBeaver Version

Version 22.3.3.202301211053

### Operating System

Windows 11

### Database and driver

Postgres

### Steps to reproduce

1. CTRL + 0 to open "database/schema" selector
2. Try to select DB from left side, via keyboard navigation

=> fails, since focus is stolen to right side list after any key press

### Additional context

_No response_

E1izabeth:


serge-rider:
Let's rename variable `isSetFocusIfNoneType` to something more meaningful. And some comment in code would be useful.,

uslss:
verified

kikonen:
:)

PR #19197, opened 2023-02-27
#19191 fix Postgres Explain Plan execution parameters saving

### LINKED ISSUE #19191: Postgres explain plan and analyze
### Description

When creating eplain-plans for Postgres-SQL-Queries could causs an error when switching on/off the "analyze" option.

### DBeaver Version

Comunity 22.3.5.202302171528

### Operating System

ubuntu 22.04

### Database and driver

Postgres 15.2
JDBC Driver 42.2.24

### Steps to reproduce

1. explain a sql-query
2. switch on all options (analyze, ...)
3. Press OK button
4. You get the plan for your query
5. switch off analyze. now options "wal" and "timing" are disabled
6. Press OK Button
7. Result is "ERROR: EXPLAIN option WAL requires ANALYZE"

### Additional context

Looks like the "wal" and "tuning" options are passed to explain event when "analyze" is off
grafik
grafik

Workarround:
first switch off "wal", "timing" then switch off "analyze".

LonwoLonwo:
Please check different PG instances, including PG > 12.

COSTS, TIMING and SUMMARY are TRUE by default, so you will not see it in the query if they enabled. If they disabled - then will be name + FALSE, like COSTS + FALSE.


E1izabeth (approved):

DariaMarkaryan:
Verified

PR #19215, opened 2023-02-28
#17703 Fix blob export to txt

### LINKED ISSUE #17703: BLOB export from Maria with new driver didn't show blob content
### Description

Maria driver has been updated from 2.7.0 to 3.0.7. And BLOBs during export to TXT format (but for example nothing changed in CSV export).
Now in TXT export we can't see BLOBs content, it replaced by string '[BLOB]'
Binaries export properties(Not depends on the encoding selected):
image

### DBeaver Version

22.2.1

### Operating System

win10

### Database and driver

Maria  3.0.7

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Blob should be inlined this way `[BLOB[��"���)]]`
(CDATA in xml and inline of same contents in our other transfers were taken into account)


DariaMarkaryan:
Verified

PR #19216, opened 2023-02-28
dbeaver/pro#1445 Update ClickHouse driver to 0.4.1

ShadelessFox:


uslss:
verified both for dbeaver and cloudbeaver

PR #19233, opened 2023-03-01
dbeaver/pro#1454 Respect 'Use Windows trust store' option in headless mode

E1izabeth:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -80,7 +80,8 @@ public class GeneralUtils {
     
     public static final String PROP_TRUST_STORE = "javax.net.ssl.trustStore"; //$NON-NLS-1$
     public static final String PROP_TRUST_STORE_TYPE = "javax.net.ssl.trustStoreType"; //$NON-NLS-1$
-    
+    public static final String VALUE_TRUST_STRORE_TYPE_WINDOWS = "WINDOWS-ROOT"; //$NON-NLS-1$
```
ShadelessFox:
```suggestion
    public static final String VALUE_TRUST_STORE_TYPE_WINDOWS = "WINDOWS-ROOT"; //$NON-NLS-1$
```

ShadelessFox (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #19274, opened 2023-03-06
dbeaver/pro#1458 return previous SQL Server driver version; disable t…

LonwoLonwo:
…rustCertificate option by default


Matvey16:
Verified

PR #19297, opened 2023-03-08
dbeaver/pro#1245 Detach result tabs

ShadelessFox:
image


E1izabeth:
Do queries and results still come to sql terminal from such result set?

ShadelessFox:
You can't detach the SQL terminal.

E1izabeth:
Do SQL Terminal still get updates when you detached tab and press Refresh?

ShadelessFox:
No. The detached tab behaves like it was disposed, so the terminal is no longer listening for its updates.

E1izabeth:
Can't we subscribe on these events for the detached tab?

ShadelessFox:
I don't think it's necessary. Why?

E1izabeth:
For me it looks like the functionality of fetching results for the concrete tab by SQL Terminal becomes broken after tab detaching.

serge-rider:
I think we should detach results to the bottom placeholder, not in editors tab. Thus it should be a view, not an editor.

ShadelessFox:
Views can't be dirty, and we must manually close them upon exiting the application. Not sure if it's a good trade-off.

MashaKorax:
Verified

PR #19304, opened 2023-03-09
Cb 2852 qm redesign

yagudin10:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/qm/QMUtils.java
```diff
@@ -194,9 +195,22 @@ public static QMEventCriteria createDefaultCriteria(DBPPreferenceStore store) {
             }
         }
         criteria.setQueryTypes(queryTypes.toArray(new DBCExecutionPurpose[0]));
+        var userId = getQmUserId(application.getWorkspace().getWorkspaceSession());
```
serge-rider: I don't like it. We shouldn't pass it as a parameter. Server must get user name from the context.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/controls/querylog/QueryLogViewer.java
```diff
@@ -1136,14 +1132,37 @@ protected void buttonPressed(int buttonId) {
...
+            if (dataSource != null) {
+                return container.getDataSource().getSQLDialect();
+            }
+            DBPDriver driver = DataSourceProviderRegistry.getInstance().findDriver(container.getDriver().getId());
+            if (driver == null) {
+                return null;
+            }
+            driver = driver.createOriginalCopy();
+            try {
+                return driver.createOriginalCopy().getScriptDialect().createInstance();
```
serge-rider: Why do we need `createOriginalCopy` here?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/controls/querylog/QueryLogViewer.java
yagudin10: fixed

serge-rider (approved):

PR #19307, opened 2023-03-09
CB-3209 Add instance ID to web session

EmYar:


plugins/org.jkiss.dbeaver.model.event/src/org/jkiss/dbeaver/model/websocket/event/WSEventTopic.java
```diff
@@ -20,6 +20,7 @@
 public enum WSEventTopic {
     SERVER_CONFIG("cb_config"),
     SESSION_LOG("cb_session_log"),
+    WEBSOCKET("cb_websocket"),
```
alexander-skoblikov: let's call this topic as `cb_session`, in the future there also will be 'expired session event'

EmYar: Renamed

alexander-skoblikov (approved):

PR #19308, opened 2023-03-09
#19288 External schema read: ignore database

### LINKED ISSUE #19288: External schemas not listed correctly in the Navigator Pane
### Description

External schemas are not recognized correctly in the 23.0.0 version. 

In the Navigator pane external schemas do not contain any tables. At the same time it is possible to query on the external tables so the connection itself is working correctly. 
- Usually in the previous versions there was a different icon that marks external schemas - this time the icon for external schema is the same as next to the non-external schemas. 
- There is no table visible under Tables in the external schema (they were visible in the previous versions).
- External schema is listed in the `SELECT * FROM svv_external_schemas` and it's possible to query on its tables. But it's not visible in the Navigator pane. 
- There are no proposals when I try to query on the external tables, but query works when typed manually

I assume it might be caused by https://github.com/dbeaver/dbeaver/commit/da9dd68aab1677d16f6ef46f0699a463cded31d2

### DBeaver Version

23.0.0

### Operating System

Windows11

### Database and driver

1. Redshift
2. com.amazon.redshift.jdbc42.Driver

### Steps to reproduce

1. Connect to Redshift using default HOST settings with provided: 
- Host/Instance
- Database
- Username
- Password
2.  Once connected I can see a list of all schemas, including the external one, but the Navigator Pane doesn't show the external schema's content.

### Additional context

_No response_

serge-rider:


ciberboy:
Issue has been resolved in scope of 23.0.1 for Amazon AWS but issue still remains for Azure Databricks. Please re-open this issue

PR #19338, opened 2023-03-11
#8394 (separate) column names copy dialog implemented

### LINKED ISSUE #8394: Copy field with type

plotn:
image

... As on picture.

Tested with Postgres


ShadelessFox (commented):
What happened to the old `CopyConfigDialog`? It had extra options for the quote character and row delimiter.

LonwoLonwo:
Hello @plotn 

So, first of all, you just ignored the current "copy column names" (CopyConfigDialog) dialog and added another version? 

The second one - not all databases wait for the `NOT NULL` modifier. Some of them use this column status by default and wait for NULL instead. Not all databases use `NOT NULL` before the `DEFAULT` keyword. We have special logic for this in our SQLTableColumnManager. 
It is hard to implement in ResultSet classes, so my suggestion - is to remove the "DEFAULT" part.

And please use the existing CopyConfigDialog.

plotn:
Lets discuss it. I think this dialog it point to replace. 
Scenario 1 (not mine):
Copy row numers. I select rectangular area in the grid, press popup menu, click "copy row numbers". The copy dialog appeared:
image

In fact - only one of the setting's used - "Row Delimitier", others are ingored.

Scenario 2 (mine):
Selection is the same, I press "Copy column names". The same dialog appeared!!! But there is only one setting "plays" - column delimitier, others are ignored.

These are two (should be) different kinds of dialogs (for different purpuses), aren't they??? We can, sure, place everything in the one dialog (showing/hiding/disabling of controls, but what for?). 
Nota bene - there is "special copy dialog" (for data) and it is differs from this one.

plotn:
> not all databases wait for the `NOT NULL` modifier. Some of them use this column status by default and wait for NULL instead

can you point me - which of them? I thought this is strict SQL standard, but I may be wrong

plotn:
> Not all databases use `NOT NULL` before the `DEFAULT` keyword

Can you point me - which of them?

plotn:
> special logic

I'll try ro investigate

plotn:
> is to remove the "DEFAULT" part

Hmm.... but what for? It can be switched off by checkbox. Even if it works not in all cases - ok, switch it off and forget. 
But if it helps - why not? Why reduce functionality?

ShadelessFox:
I wouldn't say I like the idea of removing old functionality because of a personal opinion, especially when adding new functionality on top of that.

plotn:
> removing old functionality

The functionality wasnt removed, just added. I checked the functionality,  checked the code, gave detailed explanation, examples. 

Lets make coordinated decision - should I revert to "old" dialog when "column copying"? Should I disable unapplicable options or hide them?

LonwoLonwo:
> Can you point me - which of them? 

You can take a look at the SQLTableColumnManager#getSupportedModifiers

plotn:
> You can take a look at the SQLTableColumnManager#getSupportedModifiers

Can you pay attention? This code works and collects all the existing modifiers (though I changed modifier of getSupportedModifiers from protected to public). Should I go further with this?

....
List<DBDAttributeBinding> attrs = selection.isEmpty() ? rsv.getModel().getVisibleAttributes() : selection.getSelectedAttributes();
....
for (DBDAttributeBinding attr : attrs) {
....
DBEObjectMaker manager = null; 
                        if (attr.getEntityAttribute() != null)
                            manager = DBWorkbench.getPlatform().getEditorsRegistry().getObjectManager(attr.getEntityAttribute().getClass(), DBEObjectMaker.class);
                        SQLTableColumnManager colManager = null;
                        if (manager instanceof SQLTableColumnManager) {
                            colManager = (SQLTableColumnManager) manager;
                        }                                                
                        for (ColumnModifier modifier: colManager.getSupportedModifiers(attr.getEntityAttribute(), new HashMap<String, Object>())) {
                            modifier.appendModifier(new VoidProgressMonitor(), attr.getEntityAttribute(), buffer, null);
                        }
....
}

ShadelessFox:
There have been no updates since the last review. I don't know what I'm supposed to look at.

plotn:
> There have been no updates since the last review. I don't know what I'm supposed to look at.

see my questions please: 
1) "Should I go further with this?" please. Is it right way?
2) "Lets make coordinated decision - should I revert to "old" dialog when "column copying"? Should I disable unapplicable options or hide them?"

ShadelessFox:
1. You're not supposed to use managers directly - the point was to investigate existing implementations to see which ones don't support the `NOT NULL` modifier before the `DEFAULT`.
2. Please extend the already existing dialog instead.

plotn:
> 2\. Please extend the already existing dialog instead.

Ok. I dont agree, but I'll do (the dialog will be overcomplicated, being partially enabled/disabled in different cases).



> 1. You're not supposed to use managers directly - the point was to investigate existing implementations to see which ones don't support the `NOT NULL` modifier before the `DEFAULT`.

I faced up the problems with it. There is "modificator list", that can be collected to/from/with a table's column - each database collects them in is's own way (ordered). And I see the 2 ways of realization:
1. The simple way - mine. We just add null / not null and default - this will cover almost any database (if they are SQL compliant);
2. Use managers directly for this (with some modifications of them), implementing the same logic and codebase.

By the way, SQL-92 says:

```
 <column definition> ::=
              <column name> { <data type> | <domain name> }
              [ <default clause> ]
              [ <column constraint definition>... ]
              [ <collate clause> ]

         <column constraint definition> ::=
              [ <constraint name definition> ]
              <column constraint>
                [ <constraint attributes> ]

         <column constraint> ::=
                NOT NULL
              | <unique specification>
              | <references specification>
              | <check constraint definition>
```

NB: default is before CONSTRAINT.
Postgres can accept both variants, but stores default after not null.

Anyway - all of this is too complicated, I think. Lets implement only datatype and not null (without variants whether or not we should write NULL (without not)) ? 
It's just helper function, the impact is minimal.

ShadelessFox:
You DON'T have to use managers. You can just concatenate strings like you do now.

1. You propose to remove old functionality because you don't like it. Please, don't do that. You're supposed to extend it instead.
2. Let's go with both `DEFAULT` and `NOT NULL`, but `DEFAULT` must be before `NOT NULL` to match the spec.

E1izabeth:
This PR is stale, so I'm closing it. Please, let me know if you'll continue working on this - we can reopen it.

PR #19345, opened 2023-03-13
dbeaver/dbeaver#16795 fix cache synchronization(hopefully)

### LINKED ISSUE #16795: The created index needs to be disconnected and reconnected before it can be displayed.
#### System information: 
- Operating system (distribution) and version :windows10
- DBeaver version : 22.0.4.202205011839
- Additional extensions
- 
- Database name and version :postgres
- Driver name: org.postgresql.Driver

The created index needs to be disconnected and reconnected before it can be displayed.
Create a new index through the index node below the table. Under the index node at the same level of the table, refresh does not display the created index. After disconnection and reconnection, the created index is display

https://user-images.githubusercontent.com/56383213/173764368-fa655fc6-47a6-48e8-87ad-d741147fef91.mp4

ed.

Destrolaric:
Closes #16795 
It should fix the behavior, but please review it. 
Acceptance:
Could you check if indexes are correctly synchronized and if nothing was broken?
Contains additional fixes:
* Nodes deletion for composite cache should delete both nodes in schema and table

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/cache/JDBCCompositeCache.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/cache/JDBCCompositeCache.java
@@ -390,27 +390,25 @@ public abstract class JDBCCompositeCache<
         // Fill global cache
         synchronized (this) {
             synchronized (objectCache) {
-                if (forParent != null || !parentObjectMap.isEmpty()) {
-                    if (forParent == null) {
-                        // Cache global object list
-                        List<OBJECT> globalCache = new ArrayList<>();
-                        for (Map<String, ObjectInfo> objMap : parentObjectMap.values()) {
-                            if (objMap != null) {
-                                for (ObjectInfo info : objMap.values()) {
-                                    if (!info.broken) {
-                                        globalCache.add(info.object);
-                                    }
+                if (forParent == null) {
+                    // Cache global object list
+                    List<OBJECT> globalCache = new ArrayList<>();
+                    for (Map<String, ObjectInfo> objMap : parentObjectMap.values()) {
+                        if (objMap != null) {
+                            for (ObjectInfo info : objMap.values()) {
+                                if (!info.broken) {
+                                    globalCache.add(info.object);
                                 }
                             }
                         }
-                        // Save precached objects in global cache
-                        for (List<OBJECT> objects : objectCache.values()) {
-                            globalCache.addAll(objects);
-                        }
-                        // Add precached objects to global cache too
-                        super.setCache(globalCache);
-                        this.invalidateObjects(monitor, owner, new CacheIterator());
                     }
+                    // Save precached objects in global cache
+                    for (List<OBJECT> objects : objectCache.values()) {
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/jdbc/cache/JDBCCompositeCache.java
```diff
@@ -390,27 +390,25 @@ protected void loadObjects(DBRProgressMonitor monitor, OWNER owner, PARENT forPa
...
-                    if (forParent == null) {
-                        // Cache global object list
-                        List<OBJECT> globalCache = new ArrayList<>();
-                        for (Map<String, ObjectInfo> objMap : parentObjectMap.values()) {
-                            if (objMap != null) {
-                                for (ObjectInfo info : objMap.values()) {
-                                    if (!info.broken) {
-                                        globalCache.add(info.object);
-                                    }
+                if (forParent == null) {
```
E1izabeth, line 393: I'm not sure it's ok to just throw away the condition

Destrolaric: Well, the `forParent != null` never actually did anything because, if you look at how the logic was structured, we checked `forParent != null || !parentObjectMap.isEmpty()`, then right after that, we checked `forParent == null` which makes the first part of the condition invalid. 

`parentObjectMap` is filled only if the value doesn't already exist in its parent, but we don't actually need it if we just want to update our global cache from 
```
                    for (List<OBJECT> objects : objectCache.values()) {
                        globalCache.addAll(objects);
                    }
```

E1izabeth: Thank you for explanation, it became clearer. I hope you're right

E1izabeth (approved):

Matvey16:
This breaks index deletion in the navigator. Previously index deleted directly from navigator disappeared right after deletion, now it persists until manual refresh

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNModel.java
```diff
@@ -180,11 +180,15 @@ public DBNDatabaseNode findNode(DBSObject object)
...
-    @Nullable
-    public DBNDatabaseNode getNodeByObject(DBSObject object)
-    {
+    /**
+     * @param object object to find Nodes for
+     * @param filterVirtual if virtual values should be filtered out
+     * @return list of nodes associated with object
+     */
+    @NotNull
+    public List<DBNDatabaseNode> getNodeListByObject(DBSObject object, boolean filterVirtual) {
```
ShadelessFox: Maybe call it `getNodesByObject`?

serge-rider:
Looks like too serious change in the model of cachinbg and navigator. It may break almost anything.
I had an impression that can just fix cokmposite cache items add.

serge-rider:
Please try another solution: https://github.com/dbeaver/dbeaver/pull/19440

PR #19380, opened 2023-03-15
CB-1577-H2-driver-update

EmYar:


ShadelessFox:
AFAIK you're not supposed to bump versions of plugins that are shipped with DBeaver.

It makes sense for plugins installed separately so that Eclipse can figure out that they were updated.

plugins/org.jkiss.dbeaver.ext.h2/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %Bundle-Name
 Bundle-SymbolicName: org.jkiss.dbeaver.ext.h2;singleton:=true
-Bundle-Version: 1.0.156.qualifier
+Bundle-Version: 1.0.157.qualifier
```
serge-rider: Please don't change bundle version manually

plugins/org.jkiss.dbeaver.ext.h2/pom.xml
```diff
@@ -9,6 +9,6 @@
      <relativePath>../</relativePath>
   </parent>
   <artifactId>org.jkiss.dbeaver.ext.h2</artifactId>
-  <version>1.0.156-SNAPSHOT</version>
+  <version>1.0.157-SNAPSHOT</version>
```
serge-rider: The same

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.h2/META-INF/MANIFEST.MF
EmYar: Change reverted

plugins/org.jkiss.dbeaver.ext.h2/pom.xml
EmYar: Change reverted

ShadelessFox (approved):

bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
```diff
@@ -1047,4 +1047,19 @@ public static String normalizeResourcePath(@NotNull String resourcePath) {
         resourcePath = resourcePath.replace('\\', '/');
         return resourcePath;
     }
+
+    /**
+     * Replaces the last {@code toReplace} entry in {@code string} with {@code replacement}
+     */
+    @NotNull
+    public static String replaceLast(@NotNull String string, @NotNull String toReplace, @NotNull String replacement) {
```
ShadelessFox, line 1055: Why is it here? Bad merge commit?

EmYar: This is a utility method for String used by the H2 migrator. Is this not the right place for it?

ShadelessFox: Yes, since it's not used anywhere in this PR.

serge-rider (approved):

PR #19385, opened 2023-03-16
#18300 Editor toolbar customization

### LINKED ISSUE #18300: Configure SQL editor toolbar
We want to have customization of SQL editor toolbar.
Additional actins like "Execute in new script", "Export from query" and other may be placed on the vertical toolbar.

By default only execute, execute script and explain plan should be visible.

E1izabeth:
image

By default, only Execute single query, Execute script and Explain plan should be visible.


uslss:
verified

PR #19428, opened 2023-03-20
#3371 add TOP limit support for Sybase and old SQL Server

### LINKED ISSUE #3371: Sybase SQL Anywhere is giving '@p0' error
So I can connect to the DB just fine and see all the tables and such

But when I tried to query, its giving me this error: 
example: Select * From Billing;

Query execution failed
Reason:
SQL Error [137][42W14]: SQL Anywhere Error -260: Variable '@p0' not found

Details shows:
SQL Error [137] [42W14]: SQL Anywhere Error -260: Variable '@p0' not found
  SQL Error [137] [42W14]: SQL Anywhere Error -260: Variable '@p0' not found
    SQL Error [137] [42W14]: SQL Anywhere Error -260: Variable '@p0' not found
      SQL Error [137] [42W14]: SQL Anywhere Error -260: Variable '@p0' not found
        SQL Anywhere Error -260: Variable '@p0' not found
        SQL Anywhere Error -260: Variable '@p0' not found

But if I were to run the same query again with Top 100 or something it works just fine
Example: Select Top 500 * from Billing;

I tried to set the global / local amount of rows of data it returned back but the errors still 
come back

Any ideas what is causing this?

LonwoLonwo:
A TOP limit will be added in the query for Sybase/SQL Server if the setting "Use SQL to limit fetch size" is enabled:

 
2023-03-20 12_43_57-Preferences


uslss:
verified

PR #19430, opened 2023-03-20
#19123 add extra table metadata info properties

### LINKED ISSUE #19123: Created Date/Last DDL Date missing
Would it be possible to add in "Statistics Info" metadata info like Created Date and Last DDL Date for Oracle DB?

I think that this feature is not available on DBeaver and force the developer to write the query for every table in order to know when an object has been created. 
I hope that just under Name and Table Type could be possible to show that info.

LonwoLonwo:
2023-03-20 14_30_15-DBeaver 23 0 1 - NEWTABLE_4

- [ ] Check tables
- [ ] Check views
- [ ] Check materialized views

How to change the timestamp in the "Last Changed" - with one of the following actions:
1. Tuncate Table
2. Add Table Partition
3. Drop Table Partition
4. Add Column
5. Drop Column


E1izabeth (approved):

Matvey16:
Verified

PR #19440, opened 2023-03-21
dbeaver/dbeaver#16795 Items read from composite cache: respect alread…

### LINKED ISSUE #16795: The created index needs to be disconnected and reconnected before it can be displayed.

serge-rider:
…y read parents


Matvey16:
Verified

PR #19485, opened 2023-03-26
#14511 Move erd diagram only once on search

### LINKED ISSUE #14511: Don't move the ER diagram until search is complete
Right now search function in ERD works like this: After any symbol input focus starts moving rapidly from table to table until the search is complete, this looks strange and maybe even scary.
It would be better to move a diagram to target table only after search is finished. Also I suggest adding a delay before search to allow inputting other symbols before search starts.

E1izabeth:


plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/editor/ERDEditorPart.java
```diff
@@ -1378,16 +1378,22 @@ public boolean performSearch(@NotNull String searchString, int options) {
                 }
                 if (!CommonUtils.isEmpty(nodes)) {
                     Color color = UIUtils.getColorRegistry().get(ERDUIConstants.COLOR_ERD_SEARCH_HIGHLIGHTING);
+                    DBPNamedObject focusedNode = null;
                     for (DBPNamedObject erdNode : nodes) {
                         if (matchesSearch(erdNode)) {
+                            if (resultsFound == false) {
```
LonwoLonwo:
```suggestion
                            if (!resultsFound) {
```

LonwoLonwo (approved):

uslss:
verified

PR #19503, opened 2023-03-27
#19319 Fix q-string rule chars unreading

### LINKED ISSUE #19319: [Oracle]  "q" symbol on the end of string cropped
### Discussed in https://github.com/orgs/dbeaver/discussions/19295

<sup>Originally posted by **Kectai** March  8, 2023</sup>
image

dbeaver community23.0.0

oracle19c

The old version used before did not have this problem.</div>

2023-03-10 13_09_53-DBeaver Ultimate 23 0 0 - _orcl_ Script-59

with ; and without

E1izabeth:
The case in the issue should not be reproducible anymore.

Also q-strings should work. Example: `q'[What's a quote among friends?]'`

Only for Oracle.


LonwoLonwo:
What do you think about the original query?

`SELECT q.* FROM dual q;`

Do you have plans to add it to your tests?

E1izabeth:
I'm not sure this concrete case should be tested somehow, but I added it.

LonwoLonwo (approved):

uslss:
verified

PR #19511, opened 2023-03-28
dbeaver/dbeaver#16776 refresh visuals without layout call

### LINKED ISSUE #16776: Auto arrange forced after show attributes changed even if for a table only
After manually arranging a diagram if I change the show attributes for a selected table from say none to show keys then the whole diagram is auto arranged.

I did see that a keep layout exists somewhere not sure if that would help if I can find it but basically I would simply like to not have arrangement changed after changing the show attributes for a table.

The only thing to do right now is to perhaps set show attributes for each table prior to any manual  arranging.

Destrolaric:
Closes dbeaver/dbeaver#16776 
Refreshes attribute visibility without calling relayout. We possibly want this for attribute styles.

https://user-images.githubusercontent.com/20579256/228212533-f28cd8ad-c6c1-4e9f-8185-74130081546b.mp4


E1izabeth (approved):

LonwoLonwo (approved):

MashaKorax:
verified

PR #19512, opened 2023-03-28
#19222 Support artifact classifier

### LINKED ISSUE #19222: Support artifact classifier when downloading driver files from Maven
ClickHouse JDBC driver supports multiple protocols(http, tcp, and grpc), hence it comes with multiple classifiers for user to choose. As of now, DBeaver does not support user to specify classifier when downloading driver files from Maven.
image

It would be great if we can add one more dropdown list next to `Version` column shown above to allow user to select classifier(e.g. `all`, `cli`, `grpc`, and `http` for ClickHouse, and perhaps empty for others)

N/A.

**Additional context**
The discussion started with a bug reported at ClickHouse/clickhouse-java#1257.

serge-rider:


Matvey16:
Verified

PR #19516, opened 2023-03-28
dbeaver/pro#1241 update icon

Destrolaric:
Closes dbeaver/pro#1241
Updated icon to avoid confusion of having two same icons for a different types of preferences when opened in VQB. 
image


Matvey16:
Verified

PR #19544, opened 2023-03-30
dbeaver/pro#1516 change Confirmations page dialog behaviour - keep st…

LonwoLonwo:
Change Confirmations page dialog behavior - keep the status of the confirmation dialog

2023-03-30 14_33_40-DBeaver Ultimate 23 0 1 - _language

Important - the "Confirm" column values are not editable if the "Show" column has been checked (we can't store value from the "Confirm" status then)

Behavior for now:

A confirmation dialog will be shown if the "Show dialog" button is enabled. The "Confirm" button is also enabled in this case.
If the "Show dialog" button is enabled, the "Confirm" button status will be used instead of the dialog showing. 
If the "Confirm" button is enabled - it means always do the action (ex: Revert changes during the object refresh). 
If disabled - it means "Never" (ex: Do not revert changes after object refreshing enabled action - do not refresh, wait for the user actions)


uslss:
verified

PR #19581, opened 2023-04-02
dbeaver/dbeaver#18316 INSERT and UPDATE keywords completion

### LINKED ISSUE #18316: Sql keyword `into` is not suggested after `insert`
### Description

`INSERT INTO ...` is a common well-known sql construct that doesn't depend on any specific sql dialect. However, dbeaver doesn't suggest `INTO` completion after `INSERT` when typing

### DBeaver Version

Community Edition 22.2.5

### Operating System

Windows 11 22H2

### Database and driver

PostgreSQL 14.5

### Steps to reproduce

1. Open SQL editor
2. Type `insert`
3. Type space and invoke completion list

**Expected behaviour:**
`into` is in the completion list and I can just hit `Tab` or `Enter` to insert it

**Actual behaviour:**
Suggested completion list is empty:
dbeaver_k6SMVKSiLL

serge-rider:


MashaKorax:
verified

PR #19584, opened 2023-04-02
dbeaver/dbeaver#19583 Fix statements create attempts if is not intern…

### LINKED ISSUE #19583: Too many attempts to reopen failed connection
### Description

Currently when we execute SQL query against failed connection (e.g. when connection was lost because of network failure) then we try to re-establish connection too many times.
You cann it from logs lines `Initiate connection to XXX`.

Reason 1: we are trying to re-execute every SQL statement 3 times when connection is not open (e.g. when first attempt to invalidate connection was failed).

Reason 2: when we have nested `tryExecuteRecover` calls then each of them tries to invalidate all contexts. Thus, in case of failed SQL query we are trying to reopen each connection 6 times instead of 1.

### DBeaver Version

23.0.1

### Operating System

Any

### Database and driver

PostgreSQL

### Steps to reproduce

1. Open SQL editor with some alive connection.
2. Disable network
3. Try to execute a query
4. See logs

### Additional context

_No response_

serge-rider:
…al driver error. Do not invalidate after nested invalidation fail.


uslss:
verified

Issue #17635, opened 2022-09-13
"Insert Case" for Code Completion cannot correctly insert certain case sensitive functions in Clickhouse


eschaubert:
### Description

Some functions were added to the code completion dictionary for clickhouse, but when inserting them, the "insert case" as per settings for code completion in DBEAVER can only be either UPPER or LOWER or DEFAULT. But this does not work for Clickhouse specific function, which are all case-sensitive. The case sensitivity is by design and was a deliberate decision by the clickhouse developers

For example the function: toStartOfMonth() works only if its written exactly like that and tostartofmonth() or TOSTARTOFMONTH() will not work.

This case-sensitivity for clickhouse specific function cannot be changed via server-side settings.

### DBeaver Version

Community Edition 22.1.5

### Operating System

Linux PopOs

### Database and driver

Clickhouse 
Yandex ClickHouse driver (server version 21.3+)

LonwoLonwo:
Ok, thanks for the report. We will see what we can do.

PR #19609, opened 2023-04-05
#17635 store functions SQL words in the original case

LonwoLonwo:
2023-04-05 12_04_10-DBeaver Ultimate 23 0 2 - _default_ Script-71

Please check different-case databases. Clickhouse - as mixed case, PostgreSQL as lower case, Oracle as upper case. Or other.
With Hippie Engine and without. With default case autocomplete and with upper/lower.


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -202,21 +204,25 @@ public TreeSet<String> getDataTypes(@Nullable DBPDataSource dataSource) {
...
+    @Nullable
     @Override
     public DBPKeywordType getKeywordType(@NotNull String word) {
-        return allKeywords.get(word.toUpperCase(Locale.ENGLISH));
+        for (Map.Entry<String, DBPKeywordType> entry : allKeywords.entrySet()) {
+            if (word.equalsIgnoreCase(entry.getKey())) {
+                return entry.getValue();
+            }
+        }
+        return null;
```
ShadelessFox: This can slow things down, including formatting, autocompletion, and syntax highlighting.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -202,21 +204,25 @@ public TreeSet<String> getDataTypes(@Nullable DBPDataSource dataSource) {
...
 
     @NotNull
     @Override
     public List<String> getMatchedKeywords(@NotNull String word) {
         word = word.toUpperCase(Locale.ENGLISH);
         List<String> result = new ArrayList<>();
-        for (String keyword : allKeywords.tailMap(word).keySet()) {
-            if (keyword.startsWith(word)) {
+        for (String keyword : allKeywords.keySet()) {
+            if (keyword.toUpperCase(Locale.ENGLISH).startsWith(word)) {
```
ShadelessFox: Same here

ShadelessFox (commented):
This is a performance regression, for sure. Need to think of another solution.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/format/SQLFormatterConfiguration.java
```diff
@@ -106,8 +107,14 @@ public void setSourceEncoding(@NotNull String sourceEncoding) {
         this.sourceEncoding = sourceEncoding;
     }
 
-    public boolean isFunction(String name) {
-        return syntaxManager.getDialect().getFunctions().contains(name.toUpperCase(Locale.ENGLISH));
+    public boolean isFunction(@NotNull String name) {
+        Set<String> functions = syntaxManager.getDialect().getFunctions();
```
serge-rider: It is very slow. We should keep caqche in upper case. 
We may keep original value too

serge-rider (changes requested):

serge-rider:
Let's keep originalo case for all keywords (not just functions) and all dialects (not just Clickhouse).
Turn TreeSet into TreeMap and hold original case in value.

serge-rider:
Please move CH-related changes to https://github.com/dbeaver/dbeaver/pull/19706

PR #19630, opened 2023-04-06
Refactored org.jkiss.utils class and methods using diffrent refactoring techniques

dhruvilDal:
## Rename Variable:
- The variable was named `rand` for the random number which I have changed to `randomNumber` to make it more descriptive.

## Introduce explaining variable:
- I created a new boolean variable called `isSpecialChar`, assigned the result of the long conditional statement to it, and then used this variable in place of the original expression in the if statement. This change makes the code easier for readers to understand.

## Extract Method:
- Extracted methods from the `parseLine` method to improve readability, maintainability, and modularity of the code. Created three new methods `handleNullNextLine`, `handlePendingNextLine` and `handleInQuotes`.

## Move Method:
- Moved array related method `joinStrings` from the `CommonUtils` class to `ArrayUtils` class which is more appropriate for this method. Also related changes in `ExasolTableIndex.java` file that imports the method and related changes in `ArrayUtilsTest.java` and `CommonUtilsTest.java` files.

## Extract Class:
- Extracted class `TypeUtils` from class `BeanUtils`. All the datatype related methods are moved to `TypeUtils.java` which improves the structure and understandability of the code and also makes code more maintainable. Made a new file `TypeUtilsTest.java` for the test cases of the TypeUtils class and changed the related code of the project to call the moved method from TypeUtils class instead of BeanUtils class.

## Replace Conditional with Polymorphism:
- Made a new interface `Decoder.java` Implemented classes `Decoder1`, `Decoder2` and `Decoder3`. Modified the method `decode0` of class `Base64`. Changed the switch statement logic and replaced the code to implement the logic with polymorphism using Decoder classes. Now the code is more modular, easy to understand, maintain, and extend.


E1izabeth:
Thank you for PR! Unfortunately, we can't accept such refactoring as it's a common classes for many of our projects, including private ones. This will completely break build for them.

PR #19657, opened 2023-04-10
dbeaver/pro#1386 improve external file reader for txt, xml, json

Destrolaric:
Closes dbeaver/pro#1386
Now if txt, JSON or XML editor is opened, an external file will be opened with the .txt,.json or .XML extension

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
@@ -88,6 +88,7 @@ public abstract class AbstractTextPanelEditor<EDITOR extends BaseTextEditor>
     private IValueController valueController;
     private IEditorSite subSite;
     private EDITOR editor;
+    private Path tempFile;
 
     @Override
     public StyledText createControl(IValueController valueController) {
@@ -198,6 +199,13 @@ public abstract class AbstractTextPanelEditor<EDITOR extends BaseTextEditor>
             editor.dispose();
             editor = null;
         }
+        if (tempFile != null) {
+            try {
+                Files.deleteIfExists(tempFile);
+            } catch (IOException e) {
+                log.warn(e);
+            }
+        }
     }
 
     protected EDITOR getTextEditor() {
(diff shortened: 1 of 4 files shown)


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -294,6 +299,24 @@ public void primeEditorValue(@NotNull DBRProgressMonitor monitor, @NotNull Style
         }
     }
 
+    @Nullable
+    @Override
+    public Path getExternalFilePath(@NotNull StyledText control) {
+        try {
+            Path file = Files.createTempFile(DBWorkbench.getPlatform()
+                .getTempFolder(new VoidProgressMonitor(), getFileFolderName()), "file", getFileExtension());
```
E1izabeth: What if we're opening two files simultaneously? I'm not sure it's ok to call temp file just `file`.

Destrolaric: They assigned a unique number, so it will not cause any issues. But it may be a good idea to have only one simultaneously.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -303,10 +304,10 @@ public void primeEditorValue(@NotNull DBRProgressMonitor monitor, @NotNull Style
     @Override
     public Path getExternalFilePath(@NotNull StyledText control) {
         try {
-            Path file = Files.createTempFile(DBWorkbench.getPlatform()
-                .getTempFolder(new VoidProgressMonitor(), getFileFolderName()), "file", getFileExtension());
-            Files.writeString(file, control.getText());
-            return file;
+            Files.deleteIfExists(tempFile);
+            this.tempFile = Files.createTempFile(DBWorkbench.getPlatform().getTempFolder(new VoidProgressMonitor(), getFileFolderName()), "file", getFileExtension());
+            Files.writeString(tempFile, control.getText());
```
E1izabeth: Shouldn't we check if the operation succeeded somewhere here?

Destrolaric: Method has nullable annotation, so it should be looked at later anyway

serge-rider:
Let's delete temp files in panbel dipose method. Otherwise we produce a lot of temp files while app is running.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -294,6 +307,25 @@ public void primeEditorValue(@NotNull DBRProgressMonitor monitor, @NotNull Style
         }
     }
 
+    @Nullable
+    @Override
+    public Path getExternalFilePath(@NotNull StyledText control) {
+        try {
+            Files.deleteIfExists(tempFile);
```
serge-rider: Potential NPE?

serge-rider (changes requested):

Matvey16:
Verified

serge-rider (approved):

PR #19680, opened 2023-04-12
#19502 Result Set Viewer bottom toolbar customization

### LINKED ISSUE #19502: Configure bottom toolbar in Data Editor
We want to make the bottom toolbar in Data Editor customizable.

 We need to discuss which buttons should be visible by default and which extra actions can be added.

E1izabeth:
image

Toolbar should update on preference change and relayout should be applied - so no accidential empty space appear.
Toolbar state should be saved between launches.


MashaKorax:
verified

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageToolbarCustomization.java
```diff
@@ -187,7 +187,7 @@ public PrefPageToolbarCustomization() {
         toolBarImage = findBundleImage(PlatformUI.PLUGIN_ID, "$nl$/icons/full/obj16/toolbar.png"); //$NON-NLS-1$
         
         toolBarNodes = ToolBarConfigurationRegistry.getInstance().getKnownToolBars().stream()
-            .map(ToolBarNode::new).collect(Collectors.toList());
+            .sorted((a, b) -> a.getName().compareTo(b.getName())).map(ToolBarNode::new).collect(Collectors.toList());
```
ShadelessFox:
```suggestion
            .sorted(Comparator.comparing(ToolBarConfigurationDescriptor::getName))
            .map(ToolBarNode::new)
            .collect(Collectors.toList());
```

plugins/org.jkiss.dbeaver.ui.editors.data/OSGI-INF/l10n/bundle.properties
```diff
@@ -243,4 +245,6 @@ confirm.reset_panels_content.title = Proceed and reset panel contents
 confirm.reset_panels_content.tip = Warns about panels with unsaved changes closing
 
 confirm.close_resultset_edit.message = Result set was edited. Do you want to save changes in the database?
-confirm.close_resultset_edit.title = Save result set edit
\ No newline at end of file
+confirm.close_resultset_edit.title = Save result set edit
+
+org.jkiss.dbeaver.toolBarConfiguration.resultsetViewer.bottom = Result set bottom toolbar
```
ShadelessFox: Missing trailing line

```suggestion
confirm.close_resultset_edit.title = Save result set edit

org.jkiss.dbeaver.toolBarConfiguration.resultsetViewer.bottom = Result set bottom toolbar

```

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -1642,8 +1646,37 @@ public void redrawData(boolean attributesChanged, boolean rowsChanged)
...
+    private final UIJob statusBarLayoutJob = new UIJob("Pending resultset view status bar relayout") {
+        @Override
+        public IStatus runInUIThread(IProgressMonitor monitor) {
+            boolean changed = false;
+            for (ToolBarManager toolbarMan: toolbarList) {
+                ToolBar toolbar = toolbarMan.getControl();
+                boolean wasCollapsed = toolbar.getData() != null;
+                toolbar.setLayoutData(toolbar.getItemCount() > 0 ? null : new RowData(0, 0));
+                boolean nowCollapsed = toolbar.getData() != null;
+                changed |= (wasCollapsed == nowCollapsed);
```
ShadelessFox: Perhaps it should be `wasCollapsed != nowCollapsed` instead?

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -1642,8 +1646,37 @@ public void redrawData(boolean attributesChanged, boolean rowsChanged)
         //this.updateStatusMessage();
     }
 
+    private final UIJob statusBarLayoutJob = new UIJob("Pending resultset view status bar relayout") {
+        @Override
+        public IStatus runInUIThread(IProgressMonitor monitor) {
+            boolean changed = false;
+            for (ToolBarManager toolbarMan: toolbarList) {
```
ShadelessFox:
```suggestion
            for (ToolBarManager toolbarManager : toolbarList) {
```

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -1642,8 +1646,37 @@ public void redrawData(boolean attributesChanged, boolean rowsChanged)
...
+                changed |= (wasCollapsed == nowCollapsed);
+            }
+            if (changed) {
+                ResultSetViewer.this.getControl().layout(true, true);
+            }
+            return Status.OK_STATUS;
+        }
+    };
+    
+    private void collapseToolbarIfEmpty(ToolBar toolbar) {
```
ShadelessFox: Please specify nullability annotations for the method's parameter.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -1642,8 +1646,37 @@ public void redrawData(boolean attributesChanged, boolean rowsChanged)
         //this.updateStatusMessage();
     }
 
+    private final UIJob statusBarLayoutJob = new UIJob("Pending resultset view status bar relayout") {
```
ShadelessFox: Why is layout performed in a job anyways? Can't it be run using `UIUtils.asyncExec` instead?

IIRC you don't even have to run it on the EDT thread since you're inside an event handler for a UI event.

E1izabeth: Because the job scheduling mechanism guarantees that the job would be executed only once after all the already running UI activities would be finished. It is possible to have a sequence of resize events here as a result of settings change for multiple toolbar contributions at once. So we need only one relayout then, after the configuration update would be applied. The use of asyncExec leads to an unnecessary sequence of relayouts giving an unappropriate result according to my tests.

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageToolbarCustomization.java
```diff
@@ -187,7 +188,7 @@ public PrefPageToolbarCustomization() {
         toolBarImage = findBundleImage(PlatformUI.PLUGIN_ID, "$nl$/icons/full/obj16/toolbar.png"); //$NON-NLS-1$
         
         toolBarNodes = ToolBarConfigurationRegistry.getInstance().getKnownToolBars().stream()
-            .map(ToolBarNode::new).collect(Collectors.toList());
+            .sorted(Comparator.comparing(ToolBarConfigurationDescriptor::getName)).map(ToolBarNode::new).collect(Collectors.toList());
```
ShadelessFox: Please split it over several lines to make it more readable

```suggestion
            .sorted(Comparator.comparing(ToolBarConfigurationDescriptor::getName))
            .map(ToolBarNode::new)
            .collect(Collectors.toList());
```

ShadelessFox (commented):
I think we can listen to preference changes instead and perform layout if needed.

E1izabeth:
PreferenceStore' event would be triggered for any preference change while the context property reevaluation triggerend only once when all the preference changes already finished. And also ActionUtils is responsible for triggering property reevaluation since 2018, so it is the right place for the corresponding event listeners registration, I assume.

MashaKorax:
dbeaver_OlS79hdOJc
 SQL editor does not work at all

**Event loop exception**
`An exception stack trace is not available.`

LonwoLonwo (approved):

uslss:
verified

PR #19706, opened 2023-04-14
dbeaver/dbeaver#17635 Keep original functions, data types and keyword…

### LINKED ISSUE #17635: "Insert Case" for Code Completion cannot correctly insert certain case sensitive functions in Clickhouse

serge-rider:
…s case


E1izabeth:
Why did you decide to store all function keywords in lower case?

LonwoLonwo:
Please check ClockHouse, PostgreSQL, Oracle, and other databases with different cases with "default" autocomplete.
Also, please test the performance on the big script opening.

And please check Mongo data types in EE:
        "Boolean",
        "Byte",
        "Char",
        "Double",
        "Float",
        "Int",
        "Long",
        "Short",
        "void",

uslss:
verified

PR #19736, opened 2023-04-18
Dbeaver/dbeaver#1155 sql word iterator

### LINKED ISSUE #1155: Don't consider underscore a word separator
Right now, keys like Ctrl-Right that act on words consider underscore a word separator, but I pretty much always would rather it not (navigating through column names that are entirely_too_long_and_take_forever_to_scroll_through is currently pretty painful).  Is there a way change what it considers a word separator?  Or could an option for this be added?

serge-rider:


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLPreferenceConstants.java
```diff
@@ -99,11 +96,13 @@ public static EmptyScriptCloseBehavior getByName(String name) {
     public static final String SQLEDITOR_CLOSE_COMMENTS                = "SQLEditor.closeComments";
     public static final String SQLEDITOR_CLOSE_BLOCKS                  = "SQLEditor.closeBlocks";
 
+    public static final String SMART_WORD_ITERATOR                     = "SQLEditor.smartWrdIterator";
```
maffe: Is this a typo (`Wrd` instead of `Word`) or is there some kind of (length?) restriction?

serge-rider: Yep :)

MashaKorax:
verified

PR #19738, opened 2023-04-18
#18576 read depended for identity table column sequence; add sequence…

### LINKED ISSUE #18576: No Identity Sequence Options
We cannot access options for IDENTITY SEQUENCE in the DDL tab. 

For example, we specify the following column properties in the SQL Editor:

```
CREATE TABLE public.color1 (
	color_id int4 NOT NULL GENERATED BY DEFAULT AS IDENTITY(START WITH 10 INCREMENT BY 10),
	color_name varchar NOT NULL);
```

The information in brackets is then not displayed in the DDL tab:

image

LonwoLonwo:
read depended on identity table column sequence -> add sequence body in the table DDL

2023-04-18 17_02_28-DBeaver Ultimate 23 0 3 - color1

The table reading was changed, so please check not just the case from the ticket but also:
- table reading in Redshift (table from ticket can be created, so this code is not for Redshift)
- table reading in Greenplum
- table reading in Cockroach

https://www.postgresql.org/docs/7.3/catalog-pg-depend.html this is the base system view from a long time ago, and this is great.


uslss:
verified

PR #19746, opened 2023-04-19
dbeaver/pro#1484 store special setting in the workspace file to avoid…

LonwoLonwo:
… setting resetting

aux PR


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DesktopWorkspaceImpl.java
```diff
@@ -57,6 +57,7 @@ public DesktopWorkspaceImpl(DBPPlatform platform, IWorkspace eclipseWorkspace) {
 
         loadExtensions(Platform.getExtensionRegistry());
         loadExternalFileProperties();
+        loadWorkspaceExtraProperties();
```
serge-rider, line 60: We do not need it here. It is a pointless code in this class.

serge-rider (changes requested):

PR #19759, opened 2023-04-20
#19231 change generation of ALTER table SQL for foreign tables

### LINKED ISSUE #19231: dBeaver doesn't allow table changes in foreign table in postgresql
### Description

dBeaver apparently doesn't know how to correctly generate SQL needed to alter or update foreign tables in PostgreSQL.
If you have a foreign table in PostgreSQL and you use dBeaver to add a new column, alter an existing column, etc. it fails because dBeaver apparently doesn't know how to generate the proper SQL statements.  At least in PostgresQL.

Ex:  I have a foreign table named foo and I want to add a new column named test.  This is the SQL the dBeaver GUI generates when you go to persist the change:
`ALTER TABLE bar_fdw.foo ADD test varchar NULL;`

what it should have generated is:
`ALTER FOREIGN TABLE bar_fdw.foo ADD test varchar NULL;`

So every time one tries to use dBeaver to manage foreign tables, it fails.  Typically with the error:
"Table bar_fdw.foo doesn't exist."

Can someone look into fixing dBeaver's inability to work with PostgreSQL foeign tables correctly?

thanks.

p.s. it may or may not be related to this existing issue https://github.com/dbeaver/dbeaver/issues/16372 from back on 2022-04-29, but at a minimum it demonstrates that dBeaver still wants to conflate these vastly different object types together.

### DBeaver Version

Community Edition 22.3.5.202302171528

### Operating System

Windows 10

### Database and driver

_No response_

### Steps to reproduce

1. View Table on a foreign table in a PostgreSQL database
2. Attempt to do anything with the DDL of that table (add column, alter column, drop column, etc.)
3. Hit the save button
4. See that the presented SQL to persist this change is incorrect
5. Watch the database throw an error because dBeaver generated invalid SQL

### Additional context

_No response_

LonwoLonwo:
2023-04-20 14_05_35-fb_test_table2 - Persist Changes

Please check 
- Foreign tables changing (alter table / rename table / add column / drop column / rename column / alter column) (some changes are not allowed for foreign tables - it is expected (like changing the row level security)
- Ordinary tables changing (alter table / rename table / add column / drop column / rename column / alter column)
- view/materialized view changing (must be VIEW/Materialized view in the query)

https://www.postgresql.org/docs/current/sql-alterforeigntable.html

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreTableColumnManager.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreTableColumnManager.java
@@ -168,7 +168,10 @@ public class PostgreTableColumnManager extends SQLTableColumnManager<PostgreTabl
     @Override
     protected void addObjectCreateActions(DBRProgressMonitor monitor, DBCExecutionContext executionContext, List<DBEPersistAction> actions, ObjectCreateCommand command, Map<String, Object> options) {
         options.put(OPTION_NON_STRUCT_CREATE_ACTION, true);
-        super.addObjectCreateActions(monitor, executionContext, actions, command, options);
+        PostgreTableBase table = command.getObject().getParentObject();
+        String sql = "ALTER " + table.getTableTypeName() + " " + DBUtils.getObjectFullName(table, DBPEvaluationContext.DDL) + " ADD " +
+            getNestedDeclaration(monitor, table, command, options);
+        actions.add(new SQLDatabasePersistAction("Create new table column", sql));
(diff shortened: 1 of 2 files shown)


ShadelessFox (approved):

serge-rider:
Does it works for views now?
Afair postgresql uses alter table for them

LonwoLonwo:
It works for views, yes.

uslss:
Foreign tables changing:
- alter table - ok
- rename table - ok
- add column - ok
- drop column - ok
- rename column - ok
**- alter column - impossible to change datatype**


Ordinary tables changing:
- alter table - ok
- rename table - ok
- add column - ok
- drop column - ok
- rename column - ok
- alter column - ok
view/materialized view changing - ok

uslss:
verified
related issue https://github.com/dbeaver/dbeaver/issues/19825

PR #19792, opened 2023-04-24
dbeaver/dbeaver#19691 add link to the wiki page about building from source

### LINKED ISSUE #19691: Provide instructions for building an aarch64 binary
I am trying to build DBeaver CE from source. 
I run all the instructions as described above, and
`mvn package` successfully runs to the end.
However, on my machine (an M1 Mac) I still only get an x86_64 build in the `product/community/target/products` folder.
When I try to run that binary, regardless, it will not start with 
```
JVM shared library does not contain the JNI_CreateJavaVM symbol
```
error message.
I suspect I am not building the right binary.
I have OpenJDK from Brew:
```
/usr/libexec/java_home -V
 20 (arm64) "Homebrew" - "OpenJDK 20" /opt/homebrew/Cellar/openjdk/20/libexec/openjdk.jdk/Contents/Home
```
My VM is set in `info.plist` as 
```
<string>-vm</string><string>/opt/homebrew/Cellar/openjdk/20/libexec/openjdk.jdk/Contents/Home/bin</string>
```
Could you please help with providing some instructions on this?

flambchap:


MashaKorax:
verified

serge-rider:
Please use md link (do not use absolute github references)

flambchap:
Markdown links won't work on local clones and forks. On top of that, we use absolute links to the wiki in other parts of the README file.

PR #19831, opened 2023-04-28
Sign copyright NOTICE in License files

FHomewood:
Copyright Notices were left at Apache 2.0 default. Not sure if this was intentional since it appears in many places.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2010 DBeaver Corp
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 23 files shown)


Destrolaric:
Greetings @FHomewood, thank you for submitting your work. Regrettably, there seem to be a few issues that require attention:
- There were instances where some third-party license footers were replaced with ours, which may not have been the intended approach. 
- Certain areas were not updated as part of the alterations.

Our team suggests that it might be best to close the submission for now, and allow our developers to address the necessary changes.

Destrolaric:
@FHomewood We have looked at how this footer was implemented in the Apache repository. It shouldn't be changed, for example, you could look at https://github.com/apache/ant/blob/master/LICENSE.

FHomewood:
Might just be my misunderstanding, I was under the impression these needed to be replaced in order to show attribution.

PR #19834, opened 2023-04-28
CB-3403 TM: store task's files in separate folders

alexander-skoblikov:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/serialize/SerializerDescriptor.java
```diff
@@ -33,22 +34,26 @@ public class SerializerDescriptor extends AbstractDescriptor
...
 
-    SerializerDescriptor(IConfigurationElement config)
-    {
+    SerializerDescriptor(IConfigurationElement config) {
         super(config);
 
         this.id = config.getAttribute("id");
         this.serializerType = new ObjectType(config.getAttribute("class"));
+        String distributedAttr = config.getAttribute("distributed");
+        this.distributed = CommonUtils.isEmpty(distributedAttr) || CommonUtils.toBoolean(distributedAttr);
```
ShadelessFox: Is it expected to be `true` when the `distributed` attribute is missing?

Anyways, you can use `CommonUtils#getBoolean` and pass the default value directly.

```suggestion
        this.distributed = CommonUtils.getBoolean(config.getAttribute("distributed"), true);
```

PR #19837, opened 2023-04-28
#18621 add Snowflake data types aliases; add numeric column modifiers…

### LINKED ISSUE #18621: Incorrect DDL for export with float column
### Description

Export from Oracle with float column into Snowflake creates incorrect DDL

### DBeaver Version

22.3.1

### Operating System

Any

### Database and driver

Snowflake
Oracle

### Steps to reproduce

1. Create an Oracle table with float column and insert some data
2. Open export for this table and select Snowflake database as a target
3. Make sure that the mapping is set as `create`
4. Start export

**Expected:** Export is succesfull
**Actual:** Table can't be created because of incorrect DDL

### Additional context

DDL for column `NUM3` should at least be `DOUBLE` or `NUMBER` with correct scale and precision.

dbeaver_sTUZxiIQSr
dbeaver_gvEosieg1p

LonwoLonwo:
… rules;

also BIGINT, INT, REAL, FLOAT, DOUBLE PRECISION, and DECIMAL were added as data types aliases. Only DECIMAL has parameters.

2023-04-29 00_01_13-Configure metadata structure

https://docs.snowflake.com/en/sql-reference/data-types-numeric

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.snowflake.core/src/org/jkiss/dbeaver/ext/snowflake/model/SnowflakeDataTypeCache.java
+++ b/plugins/org.jkiss.dbeaver.ext.snowflake.core/src/org/jkiss/dbeaver/ext/snowflake/model/SnowflakeDataTypeCache.java
@@ -101,9 +101,9 @@ public class SnowflakeDataTypeCache extends GenericDataTypeCache {
                 SnowflakeConstants.TYPE_DECIMAL,
                 false,
                 false,
-                38,
+                SnowflakeConstants.NUMERIC_MAX_PRECISION,
                 0,
-                37));
+                SnowflakeConstants.NUMERIC_MAX_PRECISION - 1));
         }
     }
 }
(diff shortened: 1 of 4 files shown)


plugins/org.jkiss.dbeaver.ext.snowflake.core/src/org/jkiss/dbeaver/ext/snowflake/model/SnowflakeDataTypeCache.java
```diff
@@ -0,0 +1,109 @@
...
+            genericDataTypes.add(new GenericDataType(
+                owner,
+                Types.DECIMAL,
+                SnowflakeConstants.TYPE_DECIMAL,
+                SnowflakeConstants.TYPE_DECIMAL,
+                false,
+                false,
+                38,
+                0,
+                37));
```
E1izabeth: Shouldn't 38 and 37 be a constants?

LonwoLonwo: They can be

E1izabeth (approved):

Matvey16:
Verified

Issue #19839, opened 2023-04-30
Refresh Resultset button


almiroliveira:
### Description

The "resultset refresh" button is closing active resultset tabs every time we click it.

### DBeaver Version

Community Edition 23.0.3

### Operating System

Windows 10

### Steps to reproduce

1. Execute a sequence of queries on some existing tables, for example:
SELECT * FROM customer;
SELECT * FROM order;
SELECT * FROM employee;
....

2. Select, for example, the first result tab;

3. Click on the "resultset refresh button F5";

Notice that the tabs close with each click.
The same thing happens when using the F5 shortcut.

Destrolaric:
Hi, Can you please provide me with some information about the database you are using?

PR #19309, opened 2023-03-09
dbeaver/pro#148 Allow hiding execution statistics

ShadelessFox:
image

Acceptance criteria:
- [x] If there are multiple results with result sets and the option is enabled, the statistics tab is shown
- [x] If there are multiple results with result sets and the option is disabled, the statistics tab is not shown
- [x] If there are no results, the statistics tab is always shown

PR #19844, opened 2023-05-01
dbeaver/dbeaver#19839 Refresh closes results tab fix

serge-rider:
Please check that dbeaver/pro#148 and #19309 are still fixed


MashaKorax:
verified

Issue #19854, opened 2023-05-02
SSH Tunnel: Incorrect hostname `localhost` is checked against the SSH host certificate


ryanc-me:
### Description

The `SSHJ` SSH tunnel implementation incorrectly checks `localhost` against the SSH server host certificate.

This is particularly important for StepCA or Teleport users, where the host certs likely contain the servers IP, and **not** `localhost`.

Note that this error only occurs when using a jump server. The `Local host`, `Remote host`, and `Bypass host verification` options don't have any impact.

### DBeaver Version

Community Edition 23.0.3.202304240616

### Operating System

Windows 11

### Database and driver

Not applicable - the error happens _before_ connecting to the database.

The issue is present for at least these two drivers:
- `mysql:mysql-connector-java` 8.0.29
- `org.postgresql:postgresql` 42.5.0

### Steps to reproduce

You will need:
1. An SSH jump server
2. A remote database server
   - Must be accessible via SSH
   - With _or_ without the jump server
4. Your `known_hosts` configured to accept hostkeys from both the database & jump servers
5. The host certificates for both servers are signed 

Then, in DBeaver:
1. Create a new connection. Any driver will work (I tested with PostgreSQL and MySQL).
2. Enable SSH Tunneling
   - Pick `SSHJ` as the implementation
   - All other advanced settings left as default
3. Ensure the `Use jump server` section is enabled & filled out
4. Click `Test tunnel configuration`

An error is thrown: `KeyExchange certificate check failed: Hostname 'localhost' doesn't match any of the principals: 'xxx.xxx.xxx.xxx'`.

Then, try:
1. Disable `Use jump server`
2. Click `Test tunnel configuration`

The SSH connection works without issues.

### Additional context

[dbeaver.log](https://github.com/dbeaver/dbeaver/files/11377346/dbeaver.log)

ryanc-me:
The PR above is a workaround. Using `Bypass host verification` will totally disable host verification, which isn't ideal.

I've done some digging:
1. When using a jump server, DBeaver actually sets up 2 consecutive SSH connections (i.e., we don't just use SSH's `-J` option)
2. First, we connect to the jump server using its actual hostname (e.g. jump.example.com)
[...]

PR #19856, opened 2023-05-03
#19854 Disable SSHJ hostname verification when "Bypass host verification" is enabled

ryanc-me:
Setup:
- An SSH server (e.g., `database-host.example.com:22`)
- The server uses a signed host cert, where `database-host.example.com` is **not present** in the certs allowed host names

To reproduce:
- Create a new database connection, using `SSHJ` as the SSH implementation
- Tick `Disable host verification`

DBeaver correctly uses `PromiscuousVerifier` to skip the host certificate signature check.

However, the hostname `database-host.example.com` will still be checked against the cert.

<br />

This commit makes it so `Disable host verification` also disables the host certificate "hostname check".

See #19854 for more context.

### DIFF
--- a/plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/SSHImplementationSshj.java
+++ b/plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/SSHImplementationSshj.java
@@ -161,6 +161,7 @@ public class SSHImplementationSshj extends SSHImplementationAbstract {
             configuration.getBooleanProperty(SSHConstants.PROP_BYPASS_HOST_VERIFICATION)
         ) {
             client.addHostKeyVerifier(new PromiscuousVerifier());
+            client.getTransport().getConfig().setVerifyHostKeyCertificates(false);
         } else {
             client.addHostKeyVerifier(new KnownHostsVerifier(SSHUtils.getKnownSshHostsFileOrDefault(), actualHostConfiguration));
         }


ShadelessFox:
I assume this PR can be closed in favor of #19857?

ryanc-me:
Sorry, this is actually a slightly different issue.

I've added a more detailed description. I can also open a separate issue if that makes things easier!

ShadelessFox:
I see. Thank you.

Yes, it would be better to create a separate ticket. Currently, two PRs that address different problems point to the same ticket.

E1izabeth:
In my understanding, `PromiscuousVerifier` allows to skip verifying cert, for example, if your cert is outdated. But depending on the environment we may see the one host under the different names. Some time it may be valid, some time it's not. So, I'm unsure about skipping verifying host against certificate. User can think of other things under this option and then be compromised due to attack.
It seems that the current name of this option is not quite correct.
I think it's better to separate it and make two different options: Skip certificate validation (rename Bypass host verification) and Skip hostname validation.

The other issue I see, is that we can't detect, if the host was compromised, as we can't know, if it was changed. But I'm unsure, whether we need to solve it somehow or rely on the user who turned on the option to skip hostname verification on his own.

If I'm not right somewhere, please correct me.

ShadelessFox (dismissed):

ShadelessFox:
Actually, it doesn't make sense.

If I understand correctly, the hostname you're connecting through SSH has a certificate that doesn't list its hostname? What kind of a certificate is that? Can you please elaborate?

It seems like a big security hole that would let MITM attacks happen.

ryanc-me:
Please note that I'm not a cryptography expert- only a (hopefully) well-informed devops person with some experience in this topic.

Some info about SSH host certificates:
- [`man ssh_config`](https://man7.org/linux/man-pages/man5/sshd_config.5.html) (search for `HostCertificate`)
- https://en.wikibooks.org/wiki/OpenSSH/Cookbook/Certificate-based_Authentication#SSH_Host_Certificates

In short, they work much like X.509/PKI certificates, but the format is slightly different. They still have an "SSH CA" who signs certs that will then be used by the servers, they still have a list of principals, validity dates, etc.

And maybe it will help if I describe the setup process:

1. Create a new SSH Certificate Authority (in my case via StepCA)
2. Create a new host certificate, specifying the principals (typically I'm using the server internal IP, or its hostname)
3. Update `sshd_config` with the `HostCertificate` and `HostKey`
4. Ensure a `@cert-authorty ...` line is present in your local `ssh_config`

As you can see, the setup is somewhat complicated. Most OpenSSH installs will not be using these certs because they require some infrastructure - they cannot be included in a "standard" install (i.e., probably most DBeaver users are not using an SSH host certificate).

**Rational:** DBeaver should allow users to bypass the host certificate check in case their setup causes the specified *SSH host* not to match the host certificate principals. For example, if `ssh.example.com` is the principal on the cert, but they are VPNing in to the host network and are instead using `10.0.0.100` as the host address. The principals *aren't strictly checked by OpenSSH anyway*, so the current DBeaver behaviour is more restrictive than regular SSH (mismatched host/principal are a warning in OpenSSH, not an error).

E1izabeth:
Thank you for detailed explanation.

ShadelessFox (approved):

PR #19858, opened 2023-05-03
#19318 Fix if-exists parsing for Oracle

### LINKED ISSUE #19318: If end if and TABLE IF EXISTS Conflict
### Description

If the database supports anonymous block execution, it also supports/ TABLE IF EXISTS test; Then there will be a problem to install the JDBC SQLDialect capability of the current dbeaver. The two are conflicting

The following script execution will report an error:
```
DECLARE
i int;
BEGIN
i := 0;
IF i < 5 THEN
i := i + 1;
DBMS_OUTPUT.PUT_LINE ('This is: '||i);
END IF;
END;

select * from YM8.TIMETEST t ;

DROP TABLE IF EXISTS test;
```

### DBeaver Version

22.1.1 ce

### Operating System

WINDOS 10

### Database and driver

YashanDB, a new database similar to ORACLE, from China

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
All of the following statements should be correctly determined. DECLARE block should end on END: and so on.
```
DECLARE
i int;
BEGIN
i := 0;
IF i < 5 THEN
i := i + 1;
DBMS_OUTPUT.PUT_LINE ('This is: '||i);
END IF;
END;

DROP TABLE
IF EXISTS test;

select * FROM
YM8.TIMETEST t ;
```

Note: Oracle don't have if exists inside drop table statement, so execution will lead to an error from the database, but the correct statement part should be send, which you can see in query manager.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleSQLDialect.java
+++ b/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleSQLDialect.java
@@ -68,7 +68,6 @@ public class OracleSQLDialect extends JDBCSQLDialect implements SQLDataTypeConve
 
     private static final String[][] ORACLE_BEGIN_END_BLOCK = new String[][]{
         {SQLConstants.BLOCK_BEGIN, SQLConstants.BLOCK_END},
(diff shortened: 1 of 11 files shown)


ShadelessFox (approved):

LonwoLonwo (approved):

MashaKorax:
verified

PR #19893, opened 2023-05-05
dbeaver/dbeaver#19677  SQL format in generate SQL

### LINKED ISSUE #19677: Add an option when generate SQL: "Format SQL"
Add a new checkbox when generate SQL, "Format SQL":

image

image

The expected behavior is the same as the "Format SQL" option on the context menu.

image

This new option will prevent us from doing some extra clicks to format the sql

Griitaz:
generateSQLwithcheckbox
generateSQLwithoutcheckingbox


LonwoLonwo:
Hello @Griitaz 

Thanks for the contribution.

First of all - you have an extra commit in your files.

And second - please look at the SQLFormatUtils.formatSQL method. Try to implement it instead of non-universal formatting.

E1izabeth:
What's the point of adding tabs? As for me, compact or non-compact formatting is more about new lines.
I'd rather apply the same formatter that we apply on format command in SQL Editor.

E1izabeth:
This PR is stale, so I'm closing it. Please, let me know if you'll continue working on this - we can reopen it.

PR #19908, opened 2023-05-08
#18861 Clarify error message

### LINKED ISSUE #18861: Geometric line with only one point in Postgresql/PostGIS gives unhelpful error
### Description

When viewing a field with a geometry linestring with only one point (which is allowed in PostGIS) the result tab shows the field with  a purple background colour and the text

> Invalid hex digit: 'L'

When changing to Spatial view a new error message pops up:

> Can't switch presentation
> Reason:
> Invalid hex digit: 'o'

The value does not show up in the value panel either.

It is true that an OGC-valid LineString has either zero or two or more points, but [PostGIS also allows single-point LineStrings](https://postgis.net/docs/using_postgis_dbmanagement.html#LineString). Also, other invalid geometries can be shown without getting an error message.

How to display the value spatially is an open question, however it should be possible just to see the value as-is, just as other geometric values are shown as text. In any case the error messages are unhelpful.

### DBeaver Version

22.3.3.202301211053

### Operating System

Windows 11

### Database and driver

Server: PostgreSQL 13.4, PostGIS 3.2.3
Driver: PostgreSQL JDBC Driver 42.2.5

### Steps to reproduce

1. Connect to PostgreSQL database with PostGIS
2. Create line with a single point, e.g.: `SELECT ST_MakeLine(ST_MakePoint(10, 20));`
3. View result: `Invalid hex digit: 'L'`
4. Click on Spatial tab and view result: `Invalid hex digit: 'o'`

### Additional context

Screenshot with error message and no value for field:
image

Screenshot of error message when clicking on Spatial tab:
image

ShadelessFox:
Unfortunately, the library enforces `LINESTRING` objects to be either empty or contain two or more points.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreGeometryValueHandler.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreGeometryValueHandler.java
@@ -221,9 +221,18 @@ public class PostgreGeometryValueHandler extends JDBCAbstractValueHandler {
             result.setSRID(srid);
 
             return new DBGeometry(result);
-        } catch (Throwable ignored) {
-            // May happen when geometry value was stored inside composite
-            return makeGeometryFromWKB(geometry);
+        } catch (Throwable e) {
+            try {
+                // May happen when geometry value was stored inside composite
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):
Looks strange, but as I understood there is no other options

LonwoLonwo (approved):

MashaKorax:
Now the error message looks clearer
dbeaver_WHtlzsRcap
Verified

PR #19923, opened 2023-05-09
CB-3402 task info api

alexander-skoblikov:


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamTransferConsumer.java
```diff
@@ -146,6 +146,7 @@ public class StreamTransferConsumer implements IDataTransferConsumer<StreamConsu
     private StatOutputStream statStream;
     
     public StreamTransferConsumer() {
+        System.out.println();
```
ShadelessFox:
```suggestion
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/serialize/SerializerDescriptor.java
```diff
@@ -41,7 +41,7 @@ public class SerializerDescriptor extends AbstractDescriptor
 
         this.id = config.getAttribute("id");
         this.serializerType = new ObjectType(config.getAttribute("class"));
-        this.distributed = CommonUtils.getBoolean(config.getAttribute("distributed"), true);
+        this.distributed = CommonUtils.getBoolean(config.getAttribute("supportsDistributedMode"), true);
```
ShadelessFox, line 44: Please use `RegistryConstants#ATTR_SUPPORTS_DISTRIBUTED_MODE`

alexander-skoblikov: there is no dependence on `org.jkiss.dbeaver.registry`

ShadelessFox: Oh. Okay

ShadelessFox (approved):

PR #19932, opened 2023-05-10
#19780 Fix Bind Paramater(s) dialog information status, add new info there

### LINKED ISSUE #19780: I found two problems while upgrading to Dbeaver 23.0.3.
### Description

1. When using bind parameters, the input value window pops up every time it is executed.
   However, if the @set command is used, the bind variable input window does not appear.
2. In the previous version, the case of the bind variable was distinguished, but in this version, it is changed to upper case unconditionally.
이미지 003

I am always grateful for developing good programs. :)

### DBeaver Version

Community Edition 23.0.3

### Operating System

Windows 10

### Database and driver

Oracle

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
image


LonwoLonwo (approved):

MashaKorax:
verified

PR #19936, opened 2023-05-10
dbeaver/dbeaver#19904 Show original variables/parameters names

### LINKED ISSUE #19904: Displaying variables case
### Description

At the moment, the case-insensitive parameters are shown in user-interface in upper-case, because the case does not matter. If user want to make parameter case-sensitive, it should be put into quotes.
We don't want to show case-insensitive parameters in upper-case. We need to add information about parameter's case-sensivity for that through our model.

### DBeaver Version

DBeaver 23.0.4

serge-rider:


E1izabeth:
How to distinguish variable set as case-insensitive with `@set AbC = 1`, which turns into `ABC` under the hood, and case-sensitive variable set with `@set "AbC" = 1`?

MashaKorax:
Now  case-sensitive behavior works with  bind parameters, but with variables no: 
iYg1TnRwKy

E1izabeth:
@MashaKorax to make variable or parameter case-sensitive it should be quoted

serge-rider:
@E1izabeth variables/params resolve behavior didn't change. The only change in this ticket is that we show original variable/parameter name in binding dialog

E1izabeth:
@serge-rider yes, I see, that's why I'm asking of how to distinguish visually variables in those two cases now.

serge-rider:
It will show the name from the first occurrence (AbC).
It is better than ABC I think.

E1izabeth:
But "AbC" and AbC are two different variables 🤔

MashaKorax:
verified

PR #19942, opened 2023-05-11
dbeaver/pro#1560 support Greenplum 7 version

LonwoLonwo:
https://github.com/greenplum-db/gpdb/commit/b62e06014d1237ab9b3657e30609d2630954134d about external tables - they are foreign tables partially now (probably, we need new ticket about external tables)

https://github.com/greenplum-db/gpdb/commit/f182228992b62e2023e2fac5b4971406abd35c9d - about partitions - in Greenplum 7, we can read partitions like for ordinary PG in our code

Also, please check the DDLs of regular tables - DISTRIBUTED ON case must be inside the DDL
Reding of all main metadata entities also check.


MashaKorax:
verified

PR #19953, opened 2023-05-12
#17585 ability to trim string values during export to XLSX implemente…

### LINKED ISSUE #17585: There are blank space when export the query result into spreadsheet
Hi, each time I export the query result into spreadsheet, there are a few space behind the value. I suppose it is to make up the maximum field due to the dbeaver default setting? And it should not be the dababase issue, because if I export the query result into spreadsheet from QMF, there is no space behind the value. 

My current solution is to use excel formula "=trim" to remove the space, but it is very unproductive...

Appreciate anyone can kindly advise how do I fix the issue. thank you.

image

LonwoLonwo:
ability to trim string values during export to XLSX implemented with an extra setting

Please check export to xlsx with different settings

2023-05-12 14_15_47-Data Transfer


E1izabeth (approved):

ShadelessFox (approved):

MashaKorax:
verified

PR #19957, opened 2023-05-12
#19952 fix table cache reading query

### LINKED ISSUE #19952: [PostgreSQL] Doubling tables in database navigator if there are multiple generated columns
### Description

Greetings

Just found weird visual bug in database navigator - it shows N copies of the table that has N generated columns (GENERATED ALWAYS / BY DEFAULT AS IDENTITY / *expression* STORED

These copies have the same OID and doesn't appear in pg_class, so it's only UI problem.

Minimal reproducible example is below. Tested it on latest DBeaver CE on Windows and Ubuntu, same results.
As far as i can tell, it didin't generates table copies like that in older versions.

Here are some screens from that example

image

image

image

image

### DBeaver Version

Community Edition 23.0.4

### Operating System

Windows 10 Pro 21H2 / Ubuntu 22.04

### Database and driver

PostgreSQL 15.3
PostgreSQL JDBC 42.6.0

### Steps to reproduce

```
create schema test_schema;

create table test_schema.test_table 
(
	id int generated always as identity,
	gen int generated always as (5 * id) stored
);
```

### Additional context

_No response_

LonwoLonwo:
classid must be the same as refclassid (and it is pg_class)

Please check PG tables reading in different PG versions


E1izabeth (approved):
I'd unify the case of keywords.

ShadelessFox (approved):

MashaKorax:
verified

Issue #17919, opened 2022-10-13
Shortcut for Adding Column/Field


iambudi:
Just like shortcut for adding row, it would be nice to have shortcut for adding column `(cmd +)`.

![image](https://user-images.githubusercontent.com/31809529/195727042-990c3a6c-61b6-467f-82bb-a62958badda6.png)

flambchap:
Thank you for the idea.

arhayka:
We want to implement this feature as follows:
1. Use the  [used in Data Editor ](https://github.com/dbeaver/dbeaver/wiki/Shortcuts#result-set)(`Alt+Insert` — if it works correctly on macOS).
2. When multiple objects can be added, this shortcut should open a list with available options.

flambchap:
@arhayka There is no insert button on most keyboards for Mac, and, AFAIK, no universal substitution combination.

PR #19963, opened 2023-05-15
#17919 add shortcut for create new database object

SiyaoIsHiding:
Implementing the feature requested in #17919 

There is a challenge I met:
The commands to create new database objects, including a new column and a new index, are grouped in one command `org.jkiss.dbeaver.core.object.create`. Therefore, it is hard to create a shortcut for a new column only, unless we split the commands or do not use the RCP built-in bindings extension. But I think a shortcut for other new database objects, including a new index, is also a good idea.


Matvey16:
Verified

arhayka:
Thank you for your contribution! There are some notes:
1. This `Cmd+ | Ctrl+` shortcut conflicts with the standard Windows that enables zooming.  It's better to use the one [used in Data Editor ](https://github.com/dbeaver/dbeaver/wiki/Shortcuts#result-set)(`Alt+Insert`).
2. This shortcut does nothing when you can add multiple objects. It should open a list with available options.

E1izabeth:
Would you correct your PR according to the [mentioned issues](https://github.com/dbeaver/dbeaver/pull/19963#issuecomment-1611215677)?

SiyaoIsHiding:
Sorry that we no longer work on this project :)

PR #19967, opened 2023-05-15
dbeaver/dbeaver#19692 remove limit and increase increment

### LINKED ISSUE #19692: Add option to increase external formatting timeout
I am using an external formatter, set up in 
`Preferences -> Editors -> SQL Editor -> Formatting`

In the formatter, I just send a curl request to a formatter server and return the response.

For formatting simple queries it works as expected.
However, for bigger queries, my external formatter could take a minute or two just to format the query.

This is acceptable for my use-case.

However, currently I can't use DBeaver to do this, because it has a hard cap at `10_000ms` for `Exec timeout`.
10 seconds is far too short for my use-case.

I know a place in the code where this hard cap is set, but increasing it to a different hard cap would just move the problem down the line.
Maybe we should allow users to set the timeout to an arbitrary value?

Destrolaric:
Closes dbeaver/dbeaver#19692


E1izabeth (approved):

MashaKorax:
verified

PR #19988, opened 2023-05-17
#19941 Add table and alias inspection according to parse tree

### LINKED ISSUE #19941: Rewrite completion analyzer using antlr parser
The proposals for where clause should be taken according to the parse result of antlr parser.
Also it would be nice to do the same for proposals in join clause.

E1izabeth:
Now aliases comes to autocomplete along with the aliased table.
image

Please check autocompletion cases in where and select columns.

The previous behavior could be returned back by unchecking this preference.
image


MashaKorax:
verified

PR #19996, opened 2023-05-17
#16875 Reveal obscured cursor after control resize

### LINKED ISSUE #16875: Result View hides cursor position
#### System information: 
- Operating system (distribution) and version: Debian/SID
- DBeaver version: Version 22.1.0.202206051253
- Additional extensions

- Database name and version: PostgreSQL 12
- Driver name: 

When Result View opens at the bottom of the editor, cursor position is hidden.
Seems, before resizing the editor it's already scrolled up, thus moving cursor position out of view.

Given that the actual cursor position is at the currently executed query, the automatic scrolling seems to be unnecessary anyway.

- Editor with content filling at least until the last line
- no Result View open
- add query at the last line
- execute query

ShadelessFox:


E1izabeth (approved):

MashaKorax:
verified

PR #20020, opened 2023-05-19
#20019 Add logging listeners, fix potential NPE

### LINKED ISSUE #20019: Enhance parser error logging
Currently ANTLR writes errors in to console err stream. We don't see it while running application anyway, so let's don't do it.

E1izabeth:
We may need to log error somewhere, but in completion analyzer we don't. So, let's skip errors and do not write them to somewhere. The logging error listener might be useful when we want to write something in our debug log.

**How to test it: logs shouldn't contain any messages like "no viable alternative at".**


MashaKorax:
verified

Issue #19652, opened 2023-04-10
Write tests for parser and model mapping


E1izabeth:
We need to cover parsing and model mapping at least partially.
Maybe we could use benefits of serialization as our model classes are serializable now.

PR #20028, opened 2023-05-20
#19652 Add lsm parser test about comment and GROUP

SiyaoIsHiding:
Anyway, it would be good to cover the comments following `#` and the `GROUP` keyword in the parser test.

### DIFF
--- a/test/org.jkiss.dbeaver.model.lsm.test/src/org/jkiss/dbeaver/model/lsm/test/SelectStatements.sql.txt
+++ b/test/org.jkiss.dbeaver.model.lsm.test/src/org/jkiss/dbeaver/model/lsm/test/SelectStatements.sql.txt
@@ -101,10 +101,20 @@ SELECT ALL
     ProductSubCategory.Name AS ProductSubCategory,
     Product.ProductModelID
 -- INTO Product
+# ANOTHER INFO Product
 FROM Production.Product
 INNER JOIN Production.ProductSubCategory
 ON ProductSubCategory.ProductSubcategoryID = Product.ProductSubcategoryID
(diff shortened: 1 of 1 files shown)


E1izabeth:
Thank you for contribution.
You may find that comments are replaced with the new lines in the current test. To make your changes meaningful, you need to at least modify that part of test.
Also currently ParseSelectStmtTest contains only checks that there is no errors and no checks that the parse tree or model is correct. So, it doesn't make sense to add new cases to the text file with test sql queries, while test doesn't check it.

SiyaoIsHiding:
I deleted the link to the issue. I think it is still meaningful because It does not execute the SQL queries but parses the SQL queries, which is exactly what we should do in the lsm parser test.

1. You substitute the comments after `--` or `#` to a new line. In the `SelectStatements.sql.txt` file, you add the `-- INTO Product`. For the same reason, I add `# Another INFO Product`.
2. The `SELECT` with `GROUP` statement will test the `plugins/org.jkiss.dbeaver.model.lsm/src/org/jkiss/dbeaver/model/lsm/sql/impl/GroupingSpec.java`, which is not covered  yet.

E1izabeth:
I'm happy to help you, if you want to work on it, but I can't accept it in the current state. We need to handle situation with comments at least. Current grammar is written by SQL-92 Standard and doesn't include `#` for comments. You could modify the grammar to handle this case. It's in a Sql92Lexer.g4 file, LineComment rule, line 387. Please ask me, if you have any questions of how to do it.

E1izabeth:
It's been a while since no update here, so I'm closing this PR. Please, let me know if you make corrections according to the comments - we can reopen it or you could create a new one.

PR #20044, opened 2023-05-22
dbeaver/pro#1634 Keep original driver when replacing the old one

ShadelessFox:
STR:
1. Boot up CE, edit a connection
2. Boot up EE, edit the connection
3. Boot up CE, observe that the connection uses the original driver, edit the connection
4. Boot up EE, observe that the connection uses the replacement driver


MashaKorax:
verified

PR #20074, opened 2023-05-24
dbeaver/pro#1545 Show tips in applicable products only

serge-rider:


ShadelessFox (commented):
Please ensure that localized tips under the `nl` directory are updated as well.

Matvey16:
Verified

PR #20079, opened 2023-05-24
#20073 Add support of different identifier quotes to the grammar

### LINKED ISSUE #20073: Autocomplete of tables with quotes
### Description

Now we support only quotes `"`, we need to support also `[ ]`  and ` as quotes for identifiers.

For example, if we autocomplete propose nor alias, nor table name in this case.
image

but should

image

### DBeaver Version

DBeaver 23.0.5

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Square brackets in identifiers are in MS SQL Server
image

Back quotes can be tested in Clickhouse or MariaDB
image

Double quotes are in PostgreSQL
image

Please, note, that if alias contains only 1 letter, in most cases no proposal appear as it's supposed that you've already wrote full alias. It's an old behavior.


ShadelessFox (approved):

E1izabeth:
I haven't found database with single quotes in identifiers, so I removed support of single quotes in identifier names.

MashaKorax:
verified

PR #20082, opened 2023-05-24
CB-3200 rm events resource type folder fix

yagudin10:


plugins/org.jkiss.dbeaver.model.event/src/org/jkiss/dbeaver/model/websocket/event/WSEvent.java
```diff
@@ -33,4 +33,8 @@ public interface WSEvent {
     String getUserId();
 
     long getTimestamp();
+
+    boolean isSendForce();
+
+    void setSendForce(boolean sendForce);
```
alexander-skoblikov, line 39: remove mutable method, add new constructor instead

plugins/org.jkiss.dbeaver.model.event/src/org/jkiss/dbeaver/model/websocket/event/WSAbstractEvent.java
```diff
@@ -72,4 +74,16 @@ public String getUserId() {
     public long getTimestamp() {
         return timestamp;
     }
+
+    @Override
+    public boolean isSendForce() {
```
alexander-skoblikov, line 79:
```suggestion
    public boolean isForceSend() {
```

PR #20098, opened 2023-05-25
dbeaver/pro#1652 Cancel detached jobs during editor save

serge-rider:


uslss:
For now, the query could be canceled, but new table still cannot be saved (the progress hangs on 0% and after the ten minutes of awaiting nothing happens)

serge-rider:
It is not possible to create a table in this Netezza server.

The problem we solve in this ticket is UI freeze.

uslss:
verified

PR #20109, opened 2023-05-26
#20108 support partitions for MariaDB; do not show default empty part…

### LINKED ISSUE #20108: [MySQL/MariaDB] do not show empty default partitions
According to the [documentation](https://dev.mysql.com/doc/refman/5.7/en/information-schema-partitions-table.html), any table without created partitions has a default partition:

> A table using any storage engine other than [NDB](https://dev.mysql.com/doc/refman/5.7/en/mysql-cluster.html) and which is not partitioned has one row in the [PARTITIONS](https://dev.mysql.com/doc/refman/5.7/en/information-schema-partitions-table.html) table. However, the values of the PARTITION_NAME, SUBPARTITION_NAME, PARTITION_ORDINAL_POSITION, SUBPARTITION_ORDINAL_POSITION, PARTITION_METHOD, SUBPARTITION_METHOD, PARTITION_EXPRESSION, SUBPARTITION_EXPRESSION, and PARTITION_DESCRIPTION columns are all NULL. Also, the PARTITION_COMMENT column in this case is blank.

2023-05-26 12_22_52-DBeaver 23 0 3 - account

But it is confusing - to have empty partitions in all unpartitioned tables. Let's hide them.

LonwoLonwo:
support partitions for MariaDB; 
do not show the default empty partition for all unpartitioned tables

Please check MariaDB and MySQL tables with partitions and without them.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
@@ -71,7 +71,7 @@
                                         </items>
                                     </items>
                                 </folder>
-                                <folder type="org.jkiss.dbeaver.ext.mysql.model.MySQLPartition" label="%tree.partitions.node.name" icon="#partitions" description="Table partitions" visibleIf="object.dataSource.info.databaseProductVersion>='5.1.6'">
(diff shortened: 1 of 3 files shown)


ShadelessFox (approved):

E1izabeth (approved):

uslss:
verified

PR #20118, opened 2023-05-29
#20108 remove "create partitions" button for MySQL/MariaDB

### LINKED ISSUE #20108: [MySQL/MariaDB] do not show empty default partitions

LonwoLonwo:
Please check different places:

- [ ] In the table properties in the Partitions tab
- [ ] On the table in the Database navigator tree


E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
```diff
@@ -71,9 +71,9 @@
                                         </items>
                                     </items>
                                 </folder>
-                                <folder type="org.jkiss.dbeaver.ext.mysql.model.MySQLPartition" label="%tree.partitions.node.name" icon="#partitions" description="Table partitions" visibleIf="object.dataSource.supportsPartitions()">
+                                <folder type="org.jkiss.dbeaver.ext.mysql.model.MySQLPartition" label="%tree.partitions.node.name" icon="#partitions" description="Table partitions" visibleIf="object.dataSource.supportsPartitions()" virtual="true">
```
serge-rider: Virtual nodes are supposed to be used for duplicated objects in a tree. How this helps in this case?

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/edit/MySQLPartitionTableManager.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.ext.mysql.edit;
+
+public class MySQLPartitionTableManager extends MySQLTableManager {
+
+    @Override
+    public boolean canCreateObject(Object container) {
+        // Creating partitions separately is not supported in MySQL/MariaDB
```
serge-rider, line 25: I altho thought that we want to hide dummy partitioj in MySQL tables. Is it doable?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/edit/MySQLPartitionTableManager.java
LonwoLonwo: Implemented.

uslss:
verified

serge-rider (approved):

PR #20119, opened 2023-05-29
dbeaver/pro#1636 do not use default schema/catalog from the script in…

LonwoLonwo:
… case of data editor changing

Please check:

- [x] - script opening after DBeaver reopening (script must be the same). Use default for connection and non-default schemas/databases
- [x] - case from the ticket

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
@@ -235,6 +235,7 @@ public class SQLEditor extends SQLEditorBase implements
     private ScriptAutoSaveJob scriptAutoSavejob;
     private boolean isResultSetAutoFocusEnabled = true;
     private Boolean isDisableFetchResultSet = null;
+    private boolean datasourceChanged;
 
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -690,10 +693,13 @@ private class OpenContextJob extends AbstractJob {
         private final DBSInstance instance;
         private final Runnable onSuccess;
         private Throwable error;
-        OpenContextJob(DBSInstance instance, Runnable onSuccess) {
+        private boolean readFromInstance;
+
+        OpenContextJob(DBSInstance instance, Runnable onSuccess, boolean readFromInstance) {
```
E1izabeth: As for me, it's not clear what is this argument about by the name `readFromInstance`, like read what? I'd call it `readContextDefaultsFromInstance` or `readDefaultsFromInstance`. Yes, it's longer, but clearer.

LonwoLonwo: ok

E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #20151, opened 2023-05-31
#20150 Refactor Q-string parsing rule for Oracle

### LINKED ISSUE #20150: Oracle QString refactoring
### Discussed in https://github.com/orgs/dbeaver/discussions/19898

<sup>Originally posted by **Marcono1234** May  7, 2023</sup>
### Description

`OracleDialectRules.QStringRule.doEvaluate` contains the following check:
https://github.com/dbeaver/dbeaver/blob/0642e8c736bb919041add49621b0cc4a819838bb/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/sql/OracleDialectRules.java#L55

The problem is that `quoteStartChar` is a `char` and the minimum value of `char` is 0, so the `!= -1` check will always be true.

Maybe the simplest solution is to change `quoteStartChar` to type `int` and remove the `char` cast.

I am not planning to submit a pull request though because I am not familiar with this project.

### DBeaver Version

commit 0642e8c736bb919041add49621b0cc4a819838bb

</div>

E1izabeth:
Oracle q-strings should work. Example: q'[What's a quote among friends?]'


LonwoLonwo (approved):

ShadelessFox (approved):

MashaKorax:
verified

PR #20152, opened 2023-05-31
dbeaver/pro#1502 new connection wizard deprecation

ShadelessFox:
Acceptance criteria:
- [x] A deprecated driver is shown in the new connection wizard
- [x] A deprecated driver can't be used to create a new connection
- [x] An existing connection using a deprecated driver can't be edited
- [x] An existing connection using a deprecated driver can't be used


E1izabeth (approved):

uslss:
verified

PR #20153, opened 2023-05-31
#17882 Add retry with getLocationURI for copy file path command

### LINKED ISSUE #17882: Copy File Path missing on some SQL scripts
### Description

Some open SQL scripts no longer offer the "Copy File Path" option on the tab right-click.  I first saw this in 22.2.1 and it persists in 22.2.2.  
image

I thought at first this was related to not having a assigned connection, but that was not the case.  ALso tried shortening the file path (by default, it's a MS Onedrive with a horrendously long root path), but that is not the issue.  File name length does not appear to be a factor, got same error on "C:\temp\today\t.sql" as the source file name.  

It does seem to still work for files that I originate as a script that are saved inside the dBeaver project:

image

Some other options are missing, as well.  

### DBeaver Version

22.2.2

### Operating System

Windows 10 Enterprise 2009 10.0.19041.1806

### Database and driver

1. Database:  mySQL
2. Driver:  ID: mysql8  Driver for MySQL 8 and later

### Steps to reproduce

1. Open a mySQL or Postgres script in dBeaver SQL Editor
2. Access editor tab with right-click
3. Look for "Copy FIle Path" (with intention of including full path as a comment in the file, usually). Expectation:
image

4.  Path is missing, along with other options. Reality:
image

### Additional context

_No response_

E1izabeth:
Please, check, that the command Copy File Path still appears and works correct.

Please, check that nothing changed for TE.


LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/e4/DBeaverStackRenderer.java
```diff
@@ -172,16 +172,26 @@ public void widgetSelected(SelectionEvent e) {
...
+                });
+            } else if (inputFile.getLocationURI() != null) {
+                menuItemOthers.addSelectionListener(new SelectionAdapter() {
+                    @Override
+                    public void widgetSelected(SelectionEvent e) {
+                        String filePath = inputFile.getLocationURI().getPath();
+                        UIUtils.setClipboardContents(Display.getCurrent(), TextTransfer.getInstance(), filePath);
+                    }
+                });
+            }
```
ShadelessFox: Why is this check not in the listener itself? It looks ugly and repetitive.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/e4/DBeaverStackRenderer.java
E1izabeth: OK

ShadelessFox (approved):

uslss:
Copy file path doesn't appear in the next case:
1. Open SQL Console and associate it with the connection
2. perform Save as action - path is shown in the tab tooltip, but action is still unavailable in the context menu. Even after the script reopened.

E1izabeth:
Please, check that nothing changed for TE.

uslss:
verified

PR #20157, opened 2023-05-31
#19886 Make Customize Toolbar be a property page

### LINKED ISSUE #19886: Add option to Customize Toolbar for SQL and Data editor preferences
For now we can customize SQL/Data Editor toolbars from the Main Preferences only 

It will be great if Customize toolbars options will be available from "local" preferences for SQL and Data editors
image

E1izabeth:
Please, check that this preferences still can be changed, saved and restored to default.

image

image


ShadelessFox (approved):

LonwoLonwo (approved):

MashaKorax:
verified

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageToolbarCustomization.java
```diff
@@ -53,8 +53,10 @@
 import java.util.stream.Collectors;
 
 
-public class PrefPageToolbarCustomization extends AbstractPrefPage implements IWorkbenchPreferencePage {
-    
+public class PrefPageToolbarCustomization extends TargetPrefPage {
```
serge-rider: Why TargetPrefPage? It is designed for connection-specific properties. AbstractPrefPage maybe better.

plugins/org.jkiss.dbeaver.core/plugin.xml
```diff
@@ -1303,12 +1308,4 @@
                 group="%confirmation.group.general.actions"/>
     </extension>
 
-    <extension point="org.eclipse.ui.preferencePages">
```
serge-rider: So it will disappear from global preferences?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.core/plugin.xml
E1izabeth: No, it's still exists there

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageToolbarCustomization.java
E1izabeth: I used AbstractPrefPage before, it doesn't allow to bring this page to property settings

MashaKorax:
verified

serge-rider (approved):

PR #20158, opened 2023-05-31
dbeaver/pro#1659 Fix time for disconnect from database record in QM

E1izabeth:
The time for `Disconnected from ...` record in QM should not be equal to the time for `Connected to ...`


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/controls/querylog/QueryLogViewer.java
```diff
@@ -142,6 +142,15 @@ private static class ColumnDescriptor {
...
         String getText(QMEvent event, boolean briefInfo) {
+            if (event.getObject() instanceof QMMConnectionInfo) {
+                switch (event.getAction()) {
+                    case END:
+                        return timeFormat.format(event.getObject().getCloseTime());
+                    case BEGIN:
+                    default:
+                        timeFormat.format(event.getObject().getOpenTime());
+                }
+            }
```
ShadelessFox: I don't think this `switch` statement is required here.

Moreover, the expression in the `default` branch is never returned but works because the last return produces the same result.

Lastly, I'm not sure if the check for `QMMConnectionInfo` is needed as all QM objects have "open" and "close" times.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/controls/querylog/QueryLogViewer.java
E1izabeth: We do need to check for QMMConnectionInfo  as we want to do it only for disconnect from database events.

ShadelessFox: Indeed, but what about other events?

ShadelessFox (approved):

MashaKorax:
verified/ for cloudbeaver https://dbeaver.atlassian.net/browse/CB-3223

PR #20162, opened 2023-05-31
dbeaver/pro#1647 Deep copy task configuration when copying task

ShadelessFox:


E1izabeth (approved):

bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
```diff
@@ -228,6 +231,47 @@ public static <T> List<T> copyList(@Nullable Collection<T> theList) {
         }
     }
 
+    @NotNull
+    public static <T> List<T> deepCopy(@NotNull List<T> src) {
```
LonwoLonwo: I think a "Missing a Javadoc comment." has some meaning here.

ShadelessFox: I believe the method name is already self-explanatory.

LonwoLonwo (approved):

MashaKorax:
verified

bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
```diff
@@ -228,6 +231,47 @@ public static <T> List<T> copyList(@Nullable Collection<T> theList) {
         }
     }
 
+    @NotNull
```
serge-rider: It is better to move these functions to BeanUtils

bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
```diff
@@ -228,6 +231,47 @@ public static <T> List<T> copyList(@Nullable Collection<T> theList) {
...
+        final List<T> dst = tryInstantiateOrDefault(src.getClass(), ArrayList::new);
+        for (T element : src) {
+            dst.add(deepCopy(element));
+        }
+        return dst;
+    }
+
+    @NotNull
+    public static <K, V> Map<K, V> deepCopy(@NotNull Map<K, V> src) {
+        final Map<K, V> dst = tryInstantiateOrDefault(src.getClass(), HashMap::new);
```
serge-rider: I don't like HashMap here, even as default. LinkedHashMap<> is better.

serge-rider (changes requested):

MashaKorax:
verified

serge-rider (approved):

Issue #20070, opened 2023-05-24
support TDengine as a stand-alone database in driver list


sangshuduo:
**Is your feature request related to a problem? Please describe.**

TDengine is an open-source and cloud-native time series database. It has gained 21k+ stars on GitHub and 255k+ installations from over 50 countries. TDengine supports SQL standards and provides a generic JDBC driver to access. DBeaver can be configured with a generic driver to access TDengine and works very well.
But the user needs to configure the driver strings manually now. I would like to ask the DBeaver team to add TDengine as a stand-alone database with default configuration strings to make the user experience better and simpler.

**Describe the solution you'd like**

To list TDengine in a predefined time-series database list with pre-configured connection strings. Note TDengine supports both native and REST connections. The native connection uses a 6030 port and requires the user to install the client library. The REST connection uses a 6041 port and does not need the user to install the library.

_I would suggest DBeaver supports both of them. One can be TDengine and another one can be TDengine-REST._

REVISED: I would suggest DBeaver supports REST interface only to reduce the burden user needs to install the client library. 

**Describe alternatives you've considered**

Still, use a generic JDBC driver but not good for the user.

**Additional context**

TDengine can be installed by the command 'docker run -d -p 6030:6030 -p 6041:6041 tdengine/tdengine:latest'. 
I am the contact person of the TDengine team. I will do my best to assist if needed.

Thanks,

E1izabeth:
Thank you for the idea.

sangshuduo:
Sorry, I edit the original text to revise the requirement to only support the REST interface.

The steps to verify DBeaver connecting TDengine with generic JDBC are following:

1. run TDengine Docker instance with the command 'docker run -d -p 6030:6030 -p 6041:6041 tdengine/tdengine:latest'
2. Download the latest taos-jdbcdriver jar file from https://repo1.maven.org/maven2/com/taosdata/jdbc/taos-jdbcdriver/3.2.1/taos-jdbcdriver-3.2.1-dist.jar
3. launch DBeaver and new an item from the menu Database -> Driver Manager.
4. name it 'TDengine' and add taos-jdbcdriver-xxx-dist.jar in Libraries -> Add File. Fill in the class name with 'com.taosdata.jdbc.rs.RestfulDriver'. Fill in the URL Template with 'jdbc:TAOS-RS://127.0.0.1:6041'. Fill in the Port with 6041. Fill in the Default User with 'root'
[...]

PR #20173, opened 2023-06-01
add TDengine database driver support

sangshuduo:
verified with a local TDengine instance and a TDengine Cloud instance
image


sangshuduo:
https://github.com/dbeaver/dbeaver/issues/20070

LonwoLonwo:
You need to add small icons, also.

sangshuduo:
> You need to add small icons, also.

I added a small icon. Now I can see the driver display with the icon thumbnail. However when I create a database connection, a big icon still appears on the head area with a blank rectangle. How can I fix it? 

Another question, I want to give the user only the option to input the URL for the TDengien Cloud driver since that will access the given URL only. Now it provides Host and URL options. I wish don't provide a host option to avoid confusing the user.

LonwoLonwo:
About the second question.
Removing the "`sampleURL="jdbc:TAOS-RS://{host}?useSSL=true%26token={token}`" may help.

It will be as Azure Databricks connection page:

2023-06-02 10_26_13-Window

LonwoLonwo:
About icons.
You can take a look at other plugins.

https://github.com/dbeaver/dbeaver/tree/devel/plugins/org.jkiss.dbeaver.ext.oracle/icons

Usually, we have:
- `database_name_icon.png` - 16*16 px
- `database_name_icon@2x.png` - 32*32 px
- `database_name_icon_big.png` - 64*64 px
- `database_name_icon_big@2x.png` - 128*128 px

sangshuduo:
@LonwoLonwo  Thanks. I will fix it and resubmit. Thanks.

PR #20175, opened 2023-06-01
#11016 enhance working with table partitions and subpartitions for Oracle

### LINKED ISSUE #11016: Partition picklist in table Oracle
Is it possible to add in the detail of the table the possibility to see only the data of a specific partition? Example :
Detail Table - select partition 20201231 and show only the data of that partition. A picklist of partitions ...

LonwoLonwo:
2023-06-01 14_35_10-Window

2023-06-01 14_35_45-Window

- Partitions folders are now not visible if the table does not have partitions
- Partitions and subpartitions with the correct icon in the Navigator tree
- Subpartitions now expandable
- Partitions and subpartitions are now like tables - with a Data tab. And data can be read (special SELECT used)

Knowing issue: ERD tab is not working for partition, as in MySQL. And virtual tab also needs to be removed. You can create a separate ticket for it.


plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleTablePhysical.java
```diff
@@ -203,57 +202,44 @@ public void refreshObjectState(@NotNull DBRProgressMonitor monitor) throws DBCEx
...
+            @NotNull OracleTablePhysical table,
+            @Nullable OraclePartition partition,
+            @Nullable String partitionName
+        ) throws SQLException {
             final JDBCPreparedStatement dbStat = session.prepareStatement(
-                "SELECT * FROM "+ OracleUtils.getSysSchemaPrefix(table.getDataSource()) + "ALL_TAB_SUBPARTITIONS " +
-                "WHERE TABLE_OWNER=? AND TABLE_NAME=? " +
-                (forObject == null ? "" : "AND PARTITION_NAME=? ") +
-                "ORDER BY SUBPARTITION_POSITION");
+                "SELECT * FROM " + OracleUtils.getSysSchemaPrefix(table.getDataSource()) + "ALL_TAB_PARTITIONS " +
```
E1izabeth, line 219: Why we stopped reading from `ALL_TAB_SUBPARTITIONS` and just replace it with `ALL_TAB_PARTITIONS`? Shouldn't we read from both or do something special for `ALL_TAB_SUBPARTITIONS`?

LonwoLonwo: This logic moved to OraclePartition#getSubpartirions.
It was not handy as this type of cache. Many strange flags to read and support it.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleTablePhysical.java
```diff
@@ -77,6 +79,14 @@ protected OracleTablePhysical(
         this.partitionCache = partitioned ? new PartitionCache() : null;
     }
 
+    protected OracleTablePhysical(@NotNull OracleSchema schema, @NotNull ResultSet dbResult, @NotNull String name) {
+        // Table partition
+        super(schema, name);
+        this.rowCount = JDBCUtils.safeGetLong(dbResult, "NUM_ROWS");
```
serge-rider, line 88: Please avoid copy-paste. Move NUM_ROWS and TABLESPACE_NAME read to a separate method

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OraclePartition.java
```diff
@@ -0,0 +1,213 @@
...
+import java.sql.Timestamp;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Oracle abstract partition
+ */
+public class OraclePartition extends OracleTable implements DBSTablePartition {
```
serge-rider: Why rename TablePartition->Partition?
It is still a table partition I suppose.

serge-rider: Also why Partiton now extends Table? Maybe |it should extend TablePhysical at least (to avoid lots of visibleIf annotations)

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleTablePhysical.java
LonwoLonwo: implemented

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OraclePartition.java
LonwoLonwo: renamed. inheritance changed.

uslss:
verified

serge-rider (approved):

PR #20189, opened 2023-06-02
dbeaver/dbeaver#20187 add quotation rules for snowflake identifiers

### LINKED ISSUE #20187: View Preview does not double-quote when convenient (ex : Snowflake + greek letter)
### Description

Tested in Snowflake - As a test, let's create a VIEW (SELECT 1), in a schema named with a greek letter, like that : 

`CREATE SCHEMA "ΣTEST";
CREATE VIEW "ΣTEST".VTEST AS SELECT 1 AS DUMMY;`

whereas the call is OK in DBeaver if done properly,
`SELECT * FROM "ΣTEST".VTEST `

...the default preview shows an error because the script generated for the preview does not double-quote (which is here mandatory to handle special characters)
`SELECT * FROM <my_db>.ΣTEST.VTEST
LIMIT 200`

image

Is this something to be fixed or is there a feature in the settings to force double-quoting DB/Schema/Object name ?

### DBeaver Version

23.0.4.202305071722

### Operating System

Win11

### Database and driver

Snowflake

### Steps to reproduce

Tested in Snowflake - As a test, let's create a VIEW (SELECT 1), in a schema named with a greek letter, like that : 

`CREATE SCHEMA "ΣTEST";
CREATE VIEW "ΣTEST".VTEST AS SELECT 1 AS DUMMY;`

whereas the call is OK in DBeaver if done properly,
`SELECT * FROM "ΣTEST".VTEST `

...the default preview shows an error because the script generated for the preview does not double-quote (which is here mandatory to handle special characters)
`SELECT * FROM <my_db>.ΣTEST.VTEST
LIMIT 200`

image

Is this something to be fixed or is there a feature in the settings to force double-quoting DB/Schema/Object name ?

### Additional context

_No response_

flambchap:


E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.snowflake.core/src/org/jkiss/dbeaver/ext/snowflake/model/SnowflakeSQLDialect.java
```diff
@@ -142,4 +142,32 @@ public String getColumnTypeModifiers(
...
+        // https://docs.snowflake.com/en/sql-reference/identifiers-syntax
+
+        if (forceCaseSensitive || str.isBlank()) {
+            return true;
+        }
+        int firstCodePoint = str.codePointAt(0);
+        if (!isLatinLetter(firstCodePoint) && firstCodePoint != '_') {
+            return true;
+        }
+        for (int i = 1; i < str.codePointCount(0, str.length()); i++) {
```
serge-rider: Please variable for `str.codePointCount(0, str.length())` otherwise it calculates it for every loop.

plugins/org.jkiss.dbeaver.ext.snowflake.core/src/org/jkiss/dbeaver/ext/snowflake/model/SnowflakeSQLDialect.java
```diff
@@ -142,4 +142,32 @@ public String getColumnTypeModifiers(
...
+            int codePoint = str.codePointAt(i);
+            if (!isLatinLetter(codePoint) && codePoint != '_' && !Character.isDigit(codePoint) && codePoint != '$') {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static boolean isLatinLetter(int codePoint) {
```
serge-rider: Perhaps we should move it to SQLUtils.

serge-rider (changes requested):

LonwoLonwo:
Tests also can be created for this case, like ClickhouseDialectTest.

flambchap:
Good point @LonwoLonwo. For some reason, I thought it is impossible to test a dialect without a database connection :|

flambchap:
Unit tests added

LonwoLonwo (approved):

E1izabeth (approved):

Matvey16:
Verified

PR #20195, opened 2023-06-04
#19210 Adding an option to revert the sorting of columns

### LINKED ISSUE #19210: Want to enhance the sorting function
**The content is translated by Google, please forgive me if the description is wrong**

When you open the database to view all tables, or view table properties, you can sort by display or column name by clicking the table header, but once sorted, you can only sort in ascending or descending order, and the original order cannot be restored. I hope that the sorting function can be enhanced. Revert sorting in addition to ascending and descending

image

image

SiyaoIsHiding:
Fixing #19210 

Adding an option to sort the columns according to the original order.
Screenshot 2023-06-04 at 12 09 20 PM

I assume the feature request is about reverting the sorting of the "columns" table only, excluding the "Keys", "References" etc. Therefore, we just need to sort it according to the ascending order of the "#" column.

My current implementation is just to traverse all columns, and if there is a column with the name "#", add a menu option "Revert sorting".
```java
                	@Override
                	public void run() {
                       GroupingViewerColumnController.this.sortByColumn(finalNumberColumnInd, SWT.UP);
                	}
```

This is kinda ugly but works. But if we do not use this trick, we have to override the whole `SortListener` of `ViewerColumnController`. Please let me know if there is any better implementation.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/controls/ViewerColumnController.java
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/controls/ViewerColumnController.java
@@ -192,15 +192,6 @@ public class ViewerColumnController<COLUMN, ELEMENT> {
                 repackColumns(true);
             }
         });
-        menuManager.add(new Action(UINavigatorMessages.obj_editor_properties_control_action_columns_revert_sorting) {
-        	{
-        		setDescription(UINavigatorMessages.obj_editor_properties_control_action_columns_revert_sorting_description);
-        	}
-        	@Override
-        	public void run() {
-        		System.out.println("Revert sorting called");
-        	}
-        });
     }
 
     public void addColumn(
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/controls/ViewerColumnController.java
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/controls/ViewerColumnController.java
@@ -404,7 +404,7 @@ public class ViewerColumnController<COLUMN, ELEMENT> {
     }
 
     public void sortByColumn(int index, int direction) {
-        final ColumnInfo columnInfo = columns.get(index);
+        final ColumnInfo columnInfo = this.getColumnByIndex(index);
         columnInfo.sortListener.sortViewer(columnInfo.column, direction);
     }
 
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/controls/ViewerColumnController.java
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/controls/ViewerColumnController.java
@@ -192,6 +192,15 @@ public class ViewerColumnController<COLUMN, ELEMENT> {
                 repackColumns(true);
             }
         });
+        menuManager.add(new Action(UINavigatorMessages.obj_editor_properties_control_action_columns_revert_sorting) {
+        	{
+        		setDescription(UINavigatorMessages.obj_editor_properties_control_action_columns_revert_sorting_description);
+        	}
+        	@Override
+        	public void run() {
+        		System.out.println("Revert sorting called");
+        	}
+        });
     }
 
     public void addColumn(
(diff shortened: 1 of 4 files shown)


E1izabeth:
Thank you for contribution.

Please note that according to the [Contribution guide](https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code) it's required to add the ticket number in the commit message.

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/itemlist/ObjectListControl.java
```diff
@@ -1479,6 +1479,33 @@ private String copyGridToText() {
         @Override
         public void fillConfigMenu(IContributionManager menuManager) {
             super.fillConfigMenu(menuManager);
+            // Traverse all columns, if there is a column with name "#", add a
+            // menu option "Revert sorting"
+            // which is to sort in the ascending order of the "#" column
+            int columnsCount = this.getColumnsCount();
+            int numberColumnInd = -1;
+            for (int i = 0; i < columnsCount; i++) {
+                if (this.getColumnName(i).equals("#")) {
```
E1izabeth, line 1487: Could you clarify why are you searching for a column with name # here?

SiyaoIsHiding: Because the original order of the rows is just the ascending order of the "#" column.

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/controls/ViewerColumnController.java
SiyaoIsHiding, line 1: I find my previous implementation glitchy because the column get from `columns.get(index)` is different from the column get from `this.getColumnByIndex(index)`. May you confirm which one do you want?

SiyaoIsHiding:
> Thank you for contribution.
> 
> Please note that according to the [Contribution guide](https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code) it's required to add the ticket number in the commit message.

Modified accordingly.

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/itemlist/ObjectListControl.java
```diff
@@ -1479,6 +1479,29 @@ private String copyGridToText() {
         @Override
         public void fillConfigMenu(IContributionManager menuManager) {
             super.fillConfigMenu(menuManager);
+            // Traverse all columns, if there is a column with name "#", add a
+            // menu option "Revert sorting"
+            // which is to sort in the ascending order of the "#" column
+            int columnsCount = this.getColumnsCount();
+            for (int i = 0; i < columnsCount; i++) {
+                if (this.getColumnName(i).equals("#")) {
+                    final int numberColumnInd = i;
```
E1izabeth, line 1488: Why do you need this variable? I suppose you can just pass i as an argument.

SiyaoIsHiding: It will raise the error "Local variable i defined in an enclosing scope must be final or effectively final".

E1izabeth (approved):

ShadelessFox (changes requested):
Thanks for the contribution!

I don't like this approach. As you noted, it's pretty hacky.

I see two potential improvements:
1. In `ViewerColumnController.SortListener#handleEvent`, let's cycle through `SWT.UP`, `SWT.DOWN`, and `SWT.NONE`. This way, the sorting can be toggled off by clicking one extra time on an already sorted column.
2. The "Revert sorting" menu action can be displayed if `viewer.getXXX().getSortDirection()` returns something other than `SWT.NONE`. You can add a new method to check if sorting is present. To reset sorting, you can use `sortViewer(null, SWT.NONE);`

arhayka:
[SiyaoIsHiding](https://github.com/SiyaoIsHiding), thank you for your contribution. But we discussed this issue and decided to fix it another way without adding an option to revert the sorting.

What we want to do:
1. When clicking on a column header, sort the column ASC
2. On a second click, sort it DESC
3. On the third click return the original order

E1izabeth:
This PR is stale now, so I'm closing it. Please, let me know if you continue working on this - we can reopen it.

PR #20214, opened 2023-06-06
CB-3554 fix custom driver with several lib files

alexander-skoblikov:


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/driver/DriverLibraryLocal.java
```diff
@@ -154,6 +154,19 @@ public Path getLocalFile() {
...
     }
 
+    private Path resolveCacheDir() {
+        if (DBWorkbench.isDistributed()) {
+            // we do not have any provided drivers in distributed mode
+            return DriverDescriptor.getWorkspaceDriversStorageFolder();
+        }
+
+        if (isCustom()) {
+            return DriverDescriptor.getWorkspaceDriversStorageFolder();
```
LonwoLonwo, line 161: Maybe it is better to combine it with the previous condition?

PR #20216, opened 2023-06-06
#20139 fix PG and forks routines header creation - add default value …

### LINKED ISSUE #20139: DBeaver ignores default values for GreenPlum functions
### Description

When viewing into DDl of the GreenPlum functions, the DEFAULT keyword isn't generated for parameters

### DBeaver Version

Enterprise Edition 23.0.0. & Enterprise Edition 23.1.0

### Operating System

Windows Server 2019

### Database and driver

GreenPlum drivers based on Postgres 14 binaries

### Steps to reproduce

Double click on the function

or 

Genrerate DDL for a function

### Additional context

_No response_

LonwoLonwo:
fix PG and forks routines header creation - add a default value for parameters; 
add a new enum argument for Greenplum.

https://docs.vmware.com/en/VMware-Greenplum/7/greenplum-database/ref_guide-sql_commands-CREATE_FUNCTION.html

https://docs.vmware.com/en/VMware-Greenplum/6/greenplum-database/ref_guide-system_catalogs-pg_proc.html (new enum i)

Please check the following:

- [x] DEFAULT value added only for procedures/functions DDL in PG
- [x] DEFAULT value added only for procedures/functions DDL in Greenplum
- [x] You can rename function via UI (parameter name still working without default)
- [x] You can drop function via UI (parameter name still working without default)


uslss:
verified

PR #20218, opened 2023-06-06
dbeaver/dbeaver#18017 set hostname for credentials instead of password auth

### LINKED ISSUE #18017: SSH Auth Fail after 22.2.3 update
### Description

Hi, it's me again :)
So after your fix for https://github.com/dbeaver/dbeaver/issues/17748 that came out in 22.2.3 it's better. I can click on "test tunnel" and it logs in fine using jump server and asking me for cert keys.
But - when I try to connect to database, after all the popups to enter credentials I'm getting "auth error".

From the point of view of my bastion server it seems like "test tunnel" function uses the right user name from config, but when it connects to the database it makes my user name uppercase and then server says "nope, no such user". Not sure if it does that for both bastion and server, because it dies after trying to connect to the bastion.

Rolling back to 22.2.0 again.

### DBeaver Version

22.2.3

### Operating System

Windows 10

### Database and driver

_No response_

### Steps to reproduce

Exactly same as in https://github.com/dbeaver/dbeaver/issues/17748

### Additional context

_No response_

Destrolaric:
Closes dbeaver/dbeaver#18017

### DIFF
--- a/plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceDescriptor.java
+++ b/plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceDescriptor.java
@@ -1158,7 +1158,7 @@ public class DataSourceDescriptor
                     false
                 );
                 if (dbpAuthInfo != null) {
-                    if (rc.equals(DBWTunnel.AuthCredentials.PASSWORD)) {
+                    if (rc.equals(DBWTunnel.AuthCredentials.CREDENTIALS)) {
                         tunnelConfiguration.setProperty(getJumpServerSettingsPrefix(0) + DBConstants.PROP_ID_NAME, //$NON
                             // -NLS-1$
                             dbpAuthInfo.getUserName()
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceDescriptor.java
```diff
@@ -1158,7 +1158,7 @@ private boolean askForSSHJumpServerPassword(@NotNull DBWHandlerConfiguration tun
                     false
                 );
                 if (dbpAuthInfo != null) {
-                    if (rc.equals(DBWTunnel.AuthCredentials.PASSWORD)) {
+                    if (rc.equals(DBWTunnel.AuthCredentials.CREDENTIALS)) {
                         tunnelConfiguration.setProperty(getJumpServerSettingsPrefix(0) + DBConstants.PROP_ID_NAME, //$NON
                             // -NLS-1$
```
E1izabeth: Could you fix this comment please?

LonwoLonwo (approved):

E1izabeth (approved):

Matvey16:
Verified. Still isn't reproducible so a verification from user with ea version is required

PR #20262, opened 2023-06-12
#19992 add escape parameter for bulk load copy command

### LINKED ISSUE #19992: Export data feature fails
### Description

Export data feature fails : it creates new table in cross database but does not add data to it.  fails without error messge

### DBeaver Version

Community Edition 23.0.4

### Operating System

windows

### Database and driver

postgres sql 14.0

### Steps to reproduce

right click a table.
select export data and follow the steps to copy data to another database
it starts exporting and completes.  window goes away
it creates table but does not add data to it.
sometimes it is a success.  sometimes it fails.  it is not consistent

### Additional context

_No response_

LonwoLonwo:
2023-06-12 11_39_01-DBeaver Ultimate 23 1 0 - empty newtable2

Please check

- [x] - import from another table
- [x] - import from the file
- [x] - different special characters in a source

https://www.postgresql.org/docs/current/sql-copy.html


Matvey16:
Verified

PR #20266, opened 2023-06-12
dbeaver/pro#1151 Use GDI to make image from the browser with GIS

E1izabeth:
We can't get an image from the browser control because of the graphical acceleration usage.
However, we still can handle it by using native methods.


MashaKorax:
verified

PR #20271, opened 2023-06-12
dbeaver/pro#1644 compact ds config

serge-rider:


E1izabeth:
Lots of irrelevant changes from other commits happened to be in this PR for some reason.

PR #20278, opened 2023-06-13
dbeaver/pro#148 sql editor hide statistics

ShadelessFox:
image

Acceptance criteria:
- [x] If there are multiple results with result sets and the option is enabled, the statistics tab is shown
- [x] If there are multiple results with result sets and the option is disabled, the statistics tab is not shown
- [x] If there are no results, the statistics tab is always shown
- [x] **IMPORTANT** When refreshing a tab that has tabs after it, they remain open (#19839)


LonwoLonwo (approved):

MashaKorax:
verified

PR #20300, opened 2023-06-15
dbeaver/pro#1501 soft migrate license folder

Destrolaric:
Closes dbeaver/pro#1501


bundles/org.jkiss.utils/src/org/jkiss/utils/IOUtils.java
```diff
@@ -389,4 +389,34 @@ public static String getFileExtension(String fileName) {
         }
         return null;
     }
+
+    @NotNull
+    public static String getWorkingDirectory(String defaultWorkspaceLocation) {
```
serge-rider: Move to RuntimUtils.
IOUtils has nothing to do with Eclipse.

serge-rider (changes requested):

PR #20319, opened 2023-06-19
dbeaver/pro#1641 Events handling

E1izabeth:


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/EclipseWorkspaceImpl.java
```diff
@@ -59,10 +60,22 @@ public EclipseWorkspaceImpl(DBPPlatform platform, IWorkspace eclipseWorkspace) {
...
     public final void initializeProjects() {
-        initializeWorkspaceSession();
-
-        loadWorkspaceProjects();
-
+        try {
+            initializeWorkspaceSession();
+    
+            loadWorkspaceProjects();
+        } catch (DBException e) {
```
serge-rider: This code can be executed only for auth error

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPProject.java
```diff
@@ -108,7 +109,10 @@ public interface DBPProject extends DBPObject, SMAuthSpace, DBAPermissionRealm {
 
     @NotNull
     DBPDataSourceRegistry getDataSourceRegistry();
-
+    
+    @NotNull
```
serge-rider: Let's use another approach and don't touch DBPProject

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPWorkspaceDesktop.java
```diff
@@ -44,7 +44,9 @@ public interface DBPWorkspaceDesktop extends DBPWorkspaceEclipse {
      * Reloads workspace contents. Creates missing projects, removes unexistent projects
      */
     void refreshWorkspaceContents(DBRProgressMonitor monitor) throws DBException;
-
+    
+    void reloadWorkspaceProjects(DBRProgressMonitor monitor) throws DBException;
```
serge-rider: Let's move it to TE

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/preferences/SimplePreferenceStore.java
```diff
@@ -35,14 +36,31 @@ public abstract class SimplePreferenceStore extends AbstractPreferenceStore {
     private Map<String, String> defaultProperties;
     private boolean dirty = false;
 
-    public SimplePreferenceStore()
-    {
+    public interface Mutable {
```
serge-rider: Maybe we should just show notification about preferences change?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/preferences/SimplePreferenceStore.java
E1izabeth: There is enough synchronization for SQL Templates without additional synchronization in SimplePreferenceStore. We can handle cases with other settings when we will work on it.

PR #20341, opened 2023-06-20
Enforce explicit plugin versions

Tillerino:
Hi!

I recently bumped the Maven version in Nixpgks (https://github.com/NixOS/nixpkgs/pull/238746) and ran into some small but annoying manual effort w.r.t. the this package. The issue is that this package is packaged using the [double invocation](https://ryantm.github.io/nixpkgs/languages-frameworks/maven/#double-invocation) method which unfortunately breaks on a Maven upgrade if not all maven plugin _versions_ are set explicitly.

We're extending the documentation to prevent this (https://github.com/NixOS/nixpkgs/pull/238774), but I'm reaching out to fix the packages which have already been packaged. In this PR, we explicitly set the versions of a few plugins which hadn't had their versions set explicitly before and add a check with the enforcer plugin to make sure this won't accidentally break in the future.

Note that the exact versions of the plugins that you're using are not important to us and please do change them as you see fit. We only ask that you explicitly set the versions of all plugins so that downstream upgrades can run automatically.

Thank you.

(This was copy-pasta as I have a few repositories to go through. Please just ask if anything is unclear.)


E1izabeth:
Could you clarify why are you manually building DBeaver? You could reuse our tarballs for your packages.

E1izabeth:
Thank you for contribution.

E1izabeth (approved):

Tillerino:
I believe everything in Nixpkgs is built from source. Nix' thing is to solve the issue of incompatible library versions by linking everything exactly against the library version that it was built with and keeping an arbitrary number of alternatives for each library. I don't want to go into this too deep in case you are already familiar with Nix. If you're not, I definitely want to encourage you to check it out :)

You can see here how that works for the DBeaver package:

```
$ ldd /nix/store/r1pfv7ih9idgwd229wdcz00iqypsyzg3-dbeaver-22.2.2/dbeaver/dbeaver
	linux-vdso.so.1 (0x00007efd152b5000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007efd15294000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007efd1528f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007efd15000000)
	/nix/store/3n58xw4373jp0ljirf06d8077j15pc4j-glibc-2.37-8/lib/ld-linux-x86-64.so.2 => /lib64/ld-linux-x86-64.so.2 (0x00007efd152b7000)
```

Here, DBeaver is linked against glibc 2.37-8 _exactly_. I think you can argue that this is overkill for Java packages, or that ELF patching would also work with the tarball, but it's probably good to just keep everything consistent. And Java packages also like sneaking in native libraries at times :)

Anyway, I hope this helped.

E1izabeth:
Thank you for explanation

Destrolaric (approved):

serjiokov:
Hello @Tillerino, in first comment you were asking about plugin version to change, but the last point is glibc-2.37-8 that is a part of NixOS environment. Could you please clarify what are you trying to reach?

Tillerino:
Hi @serjiokov,

I am trying to make the DBeaver build _entirely_ deterministic - independent of the Maven version used for the build.

Imagine you emptied your `~/.m2/repository` and did a `mvn install` of DBeaver. Now hash `~/.m2/repository` and remember the hash. Repeat the entire process, but with a _different_ Maven version. Since the default versions of Maven's own plugins frequently change from Maven version to Maven version, the hash will likely be different this time. (Note that this would be ignoring timestamps, of course, which can be tricky, because some of the files _contain_ timestamps.)

DBeaver is in the Nix package repository: https://search.nixos.org/packages?channel=23.05&show=dbeaver

_Because Nix_ (see short explanation above in the reply to E1izabeth), it's being built from scratch there. Everything needs to be deterministic in Nix, and you need to give hashes for packages and (let's call them) build steps. What happens when we upgrade _Maven_ in Nix, is that the hash for _DBeaver_ changes occasionally, because of the versions problem I described above.

This means that when we upgrade Maven there, we need to change the hash of the DBeaver package (hand-wavingly). Most of the package upgrades are automated, but this non-deterministic DBeaver build prevents the automated upgrade. It's not the only package that does that (see the PR that links to this one), but I'm making progress.

_(And managing all plugin versions is also best practice according to the first article that turned up when I googled "Maven best practices", but I'd rather you do it as a favor to us 😄)_

serge-rider:
@Tillerino Generally I agree with this idea.

But we cannot accept this PR mainly because now we have `dbeaver-common` repository. 
Which is supposed to be used for such configurations (probably in https://github.com/dbeaver/dbeaver-common/blob/devel/root/pom.xml).
I can make another PR for that or you could propose yours.

Tillerino:
Thanks for your response.

Yes, dbeaver-common seems like the right place. But then it can't be in a single PR - possibly the change in common will break the other projects. So I guess I'll leave that up to you guys, then. I don't want to mess with your flow here :)

serge-rider:
👍

PR #20357, opened 2023-06-22
dbeaver/dbeaver#20340 add -version flag

### LINKED ISSUE #20340: Get binary version from command-line / shell
Dbeaver does not support a command to just see binary version.

It would be nice to have a "dbeaver-ce --version" command to check via terminal.

I am developing a [streamlined way to have dbeaver installed in Linux workstations](https://github.com/staticdev/linux-workstation-playbook) (but this command can do the same in other OS). The problem is, without a command to get the version, tools such as ansible have to redownload everytime the binary to check if it has changed or not. With such command I can even create an ansible role to have better update options for dbeaver.

I also tried executing the program and getting the version shown in the output, but that is not desirable also, since the process needs to be killed in workstation configuration time (which is not a general best practice).

flambchap:
Now the `-version` flag prints something like `DBeaver 23.1.1.202306221709` and exits the application. In real-world use (scripts), the `-q` flag should also be used to suppress logging. Otherwise, we'll see something like

```
> Start DBeaver Standalone [org.jkiss.dbeaver.ui.app.standalone 23.1.1.202306221709]
> Start Eclipse IDE UI Application [org.eclipse.ui.ide.application 1.5.0.v20230526-1051]
> Start Eclipse Jobs Mechanism [org.eclipse.core.jobs 3.14.0.v20230317-0901]
> Start Expression Language [org.eclipse.core.expressions 3.9.0.v20230317-0901]
DBeaver 23.1.1.202306221709
Commands processed. Exit DBeaver.
```


flambchap:
If the PR is accepted, we need to make an addition to https://github.com/dbeaver/dbeaver/wiki/Command-Line.

serge-rider:
I added debug logging suppress for such commands (version, help).

Matvey16:
Version

PR #20358, opened 2023-06-22
dbeaver/pro#1533 Migrate instance controller to REST

ShadelessFox:


bundles/org.jkiss.utils/src/org/jkiss/utils/rest/RestServer.java
```diff
@@ -0,0 +1,279 @@
...
+
+            exchange.sendResponseHeaders(response.code, 0);
+
+            if (response.type != void.class) {
+                try (Writer writer = new BufferedWriter(new OutputStreamWriter(exchange.getResponseBody()))) {
+                    gson.toJson(response.object, response.type, writer);
+                }
+            }
+
+            exchange.close();
```
serge-rider: It must be called in `finally`

bundles/org.jkiss.utils/src/org/jkiss/utils/BeanUtils.java
```diff
@@ -431,6 +431,20 @@ public static <T> T getFieldValue(@NotNull Object object, @NotNull String name)
...
 
+    @Nullable
+    public static Object handleObjectMethod(@NotNull Object proxy, @NotNull Method method, Object[] args) {
+        switch (method.getName()) {
+            case "toString":
+                return "Proxy";
+            case "hashCode":
+                return System.identityHashCode(proxy);
+            case "equals":
+                return proxy == args[0];
```
serge-rider, line 442: Why so strange equals implementation?
Also why do you use custom Object methods instead of calling them from `proxy` object?

serge-rider (changes requested):

bundles/org.jkiss.utils/src/org/jkiss/utils/BeanUtils.java
ShadelessFox: The `args[0]` stores the argument of the `equals` method, so we compare it with the proxy.

We can't use `equals` on the proxy itself, as it will result in recursion.

PR #20368, opened 2023-06-23
dbeaver/dbeaver#20367 Mixed cased tables cache bug fixed

### LINKED ISSUE #20367: Tables cache bug
### Description

Prerequisites: database schema contains multiple tables with the same name but in different case (e.g. TABLE, table, Table)
Problem: when we firstly open a table thru search or thru ERD or some other way when we load single table then cache breaks when we load the rest. Only first resolved table remains in cache all other are renamed into first object name.

### DBeaver Version

23.1

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

serge-rider:


Matvey16:
Verified

PR #20401, opened 2023-06-27
#20210 References panel refresh fix

### LINKED ISSUE #20210: References tab doesn't always update when a data filter is applied
### Description

In this example, I have a 'contact' table with a foreign key to a 'client' table. As I select a different contact record, the corresponding client is updated, as expected.

However, if I apply a filter to the contact table, and a new contact is selected, the corresponding client does not always update.

As a workaround, if multiple records happen to remain after I apply the filter, then I can press UP/DOWN to force an update. But when only one record is left, there is nothing I can do to update the references tab.

I have seen it occasionally update correctly (it happens at the end of the attached GIF), but I can't reliably replicate that behaviour. In any case, it fails more often than it works.

dbeaver-references-tab

### DBeaver Version

Community Edition Version 23.0.5.202305211610

### Operating System

Windows 10

### Database and driver

I've experienced this on PostgreSQL and MySQL, details:

1. PostgreSQL 15.2, using PostgreSQL JDBC Driver 42.5.2
2. MySQL 5.7.38-log, using MySQL Connector/J mysql-connector-java-8.0.29 

### Steps to reproduce

1. Select a record in a table with a foreign key
2. Apply a filter that excludes the current record
3. Expect to see reference tab update, but does not

### Additional context

_No response_

ShadelessFox:


E1izabeth (approved):

Matvey16:
Verified

PR #20413, opened 2023-06-28
Dbeaver/pro#1736 new model and size computations fix

Destrolaric:
Closes dbeaver/pro#1736
Fixes size computation for token limit & adds gpt-3.5-turbo-16k model by default


serge-rider:
Let's add smart error recovery

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/gpt3/GPTCompletionEngine.java
```diff
@@ -196,13 +196,20 @@ private String requestCompletion(
...
                         break;
                     } catch (Exception e) {
-                        if (e instanceof HttpException && ((HttpException) e).code() == 429) {
-                            RuntimeUtils.pause(1000);
+                        if ((e instanceof HttpException && ((HttpException) e).code() == 429)
+                            || (e instanceof OpenAiHttpException && e.getMessage().contains("This model's maximum"))) {
+                            if (e instanceof HttpException) {
+                                RuntimeUtils.pause(1000);
+                            } else {
+                                responseSize /= 2;
```
serge-rider: It is not "smart" size detection. We can analyze error message and calculate exact maximum size of response.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/gpt3/GPTCompletionEngine.java
```diff
@@ -207,13 +209,29 @@ private String requestCompletion(
...
+                                    String numberStr = matcher.group();
+                                    int number = Integer.parseInt(numberStr);
+                                    numbers.add(number);
+                                }
+                                if (numbers.size() < 4) {
+                                    throw e;
+                                }
+                                Integer promptSize = numbers.get(2);
+                                responseSize = Math.min(responseSize,
+                                    GPTModel.getByName(getModelName()).getMaxTokens() - promptSize - 1);
```
E1izabeth: Oh, can we rely on some words in the message to get the required size instead of relying on the ordinal position of the number token in this message?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/gpt3/GPTCompletionEngine.java
```diff
@@ -196,17 +198,40 @@ private String requestCompletion(
...
-                            RuntimeUtils.pause(1000);
+                        if ((e instanceof HttpException && ((HttpException) e).code() == 429)
+                            || (e instanceof OpenAiHttpException && e.getMessage().contains("This model's maximum"))) {
+                            if (e instanceof HttpException) {
+                                RuntimeUtils.pause(1000);
+                            } else {
+                                // Extracts resulted prompt size from the error message and resizes max response to
+                                // value lower that (maxTokens - prompt size)
+                                Matcher matcher = Pattern.compile("([0-9]+)").matcher(e.getMessage());
+                                ArrayList<Integer> numbers = new ArrayList<>();
```
serge-rider: Do not use ArrayList as var type.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/gpt3/GPTCompletionEngine.java
```diff
@@ -196,17 +198,40 @@ private String requestCompletion(
...
-                        if (e instanceof HttpException && ((HttpException) e).code() == 429) {
-                            RuntimeUtils.pause(1000);
+                        if ((e instanceof HttpException && ((HttpException) e).code() == 429)
+                            || (e instanceof OpenAiHttpException && e.getMessage().contains("This model's maximum"))) {
+                            if (e instanceof HttpException) {
+                                RuntimeUtils.pause(1000);
+                            } else {
+                                // Extracts resulted prompt size from the error message and resizes max response to
+                                // value lower that (maxTokens - prompt size)
+                                Matcher matcher = Pattern.compile("([0-9]+)").matcher(e.getMessage());
```
serge-rider: Use proper regexp which matches error message. Current one just searches for any numbers.
If error message doesn't match expected one then do not try to recalculate tokens.

serge-rider (changes requested):

MashaKorax:
dbeaver_s6fJTXPpoD
` show chave,dh, folha, colaborator, tipoabono, dhi. dhf, cid from abonos. also show all from actor`

postgres 10.  schema public -model gpt-3.5-turbo-16k. scope - current schema

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/gpt3/GPTCompletionEngine.java
```diff
@@ -70,10 +72,19 @@ public class GPTCompletionEngine implements DAICompletionEngine {
...
     private static final boolean SUPPORTS_ATTRS = true;
 
+    private static final Pattern sizeErrorPattern = Pattern.compile(
+        "This model's maximum context length is [0-9]+ tokens, "
+            + "however you requested [0-9]+ tokens \\(([0-9]+) in your prompt; [0-9]+ "
+            + "for the completion\\)\\. Please reduce your prompt; or completion length");
+    private static final Pattern chatSizeErrorPattern = Pattern.compile(
+        "This model's maximum context length is [0-9]+ tokens\\. However, "
+            + "you requested [0-9]+ tokens \\(([0-9]+) in the messages, [0-9]+ "
+            + "in the completion\\)\\. Please reduce the length of the messages or completion\\.");
```
E1izabeth: They're almost the same. What's the point to keep them both?

serge-rider: And I think id doesn't make sense to add trailing part (please reduce...). 
Make regex as simple as possible to match error messagem difference between two regexps is just a few words which can be replaced with `.+`

MashaKorax:
verified

PR #20418, opened 2023-06-28
#19402 Fix regexp for object name extraction for Oracle

### LINKED ISSUE #19402: Procedure editor changes to wrong name when saving
### Description

When saving after editing a procedure (but before committing), if the procedure name has quotes and the inner text contains `procedure \w+\s` (either as comment or code), DBeaver will identify the `\w+` as the name of the procedure, causing problems until the procedure list is refreshed.

This has been happening for months, but I didn't identify what exactly was happening until now, because the name was being changed to a mysterious "WILL". Logs only show errors that happen due to the name being replaced, but don't help identify why.

### DBeaver Version

23.0.0.202303040621

### Operating System

Windows 11

### Database and driver

OraClient19Home1

### Steps to reproduce

Reproduction (MYSCHEMA does not have to exist for this test and can also be removed):
```sql
CREATE OR REPLACE PROCEDURE MYSCHEMA."TEST_PROC"
-- SomeProcedure Wrongname
IS
BEGIN
    null;
END;
```
image

The quotes around "TEST_PROC" are by DBeaver. The bug won't be triggered if they are removed.

The same thing happens when editing functions, but you need to use the "FUNCTION" keyword instead.
```sql
CREATE OR REPLACE FUNCTION MYSCHEMA."TEST_FUNC"
-- SomeFunction Wrongname
IS
BEGIN
    null;
END;
```

### Additional context

_No response_

E1izabeth:


Matvey16:
Verified

PR #20462, opened 2023-07-03
dbeaver/pro#1739 Open external file using preferred application

ShadelessFox:


E1izabeth (approved):

LonwoLonwo (approved):

Matvey16:
Doesn't work on Mac because of file permissions

Screenshot 2023-07-05 at 13 58 22

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/ShellUtils.java
```diff
@@ -40,6 +43,35 @@ public static boolean launchProgram(@NotNull String path) {
...
+            }
+        } catch (IOException | InterruptedException e) {
+            log.debug("Unable to open external program in a platform-specific way: " + e.getMessage());
+            return false;
+        }
+
+        try {
+            Desktop.getDesktop().open(path.toFile());
+            return true;
+        } catch (IOException e) {
```
Destrolaric: Linux may not have 'xdg-open' in some cases, we should use try to use 'getDesktop().open()' in that case.

ShadelessFox: That `return false;` was excessive, I agree.

Matvey16:
Verified on Linux

PR #20494, opened 2023-07-06
#20492 update data grid accessibility for the JAWs application

### LINKED ISSUE #20492: JAWs - enhance working with GRID
We need to enhance data grid cell reading with the JAWs application.

For now, the last version of JAWs reads only values from even rows in the data grid. Only values without column name and row number (as NDVA can).

2023-07-06 14_53_23-DBeaver 23 1 1 - DEPARTMENT

Unchecking of even/odd rows highlighting helps partially.

2023-07-06 14_54_31-Preferences

LonwoLonwo:
Accessibility is updated for JAWs application only in the data grid. For cells.

Please check it for JAWs and NVDA


MashaKorax:
If you change and do not save the cell value, move to another cell using the keyboard arrows - it stops reading, reads only even cell values, or stops working altogether. Nvda works correctly

Note - if you replace to another cell using a mouse - Jaws works. But the isuue is about accessibility
dbeaver_SGK89grz5M

MashaKorax:
verified

PR #20526, opened 2023-07-10
dbeaver/dbeaver#18733 Column description in metadata panel

### LINKED ISSUE #18733: Column description is missing on Panels/Metadata
### Description

Friends, thanks for great piece of software !!! To make it event better I decided to create this bug report because I'm convinced that this feature should work but it don't.

In our company most important tables and views are well documented, it means they have Table/View/Column descriptions set. This is very important for educated business users writing theirs own SQL scripts. The problem is that those descriptions does not appear on Metadata panel that is useful during sql design. I suspect that Description shall appear on LABEL column, because now LABEL contains the same value as COLUMN NAME, so I'm convince that this is a mistake.

Thanks you.

### DBeaver Version

Community edition 22.3.0

### Operating System

macOS Ventura

### Database and driver

Latest MS SQL server driver

### Steps to reproduce

see bug description

### Additional context

_No response_

serge-rider:


MashaKorax:
Name  is changed as well if alias was set.
5RrlJcLULF

serge-rider:
This is not a bug, it is how PostgreSQL driver works.
Anyhow the ticket is about description column )

MashaKorax:
verified

PR #20527, opened 2023-07-10
#18094 Show labels on the map

### LINKED ISSUE #18094: Spatial viewer should support labels on map
The spatial viewer is nice, however just showing the geometry when viewing all data for a table and selecting the Spatial tab might not help getting an overview of the corresponding data.

I would like an option of labels next to the geometry showed on the map, e.g. by selecting a corresponding column or even an expression to combine more fields.

It doesn't look like it's possible to customize the geometry output at all (besides different columns use different colours)

**Additional context**
Quick mock-up of desired result from the spatial view:

image

ShadelessFox:
Note: the used approach utilizes description columns for dictionary entities. I thought it would be suitable here.

Acceptance criteria:
- [ ] Description columns can be configured while in the value panel
- [ ] Description columns can be configured while in the spatial presentation
- [ ] Label visibility button state is preserved when closing/reopening the viewer

image


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/virtual/DBVEntity.java
```diff
@@ -483,18 +483,37 @@ public Collection<DBSEntityAttribute> getDescriptionColumns(DBRProgressMonitor m
...
+    }
+
+    @NotNull
+    public static <T extends DBSAttributeBase> Collection<T> getDescriptionColumns(
+        @NotNull Collection<? extends T> attributes,
+        @NotNull String descColumns
+    ) {
+        if (CommonUtils.isEmpty(descColumns)) {
+            return Collections.emptyList();
+        }
```
Destrolaric: Redundant second empty check

plugins/org.jkiss.dbeaver.data.gis.view/src/org/jkiss/dbeaver/ui/gis/panel/actions/ConfigureLabelsAction.java
```diff
@@ -0,0 +1,63 @@
...
+            // access to the result set. Viewers themselves don't have access,
+            // so we instead cause a complete refresh of the presentation
+
+            if (presentation instanceof GeometryPresentation) {
+                presentation.refreshData(false, false, true);
+            } else {
+                presentation.getController().updatePanelsContent(false);
+            }
+        }
+    }
```
Destrolaric, line 56: Could you describe the parts which should be tested? This looks like it may require additional testing.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/virtual/DBVEntity.java
ShadelessFox: No, it's not. During the first check, the collection might be `null`. During the second check, we verify it has elements.

plugins/org.jkiss.dbeaver.data.gis.view/src/org/jkiss/dbeaver/ui/gis/panel/actions/ConfigureLabelsAction.java
ShadelessFox: Just the presentation and the value panel, nothing fancy. I'll include that in the criteria anyways.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/virtual/DBVEntity.java
Destrolaric: In both situation we return an empty list, Maybe it better to make it nullable and do only one check instead?

E1izabeth (approved):

uslss:
Labels for polygons aren't correct on the spatial view (pg: db2.gis.public.planet_osm_polygon)

https://github.com/dbeaver/dbeaver/assets/31996417/fc1c7e50-410b-44c4-8e9b-14db2faf7816

uslss:
verified

PR #20538, opened 2023-07-11
dbeaver/dbeaver#20383 Unquote objects names on persist

### LINKED ISSUE #20383: Creating a table with quotes in the name breaks the cache
### Description

If the table is created via UI and the name is quoted, columns in priperties tab will desappear after saving and data tab will be shown as read-only untill the schema is refreshed

### DBeaver Version

23.1.1

### Operating System

Any

### Database and driver

PostgreSQL 12

### Steps to reproduce

1. Right-click on Schema and select "create table"
2. Type table name with quotes e.g. `"Table1"`
3. Add some columns
4. Save changes

**Expected:** Columns in properties tab are shown as created
**Actual:** Columns disappear

### Additional context

dbeaver_cBAWOXsRxR

serge-rider:


MashaKorax:
verified

PR #20549, opened 2023-07-11
Add CrateDB variant of PostgreServer to improve compatibility

proddata:
Hi D🦫 Team,

We traditionally recommended to used the standard PostgreSQL driver for CrateDB due to its increasing compatibility. However, we have been made aware that some [recent changes in DBeaver have impacted this compatibility.](https://github.com/crate/crate/issues/14395)

To remedy this and improve user experience, we are proposing a new pull request for the addition of a CrateDB variant to the PostgreSQL connector.

This PR provides the base implementation, designed to adjust to CrateDB features while retaining essential PostgreSQL driver functionalities. Also this moves the generic CrateDB connector using the CrateDB specific JDBC Driver to legacy, as with recent CrateDB versions the default recommendations is to use the standard PostgreSQL JDBC Driver.

Best,
Georg from Crate.io

**Screenshots**:

image

image

image


ShadelessFox (commented):
Thanks for the contribution!

I expect white icons to be practically invisible on a light theme.

Please add a contrasting background or border.

proddata:
> I expect white icons to be practically invisible on a light theme.
> 
> Please add a contrasting background or border.

Thanks for the feedback. I didn't think about light teams initially.
Changed the icons to standard CrateDB blue, which should work with both light and dark themes.

Dark
image
image

---

Light
image
image

image

ShadelessFox (approved):
👍🏻

LonwoLonwo (approved):

LonwoLonwo:
Looks nice. Thanks for the contibution!

E1izabeth (approved):

LonwoLonwo:
Hello @proddata 

We have a question now. Do we need to keep the Legacy CrateDB version? Any specific reason for that?

proddata:
> Hello @proddata
> 
> We have a question now. Do we need to keep the Legacy CrateDB version? Any specific reason for that?

Only with CrateDB 5.x (Released on 2022-07-11) and up the PostgreSQL compatibility imho was good enough to use it with the standard PG connection. We still see many 4.x and 3.x clusters around. While those versions are not officially supported anymore one would cut out those users.

So optimally one would keep it around for a few more months (maybe end year) and then sunsets the legacy version. However I would also understand if you would want to remove it right away.

LonwoLonwo:
Ok. Thanks. Can we also use a new icon image for the old driver?

proddata:
> Ok. Thanks. Can we also use a new icon image for the old driver?

Sure, should they be copied to the generic folder or referenced?

LonwoLonwo:
Yes, please change the icon (not the logo) in the generic bundle, and in the postgres bundle use:

```
icon="platform:/plugin/org.jkiss.dbeaver.ext.generic/icons/cratedb_icon.png"
iconBig="platform:/plugin/org.jkiss.dbeaver.ext.generic/icons/cratedb_icon_big.png"
```

proddata:
> Yes, please change the icon (not the logo) in the generic bundle, and in the postgres bundle use:

done ✅ 

image
image

LonwoLonwo:
M. No. The point was to avoid od duplicating images in DBeaver. 
So, You need to remove icons in the PG bundle and change the path to icons like in the example

proddata:
> M. No. The point was to avoid od duplicating images in DBeaver. So, You need to remove icons in the PG bundle and change the path to icons like in the example

Ah, sorry for the misunderstanding 🙈  
Should be fixed now.

LonwoLonwo:
Yes, that's better, thanks!

uslss:
verified

Issue #20514, opened 2023-07-09
Materialize driver


benesch:
👋🏽 , hi, DBeaver team! We'd like to collaborate with you to bring a native driver for [Materialize](https://materialize.com) to DBeaver.

Materialize is a real-time data warehouse. Like CockroachDB, Greenplum, Redshift, Yugabyte, et al, Materialize is a mostly-PostgreSQL-compatible database.

While using the existing PostgreSQL adapter mostly works, we'd like to bring a native adapter to DBeaver for two reasons:

  1. Improving discoverability and reducing user confusion. Users often aren't aware that DBeaver integrates well with Materialize using the PostgreSQL driver. Having a dedicated "Materialize" connection in the create connection screen makes it clear that Materialize and DBeaver work well together.
  2. Eventually we'd love to expose Materialize-specific concepts to DBeaver—e.g., [sources](https://materialize.com/docs/sql/create-source/) and [sinks](https://materialize.com/docs/sql/create-sink/). Since these are not concepts in PostgreSQL, there's no way for us to expose them in DBeaver via the PostgreSQL driver.

If it's acceptable to y'all, we'd love to open a PR, modeled after the [commit to add support for CockroachDB](https://github.com/dbeaver/dbeaver/commit/623b469b9c7319c6a94a5b196cc010167316cf7e), to kick things off!

Destrolaric:
@benesch Hi, I think it will be a great idea! Please let me know if you have any questions.

uslss:
implemented in 21.1.3

PR #20553, opened 2023-07-12
Materialize driver

bobbyiliev:
As discussed in #20514 adding the initial work for the Materialize driver.

dbeaver-materialize-driver-initial-work


ShadelessFox (commented):
Thanks for the contribution!

Can padding on the icons and logo be reduced so they don't look small in the UI?

bobbyiliev:
Thanks for that suggestion @ShadelessFox! Just pushed the updated logos!

ShadelessFox (approved):
👍🏻

LonwoLonwo (approved):

E1izabeth (approved):

uslss:
verified

PR #20558, opened 2023-07-12
#20313 show confirmation dialog for result set filtering even user sa…

### LINKED ISSUE #20313: Filter resultset confirmation makes disabled "filter by value"
### Description

if you apply Filter by value in the table, a dialog box will appear, if you check the Do not ask me again checkbox and click no - after that it is still possible to press Filter by value, nothing will happen. Maybe we should change **no** to **cancel**

### DBeaver Version

23.1.0

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

LonwoLonwo:
show confirmation dialog for the result set filtering even user said "no" and "do not ask me again"

2023-07-12 16_02_39-DBeaver Ultimate 23 1 2 - Customer


ShadelessFox (changes requested):
Let's just use `ConfirmationDialog.CONFIRM` instead of `ConfirmationDialog.QUESTION` as the dialog type so no "No" button is shown:

java_bfFS6WvveB

LonwoLonwo:
@ShadelessFox 
Yep, you are correct

ShadelessFox (approved):

E1izabeth (approved):

MashaKorax:
verified

PR #20561, opened 2023-07-12
#19962 Use container from active pipe rather than saved settings

### LINKED ISSUE #19962: Data Import: do not connect to the database if the source is not a table
DBeaver tries to connect to the database always when you import some data.
We need to improve this behavior:
- need to connect if the import source is a table
- don't need to connect anywhere if the import source is a CSV, XLS, or XML file.

image

ShadelessFox:
Acceptance criteria:
- [ ] Importing from a file should not cause the wizard to establish a connection to the previously selected consumer database
- [ ] Make sure import to several tables from different database at the same time works correctly (select multiple tables)


E1izabeth (approved):

uslss:
Acceptance criteria:

- [x]  Importing from a file should not cause the wizard to establish a connection to the previously selected consumer database
- [x]  Make sure import to several tables from different database at the same time works correctly (select multiple tables)

uslss:
verified

PR #20568, opened 2023-07-13
#20467 Fix displaying images in browser on Wayland

### LINKED ISSUE #20467: Images are not displayed in the value panel on Linux
### Description

On Linux, when you open the value panel for a blob column with pictures in it, the value panel does not display them regardless of User Interface ->  Web Browser settings.

### DBeaver Version

CE 23.1.1

### Operating System

Debian 12

### Database and driver

_No response_

### Steps to reproduce

1. Create a new Sample Database
2. Open table Album
3. Click on the first or second cell in the Column1 column
4. Open the value panel. It's empty

### Additional context
image

E1izabeth:


E1izabeth:
Open in external editor now suggests to choose an application on Windows, it used to open an image in the associated program previously. I suppose we can write file without extension only for linux

E1izabeth:
I haven't figure out another the way to get the file extension from input stream without additional libs. So, we write file without extension only on linux - open in external editor works for me on debian 12 even without file extension. No changes for mac and windows.

DariaMarkaryan:
Verified

PR #20572, opened 2023-07-13
CB-3268 feat: admin users pagination

Wroud:


plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMAdminController.java
```diff
@@ -74,7 +74,10 @@ void createUser(
     SMUser getUserById(String userId) throws DBException;
 
     @NotNull
-    SMUser[] findUsers(String userNameMask) throws DBException;
+    SMUser[] findUsers(@NotNull Integer first, @Nullable String after, @NotNull String userIdMask, Boolean enabledState)
```
alexander-skoblikov: What is `first`? What is `after`?
Why `first` is Integer, but `after` is string? Do you mean limit/offset?

plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMAdminController.java
```diff
@@ -74,7 +74,10 @@ void createUser(
...
+            throws DBException;
+
+    int countUsers(@Nullable String after, @NotNull String userIdMask, Boolean enabledState) throws DBException;
```
alexander-skoblikov: is this method really needed in the interface?

alexander-skoblikov: it is better to move the 'user filter' in a separate class and use one model to count / search users

plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMAdminController.java
Wroud: replaced with limit/offset

plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMAdminController.java
Wroud: filters moved to separate class

Wroud: this method is not in use right now but it is required for pagination (to calculate pages)

PR #20588, opened 2023-07-18
dbeaver/pro#1526 ability to create execution plan for Vertica added; …

LonwoLonwo:
- the ability to create an execution plan for Vertica added
- some code refactoring - removing duplicate methods
- split procedures and functions for Vertica PRO


plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/plan/MySQLPlanAnalyser.java
```diff
@@ -120,33 +118,19 @@ public void addNodeProperties(DBCPlanNode node, JsonObject nodeJson) {
                     attributes.add("extra", new JsonPrimitive(CommonUtils.notEmpty(plainNode.getExtra())));
                 } else if (node instanceof MySQLPlanNodeJSON) {
                     MySQLPlanNodeJSON jsNode = (MySQLPlanNodeJSON) node;
-                    for(Map.Entry<String, String>  e : jsNode.getNodeProps().entrySet()) {
-                        attributes.add(e.getKey(), new JsonPrimitive(CommonUtils.notEmpty(e.getValue())));
+                    for(Map.Entry<String, Object>  e : jsNode.getNodeProps().entrySet()) {
+                        Object value = e.getValue();
+                        if (value instanceof Double) {
+                            attributes.add(e.getKey(), new JsonPrimitive((Double) value));
```
E1izabeth: It would be useful to know why do we need this special workaround for double here, but don't in other places. Could you leave some comment?

LonwoLonwo: For me, it was clear but ok.

E1izabeth: Do we need to test mysql execution plan also then?

LonwoLonwo: Yes. It is on the list.

PR #20589, opened 2023-07-18
#20574 avoid procedures duplication - remove columns objects from the…

### LINKED ISSUE #20574: Duplication of functions in the list
### Description

I have 1 functions. Query return 1 line:
`SELECT * FROM pg_proc pp, pg_namespace n WHERE pp.pronamespace = n.oid AND pp.proname = 'group_calc' and n.nspname='my_sch'`

And I have many comments . Query return 45 line:
`SELECT pgd.description, pp.oid, pp.*
FROM pg_proc pp 
join pg_namespace n on pp.pronamespace = n.oid
left join pg_catalog.pg_description pgd on  pgd.objoid = pp.oid
WHERE pp.proname = 'group_calc' and n.nspname='my_sch'`

I see in Dbeaver:
Screenshot 2023-07-14 134216
This is one function, and it duplication 

### DBeaver Version

23.1.2.202307091549

### Operating System

_No response_

### Database and driver

1.PostgreSQL 15.3 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44), 64-bit
2.PostgreSQL JDBC Driver 42.5.2

### Steps to reproduce

1. Create function.
2. Create many comments.

### Additional context

_No response_

LonwoLonwo:
… query condition

From the [documentation](https://www.postgresql.org/docs/current/catalog-pg-description.html):

```
objsubid int4
For a comment on a table column, this is the column number (the objoid and classoid refer to the table itself). For all other object types, this column is zero.
```

We do not have columns in this query case, so we can add this condition in our statement to avoid strange duplicates from PG oid messing up.

Please check that procedures/functions are still available in PG and Redshift databases.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
@@ -1126,6 +1126,7 @@ public class PostgreSchema implements
                     "FROM pg_catalog." + serverType.getProceduresSystemTable() + " p\n" +
                     "LEFT OUTER JOIN pg_catalog.pg_description d ON d.objoid=p." + oidColumn + "\n" +
                     "WHERE p.pronamespace=?" +
+                    (session.getDataSource().isServerVersionAtLeast(7, 2) ? " AND d.objsubid = 0" : "") + // no links to columns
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

ShadelessFox (approved):

uslss:
Functions aren't visible for redshift in CE
image

E1izabeth (approved):

uslss:
verified

PR #20590, opened 2023-07-18
#17692 Reconnect to the last chosen database on demand

### LINKED ISSUE #17692: Import data saves the last used target container
The target container is where the data is uploaded, schema or database. We can choose any.
But there may be an attempt to connect to the last used container. In this case, the connection to the database takes a long time, and the UI may freeze. Also, you can encounter an error **'an error occurred while logging in'**.

ShadelessFox:
Wizard:
java_rCT84jo9AR

Preferences:
java_bAQ0X50YYY


E1izabeth (approved):

uslss:
verified

PR #20599, opened 2023-07-19
dbeaver/pro#1775 fix jaws issues

Destrolaric:
Closes dbeaver/pro#1775
`CTRL + ALT + PAGEDOWN/PAGEUP` replaced with `CTRL+SHIFT+PAGEUP/PAGEDOWN`|

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/folders/TabbedFolderList.java
+++ b/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/folders/TabbedFolderList.java
@@ -113,7 +113,7 @@ public class TabbedFolderList extends Composite {
          * @param index  the index in the list.
          */
         public ListElement(Composite parent, final TabbedFolderInfo tab, int index) {
-            super(parent, SWT.NO_FOCUS);
+            super(parent, SWT.NONE);
             this.tab = tab;
             hover = false;
             selected = false;
@@ -294,7 +294,7 @@ public class TabbedFolderList extends Composite {
          * @param parent the parent Composite.
          */
         public TopNavigationElement(Composite parent) {
-            super(parent, SWT.NO_FOCUS);
+            super(parent, SWT.NONE);
             addPaintListener(this::paint);
             addMouseListener(new MouseAdapter() {
 
@@ -373,7 +373,7 @@ public class TabbedFolderList extends Composite {
          * @param parent the parent Composite.
          */
         public BottomNavigationElement(Composite parent) {
-            super(parent, SWT.NO_FOCUS);
+            super(parent, SWT.NONE);
             addPaintListener(this::paint);
             addMouseListener(new MouseAdapter() {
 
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/folders/TabbedFolderList.java
```diff
@@ -1021,7 +988,7 @@ public void addSelectionListener(SelectionListener listener) {
 
     public void handleTraverse(TraverseEvent e) {
         if (e.detail == SWT.TRAVERSE_PAGE_PREVIOUS || e.detail == SWT.TRAVERSE_PAGE_NEXT) {
-            if ((e.stateMask & SWT.ALT) != SWT.ALT) {
+            if ((e.stateMask & SWT.SHIFT) != SWT.SHIFT) {
```
E1izabeth: Is it necessary to change shortcuts?

E1izabeth (commented):
Please, add a video showing how it works now, at least
- jumping to vertical tabs in and out
- traversal through all controls with tab
- pressing shortcut while editing some property

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/folders/TabbedFolderList.java
Destrolaric: Unfortunately, yes. Jaws uses ALT version for adjusting speech speed.

E1izabeth: Have you checked it in NVDA? I guess you should as it may stop working there by the similar reason

Destrolaric: Checked

Destrolaric:
> Please, add a video showing how it works now, at least

Takes too much space and can't be uploaded here, ask me directly if needed

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/folders/TabbedFolderList.java
LonwoLonwo: Ok, but the comment is still
// Only in case of CTRL+ALT+PG

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/folders/TabbedFolderList.java
```diff
@@ -1021,7 +985,7 @@ public void addSelectionListener(SelectionListener listener) {
...
                 // Only in case of CTRL+ALT+PG
```
LonwoLonwo: Change the comment also then

LonwoLonwo (approved):

LonwoLonwo:
Fixed here: https://github.com/dbeaver/dbeaver/pull/21405

PR #20614, opened 2023-07-19
dbeaver/dbeaver#20528 Initial commit for my case insensitive feature.

### LINKED ISSUE #20528: Order case insensitive as an option at the Client Side

**Additional context**

vijayap20:
Order case insensitive as an option at the Client Side #20528[](url)


E1izabeth:
Thank you for contribution.
We don't want to add such option. Let's just change the behaviour of the existing `Always on client` setting. The sorting should always be case-insensitive.

Also only strings should be compared in such a way - it doesn't make sense to do it for numbers.
And pay attention on different locales. It should correctly work with non-latin languages.

E1izabeth:
Please delete all changes that are no longer in use

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -1123,7 +1123,7 @@ public void resetOrdering() {
                     }
                     Object cell1 = getCellValue(new ResultSetCellLocation(binding, row1));
                     Object cell2 = getCellValue(new ResultSetCellLocation(binding, row2));
-                    result = DBUtils.compareDataValues(cell1, cell2);
+                    result = DBUtils.compareDataValues((Object) cell1.toString().toLowerCase(), (Object) cell2.toString().toLowerCase());
```
E1izabeth: Please do such comparison only for string types. It makes no sense to do it for numbers or binary objects.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -1123,7 +1123,12 @@ public void resetOrdering() {
                     }
                     Object cell1 = getCellValue(new ResultSetCellLocation(binding, row1));
                     Object cell2 = getCellValue(new ResultSetCellLocation(binding, row2));
-                    result = DBUtils.compareDataValues(cell1, cell2);
+                    if(cell1 instanceof String && cell2 instanceof String) {
```
ShadelessFox:
```suggestion
                    if (cell1 instanceof String && cell2 instanceof String) {
```

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -1123,7 +1123,12 @@ public void resetOrdering() {
...
+                    	result = DBUtils.compareDataValues((Object) cell1.toString().toLowerCase(), (Object) cell2.toString().toLowerCase());                    	
+                    }else {
```
ShadelessFox:
```suggestion
                    } else {
```

vijayap20: resolved

ShadelessFox (approved):

E1izabeth:
Passed to QA team.

Matvey16:
Verified

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -1123,7 +1123,12 @@ public void resetOrdering() {
                     }
                     Object cell1 = getCellValue(new ResultSetCellLocation(binding, row1));
                     Object cell2 = getCellValue(new ResultSetCellLocation(binding, row2));
-                    result = DBUtils.compareDataValues(cell1, cell2);
+                    if (cell1 instanceof String && cell2 instanceof String) {
```
serge-rider, line 1126: WE can use simple compareToIgnoreCase here.

serge-rider (changes requested):

E1izabeth:
@vijayap20 would you made the requested changes?

E1izabeth (approved):

LonwoLonwo (approved):

Matvey16:
Verified

serge-rider (approved):

PR #20622, opened 2023-07-20
#20507 check connection types first during the auto-commit recover

### LINKED ISSUE #20507: "Return to auto-commit on transaction end" setting on by default for production connections
### Description

Recently we noticed a change in the default behaviour of dbeaver at our company. All production connections now have the "Return to auto-commit on transaction end" on by default.

image

This is not exactly a bug but in my opinion it's quite a dangerous default and making this change unannounced can result in unexpected data loss. For example this will result in permanent data loss

```sql
SELECT * from foobar;
-- DBeaver has now returned to auto-commit mode
DELETE from foobar;
```

### DBeaver Version

Community 23.1.1.202306251800

### Operating System

Ubuntu 22.04

### Database and driver

PostgreSQL

### Steps to reproduce

1. Connect to a production database
2. Run a select query
3. Observe that auto-commit mode is on

### Additional context

_No response_

LonwoLonwo:
2023-07-20 16_56_52-DBeaver Ultimate 23 1 3 - _postgres_ Script-45

2023-07-20 16_57_11-DBeaver Ultimate 23 1 3 - _postgres_ Script-45

Now the "return to auto-commit" setting depends on the "smart commit mode" setting in Connection types and Transactions. And if the default value for the smart commit was false, auto recovery now also changed to false.

Specific Data source Transactions setting is prioritized under connection types for smart commit and auto-commit recovery (ask me).


Matvey16:
Verified

Issue #18312, opened 2022-11-23
Found some hard-coded cases where the localization properties file was not found (part 4)


bianyun1981:
#### System information: 
- Operating system (distribution) and version: Windows 11 Professional Edition (Simplified Chinese), 21H2
- DBeaver version: 22.2.5

#### Describe the problem you're observing:
After I updated the code of the develop branch to the latest version (last commit time: Sun Nov 20 23:35:31 2022 +0300, commit by plotn <plotnikovya@gmail.com>, commit hash: 7fb7c33fbf4fb9d99a1c42a6b3f5cdf6116ddb72), I still found the following hard-coded cases without corresponding localization properties file. 

After these properties files are created, I can continue to complete the Simplified Chinese localization of these contents.

<details>

```

---------------------------------------------------------------------------------------------------------------------
plugins/org.jkiss.dbeaver.ext.mysql.ui/plugin.xml
---------------------------------------------------------------------------------------------------------------------
  - line 59:
        description="Schema user privileges"

---------------------------------------------------------------------------------------------------------------------
plugins/org.jkiss.dbeaver.tasks.ui.view/plugin.xml
---------------------------------------------------------------------------------------------------------------------
  - line 192:
        <menu label="Project">

---------------------------------------------------------------------------------------------------------------------
plugins/org.jkiss.dbeaver.ui.editors.data/plugin.xml
---------------------------------------------------------------------------------------------------------------------
  - line 838:
        <menu label="Default command">


---------------------------------------------------------------------------------------------------------------------
org.jkiss.dbeaver.ext.generic.model.GenericObjectContainer
---------------------------------------------------------------------------------------------------------------------
  - line 438:
        return getName() == null ? "<NONE>" : getName();

---------------------------------------------------------------------------------------------------------------------
org.jkiss.dbeaver.ext.mysql.edit.MySQLDatabaseManager
---------------------------------------------------------------------------------------------------------------------
  - line 100:
        throw new DBException("Direct database rename is not yet implemented in MySQL. You should use export/import functions for that.");

---------------------------------------------------------------------------------------------------------------------
org.jkiss.dbeaver.ext.mysql.model.plan.MySQLPlanAnalyser
---------------------------------------------------------------------------------------------------------------------
  - line 60:
        throw new DBCException("Only SELECT statements could produce execution plan");

---------------------------------------------------------------------------------------------------------------------
org.jkiss.dbeaver.ext.mysql.ui.config.MySQLIndexConfigurator
[...]

PR #20671, opened 2023-07-25
#18312 Complete some Simplified Chinese localization work

bianyun:
Most of the cases in issue #18312 already have corresponding property files, and the corresponding Simplified Chinese localization work has also been completed. 

There are only one or two places where the corresponding property files are still not available. I will record these places in a new issue. 

If these places are not needed or cannot be localized, please comment them as '//$NON-NLS-1$' to avoid double-checking these places in the future.


E1izabeth (approved):

E1izabeth:
Thank you for contribution!

bianyun:
> Thank you for contribution!

@E1izabeth 
You're welcome. DBeaver is a great database tool and it's an honor to contribute a little bit to it.

MashaKorax:
verified

PR #20695, opened 2023-07-25
#20652 Respect blank line as a statement delimiter in completion

### LINKED ISSUE #20652: Autocompletion doesn't respect blank line as a statement delimiter
### Description

Script example:
```
select a.dh from abonos a,    where
select * from cb.cb_auth_configuration cac, cb.cb_auth_permissions cap
where cac.provider_id = cap.subject_id
select current_timestamp , pg_sleep(3)
select * from actor a

s|
```

No proposals for select keyword in the cursor position (`|`), but if I add `;` after `select * from actor a` it starts working.

### DBeaver Version

DBeaver 23.1.3

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Please check at least case from the issue with and without `Blank line is statement delimiter` checked.


uslss:
verified

PR #20696, opened 2023-07-25
#20654 Don't insert space after table reference completion

### LINKED ISSUE #20654: No space after aliases autocompletion
We shouldn't put whitespace after user accepted autocompletion proposal for some alias. We'd rather put dot there, than whitespace.

E1izabeth:
No space should be inserted after table or alias proposal acception.


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
```diff
@@ -682,6 +682,8 @@ private void makeProposalsFromQueryParts() {
             }
             final DBPPreferenceStore prefStore = request.getContext().getDataSource().getContainer().getPreferenceStore();
             final List<Pair<String, String>> names = extractTableNames(wordPart, true, prefStore);
+            Map<String, Object> parameters = new LinkedHashMap<>();
+            parameters.put(SQLCompletionProposalBase.PARAM_NO_SPACE, true);
```
serge-rider: Use Map.of instead of explicit map creation

serge-rider (changes requested):

uslss:
verified

PR #20710, opened 2023-07-26
dbeaver/pro#1781 Tips shortcuts

E1izabeth:
Now we name the command and tell the shortcut for it instead of just writing the shortcut.
image


MashaKorax:
verified

PR #20716, opened 2023-07-27
dbeaver/pro#1711 Redesign dictionary value viewer

E1izabeth:
Now it looks like this.
image
Also if the max number of pages is known (you are or have been on the last page), it's remembered and shown like page 2/3. We could calculate it and show always, but it requires an additional query. It's not worth it.
image

Please check at least these (separate and in different combinations) in different databases:
- [x] value editing with Value field
- [x] search
- [x] sorting
- [x] pagination


MashaKorax:
verified

PR #20736, opened 2023-07-28
#20347 Redesign network handlers

### LINKED ISSUE #20347: Redesign network handler configuration
We want to improve the user interface in the connection dialog. Right now there are too many tabs for some databases, which can confuse users and does not leave enough space for adding new features.

Screenshot 2023-06-21 at 13 13 23

We decided to do the following:

1. Leave only two tabs by default: Main and Driver Properties
2. Add a tab or button with the **+** sign after them. When users press this button, a dropdown list appears with all available options such as SSL, SSH, etc.
3. When the user selects an option, the corresponding tab appears.

If an additional tab is added, this option is enabled, and this tab appears every time this connection dialog opens. 

In this case, we want to remove the **Use SSH** and similar checkboxes, but we need to decide if we want to keep the SSH and other settings when the user removes the tab and restores them the next time they open it.

ShadelessFox:
|Adding new handlers|No handlers are available|Closing handler tab|
|---|---|---|
|java_6MGJDhxjTx|java_DlKwoiUZtK|image|


E1izabeth:
How to hide tab back?

ShadelessFox:
By unchecking the "Use XXX" checkbox. After the wizard is reopened, the tab will be hidden.

E1izabeth (approved):

E1izabeth (approved):

PR #20743, opened 2023-07-31
dbeaver/dbeaver#19949 reset all settings

### LINKED ISSUE #19949: Reset all settings of all tables for a specific connection
You can now change the view and format of the data tables. We want to add the ability to reset all of these settings for all tables for specific connections. 

We need to figure out where and how to store it all.

serge-rider:


MashaKorax:
verified

PR #20746, opened 2023-07-31
dbeaver/dbeaver#18786 Athena: support multi catalogs

### LINKED ISSUE #18786: MetadataRetrievalMethod=ProxyAPI sending wrong catalog in request to get table metadata
### Description

Issue with getting tables in one table when accessing Athena with MetadataRetrievalMethod=ProxyAPI;
Addition use can remove MetadataRetrievalMethod=ProxyAPI;  they then can see the tables in Account A, but recieves similar glue:GetTables error when running select against a table in Account A that was not displaying when MetadataRetrievalMethod=ProxyAPI; was in the connection string.

Case with AWS resulted in them being able to re-create the issue, and they have determined this is an issue specific to 
It appears that DBeaver is not handling well the situation when there are multiple data sources and sends the wrong catalog in the request to get table metadata.

### DBeaver Version

Community Edition 22.3.1

### Operating System

Microsoft Windows 10

### Database and driver

Database: Athena/Glue
Driver:  2.0.34.10000

### Steps to reproduce

1. Athena Driver has MetadataRetrievalMethod=ProxyAPI set in jdbc connection.
2. Have a role with the ability to connect to multiple databases in multiple accounts.
4. Connect to Athena using that role.
5. See all databases in Account A and Account B
5. Try to expand tables in all of the databases.  
6. On one table (in our situation, first Database in Account A), we receive the error that we do not have access glue:GetTables on a database in Account B... but the database only exists in Account A.

SQL Error [100071] [HY000]: [Simba][AthenaJDBC](100071) An error has been thrown from the AWS Athena client. is not authorized to perform: glue:GetTables
User: arn:aws:sts::Account A:assumed-role/authorizedrole/authorizeduser  is not authorized to perform: glue:GetTables on resource: arn:aws:glue:region:Account B:database/xxxxxx 
because no resource-based policy allows the glue:GetTables action (Service: AmazonDataCatalog; Status Code: 400; Error Code: AccessDeniedException; 
Request ID: xxxx-xxxx-xxxx-xxxx-xxxxx; Proxy: null) [Execution ID not available]

### Additional context

_No response_

serge-rider:


Matvey16:
Verified

PR #20756, opened 2023-08-01
dbeaver/pro#935 add ability to map new data transfer objects names

LonwoLonwo:
Please check the following:

- [x] Saving and executing Task with changed parameters from file
- [x] Saving and executing Task with changed parameters from a database
- [x] Export file with spaces to PostgreSQL
- [x] Export table with spaces to PostgreSQL
- [x] Export file to MySQL
- [x] Export table with spaces to MySQL
- [x] Export file to Oracle/Snowflake (upper case database)
- [x] Export table with spaces to Oracle/Snowflake (upper case database)
- [x] Datasource settings before global
- [x] Settings saving for datasources
- [x] Saving global settings
- [x] Settings saving from mapping dialog
- [x] Data types change with the setting from the Mapping Riles dialog
- [x] Camel case (doesn't work with non-default name case setting)
- [x] Camel case - will be quoted if needed (yes for PostgreSQL, no quotes for MySQL - expected)
- [x] Underscores - case depends on the database - no quotes for the target name (because we do not need them)
- [x] Transforming only works for new table/columns mapped names (with create status) 
- [x] a table can exist, but a column can have the 'create' status - check separately
- [x] The mapping rules button is disabled if no container is chosen on the "Tables mapping" page
- [x] column names with specific characters - must be quoted in the target DDL
- [x] table name with specific characters - must be quoted in the target DDL

2023-08-02 10_14_17-Data Transfer

2023-08-01 12_53_36-Data Transfer

2023-08-01 14_19_09-DBeaver Ultimate 23 1 3 - actorwithspaces2


plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/internal/DTUIMessages.java
```diff
@@ -225,6 +225,18 @@ public class DTUIMessages extends NLS {
 	public static String pref_data_transfer_options_title;
 	public static String pref_data_transfer_options_fallback_directory;
 	public static String pref_data_transfer_options_fallback_directory_tip;
+	public static String pref_data_transfer_mapping_group;
+	public static String pref_data_transfer_info_label_mapping;
+	public static String pref_data_transfer_name_case_label;
+	public static String pref_data_transfer_replacing_combo_label;
+    public static String pref_data_transfer_replacing_combo_tip;
```
E1izabeth, line 232:
```suggestion
        public static String pref_data_transfer_replacing_combo_tip;
```

ShadelessFox:
Checkboxes look really weird. Can you make them not consume all available space and move the label to a separate container so it won't stretch checkboxes too?

LonwoLonwo:
@ShadelessFox combo boxes, maybe?

2023-08-02 10_13_27-Preferences

I changed the data transfer page. And I don't particularly appreciate how it looks changed in the Mapping dialog, so  I left it as it is.

ShadelessFox:
Yes, thank you. I believe we should discuss it.

Matvey16:
There is a bug where global and task settings conflict with each other. I'm still in the process of trying to find exect steps to reproduce but my import task failed in a case when global mapping for spaces in names was set to "don't replace" and in the task it was set to "replace with underscore" and after executing a task table was created with a column named `c_o_l_u_m_n` but inser query used the name `c o l u m n` and failed.

Also I don't think that checking a box to save settings to global everytime is a good idea. That would overwrite global setting when the user don't want to do it sooner or later.

https://github.com/dbeaver/dbeaver/assets/82543000/3728a776-0dc6-46cd-97ce-182de40bb29e

Matvey16:
There is something strange with settings and how they are saved.
Settings can have two different states: with checkbox to save as global checked or not.

If the checkbox is unchecked everything seems to work normally, other changes in settings are reflected in wizrd


https://github.com/dbeaver/dbeaver/assets/82543000/c7bfbcd6-826a-4ed3-8b59-70f6b40fa44a

But if checkbox is checked, they work strange. Every time settings window is opened, all option are reset to a state that is saved somewhere (in cache maybe?) not reflecting the state of the current mapping in wizard. And picking those options becomes impossible


https://github.com/dbeaver/dbeaver/assets/82543000/6d71dd4c-bf9e-46d0-ac70-51a0ff70f769


Another problem is the fact, that changes to that checkbox are not saved if other settings were left the same


https://github.com/dbeaver/dbeaver/assets/82543000/2f0f1139-1147-4f5f-a2cd-88d0325e6c71

LonwoLonwo:
The setting "save current changes to global" now is global.
Other settings from the Mapping Rules dialog will be saved as data source settings if "save current changes to global" is unchecked.

CamelCase + Upper case in the Mapping Rules dialog will not work (how will it work?) - it has a tooltip.

Matvey16:
Verified

PR #20764, opened 2023-08-02
#20725 support YellowBrick procedures recognition

### LINKED ISSUE #20725: Cannot (easily) drop stored procedure in Yellowbrick
### Description

When using the GUI, attempting to right-click a stored procedure and choose "delete" will not work.
The script that is generated behind the scenes is not syntactically correct for yellowbrick.  The generated script uses "DROP FUNCTION..." but Yellowbrick expects "DROP PROCEDURE"

If I proceed with the incorrect script, I get the following error:
**SQL Error [42809]: ERROR: "blah.blah" is a stored procedure
  Hint: Use DROP PROCEDURE to drop stored procedures.**

### DBeaver Version

Community Edition 23.1.3

### Operating System

Windows 10 64 bit 22H2 build 19045

### Database and driver

Yellowbrick 5.2
Latest Postgres JDBC driver

### Steps to reproduce

1. right click any stored procedure in a YB database
2. choose "delete" from the context menu
3. In the "Delete Object" dialog, click the "View Script" button
4. Observe the incorrect syntax

### Additional context
Of course the workaround is to just write and execute a "DROP PROCEDURE" statement myself.

error log

image

LonwoLonwo:
2023-08-02 09_22_37-Delete script

You can check only YellowBrick. No impact on other forks.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
@@ -305,14 +305,18 @@ public class PostgreProcedure extends AbstractProcedure<PostgreDataSource, Postg
                 kind = PostgreProcedureKind.a;
             } else if (isWindow) {
                 kind = PostgreProcedureKind.w;
-            } else if (dataSource.getServerType().supportsProSPColumn()) {
-                if (JDBCUtils.safeGetBoolean(dbResult, "prosp", false)) {
+            } else {
+                boolean isProcedure = false;
+                try {
+                    isProcedure = dbResult.getBoolean("prosp");
+                } catch (SQLException e) {
+                    // Slip then. This column only persist in special cases
+                }
+                if (isProcedure) {
                     kind = PostgreProcedureKind.p;
                 } else {
                     kind = PostgreProcedureKind.f;
                 }
-            } else {
-                kind = PostgreProcedureKind.f;
             }
         }
     }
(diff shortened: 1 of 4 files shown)


E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Verified

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreServerExtension.java
```diff
@@ -104,6 +104,11 @@ public interface PostgreServerExtension {
     String getProceduresSystemTable();
     String getProceduresOidColumn();
 
+    /**
+     * Returns true if a database contains prosp column in the pg_proc system table. This column knows about routine type.
+     */
+    boolean supportsProSPColumn();
```
serge-rider: Lets not add yet another functin to ServerType. We can determine column presence in loadInfo (lookup in resultset metadata)

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
```diff
@@ -306,7 +306,17 @@ private void loadInfo(DBRProgressMonitor monitor, ResultSet dbResult) {
             } else if (isWindow) {
                 kind = PostgreProcedureKind.w;
             } else {
-                kind = PostgreProcedureKind.f;
+                boolean isProcedure = false;
+                try {
+                    isProcedure = dbResult.getBoolean("prosp");
+                } catch (SQLException e) {
+                    // Slip then. This column only persist in special cases
```
ShadelessFox, line 313:
```suggestion
                } catch (SQLException ignored) {
                    // Skip then. This column only persist in special cases
```

Matvey16:
Verified

PR #20770, opened 2023-08-02
#20438 Fix created with cmd folders disappearing on navigator refresh

### LINKED ISSUE #20438: When opening a connection through the cmd, if you place the connection in a folder, it will disappear
### Description

STR:
1) create a new connection using  command line
 `dbeaver-cli -con "driver=sqlite|database=C:\Users\user\AppData\Roaming\DBeaverData\workspace6\.metadata\sample-database-sqlite-1\Chinook.db|name=SQLite_TEST|openConsole=true|folder=SQLite_TEST"`
2) look at Dbeaver UI - folder with the connection disappered

If you create connection without folder parameter - its ok

### DBeaver Version

23.1.1

### Operating System

win 10

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:


LonwoLonwo (approved):

LonwoLonwo:
Code style? :)

E1izabeth:
I'll fix code style, thanks. Haven't noticed

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBPDataSourceFolder.java
```diff
@@ -36,4 +36,14 @@ public interface DBPDataSourceFolder extends DBPNamedObject2 {
     DBPDataSourceRegistry getDataSourceRegistry();
 
     boolean canMoveTo(DBPDataSourceFolder folder);
+
+    /**
+     * Specifies if the folder is not present on the file system, but should not be removed on data sources configuration reload
+     */
+    void setProtected(boolean value);
```
serge-rider: I don't like that we have it in public interface.
We also can calculate this property (all folders which contain provided datasources are protected)

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBPDataSourceFolder.java
E1izabeth: I was surprised, but detecting datasources of the folder is much more obscure

serge-rider (approved):

Issue #19866, opened 2023-05-03
An issue with keeping editors open after reconnecting when switching between simple/advanced view


Essentia-G:
### Description

If I change the connection view from advanced to simple, an opened editor won't reload. The following error message appears:
<img width="1352" alt="Screenshot 2023-05-03 at 18 31 53" src="https://user-images.githubusercontent.com/54938058/235981513-297884cc-5e29-4fd2-97a3-072642cb3657.png">

### DBeaver Version

DBeaver Ultimate Edition 23.0.4

LonwoLonwo:
Thanks for the report @Essentia-G

LonwoLonwo:
We can show here a more user-friendly message instead of the error.

Essentia-G:
Yes, I have the same suggestion

ShadelessFox:
Please see the explanation here: https://github.com/dbeaver/dbeaver/issues/13981#issuecomment-1135687498

arhayka:
Proposed solution:
1. We want to save two node objects identifiers in editors: current node pass + object id. We should try both identifiers on resolve.
2. On resolve try to fix full node pass. On fail, show an error saying that the connection view mode was changed.

PR #20772, opened 2023-08-02
#19866 Find node with missing nodes in-between

ShadelessFox:
Test cases:
- [x] Toggle from Advanced to Simple view (will find node)
- [x] Toggle from Simple to Advanced view (will not find node)

image


E1izabeth:
Why can't we always find the node?

MashaKorax:
Cant toggle from Custom to Advanced


https://github.com/dbeaver/dbeaver/assets/84867187/cf2b5313-fd4c-4a3c-b99c-110de0e19948

ShadelessFox:
@E1izabeth Node path is not unique, it's based on the folder structure in the navigator.

In the simple view, we can have:
```
DataSource1
    Table1
    Table2
    Table3
```

But in the advanced view, we can have:
```
DataSource1
    Database1
        Schema1
            Table1
        Schema2
            Table2
    Database2
        Schema3
            Table3
```

If we have the path `DataSource1/Database1/Schema1/Table1`, it's possible to find `Table1` even if `Database1` and `Schema1` are missing. In the opposite case when we have just `DataSource/Table1`, it's impossible to determine where that `Table1` is stored. It might be `Database1/Schema2`, `Database1/Schema2`, or `Database2/Schema3`. That's why this only works when switching from the advanced view to the simple view.

ShadelessFox:
@MashaKorax custom view's parameters match the advanced view's parameters. This is expected.

MashaKorax:
verified

PR #20779, opened 2023-08-02
#20096 SELECT with WITH clause

### LINKED ISSUE #20096: Table columns do not appear for SELECT with WITH clause
### Description

In Community Edition version 23.0.5, table columns do not appear after entering the table alias and the period character.

image

I am expecting behavior similar to a table column prompt in version 23.0.4

### DBeaver Version

Community Edition 23.0.5

### Operating System

Windows 11 

### Database and driver

SQL Server 2016
Microsoft JDBC Driver for SQL Server (mssql-jdbc-9.2.0.jre8)

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Case from the issue should work right now.
```
SELECT *
FROM dbo.Person PARTITION(xx,yy,zz) p WITH(NOLOCK)
WHERE p.id > 5;
```

Also I added support for mysql partitioning selection - queries having `partition(p0, p1)` after table name in from clause should work now:
```
SELECT store_id
FROM employees PARTITION (p1,p2,p3)
GROUP BY store_id HAVING c > 4;
```


ShadelessFox (approved):

MashaKorax:
verified

PR #20804, opened 2023-08-04
#17919 Add shortcut for creating new objects

### LINKED ISSUE #17919: Shortcut for Adding Column/Field

ShadelessFox:
New shortcut was added: <kbd>Alt+Insert</kbd> (<kbd>Command+N</kbd> on macOS). Can be configured (command name is "Create Object")

https://github.com/dbeaver/dbeaver/assets/35821147/5cec1903-2591-4b8c-91c9-57cbe713ebbe

Acceptance criteria:
- [x] Works when using the shortcut against a folder node
- [x] Works when using the shortcut against an object node
- [x] Works when inside the grid and **does not interfere with the "add row" command that has the same shortcut**
- [x] Vice-versa of the above


uslss:
verified

PR #20829, opened 2023-08-07
CB-3769-user-role-change-event

yagudin10:


plugins/org.jkiss.dbeaver.model.event/src/org/jkiss/dbeaver/model/websocket/event/permissions/WSSubjectPermissionEvent.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.jkiss.code.NotNull;
+import org.jkiss.code.Nullable;
+import org.jkiss.dbeaver.model.websocket.event.WSAbstractEvent;
+import org.jkiss.dbeaver.model.websocket.event.WSEventType;
+
+public class WSSubjectPermissionEvent extends WSAbstractEvent {
+
+    private final String subjectId;
+
```
alexander-skoblikov, line 29: add subject type

alexander-skoblikov (approved):

PR #20840, opened 2023-08-08
dbeaver/pro#1711 Dictionary enhancements

### LINKED ISSUE #17029: Dictionary paging button should disable if there is nothing to load
Navigation arrows buttons should disable if there is no more elements to load in that direсtion.

E1izabeth:
Two queries on opening (loading page with current value):
```
SELECT rental_id dbvrvalue, rental_id FROM public.rental WHERE rental_id >= 3407 ORDER BY rental_id ASC
LIMIT 100
SELECT rental_id dbvrvalue, rental_id FROM public.rental WHERE rental_id < 3407 ORDER BY rental_id DESC
LIMIT 100
```

On pagination:
```
SELECT rental_id dbvrvalue, rental_id FROM public.rental WHERE rental_id >= 3407 ORDER BY rental_id ASC
LIMIT 200 OFFSET 300
```

image

- [x] sorting works correctly
- [x] pagination works correctly
- [x] search works correctly
- [x] it works comparable fast on big tables

For data types that are not string, numeric or datetime only one value should appear in the dictionary panel. Search should work only on equality.

 Please, check at least on string, numeric, date/time/timestamp, uuid. You can find uuid and date fk in PostgreSQL 12: dvdrental.Elizabeth.123table. Numeric fk can be found in the streamings table of the same schema. But it would be nice to check it on some other database too.


serge-rider:
When use search pattern then selected item is ignored (it searches for any random items).
This breaks current behavior.

Also please check attribute type. Range search should work only for numbers, strings and datetimes. For other data types show only current value + search

MashaKorax:
pagination fast-forward makes the arrow inactive

MashaKorax:
verified with an additional ticket  https://github.com/dbeaver/dbeaver/issues/20897

Issue #20591, opened 2023-07-18
The map type of clickhouse is displayed as=? 


guoxiangdaxia:
Windows version. The map type of clickhouse is displayed as=? 
![image](https://github.com/dbeaver/dbeaver/assets/19209779/7b9ceb2d-557c-4412-a5bf-4d1ae6ab1f2c)

LonwoLonwo:
Hello @guoxiangdaxia 

What Clickhouse server and driver version do you use? You can check it in your connection settings with the "Test connection" button.

guoxiangdaxia:
![image](https://github.com/dbeaver/dbeaver/assets/19209779/c8500460-da95-4b72-a9af-97347b49a0e0) this is my version.

LonwoLonwo:
Ok. And what is the expected result for you in this case from main topic?

guoxiangdaxia:
![image](https://github.com/dbeaver/dbeaver/assets/19209779/aaf817b5-8961-473f-a23b-54ec699a5f16)
Hello, the above image shows the display results of my mac bookpro, and the result I want is as shown in the above image. Display in  "key: value" format.

E1izabeth:
Are your drivers on mac and windows the same?

guoxiangdaxia:
![image](https://github.com/dbeaver/dbeaver/assets/19209779/0ab84541-17d4-4cd7-8f9b-b0b242f807d7)
these are Mac's, they are different.
@E1izabeth

ShadelessFox:
Something was changed on the driver side. Looks like we need to handle maps differently.

E1izabeth:
See also #20388

uslss:
fixed in #20797

guoxiangdaxia:
thanks! @uslss @E1izabeth @ShadelessFox @LonwoLonwo

Issue #20388, opened 2023-06-26
Clickhouse Map type handling


Hubbitus:
You apparently support even `Tuple` type, but `Map` shown just as "[Object]".

Please support it in similar to `Tuple` way:
```sql
SELECT transaction_id, AsyncReadCounters
FROM system.query_log
```
![image](https://github.com/dbeaver/dbeaver/assets/720349/7a7a0096-1a23-461b-97fd-0b203aab317a)

That will be nice to see `AsyncReadCounters` in similar form of sub-headers, but with key names provided.

DBeaver version 23.1.0.202306041918 CE.

E1izabeth:
Thank you for the report.

uslss:
fixed in #20797

Hubbitus:
Awesome! Thank you.
Will it be included in next release? Which version?

E1izabeth:
@Hubbitus it's planned for DBeaver 23.1.5

Hubbitus:
```sql
SELECT transaction_id, `Settings` 
FROM system.query_log
```
Will looks like:
![image](https://github.com/dbeaver/dbeaver/assets/720349/af907fa1-9c1c-49b4-a94c-ca0a537375df)

Why form is:
```
[{Key:connect_timeout_with_failover_ms, Value:1000}, {Key:load_balancing, Value:nearest_hostname}, {Key:distributed_aggregation_memory_efficient, Value:1}, {Key:do_not_merge_across_partitions_select_final, Value:1}, {Key:os_thread_priority, Value:2}, {Key:log_queries, Value:1}, {Key:parallel_view_processing, Value:1}, {Key:allow_experimental_live_view, Value:1}, {Key:default_database_engine, Value:Ordinary}]
```
instead of:
```
[{connect_timeout_with_failover_ms:1000}, {load_balancing:nearest_hostname}, {distributed_aggregation_memory_efficient:1}, {do_not_merge_across_partitions_select_final:1}, {os_thread_priority:2}, {log_queries:1}, {parallel_view_processing:1}, {allow_experimental_live_view:1}, {default_database_engine:Ordinary}]
```
?

Looks like repeating `key:` and `value:` just like the garbage. Are you agreed??

E1izabeth:
What are you driver and database version?
I don't see this issue on my Clickhouse connection, but the versions are matter
[...]

PR #20842, opened 2023-08-08
#20797 Support `Array(T)`, `Map(K, V)`, and `Tuple(T...)` types

### LINKED ISSUE #20797: Rendered display of Clickhouse Map Type
It would be supercool if Clickhouse Map Types could be rendered as Array of key-values 

e.g.

[{ukf_src_lab=ZLAB, ukf_src_code=K, ukf_src_display=Kalium, ukf_src_group_display=Klinische Chemie}]
(for this the Preview Panel ist an empty grid - it seem that the entrys could not be parsed correctly)

### This is the expected display   

|idx|name|value|
|-|---|----|
|0|ukf_src_lab|ZLAB|
|1|ukf_src_code|K|
|2|ukf_src_display|Kalium|
|3|ukf_src_group_display|Klinische Chemie|

ShadelessFox:
Please check #20797, #20591, and #20388

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.clickhouse/grammar/ClickhouseDataTypes.g4
+++ b/plugins/org.jkiss.dbeaver.ext.clickhouse/grammar/ClickhouseDataTypes.g4
@@ -52,15 +52,32 @@ Eq: '=';
 String: '\'' (~('\''|'\r'|'\n'))+ '\''; // FIXME: string does not support escape sequences
 
 type
- : simpleType
- | enumType
+ : anyType EOF
+ ;
+
+anyType
+ : enumType
  | tupleType
  | mapType
- | arrayType;
+ | arrayType
+ | simpleType
+ | markerType
+ ;
+
+markerType
+ : 'Nullable' LeftParen anyType RightParen
+ | 'LowCardinality' LeftParen anyType RightParen
+ ;
 
 simpleType
  : intType
+ | floatType
+ | decimalType
  | stringType
+ | uuidType
+ | boolType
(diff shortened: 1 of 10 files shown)


plugins/org.jkiss.dbeaver.ext.clickhouse/grammar/ClickhouseDataTypes.g4
```diff
@@ -17,54 +17,29 @@
...
-//fragment W:[wW];
-//fragment X:[xX];
-//fragment Y:[yY];
-//fragment Z:[zZ];
-
-ENUM_KW: E N U M ('8'|'16');
+ENUM_KW: 'Enum' ('8'|'16');
+MAP_KV: 'Map';
+ARRAY_KV: 'Array';
+TUPLE_KV: 'Tuple';
```
E1izabeth, line 42: As far as I see ENUM, EnUm and enum also works, but here you removed case-insensitivity. I believe it may broke some cases. Why did you do that?

ShadelessFox: Types in Clickhouse are case-sensitive. You can't have an `enUm` type.

E1izabeth: Yes, you're right. I rechecked.

E1izabeth:
Please check array of enums as an example of something more complicated

uslss:
verified

PR #20868, opened 2023-08-10
dbeaver/dbeaver#546 Notification settings

### LINKED ISSUE #546: Additional Sound Notifications
Hey Serge,

I see that in 3.6.7 you added sound notification support "beep after query finish".

Is it possible we can expand on this capability, for example rather than beep after each query could we have a notification sound after the run of all queries within one SQL editor?

Also could there be a different notification sound based on the outcome per query/per SQL script.

See example notification sounds below:
- Success: [http://www.soundsnap.com/node/92951](http://www.soundsnap.com/node/92951)
- Failed: [http://www.soundsnap.com/error_tone_chime_blastwavefx_16379](http://www.soundsnap.com/error_tone_chime_blastwavefx_16379)

This would really help me and hopefully other people when you running many queries, want to monitor but don't always have to be at your desk. 

Thanks,

Ben

ShadelessFox:
image


plugins/org.jkiss.dbeaver.core/plugin.xml
```diff
@@ -1299,4 +1300,13 @@
             name="%org.jkiss.dbeaver.preferences.main.toolbar.customization.name"/>
     </extension>
 
+    <extension point="org.jkiss.dbeaver.notifications">
+        <notification id="commit" title="Transaction commit"/>
+        <notification id="rollback" title="Transaction rollback"/>
+        <notification id="rollback.idle" title="Transaction rollback after long idle period"/>
+        <notification id="reconnect.success" title="Datasource invalidation success"/>
+        <notification id="reconnect.failure" title="Datasource invalidation failure"/>
```
E1izabeth: You were the one who asked me to write datasource as data source :) Shouldn't we do here the same?

plugins/org.jkiss.dbeaver.core/plugin.xml
```diff
@@ -1299,4 +1300,13 @@
...
+        <notification id="agent.notify" title="Agent notification"/>
```
E1izabeth: What's Agent?

E1izabeth:
I think it's better to keep values selector right near the label this way:

image

E1izabeth:
Did we really have popup and sound for filter saving, UUID copying and finishing transformation from source code to sql? It doesn't look like a long operations. Do we need them here?

E1izabeth (approved):

Matvey16:
Notification sound doesn't play if `Show popup` is disabled

Matvey16:
Verified

PR #20881, opened 2023-08-14
dbeaver/dbeaver#20529 add username to enter passphrase dialog

### LINKED ISSUE #20529: SSH connection defaults to OS login, if server login not saved in settings
### Description

It's related to credential popup changes from #18017 - now if SSH is set to use encrypted certificates there will be a popup asking for certificate password, but there's no way to enter server login. If the login is stored in settings - it will work - but in case the login is left empty, the connection will default to OS login.

### DBeaver Version

Community Edition 23.1.2

### Operating System

Windows 10

### Database and driver

_No response_

### Steps to reproduce

1. Configure SSH with jump server (probably without the jump server will be the same)
2. Use encrypted certificate as auth, leave passphrases blank
3. Leave the "user name" blank
4. Try to connect
The connection will ask for passphrase, but not for login and then it will default to using OS user name as SSH server login.

### Additional context

_No response_

Destrolaric:
Closes dbeaver/dbeaver#20529
image


E1izabeth (approved):

Matvey16:
Verified

PR #20882, opened 2023-08-14
dbeaver/pro#1822 Add more keywords to non-reserved list

E1izabeth:
Please try to call your table or columns with these keywords and then autocomplete it.
(You won't be able to call an identifier with this name in some databases, it's okay)
`COMMITTED, REPEATABLE, SERIALIZABLE, TYPE, UNCOMMITTED, CURRENT_USER, SESSION_USER, SYSTEM_USER, USER, VALUE, DATE, YEAR, MONTH, DAY, HOUR, MINUTE, SECOND, ZONE, ACTION, ADD, AUTHORIZATION, BY, CASCADE, CASCADED, CATALOG, COALESCE, COMMIT, CONSTRAINT, CONSTRAINTS, CORRESPONDING, COUNT, DEFERRABLE, DEFERRED, IMMEDIATE, EXTRACT, FULL, GLOBAL, LOCAL, INDICATOR, INITIALLY, INTERVAL, ISOLATION, KEY, LEVEL, NAMES, NO, NULLIF, ONLY, OVERLAPS, PARTIAL, PRESERVE, READ, RESTRICT, ROLLBACK, SCHEMA, SESSION, SET, TEMPORARY, TIME, TIMESTAMP, TIMEZONE_HOUR, TIMEZONE_MINUTE, TRANSACTION, VIEW, WORK, WRITE`


Matvey16:
Verified

PR #20884, opened 2023-08-14
Cb 3814 add ssl tab for all databases that support it

yagudin10:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/net/SSLHandlerTrustStoreImpl.java
```diff
@@ -192,4 +197,23 @@ public static SSLSocketFactory createNonValidatingSslSocketFactory() throws Exce
...
     }
+
+    /**
+     * Reads trust store file contents.
+     */
+    public static byte[] readTrustStoreData(@NotNull DBWHandlerConfiguration configuration, @NotNull String property) throws DBException {
+        var propertyValue = configuration.getSecureProperty(property);
+        if (!CommonUtils.isEmpty(propertyValue)) {
+            try {
+                return Files.readAllBytes(Path.of(property));
```
alexander-skoblikov:
```suggestion
                return Files.readAllBytes(Path.of(propertyValue));
```

PR #20910, opened 2023-08-16
CB-3778-bug-use-oracle-as-internal-db

yagudin10:


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/translate/SQLQueryTranslator.java
```diff
@@ -84,7 +94,10 @@ public static String translateScript(
             sql.append(element.getText());
             sql.append(scriptDelimiter).append("\n");
         }
-        return sql.toString();
+        if (sql.length() > 0) {
```
serge-rider: Why?

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/translate/SQLQueryTranslator.java
```diff
@@ -222,25 +235,64 @@ protected List<? extends SQLScriptElement> translateStatement(
                     }
                 }
             }
-            if (defChanged) {
-                String newQueryText = SQLFormatUtils.formatSQL(null,
+            if (dialectName.equals(DIALECT_NAME_ORACLE) && !CommonUtils.isEmpty(createTable.getIndexes())) {
```
serge-rider: It is a bad idea to use if/else with dialect names. This logic must be implement inside dialect implementation.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/translate/SQLQueryTranslator.java
yagudin10: removed

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/translate/SQLQueryTranslator.java
yagudin10: fixed

serge-rider (approved):

PR #20911, opened 2023-08-16
dbeaver/pro#1823 Custom secure storage

serge-rider:


plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverCommandLine.java
```diff
@@ -232,7 +234,15 @@ static boolean executeCommandLineCommands(@Nullable CommandLine commandLine, @Nu
...
         try {
-            return new DefaultParser().parse(ALL_OPTIONS, Platform.getApplicationArgs(), false);
+            List<String> applicationArgs = Arrays.stream(Platform.getApplicationArgs()).collect(Collectors.toList());
+            int index = applicationArgs.indexOf(CoreApplicationActivator.ARG_ECLIPSE_KEYRING);
+            if (index >= 0) {
+                applicationArgs.remove(index);
+                if (applicationArgs.size() > index) {
+                    applicationArgs.remove(index);
+                }
+            }
```
E1izabeth, line 244: Is two is the max number of eclipse keyring args that could be set? I mean shouldn't we do it in cycle?

serge-rider: It can be specified only once as "-eclipse.keyring file-path"

PR #20915, opened 2023-08-17
CB-3756 add datasource permission updated event

yagudin10:


plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMObjects.java
```diff
@@ -19,6 +19,6 @@
 public interface SMObjects {
     String PROJECT_OBJECT_TYPE_ID = "project";
 
-    SMObjectType DATASOURCE = new SMObjectType("datasource");
-    SMObjectType PROJECT = new SMObjectType(PROJECT_OBJECT_TYPE_ID);
+    SMObjectType DATASOURCE = SMObjectType.datasource;
```
serge-rider: Looks crazy. If we turned SMObjectType into enum then we don't need these constants anymore.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMObjects.java
yagudin10: Fixed. Constants were removed.

PR #20922, opened 2023-08-18
#20921 Fix disappearing controls presence on the map while taking a screenshot

### LINKED ISSUE #20921: Save picture in spatial view does a screenshot which can contain os explorer
### Description

If the OS explorer overlaps the map while save path is being chosen, that exolorer will be present on the final picture

### DBeaver Version

23.1.5

### Operating System

Win11

### Database and driver

Any

### Steps to reproduce

1. Open spatial view for any table
2. Click save as picture
3. Drag the explorer so it overlaps a map
4. Save the picture
5. Open it

**Expected:** The picture contains only the map with data from the table
**Actual:** System explorer is captured before it disappears

### Additional context

nnnn

E1izabeth:


ShadelessFox (approved):

LonwoLonwo (approved):

Matvey16:
Time out is too short, explorer window is still there.

Matvey16:
Verified

PR #20927, opened 2023-08-18
dbeaver/dbeaver#546 Handle missing notifications

### LINKED ISSUE #546: Additional Sound Notifications

ShadelessFox:
Acceptance criteria:
- [x] Scheduling a task should not cause any errors


E1izabeth (approved):

Matvey16:
Verified

PR #20929, opened 2023-08-18
CB-3449. Fixed formatting sql script at macOs/linux.

DenisSinelnikov:


ShadelessFox (commented):
Can you please explain why is this needed?

DenisSinelnikov:
> Can you please explain why is this needed?

Sure, 
It is necessary for correct formatting, on Windows it defaults values for the end of the line \r\n , but if you do it on unix systems they default line ends with a different value, so you have to manually substitute here
https://stackoverflow.com/questions/426397/do-line-endings-differ-between-windows-and-linux

alexander-skoblikov (changes requested):
we should pass correct indent into SQLFormatterConfiguration from CloudBeaver side, do not change core format logic

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/format/SQLFormatUtils.java
```diff
@@ -30,11 +30,21 @@ public class SQLFormatUtils {
     public static String formatSQL(DBPDataSource dataSource, String query) {
         SQLSyntaxManager syntaxManager = new SQLSyntaxManager();
         syntaxManager.init(dataSource.getSQLDialect(), dataSource.getContainer().getPreferenceStore());
-        return formatSQL(dataSource, syntaxManager, query);
+        return formatSQL(dataSource, syntaxManager, query, null);
```
ShadelessFox: You can call the overloaded method here. It's better than duplicating code.

```java
return formatSQL(dataSource, query, null);
```

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/format/SQLFormatUtils.java
```diff
@@ -30,11 +30,21 @@ public class SQLFormatUtils {
...
 
-    public static String formatSQL(@Nullable DBPDataSource dataSource, @NotNull SQLSyntaxManager syntaxManager, String query) {
+    public static String formatSQL(DBPDataSource dataSource, String query, @Nullable String indent) {
+        SQLSyntaxManager syntaxManager = new SQLSyntaxManager();
+        syntaxManager.init(dataSource.getSQLDialect(), dataSource.getContainer().getPreferenceStore());
+        return formatSQL(dataSource, syntaxManager, query, indent);
+    }
+
+    public static String formatSQL(@Nullable DBPDataSource dataSource, @NotNull SQLSyntaxManager syntaxManager, String query,
+                                   @Nullable String indent) {
```
ShadelessFox: Quite a weird formatting. Also, please use nullability annotations whenever possible.

```suggestion
    @NotNull
    public static String formatSQL(
        @Nullable DBPDataSource dataSource,
        @NotNull SQLSyntaxManager syntaxManager,
        @NotNull String query,
        @Nullable String indent
    ) {
```

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/translate/SQLQueryTranslator.java
```diff
@@ -225,7 +225,8 @@ protected List<? extends SQLScriptElement> translateStatement(
             if (defChanged) {
                 String newQueryText = SQLFormatUtils.formatSQL(null,
                     sqlTranslateContext.getSyntaxManager(),
-                    createTable.toString());
+                    createTable.toString(),
+                    null);
```
ShadelessFox: You don't need to call the overloaded method here. It will be called anyways because the original method calls the overloaded.

ShadelessFox (approved):

Issue #18774, opened 2023-01-15
[Export Data] Add two buttons to Skip/Export All columns


ShadelessFox:
### Discussed in https://github.com/orgs/dbeaver/discussions/18768

<div type='discussions-op-text'>

<sup>Originally posted by **PsyLogic** January 13, 2023</sup>
Hello,

Recently I tried to export a table that has over 30 columns and I tried to skip the majority of them but it was a pain to do so.
I hope if there's two buttons "**Skip All**", "**Export All**"  (or maybe a checkbox with the same idea "**Select/Unselecet All**").

Thank you

![image](https://user-images.githubusercontent.com/8070135/212367550-34ba345d-8d18-40c4-978e-a6745646bbf5.png)
</div>

PR #20941, opened 2023-08-21
Add select all and deselect all buttons when exporting data

niushuai1991:
#18774 

printscreen 2023-08-22 014959


niushuai1991:
I only added English and Chinese, because I am not familiar with others.

LonwoLonwo:
Hello @niushuai1991 

Thanks for your contribution.

You helped me to find a bug. In your code and our code, the "Cancel" button will be ignored after clicking. It also works as the "Ok" button.
So, You need to think about the temporary state of actions list.

niushuai1991:
@LonwoLonwo hi, After some tries, I changed the mappings to member variables and it's already working fine.

niushuai1991:
> Hello @niushuai1991
> 
> Thanks for your contribution.
> 
> You helped me to find a bug. In your code and our code, the "Cancel" button will be ignored after clicking. It also works as the "Ok" button. So, You need to think about the temporary state of actions list.

I fixed this bug

niushuai1991:
> Hello @niushuai1991
> 
> Thanks for your contribution.
> 
> You helped me to find a bug. In your code and our code, the "Cancel" button will be ignored after clicking. It also works as the "Ok" button. So, You need to think about the temporary state of actions list.

@LonwoLonwo OK, I fixed this bug.

sbOogway (approved):
clean

plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/pages/stream/StreamConsumerPageSettings.java
```diff
@@ -415,6 +421,35 @@ protected void setValue(Object element, Object value) {
                 column.getColumn().setText(DTUIMessages.stream_consumer_page_mapping_mapping_column_name);
             }
 
+            Composite panel = new Composite(group, SWT.NONE);
+            panel.setLayout(new FillLayout());
+            Button selectAllButton = new Button(panel, SWT.NONE);
```
ShadelessFox: Please consider using `UIUtils.createDialogButton` instead. They have padding and generally look better.

ShadelessFox (changes requested):
There are too many changes IMO.

You can use the `refreshMappings` to acquire a copy of all mappings that you'll operate over in the dialog. After pressing the OK button, you can simply replace old mappings with new mappings.

niushuai1991:
@ShadelessFox hello,Your suggestion is very helpful. I have optimized my code, please review it again.

ShadelessFox (approved):

E1izabeth (approved):

DariaMarkaryan:
Verified

ShadelessFox:
We decided that we wanted to allow changing mappings directly from the tree. Thanks for the contribution, regardless. https://github.com/dbeaver/dbeaver/pull/22409

PR #20943, opened 2023-08-22
dbeaver/pro#1489 don't write bytes for printStreamProgressMonitor

Destrolaric:
Closes dbeaver/pro#1489


plugins/org.jkiss.dbeaver.tasks.native/src/org/jkiss/dbeaver/tasks/nativetool/AbstractNativeToolHandler.java
```diff
@@ -342,7 +343,9 @@ public void runDump() throws IOException {
                         totalBytesDumped += count;
                         long currentTime = System.currentTimeMillis();
                         if (currentTime - prevStatusUpdateTime > 300) {
-                            monitor.subTask(numberFormat.format(totalBytesDumped) + " bytes");
+                            if (!(monitor instanceof PrintStreamProgressMonitor)) {
```
E1izabeth: We also have at least LoggingProgressMonitor. Have you tried to create a task and look at the logs? It may not help in this case.

E1izabeth (approved):

Matvey16:
Verified

PR #20945, opened 2023-08-22
dbeaver/pro#1861 Remove ${password} variable due to security reasons

E1izabeth:
- [x] Please check ${password} variable resolution in connection dialogs, driver properties, sql editor, ... (ideally all places where it might be used) 
- [x] This case should work https://github.com/dbeaver/dbeaver/wiki/Admin-Variables#declare-external-variables-in-a-file


ShadelessFox (approved):

uslss:
Variable starts working if another actual variable was used
1. Paste password variable to database - field - fails with an error
2. Paste some actual variable, e.g. date - works ok
3. paste password variable again - password is shown in error message
https://github.com/dbeaver/dbeaver/assets/31996417/830b1f8a-3495-4402-a1b6-da3c1bf217e9

E1izabeth:
For the first time you tried $[password] with square brackets. Then you used allowed variable. For the third time {password} was resolved assuming that it's a part of a jdbc url pattern - $ remained. I'm sure we should fix it, but it's a separate bug - not about variable ${password}, but about recursive replacing of jdbc template url components. What was actually happened is `jdbc:postgresql://{host}[:{port}]/[{database}]` ->`jdbc:postgresql://myhost[:{port}]/[{database}]` -> `jdbc:postgresql://myhost:{port}/[{database}]` -> `jdbc:postgresql://myhost:5432/[{database}]` -> `jdbc:postgresql://myhost:5432/{database}` -> `jdbc:postgresql://myhost:5432/${password}` -> `jdbc:postgresql://myhost:5432/$bg0xxx`. The last transformation should not be allowed, as I assume we can't use recursive patterns in jdbc url.

E1izabeth:
I'd like to solve this exact issue separately as it's a risky to rewrite jdbc url handling right before the major release and I assume it wouldn't be merged untill freeze. Here's the issue for that https://github.com/dbeaver/pro/issues/1888

E1izabeth:
You could try doing the same with another field which doesn't go to the url - username, for example. Password shouldn't appear there.

uslss:
verified

PR #20948, opened 2023-08-22
dbeaver/pro#1859 Hide shared ssh keys in ui

E1izabeth:
- Inline text area is masked and uneditable
- Widget for text editing doesn't show text, only allows to paste new key and called "Set text" instead of "Edit text"

image


uslss:
verified

PR #20962, opened 2023-08-23
#20800 Allow choosing new data source if can't be found

### LINKED ISSUE #20800: ERD files do not open in DBeaver on other computers
### Description

Even if users have the same database connection, they will not be able to open a custom diagram based on this connection received from another user either in File -> Open file or the Project Explorer (after adding the erd to the C:\Users\user\AppData\Roaming\DBeaverData\workspace6\project\Diagrams folder on Windows).

image

### DBeaver Version

Community Edition 23.1.3

### Operating System

Windows

### Database and driver

Microsoft SQL Server

### Steps to reproduce

1. Create a new connection (SQL Server on db9)
2. Create a new custom diagram in Project Explorer by adding some of the existing tables
3. Share it with someone who has the same connection (SQL Server on db9)
4. Let this person try to open the erd in the ways mentioned above

### Additional context

_No response_

ShadelessFox:
java_88oOabQSrg


Matvey16:
I suggest automatically flagging the diagram as dirty as right now it's impossible to save it if no changes to the diagram itself are made. Because of that the diagram can be closed without saving, so the next time it is opened every connection needs to be manually defined again

LonwoLonwo (approved):

Matvey16:
Verified

PR #20971, opened 2023-08-23
dbeaver/dbeaver#18188 allow multiple certs

### LINKED ISSUE #18188: DBeaver cannot load cert file with multiple certs inside correctly
### Description

Trying to connect MySQL using SSL. On macOS, when I give the system cert bundle `/etc/ssl/cert.pem`, the connection fails and reports `Path does not chain with any trust anchors`. The server uses ISRG X1 CA so definitely it exists in `/etc/ssl/cert.pem`. But when I give the cert with only ISRG X1 CA cert inside, the connection succeed!

Seems DBeaver cannot load cert file with multiple certs inside correctly. It could only handle one file one cert. Multiple certs in a single file is allowed for a long time and most system CA bundles all goes like this. And clients could handle it well.

### DBeaver Version

Community Edition 22.2.4

### Operating System

macOS

### Database and driver

MySQL Connector/J 8.0.29

### Steps to reproduce

_No response_

### Additional context

_No response_

Destrolaric:
Closes dbeaver/dbeaver#18188
Can be tested by passing certificate file containing multiple entries


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/app/DefaultCertificateStorage.java
```diff
@@ -138,14 +135,24 @@ public void addCertificate(@NotNull DBPDataSourceContainer dataSource, @NotNull
...
             if (caCertData != null) {
-                Certificate caCert = cf.generateCertificate(new ByteArrayInputStream(caCertData));
-                keyStore.setCertificateEntry(CA_CERT_ALIAS, caCert);
+                Collection<? extends Certificate> certificates = cf.generateCertificates(new ByteArrayInputStream(
+                    caCertData));
+                int i = 0;
+                for (Certificate certificate : certificates) {
+                    keyStore.setCertificateEntry(i == 0 ? CA_CERT_ALIAS : CA_CERT_ALIAS + i, certificate);
+                    i++;
+                }
```
E1izabeth: Ugh.. how about a traditional for-loop if you need a counter?

Destrolaric: Collection<?> requires additional casts to use counter

E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/app/DefaultCertificateStorage.java
```diff
@@ -138,14 +135,24 @@ public void addCertificate(@NotNull DBPDataSourceContainer dataSource, @NotNull
             CertificateFactory cf = CertificateFactory.getInstance("X.509");
             List<Certificate> certChain = new ArrayList<>();
             if (caCertData != null) {
-                Certificate caCert = cf.generateCertificate(new ByteArrayInputStream(caCertData));
-                keyStore.setCertificateEntry(CA_CERT_ALIAS, caCert);
+                Collection<? extends Certificate> certificates = cf.generateCertificates(new ByteArrayInputStream(
+                    caCertData));
+                int i = 0;
+                for (Certificate certificate : certificates) {
```
serge-rider: Let's use indexed loop instead of separate variable

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/app/DefaultCertificateStorage.java
```diff
@@ -138,14 +135,21 @@ public void addCertificate(@NotNull DBPDataSourceContainer dataSource, @NotNull
             CertificateFactory cf = CertificateFactory.getInstance("X.509");
             List<Certificate> certChain = new ArrayList<>();
             if (caCertData != null) {
-                Certificate caCert = cf.generateCertificate(new ByteArrayInputStream(caCertData));
-                keyStore.setCertificateEntry(CA_CERT_ALIAS, caCert);
+                List<? extends Certificate> certificates =
+                    new ArrayList<>(cf.generateCertificates(new ByteArrayInputStream(caCertData)));
```
ShadelessFox, line 139: Maybe use arrays at this point?

```suggestion
                Certificate[] certificates = cf.generateCertificates(new ByteArrayInputStream(caCertData)).toArray(String[]::new)
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/app/DefaultCertificateStorage.java
```diff
@@ -138,14 +135,21 @@ public void addCertificate(@NotNull DBPDataSourceContainer dataSource, @NotNull
...
+                for (int i = 0; i < certificates.size(); i++) {
+                    keyStore.setCertificateEntry(i == 0 ? CA_CERT_ALIAS : CA_CERT_ALIAS + i, certificates.get(i));
```
ShadelessFox, line 141: Let's extract `i == 0 ? CA_CERT_ALIAS : CA_CERT_ALIAS + i` to a separate utility method so it's not duplicated.

MashaKorax:
verified

PR #20974, opened 2023-08-23
dbeaver/pro#1853 Driver list accessibility support added

serge-rider:


plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/finder/AdvancedList.java
```diff
@@ -457,4 +468,17 @@ private void showItem(AdvancedListItem item) {
         }
     }
 
+    private void initAccessible() {
+        final Accessible accessible = getAccessible();
+        accessible.addAccessibleListener(new AccessibleAdapter() {
+
+            public void getName(AccessibleEvent e) {
```
ShadelessFox, line 475: I believe this method should be annotated with the `@Override` annotation

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/finder/AdvancedList.java
```diff
@@ -403,6 +409,11 @@ void setSelection(AdvancedListItem item) {
         notifyListeners(SWT.Selection, event);
 
         redraw();
+
+        if (isFocusControl()) {
+            //getAccessible().sendEvent(ACC.EVENT_SELECTION_CHANGED, new Object[]{null, item.getData()});
```
ShadelessFox, line 414: There's already lots of commented-out code in our code base :(

uslss:
verified

PR #19808, opened 2023-04-26
#19807 Correctly handle `\r\n` sequence

ShadelessFox:
Test cases:
- [x] `select 'abc' || chr(10) || 'def'`
- [x] `select 'abc' || chr(13) || 'def'`
- [x] `select 'abc' || chr(13) || chr(10) || 'def'`
- [x] `select 'abc' || chr(10) || chr(13) || 'def'`
- [x] Other cases from #19660

The `chr(10)` expression yields `\n`, `chr(13)` yields `\r`

PR #19660, opened 2023-04-10
#19305 Render non-displayable characters in the grid

ShadelessFox:
Demonstration (click to enlarge)
java_Rbe8qn2UI6

The query for testing:
```sql
select CHR(  1) || CHR(  2) || CHR(  3) || CHR(  4) || CHR(  5) || CHR(  6) || CHR(  7) || CHR(  8)
    || CHR(  9) || CHR( 10) || CHR( 11) || CHR( 12) || CHR( 13) || CHR( 14) || CHR( 15) || CHR( 16)
    || CHR( 17) || CHR( 18) || CHR( 19) || CHR( 20) || CHR( 21) || CHR( 22) || CHR( 23) || CHR( 24)
    || CHR( 25) || CHR( 26) || CHR( 27) || CHR( 28) || CHR( 29) || CHR( 30) || CHR( 31) || CHR( 32)
    || CHR( 33) || CHR( 34) || CHR( 35) || CHR( 36) || CHR( 37) || CHR( 38) || CHR( 39) || CHR( 40)
    || CHR( 41) || CHR( 42) || CHR( 43) || CHR( 44) || CHR( 45) || CHR( 46) || CHR( 47) || CHR( 48)
    || CHR( 49) || CHR( 50) || CHR( 51) || CHR( 52) || CHR( 53) || CHR( 54) || CHR( 55) || CHR( 56)
    || CHR( 57) || CHR( 58) || CHR( 59) || CHR( 60) || CHR( 61) || CHR( 62) || CHR( 63) || CHR( 64)
    || CHR( 65) || CHR( 66) || CHR( 67) || CHR( 68) || CHR( 69) || CHR( 70) || CHR( 71) || CHR( 72)
    || CHR( 73) || CHR( 74) || CHR( 75) || CHR( 76) || CHR( 77) || CHR( 78) || CHR( 79) || CHR( 80)
    || CHR( 81) || CHR( 82) || CHR( 83) || CHR( 84) || CHR( 85) || CHR( 86) || CHR( 87) || CHR( 88)
    || CHR( 89) || CHR( 90) || CHR( 91) || CHR( 92) || CHR( 93) || CHR( 94) || CHR( 95) || CHR( 96)
    || CHR( 97) || CHR( 98) || CHR( 99) || CHR(100) || CHR(101) || CHR(102) || CHR(103) || CHR(104)
    || CHR(105) || CHR(106) || CHR(107) || CHR(108) || CHR(109) || CHR(110) || CHR(111) || CHR(112)
[...]

PR #20978, opened 2023-08-23
dbeaver/dbeaver#20957 Contiguous replacement characters rendering fix

### LINKED ISSUE #20957: Two occurrences of the whitespace dots have different meanings
### Description

In the editor, one dot is one whitespace, three whitespaces are three dots, and so on.
In the grid result set, one whitespace dot seems to have a different meaning.
New dots only appear with an odd number of whitespaces. With an even number, only a little more space is added to the right of the dot.

I like the editor variant better because it is easier to count spaces, but in general, I would like it better if both editor and grid result would be consistent and would display the whitespace dots the same way.

### DBeaver Version

CE Version 23.1.5.202308201919

### Operating System

macOS 13.5

### Database and driver

_No response_

### Steps to reproduce

1. Enable showing whitespaces in editor and grid result.
2. Compare whitespace dots in editor and result

```sql
select ' abc'
union
select '  abc'
union
select '   abc'
union
select '    abc'
union
select '     abc'
union
select '      abc'
order by 1 desc;
```
  

dots_editor
dots_grid

### Additional context

_No response_

ShadelessFox:
Please test against cases provided in:
- [x] #20957
- [x] #19808
- [x] #19660

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/GridCellRenderer.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/lightgrid/GridCellRenderer.java
@@ -326,8 +326,9 @@ class GridCellRenderer extends AbstractRenderer {
         int start = 0;
         int index = 0;
 
-        outer:
         while (index < text.length()) {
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

uslss:
verified

PR #20982, opened 2023-08-24
Expose external tables for the newest Greenplum server 7

adam8157:
This partially reverts commit cb7548c45c883a8a077e790d49fea4de9698a814 (https://github.com/dbeaver/dbeaver/pull/20540).

Greenplum 7 turned external tables into foreign tables, but that is only
a low level implementation change, it still provides the legacy external
tables syntax and the `pg_exttable` catalog (via a view instead).

References:
https://github.com/greenplum-db/gpdb/pull/10015
https://github.com/greenplum-db/gpdb/commit/3aad307cb575c6a11a159f599071391b36b21a82

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.greenplum/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.greenplum/plugin.xml
@@ -26,7 +26,7 @@
             <treeInjection path="postgresql/database/schema" after="table">
                 <folder type="org.jkiss.dbeaver.ext.greenplum.model.GreenplumExternalTable"
                         label="%tree.externaltables.node.name"
-                        icon="#folder_table" visibleIf="object.dataSource.supportsExternalTables()">
+                        icon="#folder_table">
                     <items label="%tree.externaltable.node.name" path="externalTable" property="externalTables"
                            icon="#table_external">
                         <folder type="org.jkiss.dbeaver.ext.postgresql.model.PostgreTableColumn"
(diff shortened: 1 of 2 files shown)


LonwoLonwo:
Hello @adam8157 

Ok. I thought that Greenplum was in the process of removing pg_exttable at all.
https://github.com/greenplum-db/gpdb/issues/15950#issuecomment-1628448049
So, that is the reason to keep the "External tables" folder for the newest Greenplum versions?
We have plans to create a special folder for the foreign tables: https://github.com/dbeaver/dbeaver/issues/20988
So, external (foreign) tables will be there in the future.

adam8157:
I can give some input for your decision.

The external table is one kind of foreign table, and the foreign table is one kind of table, but usually, users don't know, they think of them as three kinds of things.

So, having them all in "Foreign tables" is fine. And having "External tables", "Foreign tables" and "Tables" maybe look better, since it's in line with user intuition.

BTW, I would suggest not putting the same table into more than one folder, even theoretically it is like that.

LonwoLonwo:
Ok. Then, I have another solution. https://github.com/dbeaver/dbeaver/pull/20989
We can check the ability to use pg_exttable table/view, and then we can decide - how to check the "is exttable" status. Because without pg_exttable and pg_class.relstorage column - we can not set the "is external" status on tables.

> I would suggest not putting the same table into more than one folder

Yes, we have plans to fix it and to keep tables in different folders without duplicates.

adam8157:
> Ok. Then, I have another solution. #20989 We can check the ability to use pg_exttable table/view, and then we can decide - how to check the "is exttable" status. Because without pg_exttable and pg_class.relstorage column - we can not set the "is external" status on tables.

I like that solution, closing this PR.

PR #20989, opened 2023-08-24
dbeaver/pro#1606 hide External table in Greenplum only we can not rea…

LonwoLonwo:
Hide External table in Greenplum only then we can not read info from pg_exttable, not by the version number condition.

Please check the following:

- [x]  tables reading in Greenplum 6
- [x]  tables reading in Greenplum 7

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumDataSource.java
+++ b/plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumDataSource.java
@@ -54,13 +54,17 @@ public class GreenplumDataSource extends PostgreDataSource {
         super.initialize(monitor);
 
         // Read server version
-        try (JDBCSession session = DBUtils.openMetaSession(monitor, this, "Read Greenplum server version")) {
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumDataSource.java
```diff
@@ -54,14 +54,18 @@ public void initialize(@NotNull DBRProgressMonitor monitor) throws DBException {
...
             if (versionStr != null) {
                 Matcher matcher = Pattern.compile("Greenplum Database ([0-9\\.]+)").matcher(versionStr);
                 if (matcher.find()) {
                     gpVersion = new Version(matcher.group(1));
                 }
             }
+
+            if (hasAccessToExttable == null) {
+                hasAccessToExttable = PostgreUtils.isMetaObjectExists(session, "pg_exttable", "*");
+            }
```
Destrolaric, line 68: `NON-NLS` marking would be helpful

Destrolaric (approved):

E1izabeth (approved):

adam8157 (approved):

uslss:
verified

PR #20993, opened 2023-08-24
dbeaver/cloudbeaver#1951 Handle parsing errors

### LINKED ISSUE #1951: Array fields in Clickhouse cause a crash
Hi,
Started at 23.1.5, doesn't happen in 23.1.4.
When we try to display data that has an array column in Clickhouse, we get an error:

```
The field at path '/result/results[0]/resultSet/id' was declared as a non null type, but the code involved in retrieving data has wrongly returned a null value.  The graphql specification requires that the parent field be set to null, or if that is non nullable that it bubble up null to its parent and so on. The non-nullable type is 'ID' within parent type 'SQLResultSet'
```
or:
```
Cannot invoke "org.jkiss.dbeaver.ext.clickhouse.ClickhouseDataTypesParser$AnyTypeContext.getText()" because the return value of "org.jkiss.dbeaver.ext.clickhouse.ClickhouseDataTypesParser$ArrayTypeContext.anyType()" is null
```

**To Reproduce**
```
create table test(
  `test` Array(Int64)
  )
ENGINE = MergeTree
ORDER BY tuple()
SETTINGS index_granularity = 8192
```

Try to navigate to the data tab
**Screenshots**
image
image
image

ShadelessFox:


E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/ClickhouseTypeParser.java
```diff
@@ -104,6 +104,10 @@ public static DBSDataType getType(@NotNull DBRProgressMonitor monitor, @NotNull
         final var type = parser.type().anyType();
         final DBSDataType resolved;
 
+        if (parser.getNumberOfSyntaxErrors() > 0) {
+            return null;
```
serge-rider, line 109: Let's add some debug logging here (no stacktrace, only message)

serge-rider (changes requested):

uslss:
verified

serge-rider (approved):

PR #21008, opened 2023-08-25
#18892 Use select if the function is not a procedure for debugging

### LINKED ISSUE #18892: Problem with postgres function debug
### Description

During debugging postgres function through dbeaver debugger plugin following message pops out

"ERROR: ... is not a procedure
  Hint: To call a function, use SELECT.
  Position: 6"

### DBeaver Version

Comunity Edition 22.3.3.202301211053

### Operating System

Windows 10 Enterprise

### Database and driver

1. Postgres 13.9
2. jdbc driver 42.5.0

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
You might check PostgreSQL 12` dvdrental.public.film_in_stock` fucntion and `dvdrental.public.display_message` procedure. Both should work now (`film_in_stock` didn't work as we tried to use `CALL `with function instead of `SELECT`).

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql.debug.core/src/org/jkiss/dbeaver/ext/postgresql/debug/internal/impl/PostgreDebugSession.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql.debug.core/src/org/jkiss/dbeaver/ext/postgresql/debug/internal/impl/PostgreDebugSession.java
@@ -35,6 +35,7 @@ import org.jkiss.dbeaver.ext.postgresql.debug.core.PostgreSqlDebugCore;
 import org.jkiss.dbeaver.ext.postgresql.model.PostgreDataSource;
 import org.jkiss.dbeaver.ext.postgresql.model.PostgreDatabase;
 import org.jkiss.dbeaver.ext.postgresql.model.PostgreProcedure;
+import org.jkiss.dbeaver.ext.postgresql.model.PostgreProcedureKind;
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.ext.postgresql.debug.core/src/org/jkiss/dbeaver/ext/postgresql/debug/internal/impl/PostgreDebugSession.java
```diff
@@ -337,13 +338,21 @@ protected IStatus run(DBRProgressMonitor monitor) {
                     JDBCCallableStatement statement = null;
                     try {
                         StringBuilder query = new StringBuilder();
-                        query.append("{ CALL ").append(function.getFullyQualifiedName(DBPEvaluationContext.DML)).append("(");
+                        if (function.getKind().equals(PostgreProcedureKind.p)) {
```
serge-rider: Can we test using getProcedureType()?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql.debug.core/src/org/jkiss/dbeaver/ext/postgresql/debug/internal/impl/PostgreDebugSession.java
E1izabeth: Yes

LonwoLonwo (approved):

MashaKorax:
verified

PR #21017, opened 2023-08-28
#16693 Reset settings on startup

### LINKED ISSUE #16693: Reset UI settings doesn't work on mac
#### System information: 
- Macos
- DBeaver CE 22.1.0

image
  Steps to reproduce:
1. Hide toolbar with window->appearance->Hide toolbar
2. Edit->reset UI settings
Expected: the ui settings rejected
Actual: this action doesn't affet on ui settings and, for example, the toolbar is still hided
4.

ShadelessFox:
Acceptance criteria:
- [x] Works on Windows
- [x] Works on Linux
- [x] Works on macOS
- [x] Works with non-default workspace (using `-data` CLI parameter)

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
+++ b/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
@@ -850,7 +850,7 @@ public class DBeaverApplication extends DesktopApplicationImpl implements DBPApp
             return;
         }
 
-        final Path path = Path.of(instanceLoc.getDataArea("").toURI());
+        final Path path = new File(instanceLoc.getDataArea("").toURI()).toPath();
 
         if (Files.notExists(path) || !Files.isDirectory(path)) {
             return;
--- a/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
+++ b/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
@@ -169,8 +169,6 @@ public class DBeaverApplication extends DesktopApplicationImpl implements DBPApp
     public Object start(IApplicationContext context) {
         instance = this;
 
-        loadStartupActions();
-
         Location instanceLoc = Platform.getInstanceLocation();
 
         CommandLine commandLine = DBeaverCommandLine.getCommandLine();
@@ -228,6 +226,7 @@ public class DBeaverApplication extends DesktopApplicationImpl implements DBPApp
 
         final Runtime runtime = Runtime.getRuntime();
 
+        loadStartupActions();
         initializeConfiguration();
 
         // Debug logger
--- a/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
+++ b/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
@@ -63,16 +63,11 @@ import org.jkiss.utils.IOUtils;
 import org.jkiss.utils.StandardConstants;
 import org.osgi.framework.Version;
 
-import java.io.File;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.io.PrintStream;
+import java.io.*;
 import java.lang.reflect.Field;
 import java.net.URL;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
-import java.nio.file.StandardOpenOption;
+import java.nio.file.*;
+import java.nio.file.attribute.BasicFileAttributes;
 import java.util.*;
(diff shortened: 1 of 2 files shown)


serge-rider:
Did you try to use IStartup instead of our custom hooks?

ShadelessFox:
We need to register these hooks only when we want to.

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/startup/StartupHook.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.ui.app.standalone.startup;
+
+import org.jkiss.code.NotNull;
+
+import java.util.Map;
+
+public interface StartupHook {
+    void handle(@NotNull Map<String, Object> data) throws Exception;
```
serge-rider: Please add javadocs

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -102,7 +103,13 @@ public class DBeaverApplication extends DesktopApplicationImpl implements DBPApp
     private static final String PROP_EXIT_CODE = "eclipse.exitcode"; //$NON-NLS-1$
 
     public static final String DEFAULT_WORKSPACE_FOLDER = "workspace6";
-    
+
+    private static final Path STARTUP_HOOKS_FILE = Path.of("dbeaver-startup-hooks.json");
```
serge-rider: Why it is static and in unknown location?
Let's keep it in workspace metadata folder and resolve on demand.

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -102,7 +103,13 @@ public class DBeaverApplication extends DesktopApplicationImpl implements DBPApp
...
+    private static final Gson GSON = new GsonBuilder()
+        .enableComplexMapKeySerialization()
+        .serializeNulls()
+        .create();
+
```
serge-rider, line 106: Let's not use gson parser. Simple properties files is enough.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
ShadelessFox: Not enough for the payload. It may contain arbitrary data.

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
ShadelessFox: It doesn't point to any particular path. It's just a file name. Maybe it would be clearer if it was `String`, not `Path`.

It's located here because that file's load/store logic is handled in that enclosing class, nowhere else.

I still think that exposing implementation-specific details is a bad idea.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -168,6 +169,8 @@ public long getLastUserActivityTime() {
     public Object start(IApplicationContext context) {
         instance = this;
 
+        loadStartupActions();
```
serge-rider: Please test with non-default workspace. `-data path`

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -780,18 +785,125 @@ public void notifyVersionUpgrade(@NotNull Version currentVersion, @NotNull Versi
...
+
+    private void processStartupActions(@NotNull Location instanceLoc, @NotNull Set<String> actions) throws Exception {
+        final boolean resetUserPreferences = actions.contains(RESET_USER_PREFERENCES);
+        final boolean resetWorkspaceConfiguration = actions.contains(RESET_WORKSPACE_CONFIGURATION);
+
+        if (!resetUserPreferences && !resetWorkspaceConfiguration || !instanceLoc.isSet()) {
+            return;
+        }
+
+        final Path path = Path.of(instanceLoc.getDataArea("").toURI());
```
E1izabeth: You probably don't want to use toURI(), but I'm not 100% sure. @Destrolaric is an expert in this now.

E1izabeth (approved):

Matvey16:
Verified

serge-rider (approved):

PR #21029, opened 2023-08-29
dbeaver/pro#1888 Remove password template replacing for jdbc url

E1izabeth:
- [x] The bug from the issue should be fixed.
- [x] URL parsing for new connection creating should work with such cases: `jdbc:postgresql://usr:pwd@192.168.42.6:5432/e1izabethdb`
image


LonwoLonwo (approved):

uslss:
verified

PR #21031, opened 2023-08-29
#20869 Limit maximum size of the text widget

### LINKED ISSUE #20869: the error display panel in the results panel isn't scrollable.
### Description

When dbeaver encounters an error running code in a SQL Editor script window, it _usually_ displays the error message in a pane on the left of the results tab.

Unfortunately, there is no way to scroll that text area and so it's impossible to see the entire error message.

This is how it looks at it's default size:
dbeaver_standard_size_results_error_20230810_140504

and here's what it looks like if I make the results panel take up the entire screen, heightwise:
dbeaver_full_screen_results_panel_error_20230810_140548

As you can see, without some means to scroll, there's practically no way to see the entire error message.

Can the error message area be made scrollable?

Thanks.

### DBeaver Version

Community Edition 23.1.4

### Operating System

Windows 10

### Database and driver

_No response_

### Steps to reproduce

write some code that will throw a large detailed error
run that code
watch as it's impossible to read the entire error message.

### Additional context

_No response_

ShadelessFox:


E1izabeth (approved):

LonwoLonwo:
Code style?

E1izabeth:
The whole document is formatted with tabs. I'd better left it as is.

ShadelessFox:
It will conflict with #21032.

Matvey16:
Verified

PR #21033, opened 2023-08-29
dbeaver/pro#1901 update navigator tree elements accessibility reading labels

LonwoLonwo:
Database Navigator elements:
Tables
Folders
Views
Connections
Triggers
Procedures
Sequences
Catalogs
Columns
Indexes
Keys
Projects
Scripts
ERD files
Bookmarks


plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/NavigatorViewBase.java
```diff
@@ -345,4 +356,73 @@ protected void redrawTree() {
...
+        Accessible accessible = tree.getAccessible();
+        accessible.addAccessibleListener(new AccessibleAdapter() {
+            @Override
+            public void getName(AccessibleEvent e) {
+                TreeViewer viewer = navigatorTree.getViewer();
+                IStructuredSelection selection = (IStructuredSelection) viewer.getSelection();
+                if (selection.size() > 1) {
+                    return;
+                }
+                Object firstElement = selection.getFirstElement();
```
serge-rider, line 372: Why can't we use information from DBXTreeItem. We have item time (even plural) for every DBNDatabaseNode.
It doesn't make sense to use else/if and additional localizations

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/NavigatorViewBase.java
LonwoLonwo: Ok, changed

MashaKorax:
Still incorrect behavior

PR #21067, opened 2023-08-31
dbeaver/dbeaver#20513 Fix uuid value manager for panel

### LINKED ISSUE #20513: UUID Column in MongoDB displays "Binary" instead of string
### Description

As can we seen [here](https://github.com/orgs/dbeaver/discussions/20154#discussioncomment-6087756) there is some conditions in which a database in mongo setup to have ID as UUID is showing only "Binary" instead of the UUID output from the database

image
image

### DBeaver Version

DBeaver Enterprise 23.0.0

### Operating System

Windows 11

### Database and driver

1. org.jkiss.jdbc.mongodb.mongodbDriver (cql - Driver for MongoDB)

### Steps to reproduce

1. Create a collection on mongo
2. Add a document with the UUID as ID
3. Connect on Dbeaver PRO
4. Execute a SELECT * FROM _path_

### Additional context

_No response_

E1izabeth:


uslss:
incorrect id is shown as a header in JSON view
image

uslss:
verified

PR #21090, opened 2023-09-01
dbeaver/dbeaver#18965 use old timezone during connect

### LINKED ISSUE #18965: Change timezone from Europe/Kiev to Europe/Kyiv
### Description

timedatectl have recently changed Europe/Kiev timezone to Europe/Kyiv. As a result dbeaver doesnt work with this timezone. Please, change timezone or add new(for compatibility, some users still have Europe/Kiev timezone).
image

### DBeaver Version

Community edition 22.3.4

### Operating System

ubuntu 20.04

### Database and driver

not related

### Steps to reproduce

1. Change timezone to Europe/Kyiv
2. Open dbeaver
3. get error 
image

### Additional context

_No response_

Destrolaric:
Closes dbeaver/dbeaver#18965
Uses old timezone during connection, after connection succeded returns to the new one.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
@@ -730,11 +730,6 @@ public class PostgreDataSource extends JDBCDataSource implements DBSInstanceCont
         }
     }
 
-    @Override
-    public void shutdown(DBRProgressMonitor monitor) {
-        super.shutdown(monitor);
-    }
-
     public PostgreServerExtension getServerType() {
         if (serverExtension == null) {
             PostgreServerType serverType = PostgreUtils.getServerType(getContainer().getDriver());
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
@@ -98,8 +98,8 @@ public class PostgreDataSource extends JDBCDataSource implements DBSInstanceCont
     private String serverVersion;
     private volatile boolean hasStatistics;
 
-    private static final String LEGACY_TIMEZONE = "Europe/Kiev";
-    private static final String NEW_TIMEZONE = "Europe/Kyiv";
+    private static final String LEGACY_UA_TIMEZONE = "Europe/Kiev";
+    private static final String NEW_UA_TIMEZONE = "Europe/Kyiv";
 
     public PostgreDataSource(DBRProgressMonitor monitor, DBPDataSourceContainer container)
         throws DBException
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -497,7 +499,15 @@ protected Connection openConnection(@NotNull DBRProgressMonitor monitor, @Nullab
         if (instance != null) {
             log.debug("Initiate connection to " + getServerType().getServerTypeName() + " database [" + instance.getName() + "@" + conConfig.getHostName() + "] for " + purpose);
         }
+        boolean timezoneOverridden = false;
         try {
+            // Old versions of postgres and some linux distributions, on which docker images are made, may not contain
+            // new timezone, which will lead to the error while connecting, there is no way to know before connecting
+            // so to be sure we will use the old name
+            if ("Europe/Kyiv".equals(TimeZone.getDefault().getID())) { //$NON-NLS-1$
```
serge-rider: Please move it to constant

serge-rider: What if new PG server version will have a fix for this timezone?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
Destrolaric: > What if new PG server version will have a fix for this timezone?

Even if the new version will contain a new timezone, it still requires Linux distributions on which it is installed to have this timezone as well. We can't just be sure based on postgres version.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -94,9 +96,11 @@ public class PostgreDataSource extends JDBCDataSource implements DBSInstanceCont
     private String activeDatabaseName;
     private PostgreServerExtension serverExtension;
     private String serverVersion;
-
     private volatile boolean hasStatistics;
 
+    private static final String LEGACY_TIMEZONE = "Europe/Kiev";
```
LonwoLonwo:
```suggestion
    private static final String LEGACY_UA_TIMEZONE = "Europe/Kiev";
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -94,9 +96,11 @@ public class PostgreDataSource extends JDBCDataSource implements DBSInstanceCont
...
+    private static final String NEW_TIMEZONE = "Europe/Kyiv";
```
LonwoLonwo:
```suggestion
    private static final String NEW_UA_TIMEZONE = "Europe/Kyiv";
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
LonwoLonwo: `LEGACY_TIMEZONE` is too abstract name.

serge-rider:
This fix means that we will use old timezone always.
What happens in new PostgreSQL server version? Does it understand both old and new timezone?

MashaKorax:
verified

PR #21118, opened 2023-09-04
#21114 Handle cases with null key value

### LINKED ISSUE #21114: Dictionary associated through foreign key is shown empty for new record
### Description

When I select the value in a column that has a foreign key to another table it shows the list of values from the dictionary in the "Value" panel. Recently I have noticed that this panel is empty for new records created with the "+" button. 
When I type some random value in the column, then click another column and again click the dictionary column the "Value" panel is reloaded and shows the list of entries correctly.

### DBeaver Version

Community 23.2.0.202309031556

### Operating System

Windows 11

### Database and driver

### Steps to reproduce

1. Add new (empty) record in some table
2. Click column which has the FK to some other table.
3. Check the contents of the "Value" panel (is empty but should be filled with the record from related dictionary table)

### Additional context

_No response_

E1izabeth:


Matvey16:
Verified

PR #21130, opened 2023-09-05
Dbeaver/dbeaver#20975 validate parameter values

### LINKED ISSUE #20975: Validate parameter values
### Description

Currently we pass parameters of the procedure as is without any validation. It leads to the following cases:
- if user forgot to set value the debugging doesn't work for the Unable to rcv port number reason
- if user put invalid argument (for example, pass string as `asdf` instead of `'asdf'`, the debugging doesn't work for the same reason

We should make it more user-friendly.

### DBeaver Version

DBeaver 23.1.5

### Operating System

Windows 11

### Database and driver

PostgreSQL 12

E1izabeth:
image
image


LonwoLonwo (approved):

MashaKorax:
verified

ShadelessFox (approved):

ShadelessFox (approved):

priiduneemre:
@E1izabeth: Hi, I hope everything is great! 
Just happened to notice this PR breaks the PostgreSQL debugger for some common usecases. 

Examples: 
* Omitting an optional procedure parameter
* Passing the `NULL` value for `boolean` or `numeric` data types
* Passing other non-literal PostgreSQL expressions as arguments

I was able to work around the issue by reverting this PR locally and building from source. 
Maybe it makes sense to skip these checks altogether?

E1izabeth:
Please create a new issue and describe your case so we can reproduce it.

PR #21134, opened 2023-09-06
dbeaver/dbeaver#15951 UI unresponsive while loading a larger JSONB

### LINKED ISSUE #15951: UI unresponsive while loading a larger JSONB object
Hi,

I've just noticed that if I load larger JSONB objects from PostgreSQL, the DBeaver UI freezes and becomes completely unresponsive while trying to load the content in the view panel. The unresponsive part is that bothers me there as I have to force quite the application leaving all unsaved work.

image

serjiokov:
#15951 UI unresponsive while loading a larger JSONB

Provided functionality for restriction of large file content representation.


E1izabeth:
Please use 4 spaces instead of tabs for indentation

E1izabeth:
Could you also provide a screenshot of how does it look like?

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -271,42 +289,66 @@ public <T> T getAdapter(Class<T> adapter) {
...
-            monitor.done();
-        }
+    	try {
+			int maxContentSize = DBWorkbench.getPlatform().getPreferenceStore().getInt(ModelPreferences.EDITING_CONTENT_MAX_SIZE_KBYTES)*1000;
+			if (editor == null) {
+				log.error("Editor is null or undefined");
+				return;
+			}
+			resetEditorInput();
+			byte[] bytes = CommonUtils.toString(value).getBytes();
```
E1izabeth: Why don't you use getContentLength method?

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -271,42 +289,66 @@ public <T> T getAdapter(Class<T> adapter) {
...
+		UIUtils.asyncExec(() -> {
+			if (editor != null) {
+				editorPaintListener.updateStatusMessage("");
+				editor.setInput(new StringEditorInput("Empty", "", true, StandardCharsets.UTF_8.name()));
+			}
+		});
+	}
+
+	private void showRegularContent(final PanelEditorPaintListener editorPaintListener, byte[] bytes,
+			DBRProgressMonitor monitor) throws DBException {
```
E1izabeth: We use such formatting for a long lines
```suggestion
	private void showRegularContent(
	    @NotNull final PanelEditorPaintListener editorPaintListener,
	    byte[] bytes,
	    @NotNull DBRProgressMonitor monitor
	) throws DBException {
```

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: The main reason is in next: funtion "getContentLength" packed to ContentEditorInput here such object is unaccessable, (another logic branch is for the case of limited content Im using StringEditorInput object) that is why using getBytes is applicable here.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Let's have agrement of code style, Im currently  using default eclipse settings.

serjiokov:
The current behavior presented on the screen:
image

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
E1izabeth: We have it already.

Such formatting for long lines
Long line is 140 symbols and more
Formatting for casts is the following `(ClassName) varName` - space in between
Arrays formatting: `new String[]{"*.*", "*.jks", "*"}`
Don't use static imports
Also note that we use nullable annotations.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
E1izabeth: I referred to `DBDContent`'s method as value is a `DBDContent`

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
E1izabeth: Also we don't write if or cycles without blocks. For example:
```
if (true) System.out.println("Hey!");
```
should be 
```
if (true) {
    System.out.println("Hey!");
}
```

serjiokov: It is clear, same for each IDE,
one more point, the max line width is 140  instead of 120 (120 by defaul) is it correct?

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -269,44 +287,71 @@ public <T> T getAdapter(Class<T> adapter) {
...
-            final TextViewer textViewer = editor.getTextViewer();
-            final String encoding = getPanelSettings().get(PREF_TEXT_EDITOR_ENCODING);
-            final ContentEditorInput textInput = new ContentEditorInput(valueController, null, null, encoding, monitor);
-            boolean longContent = textInput.getContentLength() > LONG_CONTENT_LENGTH;
-            if (longContent) {
-                UIUtils.asyncExec(() -> {
-                    editor.setInput(new StringEditorInput("Empty", "", true, StandardCharsets.UTF_8.name()));
-                });
+            int maxContentSize = DBWorkbench.getPlatform().getPreferenceStore()
+                    .getInt(ModelPreferences.EDITING_CONTENT_MAX_SIZE_KBYTES) * 1000;
```
E1izabeth: Also we don't use 8 spaces for indentation in such places. Always 4 spaces :)
```suggestion
            int maxContentSize = DBWorkbench.getPlatform().getPreferenceStore()
                .getInt(ModelPreferences.EDITING_CONTENT_MAX_SIZE_KBYTES) * 1000;
```

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: ok, will do.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Already reformatted

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -1512,6 +1512,26 @@ public static void drawMessageOverControl(Control control, GC gc, String message
...
     }
 
+    /**
+     * Method designed to draw message on top of control with left bottom
+     * positioning
+     * @param control - ui control
+     * @param gc      - graphical container
+     * @param message - custom message text
+     */
+    public static void drawMessageOverControlOnLeftBottom(Control control, GC gc, String message) {
```
serge-rider: Let's leave this method in panel editor. UIUtils is already too big and this implementation looks quite specific.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -269,44 +287,69 @@ public <T> T getAdapter(Class<T> adapter) {
...
+                log.error("Editor is null or undefined");
+                return;
+            }
+            resetEditorInput();
+            if (value.getContentLength() > maxContentSize) {
+                showRestrictedContent(editorPaintListener, value);
+            } else {
+                showRegularContent(monitor);
             }
-            UIUtils.asyncExec(() -> {
```
serge-rider, line 285: Why default progress message was removed? It still makes sense I think.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: We added content size restriction, the large values do not take much loading time now.

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
serjiokov: The drawing method content moved to PanelEditorPaintListener.

MashaKorax:
If the value of **Maximum size of editing content** was set =< two-digit number (xx)---> after that it is impossible to change it to more than two digits. The same with 1 digit.

https://github.com/dbeaver/dbeaver/assets/84867187/1ab27cb7-147d-403f-b792-11465692aaa6

serjiokov:
image
Aligned ui components

E1izabeth (approved):

LonwoLonwo (approved):

MashaKorax:
additional tiket https://github.com/dbeaver/pro/issues/1952

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -107,7 +118,15 @@ public StyledText createControl(IValueController valueController) {
         initEditorSettings(editorControl);
 
         editor.addContextMenuContributor(manager -> contributeTextEditorActions(manager, editorControl));
+        editorPaintListener = new PanelEditorPaintListener();
+        editorControl.addPaintListener(editorPaintListener);
+        editorControl.addListener(SWT.RESIZE, new Listener() {
```
ShadelessFox: Wrong constant

```suggestion
        editorControl.addListener(SWT.Resize, new Listener() {
```

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -451,4 +497,34 @@ public String getEncoding() {
...
+                for (String line : statusMessage.split("\n")) {
+                    line = line.trim();
+                    Point ext = gc.textExtent(line);
+                    int x = bounds.x;
+                    int y = bounds.y + bounds.height - ext.y - height;
+                    gc.drawText(line, x, y);
+                }
+            }
+        }
+    }
```
ShadelessFox: Why do we need to paint text over a control? Can't we just create a new label control?

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/preferences/PrefPageResultSetMain.java
```diff
@@ -123,127 +127,149 @@ protected boolean supportsDataSourceSpecificOptions()
...
-                @Override
-                public void focusLost(FocusEvent e) {
-                    int newValue = CommonUtils.toInt(resultSetSize.getText());
-                    if (newValue > 0 && newValue < ResultSetPreferences.MIN_SEGMENT_SIZE) {
-                        resultSetSize.setText(String.valueOf(ResultSetPreferences.MIN_SEGMENT_SIZE));
-                    }
+        rightPane.setLayoutData(new GridData(GridData.FILL_BOTH));
+
+        // Group of queries settings
+        Group queriesGroup = UIUtils.createControlGroup(leftPane,
```
ShadelessFox: Please bring back {}. They denote a scope. You don't need the `queriesGroup` variable outside. Same for blocks down below.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/preferences/PrefPageResultSetMain.java
```diff
@@ -293,7 +319,9 @@ protected void loadPreferences(DBPPreferenceStore store)
             }
             showErrorsInDialog.setSelection(store.getBoolean(ResultSetPreferences.RESULT_SET_SHOW_ERRORS_IN_DIALOG));
             markCellValueOccurrences.setSelection(store.getBoolean(ResultSetPreferences.RESULT_SET_MARK_CELL_VALUE_OCCURRENCES));
-
+            int maxEditiingContentSize = store.getInt(ModelPreferences.EDITING_CONTENT_MAX_SIZE_KBYTES);
+            maxEditingContentSize.setText(String.valueOf(maxEditiingContentSize));
```
ShadelessFox: Redundant local variable

```suggestion
maxEditingContentSize.setText(String.valueOf(store.getInt(ModelPreferences.EDITING_CONTENT_MAX_SIZE_KBYTES)));
```

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov:
```
/**
	 * Style constant for resize box trim (value is 1&lt;&lt;4).
	 * <p><b>Used By:</b></p>
	 * <ul>
	 * <li><code>Decorations</code> and subclasses</li>
	 * <li><code>Tracker</code></li>
	 * </ul>
	 */
	public static final int RESIZE = 1 << 4;
```

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: No, it is drawn directly on canvas, here require a specific text as a message. The label require composite.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/preferences/PrefPageResultSetMain.java
serjiokov: {} - is redundunt and does not fit the Java Code Style convention.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/preferences/PrefPageResultSetMain.java
serjiokov: The variable extracted for clarity.

serjiokov:
Rework to message label panel text in place of editor.
Screenshot 2023-09-19 003745

serjiokov:
New group added for detup represention content on prefernces page.
Screenshot 2023-09-19 014610

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/internal/ResultSetMessages.properties
```diff
@@ -245,6 +245,7 @@ pref_page_content_editor_checkbox_use_navigator_filters_tip = Filter (hide) colu
 pref_page_content_editor_group_content = LOB
 pref_page_content_editor_group_hex = Binary editor
 pref_page_content_editor_group_editing = Behavior on missing key
+pref_page_content_editor_group_representation_content = Representation content
```
LonwoLonwo: I'm not a native speaker, but I think `Content representation` sounds better. What do you think?

serjiokov: I think so, thanks.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -101,16 +115,44 @@ public StyledText createControl(IValueController valueController) {
...
-
         editor.addContextMenuContributor(manager -> contributeTextEditorActions(manager, editorControl));
-
+        messageBar = createMessageBar(cmpsBase);
         return editorControl;
     }
 
+    private void reasignLayout(Composite cmpsInternalBase) {
+        Control[] children = cmpsInternalBase.getChildren();
+        Stream.of(children).forEach(c -> c.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true, 1, 1)));
```
ShadelessFox, line 141: Just use plain foreach loop, no need to use streams here. You can also reuse the created layout.

Also, I'm not even sure why this is needed. You can nest several composites to avoid changing child layout data. It looks hacky.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -101,16 +115,44 @@ public StyledText createControl(IValueController valueController) {
...
+    }
+
+    private CLabel createMessageBar(Composite cmpsBase) {
+        UIUtils.createHorizontalLine(cmpsBase);
+        messageBar = new CLabel(cmpsBase, SWT.RIGHT);
+        GridData layoutData = new GridData(SWT.FILL, SWT.FILL, true, false, 1, 1);
```
ShadelessFox: Redundant local variable

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -269,28 +311,64 @@ public <T> T getAdapter(Class<T> adapter) {
     }
 
     @Override
-    public void primeEditorValue(@NotNull DBRProgressMonitor monitor, @NotNull StyledText control, @Nullable DBDContent value) throws DBException
-    {
+    public void primeEditorValue(@NotNull DBRProgressMonitor monitor, @NotNull StyledText control, @Nullable DBDContent value)
+            throws DBException {
```
ShadelessFox:
```suggestion
    public void primeEditorValue(
        @NotNull DBRProgressMonitor monitor,
        @NotNull StyledText control,
        @Nullable DBDContent value
    ) throws DBException {
```

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -299,12 +377,25 @@ public void primeEditorValue(@NotNull DBRProgressMonitor monitor, @NotNull Style
...
-        }
+            }
+        });
+    }
+
+    private void showRestrictedContent(@NotNull DBDContent value) {
+        int maxContentSize = DBWorkbench.getPlatform().getPreferenceStore()
+            .getInt(ModelPreferences.REPRESENTATION_CONTENT_MAX_SIZE_KBYTES);
+        byte[] bytes = CommonUtils.toString(value).getBytes();
+        byte[] restrictedBytes = Arrays.copyOfRange(bytes, 0, maxContentSize);
```
ShadelessFox: You should use `Math.min(bytes.length, maxContentSize)` to avoid having zero bytes at the end if the buffer is smaller than `maxContentSize`.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -451,4 +542,34 @@ public String getEncoding() {
             return encoding;
         }
     }
+    
+    class PanelEditorPaintListener implements PaintListener {
```
ShadelessFox: Unused code.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: So, the nested ui controls out of controls and layouting here, the default class is pure Eclipse implementaion that hided in abstract root class with Stack layout, whitout this approch Label looks ugly.

serjiokov: ...mmm not get you right about streams, whats is you consern?

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Redundant style - supported only by IDEA. Jenkins style check past. No issue here.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: What issues with zero at the end? The restricted content is cutting and is in read only mode, auto formating aslo does not work for this case.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Unused code removed.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Meanwhile All checks have passed
4 successful checks

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Reorganized

ShadelessFox (commented):
Please see my other comments

serjiokov:
Please argue requested style  change if all checks have been passed (4 successful checks)
Please respond on my comments.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
ShadelessFox: I've explained my concern about streams in another comment.

Regarding layout, you can wrap the editor's component in another composite and change its layout instead.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
ShadelessFox: Then use IDEA instead. We use this style of wrapping for long methods.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
ShadelessFox: If the provided content is shorter than the limit, the resulting buffer will be filled with zero bytes to be of size `maxContentSize`.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Meanwhile All checks have passed
4 successful checks

ShadelessFox: It's not about checks - they are dumb anyway - it's about human readability.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Should not happen because of 
```
if (value.getContentLength() > maxContentSize) {
                showRestrictedContent(value);
            } else {
                showRegularContent(monitor);
            }
```

ShadelessFox: But it's happening. Also content conversion is off in SQLite (it should display trimmed data, not the displayed placeholder):

|Without my suggestion|With my suggestion|
|---|---|
|java_mpO8C4cO13|ZZPAk7FhiL|

serjiokov: Content restricted by 1000 KB, but the right part looks a 100 bytes, not sure is it correct.

ShadelessFox: You're basically doing `content.toString()`, which is incorrect. You should use `DBDContent#getContents` instead.

ShadelessFox: This is a bug, and I won't pass this PR to the QA team until it's fixed.

E1izabeth: Let's better use DBDContent#getContents. No need to call toString when we have API for that needs.

E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Inputing anything feels very odd, because any number inputed immidiately turns to `100` and other input messes up.


https://github.com/dbeaver/dbeaver/assets/82543000/9f6e48b6-d890-4ab5-934d-9a7a26631fd5

E1izabeth (approved):

Matvey16:
Verified

serge-rider:
What confuses me:
- New option on main data editor pref page. It should be on Apperance or on Binary Editor page
- Existing option "Maximum text editor content length" look very similar. And it also looks like we don't use it anymore. Let's delete it?

serjiokov:
I can propose reuse already existing pref option just rename a group name:
image
PLS give your feedback

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/preferences/PrefPageResultSetMain.java
```diff
@@ -214,36 +215,25 @@ public void widgetSelected(SelectionEvent e) {
...
+                ResultSetMessages.pref_page_content_editor_group_editing, 1, GridData.VERTICAL_ALIGN_BEGINNING, 0);
+        group.setLayout(editingGroupLayout);
+
+        alwaysUseAllColumns = UIUtils.createCheckbox(group,
+                ResultSetMessages.pref_page_content_editor_checkbox_keys_always_use_all_columns, false);
+        alwaysUseAllColumns.setLayoutData(new GridData(GridData.FILL, GridData.FILL, true, true, 2, 1));
+
+        disableEditingOnMissingKey = UIUtils.createCheckbox(group,
+                ResultSetMessages.pref_page_content_editor_checkbox_disable_editing_if_key_missing, false);
+        disableEditingOnMissingKey.setLayoutData(new GridData(GridData.FILL, GridData.FILL, true, true, 2, 1));
```
ShadelessFox: Please revert this redundant change

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -299,11 +377,27 @@ public void primeEditorValue(@NotNull DBRProgressMonitor monitor, @NotNull Style
...
+        });
+    }
+
+    private void showRestrictedContent(@NotNull DBDContent value) throws DBCException, IOException {
+        int maxContentSize = DBWorkbench.getPlatform().getPreferenceStore().getInt(ResultSetPreferences.RS_EDIT_MAX_TEXT_SIZE);
+        DBDContentStorage contents = value.getContents(new VoidProgressMonitor());
+        try (final InputStream stream = contents.getContentStream()) {
+            ByteArrayOutputStream buffer = new ByteArrayOutputStream((int) contents.getContentLength());
+            IOUtils.copyStream(stream, buffer);
+            byte[] displaingContentBytes = Arrays.copyOfRange(buffer.toByteArray(), 0, maxContentSize * 1000);
```
ShadelessFox: Can we read `maxContentSize` instead of the entire blob first and then slice it? It might be memory-expensive and won't fit in a 4GB boundary for big BLOBs.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -299,11 +377,27 @@ public void primeEditorValue(@NotNull DBRProgressMonitor monitor, @NotNull Style
                     }
                     applyEditorStyle();
                 }
+            }
+        });
+    }
+
+    private void showRestrictedContent(@NotNull DBDContent value) throws DBCException, IOException {
```
ShadelessFox: Let's also rename it as "restricted" doesn't quite fit the purpose of this method. Let's call it something like "showLimitedContent", you get the spirit.

Also, let's pass `maxContentSize` as an argument (already computed by the caller) to avoid code duplication.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -269,28 +310,65 @@ public <T> T getAdapter(Class<T> adapter) {
...
-                    if (textWidget != null && longContent) {
+                    if (textWidget != null) {
                         GC gc = new GC(textWidget);
                         try {
-                            UIUtils.drawMessageOverControl(textWidget, gc, NLS.bind(ResultSetMessages.panel_editor_text_loading_placeholder_label, textInput.getContentLength()), 0);
+                            UIUtils.drawMessageOverControl(textWidget, gc,
+                                NLS.bind(ResultSetMessages.panel_editor_text_loading_placeholder_label, textInput.getContentLength()), 0);
                             editor.setInput(textInput);
+                            messageBar.setForeground(UIStyles.getDefaultTextForeground());
+                            messageBar.setText(String.format("Content size: %s Bytes", textInput.getContentLength()));
```
ShadelessFox, line 368: I'm unsure whether this label should be always displayed or not. It's probably up to a discussion.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Yeap sure, let rework to read require content lenght.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Done

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Persistant panel looks empty and ugly without information, displaing content lenght is kind of interesting to have. But let's enchance it after UI/UX discussion.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/preferences/PrefPageResultSetMain.java
serjiokov: reverted

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -381,16 +381,15 @@ private void showRegularContent(@NotNull DBRProgressMonitor monitor) throws DBEx
...
-    private void showRestrictedContent(@NotNull DBDContent value) throws DBCException, IOException {
-        int maxContentSize = DBWorkbench.getPlatform().getPreferenceStore().getInt(ResultSetPreferences.RS_EDIT_MAX_TEXT_SIZE);
+    private void showLimitedContent(@NotNull DBDContent value, int lengthInBytes) throws DBCException, IOException {
         DBDContentStorage contents = value.getContents(new VoidProgressMonitor());
         try (final InputStream stream = contents.getContentStream()) {
-            ByteArrayOutputStream buffer = new ByteArrayOutputStream((int) contents.getContentLength());
-            IOUtils.copyStream(stream, buffer);
-            byte[] displaingContentBytes = Arrays.copyOfRange(buffer.toByteArray(), 0, maxContentSize * 1000);
+            ByteArrayOutputStream buffer = new ByteArrayOutputStream(lengthInBytes);
+            IOUtils.copyStream(stream, buffer, lengthInBytes);
```
ShadelessFox: The third argument is the internal `bufferSize`, not the maximum amount of bytes to read.

Just create a byte array and use `stream.read(buffer)` to read it, it will return the length of bytes read which you can use to trim the output array if fewer bytes were read.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
serjiokov: Better to use pure readNBytes. 
`byte[] displaingContentBytes = stream.readNBytes(lengthInBytes);`

ShadelessFox (approved):

Matvey16:
Verified

serge-rider (approved):

PR #21136, opened 2023-09-06
dbeaver/pro#1862 Redirect awt browse requests to swt browser

### LINKED ISSUE #20887: Azure SQL Server MFA Authentication repeatedly opens multiple browser windows
### Description

When authenticating with the Azure SQL database driver, it opens a browser tab/window every time a connection is opened or re-connected. In these tabs it only says

**Authentication complete. You can close the browser and return to the application.**

This is super annoying when working with multiple databases and database connections as any time a connection drops and reconnects, a tab will be opened.

I have around 40 SQL databases I work with, and in DBeaver I am normally connected to the majority of them. However, If there's hiccups in the connection (like Wifi connection fails for a moment, VPN disconnects, ...), all open database connections are reconnected automatically which is a good thing. However, for all database connections that are open, DBeaver opens a tab in my browser, which means suddenly I have 30+ browser tabs I have to close manually. This does not only happen when working in DBeaver, but also if I'm in another program like Visual Studio, Word or Excel. If DBeaver in the background decides that it needs to reconnect, my browser pops up and keeps on opening tabs. Sometimes there are a few seconds between the tabs are being opened, so I switch back to where I was working before and after a few seconds the browser pops up again until all sessions have been reconnected.

Sometimes the sessions do not reconnect automatically but when I send a new query in a script window and DBeaver realizes that the connection is dead. Love that feature, but again, a browser tab is opened. That might not be a big problem with one or two connections, but working with many connections you spend a lot of time switching between browser and dbeaver and closing tabs which is super annoying.

This is most likely intended behavior by the Azure SQL driver, however, it seriously disrupts my workflow as this happens multiple times per day. 

My suggestion would be to integrate a tiny browser in dBeaver (or in the driver) which handles re-authentication in the background. I am aware that sometimes I have to re-authenticate with my username and password, but this rarely happens as my credentials seem to be cached in the browser.

Please find a way I can use dbeaver without constantly being harassed by authentication attempts in my browser. I love dbeaver and want to continue using it

### DBeaver Version

Version 23.1.4.202308061552

### Operating System

Win 10

### Database and driver

Azure SQL / Azure SQL Server MFA Authentication

### Steps to reproduce

- Create multiple connections to Azure databases using the Azure SQL driver
- Connect to all of them in different script windows
- Disrupt the connection (turn off/on wifi, ...)
- DBeaver will attempt to reconnect and open multiple browser windows.
- MIght be that the browser window opens when a new query is sent from a browser window. 

### Additional context

_No response_

Destrolaric:
Closes dbeaver/pro#1862
I hope this will work...
image

Edge cases to test:
1. Try running with system java
2. Try running on Linux/MacOS distributions
3. Try different OAuth, excluding AWS(due to it calling auth via launching another binary)
4. Test that everything works in TE
5. Check that the checkbox is present for all systems and should be disabled if Internet Explorer is selected. No redirection should happen if IE is the default browser


Matvey16:
Verified

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/CoreMessages.java
```diff
@@ -350,6 +350,8 @@ public class CoreMessages extends NLS {
     public static String pref_page_ui_general_group_browser;
     public static String pref_page_ui_general_combo_browser;
     public static String pref_page_ui_general_combo_browser_tip;
+    public static String pref_page_ui_general_check_browser_auth;
+	public static String pref_page_ui_general_check_browser_auth_tip;
```
E1izabeth:
```suggestion
    public static String pref_page_ui_general_check_browser_auth_tip;
```

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/CoreResources.properties
```diff
@@ -366,6 +366,8 @@ pref_page_ui_general_combo_language_tip = Application language - used for user i
 pref_page_ui_general_group_browser = Client Web Engine
 pref_page_ui_general_combo_browser = Web Engine
 pref_page_ui_general_combo_browser_tip = Used for displaying web contents inside the application
+pref_page_ui_general_check_browser_auth = Open auth links in embedded browser
+pref_page_ui_general_check_browser_auth_tip = Will redirect SSO and other auth requests to be shown in the embedded window
```
E1izabeth: The sentence is malformed. It could be `Will redirect SSO and other auth requests to the embedded window` or `Will make SSO and other auth requests to be shown in the embedded window`
```suggestion
pref_page_ui_general_check_browser_auth_tip = Will redirect SSO and other auth requests to be shown in the embedded window
```

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/browser/BrowsePeerMethods.java
```diff
@@ -0,0 +1,79 @@
...
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class BrowsePeerMethods {
+    private static final Log log = Log.getLog(BrowsePeerMethods.class);
+
+    /**
+     * Checks if request can be opened in SWT browser
+     * @return true if request is redirectable
```
E1izabeth:
```suggestion
     * Checks if request can be opened in SWT browser
     *
     * @return true if request is redirectable
```

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/browser/BrowsePeerMethods.java
```diff
@@ -0,0 +1,79 @@
...
+            if (!internalWebBrowserAvailable) {
+                log.warn("Embedded Browser is disabled or unavailable");
+            }
+        });
+        return result.get();
+    }
+
+    /**
+     * Open URI via SWT browser API
+     * @param uri uri to open
```
E1izabeth:
```suggestion
     * Open URI via SWT browser API
     *
     * @param uri uri to open
```

E1izabeth (changes requested):

E1izabeth (approved):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/browser/BrowsePeerMethods.java
```diff
@@ -0,0 +1,80 @@
...
+     * @return was it opened successfully
+     */
+    public static boolean browseInSWTBrowser(URI uri) {
+        DBPPreferenceStore store = DBeaverActivator.getInstance().getPreferences();
+        if (store.getBoolean(DBeaverPreferences.UI_USE_EMBEDDED_AUTH)) {
+            AtomicBoolean result = new AtomicBoolean();
+            UIUtils.syncExec(() -> {
+                try {
+                    BrowserPopup.openBrowser("redirect.auth", uri.toURL());
```
ShadelessFox, line 69: No magic constants please

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/browser/BrowsePeerMethods.java
```diff
@@ -0,0 +1,80 @@
...
+    /**
+     * Open URI via SWT browser API
+     *
+     * @param uri uri to open
+     * @return was it opened successfully
+     */
+    public static boolean browseInSWTBrowser(URI uri) {
+        DBPPreferenceStore store = DBeaverActivator.getInstance().getPreferences();
+        if (store.getBoolean(DBeaverPreferences.UI_USE_EMBEDDED_AUTH)) {
+            AtomicBoolean result = new AtomicBoolean();
```
ShadelessFox, line 66: No need to use atomic primitives here because you're performing a blocking operation.

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/browser/BrowsePeerMethods.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * Open URI via SWT browser API
+     *
+     * @param uri uri to open
+     * @return was it opened successfully
+     */
+    public static boolean browseInSWTBrowser(URI uri) {
+        DBPPreferenceStore store = DBeaverActivator.getInstance().getPreferences();
+        if (store.getBoolean(DBeaverPreferences.UI_USE_EMBEDDED_AUTH)) {
```
ShadelessFox, line 65: Maybe invert it and perform an early return to simplify code reading?

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/browser/BrowserPopup.java
```diff
@@ -0,0 +1,85 @@
...
+        try {
+            Browser browser = new Browser(composite, SWT.NONE);
+            gd = new GridData(GridData.FILL_BOTH);
+            gd.widthHint = 600;
+            gd.heightHint = 500;
+            browser.setLayoutData(gd);
+            browserRegistry.put(id, browser);
+            browser.setUrl(url.toString());
+        } catch (SWTError e) {
+            log.error("Could not instantiate Browser", e);
```
ShadelessFox, line 67: It should be a fatal error IMO

Destrolaric: There are 100 reasons why it can go wrong, especially on linux, in that case we warn the user and open using external browser

ShadelessFox (approved):

PR #21141, opened 2023-09-06
dbeaver/dbeaver#20276 Show query in grouping panel

### LINKED ISSUE #20276: It's impossible to see the query in the grouping panel
Right now the query can only be seen in the Data grid. We want to make it possible to open it in the Grouping panel.

image

serge-rider:


MashaKorax:
verified

PR #21142, opened 2023-09-06
#19676 Save console as a script

### LINKED ISSUE #19676: Add a new item in context menu: "Read data in SQL console"
I want to open the SQL console from the context menu:

image

And it must automatically generate the SELECT query:

image

ShadelessFox:


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -846,8 +845,8 @@ public boolean isDirty() {
             return true;
         }
         if (isNonPersistentEditor()) {
-            // Console is never dirty
-            return false;
+            // Console is always dirty
+            return true;
```
E1izabeth: Won't it add * to the name of each console tab?

ShadelessFox: It will. That's the goal.

E1izabeth: But doesn't the main idea of console - to be unsaved editor? That's why we have them. And they have another icon to indicate it.

The * near the tab name shows unsaved state for editors, when the normal editor state is saved. But the normal console state is unsaved.

I don't like the idea of always having * for all console editors.

ShadelessFox: It's up for a discussion.

ShadelessFox: Perhaps we can override the editor's title to exclude the asterisk if it's a console.

ShadelessFox: Unfortunately, it's impossible to override it without hacky constructions.

Keep in mind that it WON'T prompt the user to save itself upon closing.

E1izabeth (approved):

LonwoLonwo (approved):

uslss:
verified

PR #21148, opened 2023-09-07
#20994 Reset document provider on input change

### LINKED ISSUE #20994: Empty SQL script editor after saving content of console
### Description

If I start to write SQL query using a console, and then decide I need to store the query as a script, I expect that SAVE AS function will save the content of the console into a file, starting to behave as an opened script since then.

What is happening is, that SAVE AS operation being issued from the console creates the file but the content disappears from the editor. Fortunately the code is indeed stored in the file.

BTW after saving the file, subsequent open command opens completely different location.

Both issues shown on the video
https://youtu.be/t-Yf2N3ET98

### DBeaver Version

CE 23.1.5

### Operating System

Fedora 38

### Database and driver

Postgresql 14

### Steps to reproduce

See video

### Additional context

_No response_

ShadelessFox:


E1izabeth (approved):

Matvey16:
Verified

PR #21155, opened 2023-09-07
dbeaver/pro#1897 NIO EFS integration

ShadelessFox:


plugins/org.jkiss.dbeaver.model/META-INF/MANIFEST.MF
```diff
@@ -52,6 +52,7 @@ Export-Package: org.jkiss.dbeaver,
  org.jkiss.dbeaver.model.messages,
  org.jkiss.dbeaver.model.navigator,
  org.jkiss.dbeaver.model.navigator.fs,
+ org.jkiss.dbeaver.model.navigator.fs,
```
LonwoLonwo: looks like duplicate

PR #21162, opened 2023-09-08
dbeaver/dbeaver#21053 Display errors in the result tab 

### LINKED ISSUE #21053: Display errors in the result tab
### Description

When there is a bug in a query, it's now difficult to find an error in the result tab: it is hidden behind the query view. Users cannot find it. See details: https://github.com/dbeaver/dbeaver/issues/21013

Screenshot 2023-08-30 at 17 31 27

We need to display the view with an error instead of it.

Screenshot 2023-08-30 at 17 31 44

### DBeaver Version

23.1.5

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

1. Execute any query with a bug
2. See the result

### Additional context

_No response_

serjiokov:
The state of error viewer is reseting (restore default state 50/50) in case of catching errors.

image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
@@ -1016,15 +1016,7 @@ public class ResultSetViewer extends Viewer
             }
         } else {
             if (viewerSash != null) {
-                Control control = viewerSash.getChildren()[0];
-                viewerSash.setMaximizedControl(control);
-                if (control instanceof Composite) {
-                    Control child = ((Composite) control).getChildren()[0];
-                    if (child instanceof CustomSashForm) {
-                        CustomSashForm form = (CustomSashForm) child;
-                        form.resetWeight();
-                    }
-                }
+                viewerSash.setMaximizedControl(viewerSash.getChildren()[0]);
             }
         }
 
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
@@ -76,7 +76,6 @@ import org.jkiss.dbeaver.runtime.DBeaverNotifications;
 import org.jkiss.dbeaver.runtime.jobs.DataSourceJob;
 import org.jkiss.dbeaver.tools.transfer.ui.internal.DTUIMessages;
 import org.jkiss.dbeaver.ui.*;
-import org.jkiss.dbeaver.ui.controls.CustomSashForm;
 import org.jkiss.dbeaver.ui.controls.TabFolderReorder;
 import org.jkiss.dbeaver.ui.controls.ToolbarSeparatorContribution;
 import org.jkiss.dbeaver.ui.controls.VerticalButton;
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/view/ErrorPresentation.java
```diff
@@ -98,29 +98,6 @@ public void createPresentation(@NotNull IResultSetController controller, @NotNul
...
-            if (errorWidth != null) {
-                String[] widthStrs = errorWidth.split(":");
-                if (widthStrs.length == 2) {
-                    partDivider.setWeights(Integer.parseInt(widthStrs[0]),
-                        Integer.parseInt(widthStrs[1]));
-                }
-                widthSet = true;
-            }
-            if (!widthSet) {
-                partDivider.setWeights(700, 300);
```
E1izabeth: Shouldn't we left this string to set weights? Or is it set somewhere else also?

serjiokov: Decided: remove saved state of weights and use default, to avoid hiding errors.

E1izabeth: OK

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -76,6 +76,7 @@
 import org.jkiss.dbeaver.runtime.jobs.DataSourceJob;
 import org.jkiss.dbeaver.tools.transfer.ui.internal.DTUIMessages;
 import org.jkiss.dbeaver.ui.*;
+import org.jkiss.dbeaver.ui.controls.CustomSashForm;
```
E1izabeth: This import seems like it's unintended. Is it?

serge-rider: Why?

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/view/ErrorPresentation.java
```diff
@@ -98,29 +98,6 @@ public void createPresentation(@NotNull IResultSetController controller, @NotNul
                 textWidget.setText(sqlText);
             }
         }
-
-        try {
```
serge-rider, line 102: Let's make it conditional (e.g. static variable REMEBER_SASH_RATIO). May be useful if we will change our mind after release.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/view/ErrorPresentation.java
serjiokov: Proposed varaible added to keep state.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
serjiokov: Removed, not used

serjiokov: Removed, not used

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/view/ErrorPresentation.java
E1izabeth: I believe we should keep try catch too in this case

serjiokov: Nobody throw Exception here, specification of weights is always in integers by internal contract api.

E1izabeth: OK

E1izabeth (approved):

MashaKorax:
verified

serge-rider (approved):

PR #21205, opened 2023-09-13
#17675 read enum values as database cache; fix data types renaming

### LINKED ISSUE #17675: Opening a connection takes a lot of time on `Read data types`
### Description

Related to #17117 - it happens in my case as well.
I have a DB with a dozen schemas and ~315 enums.
Whenever I open a connection to a remote DB the background task stays on `Read data types` for about 35 seconds before the connection is actually opened.
suggestions: 
1. do this as a background task AFTER the connection is made
2. lookup a data type when it's actually needed i.e describing or selecting from a table for the first time.

### DBeaver Version

Community 22.2.0

### Operating System

MacOS 12.6

### Database and driver

postgres 12.4, default JDBC drivers

### Steps to reproduce

create a remote (i.e cloud) database with many enums.
open a connection to it

### Additional context

it only happens when `Connect`ing to a DB - it doesn't happen when invalidating the connection

LonwoLonwo:
Now, we read the enum's list with all data types together. One query instead of many.

Was

2023-09-13 22_32_59-DBeaver Enterprise 21 3 4 - _none_ Script-83

Now

2023-09-13 23_47_29-DBeaver Ultimate 23 2 0 - enum_type_7

So, please check:

- [x] Enum data values are still viewable for enum data types
- [x] Enum data values are still not viewable for not enum data types
- [x] Enum data types refreshing
- [x] Data types renaming via UI
- [x] Connection and data types reading (just check your Qqery manager) - for forks (Redshift, Cockroach, etc.)

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataType.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataType.java
@@ -33,6 +33,7 @@ import org.jkiss.dbeaver.model.impl.jdbc.JDBCUtils;
 import org.jkiss.dbeaver.model.impl.jdbc.cache.JDBCObjectCache;
 import org.jkiss.dbeaver.model.impl.jdbc.struct.JDBCDataType;
 import org.jkiss.dbeaver.model.meta.ForTest;
+import org.jkiss.dbeaver.model.meta.IPropertyValueValidator;
 import org.jkiss.dbeaver.model.meta.Property;
 import org.jkiss.dbeaver.model.meta.PropertyLength;
 import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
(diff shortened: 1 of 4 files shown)


ShadelessFox (approved):

E1izabeth (approved):

uslss:
enum data is incorrectly shown in datagrid for newly created type and table - no values are available both on grid and value panel.

https://github.com/dbeaver/dbeaver/assets/31996417/bf57c9b7-e4ad-49f3-909a-8ddd7c921b72

LonwoLonwo:
Fixed.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataType.java
```diff
@@ -314,6 +314,27 @@ private void readEnumValues(@NotNull DBRProgressMonitor monitor) throws DBExcept
             .toArray();
     }
 
+    private void readNewEnumValues(DBRProgressMonitor monitor) throws DBException {
```
ShadelessFox, line 317: Maybe we can just invalidate enum cache instead of duplicating queries?

uslss:
verified

E1izabeth (approved):

PR #21236, opened 2023-09-18
CB-3839. Added async error events messages for te

DenisSinelnikov:


alexander-skoblikov (commented):
it might be better to make an asynchronous log, and not just an error 🤔

PR #21246, opened 2023-09-19
#20988 move foreign table to separate folder in the navigator tree

### LINKED ISSUE #20988: [PostgreSQL] add the Foreign Tables folder
Kind of continuations of this talk:
https://github.com/dbeaver/dbeaver/issues/16372#issuecomment-1270329764

We need to add a special folder for foreign tables for PostgreSQL.

For now, we only have a special icon for these tables, but they are in the general Tables list.

2023-08-24 12_49_39-DBeaver Ultimate 23 1 5 - external_test

LonwoLonwo:
2023-09-18 19_12_20-DBeaver Ultimate 23 2 0 - newtable

2023-09-18 19_21_28-DBeaver Ultimate 23 2 0 - newtable

A separate folder was created for PG (starting version 8.4) and other forks that support foreign data wrappers.
You will only see Foreign Tables in the tree if you see foreign wrappers and servers.
Cockroach, Redshift, and Yugabyte - do not support it.
Also, external tables in Greenplum 7 will be in the Foreign Tables folder if the view pg_catalog.pg_exttable is not readable.
External tables in Greenplum 7 will be in the External Tables folder if the view pg_catalog.pg_exttable is readable.

Also, Foreign tables can have only columns, constraints, and dependencies. It is expected behavior. 

So, please check that new folders are visible (for databases support it) and that tables inside Foreign or External folders do not have duplicates in the "Tables" folder.


plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumSchema.java
```diff
@@ -67,6 +69,26 @@ public Collection<? extends JDBCTable> getExternalTables(DBRProgressMonitor moni
         return new ArrayList<>(getTableCache().getTypedObjects(monitor, this, GreenplumExternalTable.class));
     }
 
+    @Override
+    public List<? extends PostgreTable> getTables(DBRProgressMonitor monitor) throws DBException {
+        List<? extends PostgreTable> postgreTables = super.getTables(monitor);
+        // Remove external tables from the list. Store is different folder.
+        return postgreTables.stream().filter(e -> !(e instanceof GreenplumExternalTable)).collect(Collectors.toCollection(ArrayList::new));
```
E1izabeth: Isn't there a way to query only non-external tables instead of querying all and then filtering? Don't we duplicate query requesting the tables when looking for external tables (_"stored in different folder"_)

LonwoLonwo: This a story like for views reading. We read tables and views - in one list. Because they have one parent class. And then separate them. 
Here the logic is the same.
We need a full tables-views list for SQL assistant, as an example.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumSchema.java
```diff
@@ -67,6 +69,26 @@ public Collection<? extends JDBCTable> getExternalTables(DBRProgressMonitor moni
         return new ArrayList<>(getTableCache().getTypedObjects(monitor, this, GreenplumExternalTable.class));
     }
 
+    @Override
+    public List<? extends PostgreTable> getTables(DBRProgressMonitor monitor) throws DBException {
+        List<? extends PostgreTable> postgreTables = super.getTables(monitor);
+        // Remove external tables from the list. Store is different folder.
```
ShadelessFox:
```suggestion
        // Remove external tables from the list. They are stored in a different folder.
```

plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumSchema.java
ShadelessFox: Readability nitpick

```suggestion
        return postgreTables.stream()
            .filter(e -> !(e instanceof GreenplumExternalTable))
            .collect(Collectors.toCollection(ArrayList::new));
```

plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumSchema.java
```diff
@@ -67,6 +69,26 @@ public Collection<? extends JDBCTable> getExternalTables(DBRProgressMonitor moni
...
+
+    @Override
+    public List<? extends PostgreTable> getForeignTables(DBRProgressMonitor monitor) throws DBException {
+        List<? extends PostgreTable> foreignTables = super.getForeignTables(monitor);
+        GreenplumDataSource dataSource = getDataSource();
+        if (dataSource.isServerVersionAtLeast(7, 0) && dataSource.isHasAccessToExttable()) {
+            // Starting Greenplum version 7 external tables are marked as foreign tables.
+            // Lets's remove external tables from the list foreign tables. Store is the External tables folder.
+            return foreignTables.stream()
+                .filter(e -> !(e instanceof GreenplumExternalTable)).collect(Collectors.toCollection(ArrayList::new));
```
ShadelessFox: Readability nitpick

```suggestion
            return foreignTables.stream()
                .filter(e -> !(e instanceof GreenplumExternalTable))
                .collect(Collectors.toCollection(ArrayList::new));
```

ShadelessFox (approved):

uslss:
verified

PR #21247, opened 2023-09-19
dbeaver/pro#1954 clear superInheritance for partitions

Destrolaric:
Closes dbeaver/pro#1954

Relation should not duplicate for scheme erd diagrams

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
@@ -435,10 +435,10 @@ public class PostgreSchema implements
     }
 
     private void resetPartitionsInheritance(DBRProgressMonitor monitor) throws DBException {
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
```diff
@@ -434,6 +434,14 @@ private void cacheTableInheritance(DBRProgressMonitor monitor) throws DBExceptio
         }
     }
 
+    private void resetPartitionsInheritance(DBRProgressMonitor monitor) throws DBException {
+        List<PostgreTable> partitions = getTableCache().getTypedObjects(monitor, this, PostgreTable.class).stream()
+            .filter(PostgreTableBase::isPartition).collect(Collectors.toList());
+        for (PostgreTable partition : partitions) {
+            partition.resetSuperInheritance();
+        }
```
ShadelessFox: Maybe just use a simple `for` loop with `table.isPartition()` check? What's the point of using streams here?

LonwoLonwo (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #21261, opened 2023-09-21
dbeaver/dbeaver#19915 read values as double

### LINKED ISSUE #19915: Table editor changes for table column value not saved upon save
### Description

Community 23.0.4

A table has no primary key. When changing a field of a record , the save button lights up. DBeaver uses all fields in the UPDATE command, however the data never gets updated upon a "save" even when the save button returns to grey color. This gives the use the impression that the change was saved. Quite a long time ago I could make changes with save and this problems did not happen. As a workaround now copy paste the update script and modify/generate the SQL. The save function makes changing much more easier. Also tried to change properties and enable "auto commit" but did not have effect.

P.S. Copying  the  generated script from the save button and copy/paste it into a new script and execute it, also resulted in 0 updates. Have not have time to analyze where the script is wrong.

### DBeaver Version

Community 23.0.4

### Operating System

Unbuntu 22.04

### Database and driver

 mysql  Ver 15.1 Distrib 10.8.7-MariaDB, for debian-linux-gnu (x86_64) using  EditLine wrapper

### Steps to reproduce

1. Click on table, real SQL in editor
2. change value of a column (Save higlhighted)
3. Click save dropdown and show the SQL
4. Click on Save
5. Save (save is dimmed)
6. Click on table, read SQL in edit
7. Values are the same as before the edit

### Additional context

_No response_

Destrolaric:
Closes dbeaver/dbeaver#19915

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
@@ -337,7 +337,7 @@
             label="MySQL data types provider">
 
             <datasource id="mysql"/>
-
+            <type name="FLOAT"/>
             <type name="ENUM"/>
             <type name="SET"/>
             <type standard="DATE"/>
--- a/plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
@@ -338,6 +338,8 @@
 
             <datasource id="mysql"/>
             <type name="FLOAT"/>
+            <type name="DOUBLE"/>
+            <type name="DOUBLE PRECISION"/>
             <type name="ENUM"/>
             <type name="SET"/>
             <type standard="DATE"/>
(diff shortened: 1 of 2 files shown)


E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
```diff
@@ -337,7 +337,7 @@
             label="MySQL data types provider">
 
             <datasource id="mysql"/>
-
+            <type name="FLOAT"/>
```
LonwoLonwo, line 340: How about Double(10,10)? Did you test it? Maybe can also be here?

LonwoLonwo (approved):

MashaKorax:
Still reproducible  **db** - https://github.com/dbeaver/pro/wiki/Sample-databases#mysql-8-oracle-cloud  **schema** test  **table** float

https://github.com/dbeaver/dbeaver/assets/84867187/938fc41b-ab35-4d02-ab15-77d018bbe573

MashaKorax:
verified

PR #21285, opened 2023-09-22
dbeaver/pro#1959 configure browser earlier

Destrolaric:
Closes dbeaver/pro#1959


E1izabeth:
> We should set the correct browser instance before opening auth dialog

Could you explain why?

Destrolaric:
@E1izabeth Because if not configured from our side, Internet Explorer will be opened instead of Edge, even if it is configured in preferences. We don't exactly want to use IE for authorization due to a lot of inherent issues with it.

Matvey16:
Verified

PR #21309, opened 2023-09-25
#21302 Fix tool task wizard finish

### LINKED ISSUE #21302: Backup process
I just updated to the most recent version and noticed that the whole backup (PostgreSQL dump) GUI process has changed. Now when I click "Proceed" the whole modal dialogue dissapears and the task in moved into the background task. I don't know if this is a bug or by design but I am very annoyed by the new approach. I much preffered the old one.

Improve the whole backup GUI process to be more interactive with the user.

There are none except to accept the new approach which is not the best but it does the job. The lack of feedback about the whole dump process is what bothers me.

E1izabeth:


uslss:
verified

PR #21333, opened 2023-09-26
#21281 Don't apply mysql execute script task to any DBSDataManipulator

### LINKED ISSUE #21281: Wrong shell name for postgres script task creaton called from table node
### Description

1. call the context menu on a table in PostgreSQL
2. context menu -> tools -> execute script
3. Shell for 'MySQL' instead of 'PostgreSQL' is opened

image

When calling the context menu on a database node all is OK

https://github.com/dbeaver/dbeaver/assets/43252954/f378320e-6459-45be-9d5c-82742e03ef2e

### DBeaver Version

23.2.1

### Operating System

win10

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Execute script task should be available only on database node both for postgresql and mysql


LonwoLonwo (approved):

serge-rider:
I thought that `<datasource id="mysql"/>` must restrict this task other datasources

E1izabeth:
To apply these constraints with AND (instead of OR) we need to change this logic https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/AbstractContextDescriptor.java#L84
It touches a huge number of places, we can broke a lot. Let's just fix the bug we made with this condition.

MashaKorax:
verified

PR #21364, opened 2023-09-28
dbeaver/dbeaver#20522 Rename `Entries per page` setting

### LINKED ISSUE #20522: Investigate the usage of `Entries per page` setting
### Description

We have a setting `Entries per page` in Query Manager, but we don't have pagination in Query manager. Let's find out what does this setting do and make it clear or remove this setting if there is no usage of it.

### DBeaver Version

DBeaver 23.1.2

### Steps to reproduce

https://github.com/dbeaver/dbeaver/assets/28875055/7728f8d5-19c3-474a-86ac-cde5a7ac1cda

E1izabeth:
image


uslss:
verified

PR #21368, opened 2023-09-28
CB-3930 fs gql api

alexander-skoblikov:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/DBFFileSystemManager.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class DBFFileSystemManager {
+    private static final Log log = Log.getLog(DBFFileSystemManager.class);
+    private final Map<String, FileSystemProvider> nioFileSystems = new LinkedHashMap<>();
+    private final Map<String, DBFVirtualFileSystem> dbfFileSystems = new LinkedHashMap<>();
+
+
+    public synchronized void reloadFileSystems(@NotNull DBRProgressMonitor monitor, @NotNull SMSessionContext sessionContext) {
```
serge-rider: Let's pass sessionContext in constructor.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/DBFFileSystemManager.java
```diff
@@ -0,0 +1,84 @@
...
+import java.net.URI;
+import java.nio.file.Path;
+import java.nio.file.spi.FileSystemProvider;
+import java.util.Collection;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class DBFFileSystemManager {
+    private static final Log log = Log.getLog(DBFFileSystemManager.class);
+    private final Map<String, FileSystemProvider> nioFileSystems = new LinkedHashMap<>();
```
serge-rider: This looks very strange. File systems need extra info (project, cloud config id, etc).
We create them without any of these parameters. How does it work?

serge-rider (changes requested):

serge-rider (approved):

PR #21382, opened 2023-09-29
dbeaver/pro#1841 auto fit columns during export

LonwoLonwo:
2023-10-05 15_08_25-film_202310051507 - Excel


E1izabeth (approved):

Matvey16:
Verified

Matvey16:
This feature has 1 edge case where it works incorrectly. If the cell has a lot of letters l e.g. `abclllllllllllllllllllllabclllllllllllllllllllllllllllllllllllllllabcllllllllllllllll` it won't widen the column enough to fit everything

Issue #20784, opened 2023-08-03
The Chinese description of the fields in the Oceanbase database table is garbled  code


pengyong255237:
### Description

The Chinese description of the fields in the Oceanbase database table is garbled  code
![Snipaste_2023-08-03_15-45-39](https://github.com/dbeaver/dbeaver/assets/28984017/0808f7d1-3990-4055-902c-67477f3cd5b2)

The database plugin of IDEA can display table field descriptions normally
![Snipaste_2023-08-03_10-02-47](https://github.com/dbeaver/dbeaver/assets/28984017/0859801c-5926-4002-a662-9f0682a55683)

### DBeaver Version

Community 23.1.2.202307091549

### Operating System

windows

### Database and driver

driver：oceanbase-client-2.4.3.jar

LonwoLonwo:
Hello @pengyong255237 

Please translate your ticket text into English, including the title.

pengyong255237:
> Hello @pengyong255237
> 
> Please translate your ticket text into English, including the title.

done

LonwoLonwo:
And what server version do you use?

ShadelessFox:
Hello @pengyong255237,

Does the issue occur in the "DDL" tab? Please make sure to enable the "show comments" option.

pengyong255237:
> Are you asking about the window version or the dbeaver version?

pengyong255237:
> 您使用什么服务器版本？

Are you asking about the window version or the dbeaver version?

pengyong255237:
> 你好@pengyong255237,
> 
> 问题是否出现在“DDL”选项卡中？请确保启用“显示评论”选项。
The ER chart tab displays garbled code, and the DDL tab does not display the DDL SQL for field descriptions, as shown in the following figure
![123](https://github.com/dbeaver/dbeaver/assets/28984017/bb016575-c45b-4080-a441-95e32029a1b9)
![456](https://github.com/dbeaver/dbeaver/assets/28984017/16a29de0-8ae3-4943-b986-318965d95f96)

pengyong255237:
> > 您使用什么服务器版本？
> 
> Are you asking about the window version or the dbeaver version?

After executing the update field description sql statement, only English alphabet and Arabic numerals will be displayed. Chinese is still garbled. The ddl tab does not display the field description at all

![000](https://github.com/dbeaver/dbeaver/assets/28984017/5ffae5b3-b73c-4194-9ea1-70f5e3e310e5)

[...]

Issue #20392, opened 2023-06-27
Issue with hanzi in Data Editor


ywss:
### Description
Chinese hanzi can't be displayed in a proper way. Encoding change does nothing.
As shown on the picture below it works in other Application, but not in DBeaver.

sqlite db：[hk.zip](https://github.com/dbeaver/dbeaver/files/11877281/hk.zip)

![微信截图_20230627110348](https://github.com/dbeaver/dbeaver/assets/1284357/a993195a-3fb6-4b2a-8cba-2aabb22af2f4)
![微信截图_20230627110423](https://github.com/dbeaver/dbeaver/assets/1284357/61e893df-010e-48e8-8cfd-4e05160d8c19)

### DBeaver Version

community edition  23.1.1.202306251800

### Operating System

windows10

### Database and driver

1. SQLite
2. SQLite JDBC [3.42.0.0]

E1izabeth:
Thank you for bug report.

E1izabeth:
Does it still reproducible in the latest DBeaver version? We expect that it has been solved with #21513

E1izabeth:
It's been a while since no update here. If the issue is still actual please let me know and provide additional information.

gregkotsaftis:
I can verify this issue as well for Greek characters stored in the db. I use firebirdsql with NONE character set.
Data editor was showing greek characters without any issues in older versions e.g. v6.0.0, but later versions e.g. 22.x.x never worked properly!

Issue #20054, opened 2023-05-23
TRUNCATE Tools dialog unreadable codec


gateslu:
### Description

![image](https://github.com/dbeaver/dbeaver/assets/1697700/fb5a51fa-26ba-4d1b-889e-d07dc3b41de2)

### DBeaver Version

Community Edition 23.0.5.202305211610

### Operating System

Windows 11

### Database and driver

MySQL 8

E1izabeth:
Seems that something goes wrong with encoding in this log.

I reproduced while trying to execute Truncate task with error for Database with non-latin language locale. As a result error from truncate tool was shown to me in an error window correctly, but in this task log I saw the same error with broken encoding.

E1izabeth:
Can't reproduce under debug, but it's stable reproducible on the installed version.
It happens not only for MySQL, but for PostgreSQL also. Probably, for any database.

Issue #19631, opened 2023-04-07
DBeaver doesn't read UTF-8 data base


G-Zidarov:
Hello, I have a problem with loaded data base (MariaDB) which includes Cyrillic letters. I have Macbook Pro 13' 2017 with MacOS Ventura and Dbeaver Version 22.3.4 
On the server (into the cPanel) everything is readable but when connect data base to the DBeaver the result is as it seen on the picture below (column: title).
<img width="651" alt="Screenshot 2023-04-06 at 18 00 20" src="https://user-images.githubusercontent.com/94197202/230419315-c221f6cf-d212-482e-ab95-74ee6521029a.png">
So far, I already tried to edit DB connection and inserted "characterEncoding [UTF-8]" and checked Preferences/General/Workspace but without results.
Would you advise me how to display correctly Cyrillic value from data base.
<img width="719" alt="Screenshot 2023-04-06 at 17 53 17" src="https://user-images.githubusercontent.com/94197202/230417351-09b7322e-5b00-4497-a498-c8e67ddf318b.png"> <img width="845" alt="Screenshot 2023-04-06 at 17 57 37" src="https://user-images.githubusercontent.com/94197202/230418796-2a8b5175-ba57-43cd-8adf-cd02880c19b4.png">

LonwoLonwo:
Hello @G-Ziradov 

Could you please provide your table DDL?

And value from 

```
SELECT default_character_set_name FROM information_schema.SCHEMATA 
WHERE schema_name = <your_catalog_name>;
```

G-Zidarov:
Hello @LonwoLonwo 
thanks for your response.
I've made some screenshots of DDL and SELECT statement.
DDL [screenshot: 1] is from table which is part of data base from row 4 [screenshot: 2].

[screenshot :1]
<img width="750" alt="Screenshot 2023-04-07 at 19 19 23" src="https://user-images.githubusercontent.com/94197202/230644172-59c2f07d-ec8c-4e2b-ad43-0855942e4576.png">
[screenshot:2]
<img width="1064" alt="Screenshot 2023-04-07 at 19 17 16" src="https://user-images.githubusercontent.com/94197202/230645128-a65c23cd-92d4-4595-8518-29c9857903ee.png">

E1izabeth:
Does it still reproducible in the latest DBeaver version? We expect that it has been be solved with #21513

E1izabeth:
[...]

PR #21419, opened 2023-10-04
dbeaver/pro#1905 Incorrect characters in databricks tables

serjiokov:
Fixed specification of file.encoding property require for correct string processing.
Provided mechanism in order to change default encoding value

### DIFF
--- a/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageDatabaseUserInterface.java
+++ b/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageDatabaseUserInterface.java
@@ -74,11 +74,13 @@ public class PrefPageDatabaseUserInterface extends AbstractPrefPage implements I
 
     @Nullable
     private Combo clientTimezone;
+    private Combo cmbDefaultEncoding;
 
     private boolean isStandalone = DesktopPlatform.isStandalone();
     private Combo browserCombo;
     private Button useEmbeddedBrowserAuth;
 
+    private String userEncoding;
 
     public PrefPageDatabaseUserInterface()
     {
@@ -109,12 +111,22 @@ public class PrefPageDatabaseUserInterface extends AbstractPrefPage implements I
                 GridData.FILL_HORIZONTAL | GridData.VERTICAL_ALIGN_BEGINNING,
                 0
             );
+            Control tipLabel = UIUtils.createInfoLabel(regionalSettingsGroup,
+                CoreMessages.pref_page_ui_general_label_options_take_effect_after_restart
+            );
+            tipLabel.setLayoutData(new GridData(GridData.HORIZONTAL_ALIGN_BEGINNING,
+                GridData.VERTICAL_ALIGN_BEGINNING,
+                false,
+                false,
+                2,
+                1
+            ));
             workspaceLanguage = UIUtils.createLabelCombo(regionalSettingsGroup,
                 CoreMessages.pref_page_ui_general_combo_language,
                 CoreMessages.pref_page_ui_general_combo_language_tip,
                 SWT.READ_ONLY | SWT.DROP_DOWN
             );
-            workspaceLanguage.setLayoutData(new GridData(GridData.HORIZONTAL_ALIGN_BEGINNING));
+            workspaceLanguage.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
             List<PlatformLanguageDescriptor> languages = PlatformLanguageRegistry.getInstance().getLanguages();
             DBPPlatformLanguage pLanguage = DBPPlatformDesktop.getInstance().getLanguage();
             for (int i = 0; i < languages.size(); i++) {
@@ -133,7 +145,7 @@ public class PrefPageDatabaseUserInterface extends AbstractPrefPage implements I
                 CoreMessages.pref_page_ui_general_combo_timezone_tip,
                 SWT.DROP_DOWN
             );
-            clientTimezone.setLayoutData(new GridData(GridData.HORIZONTAL_ALIGN_BEGINNING));
+            clientTimezone.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
             clientTimezone.add(DBConstants.DEFAULT_TIMEZONE);
             for (String timezoneName : TimezoneRegistry.getTimezoneNames()) {
                 clientTimezone.add(timezoneName);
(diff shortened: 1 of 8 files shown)


serjiokov:
Since now, user able to specify require encoding type
image

E1izabeth (commented):
Why the setting is called `File encoding`? Do I correctly understand, that it works not only with files since you're fixing issue with result set?

serjiokov:
I do name as File Encoding because it has correlation to original property.
The **file.encoding** value influences the default charset used by the JVM among other things

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -829,4 +847,56 @@ public static UUID getMixedEndianUUIDFromBytes(byte[] bytes) {
         target.rewind();
         return new UUID(target.getLong(), target.getLong());
     }
+
+    /**
+     * Find index of item in set
+     */
+    public static <T> int getIndexInSet(Set<T> set, @Nullable T value) {
```
serge-rider: Do not add common util methods in GeneralUtils.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -829,4 +847,56 @@ public static UUID getMixedEndianUUIDFromBytes(byte[] bytes) {
...
+        List<String> charsetList = new ArrayList<>();
+        charsetList.add(String.format("%s (%s)", "Default encoding", systemFileEncoding));
+        charsetList.addAll(Charset.availableCharsets().keySet());
+        return charsetList;
+    }
+
+    /**
+     * Return index of charset specification
+     */
+    public static int getIndexOfAvailableCharset(String encoding) {
```
serge-rider: This function doesn't make sense

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -829,4 +847,56 @@ public static UUID getMixedEndianUUIDFromBytes(byte[] bytes) {
...
+        return availableCharsets().indexOf(encoding);
+    }
+
+    /**
+     * Return charset specification by index
+     */
+    public static String getCharsetByIndex(int index) {
```
serge-rider: This too. Just use encoding name from combobox text

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -114,6 +115,23 @@ public static String getDefaultConsoleEncoding() {
...
+        } else {
+            setDeafaultFileEncoding();
+        }
+    }
+
+    public static void setDeafaultFileEncoding() {
+        System.setProperty(StandardConstants.ENV_FILE_ENCODING, UTF8_ENCODING);
+    }
+
+    public static String getFileEncoding() {
```
serge-rider, line 130: This function doesn't make sense

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
serjiokov: Done

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
serjiokov: Done

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
serjiokov: Done

serjiokov:
image

serge-rider (approved):

E1izabeth:
The issues that might be solved with this PR (_**we need to check**_):
https://github.com/dbeaver/dbeaver/issues/20784
https://github.com/dbeaver/dbeaver/issues/20392
https://github.com/dbeaver/dbeaver/issues/20054
https://github.com/dbeaver/dbeaver/issues/19631
https://github.com/dbeaver/dbeaver/issues/9890
https://github.com/dbeaver/dbeaver/issues/8875
https://github.com/dbeaver/dbeaver/issues/6858

E1izabeth (commented):
Won't this preference change will take effect only after DBeaver restart?
If yes we need to add such info label then (like in Regional settings group). And then it would be better to move this option to the upper group to prevent duplication of the info label.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -114,6 +114,23 @@ public static String getDefaultConsoleEncoding() {
...
+    public static void setFileEncoding(String string) {
+        Charset charset = Charset.availableCharsets().get(string);
+        if (charset != null) {
+            System.setProperty(StandardConstants.ENV_FILE_ENCODING, charset.name());
+        } else {
+            setDeafaultFileEncoding();
+        }
+    }
+
+    public static void setDeafaultFileEncoding() {
```
ShadelessFox, line 126: Typo

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPApplication.java
```diff
@@ -102,5 +102,14 @@ default void beforeWorkspaceInitialization() {}
      */
     long getApplicationStartTime();
 
+    /**
+     * Store system encoding specification
+     */
+    public void storeSystemEncoding(String encoding);
```
ShadelessFox, line 108: Please annotate public interface methods with nullability annotations

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageDatabaseUserInterface.java
```diff
@@ -309,4 +341,14 @@ public IAdaptable getElement()
...
     {
     }
+
+    /**
+     * Return available charsets with first item of system encoding
+     */
+    public static List<String> availableCharsets() {
+        List<String> charsetList = new ArrayList<>();
+        charsetList.add(String.format("%s (%s)", "Default encoding", BaseApplicationImpl.getInstance().getSystemEncoding()));
+        charsetList.addAll(Charset.availableCharsets().keySet());
```
ShadelessFox, line 351: I expect this set to be ordered chaotically. Maybe it's better to sort it alphabetically?

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageDatabaseUserInterface.java
```diff
@@ -74,11 +77,13 @@ public class PrefPageDatabaseUserInterface extends AbstractPrefPage implements I
 
     @Nullable
     private Combo clientTimezone;
+    private Combo cmbDefaultEncoding;
```
ShadelessFox, line 80:
```suggestion
    private Combo defaultEncodingCombo;
```

We don't really use abbreviations such as cmb, btn, etc.

Issue #21283, opened 2023-09-22
'Cannot find database for restoring' error appears when opening Restore wizard


DariaMarkaryan:
### Description

1. On a postgres schema call a context menu.
2. Tools -> Restore
3. Shell with error is appeared
![image](https://github.com/dbeaver/dbeaver/assets/43252954/2c8a504d-fe2b-407a-9475-00521beeef7b)
4. But after closing it, the restore wizard opens

### DBeaver Version

23.2.1

### Operating System

win10

PR #21420, opened 2023-10-04
#19054 use global store for native tasks; do not fix file names patte…

### LINKED ISSUE #19054: PostgreSQL - Backup tool - Keep the filename extension in the pattern between two uses
When you use the backup tool on a PostgreSQL database, the pattern used for naming the file dump is kept between two uses of the tool.

But the filename extension isn't... 😢 

For example :
I set the filename field to `$schema_$table.custom`
When I'm using the tool at another time, the filename field is set to `$schema_$table`

This is really anoying because the restoring tool is preset on the `.custom` extension name.

LonwoLonwo:
use the global store for native tasks; 
do not fix file name patterns after loading

Please check the following:

- [x] settings saving between native clients wizards using (restore, backup, MySQL, Postgres) 
- [x] case from the ticket
- [x] again https://github.com/dbeaver/dbeaver/issues/21283
- [x] like https://github.com/dbeaver/dbeaver/issues/21283 but for "Execute script" for PostgreSQL


E1izabeth (approved):

ShadelessFox (approved):

MashaKorax:
verified

PR #21422, opened 2023-10-04
#3052 Add `go to column` command, add search for metadata panel

### LINKED ISSUE #3052: Ctrl + F to find column name in Results Grid
I think a feature to Ctrl + F to find column name in Results Grid will be very useful. I don't think DBeaver can do this currently ? It is only possible to search for data in results grid

Generally in very large tables, it is very tedious to scroll left, right to search for columns.

### LINKED ISSUE #14742: Column Search in the result set - Feature request
Problem Description
  On a daily basis, I deal with tables having columns more than 200-250. And, if I need to look for a certain column, I have to pull the DDL or Transpose the result set and search for the column names that I am looking for and pull those column names to the editor.
  
  The above steps are a bit time-consuming.

Solution
  I would like to have a column search feature in the Metadata section, in the result set.( see below pic)
image

Alternative
  Alternatively, I would like to have a column search feature from the result set. We have this kind of feature in the Aginity tool.

**Additional context**
  For now, a filter feature in the Metadata section would suffice.
  In the future, we can plan for a dedicated section for column search.

E1izabeth:
what was done: https://github.com/dbeaver/dbeaver/issues/3052#issuecomment-1708102730
Shortcut for Go to Column: `CTRL+J`
Shortcut for Go to Row: `CTRL+G` - should work as usual

image

image

image


uslss:
verified

PR #21461, opened 2023-10-06
dbeaver#20446 added UUID Postgre keyword

### LINKED ISSUE #20446: Postgres data types UUID and TEXT not styled in SQL editor
### Description

A minor issue, Postgres data types **TEXT** and **UUID** appear as plain text in SQL code:
image

VARCHAR shown for expected styling.

In PostGRES: 

- UUID is a data type, not a keyword.  
- TEXT is both data type and keyword.

**REFERENCES**: 

Postgres references for [data types](https://www.postgresql.org/docs/15/datatype.html) and [keywords](https://www.postgresql.org/docs/current/sql-keywords-appendix.html)

[dbeaver source code](https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDialect.java) (maybe?)

_Looks like a simple fix, just adding TEXT and UUID as defined "keywords" (though only one is formally defined as such by PostGRES) .  If so, I'm considering having a go at it myself (would be first contribution), but if there's more to it, I won't touch it  ._  

### DBeaver Version

Community 23.1.1.202306251800

### Operating System

Windows 11 21H2 (22000.2057)

### Database and driver

PostgresSQL 15.3
PostgreSQL JDBC Driver postgresql-x64-15

### Steps to reproduce

Write SQL code.
Declare a TEXT or UUID type, or cast to UUID or TEXT type. 
Type defining word appears in plain text

### Additional context

I have DevStyle Themes installed but disabled for the example shown here.  Behavior is the same in DevStyle themes.

guilhermepereira2917:
Added UUID Postgree keyword, the highlighting for TEXT keyword mentioned in the issue was already working.

image


E1izabeth (approved):

E1izabeth:
Thank you for contribution.

LonwoLonwo:
Hello @guilhermepereira2917 

Thanks for your contibution.

But you can see on your screen that data types usually are blue in our SQL Editor. It means, that keywords list is not the correct place to add UUID word.

guilhermepereira2917:
Hi @LonwoLonwo 

Thanks for your feedback, I definitely added it in the wrong place. I think I fixed it now, at least it's the right color, tell me if you want me to change something.

image

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDialect.java
```diff
@@ -676,6 +680,10 @@ public void addExtraKeywords(String... keywords) {
     public void addExtraFunctions(String... functions) {
         super.addFunctions(Arrays.asList(functions));
     }
+    
+    public void addDataTypes(String... dataTypes) {
+        super.addDataTypes(Arrays.asList(dataTypes));
```
LonwoLonwo: But what is the reason to override it here without modifications?

guilhermepereira2917: I think it's more readable this way, and the function for adding functions and extra keywords also have the method overloaded like this, do you still want me to remove it?

LonwoLonwo: Yes, please.

LonwoLonwo (approved):

Matvey16:
Verified

PR #21482, opened 2023-10-10
dbeaver/pro#1907 Grammar improvements

E1izabeth:
Please test that autocompletion works ok for several databases (sqlite, oracle and postgresql, for example). Don't forget to turn on `Enable experimental features` setting.


MashaKorax:
verified

PR #21486, opened 2023-10-10
dbeaver/pro#1993 hide single catalog by default for OpenEdge to avoid…

LonwoLonwo:
hide single catalog by default for OpenEdge to avoid incorrect path creating for tasks

So. For some reason, schemas for OpenEdge have null as catalog.
Because they are read with empty catalog:

https://github.com/dbeaver/dbeaver/blob/da4705b9f759b7997eccd3023cf3418118bef189/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/meta/GenericMetaModel.java#L270-L273

And DBUtils create and use an incorrect path to node after that (null catalog - no entity name in the path for saving, but not for deserializing and finding)

But it seems that we can hide this one single catalog.
https://docs.progress.com/ru-RU/bundle/openedge-database-essentials-117/page/Database-schema-and-metaschema.html Something about the structure

Here are some user's example: https://stackoverflow.com/questions/67500409/how-to-get-all-schema-list-in-progress-openedge-sql-query

**One PR for two tasks:**
https://github.com/dbeaver/pro/issues/1993
https://github.com/dbeaver/pro/issues/1991


E1izabeth (approved):

uslss:
verified

PR #21489, opened 2023-10-10
dbeaver/pro#1966 open in external browser

Destrolaric:
Closes dbeaver/pro#1966


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/DBeaverActivator.java
```diff
@@ -71,7 +72,10 @@ public void start(BundleContext context)
         } catch (MissingResourceException x) {
             coreResourceBundle = null;
         }
-        if (DesktopPlatform.isStandalone() && getPreferenceStore().getBoolean(DBeaverPreferences.UI_USE_EMBEDDED_AUTH)) {
+        if (
```
serge-rider, line 75: Crazy code style

PR #21506, opened 2023-10-12
Update error msg for windows download failed due to cert error

brandonJY:
On commnunity version, dbeaver Custom Truststore feature is not avaliable. Removing the relative message to prevent misleading the users.

Reference: https://github.com/dbeaver/dbeaver/wiki/Managing-Truststore-Settings


E1izabeth:
Thank you for contribution.
The mentioned setting is available in all editions. It allows to switch between windows store and java store.
_"It might help if you haven't overridden trust store."_ is talking only about possibility of changing the trust stores - for example, this way https://github.com/dbeaver/dbeaver/wiki/Importing-CA-certificates-from-your-local-Java-into-DBeaver (works for all editions). So the the removed part of the error message is still useful. I believe we actually don't want to remove it.

brandonJY:
> Thank you for contribution. The mentioned setting is available in all editions. It allows to switch between windows store and java store. _"It might help if you haven't overridden trust store."_ is talking only about possibility of changing the trust stores - for example, this way https://github.com/dbeaver/dbeaver/wiki/Importing-CA-certificates-from-your-local-Java-into-DBeaver (works for all editions). So the the removed part of the error message is still useful. I believe we actually don't want to remove it.

Thanks @E1izabeth . Just want to double check.  Is `Use Windows trust store` features avaliable in all editions?  Can not locate it in either Version 23.2.2 or Version 21.3.5.    
image
image
image

E1izabeth:
This setting is available in all editions and should appear in `Preferences->Connections` if you are running DBeaver under Windows.
image

E1izabeth:
I don't see either this setting on your Connections preference page either some other settings like Connection name pattern with Example result. I doubt why that settings didn't come for you. Could you send the log files? You can find instructions here https://dbeaver.com/docs/wiki/Log-files/.

PR #21513, opened 2023-10-12
dbeaver/pro#1905 UTF-8 is the default file encoding

serge-rider:


E1izabeth:
We have almost the same PR from external contributor https://github.com/dbeaver/dbeaver/pull/21392/files =)

E1izabeth:
Can we merge [it](https://github.com/dbeaver/dbeaver/pull/21392/files) first if everything is okay?

Matvey16:
Verified

PR #21525, opened 2023-10-13
dbeaver/pro#1971 skip parameter numbers used instead names in old PG versions

LonwoLonwo:
Please check procedures DDLs with parameter names and without (just data types) in PG and in our Greenplum. You can check Redshift just to avoid many problems again.


E1izabeth (approved):

serge-rider:
Also test procedure execute (from navigator)

MashaKorax:
verified

PR #21597, opened 2023-10-20
dbeaver/pro#2024 Remote folders create fix

serge-rider:


uslss:
probably the incorrect name is used instead of `create folder`
image

uslss:
verified

PR #21624, opened 2023-10-23
#21618 fix new required columns creating SQL; add base table DDL and …

### LINKED ISSUE #21618: NOT NULL statement is missing in DDL for newly created SQLite tables
### Description

I try to create a SQLite db with tables that have columns with  "NOT NULL" constrains. During column creation via the GUI I check-mark the NOT NULL choice and it gets reflected in the "Columns" section of the "Properties" window (also check-marked).

The DDL has no "NOT NULL" in it though, so once I persist the table my columns can be NULL.

### DBeaver Version

Community Edition, Version 23.2.2.202310081633 (Flatpak)

### Operating System

Linux Ubuntu 22.04.3 LTS (Kubuntu)

### Database and driver

SQLite 

### Steps to reproduce

1. Start DBeaver and open "New Database Connection"
2. Choose "SQLite"
3. Choose "Connect by: Host"
4. Go to "Create..." and create new db file "test.db"
5. Leave everything as is and hit "Finish"
6. In "test.db" go to Tables, right click "Create New Table"
7. In Properties tab choose "Columns" and on lower right part click "Create New Column" 
8. Leave everything as is BUT toggle value of "Not Null" [x]
9. Observe newly created column in the "Columns" tab and the "Not Null" is set
10. Go to "DDL" tab and see that no SQL "NOT NULL" is generated, as well as when clicking on "Save..."

Screenshot_20231021_142843-1

Screenshot_20231021_142955

### Additional context

I tried to use the created database for pre-populating an android room database, but the compiler complained my that my schemas differed (as room needs NOT NULL columns but my tables created with DBeaver kept having NOT NULL = false).

The same steps on a local postgres do generate correct (i.e. "NOT NULL") columns.

LonwoLonwo:
fix new required columns creating SQL; 
add base table DDL and dialect tests for SQLite

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericDataSource.java
+++ b/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericDataSource.java
@@ -518,7 +518,7 @@ public class GenericDataSource extends JDBCDataSource implements DBPTermProvider
                 }
 
                 if (isMergeEntities() || (schemas == null || schemas.isEmpty())) {
-                    this.structureContainer = new DataSourceObjectContainer();
+                    this.structureContainer = new GenericDataSourceObjectContainer(this);
                 }
             }
         } catch (Throwable ex) {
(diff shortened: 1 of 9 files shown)


E1izabeth (approved):

ShadelessFox (approved):

MashaKorax:
verified

PR #21625, opened 2023-10-23
#20984 Add Refresh button for Generate DDL dialog

### LINKED ISSUE #20984: Always generate up-to-date DDL when selecting Generate SQL->DDL
I use right-click->Generate SQL->DDL to get the definition of a view or materialized view, in order to do some changes.
The problem is that this definition is cached from the last time I connected to the database. If in the meantime the view has been updated, I get the old-code instead of the new one.

When I press Generate SQL->DDL, DBeaver should always fetch the most up-to-date DDL, not the cached one.

I know that I can first refresh the views, but it's really aggravating to have to do those additional steps every time I need to do a change (and in my current workflow, that's very frequent).

E1izabeth:
image


LonwoLonwo (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #21634, opened 2023-10-23
dbeaver/pro#1918 Check if scheduler supports opening external settings

ShadelessFox:


plugins/org.jkiss.dbeaver.tasks.ui.view/plugin.xml
```diff
@@ -27,6 +27,12 @@
             namespace="org.jkiss.dbeaver.task"
             properties="scheduled,editable"
             type="org.jkiss.dbeaver.model.task.DBTTask"/>
+        <propertyTester
```
serge-rider: I don't like property testes which accept any objects in input. They produce a mess inside Eclipse registries.
Also maybe we should have property tested in ee only?

serge-rider (changes requested):

serge-rider (approved):

PR #21642, opened 2023-10-24
Remove outdated docs/build.txt

mfussenegger:
The file mentions Java 1.8/Maven 2 which seems outdated.
The readme links to:

  https://github.com/dbeaver/dbeaver/wiki/Build-from-sources


LonwoLonwo:
Hello @mfussenegger 

Thanks for your contribution.

But maybe you can just upgrade version numbers, not delete the entire file?

mfussenegger:
Why keep the file if there is a second place where it's documented?
Given how long it has remained stale, it doesn't seem like anybody looked at it.
The `build.txt` isn't linked to from anywhere else as far as I could tell.

serge-rider:
Agreed

PR #21662, opened 2023-10-25
#21373 Add completion for update statements

### LINKED ISSUE #21373: Code completion for update statements
### Description

We only use MySQL, so I'm not sure if this issue happens with other databases.

When typing an update query, with the previous version (23.2.0) you would get autocompletion when you used an alias. This was already a step down from earlier in version 23.1.x when you had autocomplete even without an alias.

Now, however, autocomplete doesn't work at all anymore when you type an update query. Look at my examples below. The same query has autocomplete with a select query, but not with an update query.

image
image

This happens on joins, but also on where-statements:
image

Could this please be fixed? This makes typing query's so much harder. 
If possible, could the autocomplete also work without a required alias like it did in version 23.1.x?

Thanks in advance!

### DBeaver Version

23.2.1

### Operating System

Windows 10

### Database and driver

MySQL 8
com.mysql.cj.jdbc.Driver
8.0.29

### Steps to reproduce

Type an update query, and try to autocomplete a join or where-statement

### Additional context

_No response_

E1izabeth:


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/analyzer/TableReferencesAnalyzer.java
```diff
@@ -38,7 +38,7 @@ public interface TableReferencesAnalyzer {
      * The method returns map of table by alias base on query expression
      */
     @NotNull
-    Map<String, String> getTableAlicesFromQuery(@NotNull String query);
+    Map<String, String> getTableAliacesFromQuery(@NotNull String query);
```
ShadelessFox:
```suggestion
    Map<String, String> getTableAliasesFromQuery(@NotNull String query);
```

LonwoLonwo (approved):

ShadelessFox (approved):

uslss:
values from priveously mentioned column are proposed after comma
sqlite:
image
pg:
image

E1izabeth:
As far as I see it's reproducible on the previous version with experimental feature off 
image

image


Please check. I suspect this behavior didn't come with this PR.

uslss:
verified

PR #21693, opened 2023-10-30
dbeaver/dbeaver#21278 add left align to title and add spaces to title

### LINKED ISSUE #21278: Improve query readability on HTML export by aligning text to the left
Hey! As a data professional, I heavily use this "Copy as HTML" feature. (Thank you)
Ref.: [DataExporterHTML.java#L94](https://github.com/dbeaver/dbeaver/blob/e7989f5547716d31d1ec8d10247b9079173c730e/plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterHTML.java#L94)

When the query is exported along with the data, the text-align: center makes the query super unreadable. Changing that to "left" would already be a significant improvement.

I always have to edit this before sharing the query results with others. Sometimes, the tool I paste the results as HTML can't edit the alignment (e.g.: MS Teams), so I have to delete the query from the export and then copy and paste the query separately.

I know this is also used for the table name when the results are exported from a table. The table name would look better aligned to the left as well.

Just change the alignment to: left
Bonus points: export the query as coloured source code using [HTMLSQLConverter.java#L38](https://github.com/dbeaver/dbeaver/blob/e7989f5547716d31d1ec8d10247b9079173c730e/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/convert/impl/HTMLSQLConverter.java#L38)

I usually erase the query in the export, then copy and paste it again and share it outside the HTML table provided.

**Additional context**
Below are screenshots of the "Copy as HTML" when pasted to MS Teams. The query is only readable because it is simple. But the screenshots show that the query does not even appear unless you scroll to the center.
Snag_2e76ddd0
Snag_2e76f65a
Snag_2e770a7e

Destrolaric:
Closes dbeaver/dbeaver#21278
image

### DIFF
--- a/bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
+++ b/bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
@@ -933,11 +933,13 @@ public class CommonUtils {
         }
         return text
             .replace("&", "&amp;")
+            .replace(" ", "&nbsp")
             .replace("<", "&lt;")
             .replace(">", "&gt;")
             .replace("\r\n", "<br>")
             .replace("\r", "<br>")
-            .replace("\n", "<br>");
+            .replace("\n", "<br>")
+            .replace("\t", "&emsp");
     }
 
     /**
--- a/bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
+++ b/bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
@@ -939,7 +939,7 @@ public class CommonUtils {
             .replace("\r\n", "<br>")
             .replace("\r", "<br>")
             .replace("\n", "<br>")
-            .replace("\t", "&emsp");
+            .replace("\t", "&emsp;");
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ShadelessFox:
Wrong issue ID in the commit's message

bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
```diff
@@ -933,11 +933,13 @@ public static String escapeHtml(@Nullable String text) {
         }
         return text
             .replace("&", "&amp;")
+            .replace(" ", "&nbsp")
```
ShadelessFox, line 936: Is the semicolon optional here?

bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
```diff
@@ -933,11 +933,13 @@ public static String escapeHtml(@Nullable String text) {
...
             .replace("<", "&lt;")
             .replace(">", "&gt;")
             .replace("\r\n", "<br>")
             .replace("\r", "<br>")
-            .replace("\n", "<br>");
+            .replace("\n", "<br>")
+            .replace("\t", "&emsp");
```
ShadelessFox: And here

bundles/org.jkiss.utils/src/org/jkiss/utils/CommonUtils.java
Destrolaric: Works without semicolon, as shown at screenshot

ShadelessFox: Although some named characters in HTML don't have the semicolon, the `&emsp;` is an exception.

Please see https://html.spec.whatwg.org/multipage/named-characters.html

ShadelessFox (approved):

Matvey16:
Verified

PR #21699, opened 2023-10-30
CB-3844 tm use qm

yagudin10:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/auth/SMSessionUtils.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.jkiss.dbeaver.DBException;
+import org.jkiss.dbeaver.model.app.DBPWorkspace;
+import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
+import org.jkiss.dbeaver.runtime.DBWorkbench;
+
+public class SMSessionUtils {
+
+    public static SMSession getSpaceSession(DBRProgressMonitor monitor) throws DBException {
+        DBPWorkspace workspace = DBWorkbench.getPlatform().getWorkspace();
```
serge-rider: We have identical code in EnterpriseWorkbenchWindowAdvisor and in QMUtils.
Maybe it is better to move it to QMUtils and make more generic?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/auth/SMSessionUtils.java
yagudin10: I extracted this method from EnterpriseWorkbenchWindowAdvisor to utils class. I don't think that it's good to add method to QMUtils because SMSession is not related to QM.

alexander-skoblikov (approved):

serge-rider (approved):

PR #21704, opened 2023-10-30
dbeaver/dbeaver#21522 Message bar content improved, added logic to hide/show panel.

### LINKED ISSUE #21522: Grammar improvement "Warning: Content representation size is limited by: 100 KB"
### Description

New version does not show full value in 
Result Panels -> Value 
anymore.
Instead a warning is displayed in red below the area
"Warning: Content representation size is limited by: 100 KB"

Screenshot_20231013_112211

Linguistic issues (it's hardly limited "by" but rather "to") aside -- that limited hasn't been there before, but the panel was an easy way to overcome the limit in the Result View, where the value length was limited already.

I think the value length should not be limited in the Panel at all.
If that limit cannot be waived at all (why not, though -- this can't just be the same data as in the View, but has to be fetched separately, hasn't it?) it has to be configurable by the user.

In fact, there's now no (easy) way to fetch the full value at all.

### DBeaver Version

Community Edition 23.2.2.202310081633

### Operating System

Linux, Debian/Sid

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

serjiokov:
Message bar content improved, added logic to hide/show panel.
1.  Show panel when limit reached 
 
image

2. Hide for all other cases
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
@@ -149,6 +149,21 @@ public abstract class AbstractTextPanelEditor<EDITOR extends BaseTextEditor>
         return messageBar;
     }
 
+    private void hideMessageBar() {
+        GridData gridData = new GridData(SWT.FILL, SWT.FILL, true, false, 1, 1);
+        gridData.heightHint = 0;
+        messageBar.setLayoutData(gridData);
+        messageBar.getParent().layout();
+    }
+
+    private void showMessageBar(String msg) {
+        messageBar.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false, 1, 1));
+        messageBar.setForeground(UIStyles.getErrorTextForeground());
+        messageBar.setText(msg);
+        messageBar.setImage(UIUtils.getShardImage(ISharedImages.IMG_OBJS_WARN_TSK));
+        messageBar.getParent().layout();
+    }
+
     protected abstract EDITOR createEditorParty(IValueController valueController);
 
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -149,6 +149,21 @@ private CLabel createMessageBar(Composite cmpsBase) {
         return messageBar;
     }
 
+    private void hideMessageBar() {
+        GridData gridData = new GridData(SWT.FILL, SWT.FILL, true, false, 1, 1);
+        gridData.heightHint = 0;
+        messageBar.setLayoutData(gridData);
+        messageBar.getParent().layout();
```
ShadelessFox: Use the `GridData.exclude` flag on the existing layout instead of creating a new one every time.

Maybe you don't even need to create that message if the content size doesn't exceed the limit?

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -149,6 +149,21 @@ private CLabel createMessageBar(Composite cmpsBase) {
...
+    }
+
+    private void showMessageBar(String msg) {
+        messageBar.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false, 1, 1));
+        messageBar.setForeground(UIStyles.getErrorTextForeground());
+        messageBar.setText(msg);
+        messageBar.setImage(UIUtils.getShardImage(ISharedImages.IMG_OBJS_WARN_TSK));
```
ShadelessFox, line 167: Please don't use outdated Eclipse icons. Use ours: `DBIcon#STATUS_WARNING`

ShadelessFox (changes requested):

Matvey16:
Verified

PR #21714, opened 2023-10-31
#21151 do not update attribute names for quoted attributes

### LINKED ISSUE #21151: When using Data Transfer to import a file, doublequotes in "Tables Mapping" are accepted... then ignored.
### Description

I tried to import a CSV file into Postgres through the Data Transfer dialog. My column headings are mixed case, and all are doublequoted in the CSV file.
Under 'Tables mapping', in the 'Source' column, some of the headings appeared in doublequotes, but most didn't. I noticed that editing the name in the 'Target' field, by putting doublequotes around the name, the name retained its mixed case. After editing some 240 column names (yes, this is a whinge!), I clicked through to 'Proceed' only to find that my mixed case headings had returned to lowercase, causing the TABLE CREATE to fail due to duplicate column names.

### DBeaver Version

Community Edition 23.2.0

### Operating System

OpenSUSE Tumbleweed

### Database and driver

Database: PostgreSQL 15.4
Driver: JDBC 42.5.2

### Steps to reproduce

See description!

### Additional context

_No response_

LonwoLonwo:
2023-10-31 13_14_36-Data Transfer

2023-10-31 14_16_45-Data Transfer

You need to test many cases. With different upper/camel/lower case attribute names. Names like keywords or with special symbols inside. From lower case databases to upper case. And back. From CSV files.


uslss:
verified

PR #21729, opened 2023-11-02
#21720 add the drop statement part for routines definition

### LINKED ISSUE #21720: Generate drop comment into function DDL
In Postgres, when I generate a function DDL script, there is no commented drop SQL command for that function in the script. When I generate DDL for a table, there it is.  The only way I can generate a drop command for a function is by clicking "delete" and then "view script". It would be more convenient and safer to generate it as a commented part of a DDL script.

LonwoLonwo:
It will work only with the "Show header" command

You can check it on Redshift and Greenplum as well.

2023-11-02 12_21_47-DBeaver Ultimate 23 2 3 - add_em(int4, int4)


E1izabeth (approved):

ShadelessFox (approved):

MashaKorax:
verified

PR #21752, opened 2023-11-03
dbeaver/pro#2063 NIO FS events propogation fix

serge-rider:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/fs/DBNFileSystem.java
```diff
@@ -96,7 +96,7 @@ protected void dispose(boolean reflect) {
 
     @Override
     public String getNodeType() {
-        return NodePathType.dbvfs.name() + ".fileSystem";
```
alexander-skoblikov, line 99: `dbvfs.fileSystem ` used in CloudBeaver to determinate the required authentication.

MashaKorax:
verified

PR #21758, opened 2023-11-06
#21757 support conditional NULL/NOT NULL modifiers

### LINKED ISSUE #21757: [Oracle] Change type of a column adds unecessary "NULL" at the end of the alter column query
### Description

I think this line https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableColumnManager.java#L81C46-L81C46 is wrong because in oracle you don't need to add "NULL" at the end

Example
ALTER TABLE SCHEMA_NAME.TABLE_NAME MODIFY COLUMN_NAME NUMBER(5,0) NULL;
it should be simply
ALTER TABLE SCHEMA_NAME.TABLE_NAME MODIFY COLUMN_NAME NUMBER(5,0);
or
ALTER TABLE SCHEMA_NAME.TABLE_NAME MODIFY COLUMN_NAME NUMBER(5,0) NOT NULL;

### DBeaver Version

23.2.3.202310221533 community

### Operating System

Win 10

### Database and driver

Oracle Latest driver

### Steps to reproduce

Alter a column type of an oracle table which is nullable (so it's not "NOT NULL")

### Additional context

_No response_

LonwoLonwo:
2023-11-06 10_09_17-NUMBER_ONE - Persist Changes

Now, the NULL/NOT NULL modifier will be added only if it is changed.

Please check for existing and new columns.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/edit/OracleTableColumnManager.java
+++ b/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/edit/OracleTableColumnManager.java
@@ -81,9 +81,8 @@ public class OracleTableColumnManager extends SQLTableColumnManager<OracleTableC
         return object.getParentObject().getContainer().tableCache.getChildrenCache(object.getParentObject());
     }
 
-    protected ColumnModifier[] getSupportedModifiers(OracleTableColumn column, Map<String, Object> options)
-    {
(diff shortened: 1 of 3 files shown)


E1izabeth (approved):

ShadelessFox (approved):

angeousta (approved):

MashaKorax:
verified

PR #21778, opened 2023-11-07
#21773 fix quoted columns search for correct mapping

### LINKED ISSUE #21773: 23.2.4 - Issue with Importing Data from a CSV File
### Description

When I try to import data from a CSV file, the mapping of the columns from the CSV file to an existing table does not set the "Mapping" field to "existing" if the table's column name contains a space. 

I tested my diagnosis by changing the column names in the existing table by replacing the spaces with an underscore and the column mapping worked fine during the data import process.

### DBeaver Version

23.2.4

### Operating System

Sonoma 14.0

### Database and driver

Database: MySQL 8.0.33
Driver: MySQL

### Steps to reproduce

Here is a video showing the actions I took to experience this bug: https://app.screencast.com/wpOVTr8aMDC8m

1. Right-click table name and choose "Import Data"
2. Choose CSV and click the "Next >" button
3. Select the CSV file and click the "Open" button
4. Click the "Next >" button
5. Click the arrow (>) in front of the Source data file name
6. Click the Target column name and choose a column name that contains a space, such as: `First Name`
7. The Mapping column for that row does not change to "existing", it remains at "create" which causes an error during the import process
image

### Additional context

_No response_

LonwoLonwo:
Please check

- [ ] transfer from csv into existing table and columns mapping
- [ ] transfer from another table into an existing table and columns mapping

2023-11-07 10_58_35-Data Transfer


Matvey16:
Verified

Issue #10658, opened 2020-12-10
Support more distinct SQL Editor text colors


lukaseder:
#### System information: 
- Operating system (distribution) and version
Microsoft Windows [Version 10.0.18363.1198]

- DBeaver version
7.3.0.202011291229

- Java version 
OpenJDK Runtime Environment (build 14-ea+27-1339)

- Additional extensions
N/A

#### Connection specification:
N/A

#### Describe the problem you're observing:
As discussed here:
https://twitter.com/dbeaver_news/status/1336744981789364224

I accidentally renamed a column because of a missing comma:

```sql
SELECT
  col1 -- Missing comma here
  col2,
  col3
FROM t
```

It took me way longer than it should have to see my mistake. More sophisticated syntax highlighting as e.g. in IntelliJ could have helped spot the problem more quickly. An example can be seen here:
https://twitter.com/missingfaktor/status/1336734896908349443

![image](https://user-images.githubusercontent.com/734593/101748578-04e8bb80-3acd-11eb-8ab9-38673edc07ad.png)

I'm not sure if the colors should be so "aggressive" by default, but having the option would certainly be useful. Not just for aliases, but also for other things. Often, a column reference has the same name as a table reference, and it would be interesting to be able to visually distinguish between the two types of object.

uslss:
Thanks for suggestion

E1izabeth:
It's done in https://github.com/dbeaver/dbeaver/pull/21788 and will be available in the next release (DBeaver 23.2.5)
The feature is turned off by default, you can enable it here 

![image](https://github.com/dbeaver/dbeaver/assets/28875055/dca4fef3-3bd6-484b-a11b-21cf916d1423)

E1izabeth:
Done in scope of this issue #10831

Issue #4099, opened 2018-08-28
Feature request: table name color in editor


jpina10:
Hey, like on TOAD for exemple, when I type a table name in the schema it changes is color. It's very good when I have a very long query with many joins

serge-rider:
Agreed, this would be handy.
Although it is not so easy because we'll need to cache all database table names when opening a SQL editor.

jpina10:
So TOAD caches all database table names?!That seems to be 'a big price to pay' just to have the color

p91paul:
@jpina10 that would explain why TOAD is so freaking slow :)

serge-rider:
Potentially this could be implemented without full table list load. Property SQL semantic analysis + table name cache might solve this.
We have SQL analysis improvement in plans. But it is much bigger feature than just table name highlighting.
TOAD needs to know only PL/SQL while DBeaver will need to understand all existing SQL dialects.

jpina10:
> 
> 
> @jpina10 that would explain why TOAD is so freaking slow :)

I have the 10.1 version and it's really fast. Dbeaver has a lot more features that is why I'm willing to change

p91paul:
Really? I have used versions from 9 to 12, and opening a new editor or worse, a new object explorer, to not talk about a new connection, the whole ui is locked for seconds while dbeaver is almost istantaneous. Moreover, TOAD excels in how it is able to crash or go in a deadly freeze from which it never wakes up.

I still miss some of TOAD dbadmin features (there are open issues for most of them, but I understand those are not priority goals for dbeaver since those would be features tied to a single database), but I strongly prefer dbeaver because it _feels_ way faster and it's amazingly more user-friendly when querying and editing data :smile: .

DavidMCook:
This would be tremendously helpful in a project I'm running right now. So much so that I'm actually considering finding a new IDE, even though I love DBeaver.

akaCarbone:
Seriously that there is no evolution on this since 2018?
This is VERY useful and a common feature on many other SQL tools.

It shoud have more traction than that...

E1izabeth:
It's done in https://github.com/dbeaver/dbeaver/pull/21788 and will be available in the next release (DBeaver 23.2.5)
The feature is turned off by default, you can enable it here 

![image](https://github.com/dbeaver/dbeaver/assets/28875055/dca4fef3-3bd6-484b-a11b-21cf916d1423)

E1izabeth:
Done in scope of this issue #10831

PR #21788, opened 2023-11-08
#10831 Color column and table names separately

### LINKED ISSUE #5714: Syntax coloring: table names
Is there any way to set tables names coloring in query editor ?

This is very handy feature - eg you can easily see opening bigger script if table exists or not.

### LINKED ISSUE #10831: Color column and table names separately
When I code review, I spend more energy than I want looking for the period separating columns from tables. Our standard is all lowercase and full table names. It makes the code easier to share but also makes the lines longer to read and more cumbersome to navigate.

I'd like to specify the color of the table separately from the column name. By greying out the table name, I can treat it as context and not give it as much visual prominence.

image

I use external text editors when code reviewing as I can control the colors of these two pieces. I have used both VS Code & Atom (shown above).

### LINKED ISSUE #11549: SQL parser: support semantic rules for tables, columns and aliases

### LINKED ISSUE #19641: Enhance highlighting with the parse tree
We have parse tree by the grammar. Let's enhance our highlighting as it could be more "smart" with this knowledge. 

Maybe we may focus on these issues idea #10658 #7395

_______________

https://github.com/dbeaver/pro/wiki/Plans-on-ANTLR-based-parser-and-editor-services

E1izabeth:
image

image

**Should be tested against select clauses.**

How does it work from the user view:
- it highlights the column name and alias, the table name and alias and the schema name.
- if you used something in selection list that is not present in from clause, it will be highlighted with red color like an error
for example, here b is missing in from clause
image
- if you established connection to the database and use something that is missing in metadata context, it will be highlighted as an error too

### The feature is turned on by default. Metadata usage is turned on by default too. Also I turned on experimental autocompletion.

image

image

### Granular re-parsing on editing will be finished in a separate PR.

_P.S. Primary logic of the OffsetKeyedTreeMap seems implemeted for now and requires test coverage now. Highlighting granular update on input will be introduced as soon as OffsetKeyedTreeMap will be tested as stable._

_P.P.S. There is a wide range for the future improvement - adding more features, I have a huge list of ideas and ToDo-s_

Also related issues #11549 #5714 #4099 #10658 #19641


uslss:
verified

PR #21792, opened 2023-11-08
dbeaver/dbeaver#21779 Hana dialect enhancement with tests

### LINKED ISSUE #21779: Identifiers with special characters are not escaped
### Description

I have a table with some special characters (e.g. `Ĕ` and `Ʋ`). I am not able to work at all with it (not able to view the data or to delete it). It seems that all queries that DBeaver generates do not escape the table name.

### DBeaver Version

23.2.4

### Operating System

any

### Database and driver

1. SAP HANA Cloud 2023.4.35
2. com.sap.db.jdbc.Driver 2.18.13

### Steps to reproduce

_No response_

### Additional context

If I right click on the table and choose Delete, the following SQL query is executed:
```sql
DROP TABLE MY_SCHEMA.Y105ĔC4Ʋ6FBCAA03;
```

This is incorrect SQL, since the identifier contains special characters which need escaping. The correct query would be:
```sql
DROP TABLE MY_SCHEMA."Y105ĔC4Ʋ6FBCAA03";
```

LonwoLonwo:
2023-11-08 13_46_26-Delete script

You can check objects creation/modification/deleting for non-Latin symbol names

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANASQLDialect.java
+++ b/plugins/org.jkiss.dbeaver.ext.hana/src/org/jkiss/dbeaver/ext/hana/model/HANASQLDialect.java
@@ -211,4 +211,15 @@ public class HANASQLDialect extends GenericSQLDialect implements TPRuleProvider 
     public boolean isStripCommentsBeforeBlocks() {
         return true;
     }
+
+    @Override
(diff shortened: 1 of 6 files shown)


E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #21800, opened 2023-11-09
#19188 Strip comments before JSQL parser parsing

### LINKED ISSUE #19188: Columns are read-only if query has comment block before
### Description

If there's "#\n" before the sql, its query result will not be editable even if it has a primary key.
For example:
\#
select * from foo

### DBeaver Version

Community Edition 22.3.5

E1izabeth:
image


LonwoLonwo (approved):

ShadelessFox (approved):

uslss:
verified

PR #21811, opened 2023-11-10
(fix) misspelled words in Pt-BR translation

### LINKED ISSUE #21786: Grid/Record in portuguese is translated Grade/Gravar, when It would make more sense Grade/Registro or Grade/Ficha
Gravar in portugues is save, which appears to be a out of context translation of Record.

What one would want is Registro that leads to Record but not as Saving or even better (in my opinion) Ficha that points to a paper with data of a entity.

Regards,

Geraldo Lopes de Souza

brunodmartins:
Closes #21786 

Screenshot from 2023-11-09 21-46-05
Screenshot from 2023-11-09 21-46-27


E1izabeth:
Thank you for contribution!

E1izabeth (approved):

LonwoLonwo (approved):

MashaKorax:
verified

PR #21814, opened 2023-11-10
#21697 Auto adjust result's grid size to the row's count

### LINKED ISSUE #21697: Improve vertically display by adjusting windows size for results with few lines
Hi,

Dbeaver introduce recently the vertically display  which allows you to display the results of several queries one above the other (https://github.com/dbeaver/dbeaver/issues/6861)

As discussed in https://github.com/dbeaver/dbeaver/pull/21280, would it be possible to improve this fonctionnality with an  "intelligent" auto-sizing of the results. Something like "size of each result" = min(returned rows, 10)
This would allow us to better understand the different query results without having to scroll when they only return a few lines.
"10" should be an option whose value would be left to the user's choice

Regards,
Thanks,

E1izabeth:
image


ShadelessFox (commented):
It looks a bit weird with the example containing just one row.

Maybe show **at least** N rows, like 5?

E1izabeth:
I reverted removing filters and added minimal height

LonwoLonwo (approved):

ShadelessFox (approved):

uslss:
verified

PR #21833, opened 2023-11-13
dbeaver/pro#2071 ERD overlapping paths should be splitted.

serjiokov:
Adjusted routing parent class for collision detection.


serjiokov:
The PR closed in iorder to keep original branch.

PR #21845, opened 2023-11-14
dbeaver/pro#2103 Fix errors appeared during highlighting

E1izabeth:
See the issue description.
Also check UPDATE statements briefly, I changed the way of handling incomplete queries by parser.
(update Fruits fr set fr.| <- completion should work here)


Matvey16:
Verified

PR #21186, opened 2023-09-12
Add support driver for Dameng

Stkai:
In China, Dameng is a very popular database, many governments and state-owned enterprises use the Dameng database, PR for the Dameng dialect.
[Dameng Offical Website](https://www.dameng.com/)
Hopefully it can be merged.
Thanks.

PR #825, opened 2016-10-08
Enhanced Support for Exasol

Sargul:
Hi,

I've been working on Exasol support the past days and finished my first working release.

What's supported:
- Viewing of Tables, Scripts, Views, Users
- DDL for Tables, Scripts, Views
- Session Management

I'm stilling missing the maven pom file. I'm having difficulties generating this.

TODOs: Support for Roles, Admin Recompress Tables, Reorganize Tables ...

Kind Regards

Karl Griesser

PR #21847, opened 2023-11-14
dbeaver/dbeaver#21832 Add Cubrid Plugin Features

### LINKED ISSUE #21832: Add Cubrid Plugin Features
- Driver download from maven repository 
- Connection properties
- Create folders to categorized by user
- Display list of user
- Display table based on user
- Display system table
- Display view based on user
- Display system view

longhaseng52:
- Driver download from maven repository 
- Connection properties
- Create folders to categorized by user
- Display list of user
- Display table based on user
- Display system table
- Display view based on user
- Display system view


Destrolaric:
@longhaseng52, Hi, it looks like currently the UI components, like the connection page, are stored in the same plugin. They should be transferred to the `org.jkiss.dbeaver.ext.cubrid.ui` plugin.

serge-rider:
Note: you cannot add any dependencies on *ui bundles in the main plugin.
We use the same codebase for headless tests and for CloudBeaver build where desktop UI is not available. That's why all database specific extensions are divided on two bundles.

longhaseng52:
Thank you for guidance. I will create another commit which include "org.jkiss.dbeaver.ext.cubrid.ui" and transfer UI component to it.

LonwoLonwo:
Additionally, please use JAVA inheritance practice in your code. No need to copy all logic from classes GenericSomething. Just add GenricParentClass extending. Or use generic bundle classes instead of yours. Like GenericUtils instead of CubridUtils. 
I see it everywhere. It is harder for us to check your code, and has no meaning at all.
You can take a look at this PR as an example: https://github.com/dbeaver/dbeaver/pull/21186

longhaseng52:
> Additionally, please use JAVA inheritance practice in your code. No need to copy all logic from classes GenericSomething. Just add GenricParentClass extending. Or use generic bundle classes instead of yours. Like GenericUtils instead of CubridUtils. I see it everywhere. It is harder for us to check your code, and has no meaning at all. You can take a look at this PR as an example: #21186

Thanks for you suggestion. Actually, my team has planned to update the Cubrid plugin feature continuously from now on. We wanted to own a plugin structure because in the next upcoming PR, there will be a modification to those files. And I think our case is kind of similar case to #825.

LonwoLonwo:
> Actually, my team has planned to update the Cubrid plugin feature continuously from now on. We wanted to own a plugin structure because in the next upcoming PR, there will be a modification to those files.

Big plans, great.
But it sounds like a risk for us. Maybe you will come up with future changes in your code. And maybe - not.
And support of your code will be on our side. For us, it is better to not support two identical code places in our code base.

longhaseng52:
> > Actually, my team has planned to update the Cubrid plugin feature continuously from now on. We wanted to own a plugin structure because in the next upcoming PR, there will be a modification to those files.
> 
> Big plans, great. But it sounds like a risk for us. Maybe you will come up with future changes in your code. And maybe - not. And support of your code will be on our side. For us, it is better to not support two identical code places in our code base.

Now, I understand your concern. We will refactor our code to follow your suggestion. We might need some time so we will close this PR for now, after completed we will create a new PR. Thank you for your time to review our code.

LonwoLonwo:
Will wait for your new PR 🤗

PR #21859, opened 2023-11-14
#21666 Single tab mode improvements

### LINKED ISSUE #21666: Singletab mode improvements
We want to
- [x] Add setting for removing filters in single tab mode
- [x] Extend functionality not only for Script execution
- [x] Remove warning about parallel execution
- [x] Reuse existing tab on new execution
- [x] Rename command to Show results in a single/multiple tab(s)
- [x] Think about scrolling

Auto calculate the size of the grid by the rows number will be done in a separate issue (#21697)

E1izabeth:
- [x] Add setting for removing filters in single tab mode
image

- [x] If you select several queries and press `Execute SQL Query` button, they should be handled as a script if there is more than one query
javaw_6hFAXbthuF

- [x] Remove warning about parallel execution
- [x] Reuse existing tab on new execution
javaw_ueezUPgVwU

- [x] Command should be called `Show results in a single/multiple tab(s)`
image

- [x] If you scroll over result grid, then the whole tab with several result sections should be scrolled instead of grid with data under the cursor

javaw_aprYDFhgtA


plugins/org.jkiss.dbeaver.ui.editors.sql/OSGI-INF/l10n/bundle.properties
```diff
@@ -105,7 +105,7 @@ command.org.jkiss.dbeaver.ui.editors.sql.show.variables.name=Show SQL variables
 command.org.jkiss.dbeaver.ui.editors.sql.show.variables.description=Show active SQL variables
 command.org.jkiss.dbeaver.ui.editors.sql.toggle.extraPanel.name=Show panels in result tabs
 command.org.jkiss.dbeaver.ui.editors.sql.toggle.extraPanel.description=Show panels in result tabs instead of SQL editor pane
-command.org.jkiss.dbeaver.ui.editors.sql.multipleResultsPerTab.name=Toggle execution results orientation
+command.org.jkiss.dbeaver.ui.editors.sql.multipleResultsPerTab.name= Show results in a single/multiple tab(s) Show results in a single/multiple tab(s)
```
ShadelessFox:
```suggestion
command.org.jkiss.dbeaver.ui.editors.sql.multipleResultsPerTab.name= Show results in a single/multiple tabs
```

DariaMarkaryan:
Ferified

PR #21861, opened 2023-11-15
dbeaver/pro#2086 Granular re-parsing and re-highlighting

E1izabeth:
Setting for highlighting is turned off by default now, so turn it on for testing.

Please check highlighting on text editing.


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/SQLDocumentScriptItemSyntaxContext.java
```diff
@@ -0,0 +1,63 @@
+package org.jkiss.dbeaver.ui.editors.sql.semantics;
```
LonwoLonwo, line 17: Copyright 👼

uslss:
veriffied

PR #21884, opened 2023-11-16
Define maxZoom as maxNativeZoom

kalkun:
This changes the existing `maxZoom` properties to `maxNativeZoom` properties for Leaflet TileLayers. This ensures that the map will not load tiles beyond the defined max zoom level, but will allow the map to be zoomed in beyond this point. I put `maxZoom` as 24 for existing layers. It allows geometries visualised on a map at a closer zoom, while keeping the last defined zoom layer as the background.


LonwoLonwo:
Thanks for the contribution.
We will look into it.

kalkun:
Some sources don't specify a `maxZoom` property explicitly so I suppose the default `18` is used then. The above could be added to these layers too e.g. for the Arcgis Cosmo layer to allow zoom to 24 for all default sources.

serjiokov (approved):

Matvey16:
Verified

PR #21886, opened 2023-11-17
Update MySQL (Old) jdbc driver

m1ngyuan:
Update mysql jdbc driver

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
@@ -1423,7 +1423,7 @@
                         webURL="https://www.mysql.com/products/connector/"
                         supportedConfigurationTypes="MANUAL,URL"
                         description="Driver for MySQL (4.x and earlier)">
-                    <file type="jar" path="maven:/mysql:mysql-connector-java:RELEASE"/>
+                    <file type="jar" path="maven:/mysql:mysql-connector-java:RELEASE[5.1.49]"/>
                     <property name="rewriteBatchedStatements" value="true"/>
                 </driver>
 


LonwoLonwo:
Hello @m1ngyuan 

Thanks for your contribution.
We will look into it.

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -1423,7 +1423,7 @@
                         webURL="https://www.mysql.com/products/connector/"
                         supportedConfigurationTypes="MANUAL,URL"
                         description="Driver for MySQL (4.x and earlier)">
-                    <file type="jar" path="maven:/mysql:mysql-connector-java:RELEASE"/>
+                    <file type="jar" path="maven:/com.mysql:mysql-connector-j:RELEASE"/>
```
LonwoLonwo:
```suggestion
                    <file type="jar" path="maven:/com.mysql:mysql-connector-j:RELEASE[8.2.0]"/>
```

LonwoLonwo: Please add a specific version. We can not be sure that the next release will not have vulnerabilities.

LonwoLonwo (approved):

serge-rider:
https://github.com/dbeaver/dbeaver/issues/21906
Now I am not sure that we shoulod update legacy driver. It may break backward compatibility

LonwoLonwo:
Because of this ticket #21906, we are not sure about backward compatibility.

My suggestion here is not to change the link for the latest but to use a specific old MySQL driver version:

maven:/mysql:mysql-connector-java:RELEASE[5.1.44]

m1ngyuan:
> Because of this ticket #21906, we are not sure about backward compatibility.
> 
> My suggestion here is not to change the link for the latest but to use a specific old MySQL driver version:
> 
> maven:/mysql:mysql-connector-java:RELEASE[5.1.44]


Thank you for your advice. I believe you are right. 
Here are the relevant release notes. https://downloads.mysql.com/docs/connector-j-5.1-relnotes-en.pdf 
and I think we should upgrade to 5.1.49, which is the latest GA version that fixes some bugs in the MySQL 5. x branch

LonwoLonwo (approved):

DariaMarkaryan:
Verified

PR #21917, opened 2023-11-20
#21898 Don't separate selected region into several queries

### LINKED ISSUE #21898: Error occurred during SQL Server CTE expression execution after upgrading to version 23.2.5.202311191730
### Description

After upgrading to version 23.2.5 (Community Edition), an error 'Must declare the scalar variable "@VAR_NAME"' occurs when trying to execute the following CTE expression.

image

I have tried solution mentioned in issue #15543 , doesn't work.

```
IF OBJECT_ID('tempdb..#TempTb') IS NOT NULL DROP TABLE #TempTb

CREATE TABLE #TempTb (Idx INT IDENTITY, Content NVARCHAR(200) NULL, BizTime DATETIME NULL )

INSERT INTO #TempTb(Content, BizTime)
SELECT N'A', N'2023-11-01'
UNION ALL
SELECT N'B', N'2023-11-02'
UNION ALL
SELECT N'C', N'2023-11-03'
UNION ALL
SELECT N'D', N'2023-11-04'
UNION ALL
SELECT N'E', N'2023-11-05'

DECLARE @BizTime DATETIME = N'2023-11-03'

-- Error occurs
; WITH C AS (
    SELECT *
    FROM #TempTb
    WHERE BizTime > @BizTime )
SELECT * FROM C


-- Works fine
-- SELECT * FROM #TempTb WHERE BizTime > @BizTime


``` 

### DBeaver Version

Community Edition Version 23.2.5.202311191730

### Operating System

Windows 11 

### Database and driver

Database : SQL Server 2016
Driver : Microsoft JDBC Driver for SQL Server 9.2.0

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
I returned back the previous behavior.
In the past sprint I tried to implement the behavior, when we select several queries in single-tab mode and press Run to get multiple results one behind another. However if we start parsing selected region, we broke a workaround for executing queries that we can't parse correctly or that are intentionally should be executed in one batch.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
@@ -2504,9 +2504,10 @@ public class SQLEditor extends SQLEditorBase implements
 
         List<SQLScriptElement> elements;
         ITextSelection selection = (ITextSelection) getSelectionProvider().getSelection();
-        if (script ||
-            selection.getLength() > 1 // if we selected several queries - they should not be inside one SQLQuery instance
-        ) {
+        // if we selected several queries and press Run, they're intentionally goes into one SQLQuery
+        // it's a workaround for cases where we can't correctly parse whole query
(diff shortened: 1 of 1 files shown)


uslss:
verified

PR #21922, opened 2023-11-21
dbeaver/pro#2137 auto relayout

serjiokov:
Added relayout diagram support for cases:
1. Modify visible attributes - view styles and show attributes
2. Switch between notaions and routing


serjiokov:
image

plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/editor/ERDEditorPart.java
```diff
@@ -1215,6 +1216,7 @@ public void run()
                     }
                 }
             }
+            refreshDiagram(true, true);
```
E1izabeth: As far as I remember, we intentionally don't refresh full diagram here as if the diagram is large it take noticeable amount of time. Isn't there another way of rebuilding relations without it?

serge-rider:
Diagram refresh is too expensive. Use rearrangeDiagram instead.

serjiokov:
Unfortunatly method use "getDiagramPart().rearrangeDiagram();" doesn't help
image

E1izabeth:
Can you don't reload metadata than at least?

E1izabeth (approved):

Matvey16:
Verified

PR #21927, opened 2023-11-21
dbeaver/pro#2690 Distributed file configuration UI fix

serge-rider:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/fs/DBNPath.java
```diff
@@ -98,6 +100,12 @@ public boolean allowsChildren() {
         return isDirectory;
     }
 
+    @Override
+    public DBNNode refreshNode(DBRProgressMonitor monitor, Object source) throws DBException {
+
+        return super.refreshNode(monitor, source);
+    }
```
E1izabeth, line 107: What's the point of this method if it just call super?

PR #21943, opened 2023-11-22
dbeaver/pro#2146 use dbname for restore operations

LonwoLonwo:
Please check the following:

- [x] restore process started from the navigator database object
- [x] restore process started from the navigator schema object
- [x] restore process started for the default database
- [x] restore process started for a non-default database
- [x] restore for a database with quoted name


E1izabeth (approved):

MashaKorax:
verified

PR #21949, opened 2023-11-22
#21902 show output panel for statistics tab

### LINKED ISSUE #21902: Don't display server ouput msg after execute plpgsql blocks
### Description

Example:
image

image

### DBeaver Version

 23.2.5.202311191730

### Operating System

Windows 7

### Database and driver

Postgres Pro 15.4

### Steps to reproduce

_No response_

### Additional context

_No response_

LonwoLonwo:
2023-11-22 14_51_39-Window

Important:
SQL Output button must be in the toolbar

Our case is - we have only one result panel, and it is statistic. DBeaver should read and show output in this case as well as for result + output, like for this procedure call:

```sql
CREATE OR REPLACE PROCEDURE pr_test(
    INOUT p_rows INTEGER DEFAULT NULL 
)
AS
$body$
BEGIN
    RAISE NOTICE 'Inside SP, p_rows before modify = %', p_rows;
    p_rows := 100;
    RAISE NOTICE 'Inside SP, p_rows after modify = %', p_rows;
    RETURN;
END
$body$
LANGUAGE 'plpgsql'
;
```

result for: `CALL pr_test();`


E1izabeth (approved):

ShadelessFox (approved):

uslss:
verified

Issue #21615, opened 2023-10-20
Prevent Dbeaver from closing output tabs automatically after CSV export?


kevinfusmithrx:
### Description

I have 3 queries I need to run.

I would like to run all of them and export all of them to their own CSV.

DBeaver populates the results of the 3 queries in each of their own tab (great), however, after I export 1 of them, a tab automatically closes. How do I prevent this from happening?

### DBeaver Version

Dbeaver Community 23.1.4

E1izabeth:
Thank you for bug report. I tried to reproduce your issue, however tab is not closing for me. I see that tab selection is switched to the first tab from the third one, but only when I do export for the first time, subsequent exports doesn't change the tab selection. Could you record a short video of how is it happening to you?

kevinfusmithrx:
Hi Elizabeth,

I'm sorry, I don't think so since I am currently doing this for work and it contains sensitive customer data. 
[Someone on stack overflow](https://stackoverflow.com/questions/77333127/prevent-dbeaver-from-closing-output-tabs-automatically-after-csv-export) was able to reproduce the issue though, its really easy to reproduce.

1. Just write 3 different queries on the same .sql file, ending in ;, and then "run all".
2. The results should come out in 3 different tabs on DBeaver.
3. Then export one of them to excel/csv. 
4. After it finishes exporting, there should be 3 tabs of the executed queries on DBeaver. However, for me, one of them automatically closed after the export and there are only 2 remaining.

Please let me know if this is helpful and if you can reproduce this same issue! Thank you!

Best,
Kevin

Matvey16:
@kevinfusmithrx Thank you for the report. I have reproduced this issue

Matvey16:
https://github.com/dbeaver/dbeaver/assets/82543000/a87e6abd-2f09-4d5c-b450-6e1b764dff4a

kevinfusmithrx:
Yes exactly thank you! This is a bug correct?

kevinfusmithrx:
By the way, the stackoverflow answer is incorrect. Even after pinning all tabs, one of the tabs still disappears.

E1izabeth:
Yes, it's a bug.

E1izabeth:
@kevinfusmithrx 
Thank you for reporting it.

kevinfusmithrx:
@E1izabeth In the meantime, is there a way to export all the result tabs at once?

E1izabeth:
[...]

PR #21953, opened 2023-11-22
#21615 Don't show execution results when extracting data

ShadelessFox:
It only affects data extraction from the editor, meaning loading data in the editor should work as before.


E1izabeth (approved):

MashaKorax:
verified

PR #21966, opened 2023-11-23
dbeaver/pro#2032 use unique indexes as well as PK to update imported data

LonwoLonwo:
with this setting:

2023-11-23 14_59_08-Window

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreInsertReplaceMethod.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreInsertReplaceMethod.java
@@ -67,7 +67,7 @@ public class PostgreInsertReplaceMethod implements DBDInsertReplaceMethod {
                 }
             }
             if (constraintAttrNames.isEmpty()) {
-                // Let's search for an unique index
+                // Let's search for a unique index
                 Collection<? extends DBSTableIndex> indexes = table.getIndexes(monitor);
                 if (!CommonUtils.isEmpty(indexes)) {
                     Optional<? extends DBSTableIndex> optional = indexes.stream().filter(DBSTableIndex::isUnique).findFirst();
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreInsertReplaceMethod.java
```diff
@@ -40,48 +42,82 @@ public class PostgreInsertReplaceMethod implements DBDInsertReplaceMethod {
...
     }
 
-    private void addAttributesNamesList(DBSTable table, DBSAttributeBase[] attributes, boolean isExcluded, StringBuilder updateExpression) {
+    private void getConstraintAttributesNames(
+        @NotNull StringBuilder constraintAttrNames,
+        @NotNull List<? extends DBSEntityAttributeRef> attributeReferences
+    ) {
+        boolean hasKey = false;
+        for (DBSEntityAttributeRef column : attributeReferences) {
+            if (hasKey) constraintAttrNames.append(",");
```
ShadelessFox: This should be placed after line 109; otherwise, if `attribute` is `null` for some reason, you can get multiple consecutive commas.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreInsertReplaceMethod.java
```diff
@@ -40,48 +42,82 @@ public class PostgreInsertReplaceMethod implements DBDInsertReplaceMethod {
...
-                            DBSEntityAttribute attribute = column.getAttribute();
-                            if (attribute == null) continue;
-                            pkNames.append(DBUtils.getQuotedIdentifier(attribute));
-                            hasKey = true;
+                        getConstraintAttributesNames(constraintAttrNames, attributeReferences);
+                    }
+                }
+            }
+            if (constraintAttrNames.isEmpty()) {
+                // Let's search for an unique index
```
ShadelessFox:
```suggestion
                // Let's search for a unique index
```

ShadelessFox (approved):

E1izabeth (approved):

uslss:
verified

PR #21972, opened 2023-11-23
dbeaver/pro#2134 Rendering improvements

ShadelessFox:


E1izabeth (approved):

serge-rider:
Not compiling

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/e4/DBeaverStackRenderer.java
```diff
@@ -49,21 +54,55 @@
...
 
     private static final Log log = Log.getLog(DBeaverStackRenderer.class);
 
+    @Inject
+    @Optional
+    void subscribePerspectiveSwitched(@NotNull IEventBroker broker) {
+        final EventHandler handler = new EventHandler() {
+            @Override
+            public void handleEvent(Event event) {
+                final Object element = event.getProperty(UIEvents.EventTags.ELEMENT);
```
serge-rider: Let's protect it with try/catch throwable.
I am afraid that old Eclipse version or some other thing may cause an unexpected error and may break entire app.

serge-rider (changes requested):

uslss:
verified

serge-rider (approved):

pzygielo:
- Seems that this _improvement_ caused #22275 which was observed almost immediately, but this feature for kids is still there, enabled by default.

PR #21978, opened 2023-11-23
dbeaver/pro#2045 Improve gcp fs folders support

E1izabeth:
Please check creation and deletion for empty and non-empty folders.


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/nio/EFSNIOFolder.java
```diff
@@ -49,7 +55,8 @@ public void create(boolean force, boolean local, IProgressMonitor monitor) throw
 
     public void create(int updateFlags, boolean local, IProgressMonitor monitor) throws CoreException {
         try {
-            Files.createDirectory(getNioPath());
+            Path result = Files.createFile(getNioPath().resolve(EMPTY_FOLDER_FAKE_FILE_NAME));
```
serge-rider: You cannot do it for all file systems. We need it only for GCP

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/nio/EFSNIOFolder.java
E1izabeth: Then I need this class to not be final

serge-rider: I think it is better to use some flag in FS provider (DBFVirtualFileSystem? or maybe additional interface)

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/DBFVirtualFileSystem.java
```diff
@@ -62,6 +62,10 @@ public interface DBFVirtualFileSystem extends Closeable {
 
     default void refreshRoots(DBRProgressMonitor monitor) throws DBException {}
 
+    default boolean isFoldersExists() {
```
serge-rider: Let's rename it to supportsEmptyFolders

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/nio/EFSNIOFolder.java
```diff
@@ -66,7 +79,14 @@ public void createLink(URI location, int updateFlags, IProgressMonitor monitor)
 
     public void delete(boolean force, boolean keepHistory, IProgressMonitor monitor) throws CoreException {
         try {
-            Files.delete(getNioPath());
+            if (getRoot().getRoot().getFileSystem().supportsEmptyFolders()) {
+                Files.delete(getNioPath());
+            } else {
+                // We can't delete pseudo-folder, so this is a workaround
```
serge-rider, line 85: Looks dangerous and unsafe.
Can we delete only our FOLDER_EMULATION_FILE_NAME?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/nio/EFSNIOFolder.java
E1izabeth: Then folder will not be deleted on delete action

MashaKorax:
verified

serge-rider (approved):

PR #21980, opened 2023-11-24
dbeaver/pro#2095 [Cloud Explorer] Fixed message in case of exception

serjiokov:
[Cloud Explorer] Fixed message in case of exception


serjiokov:
image

PR #21981, opened 2023-11-24
#15092 change "native" to "local" for clients; add more links about configuration

### LINKED ISSUE #15092: Change UI for local clients settings
We need to change the UI that is used when a user seys buck up or restore tasks. Currently the UI doesn't provide enough information for user to understend what is needed to start a task, this is especially confusing on Linux and Mac where MySQL and PostreSQL binaries are not downloaded automatically and nothing in the UI points to the fact that user needs to install them manually. The main confusing points of the UI are:
1. User doesn't understand what is a native/local client
2. If no client is installed on a machine, user doesn't get any tips that he needs to do that manually outside of DBeaver
3. While adding a directory with binaries in DBeaver, there is no tips to what directory user needs to poin to

LonwoLonwo:
2023-11-24 13_22_55-DBeaver Enterprise 23 2 5 - _postgres_ Script-32


E1izabeth (approved):

serge-rider:
LEt's not hardcode external links in localization files.
We can pass this as parameter + use HelpUtils.getHelpExternalReference

uslss:
verified

PR #21986, opened 2023-11-24
dbeaver/dbeaver#21202 Filter by value - search fix

### LINKED ISSUE #21202: Filter -> Filter by value - UUID FK value is not in the list since version 23.1.5
### Description

When I want to filter by value FK column with UUID type and input the correct value, it's not shown in the list.

The screen is from version 23.1.4 where it works.
It does not work with 23.1.5 and 23.2.0

image

### DBeaver Version

23.2.0

### Operating System

Windows 10

### Database and driver

PostgreSQL 14.8

### Steps to reproduce

_No response_

### Additional context

_No response_

serge-rider:


uslss:
verifed

PR #21987, opened 2023-11-24
#21650 Identifier escape fix

### LINKED ISSUE #21650: Autocomplete does not handle column names with international characters correctly
### Description

We have tables in a databricks database that have swedish column names. One example is the swedish word Resultatställe. Swedish words often contains the characters åäö ÅÄÖ.

DBeaver lists these columns correctly in the autocomplete dropdown:

image

But when selecting a column the generated code causes an error in Databricks:

image

To make the code work we need to manually add backquoutes around the name:

image

It would be much better if DBeaver automatically added these backticks if a column name contains any non-ascii characters.

The problem also affects the * expansion. The generated list of columns should have backticks around each column name with non-ascii characters.

### DBeaver Version

Community edition Version 23.2.3.202310221533

### Operating System

Windows 11

### Database and driver

Azure Databricks
Downloads\SimbaSparkJDBC42-2.6.22.1040\SparkJDBC42.jar

### Steps to reproduce

See description

### Additional context

_No response_

ShadelessFox:
Sample query for testing:
```sql
CREATE TABLE test (
  normalcolumn STRING,
  `Ställe` STRING,
  `åäöÅÄÖ` STRING,
  `Resultatställe` STRING
) USING delta;
```


Matvey16:
Verified

PR #21996, opened 2023-11-26
dbeaver/dbeaver#21995 New foreign key name generator fix

### LINKED ISSUE #21995: New foreign key constraint has wrong name
It ignores referenced table name.
New FK names look like `{owner_table}_FK`.
It has to look like `{owner_table}_{ref_table}_FK`.

serge-rider:


uslss:
verified

Issue #21617, opened 2023-10-20
Reserved word current_date is not being quoted in generated SQL for MariaDB


stringfold:
### Description

As the title says, if you have a database table column with the name "current_date", you cannot perform any operations in DBeaver that includes that column in the generated SQL.

It looks like other column names that use reserve words, like "length" are quoted correctly in the SQL, but "current_date" is not quoted, which makes it impossible to do things like modify the column definition or export data from the table containing a column of that name directly.

I checked to see if anyone else has flagged this particular error, but didn't see an issue. Other reserved word issues were resolved several years ago.

### DBeaver Version

Community Edition 23.2.2.202310081633

### Operating System

Windows 11

### Database and driver

MariaDB version 10.3.38

### Steps to reproduce

1. Using the UI, create a new database table and add a column named current_date.
2. When you attempt to persist the change, the operation fails with the following error:

> You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'current_date varchar(100) NULL) 
> ENGINE=InnoDB
> DEFAULT CHARSET=utf8mb4
> COLLA...' at line 2

Exporting a table with a column of that name also fails.

E1izabeth:
Thank you for bug report.

LonwoLonwo:
CURRENT_TIMESTAMP and current_time are also reserved. Probably, as other functions.

LonwoLonwo:
I can't reproduce it against DBeaver 23.3.5

![2024-02-26 11_02_10-Window](https://github.com/dbeaver/dbeaver/assets/45152336/e809c537-79e1-4ae8-bd63-f300253d8a82)

Driver update helped, probably.
I'm closing this issue.

PR #22000, opened 2023-11-27
#21617 update MySQLDialect for keywords setting

PzaThief:
This PR has whole update of MySQL dialect especially on keywords.
I listed up all keywords of MySQL, So, When initialization of dialect it process clear all default keywords and then register correctly.
In result, there is no more dirty correction like `removeSQLKeyword("SOURCE");`.

But, the problem remains about MySQL version and MySQL-like DB compatibility. (So, I think this patch is just temporary to make user feel at ease.)
Some reserved keywords can be removed always like `NONBLOCKING` ([MySQL Reserved Words](https://dev.mysql.com/doc/mysqld-version-reference/en/keywords.html)) , and MariaDB has another keywords like `OFFSET` ([MariaDB Reserved Words](https://mariadb.com/kb/en/reserved-words/)).

So, I hope to make a disscusion about keeping this design.
I just naively think to make an option(as check-box) to quote all column names whether it is reserved or not.
This approach is used in hibernate (option names hibernate.globally_quoted_identifiers) and I think it not will be hard works.
Could I get your opinion on this?

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
@@ -33,7 +33,6 @@ import org.jkiss.dbeaver.model.struct.rdb.DBSProcedureType;
 import org.jkiss.utils.ArrayUtils;
 
 import java.util.Arrays;
-import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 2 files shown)


E1izabeth:
Thank you for contribution.

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
```diff
@@ -43,157 +42,183 @@
...
-        "REPAIR",
-        "ENGINE"
-    };
-
-    public static final String[][] MYSQL_QUOTE_STRINGS = {
-            {"`", "`"},
-            {"\"", "\""},
-    };
-
+            BasicSQLDialect.NON_TRANSACTIONAL_KEYWORDS, new String[] { "USE", "SHOW", "CREATE", "ALTER", "DROP" });
```
E1izabeth, line 45: Why did you delete DESCRIBE and DESC keywords?

PzaThief: because BasicSQLDialect.NON_TRANSACTIONAL_KEYWORDS already has DESCRIBE, DESC and EXPLAIN I removed it.

E1izabeth: Ok, thanks

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
```diff
@@ -338,7 +363,8 @@ public boolean validIdentifierStart(char c) {
 
     @NotNull
     @Override
-    public String getTypeCastClause(@NotNull DBSTypedObject attribute, @NotNull String expression, boolean isInCondition) {
+    public String getTypeCastClause(@NotNull DBSTypedObject attribute, @NotNull String expression,
+            boolean isInCondition) {
```
E1izabeth, line 367: Could you revert this unintended change or reformat it?
If the line doesn't fit in 140 characters we format it like this
```suggestion
    public String getTypeCastClause(
        @NotNull DBSTypedObject attribute,
        @NotNull String expression,
        boolean isInCondition
    ) {
```

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
```diff
@@ -43,157 +42,183 @@
...
-
-    public static final String[][] MYSQL_QUOTE_STRINGS = {
-            {"`", "`"},
-            {"\"", "\""},
-    };
-
+            BasicSQLDialect.NON_TRANSACTIONAL_KEYWORDS, new String[] { "USE", "SHOW", "CREATE", "ALTER", "DROP" });
+
+    private static final String[] ADVANCED_KEYWORDS = { "AUTO_INCREMENT", "DATABASES", "COLUMNS", "ALGORITHM", "REPAIR",
+            "ENGINE" };
```
E1izabeth, line 48: Please don't change existing formatting like this where it's not needed.

E1izabeth:
Also we usually ask to create an issue on our issue tracker before making a pull request. And you can leave a comment saying that you are going to implement this feature (or fix a bug) by yourself.
https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code

E1izabeth:
@PzaThief could you fix code style and merge conflict?

LonwoLonwo:
Hello.
I checked. MariaDB and MySQL both return the correct colors for the words from the ticket. And for the whole of this list: https://github.com/mysql/mysql-server/blob/mysql-8.2.0/plugin/x/src/mysql_function_names.cc

2024-02-26 11_02_33-Window

So, thanks for the PR, but I do not see any reason to implement it.

PR #22012, opened 2023-11-28
#21991 show default procedure parameters

### LINKED ISSUE #21991: The default values of a procedure parameters do not appear.
### Description

As I explained in the "Steps to reproduce", the Procedure->Properties->Source form does not show the default values of the procedure.

### DBeaver Version

Community Edition 23.2.0

### Operating System

Windows Server 2019

### Database and driver

1. DB : Firebird 4.0
2. Driver : Firebird Jaybird driver org.firebirdsql.jdbc.FBDriver 5.0.2

### Steps to reproduce

1. Create a procedure with default values in its parameters.
`SET TERM # ;
CREATE OR ALTER PROCEDURE SP_PARAMS_WITH_DEFAULT_VALUES(P0 integer not null, P1 integer=10, P2 integer default 20, P3 VARCHAR(5) DEFAULT 'TEST')
RETURNS(R0 integer, R1 integer, R2 integer, R3 VARCHAR(5))
as
begin
  R0 = :P0;
  R1 = :P1;
  R2 = :P2;
  R3 = :P3;
  SUSPEND;
end#`
2. Execute this line (with only P0 parameter) in the Sql Editor `SELECT * FROM SP_PARAMS_WITH_DEFAULT_VALUES(12)`
3. Refresh the "Procedures" node in Database Navigator.
4. Double click on the procedure. On the right side, in the "Source" section, the default values of the parameters do not appear.
5. When I make a change in the procedure and press Save and then Persist in the dialog screen, the procedure does not work because the default values of the procedure are lost.

### Additional context

_No response_

LonwoLonwo:
2023-11-28 12_36_11-DBeaver Ultimate 23 3 0 - SP_PARAMS_WITH_DEFAULT_VALUES

Impact only on the **procedure** parameters reading. You can check it in different Firebird versions.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.firebird/src/org/jkiss/dbeaver/ext/firebird/model/FireBirdProcedure.java
+++ b/plugins/org.jkiss.dbeaver.ext.firebird/src/org/jkiss/dbeaver/ext/firebird/model/FireBirdProcedure.java
@@ -111,7 +111,7 @@ public class FireBirdProcedure extends GenericProcedure implements DBSObjectWith
                         "F.RDB$CHARACTER_LENGTH AS CHAR_LEN,\n" +
                         "PP.RDB$PARAMETER_NUMBER + 1 AS ORDINAL_POSITION,\n" +
                         "F.RDB$CHARACTER_SET_ID,\n" +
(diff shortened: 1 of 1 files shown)


uslss:
verified

PR #22021, opened 2023-11-28
#21615 Fix statistics tab appearing prematurely

### LINKED ISSUE #21615: Prevent Dbeaver from closing output tabs automatically after CSV export?

ShadelessFox:
The statistics tab was created prematurely because a tab is required when filling in script variables (to disable the counter while the modal window blocks the editor) even if variables might not be present (which is always true in the case of the statistics tab). So I have rearranged the logic a bit, so a tab is only acquired when it's actually needed.


MashaKorax:
verified

PR #22029, opened 2023-11-29
#22027 Fix special symbol quoting in HTML SQL converter

### LINKED ISSUE #22027: Correct HTML format on "Copy SQL as source code". Less and greater than symbols.
**Problem**
Some times we need to copy the source code of a query in a ticket platform (OsTicket) that only support HTML, so we use the HTML format of "Copy SQL as source code" often. The problem is that the < and > symbols aren't formatted correctly.

For example, if we have the SQL sentence:

    select * from table where id <> 10;

Is copied as:

    <pre><span style='color:#739eca;font-weight:bold;'>select</span> * <span style='color:#739eca;font-weight:bold;'>from</span><span style='color:#739eca;font-weight:bold;'>table</span> <span style='color:#739eca;font-weight:bold;'>where</span> <span style='color:#9e9e9e;'>id</span> <> <span style='color:#c0c0c0;'>10</span><span style='color:#eecc64;'>;</span></pre>

And that is showed as:
image

Where the `<>` symbols are lost.

**Solution**
To replace the symbols `<>` with `&lt; &gt;` HTML character entities in the HTML format on "Copy SQL as source code".

ShadelessFox:


LonwoLonwo (approved):

Destrolaric (approved):

Matvey16:
Verified

PR #22037, opened 2023-11-29
#21907 text editor limit improvements

### LINKED ISSUE #21907: SQLite connection-specific `Maximum text editor content length` setting does not apply
### Description

When I configure `Maximum text editor content length` on a specific SQLite connection, the setting does not apply. After reconnecting/restarting there is still a `Warning: Content representation size is limited to: 100 KB`. If I change it in the global settings, the configuration does apply and the warning disappears.

### DBeaver Version

23.2.5

### Operating System

Void Linux

### Database and driver

1. SQLite 3.44.0
2. SQLite JDBC driver 3.44.0.0

### Steps to reproduce

1. `Edit connection` on a SQLite database.
2. Configure `Maximum text editor content length` and set it to e.g. `1000`
3. Reconnect and open a large cell in the results panel; still see `Warning: Content representation size is limited to: 100 KB` (and only part of the content is loaded).
4. If I change the same setting in the global settings, the configuration does apply and the warning disappears as expected.

### Additional context

_No response_

ShadelessFox:
- Now the preference is read from datasource-specific preferences
- Some visual overhaul of the message bar

image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
@@ -342,6 +342,8 @@ public abstract class AbstractTextPanelEditor<EDITOR extends BaseTextEditor>
                 return;
             }
             resetEditorInput();
+            int maxContentLength = valueController.getExecutionContext().getDataSource().getContainer()
+                .getPreferenceStore().getInt(ResultSetPreferences.RS_EDIT_MAX_TEXT_SIZE) * 1000;
             if (value.getContentLength() > maxContentLength) {
                 showLimitedContent(value, maxContentLength);
             } else {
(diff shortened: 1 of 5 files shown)


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -336,12 +305,13 @@ public void primeEditorValue(
             return;
         }
         try {
-            int maxContentLength = DBWorkbench.getPlatform().getPreferenceStore().getInt(ResultSetPreferences.RS_EDIT_MAX_TEXT_SIZE) * 1000;
             if (editor == null) {
                 log.error("Editor is null or undefined");
                 return;
             }
             resetEditorInput();
+            int maxContentLength = valueController.getExecutionContext().getDataSource().getContainer()
```
LonwoLonwo: getExecutionContext() has @Nullable annotation. How about to check it?

ShadelessFox: We do the same basically in every other editor, so it was kind of a no-brainer to me to use the same construction

LonwoLonwo (approved):

Matvey16:
Verified

PR #22085, opened 2023-12-04
dbeaver/pro#1771 add gpt4

Destrolaric:
Closes dbeaver/pro#1771 
Adds gpt4 to selection.
It will not be a default model due to limited OpenAI rates for requests to that model and possible limitations for users with trial keys. gtp4-32k will not be added due to it still being invite-only.


MashaKorax:
verified

PR #22088, opened 2023-12-04
dbeaver/pro#2204 code cleanup

LonwoLonwo:
Please check the event triggers showing in the navigator tree. And working with them


uslss:
verified

Issue #21574, opened 2023-10-18
DBeaver creates a ".eclipse/configuration" directory in your profile for each version


qris:
### Description

DBeaver creates a .eclipse/<version>/configuration directory in your home directory (e.g. `c:\users\chris`). It contains quite a lot of files, a new one is created each time you upgrade DBeaver, and they are never deleted.

![image](https://github.com/dbeaver/dbeaver/assets/754175/91eee796-09de-46b8-8fb3-269b5e9f379e)

Some of them seem pretty egregious, for example there is a PNG splash screen. Why is that part of my configuration anyway?

![image](https://github.com/dbeaver/dbeaver/assets/754175/937ee874-2711-4b27-89ae-b95630922076)

In any case, this causes significant profile bloat over time. shouldn't DBeaver either use the same directory (maybe backing up the old one after an upgrade, keeping only one such backup) or delete old version-stamped directories after a while?

I had 90 MB in old configuration directories (about 15 of them) after using DBeaver for a few years.

### DBeaver Version

Community Version 23.1.4.202308061552

### Operating System

Windows 10

### Database and driver

n/a

### Steps to reproduce

1. Install and run DBeaver
2. Upgrade and run new version
3. Upgrade again and run new version
4. Look in %USERPROFILE%\.eclipse

serge-rider:
@qris thank you for the report.
It is a known issue, this happens when you install DBeaver in a write-protected folder (e.g. Program Files). DBeaver needs to extract several artifacts (including native binaries) native binaries during first start. When it is installed in a local user folder then all artifacts are extracted in the same folder where DBeaver was installed.
Otherwise they are extracted in user home `.eclipse` folder (because it is a part of Eclipse platform behavior).

Generally it is better to use different temp files location (e.g. `%APPDATA%\DBeaverData`). But we will need to patch core Eclipse libraries for that.

Workaround: install DBeaver for current user only (the default option).

qris:
Thanks, could the installer not extract these into a read-only system-wide directory, shared between all users?

E1izabeth:
Looks like a duplicate of #2118

PR #22089, opened 2023-12-04
dbeaver/dbeaver#21574 eclipse configuration directory in your profile

serjiokov:
The configuration options specified in order to store data in OS specific folders.


product/community/DBeaver.product
```diff
@@ -94,6 +94,9 @@
         <plugin id="org.eclipse.equinox.event" autoStart="true" startLevel="2"/>
         <!-- This is important because simpleconfigurator conflicts with update.configurator -->
         <plugin id="org.eclipse.update.configurator" autoStart="false" startLevel="10"/>
```
serge-rider: conf -> "install-data"

serge-rider (changes requested):

product/community/DBeaver.product
```diff
@@ -94,6 +94,9 @@
...
+        <property name="osgi.configuration.area" value="@user.home/AppData/Roaming/DBeaverData/conf" os="win32" />
+        <property name="osgi.configuration.area" value="@user.home/.config/DBeaverData/conf" os="linux" />
```
serge-rider: Not sure about Linux path. It has to be in $XDG_DATA_HOME

serge-rider (changes requested):

product/community/DBeaver.product
serjiokov: on OS the $XDX_DATA_HOME have to be resolved into 

image

product/community/DBeaver.product
serjiokov: done

serjiokov:
For now reached next points:
1. Organized own dbeaver launcher with implemetation of store relative configurtaion (../DBeaverData/install-data) in OS specific location:
    a. WIN - APPDATA
    b. UNIX - XDG_DATA_HOME 
In order to achive it next properties in product require:
        <property name="osgi.configuration.area" value="**@data.home**/DBeaverData/install-data" os="win32" />
        <property name="osgi.configuration.area" value="**@data.home**/DBeaverData/install-data" os="linux" />
        <property name="osgi.configuration.area" value="**@data.home**/DBeaverData/install-data" os="macosx" />

Not achived:
Generation .ini file option (or phaze of product materialization) still paste 
"-startup 
  plugins/org.eclipse.equinox.launcher_1.6.500.v20230717-2134.jar" that does not fit requiremnts.
This line have to be changed to 
" -startup
plugins/org.jkiss.dbeaver.launcher_[version]jar" 

Result:
Product configuration data will be stored at `"~/AppData/DBeaverData/install-data"`
The "~/.eclipse" not used any more.

plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Constants.java
```diff
@@ -0,0 +1,149 @@
+/*
```
serge-rider: Can we just import all this stuff from original launcher and not copy-paste it?
We have it in plugin list anyway.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Constants.java
serjiokov: I don't think so we targeted to reimplement a private method 
` private URL buildLocation(String property, URL defaultLocation, String userDefaultAppendage) `

serge-rider: Yes. But why can't we just import constants, JNI bridge and other stuff which is simply copy-pasted?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Constants.java
serjiokov: Done, but one thing the equinox classes is in internal package.

Matvey16:
Verified

Matvey16:
Verified

PR #22092, opened 2023-12-04
dbeaver/pro#2152 advanced paste improvements

ShadelessFox:
Test cases:
- [x] Advanced paste without selection
- [x] Advanced paste with single range selection
- [x] Advanced paste with multiple selections
- [x] Copy row from below/above
- [x] Copy multiple rows within a single range from above/below
- [x] Copy multiple rows within multiple selections from above/below


uslss:
verified

PR #22112, opened 2023-12-06
#21495 set query limits flag for MySQL/MariaDB

### LINKED ISSUE #21495: MariaDB JDBC Drivers do not allow LOAD DATA LOCAL INFILE if ResultSet fetch size is set
> Hi, I experienced exactly the same error on DBeaver 23.1.1 on macOS Big Sur & Windows, MariaDB version "10.5.20-MariaDB-log" on FreeBSD 13, JDBC driver MariaDB Connector/J 3.1.4. The fix mentioned here did not work. Rolling back to MariaDB Connector/J 3.0.4 fixed this issue. The same error occurred for each of the JDBC versions between 3.0.5 and 3.1.4.
> 
> Using the latest mySQL JDBC connector for MariaDB works as expected, i.e. it does not show the error described here.

Thanks for the info and testing, hkirschk!

Version 3.2.0 of the MariaDB Connector/J also does not work but rolling back to 3.0.4 does the trick

It should also be noted that the following allows the 3.2.0 driver to work:

Set the ResultSet fetch size parameter to 0 

image

This, I believe, is a bug in DBeaver. DBeaver _should_ detect a LOAD DATA statement and turn off the fetch size for the transaction.

Setting this value can be accomplished through several clicks, which is an annoyance, or running a simple select (such as SELECT 1) and then changing the value in the row limit box and THEN running the LOAD DATA command. 

In both instances, it is necessary to remember to put the limit back after the load is done.

_Originally posted by @JTMosaic in https://github.com/dbeaver/dbeaver/issues/19115#issuecomment-1755506069_

LonwoLonwo:
Please check the case from the ticket and some different SELECT queries with a result set fetch size and without it (set to 0). 
(Expected result: `LIMIT 0, 200` will be added to your query. Depends on your settings in Preferences -> Editors -> Data Editor)

Also, you can ask me about how work with the original query from the ticket

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDataSource.java
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDataSource.java
@@ -110,6 +110,8 @@ public class MySQLDataSource extends JDBCDataSource implements DBPObjectStatisti
                 } else {
                     return 255;
                 }
+            case DBPDataSource.FEATURE_LIMIT_AFFECTS_DML:
+                return true;
         }
         return super.getDataSourceFeature(featureId);
     }


ShadelessFox (approved):

uslss:
verified

PR #22122, opened 2023-12-07
CB-4216. Implemented new identity provider - JWT Token.

DenisSinelnikov:


ShadelessFox (commented):
What's up with icons? They look stretched and compressed like they are from 00s.

ShadelessFox (approved):

alexander-skoblikov (changes requested):
these icons should not be in this repository

DenisSinelnikov:
> these icons should not be in this repository

moved

PR #22128, opened 2023-12-07
dbeaver/pro#2195 Reuse auth credentials for multiple DS connection + …

serge-rider:
…annotations


Matvey16:
Still opens a separate browser window for every connection for Azure connections with MFA auth
image

PR #22133, opened 2023-12-08
dbeaver/pro#2221 Remote file selector fix (support root folders)

serge-rider:


MashaKorax:
AWS - ok
GCP - ok
Azure- 

1. Create task with output in root bucket, save
2. Run task - see success in task run panel
3. Go to  root bucket - **no exported file there**
error log is empty

PR #22143, opened 2023-12-11
dbeaver/pro#2135 Optimize rules scanner

E1izabeth:
Fixed an issue with lagged highlighting on editing (was not stable reproducible).
Bug illustration (no highlighting on the first line) :
image

Optimized rules scanner to remove redundant context requests, so highlighting should work a bit faster now.

### What should be tested:
At first, turn these settings on image
Then:
- check mentioned bug existence 
- tables should be colored with one color, columns - with another one. Also if you wrote select with invalid column (which doesn't exist in the database) it should be highlighted with red. 
image
- try editing the query and check that highlighting updates


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/SQLBackgroundParsingJob.java
```diff
@@ -356,6 +357,9 @@ public Interval runWithResult() {
             SQLParserContext parserContext = new SQLParserContext(this.editor.getDataSource(), this.editor.getSyntaxManager(), this.editor.getRuleManager(), this.document);
             List<SQLScriptElement> elements = SQLScriptParser.extractScriptQueries(parserContext, workOffset, workLength, false, false, false);
             if (elements.isEmpty()) {
+                if (DEBUG) {
+                    System.out.println("No script elements to parse in range " + workOffset + "+" + workLength);
```
ShadelessFox, line 361: But why? We have loggers, use them instead

E1izabeth: I don't want to log it at all. I need it to debug if something goes wrong. When this feature become stable, it will be deleted.

ShadelessFox (approved):

uslss:
verified

PR #22144, opened 2023-12-11
CB-4336 update derby driver

yagudin10:


LonwoLonwo:
Why not the https://mvnrepository.com/artifact/org.apache.derby/derby/10.17.1.0?
Derby 10.16 has a direct vulnerability.

yagudin10:
> Why not the https://mvnrepository.com/artifact/org.apache.derby/derby/10.17.1.0? Derby 10.16 has a direct vulnerability.

the latest driver requires Java 21 and higher

LonwoLonwo (approved):

PR #22163, opened 2023-12-13
dbeaver/pro#1974 update azure UI & detect azure model

Destrolaric:
Closes dbeaver/pro#1974


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/openai/OpenAICompletionEngine.java
```diff
@@ -158,6 +158,7 @@ protected String requestCompletion(
         DBSObjectContainer mainObject = getScopeObject(context, executionContext);
 
         final GPTModel model = getModel();
+        GPTCompletionAdapter service = getServiceInstance(executionContext);
```
ShadelessFox, line 161: Why this change?

Destrolaric: Because we requested getMaxTokenSize() before getting info about max tokens from the instance

LonwoLonwo (approved):

ShadelessFox (approved):

PR #22200, opened 2023-12-18
dbeaver/pro#2082 context variables

serge-rider:


DariaMarkaryan:
Verified

E1izabeth:
This case doesn't work #18998

E1izabeth:
Case sensitivity is also broken. Here I expect to get `select * from foo` query.
image

DariaMarkaryan:
> This case doesn't work #18998

checked this case one more time - it works for me .-. checked in both CE and EE

E1izabeth:
@DariaMarkaryan yes, you're right. I rechecked and it works for me too. Maybe I did something wrong last time.

PR #22201, opened 2023-12-18
dbeaver/pro#2136 Quoted identifiers highlighting

E1izabeth:
- Quoted identifiers are recognized according to dialect parameters now!
- MS SQL and SQLite -specific identifiers quotation parametrized with plugin.xml, that should have been like that by design.
- Now dialect-specific keywords are highlighted as keywords, where they were treated as identifiers before.

1. When validation by metadata is turned off:

image

then we see this (note that these databases has different quotes set for identifiers quotation): 

image

image

image

2. When validation by metadata is turned on:

image

image

image

Note that if quotes are wrong for the identifier for some dialect - the thing in some quotes cannot be a quoted identifier, we fallback to our regular highlighting with blue and green for the light theme.

To be tested:
- Quoted columns should be highlighted as a regular columns.
- Quoted tables should be highlighted as a regular tables.
- Test for [ ], " ". ` ` and ' ' quotes for SQLite (all of them can quote identifier), and MS SQL (some of them can quote identifier).


plugins/org.jkiss.dbeaver.model.lsm/src/org/jkiss/dbeaver/model/lsm/sql/dialect/LSMDialectRegistry.java
```diff
@@ -61,27 +62,34 @@ private void loadExtensions(IExtensionRegistry registry) {
...
-            analyzer = knownLsmAnalyzerByDialects.get(dialectClass);
+            analyzerFabric = knownLsmAnalyzerByDialects.get(dialectClass);
             dialectClass = dialectClass.getSuperclass();
-        } while (analyzer == null && dialectClass != null);
-        return analyzer;
+        } while (analyzerFabric == null && dialectClass != null);
+        
+        if (analyzerFabric == null) {
+            throw new IllegalStateException(
+                "Failed to resolve LSMAnalyzer for " + dialect.getDialectName() + " dialect. Illegal database driver configuation."
```
LonwoLonwo:
```suggestion
                "Failed to resolve LSMAnalyzer for " + dialect.getDialectName() + " dialect. Illegal database driver configuration."
```

E1izabeth: Thanks!

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -552,6 +552,7 @@
                     <parameter name="supports-indexes" value="true"/>
                     <parameter name="supports-stored-code" value="false"/>
                     <parameter name="supports-truncate" value="false"/>
```
serge-rider: Doesn't look like a good idea. We have many SQL Server drivers, we don't want't to copy-paste it everywhere.
Let's use Java level function for that (we already have one in SQLDialect)

plugins/org.jkiss.dbeaver.ext.postgresql/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -1,18 +1,13 @@
 eclipse.preferences.version=1
-org.eclipse.jdt.core.compiler.codegen.inlineJsrBytecode=enabled
```
serge-rider, line 2: Let's delete this file from Git

serge-rider (dismissed):

Matvey16:
Incorrect highlighting for fully qualified name in postgres

dbeaver_Tdq9bmncqH

Matvey16:
https://github.com/dbeaver/dbeaver/assets/82543000/b7226f38-b93e-4f49-b07e-32af0a0599ae

LonwoLonwo (approved):

Matvey16:
Verified

PR #22214, opened 2023-12-20
dbeaver/pro#2205 Turn on experimental highlighting features by default

E1izabeth:
Highlighting with validation by metadata should be turned on by default.


LonwoLonwo (approved):

Matvey16:
Verified

PR #22224, opened 2023-12-20
dbeaver/pro#2292 Connection collapse fix

serge-rider:


uslss:
some connections to disabled database produces endless error message loop 

https://github.com/dbeaver/dbeaver/assets/31996417/5aa70b99-e6e5-43b6-8006-08681e1ca9b3

uslss:
verified

PR #22226, opened 2023-12-21
dbeaver/pro#2293 Fix small bugs

E1izabeth:
- Fix NPE - you should not catch it while writing some query and jumping from aliases to tables with Ctrl+Click
- Color descriptions are added. You should find descriptions for all SQL Editor colors on a human natural language
image
- If you change SQL table alias foreground and SQL derived column foreground, new color should be applied.


Matvey16:
Verified

PR #22237, opened 2023-12-21
dbeaver/pro#2197 Handle encrypted project opening failure

ShadelessFox:
Test cases:
- [x] Open an encrypted project with configuration encryption, will be notified that the project can't be opened
- [x] Open an encrypted project without configuration encryption, will be opted to either load and lose all credentials or abort:
- [x] If proceeded, then a file called `credentials-config.json.bak` should be created inside the project directory in `.metadata`.

Screenshots:
- java_TG6aJbYhF5
- java_1dvMyzakGu


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceSerializerModern.java
```diff
@@ -796,6 +775,89 @@ public boolean parseDataSources(
...
+        return null;
+    }
+
+    @Nullable
+    private Map<String, Object> readConfiguration(
+        @NotNull DBPDataSourceConfigurationStorage configurationStorage,
+        @NotNull DataSourceConfigurationManager configurationManager,
+        @Nullable Collection<String> dataSourceIds
+    ) throws DBException, IOException {
+        final InputStream is;
```
serge-rider, line 834: Input stream is never closed

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceSerializerModern.java
```diff
@@ -796,6 +775,89 @@ public boolean parseDataSources(
...
+    @Nullable
+    private Map<String, Map<String, Map<String, String>>> readSecureCredentials(
+        @NotNull DBPDataSourceConfigurationStorage configurationStorage,
+        @NotNull DataSourceConfigurationManager configurationManager,
+        @Nullable Collection<String> dataSourceIds
+    ) throws DBException {
+        if (configurationManager.isSecure()) {
+            return null;
+        }
+        final String name = "%s%s%s".formatted(
```
serge-rider: Why use formatted instead of string concatenation?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceSerializerModern.java
ShadelessFox: It's closed inside the `IOUtils#copyStream` method below.

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceSerializerModern.java
ShadelessFox: I though that the concatenation looked ugly, so I replaced it with formatting that's at least more readable IMO.

LonwoLonwo (approved):

E1izabeth (approved):

MashaKorax:
Can't leave this dialogue on Mac

https://github.com/dbeaver/dbeaver/assets/84867187/0d8c4bea-118c-4fe2-b433-704229a2ade4

E1izabeth (approved):

LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceSerializerModern.java
```diff
@@ -814,7 +811,7 @@ private Map<String, Map<String, Map<String, String>>> readSecureCredentials(
                         log.error("Error backing up secure credentials", e1);
                     }
                 } else {
-                    throw new DBException("Project opening canceled by user");
+                    log.error("Project opening canceled by user");
```
ShadelessFox: This **won't** prevent the project from opening. The main idea was to throw an exception so the node wouldn't be fully loaded. See `DBNProject#readChildNodes`.

serjiokov: @ShadelessFox  this case specifically designed when user hit a cancel button in dialog, it is prevent opening project. Throwing exception by the cancellation really not needed.

serjiokov: In result, user able to operate with unsecured projects as before, the secured assets present in tree but do not available for use.

ShadelessFox: This is current behavior (before this ticket). The idea was to completely block the user from opening the project because as soon they make changes to it (change connection configuration, etc), it will be persisted and all credentials will be lost.

My approach was completely valid, the only issue that stopped my PR from being merged is that it behaves weirdly on macOS. This is what needs to be fixed.

serjiokov: No objection of your approach, it's valid, the main issue on both OS's (Win Mac) is hitting "Cancel" button in the dialog.
Pls check the snapshort on win os

Uploading Recording 2024-01-17 122418.mp4…

ShadelessFox: No recording is attached.

serjiokov: It happens due to

serjiokov: Uploading Recording 2024-01-17 122418.mp4…

serjiokov: image

ShadelessFox: Still no recording attached.

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceSerializerModern.java
```diff
@@ -373,8 +373,14 @@ public boolean parseDataSources(
         var connectionConfigurationChanged = false;
 
         // Read in this particular order to handle configuration reading errors first, but process in reverse order later
+        Map<String, Map<String, Map<String, String>>>  secureCredentialsMap = null ;
         final var configurationMap = readConfiguration(configurationStorage, configurationManager, dataSourceIds);
-        final var secureCredentialsMap = readSecureCredentials(configurationStorage, configurationManager, dataSourceIds);
+        try {
+            secureCredentialsMap = readSecureCredentials(configurationStorage, configurationManager, dataSourceIds);
```
ShadelessFox: This doesn't cover the case where the `readConfiguration` might fail as well.

serjiokov: Generally, yes better to cover both functions

serge-rider (approved):

DariaMarkaryan:
Verified

DariaMarkaryan:
Verified

PR #22254, opened 2023-12-26
CB-4424 Restrict access to system variables in DPI processes for Clou…

alexander-skoblikov:
…dBeaver


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/connection/DataSourceVariableResolver.java
```diff
@@ -47,6 +49,10 @@ protected DBPConnectionConfiguration getConfiguration() {
 
     @Override
     protected boolean isResolveSystemVariables() {
+        DBPApplication application = DBWorkbench.getPlatform().getApplication();
```
serge-rider: I think we add function isEnvironmentVariablesAccessible into DBPApplication to make it more general. 
Usage of `instanceof DBPApplicationDesktop` is a silly hack.

serge-rider (changes requested):

serge-rider (approved):

PR #22256, opened 2023-12-26
#3087 Format execution time

### LINKED ISSUE #3087: Feature Request: format elapsed time for running queries
It would be nice if the elapsed time displayed for running queries was formatted as HH:MM:SS instead of just displaying seconds:

dbeaver

I use dbeaver for a lot of long-running analytics queries and this would be useful for monitoring run times.

ShadelessFox:
java_wMStkAYJWQ

The following query can be used for testing (against PostgreSQL):
```sql
select pg_sleep(<seconds>);
```


E1izabeth (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #22271, opened 2023-12-27
dbeaver/pro#1463 read notifications jaws

serjiokov:
Provided accessibility for read notifications by JAWs


LonwoLonwo:
So, will the focus now be on the notification after operation by default?

E1izabeth (approved):
If it's not possible without setting a focus, than ok. But I hope it's possible to stop notification reading and return focus back with Esc

serjiokov:
The main idea is in by default the focus will not switch to notification, it is only possible when screen reader support enabled.

ShadelessFox (approved):

MashaKorax:
verified

PR #22281, opened 2023-12-28
#22169 Add multiline comment rule to lexer

### LINKED ISSUE #22169: Autocomplete column names disabled when multi-line comment syntax is used
### Description

When using the multi-line comment syntax, this seems to disable autocomplete of column names.
Comments like this will disable the autocomplete.
/*---
test comment
---*/
 
Autocomplete not working example (multi-line comments below the select statement also don't work):
image

Working with inline comment:
image

### DBeaver Version

Community Version 23.3.0

### Operating System

Windows 11 Enterprise

### Database and driver

SQL Server 2017

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
I believe we don't want to handle nested comments for simplicity. It makes comments parsing not easy and I'm not sure we really need it. Maybe let's wait for requests?
https://groups.google.com/g/antlr-discussion/c/OEk5ug3GNNM


ShadelessFox (approved):
Some databases support nested multiline comments, but I don't remember which. It's okay not to support them for now.

LonwoLonwo (approved):

Matvey16:
Verified

PR #22283, opened 2023-12-28
dbeaver/pro#2308 Use AbstractJob instead of TimeTask in SQLBackgroundParsingJob

E1izabeth:
Please test on editing of some huge query like query from this issue https://github.com/dbeaver/pro/issues/1907
There should be no DBeaver freezes, because the parsing job should run in background and re-run on editing.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/SQLBackgroundParsingJob.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/SQLBackgroundParsingJob.java
@@ -18,12 +18,16 @@ package org.jkiss.dbeaver.ui.editors.sql.semantics;
 
 import org.antlr.v4.runtime.misc.Interval;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/SQLBackgroundParsingJob.java
```diff
@@ -214,17 +215,19 @@ private void schedule(@Nullable DocumentEvent event) {
             }
 
             // TODO should we really schedule a new task each time this method called? or maybe at least cancel it at first
-            this.task = new TimerTask() {
+            this.task = new AbstractJob("Background parsing job") {
```
ShadelessFox, line 218: Have you considered marking it as a system job so it won't be shown in the background tasks?

E1izabeth: Actually no. I want it to be shown as a separate background task to have ability to cancel it. I even added subtasks with progress. So, it user open a really big query or connection is slow, it's shown in the Background tasks - you see that something is happening, you see some progress.

ShadelessFox (approved):

LonwoLonwo (approved):

serjiokov (approved):

Matvey16:
Verified

PR #22285, opened 2023-12-28
dbeaver/pro#2199 Handle responses with code blocks

ShadelessFox:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/metadata/MetadataProcessor.java
```diff
@@ -116,10 +116,13 @@ public DAICompletionMessage createMetadataMessage(
         final DBCExecutionContext executionContext = context.getExecutionContext();
 
         final StringBuilder sb = new StringBuilder();
-        sb.append("Perform SQL completion");
 
         if (model.isChatAPI()) {
-            sb.append(". Start response with SELECT keyword");
+            sb.append("You must perform SQL completion against a SQLite database. " +
```
E1izabeth: Why SQLite?

ShadelessFox: Oops, nice catch.

E1izabeth (approved):

LonwoLonwo (approved):

Issue #13529, opened 2021-08-13
SQL Server: Source tab is empty in the index properties


kseniiaguzeeva:
#### System information: 
- Operating system (distribution) and version Win10
- DBeaver version 21.1.5 EA

#### Connection specification:
- Database name and version: SQL Server Azure
#### Steps to reproduce, if exist:
Open a table clustered index in Metadata Editor

**Expected**: Source tab has Index DDL
**Actual result**: the Source tab is empty.
![dbeaver_wuaK5Era1S](https://user-images.githubusercontent.com/51405061/129363781-762710d8-d78d-4606-9fb8-e0825b4c9592.png)

willturcios1:
@LonwoLonwo thanks for updating the one I opened, I looked around and did not find this one, but yes this is the same behavior.

willturcios1:
Also want to point out it does work for non-clustered indexes so perhaps is just a filter issue when extracting.

LonwoLonwo:
> Also want to point out it does work for non-clustered indexes so perhaps is just a filter issue when extracting.

Thanks, I fixed the name of this issue.

Nrj27:
is this issue still open?

LonwoLonwo:
Hello @Nrj27 

Yes, it is.

The problem here is in the SQLServerTableIndex#readIndexDefinition
The query there is not for unique/primary indexes.

Nrj27:
@LonwoLonwo can you please share complete problem query along with complete screenshot.
please share the system requirements.

LonwoLonwo:
You can find the query in SQLServerTableIndex#readIndexDefinition. No special system requirements

SaumilP:
Is this issue still available for fix?

HansenChristoffer:
I added PR with fix. Would need couple of eyes on it.

PR #22294, opened 2023-12-30
#13529 Add is_primary_key constraint check for both boolean equivalent values

SaumilP:
By default if created table has PK part of clustered index, create index structure sql statement excludes it. SQL statement changed to include `sys.indexes#is_primary_key`.


E1izabeth:
Thank you for contribution

Issue #21565, opened 2023-10-18
Generate DDL missing square brackets in MSSQL Clustered Index definition


robert-sjoblom:
### Description

When generating DDL for a table, the index definitions are missing square brackets around column names containing spaces or keywords.

### DBeaver Version

DBeaver CE Version 23.2.2.202310081633

### Operating System

Ubuntu 22.04

### Database and driver

Database name and version: Microsoft SQL Server 2019
Driver name and version: MS SQL Server / SQL Server Microsoft; JDBC Driver for SQL Server

### Steps to reproduce

Given the following table:
```
CREATE TABLE db.dbo.DetailedCust_Ledg_Entry (
    [timestamp] timestamp IDENTITY(0,1) NOT NULL,
    [Entry No_] int NOT NULL,
    [Cust_ Ledger Entry No_] int NOT NULL,
    [Amount (LCY)] decimal(38,20) NOT NULL,
    [Client No_] varchar(20) COLLATE Latin1_General_100_CI_AS NOT NULL,
    ... -- more columns here
);
```
And an existing clustered index on `Cust_ Ledger Entry No_, Entry Type`, the following DDL for the clustered index is generated:

```
CREATE NONCLUSTERED INDEX ix_name ON dbo.DetailedCust_Ledg_Entry (  Cust_ Ledger Entry No_ ASC  , Entry Type ASC  )  
     INCLUDE ( Amount , Amount (LCY) , Client No_ ) 
     WITH (  PAD_INDEX = OFF ,FILLFACTOR = 100  ,SORT_IN_TEMPDB = OFF , IGNORE_DUP_KEY = OFF , STATISTICS_NORECOMPUTE = OFF , ONLINE = OFF , ALLOW_ROW_LOCKS = ON , ALLOW_PAGE_LOCKS = ON  )
     ON [PRIMARY ] ;
```
which obviously won't work.

I would expect the following ddl:
```
 CREATE NONCLUSTERED INDEX ix_DetailedCustLedgEntry_CustLedgerEntryNo ON dbo.DetailedCust_Ledg_Entry (  [Cust_ Ledger Entry No_] ASC  , [Entry Type] ASC  )  
     INCLUDE ( Amount , [Amount (LCY)] , [Client No_] ) 
[...]

PR #22296, opened 2023-12-30
#21565 Add missing square brackets around selected columns

SaumilP:
Index generation SQL query change to introduce index header.


LonwoLonwo:
Hello @SaumilP 

Thanks for your contribution.

Please provide more info about your PR. 
What issue does it help to resolve? 
And more examples.

SaumilP:
Hi @LonwoLonwo,

This change is more a cosmetic change attended to help visually differentiate between reserved keywords from other escaped words/non-reserved words.

As you can see from this snapshot, generated index source represents reserved word specific color coding due to no use of square brackets.

__Incorrect color-coding for reserved keyword__ :
image

Square brackets are optional and are recommended to be used to enclose the column/table names - if it contains any special characters (like punctuation marks), any spaces or contains reserved words. It also assist with better code reading. After the code change, index source looks like this:

image

PR #22357, opened 2024-01-05
dbeaver/dbeaver#22355 look for clients in /usr/lib on Linux & display real paths

### LINKED ISSUE #22355: Finding of local clients on Linux & macOS improvements
### Description

We can add two small improvements to https://github.com/dbeaver/dbeaver/issues/14620.

### DBeaver Version

CE 23.3.2 devel

### Operating System

macOS & Linux

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

1. When we find a symlink to a local client, we can display the original path to the file in the `Local Client` combo, not the symlink path.
2. We can expand the search area on Linux by adding `/usr/lib`. This will allow us to find all clients installed using a package manager, not just the one symlinked to `/usr/bin` (this is the way in work on openSUSE, at least).

flambchap:
image


ShadelessFox (approved):

E1izabeth (approved):

Matvey16:
Verified

PR #22375, opened 2024-01-08
#22369 Change confirmation message for clean option of restore wizard for PostgreSQL

### LINKED ISSUE #22369: Misleading info on Clean option for PostgreSQL restore
### Description

When you restore a backup on an existing database there is the "Clean" option.  
image

The text says that the current database will be delete and recreated. But it seems that it will only drop the tables that are contained in the restored backup. So tables existing in the database but not in the backup will remain and with this there were further issues with primary and foreign keys after the restore.

Either the database is really recreated if this option is checked or the description of this option shall be adjusted.

### DBeaver Version

Community Edition 23.3.2.202401072024

### Operating System

Windows 10

### Database and driver

1. Database version PostgreSQL 15.0
2. Driver PostgreSQL JDBC Driver

### Steps to reproduce

Restore a backup on an existing database

### Additional context

_No response_

E1izabeth:
image


LonwoLonwo (approved):

MashaKorax:
verified

PR #22388, opened 2024-01-09
CB-4403. Implement logic for mapping AD roles if they are specified, …

DenisSinelnikov:
…when import users


plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/user/SMUserProvisioning.java
```diff
@@ -42,4 +47,8 @@ public Map<String, String> getMetaParameters() {
         return metaParameters;
     }
 
+    @Nullable
+    public String getAuthRole() {
```
alexander-skoblikov, line 51: unused

alexander-skoblikov (changes requested):
redundant changes

alexander-skoblikov (approved):

PR #22389, opened 2024-01-09
#21789 Do not rely on whether the truncate feature is supported or not

### LINKED ISSUE #21789: Import data into SQlite: truncate target tables before load does not work
### Description

SQLite does not have a TRUNCATE command it used DELETE FROM instead.  Maybe this is why this option in data import fails.  Even if it is selected, old data is not deleted from the target table before new data is imported.

### DBeaver Version

All (including 23.2.4 CE)

### Operating System

Windows 10 professional

### Database and driver

SQLite

### Steps to reproduce

Make a CSV file to import into a SQLite table
use the import data command and select 'truncate target tables before load ' option
Import the data, the see that the old data is still in the SQLite table

### Additional context

_No response_

ShadelessFox:


LonwoLonwo (approved):

E1izabeth (approved):

MashaKorax:
verified

PR #22392, opened 2024-01-09
dbeaver/pro#2304 remove check constraints from the edit page; return …

LonwoLonwo:
1. Check constraints removed from the unique keys page creation for SQL Server.
You can create a new check constraint via a special folder.

2. Ability to choose the check constraints type returned for Vertica, DB2i, H2, and Informix (no need to check them all, just a few) (please do not create check definition like 1 = 1 on Vertica - we broke our partner's database last time with this syntax)

2024-01-09 17_46_46-DBeaver Ultimate 23 3 2 - Employees


E1izabeth (approved):
Hope we won't lost something useful

ShadelessFox (approved):

Matvey16:
Check constraint option disappeared from it's own folder too

https://github.com/dbeaver/dbeaver/assets/82543000/ae53e0e8-3d28-49d8-877b-33e1cd2df4f8

Matvey16:
Verified

PR #22393, opened 2024-01-09
dbeaver/pro#2331 add ability to create unique keys only for databases…

LonwoLonwo:
add the ability to create unique keys only for databases that support this (and remove for others).

We do not read unique keys for all generic databases, only for a few:

- Vertica
- H2
- Hana
- FireBird
- DB2i
- Derby
- Netezza (PRO)
- Informix (PRO)
- Snowflake (PRO)
- Teradata (added) (PRO)

So, there is no reason to add an ability to create a unique index if we do not see it.


E1izabeth (approved):

ShadelessFox (approved):

JoseLourenc0 (commented):
:fire: :fire: :fire:

MashaKorax:
verified

PR #22401, opened 2024-01-10
CB-4301 qm autoreconnect

alexander-skoblikov:


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/storage/InternalDatabaseConfig.java
```diff
@@ -35,4 +35,43 @@ public interface InternalDatabaseConfig {
 
     String getSchema();
 
+    Pool getPool();
+
+    class Pool {
+        private int minIdleConnections = 2;
+        private int maxIdleConnections = 10;
+        private int maxConnections = 1000;
+        private String validationQuery = "SELECT 1";
```
ShadelessFox, line 44: Are they expected to be mutable? Maybe this is a good candidate for turning it in a `record`?

PR #22403, opened 2024-01-10
dbeaver/pro#2309 Column alias visibility

### LINKED ISSUE #22374: aliases' re-use auto-coloration are inconsistant
### Description

Hello, small example : in Snowflake, one can reuse in WHERE clause (for instance) aliases declared in SELECT section. But the syntaxic color feat loses its mind and does not manage to remain consistant throughout the SQL sentence.... (3 ways to test it : )

image
image
image

### DBeaver Version

CE Version 23.3.2.202401072024

### Operating System

Win11 23H2 (Build 22631.2715)

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
### Solves both https://github.com/dbeaver/dbeaver/issues/22374 and https://github.com/dbeaver/pro/issues/2309

Column aliases has restricted visibility:
https://stackoverflow.com/questions/13031013/how-do-i-use-alias-in-where-clause

Currently we made them visible by default for all databases everywhere.
And specified some concrete visibility scopes for several databases (see below). Please test specified databases. If column alias is not visible in some scope, it should be highlighted with red. If it should be visible in some scope, no red highlighting should appear. Note: it should be tested with active connection and `Read database tables/columns for semantic analysis` on.

Column alias is visible in:
- [x]  **SQLite**: everywhere
- [x] **PostgreSQL**: GROUP BY, ORDER BY https://dba.stackexchange.com/a/225883
- [x] **DB2**:  ORDER BY https://www.ibm.com/docs/en/ias?topic=nps-using-column-aliases-in-where-clause
- [x] **MySQL**: GROUP BY, HAVING, ORDER BY https://dev.mysql.com/doc/refman/8.0/en/problems-with-alias.html
- [x] **SQL Server**: ORDER BY https://learn.microsoft.com/en-us/previous-versions/sql/compact/sql-server-compact-4.0/ms173451(v=sql.110)?redirectedfrom=MSDN
- [x] **Snowflake**: everywhere
- [x] **Oracle**: ORDER BY
- [x] **Clickhouse**: everywhere
- [x] **Exasol**: ORDER BY

You can test it on the following query:
```
SELECT
1 AS TEST_ALIAS
FROM
TPCH_SF1.CUSTOMER  a
WHERE TEST_ALIAS IS not NULL AND 1 > TEST_ALIAS
GROUP BY TEST_ALIAS
HAVING TEST_ALIAS > 0 AND 8 > TEST_ALIAS
ORDER BY TEST_ALIAS;
```
Main idea
If **nothing** is **highlighted** with red, then query should be **executed** **without errors**. If **something** is **highlighted** with red, then query should **NOT** be **executed** **without errors**.
Also you can investigate visibility scopes by trying to execute such query and reading database response.


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialect.java
```diff
@@ -503,4 +512,13 @@ void generateStoredProcedureCall(
      * @return a set of SQLBlockCompletions with information about blocks for autoedit
      */
     SQLBlockCompletions getBlockCompletions();
+
+    default EnumSet<ProjectionAliasVisibilityScope> getProjectionAliasVisibilityScope() {
```
serge-rider, line 516: Why EnumSet? Can use good old java array instead?

serge-rider (changes requested):

uslss:
verified

serge-rider (approved):

PR #22408, opened 2024-01-10
dbeaver/dbeaver#16544 BigQuery structs and arrays support

### LINKED ISSUE #16544: Poor Result Formatting of BigQuery Nested Objects
#### System information: 
- Operating system (distribution) and version: macOS Monterey - 12.3.1 (21E258) 
- DBeaver version: 22.0.4.202205011839
- Additional extensions: 

- Database name and version: Google Big Query
- Driver name: SimbaJDBCDriverforGoogleBigQuery42_1.2.22.1026

BigQuery has a datatype called STRUCT, which is essentially a key-value type. DBeaver does not display the key names, instead of using `..."f":..` as a stand-in for _all keys_. And similarly uses `..."v":...` to indicate values before displaying the value, which seems unnecessary and convoluted. It also might be reading those objects as strings (i.e., an "ABC" field).

Screen Shot 2022-05-19 at 3 46 42 PM

Anyway, all this makes using the data more difficult. Not being able to see the keys means I have to use the web app to look up what keys are available. 

Screen Shot 2022-05-19 at 3 54 58 PM

```sql
select 
	attributes
	, attributes.id
	, attributes.name
from bigquery-public-data.genomics_cannabis.cs10_gff
limit 100;
```

serge-rider:


uslss:
Structs are incorrectly shown if "Show complex column" option is disabled
image

serge-rider:
Fixed (if complex transformer is disabled then value will rendered as before in plain text mode)

uslss:
still not fixed
image

serge-rider:
One more fix

uslss:
verified

PR #22421, opened 2024-01-11
dbeaver/dbeaver#19821 exclude all suggestions with dots

### LINKED ISSUE #19821: Autocomplete duplicates table aliases
### Description

The user said that DBeaver autocomplete adds an additional alias:

In the following sample query, When we start typing "a." DBeaver generates suggestions of columns and when we pick a column by pressing enter on the required column DBeaver ends up typing not just the column name but a.column_name resulting in text a.a.column_name. Then we have to go back and delete the second alias. It is frustrating. Is this something that can be fixed? Attached is the screenshot of the DBever verison we have also recording of the issue with test datasets in snowflake.

```
select 
a.*
from table_a a inner join table_b b on a.id=b.id
```

https://user-images.githubusercontent.com/12581569/234844531-ceee5aa0-7144-4db5-8b0f-aab4de498bf1.mov

image

### DBeaver Version

СЕ 23.0.2

### Operating System

macOS

### Database and driver

Snowflake

### Steps to reproduce

_No response_

### Additional context

_No response_

Destrolaric:
Closes dbeaver/dbeaver#19821 
Exclude all suggestions with dots from Hippie due to them being error prone.


E1izabeth (approved):
I'm ok with it, only because Hippie is not our main autocompletion engine.

ShadelessFox (commented):
Looks hacky. Can't it be better to detect the existing part of the proposal in the script and replace it?

Matvey16:
Verified

Issue #2247, opened 2017-10-02
Java 9 + DBeaver 4.2.2 fails at start with NoClassDefFoundError / IllegalStateException


mmoehring:
Hello,
I downloaded the new 4.2.2. to test it with Java9, deleted all settings, started it and got an error preventing the startup. See attached log files:

[dbeaver-debug.log](https://github.com/serge-rider/dbeaver/files/1348730/dbeaver-debug.log)
[dot.log](https://github.com/serge-rider/dbeaver/files/1348731/dot.log)

mmoehring:
Thanks!

syhily:
Yeah, add `--add-modules=ALL-SYSTEM` after `-vmargs` fixed this issue.

JonnyTech:
Worked for me too - this is my file now looks like:

```
$ cat /usr/share/dbeaver/dbeaver.ini
-startup
plugins/org.eclipse.equinox.launcher_1.4.0.v20161219-1356.jar
--launcher.library
plugins/org.eclipse.equinox.launcher.gtk.linux.x86_64_1.1.500.v20170531-1133
-showsplash
--launcher.GTK_version
2
-vmargs
--add-modules=ALL-SYSTEM
-Duser.language=en_US
-Xms64m
-Xmx1024m
```

PR #22431, opened 2024-01-11
dbeaver/pro#2247 Close idle connections #2247

serjiokov:
Added reference dependency close connection IDLE time per connection types


serjiokov:
Here is a short movie in order to describe behaviour of wizards for value initialization and modification:

https://github.com/dbeaver/dbeaver/assets/32031248/7f327f2a-d3fa-490b-8d75-77cd16847ff8

Matvey16:
Verified

PR #22469, opened 2024-01-16
dbeaver/dbeaver#22467 Use form editor in object edit page + null valu…

### LINKED ISSUE #22467: New object dialog: use forms instead of property editor
Forms are much better from UIX perspective

serge-rider:
…es for combos


uslss:
verified for column, index, constraints creation

Issue #2256, opened 2017-10-03
Toggle Line Comment improper


OSmeilus:
I just upgraded to Windows 10 and now running DBeaver with Version 4.2.2.
Now I've noticed that formatting with "Toggle Line Comment" is now longer working properly. There is no toggle but a unique change.

Initial Situation:
SELECT *
  FROM tab_a
-- WHERE field_1 = 4711
 WHERE field_2 = 'abc'    

Formatting before updates: (by selecting last 2 lines and pressing Ctrl-Shift-7)
SELECT *
  FROM tab_a
 WHERE field_1 = 4711
 -- WHERE field_2 = 'abc'    

Formatting after updates:
either:
SELECT *
  FROM tab_a
-- WHERE field_1 = 4711
---- WHERE field_2 = 'abc'    

or:
SELECT *
  FROM tab_a
 WHERE field_1 = 4711
 WHERE field_2 = 'abc'

serge-rider:
This behavior was changed on purpose - see #1850.
I think this is less confusing if you have a big query with many commented lines.

OSmeilus:
Thank you for your response.

Well, this might fit the requirement when you mark commented lines and executable lines.
But this don't work, when you just want to **toggle** between commented and executable lines.

prog-rajkamal:
Adding my issue to this one as they are related
------------------------

toggle line comment fails when there is a space at beginning of line.

eg: 

When toggling comment

     -- select * from emp; (there is a space at beginning of line)
 
results in

    -- -- select * from emp;

@serge-rider let me know if I should open a new issue for this

serge-rider:
@prog-rajkamal It is a bug in comments processing. Thanks for detailed report.

PR #22472, opened 2024-01-16
dbeaver/dbeaver#2256 splash screen fixed for mac sonoma

serjiokov:
Splash screen fixed for mac sonoma

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverSplashHandler.java
+++ b/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverSplashHandler.java
@@ -21,6 +21,7 @@ import org.eclipse.core.runtime.IProduct;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.jface.resource.StringConverter;
+import org.eclipse.jface.util.Util;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.graphics.*;
 import org.eclipse.swt.widgets.Shell;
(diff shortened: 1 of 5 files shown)


plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverSplashHandler.java
```diff
@@ -85,7 +85,14 @@ private void initVisualization() {
             versionInfoSizeString = product.getProperty("versionInfoSize");
             versionInfoColorString = product.getProperty("versionInfoColor");
         }
-
+        /*
+         * Related to a bug in MacOS Sonoma
+         * (https://github.com/eclipse-platform/eclipse.platform.swt/issues/772)
+         */
```
serge-rider, line 92: 1. Check that it is MacOS.
2. Return splash screen in MacOS (see previous PR)

serge-rider: And why use absolute coordinates? What are these numbers and why they are hardcoded?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverSplashHandler.java
serjiokov: 1. done
2. done
3. This is a property value for product in general all specified in abs coords
```
      <property name="startupMessageRect" value="323,270,242,25"/>
      <property name="startupProgressRect" value="323,255,242,10"/>
```

serge-rider: These coordinates may change at any moment.
You can create new properties like startupMessageRectMacOS

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverSplashHandler.java
serjiokov: Not sure if I get it right, these approach of "man on the middle", properties are modified specifically for Mac OS after general reading procedure line:81, initially properties specified in plugin.xml.

serjiokov:
@serge-rider  one more thing, the splash image in bmp format looks little bit accurate, would we replace the current (.png) ?

PR #22492, opened 2024-01-17
#22490 Add double-dash single-line comments for Snowflake dialect

### LINKED ISSUE #22490: Manage syntaxic coloration for comments starting with double-slashes
In snowflake, in-line comment in your SQL code is usually starting with double dash ( -- ), but it can also support in-line comment starting with double-slash ( // ) 
Example : 

image

In DBeaver, only double dash is taken into consideration, and not double-slash...
image

Is there a way to activate the syntaxic coloration in the latter case ? If not, please take it as a feature request
Thanks

E1izabeth:
Double-dash comments should be colored with grey for Snowflake


LonwoLonwo (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #22493, opened 2024-01-17
dbeaver/pro#2247 Close idle connections

serjiokov:
Dedicated attributes added for setup auto closable value(period) for connection configuration based on templates of connection types


serjiokov:
image

serjiokov:
image

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/CoreMessages.java
```diff
@@ -53,6 +53,9 @@ public class CoreMessages extends NLS {
     public static String action_menu_transaction_auto_close_enabled;
     public static String action_menu_transaction_auto_close_enabled_tip;
     public static String action_menu_transaction_auto_close_ttl_tip;
```
serge-rider: We already have this localization (dialog_connection_wizard_final_label_close_idle_connections)

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/CoreMessages.java
serjiokov: These dedicated prefereces page localization
image

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/connection/DBPConnectionType.java
```diff
@@ -274,6 +292,22 @@ public void setCloseIdleTransactionPeriod(long closeIdleTransactionPeriod) {
         this.closeIdleTransactionPeriod = closeIdleTransactionPeriod;
     }
 
+    public boolean isAutoCloseConnections() {
```
serge-rider, line 295: This option is never used anywhere

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageInitialization.java
```diff
@@ -349,7 +350,11 @@ public void saveSettings(DBPDataSourceContainer dataSource) {
         bootstrap.setInitQueries(bootstrapQueries);
 
         confConfig.setKeepAliveInterval(keepAliveInterval.getSelection());
```
serge-rider, line 360: You mustn't set connection settings from connection type.
DBPConnectionType must be used in runtime (as any other DBPConnectionType  properties)

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageInitialization.java
serjiokov: Get it, m I right that require new logic for close idle connections in class DataSourceMonitorJob ?

Matvey16:
Connection type properies are not being saved

https://github.com/dbeaver/dbeaver/assets/82543000/74a2180f-ec83-4032-9d03-baed0260e533

Matvey16:
Verified

PR #22499, opened 2024-01-18
#22351 check the "grant" option only if grant permission exist

### LINKED ISSUE #22351: User properties shows incorrect setting for "Grant option"
### Description

When a user is granted "ALL" access to all schemas in MySQL or MariaDB they are not granted the right to grant access to other users. This is called the "grant option". This appears incorrectly in user properties pages.

### DBeaver Version

23.3.0+

### Operating System

macOS 14.2.1

### Database and driver

MySQL 8
MariaDB 11

### Steps to reproduce

Granting a user permissions like so:

```
grant all on *.* to user;
```

Will create a grant entry like:

```show grants for user```
image

Granting a user permissions like so:

```
grant all on *.* to user with grant option;
```

Will create a grant entry like:
```show grants for user```
image

In both cases, the "Grant option" box is checked on the user properties page

image

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by LonwoLonwo**
Check the "grant" option only if grant permission exists

2024-01-18 13_14_29-DBeaver Ultimate 23 3 3 - testuser1@%

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/controls/PrivilegeTableControl.java
+++ b/plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/controls/PrivilegeTableControl.java
@@ -236,10 +236,12 @@ public class PrivilegeTableControl extends Composite {
                 if (isStatic && !grant.isStatic()) {
                     continue;
                 }
-                if (privilege.getName().equalsIgnoreCase(MySQLConstants.PRIVILEGE_GRANT_OPTION_NAME) && grant.isGrantOption()) {
-                    // WITH GRANT OPTION is enabled only in this case
-                    privilegeEnabled = true;
-                    break;
+                if (privilege.getName().equalsIgnoreCase(MySQLConstants.PRIVILEGE_GRANT_OPTION_NAME)) {
+                    if (grant.isGrantOption()) {
+                        // WITH GRANT OPTION is enabled only in this case
+                        privilegeEnabled = true;
+                        break;
+                    }
                 } else if (grant.isAllPrivileges() || ArrayUtils.contains(grant.getPrivileges(), privilege)) {
                     privilegeEnabled = true;
                     break;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/controls/PrivilegeTableControl.java [hunk 1]
```diff
@@ -236,10 +236,12 @@ public void fillGrants(List<MySQLGrant> grants) {
...
-                if (privilege.getName().equalsIgnoreCase(MySQLConstants.PRIVILEGE_GRANT_OPTION_NAME) && grant.isGrantOption()) {
-                    // WITH GRANT OPTION is enabled only in this case
-                    privilegeEnabled = true;
-                    break;
+                if (privilege.getName().equalsIgnoreCase(MySQLConstants.PRIVILEGE_GRANT_OPTION_NAME)) {
+                    if (grant.isGrantOption()) {
+                        // WITH GRANT OPTION is enabled only in this case
+                        privilegeEnabled = true;
+                        break;
+                    }
```
JoseLourenc0, line 244: Excuse me, but is this change really needed since the condition is still the same?

LonwoLonwo: Nope, it is not the same.
Now, if we have `privilege.getName().equalsIgnoreCase(MySQLConstants.PRIVILEGE_GRANT_OPTION_NAME)`, we will go inside the block and avoid going to the condition `else if (grant.isAllPrivileges() || ArrayUtils.contains(grant.getPrivileges(), privilege))` for case `!grant.isGrantOption()`

JoseLourenc0: Got it. Thanks

E1izabeth (approved):

uslss:
verified

PR #22505, opened 2024-01-18
dbeaver/dbeaver#22479 Remove JNA deps from MySQL, PostgreSQL an Oracl…

### LINKED ISSUE #22479: JNA loader constraint violation with MariaDB & waffle-jna
### Description

when loading latest mariadb connector version (3.3.2), connection throw an error 
```
Unexpected driver error occurred while connecting to the database
  loader constraint violation: loader org.jkiss.dbeaver.registry.driver.DriverClassLoader @76f31bcb wants to load class com.sun.jna.Pointer. A different class with the same name was previously loaded by org.eclipse.osgi.internal.loader.EquinoxClassLoader @4cf5d999. (com.sun.jna.Pointer is in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @4cf5d999, parent loader 'platform')

```
This is a problem, because since a few version, mariadb 3.3.2 has a waffle-jna dependency. So this always occurs for all connections, even if GSS authentication is not used. 

Waffle-jna dependency and seems incompatible with dbeaver (reading from #8247), i've tryed to checking "ignore transcient dependencies" that permit connection when not using GSS authentication. 
When needing GSS authentication, adding waffle artificts manually without JNA, that results in error" com/sun/jna/ptr/IntByReference", probably because JNA version doesn't match.

### DBeaver Version

23.3.2.202401072024

### Operating System

windows 11

### Database and driver

mariadb database 10.11 and java connector 3.3.2

### Steps to reproduce

_No response_

### Additional context

This is similar to #8247 and #15289

serge-rider:
…e. Win registry read moved to the registry bundle.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql/META-INF/MANIFEST.MF
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/META-INF/MANIFEST.MF
@@ -11,7 +11,6 @@ Require-Bundle: org.eclipse.core.runtime,
  org.jkiss.dbeaver.data.gis,
  org.jkiss.dbeaver.registry,
  org.jkiss.dbeaver.tasks.native,
- com.sun.jna.platform,
  com.google.gson
 Bundle-ActivationPolicy: lazy
 Bundle-RequiredExecutionEnvironment: JavaSE-17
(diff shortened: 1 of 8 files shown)


uslss:
verified

rusher:
correction confirmed. thanks !

PR #22534, opened 2024-01-22
dbeaver/dbeaver#22530 Do not read lazy props for new objects

### LINKED ISSUE #22530: Table/Column editor very slow in Fedora Linux
### Description

I upgraded from 23.0.1 to 23.3.3 today. Now when I try to add column in table editor it shows a popup dialog instead of inline editing. And the popup dialog takes around 5-7 seconds to pop. Now if I have to add 50 columns then its like way too much time. I don't mind popup dialog either but it cannot take 5-7 secs to pop. Also if there is any way I can revert back to inline column editing UI then tell me please. 

### DBeaver Version

Community Edition 23.3.3

### Operating System

Fedora Linux 39

### Database and driver

PostgreSQL 16.1
PostgreSQL JDBC Driver 45.5.2

### Steps to reproduce

1. Open **_Database Navigator_**
2. Select and connect to some database (assuming there is already an existing connection)
3. Expand **_Databases_**
4. Select and expand some particular database 
5. Expand **_Schemas_** under selected database
6. Select and expand some particular schema
7. Right click on **_Tables_** and click on **_Create New Table_**. This will open a table editor
8. Select **_Columns_** tab from left 
9. Right click on white space in **_Columns_** tab and then click on **_Create New Column_** in the popup menu
10. Now the UI will hang for 5-7 secs before showing the popup dialog for creating new column.

### Additional context

_No response_

serge-rider:


serge-rider:
Please test metadata editor and lazy properties loading (like table statistics)

PR #22544, opened 2024-01-22
dbeaver/pro#2317 skip filter errors and not close filter dialog

LonwoLonwo:
Test ticket's case with different options enabled/disabled in the filter dialog.


E1izabeth (approved):

Matvey16:
Verified

PR #22556, opened 2024-01-23
dbeaver/pro#2236 add an ability to send all messages to AI

Destrolaric:
Closes dbeaver/pro#2236


LonwoLonwo (approved):

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/completion/DAICompletionSession.java
```diff
@@ -16,35 +16,52 @@
...
 import java.util.List;
 
+import org.jkiss.code.NotNull;
+
 /**
  * Completion session
  */
 public class DAICompletionSession {
-    private final List<DAICompletionMessage> messages;
+    private final List<DAICompletionMessage> allMessages;
```
ShadelessFox: Let's store all messages in the session. It's the responsibility of the engine to collect and filter messages in which it's interested.

Please revert all changes in this file.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/completion/DAICompletionEngine.java
```diff
@@ -50,13 +50,27 @@ List<DAICompletionResponse> performQueryCompletion(
...
+    /**
+     * Do query completion in a session
+     */
+    @NotNull
+    List<DAICompletionResponse> performQueryCompletion(
+        @NotNull DBRProgressMonitor monitor,
+        @NotNull DAICompletionContext context,
+        @NotNull DAICompletionSession session,
+        @NotNull IAIFormatter formatter,
+        boolean allMessages
```
ShadelessFox, line 70: I'd call it `includeAssistantMessages` or `includeResponseMessages`, but even these are not really descriptive.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/completion/AbstractAICompletionEngine.java
```diff
@@ -48,14 +48,25 @@ public List<DAICompletionResponse> performQueryCompletion(
...
+    @NotNull
+    @Override
+    public List<DAICompletionResponse> performQueryCompletion(
+        @NotNull DBRProgressMonitor monitor,
+        @NotNull DAICompletionContext context,
+        @NotNull DAICompletionSession session,
+        @NotNull IAIFormatter formatter,
+        boolean sendAllMessages
+    ) throws DBException {
+        final String result = requestCompletion(monitor, context, session.getMessages(sendAllMessages), formatter);
```
ShadelessFox: Filter messages here. Don't make the session do all the heavy lifting for you.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/completion/DAICompletionSession.java
Destrolaric: Well, most chat engines will have a similar behavior regarding answers from AI, so this may lead to the code duplication

ShadelessFox: No? All implementations inherit from the `AbstractAICompletionEngine` class, and that's where you want to filter messages. See https://github.com/dbeaver/dbeaver/pull/22556#discussion_r1463271034

ShadelessFox (approved):

PR #22559, opened 2024-01-23
dbeaver/pro#2377 App becomes unresponsive when switching router on a certain schema

serjiokov:
Fixed circling graph, enhanced layout
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/layout/algorithm/direct/OrthoDirectedGraphLayout.java
+++ b/plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/layout/algorithm/direct/OrthoDirectedGraphLayout.java
@@ -33,9 +33,9 @@ public class OrthoDirectedGraphLayout extends DirectedGraphLayout {
 
     private static final int DISTANCE_BTW_LEVELS = 2;
     private AbstractGraphicalEditPart diagram;
-    private TreeMap<String, List<Node>> nodeByLevels;
+    private TreeMap<Integer, List<Node>> nodeByLevels;
     private List<Node> isolatedNodes = new LinkedList<>();
-    private static final int DEFAUL_OFFSET_FROM_TOP_LINE = 40;
+    private static final int DEFAULT_OFFSET_FROM_TOP_LINE = 40;
     private static final int DEFAULT_ISO_OFFSET_HORZ = 140;
     private static final int DEFAULT_OFFSET_BY_X = 250;
     private static final int DEFAULT_OFFSET_BY_Y = 80;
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/layout/algorithm/direct/OrthoDirectedGraphLayout.java
```diff
@@ -225,6 +227,13 @@ private TreeMap<String, List<Node>> recomputeGraph(TreeMap<String, List<Node>> g
         int idx = 0;
         while (idx < graph.keySet().size()) {
             createGraphLayers(graph, idx);
+            // catch empty nodes   
+            List<Node> nextLevelNodes = graph.get(String.valueOf(idx+1));
```
ShadelessFox: Please fix the code style.

Also, could you tell me why you're using integers converted to strings as keys? Is there a particular reason for that? It requires more computations than just boxing that integer.

plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/layout/algorithm/direct/OrthoDirectedGraphLayout.java
```diff
@@ -287,7 +295,26 @@ private void createGraphLayers(TreeMap<String, List<Node>> nodeByEdges, int idx)
...
                 if (trg != null) {
                     String nodeIndex = getNodeIndex(nodeByEdges, trg);
+                    // skip by distance
+                    boolean skip = false;
+                    for (Edge e : trg.incoming) {
+                        Node incomingSourceNode = e.source;
+                        String childIndexStr = getNodeIndex(nodeByEdges, incomingSourceNode);
+                        if (!childIndexStr.isEmpty() && !childIndexStr.equals("0")) {
+                            Integer childIndex = Integer.valueOf(childIndexStr);
+                            if ((idx - childIndex) > DISTANCE_BTW_LEVELS) {
```
ShadelessFox, line 303: I'm not sure how it works, but will it handle when there are two tables that point to each other?

plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/layout/algorithm/direct/OrthoDirectedGraphLayout.java
serjiokov: Generally yes it is not needed, I think all enhancement will be done in another ticket.

plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/layout/algorithm/direct/OrthoDirectedGraphLayout.java
serjiokov: If two or more tables will point to each other tables will be on the same level.  This part just handle distance btw entities.

plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/layout/algorithm/direct/OrthoDirectedGraphLayout.java
serge-rider: Let's fix code style though. 

And generally changing String to Integer is easy, why not do it right here?

serge-rider (changes requested):

ShadelessFox (approved):

uslss:
verified

PR #22561, opened 2024-01-23
dbeaver/pro#2365 read global properties if new connection

Destrolaric:
Fixes the situation when global properties were changed, and no settings page was previously loaded, leading to settings not being applied
Criteria:
* **No changes should apply retroactively to existing connections**
* New connections should use global settings even if the advanced settings page weren't loaded previously

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/PostgreConnectionPage.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/PostgreConnectionPage.java
@@ -271,7 +271,9 @@ public class PostgreConnectionPage extends ConnectionPageWithAuth implements IDi
     @Override
     public void saveSettings(DBPDataSourceContainer dataSource) {
         DBPConnectionConfiguration connectionInfo = dataSource.getConnectionConfiguration();
-        applyGlobalProperties(connectionInfo);
+        if (getSite().isNew()) {
+            applyGlobalProperties(connectionInfo);
+        }
         if (typeURLRadio != null) {
             connectionInfo.setConfigurationType(
                 typeURLRadio.getSelection() ? DBPDriverConfigurationType.URL : DBPDriverConfigurationType.MANUAL);
--- a/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/PostgreConnectionPage.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/PostgreConnectionPage.java
@@ -16,6 +16,8 @@
  */
 package org.jkiss.dbeaver.ext.postgresql.ui;
 
+import java.util.Locale;
+
 import org.eclipse.jface.dialogs.IDialogPage;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.ModifyListener;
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -16,6 +16,17 @@
...
+import java.nio.file.Path;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.time.ZoneId;
+import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
```
E1izabeth: Please change order of imports back.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/PostgreConnectionPage.java
```diff
@@ -16,6 +16,8 @@
  */
 package org.jkiss.dbeaver.ext.postgresql.ui;
 
+import java.util.Locale;
+
```
E1izabeth: Why did you move it at the beginning of imports?

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/PostgreConnectionPage.java
```diff
@@ -297,6 +301,24 @@ public void saveSettings(DBPDataSourceContainer dataSource) {
...
+    private static void applyGlobalProperties(DBPConnectionConfiguration connectionInfo) {
+        // Additional page can be not loaded at all, if this is the case just to be sure we need to apply global settings
+        String[] propertiesToCheck = {
+            PostgreConstants.PROP_SHOW_NON_DEFAULT_DB,
+            PostgreConstants.PROP_SHOW_UNAVAILABLE_DB,
+            PostgreConstants.PROP_SHOW_TEMPLATES_DB,
+            PostgreConstants.PROP_READ_ALL_DATA_TYPES,
+            PostgreConstants.PROP_READ_KEYS_WITH_COLUMNS,
+            PostgreConstants.PROP_SHOW_DATABASE_STATISTICS
+        };
```
E1izabeth, line 313: I don't like an idea of enumerating properties.

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/PostgreConnectionPage.java
Destrolaric: Fixed

plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/PostgreConnectionPage.java
serge-rider: The idea is that we don't need to set properties from additional page. Setting all properties looks really bad (we will push a lot of useless stuff in all connections and we won't be able to change defaults after it)

serge-rider (changes requested):

PR #22562, opened 2024-01-23
dbeaver/pro#2368 Use monitor from recognition context for query validation

E1izabeth:


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/context/SQLQueryCombinedContext.java
```diff
@@ -43,14 +44,14 @@ public SQLQueryRowsSourceModel findRealSource(@NotNull DBSEntity table) {
 
     @NotNull
     @Override
-    public DBSEntity findRealTable(@NotNull List<String> tableName) {
-        return anyOfTwo(parent.findRealTable(tableName), otherParent.findRealTable(tableName)); // TODO consider ambiguity
+    public DBSEntity findRealTable(List<String> tableName, DBRProgressMonitor monitor) {
```
LonwoLonwo: No @NotNull annotation anymore for tableName?

E1izabeth: thank you

LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/SQLQueryModelRecognizer.java
```diff
@@ -559,7 +572,7 @@ public void traverseObjs(Object obj) {
     }
 
     @Nullable
-    public SQLQuerySelectionModel recognizeQuery(@NotNull String text) {
+    public SQLQuerySelectionModel recognizeQuery(@NotNull String text, IProgressMonitor monitor) {
```
ShadelessFox: I'd use `DBRProgressMonitor` everywhere as possible. When calling such a method, convert the monitor.

I also want to note we mostly pass monitors as the very first parameter of a method. Additionally, nullability annotations would be nice to have here.

This applies to most changes in this PR

ShadelessFox (approved):

Matvey16:
Verified

Issue #2363, opened 2017-10-18
Bugs in version 4.2.3


Apcio:
After Update to 4.2.3 I noticed:
- Tab key change to result view instead input tabulator character into SQL Editor
- In result view I can't open popup menu for export wizard

I couldn't find post with that bugs, so this could be doubled, sorry for that.

//edit
I know why I couldn't open popup menu - after save SQL in menu SQL Editor -> Save SQL script popup menu is disabled. Run the same SQL in another SQL Editor allow popup menu for show and export results

serge-rider:
Duplicate of #2347 (both these bugs were resolved in hotfix version, just re-download 4.2.3)

PR #22570, opened 2024-01-23
#2363 Fix NPE in DDL view

E1izabeth:
Issue was reproducible this way:
1. Connect to PostgreSQL
2. Open any table
3. Switch to DDL tab
4. See NPE in logs

Should not be reproducible anymore.


LonwoLonwo (approved):

Matvey16:
Verified

PR #22575, opened 2024-01-24
dbeaver/pro#2318 Unix policy file

ShadelessFox:
A new file must be created in order to configure the policy: `~/.config/dbeaver/policy.properties`

An example:
```properties
policy.master.password.disable = true
policy.master.password.policy = { "minLength": 8, "minLetterCount": 1, "minNumberCount": 1, "minSymbolCount": 0, "requireMixedCase": true }
```

**Please test on Windows (using the registry), Linux, and macOS.**


LonwoLonwo (approved):

ShadelessFox:
We decided we don't need such a mechanism because we already have JVM properties that can be passed through the dbeaver.ini file.

PR #22598, opened 2024-01-26
#22478 Introduce CTE validation and outline support

### LINKED ISSUE #22478: MS SQL Syntax highlight for CTEs not working
### Description

I love the new syntax highlighting, but it doesn't seem to parse CTEs correctly in MS SQL: 
Screen Shot 2024-01-16 at 12 02 23 PM

Problems:
1. Items within the CTE are not colored correctly
2. In the main query which reference the CTE are (excepting the CTE's table alias) colored as errors

### DBeaver Version

Community Edition 23.3.2.202401072024

### Operating System

macOS Monterey 12.7.1

### Database and driver

### Steps to reproduce

1. Connect to a MS SQL db
2. Construct a query using a CTE.

### Additional context

_No response_

E1izabeth:
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/model/SQLQueryRowsCteModel.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/model/SQLQueryRowsCteModel.java
@@ -1,0 +1,156 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2024 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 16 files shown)


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/model/SQLQueryRowsCteModel.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.List;
+
+
+public class SQLQueryRowsCteModel extends SQLQueryRowsSourceModel {
+
+    private final boolean isRecursive;
+    private final List<SQLQueryRowsCteSubqueryModel> subqueries = new ArrayList<>();
+    private final SQLQueryRowsSourceModel resultQuery;
+
+    public static class SQLQueryRowsCteSubqueryModel extends SQLQueryRowsSourceModel {
```
ShadelessFox, line 38: Why is it nested?

E1izabeth: The idea is that it shouldn't exist without containing CTE model.
But it's public, because we want to see it in Outline, for example.
Do you think it's better to separate them?

ShadelessFox: Makes sense, okay

ShadelessFox (approved):

Matvey16:
Verified

PR #22616, opened 2024-01-29
dbeaver/dbeaver#20602 load controls for extra pages

### LINKED ISSUE #20602: Postgres: general setting 'show all databases' doesn't work, only the default database is shown
### Description

In the preferences under Connections/Drivers/PostgreSQL 'show all databases' and 'show databases not available for connection' is ticked. 
When creating a new connection you would assume that automatically all databases would be shown but that is not the case.
Nevertheless, if I check the connection settings of a connection I just created, under the PostgreSQL tab indeed 'show all databases' is ticked. Why doesn't it become visible? 
So I went to have a look at data-sources.json and saw that the driver properties part is missing.
The strange thing is when I edit the connection , save and reconnect suddenly the driver properties part appears:
e.g.
"provider-properties": {
					"@dbeaver-show-non-default-db@": "true",
					"@dbeaver-show-template-db@": "false",
					"@dbeaver-show-unavailable-db@": "false",
					"show-database-statistics": "true",
					"@dbeaver-read-all-data-types-db@": "false",
					"read-keys-with-columns": "false",
					"@dbeaver-use-prepared-statements-db@": "false",
					"postgresql.dd.plain.string": "false",
					"postgresql.dd.tag.string": "false"
				}
Why do I need to edit the connection to get the provider properties part in the data-sources.json? 
So as a workaround I figured, I could adapt the drivers.xml and add parameter lines in the driver id part. But I don't know if it would help and if so what the syntax would be?
I also checked the debug logging and there it states that it will start up the default database, no errors 

### DBeaver Version

Community edition 23.0.2

### Operating System

windows

### Database and driver

postgresql-42.6.0.jar

### Steps to reproduce

In the preferences under Connections/Drivers/PostgreSQL 'show all databases' and 'show databases not available for connection' is ticked. 
When creating a new connection you would assume that automatically all databases would be shown but that is not the case.
Nevertheless, if I check the connection settings of a connection I just created, under the PostgreSQL tab indeed 'show all databases' is ticked. Why doesn't it become visible? 
So I went to have a look at data-sources.json and saw that the driver properties part is missing.
The strange thing is when I edit the connection , save and reconnect suddenly the driver properties part appears:
e.g.
"provider-properties": {
					"@dbeaver-show-non-default-db@": "true",
					"@dbeaver-show-template-db@": "false",
					"@dbeaver-show-unavailable-db@": "false",
					"show-database-statistics": "true",
					"@dbeaver-read-all-data-types-db@": "false",
					"read-keys-with-columns": "false",
					"@dbeaver-use-prepared-statements-db@": "false",
					"postgresql.dd.plain.string": "false",
					"postgresql.dd.tag.string": "false"
				}
Why do I need to edit the connection to get the provider properties part in the data-sources.json?
So as a workaround I figured, I might need to adapt the drivers.xml and add parameter lines in the driver id part. I don't know if it would help and if so what the syntax would be?

### Additional context

_No response_

Destrolaric:
Now control for extra pages for Oracle and Postgre are loaded during main page initialization which should allow properties to be saved even if no page were loaded previously
Closes dbeaver/dbeaver#20602


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageSettings.java
```diff
@@ -192,6 +191,31 @@ public void activatePage() {
...
+                        if (item.getData().equals(requiredDialogPage)) {
+                            Composite panel = (Composite) item.getControl();
+                            try {
+                                panel.setRedraw(false);
+                                requiredDialogPage.createControl(panel);
+                                Dialog.applyDialogFont(panel);
+                                panel.layout(true, true);
+                                requiredDialogPage.setVisible(false);
+                            } finally {
+                                panel.setRedraw(true);
```
serjiokov, line 210: The finally - clause generally defined for completing exceptional operation like: close resource, streams etc Are you sure for setRedraw here?

serge-rider: I think it is ok. We use it widely.
finally is not for resources close but for any operations which must be called in the end of block.

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/IDialogPageProvider.java
```diff
@@ -28,4 +28,17 @@ public interface IDialogPageProvider {
     @Nullable
     IDialogPage[] getDialogPages(boolean extrasOnly, boolean forceCreate);
 
+    /**
+     * Pages what should be saved during creation, even if they were
```
E1izabeth, line 32:
```suggestion
     * Pages that should be saved during creation, even if they were
```

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/IDialogPageProvider.java
```diff
@@ -28,4 +28,17 @@ public interface IDialogPageProvider {
...
+     * never opened
+     *
+     * @return array of IDialogPage[] to create at the initialisation,
+     *      null if no additional pages should be created.
+     */
+    @Nullable
+    default IDialogPage[] getRequiredDialogPages() {
```
E1izabeth, line 39: Method name is not descriptive enough. I couldn't understand what does it mean, until I read the comment. I'd propose something like `getPagesContainingSettings`. Maybe you can figure out better name, but don't left just as is, please. It's not clear what does Required means - required for what?

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageSettings.java
```diff
@@ -192,6 +191,31 @@ public void activatePage() {
         //getContainer().updateTitleBar();
     }
 
+    private void createAdditionalRequiredPages() {
```
E1izabeth, line 194: I doubt about calling this method `create..`. I would rather expact to call it like `load..` or smth like that. Yes, you call createControl method, but the control is not visible and so on..

ShadelessFox (commented):
I'm still concerned with the proposed solution.

PR #22623, opened 2024-01-29
dbeaver/pro#2285 Fix parallel execution and end of transfer notification

ShadelessFox:
**STR**:
1. Select several tables and create a new export task (there must be one task with several tables)
2. On the **Extraction settings** page, set the **Maximum threads** to a value bigger than one, e.g. 5
3. Execute task

Expected behavior:
- [x] Email is no longer sent several times (https://github.com/dbeaver/pro/issues/2285)
- [x] Tasks with multiple source tables now should execute a bit faster than before
- [x] Email includes just one attachment, not all attachments for all tables (dbeaver/pro#2409)


LonwoLonwo (approved):

E1izabeth (approved):

MashaKorax:
verified

PR #22628, opened 2024-01-29
Cb 4457 subcollections firestore

DenisSinelnikov:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBCHierarchicalDocumentProvider.java
```diff
@@ -0,0 +1,39 @@
+/*
```
serge-rider: It must be in PRO repo

serge-rider (changes requested):

serge-rider (approved):

PR #22630, opened 2024-01-29
dbeaver/pro#2364 Error in Outline when saving Console

serjiokov:
dbeaver/pro#2364 Fixed exception in outline after save console result as script


serjiokov:
https://github.com/dbeaver/dbeaver/assets/32031248/6f590734-f674-4ae4-9d3b-fe53e6f26449

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorBase.java
```diff
@@ -640,6 +640,9 @@ public <T> T getAdapter(Class<T> required) {
 
     @NotNull
     private IContentOutlinePage getOverviewOutlinePage() {
+        if (this.getSyntaxContext() == null) {
+            this.reloadSyntaxRules();
+        }
```
E1izabeth, line 645: What concrete scenario requires this code?

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorOutlinePage.java
```diff
@@ -148,7 +153,7 @@ public void updateChildCount(@NotNull Object element, int currentChildCount) {
                 } else if (element instanceof OutlineNode node) {
                     treeViewer.setChildCount(element, node.getChildrenCount());
                 } else {
-                    throw new UnsupportedOperationException();
+                    treeViewer.setChildCount(element, rootNodes.size());
```
E1izabeth: Why intentionally supporting root object model (which is EditorInput) being out of sync with the editor instead of just listen for the corresponding event of it being changed and update the outline model to be in sync?

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorOutlinePage.java
```diff
@@ -127,12 +130,14 @@ public void createControl(@NotNull Composite parent) {
...
             @Override
             public void updateElement(@NotNull Object parent, int index) {
-                if (parent == editor.getEditorInput()) {
-                    this.updateChildNode(parent, index, rootNodes.get(index));
-                } else if (parent instanceof OutlineNode node) {
+                if (parent instanceof OutlineNode node) {
                     this.updateChildNode(parent, index, node.getChild(index));
                 } else {
-                    throw new UnsupportedOperationException();
+                    if (index <= rootNodes.size() - 1) {
```
E1izabeth: This is wrong. Such a condition can never be false while the data model is in sync. Why don't just support the data model contract as it is required by the OutlinePage - so that the data model of the outline is always related with the real state of the owning editor?

E1izabeth (changes requested):
All outline parts should be in sync with editor in each moment of time. At least because outline is not purely ours and editor input may be used by the platform.

What happened: editor input changed and the old object was disposed, but stayed as a data model of the outline tree.
What the correct fix is: subscribe to the editor input change event and update cached editor input in the outline page.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorBase.java
serjiokov: The case:  Open outline view first when no editor exist, next try to open any script in SQL Editor (Outline is empty)

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorOutlinePage.java
```diff
@@ -191,12 +177,16 @@ public void selectionChanged(@NotNull SelectionChangedEvent event) {
...
-        }
-        this.refreshJob.schedule(500);
+        UIUtils.asyncExec(new Runnable() {
+            @Override
+            public void run() {
+                if (!treeViewer.getTree().isDisposed()) {
+                    treeViewer.refresh();
+                }
+            }
+        });
```
ShadelessFox: We're using Java 17, not Java 7.

```suggestion
        UIUtils.asyncExec(() -> {
            if (!treeViewer.getTree().isDisposed()) {
                treeViewer.refresh();
            }
        });
```

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorOutlinePage.java
ShadelessFox: Is this resolved?

E1izabeth: I'm still not sure that we should do this change, but ok, let's try.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorOutlinePage.java
serjiokov: Done

E1izabeth (changes requested):
All commits reference wrong issue

E1izabeth:
I tried to fix the issue myself. From my point of view, here are all required changes to fix error described in the issue.

IMG_20240131_122742_802

serjiokov:
As I see, we already have listener on editor (document ) not needed another one
 final SQLDocumentSyntaxContextListener syntaxContextListener = new SQLDocumentSyntaxContextListener()

E1izabeth:
Disposed and invalidated input model should not be referenced by any piece of infrastructure, especially when we have the other actual one. Violation of this leads to memory leaks and hidden errors, which may not be observable until an unforseen combination of circumstances araise.

We have already seen such things a number of times in a various parts of the product.

E1izabeth:
> As I see, we already have listener on editor (document ) not needed another one final SQLDocumentSyntaxContextListener syntaxContextListener = new SQLDocumentSyntaxContextListener()

This listener is about script elements discovered by the parsing job for the syntax model of the observable fragment of the document. They are being updated by the parsing job when the observable fragment changes and when its syntax model affected. It's not intended to update the root and the whole model.

The proposed listener works when the whole editor input model is changed meaning that previous model was disposed and not used by the infrastructure anymore, so all the corresponding resources should be released.

E1izabeth:
Please note that UnsupportedOperationException was written in these two places intentionally - it helps to detect invalid state and find the bug. Let's don't remove it.

serjiokov:
Implemented Inputchange listener in order to handle switch of editor input.

E1izabeth:
Please call removeTextInputListener here https://github.com/dbeaver/dbeaver/pull/22630/files#diff-05434f606f2d67adb338d5c36f8e52fe7347ce10a6be9d1aab6d80fe030f1b2fR236

E1izabeth (approved):

E1izabeth:
@ShadelessFox now it's fine. Please review too.

MashaKorax:
verified

PR #22635, opened 2024-01-30
#22619 fix custom name in the FK create dialog; add custom name field…

### LINKED ISSUE #22619: Cannot rename foreign key from ui
### Description

STR

1. Connect to Oracle DB
2. Open any table with primary key or unique key
3. Open tab "foreign keys" and create FK **(Rename the FC at the time of creation**)

Actual result - The FK name has not changed from default to custom, 
The problem is that when creating several FKs there will be a name duplication error

https://github.com/dbeaver/dbeaver/assets/84867187/4542d194-7084-479b-af12-a6e6bc4b4d15

### DBeaver Version

DBeaver 23.3.4

### Operating System

windows 10

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

### LINKED ISSUE #22620: [Postgres] Add name field for FK
### Description

dbeaver_5FJyJwjx1W
dbeaver_akXfKk6xaT

### DBeaver Version

DBeaver 23.3.4

### Operating System

windows 10

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

LonwoLonwo:
The name field changed to the "Custom name" in the FK creation dialog
Now, this field is empty by default
A new FK name will be created automatically if you leave the "Custom name" field empty.
A new FK will be used from the "Custom name" field if you change it.
Added the custom name field for PG.

2024-01-30 12_27_49-DBeaver 23 3 4 - child


plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/internal/EditorsResources.properties
```diff
@@ -29,7 +29,7 @@ dialog_struct_edit_fk_column_ref_col_type = Ref Column Type
 dialog_struct_edit_fk_combo_on_delete = On Delete
 dialog_struct_edit_fk_combo_on_update = On Update
 dialog_struct_edit_fk_combo_unik = Unique Key
-dialog_struct_edit_fk_name = Name
+dialog_struct_edit_fk_name = Custom Name
```
ShadelessFox, line 32: But why, tho? What does "custom" mean in the context of a foreign key creation?

LonwoLonwo: Here, custom means - not auto-generated by DBeaver. 
For users who want to specify this FK name. 
Otherwise, it will be generated by DBeaver.

ShadelessFox: Is it possible to generate a name beforehand and show it in this field (or even using `Text#setMessage`)? It would be a nicer UX IMO

LonwoLonwo: Nope, I tried - it is a timekiller work. You can create a separate ticket for this.

ShadelessFox (approved):

uslss:
verified

PR #22661, opened 2024-02-01
#22641 Fix keywords highlighting and schema name in conditions highlighting

### LINKED ISSUE #22641: Wrong syntax highlight for Oracle DB
### Description

Sometimes syntax highlighting doesn't work for schema names (AUTH for example)

### DBeaver Version

23.3.3

### Operating System

macOS 14.3

### Database and driver

Oracle 19
org.jkiss.dbeaver.ext.oracle, Version: 2.1.208.202401211839
org.jkiss.dbeaver.ext.oracle.ui, Version: 1.0.110.202401211839

### Steps to reproduce

I don't know how to reproduce it exactly, I have this bug in 2 queries on my screen. These requests work without problems.

E1izabeth:
- [x] Keywords should be highlighted with dark red no matter which case is used
- [x] Schema name should be highlighted with different column in where and select list, when it's valid. For example:
image
(To be highlighted, schema name should be in from clause too - otherwise the query is most probably invalid)


ShadelessFox (approved):

serjiokov (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #22684, opened 2024-02-02
dbeaver/pro#2428 Change label of checkbox to save ssh credentials

E1izabeth:
image


ShadelessFox (approved):

Destrolaric:
@E1izabeth If we want to change the label, you can also change it in `dialog_connection_auth_checkbox_save_password`

E1izabeth:
I would not change that label

MashaKorax:
verified

PR #22690, opened 2024-02-02
dbeaver/pro#2283 fix incorrect prompt conditions & save settings

Destrolaric:
Closes dbeaver/pro#2283
Fix incorrect prompt conditions

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/metadata/MetadataProcessor.java
+++ b/plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/metadata/MetadataProcessor.java
@@ -122,9 +122,10 @@ public class MetadataProcessor {
                 "Your query must start with \"SELECT\" and be enclosed with triple backslash on new lines. " +
                 "Talk naturally, as if you were talking to a human.");
         } else if (isChatAPI) {
-            sb.append("Perform SQL completion. Start response with SELECT keyword. Start every line with \"--\".");
+            sb.append("Perform SQL completion. Start response with SELECT keyword.\n" +
+                " Avoid using Markdown. Start every comment line with \"--\".");
         } else {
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

ShadelessFox (approved):
Please test against OpenAI and Gemini models.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/metadata/MetadataProcessor.java
```diff
@@ -122,9 +122,10 @@ public DAICompletionMessage createMetadataMessage(
                 "Your query must start with \"SELECT\" and be enclosed with triple backslash on new lines. " +
                 "Talk naturally, as if you were talking to a human.");
         } else if (isChatAPI) {
-            sb.append("Perform SQL completion. Start response with SELECT keyword. Start every line with \"--\".");
+            sb.append("Perform SQL completion. Start response with SELECT keyword.\n" +
+                " Avoid using Markdown. Start every comment line with \"--\".");
         } else {
-            sb.append("Perform SQL completion.");
+            sb.append("Perform SQL completion. Avoid using Markdown");
```
ShadelessFox, line 128: Probably lines 124-129 could be just
```java
        } else {
            sb.append("Perform SQL completion. Start your response with \"SELECT\".");
        }
```
Like it was before

ShadelessFox (changes requested):

Matvey16:
Verified

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/metadata/MetadataProcessor.java
Destrolaric: No, unfortunately Gemini tends to use markdown if not discouraged

PR #22713, opened 2024-02-05
#22594 Remove keyword VIEWS from Snowflake

### LINKED ISSUE #22594: Table names identical to any sql command must be enclosed in quotation marks or acting strangely.
### Description

dbeaver autocomplete puts table names that match any sql command in quotation marks.
If I do not write the name of the object name in lowercase letters or put it in quotation marks, then Query Analysis feature does not recognize it as an object.

If I write the object name in lowercase letters then the column name auto completion doesn't works, but the Query Analysis does

Other objects whose name does not match any sql command, works well with uppercase, lowercase and quotation marks too.

kép

### DBeaver Version

Community Edition 23.3.3.202401211839

### Operating System

Windows 10

### Database and driver

I could only test this with Snowflake.

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Should fix the described issue


ShadelessFox (approved):

LonwoLonwo (approved):

uslss:
verified

PR #22719, opened 2024-02-06
#22552 Show query segment instead of question marks in the outline view

### LINKED ISSUE #22552: Show query segment instead of question marks in the outline view
Take a look at the following query:
```sql
select 1, 2, 3, 4, 5;
```

It will be shown as this in the outline view:
image

Note that columns are shown as question marks. It will make much more sense if textual presentation of these values was used instead.

E1izabeth:
image


LonwoLonwo (approved):

serjiokov (approved):

uslss:
verified

PR #22725, opened 2024-02-06
dbeaver/dbeaver#22721 Idle connection config + UI

### LINKED ISSUE #22721: Close idle connection setting not saved when restarting dBeaver
### Description

Recently I noticed that DBs get disconnected after a while, so I found the "Close idle connection after (seconds)" setting in the Connection Types. I uncheck it, and apply. After dBeaver restart it is checked again...
I notice the connection is closed even if I have an open dashboard, so the connection shouldn't be idle...
image

### DBeaver Version

23.3.4.202402041841

### Operating System

Linux Mint 20.3

### Database and driver

Postgres 14

### Steps to reproduce

_No response_

### Additional context

_No response_

serge-rider:


Matvey16:
Verified

PR #22729, opened 2024-02-06
CB-4565. Added validation for hsqldb file connection

DenisSinelnikov:


plugins/org.jkiss.dbeaver.ext.hsqldb/src/org/jkiss/dbeaver/ext/hsqldb/model/HSQLMetaModel.java
```diff
@@ -47,13 +47,18 @@
...
+    private static final String PROHIBITED_FUNCTION = "file";
 
     public HSQLMetaModel() {
         super();
     }
 
     @Override
     public GenericDataSource createDataSourceImpl(DBRProgressMonitor monitor, DBPDataSourceContainer container) throws DBException {
+        String url = container.getConnectionConfiguration().getUrl();
+        if (url != null && url.contains(PROHIBITED_FUNCTION)) {
```
ShadelessFox: It's too aggressive, I think. What about checking if the URL starts with `jdbc:hsqldb:file`?

https://www.hsqldb.org/doc/2.0/guide/running-chapt.html#rgc_inprocess

PR #22743, opened 2024-02-07
#19401 avoid functions duplication

### LINKED ISSUE #19401: Duplicates procedures names
### Description

Hello,
We are encoutering an issue with duplicate functions names when using DBeaver with PostgreSQL database.
Here is an example :

[](url)
image

The functions are exactly identical, and queries in system tables show that the duplicate functions are existing only once in the database.
pgAdmin, which is the official tool to connect to PostgreSQL database does not show the problem.

### DBeaver Version

23.0.0.202303040621

### Operating System

windows

### Database and driver

Database PostgreSQL 13.10
Driver postgresql-42.5.0.jar

### Steps to reproduce

Not sure how you can reproduce. 

The problem seems to be linked to the following query launched by DBeaver to display procedures list:

 SELECT p.oid as poid,p.*,pg_catalog.pg_get_expr(p.proargdefaults, 0) as arg_defaults,d.description
	FROM pg_catalog.pg_proc p
	LEFT OUTER JOIN pg_catalog.pg_description d ON d.objoid=p.oid
	WHERE p.pronamespace=$1
	ORDER BY p.proname;

In our system, some procedures have the same oid as some tables, which seems to happen in postgres.
And in this case, the query has collected table and columns descriptions for the procedure, as they have the same oid.

Procedure description is only 1 row, but you may have multiple rows for table and columns, wich leads to duplicate display.

Hope this is clear enough.

### Additional context

_No response_

LonwoLonwo:
>  The regclass input converter handles the table lookup according to the schema path setting, and so it does the "right thing" automatically.
https://www.postgresql.org/docs/8.1/datatype-oid.html

Comment from the ticket: 

> Table pg_description has namespace for objects like tables, types etc. also parameter oid is unique only in particular classoid not globaly. If you join to this table you must specify classoid otherwise, duplicates may occur

https://github.com/dbeaver/dbeaver/issues/19401#issuecomment-1929653603

Please check the functions list with or without descriptions in any old PG and in the last PG 16

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
@@ -1181,12 +1181,14 @@ public class PostgreSchema implements
         public JDBCStatement prepareLookupStatement(@NotNull JDBCSession session, @NotNull PostgreSchema owner, @Nullable PostgreProcedure object, @Nullable String objectName) throws SQLException {
             PostgreServerExtension serverType = owner.getDataSource().getServerType();
             String oidColumn = serverType.getProceduresOidColumn(); // Hack for Redshift SP support
+            boolean versionAtLeast7 = session.getDataSource().isServerVersionAtLeast(7, 2);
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

Destrolaric (approved):

Matvey16:
Verified

PR #22747, opened 2024-02-07
dbeaver/pro#2451 Remove unnecessary default table model from CTE handling common for different SQL statements (not only select, but delete, update, insert)

E1izabeth:


Destrolaric (approved):

LonwoLonwo (approved):

ShadelessFox (approved):

LonwoLonwo (approved):

Matvey16:
Verified

Matvey16:
Verified

Matvey16:
Now column names in any DDL or DML are highlighted in dark blue if they are quoted
image

E1izabeth:
It is expected behavior. We haven't implemented any validation for create/delete/update/insert statements. Now only fallback is implemented to make it work as before.

Matvey16:
Verified

PR #22755, opened 2024-02-08
CB-4365. Support idle connection closing

DenisSinelnikov:


yagudin10 (commented):
i think it's better use
`DBWorkbench.getPlatform().getApplication().getLastUserActivityTime()` in SQLEditor

PR #22768, opened 2024-02-09
dbeaver/dbeaver#22126 replace load next page binding

### LINKED ISSUE #22126: Replace keybinding for "Fetch Next Page"
When I have a query result that has more rows than the number of fetched rows (200 for my example) and I want to write a semicolon, it doesn't write ; but instead fetches the next page.
I found that this can be done with the ctrl+alt+; key combination.
However, I can on a Hungarian keyboard the semicolon sign can be written by ctrl+alt+, which will act the same way as ctrl+alt+;
Therefore, semicolon and fetch next page is on the same keyboard combination for me.

Having a different keyboard combination for "Fetch Next Page" fixed the problem on my end, so it can be useful for others too.

Destrolaric:
Closes dbeaver/dbeaver#22126


E1izabeth (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #22771, opened 2024-02-09
#22766 Fix user credentials dialog not having a title

### LINKED ISSUE #22766: Window Title missing from authentication prompts in 23.3.4, so Password Manager cannot identify the Window for Auto Type
### Description

Since upgrading to 23.3.4, the popup authentication prompt when connecting to a database no longer has a Window Title. This means my Password Manager cannot Auto-Type the password for the database, as it uses the Window Title to determine which password should be chosen to be auto-typed for that database.

The same occurs for the 'SSH Tunnel' authentication also, this popup's Window Title is also missing and again the Password Manager cannot select the right password to Auto-Type.

I narrowed this down to `plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ui/DBPPlatformUI.java`, lines 124 and 146.

In a recent commit, these methods have been added to add a description to the user credentials prompt. On these lines, the first parameter refers to the Window Title, and the second parameter to the new description label on the popup window. The Window Title (parameter name `prompt`) is being set to `null`, and the inputted `prompt` is passed to the `description` parameter.

I tried replacing the `null` parameters with the variable `prompt` (so that `prompt` is used twice in the call), and can confirm that the Window Title returned and the Password Manager can once again Auto-Type the password.

Screenshot of DBeaver 23.3.4 version (missing Window Title):

image

Screenshot of patched version (restoring Window Title):

image

Here is a patch file, but it may be easier to just overwrite both `null`s with `prompt`.

[fix-window-title-on-authentication-popups.patch](https://github.com/dbeaver/dbeaver/files/14217359/fix-window-title-on-authentication-popups.patch)

### DBeaver Version

Community 23.3.4

### Operating System

Windows 10

### Database and driver

Independent of database and driver, but noted occurring when connecting using drivers mysql8, MariaDB, Microsoft SQL Server.

### Steps to reproduce

1. Create a database connection; enter the username, but do not choose to Save the password

Screenshot of my `localhost` configuration:
image

2. If connected, disconnect from the database
3. Right-click on the database connection - click Connect

Observe that there is no Window Title on the authentication popup box

### Additional context

As noted in 'Steps to reproduce', my database connection does not have the password saved in DBeaver.

My database passwords are stored in my Password Manager, for which I use KeePassXC on Windows.

I can't take a screenshot of KeePassXC for whatever reason, but the KeePassXC entry has under Auto-Type a Window Association for the 'Window title' being `'localhost' Authentication`, that is what chooses the right KeePassXC entry to Auto-Type for that database connection.

My KeePassXC Auto-Type global keyboard shortcut is Alt-Shift-R so that it doesn't conflict with other shortcuts. That means that I can press Alt-Shift-R at any time, and if a DBeaver authentication popup is open (including for SSH Tunnels) KeePassXC will Auto-Type the password quickly and press Enter. Such a time-saver compared to opening KeePassXC and finding the right password.

I noticed issue #22748 . I use KeePassXC whereas they appear to be using KeePass, both use the same database but have different user interfaces. However, it is possible that the lack of a Window Title might be confusing KeePass into typing the password into the wrong place? I could not replicate that issue in my setup either pre-patch or post-patch.

ShadelessFox:


E1izabeth (approved):

LonwoLonwo (approved):

Matvey16:
Verified

Issue #22657, opened 2024-02-01
automatic row count


JeroenNienhuis:
For a row count on a (result) table yu have to click on the count button on the bottom toolbar.
I often have large files and want to know the number of records or use the count as a check on sql command

Is it possible to calculate it by default? Perhaps add it as an option to do so on the data editor settings.

E1izabeth:
Thank you for the idea

GiorgioSgl:
This will be useful!

ShadelessFox:
We have discussed this feature and propose the following requirements:
1. Row count must be calculated only once when opening the data viewer. It should not be triggered when new data is fetched.
2. The "data editor" preferences must have a setting to turn it on/off.

JeroenNienhuis:
Does "only once when opening the data viewer" mean "every time an sql is executed"
and "not be triggered when new data is fetched" mean "not every time a next portion from LIMIT OFFSET is fetched"?
That would be nice.

ShadelessFox:
Correct.

PR #22775, opened 2024-02-09
feat: automatic row count

utsav1704:
feat: #22657
Automatic Row Count

https://github.com/dbeaver/dbeaver/assets/75378586/b1c163cf-3a40-454a-a6c5-bd2277926530

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
@@ -2386,6 +2386,7 @@ public class ResultSetViewer extends Viewer
         }
         this.curRow = null;
         this.model.setData(rows);
+        updateRowCount();
         this.curRow = (this.model.getRowCount() > 0 ? this.model.getRow(0) : null);
         if (focusRow > 0 && focusRow < model.getRowCount()) {
             this.curRow = model.getRow(focusRow);


LonwoLonwo:
Hello @utsav1704 
Thanks for your contribution.
We will take a look at it.

ShadelessFox (commented):
Please demonstrate how it would work when fetching batches of data. The default fetch size is 200 rows. What if you have more than that? Will it show "200, 400, etc" or "200+, 400+, etc"?

utsav1704:
If we have more rows than fetch size, it will show the exact number of rows.
As you can see in the video, I set the fetch size to 20, and I have 21 rows in the student table and 24 rows in the product table.

ShadelessFox:
21 rows are not enough for testing, they all were loaded in the memory at the same time. Please test against a table with several hundreds rows at least.

utsav1704:
I tried this changed code on the table, which has around 16000 rows and a default fetch size of 200. When I'm loading the table or refreshing the table, it shows the exact number of rows.

ShadelessFox:
Ah, I see. It basically executes the "Row Count" action. This change means that:
1. This action is now redundant
2. This might be an expensive operation that will be called every time we fetch rows. We probably don't want that.

I'm going to discuss this change with my colleagues.

ShadelessFox:
I'm sorry, we don't like this idea, even if there were an option to turn it on/off.

As noted, this might be a costly operation for some databases.

utsav1704:
Okay. I try to see if we can do this one time while loading the table.

ShadelessFox:
I've added some information to the original ticket. Please see https://github.com/dbeaver/dbeaver/issues/22657#issuecomment-1938736014

PR #22781, opened 2024-02-11
#22356 Fix scrolling in single tab result sets mode

### LINKED ISSUE #22356: Vertical scrolling with the mouse wheel is disabled when in multiple-tab layout
### Description

It seems as though vertical scrolling with the mouse wheel is disabled when in the multiple-tab layout. I've directly compared functionality to the single tab layout, which is still scrollable. Ideally, each pane should scroll while hovering, but at a minimum could at least scroll after activating the specific result pane.

### DBeaver Version

Community Edition 23.3.1.202312241705

### Operating System

Windows 10 10.0.19045

### Database and driver

SQL Server 2019 / Microsoft JDBC Driver for SQL Server

### Steps to reproduce

1. Toggle "Show results in a single/multiple tabs" setting to "multiple tabs" in the "SQL Editor" menu.
image

3. Select multiple individual queries and simultaneously execute with the "execute SQL script" button.
4. Results appear in multiple panes at the bottom. **Attempt to scroll with the mouse wheel while hovering over a specific pane and the results do not scroll.**
5. Click within a result pane and **attempt to scroll with the mouse wheel while the pane is active. The results still do not scroll.**
image

Check against default functionality:
6. Toggle the "Show results in a single/multiple tabs" setting back to "single tab" in the "SQL Editor" menu.
7. Select multiple individual queries and simultaneously execute with the "execute SQL script" button.
8. Results appear in multiple tabs at the bottom. **Scrolling with the mouse wheel while hovering over the results pane produces the desired effect of scrolling through the results.**

### Additional context

Initiated from discussion in issue #6861, fixed by #21280. 
https://github.com/dbeaver/dbeaver/issues/6861#issuecomment-1878845621

Referred by @E1izabeth to create new issue.

E1izabeth:
If the cursor on the results grid, then we scroll data grid.
If the cursor is outside data grid or other scrollable control, then we scroll outside.
If `ctrl` is pressed and we're trying to scroll, then we always scroll outside the data grid.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
@@ -3901,7 +3901,7 @@ public class SQLEditor extends SQLEditorBase implements
 
             Listener scrollListener = event -> {
                 Control underScroll = (Control) event.widget;
-                if (underScroll.getShell() == tabContentScroller.getShell() && tabContentScroller.isVisible()) {
+                if (underScroll.getShell() == tabContentScroller.getShell() && tabContentScroller.isVisible() && ((event.stateMask & SWT.CTRL) == SWT.CTRL)) {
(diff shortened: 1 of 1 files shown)


Matvey16:
Verified

PR #22785, opened 2024-02-12
 #21209 check data source settings and connection type for correct settings showing

### LINKED ISSUE #21209: Smart commit mode not working by default on connections made through CLI
### Description

When adding a new connection in DBeaver through the Command Line Interface, the connections never use Smart Commit, even when you have set your connection types to use Smart commit mode by default.

All my connection types are set to use 'Auto-commit mode by default' and 'Smart commit mode'.
image

The command I use to add a new connection:
dbeaver-cli.exe -con "driver=mysql|autoCommit=true|host=Hostname.host.com|port=3306|user=USER|database=DATABASE|password=PASSWORD|name=Dummy connection name"

After adding the connection and opening a script, the connection is not using smart commit mode:
image

Note: This feature was working fine in an earlier version of DBeaver, before the 'smart commit mode' checkbox was moved from the 'Transactions' menu to the 'Connection Types' menu in the preferences.

### DBeaver Version

23.2.0

### Operating System

Windows 10

### Database and driver

MySQL 8
com.mysql.cj.jdbc.Driver
8.0.29

### Steps to reproduce

1. Set all your connection types to use Smart commit mode
2. Add a new connection through the CLI using the earlier command
3. Open a script with the new connection, and see that Smart Commit is not turned on

### Additional context

_No response_

LonwoLonwo:
Now, the transaction page will first check the data source settings, then the connection type, and then - the global settings. To show correct checkboxes states and idle transactions period.


E1izabeth (approved):

uslss:
smart commit is enabled in connection type and is not set in the data source settings - smart commit doesn't work


https://github.com/dbeaver/dbeaver/assets/31996417/3ec7f2ef-1fdb-4373-a813-1c7534093611

LonwoLonwo:
We need to create a separate ticket to synchronize connection type settings changes. Now, they are working only after DBeaver restarts.

uslss:
verified

Issue #10561, opened 2020-11-30
Drop down menu for enum field in table editor not populating.


cbf305:
#### Client system info: 
- Linux (Arch - 5.9.9 Kernel)
- DBeaver version 7.2.5.202011152110
- Additional extensions - None

#### Server Info:
- Database MariaDB 10.5.8-1 (Remote server is running Ubuntu 18.04)
- Driver 2.4.3
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No

#### Describe the problem you're observing:
This has been an issue for a long time, but I just thought it was because I was running herbstluftwm and not a full DE, but recently I had a need to use dbeaver on a system with KDE and a Windows 10 system and both also did the same thing.  Since I've tested this on several Arch and Ubuntu systems running, i3, KDE and herbstluftwm, as well as several different Windows 10 machines.  I've also tested both MariaDB and MySQl servers and no matter the combination, they all do the same thing.  When in the table editor and I click on a enum field to edit, it used to give a drop down list of the possible enum values.  Now it displays the down arrow, but it never shows the values.  If you use the scroll wheel on the mouse or two finger scrolling on a touch pad it will scroll through the values, but the menu won't display.

I am using the built-in Dark theme, but it exhibits the same behavior under any theme.

Screenshot showing the incomplete down down menu.  There are 5 values for this enum field.
![2020-11-30-115447_2560x1440_scrot](https://user-images.githubusercontent.com/2612500/100639855-b3d60c00-3303-11eb-8c74-b21a69b06a0c.png)

I am not sure if it is related to closed issue #2468 or not since there was not a description given.

#### Steps to reproduce, if exist:

1. Create fresh DB on MariaDB or MySQL server
2. Create fresh table with at least one field of type ENUM with more then one value
3. Open table editor and try to change the enum field value by double clicking on it

#### Include any warning/errors/backtraces from the logs
There are no log errors

Thanks!

uslss:
Thanks for the report.
As a workaround, you can use Cell editor (Choose cell and then use Shift+Enter) or Value panel (Ctrl+7)

cbf305:
Thanks for the tip.  I've been using the value panel as a workaround, but I didn't know about the cell editor.

flambchap:
I found out that clicking on an inline editor on Linux triggers a FocusLost event. However, the reason for that is still a mystery to me. Needs more investigation.

flambchap:
Fixed.

[...]

PR #22788, opened 2024-02-12
dbeaver/dbeaver#22103 replace autosave focus listener with dispose & tabbing support for datetime

### LINKED ISSUE #22103: tab away from datetime field doesn't work as expected
When I am editing a datetime field, and I tab away from it, I am taken first to the "value" tab button.  A second tab takes me to the value input field.  This is different than what happens in a string field, in which a tab acts as it does in excel and moves me to the next field.  If I continue tabbing, I end up with a null in the date field.  The only way to preserve the value that I have entered in the date field is to move my mouse and click on a different field.

I believe that tabbing away from a date field should act the same as tabbing away from a text field

**Additional context**

Destrolaric:
Closes dbeaver/dbeaver#22103
Removes hacky way of autosaving for Linux, so it should be tested #10561
*Needs to be tested on MacOS and Linux*
Focus listener is unfortunately not viable for supporting traverse inside the multi control composites, so it was replaced with dispose listener.

Test that all values save properly after inline editing


MashaKorax:
**Mac issue** : if the cell is in edit mode and open any application or editor over the app, it becomes full screen
**Common issue:** if the cell is in edit mode and go to the value panel, make changes there - nothing happens when you press the apply button, but only when the focus changes 

https://github.com/dbeaver/dbeaver/assets/84867187/ce556226-41e8-4d0c-a0ea-216afed2b311

MashaKorax:
value panel is not refreshed
https://github.com/dbeaver/dbeaver/assets/84867187/aa0220f1-89d7-4140-8769-7e01e472177f

MashaKorax:
This case is broken again https://github.com/dbeaver/dbeaver/issues/10561

MashaKorax:
https://github.com/dbeaver/dbeaver/issues/10561 broken on linux

MashaKorax:
verified

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/spreadsheet/Spreadsheet.java
```diff
@@ -370,6 +374,9 @@ public void handleEvent(final Event event) {
                 // we don't want to mess current grid state
                 UIUtils.asyncExec(() -> presentation.navigateLink((GridCell) event.data, event.stateMask));
                 break;
+            case SWT.MouseExit:
+                System.out.println("Spreadsheet.handleEvent()");
```
serge-rider: Please remove

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/spreadsheet/Spreadsheet.java
serjiokov: fixed

PR #22791, opened 2024-02-12
Cb 4616 add ip to qm

DenisSinelnikov:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/qm/QMSessionInfo.java
```diff
@@ -32,4 +35,8 @@ public String getUserName() {
     public String getUserDomain() {
         return userDomain;
     }
+
+    public String getUserIp() {
+        return userIp != null ? userIp.replaceAll("[\\[\\]]", "") : null;
```
serge-rider: Let's not do it here (do it when passing IP from servlet)

serge-rider (changes requested):

PR #22811, opened 2024-02-13
dbeaver/pro#2373 [ERD] Overlapping entity by connections

serjiokov:
Recompute route with checks of bend point on a route


serjiokov:
https://github.com/dbeaver/dbeaver/assets/32031248/486b22f0-536f-481c-b568-110bb80a5c1a

uslss:
verified

Issue #19414, opened 2023-03-17
Table last update info


JoaoMoro:
I'm new to dbeaver and I don't know if there's a way, if there is please tell me how, but how can I see the last time a specific table was updated?

ftsfranklin:
What database are you using?

If it's Oracle, it's a duplicate of #11179.

If it's not, then @ShadelessFox said,
> Not every database offer such information, so we can't retrieve it and display somewhere.

JoaoMoro:
I am using SQL SERVER

LonwoLonwo:
Hello @JoaoMoro 

Now we do not have representation of this info for SQL server tables.

We can add your request to our backlog.

For now, you can use this query:

`select * FROM master.sys.all_objects o WHERE o.type IN ('U','S','V','TT') AND o.schema_id = 1`

with your schema id (you cab find it with double click on the schema object in the database navigator)

PR #22821, opened 2024-02-14
#19414 add more table/view properties; fix tw localizations

LonwoLonwo:
For tables and views added more properties:

2024-02-14 11_52_04-DBeaver Ultimate 23 3 5 - Department

Localization fix - removed these symbols:

2024-02-14 14_42_05-DBeaver – org jkiss dbeaver ext postgresql_…_bundle_tw properties  org jkiss dbe


E1izabeth:
Please revert all these changes in properties files :(
I'm sure we don't want to lowercase all symbols

LonwoLonwo:
We need to have some standards. And we decided to choose lowerCase - https://github.com/dbeaver/dbeaver/wiki/Localization#intellij-idea-community

So, if we change formatting from tab to 4 spaces, why can't we change localization formatting?

PR #22829, opened 2024-02-15
feat: automatic row count

### LINKED ISSUE #22657: automatic row count

utsav1704:
feat: #22657 
Automatic Row Count

I added the "Automatic Row Count" option to Data Editor Preferences.

image

If the Automatic Row Count option is enabled, then the number of rows is calculated only once when the data viewer opens. Once it is calculated and the data is being fetched batch-wise, there will be no execution of the row count. But if you click on the refresh button, then it will show "200+, 400+" based on batch size.

https://github.com/dbeaver/dbeaver/assets/75378586/5908310d-e041-4a07-9911-09eadaea2183


E1izabeth:
Thank you for contribution

E1izabeth (approved):

ShadelessFox (approved):

uslss:
verified

scscgit:
Note that after enabling the setting, the count keeps displaying 200+ on subsequent query runs until you close and reopen the SQL Editor. (DBeaver Version 25.0.1.202503231807)

PR #22842, opened 2024-02-16
dbeaver/pro#2491 Improve support of CASE statements

E1izabeth:
The query from the issue should be highlighted and you should be able perform jumping from aliases references to declarations inside query.


LonwoLonwo (approved):

uslss:
verified

PR #22865, opened 2024-02-19
dbeaver/dbeaver#22776 Idle connections: increase timeouts and disable…

### LINKED ISSUE #22776: CE Ver 23.3.4.  dropping/losing SQLite3 connection since upgrade
### Description

- SQLite3 DB accessed via DBeaver and Visual Studio C# Winforms GUI app on Win 10 Pro x64
- all table definitions done in DBeaver; GUI app accesses/updates data
- GUI app has a high level of dependence on the DB; DB required for app startup/operation
- successfully using config more than 2 1/2 yrs
- only change to dev environment preceding emergence of issue: DBeaver upgrade
- all connection Settings left at default since initial installation
- automatically opens connection to "last used" DB when DBeaver starts

The disconnection appears to be time based: if I leave DBeaver open but idle over lunch (approx. an hour) the disconnect occurs.

If the workspace active Tab is a Table the data listing is replaced with  a "Reconnect to Database" button.  Although that successfully reconnects to the DB it is mildly disruptive to my workflow.

If the active Tab is an SQL script the Reconnect button is not displayed.  Instead the Table columns all show the disconnected state.  Reconnecting by right-clicking on connection name appears to reconnect for the active Table tab but when a different Table tab is activated its columns show as disconnected.  The quickest way to solve the problem seems be closing/restarting DBeaver.

From the log file (attached): 
!ENTRY org.eclipse.ui 4 0 2024-02-10 15:52:33.172
!MESSAGE Unhandled event loop exception

### DBeaver Version

Community Edition Version 23.3.4.202402041841

### Operating System

Edition Windows 10 Pro Version 22H2 (Bld 19045.3803)

### Database and driver

SQLite 3.39.3
Driver sqlite_jdbc (no version provided)

BTW: your link for driver documentation for SQLite in the Edit Connection window landed at SQLite.org's PRAGMA documentation ...

### Steps to reproduce

1. Start Visual Studio 2022
2. Start DBeaver CE v 23.3.4 ; auto connect on startup to SQLite 3.39.3
3. click a couple Table tabs to ensure connectivity
4. leave idle in background for a few hours

### Additional context

_No response_

serge-rider:
… for embedded drivers

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/connection/DBPConnectionType.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/connection/DBPConnectionType.java
@@ -55,7 +55,7 @@ public class DBPConnectionType implements DBPDataSourcePermissionOwner {
             true,
             1800, //30 minutes
             true,
-            3600, //1 hour
+            14400, //1 hour
             true,
             null); //$NON-NLS-1$ //$NON-NLS-3$
         TEST = new DBPConnectionType(
@@ -71,7 +71,7 @@ public class DBPConnectionType implements DBPDataSourcePermissionOwner {
             true,
             900, //30 minutes
             true,
-            1800, //30 minutes
+            7200, //2 hours
             true,
             null); //$NON-NLS-1$ //$NON-NLS-3$
         PROD = new DBPConnectionType(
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/connection/DBPConnectionType.java
```diff
@@ -55,7 +55,7 @@ public class DBPConnectionType implements DBPDataSourcePermissionOwner {
             true,
             1800, //30 minutes
             true,
-            3600, //1 hour
+            14400, //1 hour
```
LonwoLonwo, line 58:
```suggestion
            14400, //4 hours
```

Matvey16:
Verified

PR #22872, opened 2024-02-20
dbeaver/pro#2494 Improve support for type casts

E1izabeth:
- [x] It typecast `::` is used, then highlighting and Ctrl+Click should work.
- [x] Outline should reflect data type from typecast
image

Please check when connection is active and not.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.lsm/grammar/SQLStandardLexer.g4
+++ b/plugins/org.jkiss.dbeaver.model.lsm/grammar/SQLStandardLexer.g4
@@ -328,6 +328,7 @@ ZONE: Z O N E ;
 
 
 // symbols
+TypeCast: '::';
 EqualsOperator: '=';
 NotEqualsOperator: '<>' | '!=';
 RightParen: ')';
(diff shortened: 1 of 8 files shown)


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/model/SQLQueryValueTypeCastExpression.java
```diff
@@ -0,0 +1,79 @@
...
+    public String getTypeRefString() {
+        return this.typeRefString;
+    }
+
+    @NotNull
+    public SQLQueryValueExpression getValueExpr() {
+        return this.value;
+    }
+
+    @NotNull
```
LonwoLonwo: Inconsistency here I see. Maybe nullable?

LonwoLonwo (approved):

plugins/org.jkiss.dbeaver.model.lsm/grammar/SQLStandardLexer.g4
```diff
@@ -328,6 +328,7 @@ ZONE: Z O N E ;
 
 
 // symbols
+TypeCast: '::';
```
ShadelessFox, line 331: It's not a part of the SQL standard. Is it OK that it's included here?

E1izabeth: yes, as we're making a common grammar, not a standard one and dialect-specifics

ShadelessFox: I see, okay, I thought the word "Standard" in the grammar was supposed to denote this grammar as SQL standard-compliant.

ShadelessFox (approved):

uslss:
verified

PR #22883, opened 2024-02-20
dbeaver/pro#2505 DBeaver Edition mismatch

serjiokov:
Fixed path of location for product configuration

### DIFF
--- a/plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
+++ b/plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
@@ -1817,9 +1817,13 @@ public class Main {
         } else {
             base = resolveLocation(productConfigurationLocation, DB_DATA_HOME, LOCATION_DATA_HOME_UNIX);
         }
-        String appVersion = getProductVersion();
         Path basePath = null;
-        if (appVersion != null && !appVersion.isEmpty()) {
+        String appId = getProductId();
+        String appVersion = getProductVersion();
+        if (appId != null &&
+            !appId.isEmpty() &&
+            appVersion != null &&
+            !appVersion.isEmpty()) {
             basePath = Paths.get(base, appVersion);
         } else {
             basePath = Paths.get(base);
@@ -1849,7 +1853,7 @@ public class Main {
         }
 
         File installDir = new File(installURL.getFile());
-        File eclipseProduct = new File(installDir, ".eclipseproduct"); //$NON-NLS-1$
+        File eclipseProduct = new File(installDir, PRODUCT_SITE_MARKER);
         if (eclipseProduct.exists()) {
             Properties props = new Properties();
             try (FileInputStream inStream = new FileInputStream(eclipseProduct)) {
@@ -1863,6 +1867,30 @@ public class Main {
             }
         }
         return appVersion;
+    }
+
+    private String getProductId() {
+        String appId = "";
+        URL installURL = getInstallLocation();
+        if (installURL == null) {
+            return null;
+        }
+
+        File installDir = new File(installURL.getFile());
+        File eclipseProduct = new File(installDir, PRODUCT_SITE_MARKER);
+        if (eclipseProduct.exists()) {
+            Properties props = new Properties();
+            try (FileInputStream inStream = new FileInputStream(eclipseProduct)) {
+                props.load(inStream);
+                appId = props.getProperty(PRODUCT_SITE_ID);
+                if (appId == null || appId.trim().length() == 0)
+                    appId = ""; //$NON-NLS-1$
+
+            } catch (IOException e) {
+                // nothing
+            }
+        }
+        return appId;
     }
 
     private void processConfiguration() {
--- a/plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
+++ b/plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
@@ -1824,7 +1824,7 @@ public class Main {
             !appId.isEmpty() &&
             appVersion != null &&
             !appVersion.isEmpty()) {
-            basePath = Paths.get(base, appVersion);
+            basePath = Paths.get(base, appId, appVersion);
         } else {
             basePath = Paths.get(base);
         }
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
```diff
@@ -1865,6 +1869,30 @@ private String getProductVersion() {
         return appVersion;
     }
 
+    private String getProductId() {
```
serge-rider: Copy-paste of getProductVersion
Use single function for product info read

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
serjiokov: Rework to one line

plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
```diff
@@ -1841,28 +1835,28 @@ private String resolveLocation(String source, String var, String location) {
...
+        File eclipseProduct = new File(installDir, PRODUCT_SITE_MARKER);
         if (eclipseProduct.exists()) {
             Properties props = new Properties();
             try (FileInputStream inStream = new FileInputStream(eclipseProduct)) {
                 props.load(inStream);
-                appVersion = props.getProperty(PRODUCT_SITE_VERSION);
-                if (appVersion == null || appVersion.trim().length() == 0)
-                    appVersion = ""; //$NON-NLS-1$
+                appProperty = props.getProperty(property);
+                if (appProperty == null || appProperty.trim().length() == 0)
```
serge-rider: I'd prefer to throw exception here. These are required properties.

Also we can read this file once, as it was done in original launcher in computeDefaultUserAreaLocation. No need to read the same file to read a single property.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
serjiokov: Reworked a bit

plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
```diff
@@ -1841,28 +1843,31 @@ private String resolveLocation(String source, String var, String location) {
...
-        File eclipseProduct = new File(installDir, ".eclipseproduct"); //$NON-NLS-1$
+        File eclipseProduct = new File(installDir, PRODUCT_SITE_MARKER);
         if (eclipseProduct.exists()) {
             Properties props = new Properties();
             try (FileInputStream inStream = new FileInputStream(eclipseProduct)) {
                 props.load(inStream);
-                appVersion = props.getProperty(PRODUCT_SITE_VERSION);
-                if (appVersion == null || appVersion.trim().length() == 0)
+                String appId = props.getProperty(PRODUCT_SITE_ID);
+                if (appId == null || appId.trim().length() == 0) {
```
ShadelessFox, line 1860:
```suggestion
                if (appId == null || appId.isBlank()) {
```

plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
```diff
@@ -1841,28 +1843,31 @@ private String resolveLocation(String source, String var, String location) {
...
+                    appId = ECLIPSE;
+                }
+                String appVersion = props.getProperty(PRODUCT_SITE_VERSION);
+                if (appVersion == null || appVersion.trim().length() == 0) {
```
ShadelessFox, line 1864:
```suggestion
                if (appVersion == null || appVersion.isBlank()) {
```

plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
```diff
@@ -1841,28 +1843,31 @@ private String resolveLocation(String source, String var, String location) {
...
                     appVersion = ""; //$NON-NLS-1$
-
-            } catch (IOException e) {
-                // nothing
+                }
+                productPath = appId + File.separator + appVersion;
```
ShadelessFox, line 1867: Please use types designed for handling with paths, such as `Path`. Rework this method to return `Path` instead of `String`. We're using Java 17, not Java 6.

plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
```diff
@@ -1841,28 +1843,31 @@ private String resolveLocation(String source, String var, String location) {
         return result.replaceFirst("^~", System.getProperty(PROP_USER_HOME));
     }
 
-    private String getProductVersion() {
-        String appVersion = "";
+    private String getProductProperties() throws IOException {
```
ShadelessFox, line 1846: It's not obvious that this method returns a path to product properties and not product properties themselves.

Also, add annotations, please.

plugins/org.jkiss.dbeaver.launcher/src/org/eclipse/equinox/launcher/Main.java
```diff
@@ -1810,22 +1810,24 @@ private URL getConfigurationLocation() {
...
-        } else {
-            basePath = Paths.get(base);
+        try {
+            String productPath = getProductProperties();
+            Path basePath = Paths.get(base, productPath);
+            productConfigurationLocation = basePath.toFile().getAbsolutePath();
+            url = buildURL(productConfigurationLocation, true);
+        } catch (IOException e) {
+            if (debug)
+                System.out.println("Can not read product properties. " + e.getMessage()); //$NON-NLS-1$
```
ShadelessFox, line 1828:
```suggestion
            if (debug) {
                System.out.println("Can not read product properties. " + e.getMessage()); //$NON-NLS-1$
            }
```

Matvey16:
Verified

PR #22888, opened 2024-02-20
#22832 Fix metadata edit permission check

### LINKED ISSUE #22832: Table creation restriction blocks data import
### Description

Data import can't be completed into existing table if structure edit id resricted

### DBeaver Version

23.3.5

### Operating System

Any

### Database and driver

Any

### Steps to reproduce

1. Open table type settings
2. Click edit permissions
3. Check only "restrict structure edit" box
4. Open import wizard for this connection
5. Ensure that table mapping is set to existing and no changes to table are needed
6. Start import

**Expected:** Data is imported
**Actual:** Error is shown, saying that table creation is restricted

### Additional context

dbeaver_okfGXt7zdT
dbeaver_xsz6vPHH3q

ShadelessFox:


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/database/DatabaseTransferUtils.java
```diff
@@ -547,6 +547,10 @@ private static void appendAttributeClause(DBPDataSource dataSource, StringBuilde
     }
 
     public static void executeDDL(DBCSession session, DBEPersistAction[] actions) throws DBCException {
+        if (actions.length == 0) {
+            return;
```
LonwoLonwo, line 551: Maybe this is a good place for logging? Looks like an unusual situation for this method.

ShadelessFox: It's not logging. If we don't have such permission, an exception will be thrown. Moreover, this method is only used for data transfer-related tasks, so this place is ideal for this check.

E1izabeth (approved):

LonwoLonwo (approved):

DariaMarkaryan:
Verified

PR #22892, opened 2024-02-21
dbeaver/pro#2359 remove experimental sql support

serjiokov:
Experimental features for SQL Editor removed:

1. AUTOCOMPLETION
2. HIGHLIGHTING

image


E1izabeth (changes requested):
We wanted to remove word 'experimental' from the setting name, not the setting itself 😳

PR #22896, opened 2024-02-21
dbeaver/dbeaver#22229 return to autocommit if just switched to manual mode

### LINKED ISSUE #22229: DBeaver switches to manual commit mode when Smart Commit is enabled and you execute an invalid query
### Description

When you use Smart Commit mode, and you execute a wrong query, like `where id = 123;` DBeaver switches to Manual Commit mode.

I am using smart commit here with autocommit:
image

After executing the query it switched to manual commit, which means I am now keeping a transaction open.
image

If I don't notice this I could potentially lock a database if I execute a correct select statement afterwards.

### DBeaver Version

Community Edition 23.3.0

### Operating System

Windows 11

### Database and driver

MySQL 8
com.mysql.cj.jdbc.Driver
8.0.29

### Steps to reproduce

1. Enable Autocommit and Smart commit
2. Excute an invalid query like 'where id = 123;'
3. DBeaver switches to Manual Commit

### Additional context

_No response_

Destrolaric:
Closes dbeaver/dbeaver#22229
Now if the failed statement is the same one from which change to manual mode happened we will return to auto commit mode

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
@@ -370,26 +370,28 @@ public class DBExecUtils {
         }
     }
 
-    public static void checkSmartAutoCommit(DBCSession session, String queryText) {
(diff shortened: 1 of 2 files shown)


DariaMarkaryan:
Verified

PR #22925, opened 2024-02-22
#17640 Override exposeModelRange to not always expand all collapsed nodes

### LINKED ISSUE #17640: Do not expand code groups after query execution
DBeaver 22.2.0
Database - ClickHouse 22.

I have long query with several CTE in beginning.
So I decide to collapse CTE text by grouping feature of text editor, something like this (comments is in russian, nevermind):
Снимок экрана 2022-09-14
But after executing query all this groups become expanded, and I again can't see whole query structure on one screen.

I think, it is not nessesary to open code group after query execution. Please, change behaviour of DBeaver and let me control state of this groups mannualy.

E1izabeth:
Issue should not happen anymore


LonwoLonwo (approved):

DariaMarkaryan:
Verified

PR #22937, opened 2024-02-23
dbeaver/pro#2524 Fix memory leak in the outline script node image

E1izabeth:
Please, test that on editor image change on editor disposal/closing/opening, outline image is correctly updated and no errors appear.


uslss:
verified

PR #22938, opened 2024-02-23
#22912 add catalog cache case sensitivity flag

### LINKED ISSUE #22912: Set as default (DB) doesn't work
### Description

Hello,
Switching the default SQL DB in connection when the name is "same" (case sensitivity) doesn't work for me.

I have 2 DB in the single connection '**customdatabase**' and '**CustomDatabase**' (don't ask why they have the same name, I did not create them).
And when I want to switch to 'customdatabase':
1. Switch between them in Open SQL console
2. Set as default in Databases list
3. Or setup default in Connections settings -> Initialization -> Default database
DBeaver switches to 'CustomDatabase' every time. I don't know why. 
Yesterday I updated DBeaver to the latest version, everything worked before.

### DBeaver Version

Community Edition Version 23.3.5.202402181738

### Operating System

macOS 13.4

### Database and driver

_No response_

### Steps to reproduce

1. Create two SQL databases in single connection and name them '**customdatabase**' and '**CustomDatabase**'.
2. Try to switch default between them.

### Additional context

I don't think it matters but I use ssh tunnel to MariaDB.

LonwoLonwo:
Like we already do for MySQL table cache

2024-02-23 14_32_02-Window


uslss:
verified

PR #22955, opened 2024-02-26
dbeaver/pro#2112 show only models with priority

LonwoLonwo:
aux PR


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceAuthModelDescriptor.java
```diff
@@ -42,6 +42,27 @@ public class DataSourceAuthModelDescriptor extends DataSourceBindingDescriptor i
 
     public static final String EXTENSION_ID = "org.jkiss.dbeaver.dataSourceAuth"; //$NON-NLS-1$
 
+    public enum DSPriority {
+        EXTRA_HIGH(1), // For necessary models
+        HIGH(2); // For priority models
```
E1izabeth: It looks strange. If high priority exists, there should also be normal priority somewhere

LonwoLonwo: And it is for the future. We can mark all our drivers with priority marks, but why do we need to do it without a special request?

E1izabeth: I just don't like the naming, that there are only high and extra high priorities in the enum

LonwoLonwo: Ok, what do you suggest?

E1izabeth: I would do it this way:
Low = 0, Normal = 1, Medium = 2, High = 3

E1izabeth: Or
Normal = 1, Medium = 2,  High = 3, Urgent = 4

ShadelessFox (commented):
This approach looks a bit weird. I think there should be other ways.

LonwoLonwo:
Approach changed.

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceProviderDescriptor.java
```diff
@@ -109,6 +110,7 @@ public DataSourceProviderDescriptor(DataSourceProviderRegistry registry, IConfig
             this.treeDescriptor = this.loadTreeInfo(trees[0]);
         }
         this.supportsDriverMigration = CommonUtils.toBoolean(config.getAttribute("supports-migration"));
+        this.skipParentsAuthModels = CommonUtils.toBoolean(config.getAttribute("skipParentsAuthModels"));
```
ShadelessFox: Please rename to `skipParentAuthModels`

serge-rider: I'd prefer `inheritAuthModels` 😎

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceProviderDescriptor.java
```diff
@@ -110,7 +110,7 @@ public DataSourceProviderDescriptor(DataSourceProviderRegistry registry, IConfig
             this.treeDescriptor = this.loadTreeInfo(trees[0]);
         }
         this.supportsDriverMigration = CommonUtils.toBoolean(config.getAttribute("supports-migration"));
-        this.skipParentsAuthModels = CommonUtils.toBoolean(config.getAttribute("skipParentsAuthModels"));
+        this.inheritAuthModels = CommonUtils.toBoolean(config.getAttribute("inheritAuthModels"));
```
ShadelessFox: I believe the new attribute name means the opposite thing - so it should be `true` by default.

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceProviderDescriptor.java
```diff
@@ -241,7 +241,7 @@ public DataSourceProviderDescriptor getParentProvider() {
     @Override
     public boolean matchesId(String id) {
         if (id.equals(this.id)) return true;
-        if (skipParentsAuthModels) {
+        if (inheritAuthModels) {
```
ShadelessFox: And here, it should be inverted

ShadelessFox (approved):

PR #22968, opened 2024-02-27
dbeaver/pro#2539 fix connection page NPE

LonwoLonwo:
Also, the default port was added on the main page for the Generic Instance. And the Instance field has been increased.

2024-02-27 14_33_08-Connect to a database


MashaKorax:
verified

PR #22969, opened 2024-02-27
dbeaver/pro#2264 Don't stop proposing after dot character

E1izabeth:
Please check proposals in other dialogs where they exist. Try writing dot there and see if it works okay.


uslss:
verified

PR #22970, opened 2024-02-27
dbeaver/pro#2171 clear preferences before adding new custom parameters

LonwoLonwo:
please check case from the ticket in EE and TE versions

with and without client restarting.

also please check datasource settings saving after client restart.


E1izabeth (approved):

uslss:
verified

PR #22974, opened 2024-02-27
dbeaver/pro#2395 [TE] Fix or remove Profile auth method

serjiokov:
Part-2 The profile can't be deleted


plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/AuthModelSelector.java
```diff
@@ -62,18 +62,20 @@ public class AuthModelSelector extends Composite {
     private final Runnable changeListener;
     private Combo authModelCombo;
     private boolean authSettingsEnabled = true;
+    private boolean isEnableSharedConfigurator = true;
```
serge-rider, line 65: no need to init here

serge-rider (changes requested):

PR #22977, opened 2024-02-27
dbeaver/pro#2540 Task execution and scheduling improvement

ShadelessFox:


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/task/TaskManagerImpl.java
```diff
@@ -302,8 +302,29 @@ public Path getStatisticsFolder(@NotNull DBTTask task) {
         return statisticsFolder.resolve(task.getId());
     }
 
+    @NotNull
+    @Override
+    public DBTTaskRunStatus runTask(@NotNull DBTTask task, @NotNull DBTTaskExecutionListener listener) throws DBException {
+        final TaskRunJob job = scheduleTask(task, listener);
```
serge-rider: Do we really need schedule+join? Can we just call run?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/task/TaskManagerImpl.java
ShadelessFox: This method is designed to be a synchronous alternative to the `scheduleTask` method; so yes, we need to `join()` here.

ShadelessFox: I get what you mean. Fixed in the following commit.

PR #22987, opened 2024-02-28
dbeaver/pro#2334 Highlighting, validation, outline for INSERT, UPDATE, DELETE

E1izabeth:
Now we support INSERT, UPDATE and DELETE in Outline

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorOutlinePage.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorOutlinePage.java
@@ -560,7 +560,11 @@ public class SQLEditorOutlinePage extends ContentOutlinePage implements IContent
      */
     private enum OutlineQueryNodeKind {
         DEFAULT,
-        NATURAL_JOIN_SUBROOT
+        NATURAL_JOIN_SUBROOT,
+        PROJECTION_SUBROOT,
+        DELETE_SUBROOT,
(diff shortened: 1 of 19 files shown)


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorOutlinePage.java
```diff
@@ -1003,6 +1010,113 @@ public Object visitSelectColumnSpec(@NotNull ColumnSpec columnSpec, @NotNull Out
             this.makeNode(arg, columnSpec, text, extraText, DBIcon.TREE_COLUMN, columnSpec.getValueExpression());
             return null;
         }
+        
+
+        @Nullable
+        @Override
+        public Object visitTableStatementDelete(@NotNull SQLQueryTableDeleteModel deleteStatement, @NotNull OutlineQueryNode node) {
+            String tableName = deleteStatement.getTableModel() == null ? "?" : deleteStatement.getTableModel().getName().toIdentifierString();
```
LonwoLonwo: Maybe we can add a constant for the "?" symbol? We have 3 in this PR and two more in the ui.editors.sql

E1izabeth: Yes, I wanted to do it in scope of refactoring ticket https://github.com/dbeaver/pro/issues/2529, but if you think that it's better to do now, okay then

uslss:
verified

PR #23010, opened 2024-02-29
#21748 Fix SRID being ignored when editing geometry data in MySQL

### LINKED ISSUE #21748: How to pick a point in the map? + Other issues using spatial data
Hi, I'm working in a project with geospatial data and I need to fill the database with some values, but I can't find a way to pick or edit a location using the map.

In fact, my experience is that the spatial data in MySQL in dbeaver are completely broken: (I can't view, edit or create properly)
1. The map doesn't show the correct location with the SRID I'm using (25831).
   - I noticed that if I manually change the SRID to 4326, then it shows the right location.
3. I can't manually edit the point text (see screenshot below), there's an error when I save.
4. I can't find a way to select a point in the map to edit the value. 

error messsage

I guess I'm doing something wrong. But the things I need are so simple that I'm surprised that I can't find solutions.

ShadelessFox:
Reproducible against MySQL 8.0+

```sql
CREATE TABLE `place` (
  `id` bigint UNSIGNED NOT NULL AUTO_INCREMENT,
  `location` POINT SRID 25831 NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `id` (`id`)
);

INSERT INTO `place` (`location`) VALUES (ST_GeomFromText('point(3.210907 41.972898)', 25831));
```

### DIFF
--- a/plugins/org.jkiss.dbeaver.data.gis/src/org/jkiss/dbeaver/model/gis/GisAttribute.java
+++ b/plugins/org.jkiss.dbeaver.data.gis/src/org/jkiss/dbeaver/model/gis/GisAttribute.java
@@ -16,7 +16,6 @@
  */
 package org.jkiss.dbeaver.model.gis;
 
-import org.jkiss.code.NotNull;
 import org.jkiss.code.Nullable;
 import org.jkiss.dbeaver.model.exec.DBCException;
 import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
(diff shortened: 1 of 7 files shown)


LonwoLonwo:
What about the getAttributeGeometryDimension method deleting? What was the reason?

ShadelessFox:
Not used.

E1izabeth (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #23032, opened 2024-03-01
CB-3582 fix package objects duplication

yagudin10:


LonwoLonwo (approved):

LonwoLonwo:
You also need to test DBeaver CE - both Oracle versions (to check the query syntax)

PR #23037, opened 2024-03-02
scripts to build DBeaver for all platforms

chirontt:
These build scripts make use of the `all-platforms` Maven profile to produce DBeaver build artifacts for all supported platforms, including the `aarch64` (i.e. `Arm64`) platforms for Linux and MacOS.

### DIFF
--- a/build_all_platforms.cmd
+++ b/build_all_platforms.cmd
@@ -1,0 +1,5 @@
+set MAVEN_OPTS=-Xmx2048m
+IF NOT EXIST ..\dbeaver-common git clone https://github.com/dbeaver/dbeaver-common.git ..\dbeaver-common
+cd product/aggregate
+call mvn clean install -T 1C -Pall-platforms
+cd ../..
--- a/build_all_platforms.sh
+++ b/build_all_platforms.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+[ ! -d ../dbeaver-common ] && git clone https://github.com/dbeaver/dbeaver-common.git ../dbeaver-common
+
+cd product/aggregate
+mvn clean install -T 1C -Pall-platforms
+cd ../..
+


LonwoLonwo:
Hello @chirontt 

Thanks for your contribution. Please add a description in the PR body.

chirontt:
description updated.

LonwoLonwo:
> These build scripts make use of the all-platforms Maven profile to produce DBeaver build artifacts for all supported platforms, including the aarch64 (i.e. Arm64) platforms for Linux and MacOS.

Please add more info about what issue you are trying to solve with this PR.

ShadelessFox (commented):
Perhaps this option can be used by default? I believe it shouldn't slow down the build process significantly. It might be better to change `build.cmd`/`build.sh` scripts instead.

chirontt:
> Perhaps this option can be used by default? I believe it shouldn't slow down the build process significantly. It might be better to change `build.cmd`/`build.sh` scripts instead.

Yes, it's better to incorporate the scripts in this PR into the existing build scripts, but I don't know why the `aarch64` platforms were left out of the existing scripts; I guess they were left out for a reason, so I leave them alone and devise new scripts to accommodate the `aarch64` builds in this PR, where I need them anyway.

I think, existing build scripts should default to build all supported platforms, regardless of how long they would take. Then, to enhance it further, there should be some parameter option to build only for specific platform, say Linux, or Windows, to save some build time. But this change won't come from me.

Please close this PR as you see fit, or I will close it myself, while we wait for better PR implementing the above.

ShadelessFox:
I'm going to discuss it with my colleagues first, so please keep it open for now.

ShadelessFox:
After further discussion, we decided to add this flag to our existing build scripts.

You can keep this PR open and just amend the changes.

chirontt:
If the existing build scripts will build the `aarch64` platforms (by default or otherwise), then this PR is not needed anymore; no point of keeping it around.

ShadelessFox:
You can submit a new PR that will modify existing scripts.

PR #23047, opened 2024-03-04
dbeaver/pro#2568 Reset user/password for shared credentials mode

serge-rider:


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceDescriptor.java
```diff
@@ -532,6 +532,11 @@ public void forgetSecrets() {
         this.secretsContainsDatabaseCreds = false;
         this.availableSharedCredentials = null;
         this.selectedSharedCredentials = null;
+        if (sharedCredentials) {
+            // For shared credentials reset cache also
+            connectionInfo.setUserName(null);
+            connectionInfo.setUserPassword(null);
```
alexander-skoblikov, line 538: looks like authProperties also must be cleared

PR #23052, opened 2024-03-05
dbeaver/pro#2565 Update syntax on metadata context change

E1izabeth:
1) Issue from the ticket should be solved.
2) If you open several scripts for one database and change schema only in one script, other scripts should not be updated. You can check it by opening script 1 with some select by entities existing only in one schema, and another script with any huge query. Then open Background tasks view and see that no new jobs appear there when you switch schema on the small script.


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -1502,10 +1502,10 @@ public static void fireObjectRemove(DBSObject object) {
         }
     }
 
-    public static void fireObjectSelect(DBSObject object, boolean select) {
+    public static void fireObjectSelect(DBSObject object, boolean select, DBCExecutionContext context) {
```
ShadelessFox: It would be nice to have annotations and javadoc here. It's a commonly used method, after all.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -1893,26 +1893,30 @@ public static DBSObject[] getSelectedObjects(@NotNull DBCExecutionContext contex
         return new DBSObject[0];
     }
 
-    public static void refreshContextDefaultsAndReflect(DBRProgressMonitor monitor, DBCExecutionContextDefaults contextDefaults) {
+    public static void refreshContextDefaultsAndReflect(DBRProgressMonitor monitor, DBCExecutionContextDefaults contextDefaults, DBCExecutionContext context) {
```
ShadelessFox: And here

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -1893,26 +1893,30 @@ public static DBSObject[] getSelectedObjects(@NotNull DBCExecutionContext contex
...
+                fireObjectSelectionChange(defaultCatalog, contextDefaults.getDefaultCatalog(), context);
+                fireObjectSelectionChange(defaultSchema, contextDefaults.getDefaultSchema(), context);
             }
         } catch (Exception e) {
             log.debug(e);
         }
     }
 
-    public static void fireObjectSelectionChange(DBSObject oldDefaultObject, DBSObject newDefaultObject) {
+    public static void fireObjectSelectionChange(
```
ShadelessFox, line 1922: And here

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -3140,7 +3146,23 @@ public void handleDataSourceEvent(final DBPEvent event) {
                             break;
                     }
                     updateExecutionContext(null);
-                    onDataSourceChange();
+                    
+                    boolean contextChanged = false;
+                    if (event.getAction().equals(DBPEvent.Action.OBJECT_SELECT)
+                        && event.getData() == this.getExecutionContext()
+                        && Boolean.TRUE.equals(event.getEnabled())
```
ShadelessFox: Can be just
```suggestion
                        && event.getEnabled()
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
E1izabeth: I don't think it's a good idea to add javadocs and annotation to the method I haven't created

LonwoLonwo (approved):

ShadelessFox (approved):

PR #23080, opened 2024-03-07
dbeaver/pro#2589 Project reference fix

ShadelessFox:
STR: https://github.com/dbeaver/pro/issues/2589

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/jobs/DataSourceMonitorJob.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/jobs/DataSourceMonitorJob.java
@@ -95,21 +95,21 @@ public class DataSourceMonitorJob extends AbstractJob {
 
     protected void doJob() {
         final DBPWorkspace workspace = platform.getWorkspace();
-        checkDataSourceAliveInWorkspace(workspace, () -> getLastUserActivityTime(lastPingTime));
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/jobs/DataSourceMonitorJob.java
```diff
@@ -95,21 +95,21 @@ protected IStatus run(DBRProgressMonitor monitor) {
...
     }
 
-    protected void checkDataSourceAliveInWorkspace(DBPWorkspace workspace, LongSupplier supplier) {
+    protected void checkDataSourceAliveInWorkspace(DBPWorkspace workspace, long lastUserActivityTime) {
         for (DBPProject project : workspace.getProjects()) {
             if (project.isOpen() && project.isRegistryLoaded()) {
                 DBPDataSourceRegistry dataSourceRegistry = project.getDataSourceRegistry();
                 for (DBPDataSourceContainer ds : dataSourceRegistry.getDataSources()) {
-                    checkDataSourceAlive(ds, supplier, workspace.getActiveProject().getWorkspaceSession());
+                    checkDataSourceAlive(ds, lastUserActivityTime);
```
ShadelessFox, line 106: Note for reviewers: here was an NPE caused by the return value of `workspace.getActiveProject()` being null. We can get a project from a data source container that we conveniently pass to the same function.

MashaKorax:
verified

PR #23122, opened 2024-03-13
CB-4568 drop unused internal tables

yagudin10:


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/schema/SQLSchemaScriptSource.java
```diff
@@ -38,4 +38,9 @@ public interface SQLSchemaScriptSource {
     @Nullable
     Reader openSchemaUpdateScript(DBRProgressMonitor monitor, int versionNumber) throws IOException, DBException;
 
+    /**
+     * Opens db specific scripts.
+     */
+    @Nullable
+    Reader openSpecificSchemaUpdateScript(DBRProgressMonitor monitor, int versionNumber, String specificPrefix);
```
LonwoLonwo: Please add NotNull/Nullable annotations to arguments.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/schema/SQLSchemaManager.java
```diff
@@ -153,6 +154,18 @@ private void upgradeSchemaVersion(
...
+    private void executeSpecificSchemaScript(
+        @NotNull DBRProgressMonitor monitor,
+        @NotNull Connection connection,
+        int updateToVer
+    ) throws SQLException, IOException, DBException {
+        Reader specificDdl = scriptSource.openSpecificSchemaUpdateScript(monitor, updateToVer, targetDatabaseDialect.getDialectId());
+        if (specificDdl != null) {
+            executeScript(monitor, connection, specificDdl, true);
+            specificDdl.close();
+        }
```
ShadelessFox: Manual `close()` is a no-no, use `try-with-resources` (it's null-safe).

ShadelessFox (approved):

PR #23130, opened 2024-03-14
CB-4528 Snowflake SSO auth

alexander-skoblikov:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/access/DBAAuthModelExternal.java
```diff
@@ -0,0 +1,27 @@
...
+
+package org.jkiss.dbeaver.model.access;
+
+import org.jkiss.dbeaver.model.connection.DBPConnectionConfiguration;
+
+/**
+ * External auth model.
+ */
+public interface DBAAuthModelExternal<CREDENTIALS extends DBAAuthCredentials> extends DBAAuthModel<CREDENTIALS> {
+    String getRequiredExternalAuth(DBPConnectionConfiguration configuration);
```
ShadelessFox: Nullability annotations, please 🙂

ShadelessFox (changes requested):

ShadelessFox (approved):

Issue #23150, opened 2024-03-15
"Export data" from footer wrongly uses pinned tab query 


2019-05-10:
### Description

Run a query, pin result tab (tab a).
Run another query, do _not_ pin result tab (tab b).

With tab b active, hit "Export data" in the footer below to export the results of that tab **b**.
DBeaver exports result of tab **a** instead.
With (multiple) pinned tabs always the first pinned tab to the left is used.

### DBeaver Version

Version 24.0.0.202403031740

### Operating System

Linux, Debian/SID

### Steps to reproduce

1. `SELECT * FROM a`
2. pin result tab (*tab a*)
3. `SELECT * FROM b`
4. do NOT pin result tab (*tab b*)
5. with *tab **b*** active hit `Export data` in the footer below to export results of *tab **b***
6. exported file contains results of *tab **a***!

ShadelessFox:
Thanks for the bug report

E1izabeth:
Can't reproduce the issue on the latest DBeaver version. So, probably it was fixed.
If it's still relevant, please let me know

PR #23151, opened 2024-03-17
#23150 Fix active result set obtaining

ShadelessFox:
Please check in single/multiple tab modes


ShadelessFox:
This doesn't work for single-tab presentation mode. Without passing additional data to the action handler, we can't determine the active result set.

PR #23156, opened 2024-03-18
CB-4867 remote preference store

alexander-skoblikov:


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/DesktopPlatform.java
```diff
@@ -100,18 +102,18 @@ static void setClosing(boolean closing) {
...
-        return DBeaverActivator.getInstance().getPreferences();
-    }
-
     public DesktopPlatform() {
         instance = this;
+        if (getApplication() instanceof DBPPreferenceStoreProvider preferenceStoreProvider) {
+            this.preferenceStore = preferenceStoreProvider.getPreferenceStore();
+        } else {
+            this.preferenceStore = DBeaverActivator.getInstance().getPreferences();
+        }
```
ShadelessFox: Let's have fewer `instanceof` checks everywhere, maybe? Let's define this method on the `DBPApplication` interface itself.

ShadelessFox (changes requested):

ShadelessFox (approved):

PR #23164, opened 2024-03-18
dbeaver/dbeaver#22666 Global configuration

### LINKED ISSUE #22666: Language preference should be locked whenever -nl is passed
### Description

This is an UX bug with the **User Interface** page of the **Preferences** window:

image

If `dbeaver.exe` was given the `-nl` option through the command line, then the **Language** field should be locked. An informative text label should pop right below it, explaining why so.

The thing is that the data store of this setting is ambiguous. It can come either...

1. from the file `%AppData%\DBeaverData\workspace6\.metadata\.plugins\org.eclipse.core.runtime\.settings\org.jkiss.dbeaver.core.prefs` ; or
1. from the command line, as the argument to the option `-nl` .

In the second case, the **Preferences** window would be hopeless trying to configure anything. By letting the user pick something in the **Language** field, we are selling the user an illusion that something is about to change. Then this expectation is bound to be frustrated.

The frustration manifests in two moments: immediate and long-term.

image

The immediate frustration arises right after clicking the **Apply** button. If `-nl` was given, the workbench is restarted in the very same language in that it previously was. That is because the argument is not cleared, internally, before reset:

https://github.com/dbeaver/dbeaver/blob/d83e5838dc4a5f6a50b891eaf730d0acd8923a56/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageDatabaseUserInterface.java#L306

The long-term frustration comes from the fact that the overriding data store was not touched at all so, when the user re-opens DBeaver the next day, they are getting the old setting once again:

image

### DBeaver Version

DBeaver Community Edition 23.3.0.202312031432

### Operating System

Windows 11 (21H2/10.0.22000.2652)

### Database and driver

Driver-independent.

### Steps to reproduce

1. Run `"%ProgramFiles%\DBeaver\dbeaver.exe" -nl en`
1. try to switch to any other language, from within the UI
1. when prompted to restart the workbench, press **Yes**
1. DBeaver restarts in English, once again, no matter what

### Additional context

1. https://github.com/dbeaver/dbeaver/issues/14694#issue-1064162326
1. https://github.com/dbeaver/dbeaver/issues/17857#issuecomment-1287051415
1. https://github.com/dbeaver/dbeaver/issues/19253#issue-1608940980
1. https://github.com/dbeaver/dbeaver/issues/19421#issue-1630991792
1. https://github.com/dbeaver/dbeaver/issues/20346#issue-1767294503
1. https://github.com/dbeaver/dbeaver/issues/20497#issue-1792045217
1. https://github.com/dbeaver/dbeaver/issues/20693#issue-1820764035
1. https://github.com/dbeaver/dbeaver/issues/20878#issuecomment-1719618325
1. https://github.com/dbeaver/dbeaver/issues/21002#issuecomment-1751120876

Also, this other proposal: *(with which I disagree, btw...)*

* https://github.com/dbeaver/dbeaver/issues/16187#issue-1204271562

ShadelessFox:
The configuration file resides in `DBeaverData/settings/global-settings.ini`.

List of currently supported properties:
|Name|Description|
|---|---|
|`language`|Language of the application. Has lower priority than `-nl`|


MashaKorax:
error on start 
`Can't open 'C:\Users\Maria Chubarova\AppData\Local\DBeaverEE\Chubarova\AppData\Local\DBeaverEE\dbeaver.exe': file doesn't exist`

icons in windows menu
explorer_n2kInwldQM

MashaKorax:
verified

PR #23169, opened 2024-03-19
dbeaver/pro#2570 support Romanian language by installation

LonwoLonwo:
The Romanian language is set alphabetically.


plugins/org.jkiss.dbeaver.registry/OSGI-INF/l10n/bundle.properties
```diff
@@ -27,7 +27,6 @@ language.it.label=Italian
 language.ja.label=Japanese
 language.uk.label=Ukrainian
 language.ko.label=Korean
-language.uk.label=Ukrainian
```
ShadelessFox, line 30: Why was it removed?

LonwoLonwo: Duplicate (check 28 string)

ShadelessFox: Nice catch.

PR #23208, opened 2024-03-22
dbeaver/pro#2663 Refactor SSH configuration model

ShadelessFox:


Matvey16:
Verified in Desktop DBeaver

PR #23234, opened 2024-03-26
#23079 Add support for WHERE TRUE, add simple types to outline

### LINKED ISSUE #23079: Support WHERE TRUE
As this syntax helps to comment / uncomment specific condition in the where clause I usually us a first condition which is always true and add a seperate line for each additional condition.

If I use `WHERE TRUE` as first condition the syntax highlighting partially stops working after this condition (`v.name` and `v.brand` ),

grafik

while syntax highlighting works fine if I use `WHERE 1=1` instead
grafik

I am using a MySQL database.

@shenzj1994: as you can see in the screenshot multiple conditons work fine for me (using MySQL at least). Maybe it's rater the `num::int<>0`?#

DBeaver Version
Community Edition Version 23.3.5.202402181738

Operating System
Windows 10 Enterprise

Database:
MySQL

_Originally posted by @TK-517 in https://github.com/dbeaver/dbeaver/issues/22887#issuecomment-1956433306_

E1izabeth:
- where true support was added
- now you can see boolean and constant in the outline as nodes


LonwoLonwo (approved):

uslss:
verified

PR #23242, opened 2024-03-26
dbeaver/pro#2390 Allow specifying case for column names in the header

ShadelessFox:
image


plugins/org.jkiss.dbeaver.data.transfer/OSGI-INF/l10n/bundle.properties
```diff
@@ -35,6 +35,8 @@ dataTransfer.processor.csv.property.header.name=Header
 dataTransfer.processor.csv.property.header.description=CSV header settings
 dataTransfer.processor.csv.property.headerFormat.name=Header format
 dataTransfer.processor.csv.property.headerFormat.description=Header format
+dataTransfer.processor.csv.property.headerCase.name = Header case
+dataTransfer.processor.csv.property.headerCase.description = You can choose lower or upper case for column names in the header
```
LonwoLonwo:
```suggestion
dataTransfer.processor.csv.property.headerCase.description = You can choose lower or upper case for column names or descriptions in the header
```

LonwoLonwo: Because we have dataTransfer.processor.csv.property.headerFormat

LonwoLonwo (approved):

uslss:
verified

PR #23245, opened 2024-03-26
#23072 SQLite column drop fix

### LINKED ISSUE #23072: SQLite: column delete fails
### Description

When we delete column in SQLite table then DDL script fails with error like
```
SQL error or missing database (table Invoice has 9 columns but 10 values were supplied)
```
Because of query `INSERT INTO Invoice SELECT * FROM temp`. 
Table `temp` contains more columns than new table (because we just deleted one of them).

We must list column names instead of `*`.

### DBeaver Version

24.0.0

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

ShadelessFox:
To reproduce the original issue:
1. Drop an existing FK
2. Create a new column
3. Persist changes

Please also test different mixes of column creation/deletion.


uslss:
verified

Issue #23135, opened 2024-03-14
Enhance the 'Can't create driver instance' error


flambchap:
**Is your feature request related to a problem? Please describe.**
When users try to establish a database connection, but there is no appropriate JDBC driver found, they get this error:

![image](https://github.com/dbeaver/dbeaver/assets/26031301/021fde72-9915-413e-b105-e68b4a9d3175)

If they click details, the text is not that helpful:

![Screenshot_20240314_184259](https://github.com/dbeaver/dbeaver/assets/26031301/e29fc8bc-5f25-483d-a1f8-a640c0213dc2)

It's especially frustrating in the case of databases like Netezza, for which we don't ship any drivers at all, and users need to provide them by hand. We can do better.

**Describe the solution you'd like**
We need to show a more user-friendly error. This error needs to be:

* Properly formatted
* Seen without clicking 'details'
* More accurate text.
  * If there is nothing in the driver manager's 'libraries' tab because we don't suggest any artifacts in the first place, we need to say something like, "We don't ship a Java driver ourselves. Please find it somewhere and add it manually".
  * If there is a file in the driver manager's 'libraries' tab that is not downloaded, suggest going to the driver manager and downloading the file.
  * If there is nothing in the driver manager's 'libraries' tab because the user deleted every artifact by hand, suggest resetting driver settings to defaults. Do the same if the user changed the class name.
  * If it's a custom driver, suggest checking the libraries tab and the class name.
* It must have a link that opens the driver manager and a link to the relevant documentation page.

**Additional context**
A similar issue propagates to CloudBeaver.

![Screenshot_20240314_184515](https://github.com/dbeaver/dbeaver/assets/26031301/1b0c93b9-00b3-42e0-96b5-878c42af8426)

![Screenshot_20240314_184529](https://github.com/dbeaver/dbeaver/assets/26031301/bca49926-e526-495b-ad8e-cf4c1737b589)

PR #23256, opened 2024-03-27
dbeaver/dbeaver#23135 Can't create driver instance' error

serjiokov:
Provided convenient error handling with a driver configuration option.


serjiokov:
image

PR #23273, opened 2024-03-28
#21654 subquery autocompletion

### LINKED ISSUE #21654: Subquery autocompletion does not work with experimental feature
### Description

dbeaver_9RpmBIC0Fx

### DBeaver Version

DBeaver 23.2.3

### Operating System

windows 10

### Database and driver

Oracle

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
image

It doesn't work well enough to test it or announce, but it works! We need more time to tune it. Also currently it doesn't look for the preferences about completion proposals formatting.

Here is a new settings - we can switch to use new engine or combine proposals from old and new engines. Combination doesn't remove duplicates for now.

image

For QA: please, test that old autocomplete works as it was before.

Further issues: dbeaver/pro#2718 dbeaver/pro#2719


Matvey16:
Verified

PR #23302, opened 2024-04-02
dbeaver/pro#2689 Don't rollback/disconnect during query execution

ShadelessFox:
This change affects just the "rollback idle transactions" option, but please also verify the "close idle connections" option works as before.


MashaKorax:
still reproduced

MashaKorax:
verified

PR #23304, opened 2024-04-02
dbeaver/dbeaver#23135 Driver init exception handle fix

### LINKED ISSUE #23135: Enhance the 'Can't create driver instance' error

serge-rider:


serjiokov:
image

serjiokov (approved):

uslss:
verified

PR #23319, opened 2024-04-04
dbeaver/pro#2705 Disable credentials save (by policy)

serge-rider:


plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/DatabaseNativeAuthModelConfigurator.java
```diff
@@ -56,6 +58,9 @@ public class DatabaseNativeAuthModelConfigurator implements IObjectPropertyConfi
 
     protected DBPDataSourceContainer dataSource;
 
+    public final boolean CREDENTIALS_SAVE_RESTRICTED = BasePolicyDataProvider.getInstance()
```
ShadelessFox:
```suggestion
    public static final boolean CREDENTIALS_SAVE_RESTRICTED = BasePolicyDataProvider.getInstance()
```

PR #23333, opened 2024-04-06
dbeaver/dbeaver#23332 warn users about possibility of having sensitive info in diagnostic archives

### LINKED ISSUE #23332: Warning about potentially sensitive data in diagnostic info archive
We need to warn users about the possibility of having sensitive information in diagnostic info archives extracted by feature #17003.

flambchap:
Screenshot 2024-04-10 at 12 40 54


ShadelessFox:
Won't it be irrelevant after dbeaver/pro#2666?

flambchap:
It'll still be relevant. If you don't mind, I'll explain it in person since you're a meter away from me right now.

Matvey16:
Verified

PR #23356, opened 2024-04-08
dbeaver/pro#2666 Mask sensitive information in the log

ShadelessFox:
A new VM flag was added to disable the masking:
```bash
-Ddbeaver.disable.secure.masking=true
```


E1izabeth (approved):

Matvey16:
Verified

PR #23358, opened 2024-04-08
dbeaver/pro#2729 SSH page settings load fix

ShadelessFox:
To reproduce:
1. Choose the "public key" authentication method and specify the path to a key
2. Save the connection
3. Delete the specified key
4. Open the connection, open the SSH page

Also, check:
- [ ] Changing the active profile updates the current configuration, including the "Enable XXX" checkbox next to the combo


MashaKorax:
verified

PR #23406, opened 2024-04-11
dbeaver/pro#2716 [ERD] Ortho layout should be reorganized

serjiokov:
New composition applied for graph
Ortho layout should be reorganized


uslss:
verified

PR #23441, opened 2024-04-16
#23359 JSCH session creation fix

### LINKED ISSUE #23359: DBeaver unable to re-establish secure tunnel
### Description

Since upgrading to DBeaver 24.0.1, after initial successful connection to DB via secure SSH tunnel, DBeaver disconnects SSH tunnel at some point, and is no longer able to invalidate/reconnect to DB via secure tunnel properly.

The following error occurs after invalidate/reconnect is attempted

SQL Error [08000]: Could not connect to address=(host=127.0.0.1)(port=23363)(type=master) : Socket fail to connect to host:127.0.0.1, port:23363. Connection refused

### DBeaver Version

CE 24.0.2

### Operating System

macOS 12.1

### Database and driver

JDBC

### Steps to reproduce

Connect to DB via SSH Tunnel
Allow tunnel to time out
attempt to reconnect 

### Additional context

using SSH tunnel

ShadelessFox:


E1izabeth (approved):

Matvey16:
This creates a similair bug for connections with jump server enabled. Was not present in devel branch. reproducebla with the same settings as in the attached issue but with addition of jump server

Matvey16:
Verified

PR #26274, opened 2024-04-22
#23105 Ensure stream column names are unique

### LINKED ISSUE #23105: Data importing problem from CSV files.
### Description

I think there is a problem with importing table data from CSV files(MS Excel).

### DBeaver Version

Version 24.0.0.202403031740 (Zip version)

### Operating System

Windows 11 Pro

### Database and driver

MARIADB 10.3

### Steps to reproduce

1. Create a new database table which has three columns ( ex: table_cat1, table_cat2, table_cat3) 
2. Try to import data from MS Excel CSV file.
3. At the "Tables mapping", you can see the data is mapped with table_cat1, table_cat2, table_cat2(not 3).
4. Even though I specified the columns one by one from source to target.

### Additional context
I tested it again, and I think at least 5 fields are needed to repaet this problem.
Always my fifth field data is linked to fourth field of target.
_No response_

ShadelessFox:
STR:
1. Create a CSV file with the first row containing duplicate values
2. Import as CSV and make sure the **header position** property is set to **top**

Example:
```csv
0,0,1,0,3
1,2,3,4,5
2,6,7,8,9
```


Matvey16:
Verified

PR #26423, opened 2024-04-23
dbeaver/pro#2706 Usage Statistics dialogue appears after double Workspace reset

serjiokov:
The order of operation is fixed, the current order is:
On startup:
1. Invoke reset (delete) preferences files
2. Invoke bundle preferences initialization


E1izabeth (approved):

ShadelessFox (approved):

ShadelessFox (commented):
Will these log messages be reverted before merging the PR?

Otherwise, please change the logging level to at least `trace` or `debug`

MashaKorax:
App cant start on Linux
[1713994616608.log](https://github.com/dbeaver/dbeaver/files/15101013/1713994616608.log)

MashaKorax:
verified

plugins/org.jkiss.dbeaver.ui.statistics/src/org/jkiss/dbeaver/ui/statistics/UIStatisticsActivator.java
```diff
@@ -110,6 +109,9 @@ public static ImageDescriptor getImageDescriptor(String path) {
     }
 
     public DBPPreferenceStore getPreferences() {
+        if (preferences == null) {
```
serge-rider: Why? It must be instantiated on bundle activation

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.statistics/src/org/jkiss/dbeaver/ui/statistics/UIStatisticsActivator.java
serjiokov: Generally, avoiding hanging bundle on start and case preferences reset (remove pref files) is better to create preferences at first call.

MashaKorax:
verified

PR #26441, opened 2024-04-23
dbeaver/pro#2752 Add a new setting to Disable (grey out) the Automatic updates check checkbox in Preferences

serjiokov:
Disable all automatic update operations


MashaKorax:
verified

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/policy/ApplicationPolicyService.java
```diff
@@ -67,4 +67,11 @@ private void disableCommand(@NotNull ICommandService commandService, @NotNull St
         }
     }
 
+    /**
+     * Return true, if software install/update policy enabled
+     */
+    public boolean isPolicySoftwareInstallUpdateEnabled() {
```
serge-rider: Naming hell.
- POLICY_SOFTWARE_INSTALL actually means "install disabled"
- isPolicySoftwareInstallUpdateEnabled actually means "install update disabled"

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/policy/ApplicationPolicyService.java
serjiokov: Get it, fixed to name isInstallUpdateDisabled()

PR #26450, opened 2024-04-23
dbeaver/pro#2744 Fix parallel transfer job cancellation

ShadelessFox:
It should now be possible to cancel the topmost job, and all other transfer jobs should be canceled as well.


serge-rider:
Please test with single and multiple tables export/import.

Matvey16:
It breaks the monitor of multiple tables transfer in one thread. Now it shows 0% until the end

Matvey16:
Verified

PR #26459, opened 2024-04-24
[CUBRID] Insert Color in Query plan's TableView

rathana-pvs:
- Change text color in cell to red when it is Full Scan
- Change text color in cell to green when it is Index Scan

_Noted: This PR has modified common source (shared modules). However, if this is not acceptable we will try to create a custom view for the Cubrid plugin._


LonwoLonwo:
Hello @rathana-pvs 

Thanks for your contribution.
We will look at it.

ShadelessFox (commented):
You should not bring UI dependencies into model bundles. Moreover, the navigator is now aware of the execution plan, leaking the incapsulation again.

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/plan/CubridPlanNode.java
```diff
@@ -99,6 +100,17 @@ public CubridPlanNode getParent() {
...
     }
+    
+    @Nullable
+    @Override
+    public Color getNodeColor() {
+        if (name.equals("iscan")) {
+            return new Color(0, 255, 0);
+        } else if (name.equals("sscan")) {
+            return new Color(255, 0, 0);
+        }
```
ShadelessFox, line 111: The `Color` class is backed by a native OS resource, so it must be disposed of after use.

Moreover, the colors chosen might (and will) look bad against the dark theme, so it's always better to make them customizable through the `org.eclipse.ui.themes` extension point whenever possible.

ShadelessFox (commented):
The specialized version of the `ObjectListControl` class is `org.jkiss.dbeaver.ui.editors.sql.plan.simple.PlanNodesTree`. You need to make your changes here.

Speaking of colors, it might be acceptable to make a mapping table from `DBCPlanNode#getNodeKind` to its respectful color. Not sure if the kind has the desired granularity, but the list of existing node kinds can be extended.

ShadelessFox:
Sorry, we won't accept this PR in its current state. Feel free to submit a new PR addressing all the issues mentioned.

PR #26614, opened 2024-04-24
CB-5055 trim lines in key data

yagudin10:


plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/SSHTunnelConfiguratorUI.java
```diff
@@ -758,7 +758,7 @@ public SSHHostConfiguration saveSettings() throws DBException {
                 case PUBLIC_KEY -> {
                     final String privateKey = privateKeyText.getText().trim();
                     if (DBWorkbench.isDistributed()) {
-                        yield new SSHAuthConfiguration.KeyData(privateKey, password, savePassword);
+                        yield new SSHAuthConfiguration.KeyData(SSHUtils.getKeyData(privateKey), password, savePassword);
```
ShadelessFox: I think it's better to do this inside `SSHUtils#saveHostConfiguration`

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -278,7 +278,8 @@ private static SSHHostConfiguration loadHostConfiguration(
                     if (validate && privKeyValue == null) {
                         throw new DBException(kind.formatErrorMessage("private key is not specified"));
                     }
-                    yield new SSHAuthConfiguration.KeyData(CommonUtils.notEmpty(privKeyValue), password, savePassword);
+                    String keyData = getKeyData(privKeyValue);
```
ShadelessFox: Please inline the variable

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -293,6 +294,11 @@ private static SSHHostConfiguration loadHostConfiguration(
         return new SSHHostConfiguration(username, hostname, port, auth);
     }
 
+    @NotNull
+    public static String getKeyData(String keyValue) {
```
ShadelessFox: It doesn't really "get" data from the input but rather sanitizes/normalizes the input.

You may also want to write a short javadoc for it if it's left public.

plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/SSHTunnelConfiguratorUI.java
yagudin10: we don't use `SSHUtils#saveHostConfiguration` in CloudBeaver

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
yagudin10: javadoc was added

plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/SSHTunnelConfiguratorUI.java
ShadelessFox: But we do. This particular class (`SSHTunnelConfiguratorUI`) is only used in the desktop product, hence it makes sense to perform sanitization where the data is actually serialized/deserialized.

yagudin10: Actually, we don't need to trim key data here - it will be trimmed on handler init.

ShadelessFox (approved):

PR #26616, opened 2024-04-24
dbeaver/pro#2784 Acquire the onboarding container in the "eclipse" way

ShadelessFox:
There should be no more errors in the log, and holiday decorations should work again (try changing the current date to Dec 20th)


MashaKorax:
verified

PR #26763, opened 2024-04-25
dbeaver/pro#2461 Add support for drop table statement

E1izabeth:
- Highlighting and new autocompletion works 
image

- Outline support works
image


uslss:
verified

PR #26449, opened 2024-04-23
dbeaver/pro#2594 Fix NPE in SpreadsheetPresentation

E1izabeth:
Can't reproduce

PR #26856, opened 2024-04-26
fix typo in SpreadsheetPresentation.java

angeousta:


angeousta:
needs to be merged before https://github.com/dbeaver/dbeaver/pull/26449

ShadelessFox:
Sorry, but we'd rather fix the typo in #26449 instead.

PR #26893, opened 2024-04-26
dbeaver/pro#2730 Connection wizard page validation fix

ShadelessFox:
Try creating or editing an existing connection and setting up an SSH tunnel without hostname/key - an error should be displayed and it should be impossible to save the wizard


Matvey16:
Proxy and SSL pages are always flagged as incomplete

Matvey16:
Verified

PR #29273, opened 2024-04-29
Fix potential github action smells

### LINKED ISSUE #29272: Fix potential github action smells
### Description

Hey! 🙂
I want to contribute the following changes to your workflow:

- Stop running workflows when there is a newer commit in PR
- Use commit hash instead of tags for action versions
- Use fixed version for runs-on argument
- Define permissions for workflows with external actions

These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html)

### DBeaver Version

24.0.3

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

ceddy4395:
Hey! 🙂
I want to contribute the following changes to your workflow:

- Stop running workflows when there is a newer commit in PR
- Use commit hash instead of tags for action versions
- Use fixed version for runs-on argument
- Define permissions for workflows with external actions

These changes are part of a research Study at TU Delft looking at GitHub Action Smells. [Find out more](https://ceddy4395.github.io/research/gha-smells.html)

closes #29272

### DIFF
--- a/.github/workflows/commit-msgs-validator.yml
+++ b/.github/workflows/commit-msgs-validator.yml
@@ -14,4 +14,6 @@ on:
 
 jobs:
   check-commit-message:
+    permissions: 
+      contents: write
     uses: dbeaver/dbeaver/.github/workflows/reused-commit-msgs-validator.yml@devel
--- a/.github/workflows/reused-commit-msgs-validator.yml
+++ b/.github/workflows/reused-commit-msgs-validator.yml
@@ -8,10 +8,12 @@ on:
 jobs:
   validate-commit-message:
     name: Validate Commit Message
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
+    permissions: 
+      contents: write
     steps:
       - name: Validate Commit Message Content
-        uses: gsactions/commit-message-checker@v1
+        uses: gsactions/commit-message-checker@b88ee88552f16f594ca19cb2c7fcd90df95c9bd4  # v1
         with:
           pattern: '^(#\d{3,6}|((Merge|dbeaver|cloudbeaver).+?#\d{3,6})|((DB-|CB-|web-)\d+)).*$'
           excludeDescription: 'true'


.github/workflows/commit-msgs-validator.yml
```diff
@@ -14,4 +14,6 @@ on:
 
 jobs:
   check-commit-message:
+    permissions: 
+      contents: write
```
ShadelessFox, line 18: Why? This job doesn't write anything

.github/workflows/reused-commit-msgs-validator.yml
```diff
@@ -8,10 +8,12 @@ on:
 jobs:
   validate-commit-message:
     name: Validate Commit Message
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
ShadelessFox, line 11: What's the point of specifying the exact runner machine? This job just validates commit messages; it doesn't build anything, so we don't care about what runs it.

.github/workflows/reused-commit-msgs-validator.yml
```diff
@@ -8,10 +8,12 @@ on:
...
+    permissions: 
+      contents: write
     steps:
       - name: Validate Commit Message Content
-        uses: gsactions/commit-message-checker@v1
+        uses: gsactions/commit-message-checker@b88ee88552f16f594ca19cb2c7fcd90df95c9bd4  # v1
```
ShadelessFox, line 16: Same question here. Why is it better than using a particular tag?

ShadelessFox:
I don't think these changes are worth adding.

.github/workflows/reused-commit-msgs-validator.yml
ceddy4395: It has to do with security, given that code attached to a tag can be altered by malicious users whereas commits cannot. [Here is a scientific paper](https://www.usenix.org/conference/usenixsecurity22/presentation/koishybayev) and [a blog post](https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash) about potential security risks related to GitHub Actions

ShadelessFox: This can be mitigated by limiting the set of permissions given to this job, and you have proposed to do the opposite.

Again, the whole purpose of this job is to validate commit messages so they match a particular pattern. Nothing fancy.

PR #29292, opened 2024-05-01
#29291 Improve page validation marks in multi-page wizards

### LINKED ISSUE #29291: Task editing wizard shows errors which are not actually present
### DBeaver Version

24.0.4

### Operating System

Any

### Database and driver

Any

### Steps to reproduce

1. Create an export/import task and save it
2. Open the task from tasks view

**Expected:** No error indication is present
**Actual:** Some of the steps are marked with red exclamation marks

### Additional context

dbeaver_K6ixqYbYwy

ShadelessFox:
Please test in the following scenarios:
1. Task wizard for a new task
2. Task wizard for an existing task
3. Mockdata wizard
4. Native tools wizard
5. Connection wizard

The incomplete mark (red cross) should only appear in the connection wizard. In other cases, just the green mark should be shown.


Matvey16:
Verified

PR #29830, opened 2024-05-06
dbeaver/pro#2803 Fix colors in navigator and ItemListControl

E1izabeth:
Was:
image
Become:
image

Was:
image

Become:
image


ShadelessFox (approved):

Matvey16:
Verified

PR #29935, opened 2024-05-07
dbeaver/dbeaver#28259 Portable DBeaver product distribution

### LINKED ISSUE #28259: before 24.0 it was portable
Hi,

we're using Dbeaver in a virtual maschine, bcoz of this we need to have a portable version (the network is closed we just put Dbeaver folder) before 24.0 version it was so and the life was easy. after version 24.0 Office etc. addins saving in folder users/dbeaverdata/instal-data or something like this.. it makes harder using the Dbeaver portable. Please could you do like before this feature.

Thank you
(sorry for my English)

serjiokov:
Fix for identify install location and configuration location

### DIFF
--- a/plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/DBeaverLauncher.java
+++ b/plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/DBeaverLauncher.java
@@ -49,7 +49,7 @@ public class DBeaverLauncher {
     /**
      * Indicates whether this instance is running in debug mode.
      */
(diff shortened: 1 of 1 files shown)


serjiokov:
Require to test:
case 1: If the product is installed in a public location (where the access rights is enough to create files/folder), the configuration will be in the same folder as the installation (unpacked)
case 2: if the product is installed in a system location C:/Program Files/...  or other restricted  user area, the configuration will be in a common location 
     ~/user/APP_DATA - WinOS
     ~/Library - MacOS
     ~/.local/share - Unix

DariaMarkaryan:
Verified

PR #29994, opened 2024-05-08
dbeaver/pro#2824 Custom scope performance fix

ShadelessFox:
Try selecting several objects in Oracle DB1. It would take several minutes to find these objects against the `devel` branch.


MashaKorax:
verified

PR #30356, opened 2024-05-10
CB-4889. Fixed deleted newly firestore column

DenisSinelnikov:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBSDocumentLocator.java
```diff
@@ -37,7 +37,12 @@ public interface DBSDocumentLocator extends DBSDocumentContainer
      * @throws DBException on any DB error
      */
     @Nullable
-    DBDDocument findDocument(@NotNull DBCSession session, Map<String, Object> key, Map<String, Object> metaData) throws DBException;
+    DBDDocument findDocument(
+        @NotNull DBCSession session,
+        Map<String, Object> key,
+        Map<String, Object> metaData,
+        @Nullable DBSDocumentLocator locator
```
alexander-skoblikov: use this?

alexander-skoblikov (approved):

PR #32368, opened 2024-05-13
CB-931 ldap identity provider

alexander-skoblikov:


ShadelessFox (dismissed):
These icons are completely black and won't be visible on the dark theme. Please change their color to use our accent color.

PR #33040, opened 2024-05-16
Insert Color in Query plan's treeView

### LINKED ISSUE #23407: [CUBRID] Insert Color in Query plan's TableView
**Problem**
There is no color in query plan's Tableview. 

**Solution**
- Change text color in cell to **red** when it is **Full Scan**
- Change text color in cell to **green** when it is **Index Scan**

hwany7seo:
- Change color in cell to red when it is Full Scan (Table Scan)
- Change color in cell to green when it is Index Scan

Issue : https://github.com/dbeaver/dbeaver/issues/23407
PR : https://github.com/dbeaver/dbeaver/pull/26459

We proceeded by referring to the above issues and PR.
image
image

And I wanted to display oracle as well, but the kind information seems to be different. 
(I don't know if it's different for each version, but 'foregin' is not displayed until versions 9 to 11.)

I think need to add "table access" to allowedkind in OraclePlanNode 
and distinguish between 'FULL SCAN' and 'NOT_FULL_TABLESCAN'.

	case "table access":
                    if (options.toLowerCase().equals("full")) {
                        return DBCPlanNodeKind.TABLE_SCAN;
                    }
                    return DBCPlanNodeKind.NOT_FULL_TABLE_SCAN;

If do this, it will display normally as shown below.
image
image

If possible, I'd like to include this in the PR as well.


E1izabeth (approved):

ShadelessFox (approved):

E1izabeth:
Your proposal sounds good. Yes, it might work different on different versions.
Let's add this for Oracle in a separate PR.

MashaKorax:
verified

serge-rider:
There are some merge conflicts. But GH doesn't say which exactly because it is fork PR.

hwany7seo:
The conflict has been resolved.

Issue #18644, opened 2023-01-03
The PG query array data log appears  Can't parse array


travelliu:
### Description

PostgreSQL

```
PostgreSQL 13.4 (Debian 13.4-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
```

Query pg_proc

```
select * from pg_proc where proname='to_timestamp';
```

log 

```
2023-01-03 21:28:47.778 - Can't parse array
2023-01-03 21:28:47.781 - Can't parse array
```
![2023-01-03 at 21 29](https://user-images.githubusercontent.com/7792329/210368138-8f3ed632-9a47-4705-a69f-15f3c7aca02c.png)

General problem: There is an error in the conversion array data

![2023-01-03 at 21 42](https://user-images.githubusercontent.com/7792329/210368828-86722823-732e-44ab-b85b-aaca7f8f95ee.png)

### DBeaver Version

22.3.1

### Operating System

MacOS 10.12

### Database and driver

1. PostgreSQL 13.4
2. PostgreSQL JDBC Driver [42.2.20]

E1izabeth:
Thank you for bug report.

Issue #17467, opened 2022-08-25
PostgreSQL - _box (box[]) isn't formatted as an array correctly.


Bas950:
### Description

Box arrays aren't getting parsed correctly
![image](https://user-images.githubusercontent.com/34792838/186643633-454fff7f-c751-43a9-a21f-d8167ab428d4.png)

Should be like any other arrays:
![image](https://user-images.githubusercontent.com/34792838/186643798-107c5807-86ca-4d8b-89df-c1120ac902fa.png)

### DBeaver Version

22.1.3.202207240614

### Operating System

Windows 11 build: 22000.856

### Database and driver

Database: PostgreSQL 13
Driver: PostgreSQL JDBC Driver 42.2.25

### Steps to reproduce

1. Connect to a database
2. Open a new sql script
3. run `select '{(1\,2)\, (3\,4); (1\,2)\,(3\,4)}'::box[]`

### Additional context

https://www.postgresql.org/docs/current/arrays.html#ARRAYS-INPUT
![image](https://user-images.githubusercontent.com/34792838/186644918-e4780f25-3d34-4454-a7cf-2f879461ab8f.png)
I am guessing since box has a different delimiter character DBeaver's formatter is messing up.

uslss:
thanks for the report

Issue #17031, opened 2022-07-08
`ArrayIndexOutOfBoundsException` thrown when expanding a list of maps in the data editor


WEGFan:
#### System information: 

- Operating system (distribution) and version: Windows 10 20H2 (19042.1151)
- DBeaver version: 22.1.1.202206241354
- Additional extensions: None

#### Connection specification:

- Database name and version: MongoDB v5.0.9
- Driver name: MongoDB
- Do you use tunnels or proxies (SSH, SOCKS, etc)? No

#### Describe the problem you're observing:

`ArrayIndexOutOfBoundsException` thrown when expanding a list of maps in the data editor

#### Steps to reproduce, if exist:

1. Insert sample data:

    ```js
    db.test.insert({
        "a": "1",
        "b": [
            {
                "d": 1,
                "e": 2
            },
            {
                "d": 11,
                "e": 22
            },
            {
                "d": 111,
                "e": 222
            }
        ]
    });
    ```

2. When expanding column `b.d` and `b.e`, an `ArrayIndexOutOfBoundsException` was thrown.

    ![](https://user-images.githubusercontent.com/7558201/178012035-1f490d43-9f8a-4323-aa37-ca503bc0ac0b.png)

    It should be shown like this:

    <table>
        <thead>
            <tr>
                <th rowspan="2">_id</th>
                <th rowspan="2">a</th>
                <th colspan="2">b</th>
            </tr>
            <tr>
                <th>d</th>
                <th>e</th>
            </tr>
        </thead>
        <tbody>
            <tr>
                <td>62c83b8c3ff93e3da2f976b4</td>
                <td>1</td>
                <td>1</td>
                <td>2</td>
            </tr>
            <tr>
                <td>62c83b8c3ff93e3da2f976b4</td>
                <td>1</td>
                <td>11</td>
                <td>22</td>
            </tr>
            <tr>
                <td>62c83b8c3ff93e3da2f976b4</td>
                <td>1</td>
                <td>111</td>
                <td>222</td>
            </tr>
        </tbody>
    </table>

#### Include any warning/errors/backtraces from the logs

```
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at org.jkiss.dbeaver.model.DBUtils.getAttributeValue(DBUtils.java:812)
	at org.jkiss.dbeaver.ui.controls.resultset.ResultSetModel.getCellValue(ResultSetModel.java:400)
[...]

PR #33113, opened 2024-05-16
dbeaver/pro#2574 Support nested rows in grid

ShadelessFox:
image

Nested values don't support editing.


E1izabeth (commented):
I suppose, at least these cases should be tested to check that we don't break something

- [ ] #17467
- [ ] #17248
- [ ] #17031
- [ ] #18644
- [ ] #17016
- [ ] #15429
- [ ] #14594
- [ ] #14327

E1izabeth (approved):

Matvey16:
Now bottom level objects in arrays have expand button that doesn't do anything. Is it expected?

https://github.com/dbeaver/dbeaver/assets/82543000/90bf6f99-622b-486b-9782-504710037d1f

Matvey16:
The expand button disappeared from the grid cells in Dynamo, now it is shown only near row numbers
image

ShadelessFox:
Let's fix them in a separate ticket.

Matvey16:
I hope we can fix them sooner than later because if feel like it's a regression
https://github.com/dbeaver/pro/issues/2871

PR #33193, opened 2024-05-17
dbeaver/dbeaver#22020 Fix "before connect" not terminate after failing connect

### LINKED ISSUE #22020: Shell сommand "Before Connect" not terminates then connection failed
### Description

I expect, if connection failed then `Connection settings -> Shell Commands -> Before Connect` shell script should be automatically terminated.

But unfortunately the script is still running, and there is no way to interrupt it.

### DBeaver Version

Community Edition 23.2.5

### Operating System

macOS 13.3.1

### Database and driver

PostgreSQL - any

### Steps to reproduce

1. Create posgrtess connection with default configuration
image

2. Edit that connection and go to `Connection settings -> Shell Commands -> Before Connect` and set and shell cmomand, which should work until terminated, for instance `ping 8.8.8.8` and set `Terminate process at disconnect` to enabled.
image

3. Try to connect. Connection error should be shown, but script still running
image

### Additional context

Also unable to terminate processes via "Background tasks" panel

image

HocKu7:


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceDescriptor.java
```diff
@@ -1296,6 +1297,18 @@ private boolean connect0(DBRProgressMonitor monitor, boolean initialize, boolean
         }
     }
 
+    private void terminateChildProcess() {
```
ShadelessFox:
```suggestion
    private void terminateChildProcesses() {
```

E1izabeth: Agreed

E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #33259, opened 2024-05-20
#29999 Fix race condition when loading navigator nodes

### LINKED ISSUE #29999: DBeaver 24.0.0+ disconnect doesn't always work
### Description

See the Step 11 of "Steps to Reproduce" section for the description of the bug. I don't know how to describe it in isolation of the steps to reproduce.

### DBeaver Version

Community Edition 24.0.4

### Operating System

Windows 11 22H2

### Database and driver

Database Server: MySQL 5.7.44-google-log
Database Driver: MySQL Connector/J mysql-connector-j-8.4.0 (Revision: 1c3f5c149e0bfe31c7fbeb24e2d260cd890972c4)

### Steps to reproduce

1. Start DBeaver.
2. Make sure no Editors opened (middle part of the screen).
3. Connect to a Database server from the Projects view.
4. Double-click on any TABLE in a DB on the connected server, so the TABLE View (Properties tab) will be opened in the middle of the screen. This is the state you should be in:
image
5. Disconnect from the Database in the Project View using context menu. The Middle of the screen should still have an opened TAB with the table, but it says "Disconnected" with "Reconnect to the database" in the middle of the screen. 
 image
 image

6. Close DBeaver.
7. Start DBeaver.
8. The Middle of the screen should have an opened TAB with the table (Properties tab) in a Connected state.
image
9. Find the DB Server in the Projects View again (by expanding <Group_name> -> Connections -> <Folder_name> -> <DB_Connection_name>. The UI shows a green checkbox on the DB Connection, that means it is connected.
image
10. Right click on the DB Connection name -> Disconnect. It is the same step as step 5.
 image
11. **Here is the bug:** The DB Connection name in the UI still has a green checkbox, like it is still connected. The TABLE tab in the middle of the screen also still has active Properties tab. If you try to right-click on the DB Connection name again, the Context Menu will show "Connect" action as available. "Disconnect" action will be grayed out.
 image
 From this broken state there are plenty of incorrect behavior happens. E.g. trying to open some other table from this connection says "Default context not found". Trying to connect again shows the same error message. Basically the connection is broken and weird at this point and the only way to fix it is to restart DBeaver. "Invalidate/Reconnect" also doesn't help. Sometimes a "Data read" error window pops up with "Can't read data - no active connection" message.
  Trying to refresh the opened table gives this error message: 
image
  Trying to refresh the "Projects" tree removes the green checkbox from the UI, but if you try to connect again, the green checkbox will still be missing. And if you try to open any other table from this connection, a "Default context not found" error will shown again. Trying to access "Data" tab of the table shows this error: 
image
Basically, without DBeaver restarting it is impossible to correctly Disconnect. Every time I need to Disconnect some connection I'm forced to restart DBeaver.

### Additional context

This bug never happened before I upgraded to version 24.0.
I tried to install 23.3.5 now and I can't reproduce this bug in this version. So, probably some changes in 24.0.0 caused this bug.

ShadelessFox:


E1izabeth (approved):

Destrolaric (approved):

Matvey16:
Verified

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNResource.java
```diff
@@ -216,7 +216,11 @@ public boolean allowsChildren() {
     @Override
     public DBNNode[] getChildren(DBRProgressMonitor monitor) throws DBException {
         if (children == null) {
-            this.children = readChildNodes(monitor);
+            synchronized (this) {
```
serge-rider: I don't think that we can use sync here. It may lead to deadlocks because readChildNodes may interact with UI.

serge-rider (changes requested):

Matvey16:
Verified

Issue #23314, opened 2024-04-03
Add Possibilty to Hide Folders


lbassuncao:
It would be nice to add the possibility to hide folders unrelated to projects like ".git", ".vscode", ".idea", etc.

Maybe this could be implemented using a ".hidden" file like in Linux. Which would contain all the files and folders to hide from the file viewer, and of course the file ".hidden" is hidden itself by default.

Something like this: https://delightlylinux.wordpress.com/2015/06/03/hide-files-with-hidden/

Thanks.

LonwoLonwo:
Hello @lbassuncao 

Could you please provide a screen of how it looks in your case right now?

lbassuncao:
Hi @LonwoLonwo 

Of Course. Here it is:
![DBeaverCE](https://github.com/dbeaver/dbeaver/assets/8853276/b0afdd25-b8d2-4500-b98b-41896bb515a8)

Thanks.

lbassuncao:
Hi @LonwoLonwo 

Of course.

Here it is:

![DBeaverCE](https://github.com/dbeaver/dbeaver/assets/8853276/de75b5e3-5db1-4646-a620-5aa2777e0972)

Thanks.

LonwoLonwo:
Ok, thanks for the idea!

arhayka:
We discussed this ticket with the team and decided to do the following:
1/ Add the new setting in Preferences (User interface —> Navigator) and in the Navigator toolbar 
2/ This setting will allow users to show or hide hidden files: Windows hidden files, dotfiles, .hidden, gitignore

SASUKE38:
I'm interested in working on this issue but am having some trouble figuring out where/in which file to place the logic for hiding the files and folders. Specifically, I'm unsure where the logic that loads project files and folders is. Does anyone have any advice for this?

Issue #20759, opened 2023-08-01
Killling saved connections


altermodus:
### Description

All my database connections list is lost!

... after replace `,` to `DEFAULT NULL,`  in sql script editor.

### DBeaver Version

23.1.3

### Operating System

Ubuntu 22.10

### Database and driver

Any driver (tested on MariaDB, PostgreSQL)

### Steps to reproduce

1) Create and save any connection on **Database tab**
2) Open database and open **Script editor** for opened database and print `,`
3) Press **Ctrl+H**, press **"Replace ..."**
4) Replace `,` to `DEFAULT NULL,` and press `OK`

and i get an exception:

```
com.google.gson.stream.MalformedJsonException: Unterminated object at line 3 column 18 path $..Mydb

Reason:
com.google.gson.stream.MalformedJsonException: Unterminated object at line 3 column 18 path $..Mydb
```

Restart Dbeaver and I get and exception:

```
com.google.gson.stream.MalformedJsonException: Unterminated object at line 2 column 17 path $.

Reason:
com.google.gson.stream.MalformedJsonException: Unterminated object at line 2 column 17 path $.
```

E1izabeth:
Thank you for bug report.

We had something similar some time ago, but couldn't reproduce #20065

I can reproduce it now.

E1izabeth:
This suggestion should help to eliminate the error message and restore your connections  https://github.com/dbeaver/dbeaver/issues/20065#issuecomment-1560624522

altermodus:
I found backup json files in 
```
.local/share/DBeaverData/workspace6/General/.dbeaver/.data-sources.json.bak
.local/share/DBeaverData/workspace6/General/.dbeaver/.credetials-config.json.bak
.local/share/DBeaverData/workspace6/General/.dbeaver/.project-metadata.json.bak
```
and restore (rename) to
```
.local/share/DBeaverData/workspace6/General/.dbeaver/data-sources.json
.local/share/DBeaverData/workspace6/General/.dbeaver/credetials-config.json
.local/share/DBeaverData/workspace6/General/.dbeaver/project-metadata.json
```
but in **data-sources.json** I was found
```
    "folders": {
        "Folder1": {}default null DEFAULT NULL DEFAULT NULL,
        "Folder2": {}default null DEFAULT NULL DEFAULT NULL,
        "Folder3": {}default null DEFAULT NULL DEFAULT NULL,
[...]

PR #33265, opened 2024-05-21
dbeaver/dbeaver#23314 Hide unrelated folders and files in projects

SASUKE38:
Added a method to read an optional file in the top level directory of a project that contains names of files and folders to avoid displaying.


ShadelessFox:
Thank you for your contribution!

Unfortunately, we're not happy with the proposed solution:
1. No caching is done, so the `.hidden` file is read every time a node has its child nodes loaded
2. Just DBNResource nodes are affected; filtering nodes won't prevent them from being discoverable via other facilities or the Find/Replace dialog (might cause something like in #20759)
3. The `.hidden` mechanism is Linux-specific and only applies to graphical environments; it's not standardized and is cumbersome to get right. The general approach is to hide all `.*` files on Linux. Meanwhile, there's no such mechanism on Windows, and files are hidden by changing their attributes.

You can read more about the proposed changes here: https://github.com/dbeaver/dbeaver/issues/23314#issuecomment-2125060418

PR #33271, opened 2024-05-21
dbeaver/pro#2674 [TE] Shared credentials button is not visible on small screens

serjiokov:
Scrolled composite added


serjiokov:
https://github.com/dbeaver/dbeaver/assets/32031248/ac1d5910-e9be-4640-b755-63ed1acb9927

E1izabeth:
Why these "Save credentials" checkbox and "Manage" button are so far from Authentication method combo?

ShadelessFox:
Perhaps because of the "Save credentials" checkbox. I wouldn't say I like it either, but fixing it wouldn't be that trivial.

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageSettings.java
```diff
@@ -425,9 +422,13 @@ private CTabItem createPageTab(@NotNull IDialogPage page, int index) {
         item.setToolTipText(page.getDescription());
 
         if (page.getControl() == null) {
-            final Composite placeholder = new Composite(tabFolder, SWT.NONE);
+            final ScrolledComposite sc = new ScrolledComposite(tabFolder, SWT.BORDER | SWT.H_SCROLL);
+            sc.setExpandHorizontal(true);
+            sc.setExpandVertical(true);
+            final Composite placeholder = new Composite(sc, SWT.NONE);
```
ShadelessFox: I don't think you need a placeholder here. Put controls directly inside your `ScrolledComposite`

serjiokov: Just for use 'setContent' - require to specify scrollable content

E1izabeth (approved):

ShadelessFox (approved):

PR #33402, opened 2024-05-22
dbeaver/dbeaver#23349 [CrateDB] ER diagrams can't be opened

### LINKED ISSUE #23349: [CrateDB] ER diagrams can't be opened
### Description

Opening ER diagrams for any object lead to an SQL Error because the postgres function used doesn't work in CrateDB

### DBeaver Version

24.0.2

### Operating System

Any

### Database and driver

CrateDB 5.3.3 (built a0a52be/2023-06-26T19:53:14Z, Linux 5.4.0-122-generic amd64, OpenJDK 64-Bit Server VM 19.0.2+7)

PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

1. Open a table
2. Click on ER Diagram tab

**Expected:** Diagram is opened
**Actual:** Error appears

### Additional context

_No response_

serjiokov:


E1izabeth:
Please give some explanations of your PR - we need to know why we can't show ER diagrams in CrateDB in the future. Also I wonder how does it look like (where do we see this error message)?

LonwoLonwo:
I think the problem here is the inability to read table references. You need to hide foreign keys or find the correct way to read them.

serjiokov:
Not, exactly true, not needed to fix data base if it has implementation issue. The idea behind, don't to throw any exceptions if you already have enough details. Instead we have to log non critical exception and proceed visialisation.

E1izabeth:
So, do ER diagrams work now? Please add description to this PR.

serjiokov:
Yes it is working for crateDB
image

E1izabeth (approved):

Matvey16:
Verified

PR #33960, opened 2024-05-22
dbeaver/pro#2861 Add block bound strings for db2

E1izabeth:
If you get create procedure statement from the txt file in the issue, DBeaver will handle full declaration until end. You can check it in query manager or by foldings 
image


uslss:
verified

PR #34052, opened 2024-05-23
CB-4906. Fixed ConcurrentModificationException in WebDataSourceMonito…

DenisSinelnikov:
…rJob

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/jobs/DataSourceMonitorJob.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/jobs/DataSourceMonitorJob.java
@@ -96,10 +96,12 @@ public class DataSourceMonitorJob extends AbstractJob {
     }
 
     protected void checkDataSourceAliveInWorkspace(DBPWorkspace workspace, long lastUserActivityTime) {
-        for (DBPProject project : workspace.getProjects()) {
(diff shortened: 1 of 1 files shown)


alexander-skoblikov (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/jobs/DataSourceMonitorJob.java
```diff
@@ -96,10 +96,12 @@ protected void doJob() {
     }
 
     protected void checkDataSourceAliveInWorkspace(DBPWorkspace workspace, long lastUserActivityTime) {
-        for (DBPProject project : workspace.getProjects()) {
+        List<DBPProject> projects = new ArrayList<>(workspace.getProjects());
```
E1izabeth, line 99: As far as I understood, another thread may be making changes to the array element while you are copying it. This could lead to an IndexOutOfBoundsException, uninitialized array values or even some sort of memory access exception, as System.arraycopy() is implemented in native code.

So, this may not help, depending on the concrete situation

alexander-skoblikov: in very extreme cases this can happen, then we need to add a lot of locks to workspace (there are none there now, and by some miracle everything has been working for years)

PR #34059, opened 2024-05-23
dbeaver/dbeaver#26613 Fix json format for binary column exporting

### LINKED ISSUE #26613: Binary columns cause JSON export to be invalid
### Description

A table that contains a binary column causes the JSON generated by the "Export data" feature to be invalid.

For example,

```
{
"select * from t1": [
	{
		"id" : 0x1234
	},
	{
		"id" : 0x5678
	}
]}
```

The export menu offers to choose an "Encoding" for "Binaries", but none of the options produces valid JSON:

* Native: `"id" : 0x1234`
* Binary: `"id" : 4` (including a non-printable character before the "4"!)
* Base64: `"id" : EjQ=`
* Hex: `"id" : 0x1234` (same as Native)

### DBeaver Version

Community Edition 24.0.3.202404222221

### Operating System

Linux

### Database and driver

Database: MySQL (Percona) v8.1.0
Driver: MySQL (I don't see any version number)

### Steps to reproduce

Create a simple table, insert some binary data, select it:

```
create table t1 (
	id varbinary(10)
);

insert into t1 values (0x1234),(0x5678);

select * from t1;
```

Click "Export data", choose "JSON / Export to JSON file(s)"

### Additional context

_No response_

HocKu7:


E1izabeth (approved):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/IStreamDataExporterSite.java
```diff
@@ -53,6 +53,8 @@ public interface IStreamDataExporterSite {
 
     void writeBinaryData(@NotNull DBDContentStorage cs) throws IOException;
 
+    void writeBinaryData(@NotNull DBDContentStorage cs, String typeOfOutput) throws IOException;
```
ShadelessFox: Annotations 🙁

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamTransferConsumer.java
```diff
@@ -1061,23 +1074,37 @@ public void writeBinaryData(@NotNull DBDContentStorage cs) throws IOException {
...
+                                    binaryString = CommonUtils.getQuotedString(binaryString);
+                                }
                                 writer.write(binaryString);
                                 break;
                             }
                         }
                         default: {
                             // Binary stream
                             try (Reader reader = new InputStreamReader(stream, cs.getCharset())) {
+                                String string = IOUtils.readToString(reader);
```
ShadelessFox: This is a big no. It might be a really big blob (like several GBs in size) that we would otherwise just pipe to the output stream (writer).

ShadelessFox (commented):
I don't like the general approach. Let's discuss it tomorrow.

DariaMarkaryan:
verified

PR #34068, opened 2024-05-24
Cb 4793 schema name column

DenisSinelnikov:


plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericExecutionContext.java
```diff
@@ -387,4 +398,12 @@ public GenericObjectContainer getDefaultObject() {
         return null;
     }
 
+    @NotNull
+    @Override
+    public DBCCachedContextDefaults getCachedDefault() {
```
serge-rider, line 417: You cannot call getDefaultSchema or getDefaultCachedCatalog here. 
you cannot interact with dynamic catalog or schema.
Use just selectedEntityName

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/qm/meta/QMMConnectionInfo.java
```diff
@@ -237,12 +237,22 @@ public QMMStatementExecuteInfo beginExecution(DBCStatement statement)
                 isTransactional() && getTransaction() != null ?
                     getTransaction().getCurrentSavepoint() : null;
             var sqlDialect = statement.getSession().getDataSource().getSQLDialect();
+            String schema = null;
+            String catalog = null;
+            DBCExecutionContextDefaults contextDefaults = statement.getSession().getExecutionContext().getContextDefaults();
+            if (contextDefaults != null) {
+                schema = contextDefaults.getCachedDefault().schemaName();
```
serge-rider: Put getCachedDefault to var.

serge-rider (changes requested):

PR #34119, opened 2024-05-24
dbeaver/dbeaver#33618 Change "Delete" to "Delete Connection"

### LINKED ISSUE #33618: Change "Delete" to "Delete Connection" in connection dropdown menu
### Discussed in https://github.com/orgs/dbeaver/discussions/33393

<sup>Originally posted by **troogyt** May 22, 2024</sup>
In the screenshot below, does "**Delete**" mean "_Delete the database_", or does it mean "_Delete the connection to the database, but leave the database in tact_"?
image

If my guess is correct, and it means "_Delete only the connection to the database but do not delete the database_", then can we please change the menu to read as follows?
image

I am not that smart, so when I read this I get worried that I'm going to delete the actual database.
Making this minor change would remove this stress and make it much clearer (at least, for us dummies :))</div>

HocKu7:


E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorHandlerObjectDelete.java
```diff
@@ -209,21 +211,21 @@ private static void changeDependentObjectsList(@NotNull DBRProgressMonitor monit
...
-//            if (selection instanceof IStructuredSelection && ((IStructuredSelection) selection).size() > 1) {
-//                element.setText(UINavigatorMessages.actions_navigator_delete_objects);
-//            } else {
-//                DBNNode node = NavigatorUtils.getSelectedNode(selection);
-//                if (node != null) {
-//                    element.setText(UINavigatorMessages.actions_navigator_delete_ + " " + node.getNodeTypeLabel()  + " '" + node.getNodeName() + "'");
-//                }
-//            }
-//        }
+        if (!updateUI) {
```
serge-rider, line 214: It is still not clear why we commented it last time..

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorHandlerObjectDelete.java
```diff
@@ -209,21 +211,21 @@ private static void changeDependentObjectsList(@NotNull DBRProgressMonitor monit
...
+        if (selectionProvider != null) {
+            ISelection selection = selectionProvider.getSelection();
+            if (selection instanceof IStructuredSelection) {
+                List<DBNNode> nodes = NavigatorUtils.getSelectedNodes(selection);
+                for (DBNNode node : nodes) {
+                    if (!(node instanceof DBNDataSource)) {
+                        return;
+                    }
+                }
+                element.setText(UINavigatorMessages.actions_navigator_delete_ + " " + nodes.get(0).getNodeTypeLabel());
```
serge-rider: Are we sure that we want to update label only for connections?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorHandlerObjectDelete.java
HocKu7: I guess so. The reason to write it is only for connection, because it could be a bit confusing what exactly do you want to delete? A database or connection? But it's clear enough if you try to delete for example a table. You don't need extra explonation for it as well as we don't need to overload our UI

HocKu7:
We decided to write a type for any object. So it should be like "Delete Connection", "Delete Table", etc.

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorHandlerObjectDelete.java
```diff
@@ -209,21 +211,20 @@ private static void changeDependentObjectsList(@NotNull DBRProgressMonitor monit
...
-//            }
-//        }
+        if (!updateUI) {
+            return;
+        }
+        final ISelectionProvider selectionProvider = UIUtils.getSelectionProvider(element.getServiceLocator());
+        if (selectionProvider != null) {
+            ISelection selection = selectionProvider.getSelection();
+
+            if (selection instanceof IStructuredSelection && ((IStructuredSelection) selection).size() == 1) {
```
ShadelessFox:
```suggestion
            if (selection instanceof IStructuredSelection ss && ss.size() == 1) {
```

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorHandlerObjectDelete.java
```diff
@@ -209,21 +211,20 @@ private static void changeDependentObjectsList(@NotNull DBRProgressMonitor monit
...
+                DBNNode node = NavigatorUtils.getSelectedNode(selection);
+                if (node != null) {
+                    element.setText(String.format(
+                        "%s %s", UINavigatorMessages.actions_navigator_delete_, node.getNodeTypeLabel()));
```
ShadelessFox: I kind of dislike the concatenation here. In some languages, the order is reversed (IIRC in Chinese, you say "Table Delete" instead of "Delete Table"). Let's replace it with an NLS pattern.

ShadelessFox (changes requested):

ShadelessFox (approved):

E1izabeth (approved):

Matvey16:
Verified

PR #17870, opened 2022-10-10
#17619 Add ` to escaping for mysql

E1izabeth:

Issue #17619, opened 2022-09-12
Grave accent isn't escaped during table creation


uslss:
### Description

![image](https://user-images.githubusercontent.com/31996417/189613669-6db4893a-ba99-4e91-add1-8d10737cd552.png)
Grave accent isn't escaped during table creation

### DBeaver Version

CE 22.2.0

### Operating System

win10

### Database and driver

MySQL 5.7.25-28

PR #34167, opened 2024-05-26
#34060 Backticks are exported as double backticks when exporting resaults from query as SQL

### LINKED ISSUE #34060: Backticks are exported as double backticks when exporting resaults from query as SQL
### Description

When trying to export the data that came out from a query result single backticks are converted to double.

Here is the data I got from the query:

Screenshot 2024-05-23 at 4 47 14 PM

And here is how it's generated after I click Export to SQL INSERT statements:

Screenshot 2024-05-23 at 4 51 57 PM

### DBeaver Version

Version 24.0.4.202405051837

### Operating System

_No response_

### Database and driver

MySQL 8

### Steps to reproduce

Just select any column that has a backtick (`) and try to export it as SQL statement and the result will be double backticks (``)

### Additional context

_No response_

diashalabi:


plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialect.java
```diff
@@ -270,7 +270,7 @@ protected String escapeString(@NotNull String string, @Nullable String[] quotes)
         if (quotes != null) {
             string = string.replace(quotes[0], quotes[0] + quotes[0]);
         } else {
-            string = string.replace("'", "''").replace("`", "``");
+            return super.escapeString(string);
```
E1izabeth, line 273: I'm sure this was done for some reason and I don't like the idea of just removing it.

diashalabi: I agree, it might have been put for a reason and from git blame I can see that you added it? 😅 

Either way the reason I thought it should be removed it that I checked Sqlite and Postgresql don't use this escaping so it might have been added for a reason but I believe this wasn't the correct way of fixing it so maybe remove it now and let me know the reason and I will try to fix the other problem without having to compromise the export feature

E1izabeth: Here is the initial issue about these escaping #17619 and the following fix was done #17870
I checked that your fix doesn't broke that issue.

So, maybe we can rely on my investigation and merge this PR.

E1izabeth: Also the build of this PR passes our unit test `MySQLDialectTest::quoteStringTest`

E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #34180, opened 2024-05-28
Databricks Identifier Test

SASUKE38:
Test for Databricks identifiers.


test/org.jkiss.dbeaver.ext.databricks.test/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,20 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: DBeaver MySQL Tests
```
LonwoLonwo:
```suggestion
Bundle-Name: DBeaver Databricks Tests
```

serge-rider:
The test looks a bit pointless. Extra bundle for a test which verifies identifiers for a databricks is too much.

PR #34183, opened 2024-05-28
dbeaver/dbeaver#33618 Revert label update

### LINKED ISSUE #33618: Change "Delete" to "Delete Connection" in connection dropdown menu

serge-rider:


Matvey16:
Verified

PR #34211, opened 2024-05-30
dbeaver/pro#2646 Fix Redis key filters

HocKu7:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -221,6 +231,23 @@ public static String makeSQLLike(String like)
...
+
+        StringBuilder result = new StringBuilder();
+        for (int i = 0; i < sqlLikePattern.length(); i++) {
+            char charAtI = sqlLikePattern.charAt(i);
+            if (i + 1 < sqlLikePattern.length() && charAtI == '\\'
+                && (sqlLikePattern.charAt(i + 1) == '_' || sqlLikePattern.charAt(i + 1) == '%')) {
+                result.append(sqlLikePatternToGlobSymbolsMap.get(
+                    String.format("%c%c", charAtI, sqlLikePattern.charAt(i + 1))));
+                i++;
+            }
```
ShadelessFox: This condition only checks for `\\_` and `\\%` sequences. Is this expected? If so, I'd rather remove the map lookup.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -221,6 +231,23 @@ public static String makeSQLLike(String like)
...
+            result.append(sqlLikePatternToGlobSymbolsMap.computeIfAbsent(String.valueOf(charAtI), Function.identity()));
```
ShadelessFox: ... and here we are only checking for `_`, `%`, and `?`, really. Since you're calling `computeIfAbsent` to avoid introducing a local variable, you will poison the map with all possible characters, wasting memory for nothing. You either want to use `Map#getOrDefault`, or check explicitly for these two cases with an `if` statement.

ShadelessFox: You don't need a lookup map for such a simple function. Please refactor it.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
HocKu7: not only. In the map also is a 3 another values

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
HocKu7: Totally agree about computeIfAbsent. There is no need to push new elements into map, thank you about that. But getting rid of map leads to overcomplication of code with multiply if-statement. I don't think that it is something that we want

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
ShadelessFox: No, this particular piece of code will only result in either of these two because your key is two characters long while others are one character long.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
serge-rider: I agree that we don't need map for such a simple algorithm. static switch/case is much better

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -221,6 +221,21 @@ public static String makeSQLLike(String like)
...
+
+        StringBuilder result = new StringBuilder();
+        for (int i = 0; i < sqlLikePattern.length(); i++) {
+            char charAtI = sqlLikePattern.charAt(i);
+            switch (charAtI) {
+                case '_' -> result.append('?');
+                case '%' -> result.append('*');
+                case '?' -> result.append("\\?");
+                default -> result.append(charAtI);
+            }
```
ShadelessFox: You also need to escape the `*` character

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -221,6 +221,21 @@ public static String makeSQLLike(String like)
         return like.replace("*", "%").replace("?", "_");
     }
 
+    public static String makeGlobFromSqlLikePattern(String sqlLikePattern) {
```
ShadelessFox: Maybe `makeGlobFromLike`? Also, please specify nullability annotations. They are essential in our codebase.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
HocKu7: No, there is no need to do it

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
HocKu7: I guess to have a more specific function name is more readable for such big namespace of the class. And second one, why do we need it if IDEA provide default annotation from the box 
image

Maybe we don't need it anymore?

ShadelessFox: Not everyone uses IntelliJ as their preferred IDE. Moreover, it's possible to generate compile-time checks for these which is impossible without explicit annotations.

ShadelessFox: The class itself is called `SQLUtils`, suggesting that its methods are somehow related to SQL. If you look at other methods here, you will notice that they don't contain the term "SQL" in their names.

Therefore, it should be called `makeGlobFromLike` to match other methods.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
ShadelessFox: Yes, there is. Since both `?` and `*` are special markers when in the glob, you want to escape all occurrences of these in the original string. You escape `?` but not `*`, why?

HocKu7: image
That's why

ShadelessFox: I see. I'd consider it a bug. Can you please add a comment somewhere in the method about this?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
HocKu7: Actually, I see a number of methods with SQL in their name in the class. Also such util method is avaliable in almost entire codebase of our, so it's better to have a name exhaustive name for such method. Moreover, you are always able to use static import, what brings confusing to the name without SQLUtils.

ShadelessFox: Static imports are forbidden in our codebase.

PR #34242, opened 2024-06-03
#34239 Fix node loading from non-default project

### LINKED ISSUE #34239: Objects from non-default project open with error after restart
### Description

If any object from a connection in non-default project is still open whe DBeaver is closed it would show an error after restart

### DBeaver Version

24.1.0

### Operating System

Win11

### Database and driver

Any

### Steps to reproduce

1. Open a table from a connection in non-active project
2. Close DBeaver
3. Start DBeaver

**Expected:** Table is loaded
**Actual:** Table tab shows an error

### Additional context

image

ShadelessFox:


Matvey16:
Verified

PR #34247, opened 2024-06-04
dbeaver/dbeaver#26613 Fix json binary

### LINKED ISSUE #26613: Binary columns cause JSON export to be invalid

HocKu7:


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamTransferConsumer.java
```diff
@@ -1065,14 +1066,22 @@ public void writeBinaryData(@NotNull DBDContentStorage cs) throws IOException {
...
-                            // Binary stream
-                            try (Reader reader = new InputStreamReader(stream, cs.getCharset())) {
-                                IOUtils.copyText(reader, writer);
+                            byte[] readBuffer = new byte[1000];
+                            for (; ; ) {
+                                int count = stream.read(readBuffer);
+                                if (count <= 0) {
+                                    break;
+                                }
+                                String content = new String(readBuffer, 0, count, settings.getOutputEncoding());
```
ShadelessFox: You should be using `cs.getCharset()` instead.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamTransferConsumer.java
```diff
@@ -1065,14 +1066,22 @@ public void writeBinaryData(@NotNull DBDContentStorage cs) throws IOException {
                                 writer.write(binaryString);
                                 break;
                             }
+                            break;
```
ShadelessFox, line 1073: Remove the `break` above then

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/json/JSONUtils.java
```diff
@@ -104,7 +104,13 @@ public static String escapeJsonString(String str) {
                 case '\f' -> result.append("\\f");
                 case '\b' -> result.append("\\b");
                 case '"', '\\', '/' -> result.append("\\").append(c);
-                default -> result.append(c);
+                default -> {
+                    if ((int) c < 32) {
```
ShadelessFox, line 108:
```suggestion
                    if (c < ' ') {
                        // escape non-displayable ASCII characters
```

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamTransferConsumer.java
HocKu7: I guess it was a mistake to use `cs.getCharset()` before. As I see there is need to take charset from params, cause we have such properties in configuration page

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/json/JSONUtils.java
HocKu7: It has not so clear intention. I prefer to leave it

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamTransferConsumer.java
ShadelessFox: But this is not correct. You are reading data from `cs` and should use its charset, not the output charset. The output charset is already used by the underlying stream.

ShadelessFox (approved):

MashaKorax:
verified

PR #34252, opened 2024-06-04
dbeaver/dbeaver#29299 Securely save credentials

### LINKED ISSUE #29299: Stop using DES and avoid hardcoded key
## Why you need it?
We've identified that in the file SecuredPasswordEncrypter.java, the encryption key is hardcoded, and vulnerable algorithms DES and 3DES are still in use [here](https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/encode/SecuredPasswordEncrypter.java#L37). 
```java
    private static final byte[] PASSWORD_ENCRYPTION_KEY = "sdf@!#$verf^wv%6Fwe%$$#FFGwfsdefwfe135s$^H)dg".getBytes(Charset.defaultCharset());

    public static final String SCHEME_DESEDE = "DESede";
    public static final String SCHEME_DES = "DES";
```
## How could it be?
The hardcoded encryption key may be exposed to the attacker, and the password storage will be insecure.
The DES and 3DES have long been regarded as vulnerable. Applying these two algorithms for password storage may not be considered as secure.

### DBeaver Version

24.0.3

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

ShadelessFox:
**Please save credentials in the previous version FIRST and then switch to this branch to ensure that backward compatibility is present.**

Check against:
1. PostgreSQL native tools, such as backup/restore/etc. They allow specifying custom credentials when saved as a task
2. Proxy for downloading drivers (Preferences -> Connections -> Drivers)
3. Maven authentication (Preferences -> Connections -> Drivers -> Maven)

UPDATE:
1. Please test against **community edition** and **any other PRO edition**
2. Not sure what else this could break. I have a suspicion that it might somehow have an effect on CloudBeaver, but I'm not sure.


MashaKorax:
Dbeaver CE asks about master pass hint when you set a password in drivers
STR

1. Delete secrete storage
2. Go to drivers and set credentials (on video)

https://github.com/dbeaver/dbeaver/assets/84867187/be9ba408-11ee-48d1-a09f-a9df45165367

MashaKorax:
STR:
1)Open Ce app from devel
2)Go to drivers page in preferences and save creds
3)Install CE app from PR
4)Open drivers page in preferences  and observe that creds are dissapeared

MashaKorax:
verified

PR #34255, opened 2024-06-04
dbeaver/dbeaver#23203 smart metadata read

### LINKED ISSUE #23203: Option to disable all metadata queries
---------------------------------------------------------
--TEST 1: when executing SQL queries in Snowflake, I see 2 extra queries done by dbeaver on same session:
image

--I found 2 settings to turn off suggested in:
[](https://github.com/dbeaver/dbeaver/issues/22637)

> preferences > editors > data editor >
    [ ] Read table metadata (unique keys)
    [ ] Read table references (foreign keys)

But that only removes one of the rogue metadata queries from being executed:
image

---------------------------------------------------------
--TEST 2 - another test, simple disconnect/connect and execute 3 USE + query to grab all queries executed 
image

I executed 4 queries but Snowflake says DBEaver executed 11 queries in total:
my 4 queries:
`
use role sysadmin;
use database db;
use schema s;
`

the 7 extra metadata queries that seems DBeaver enjoying doing on its own :(
`
--1st dbeaver query:
SELECT DISTINCT
  CASE TABLE_TYPE
    WHEN 'BASE TABLE' THEN 'TABLE' // For backward compatibility. 
    ELSE TABLE_TYPE
  END AS TABLE_TYPE
FROM INFORMATION_SCHEMA.TABLES;

--2nd dbeaver query:
SELECT CURRENT_DATABASE(), CURRENT_SCHEMA();

--3rd dbeaver query:
SELECT CURRENT_DATABASE(), CURRENT_SCHEMA();

--4th dbeaver query:
show /* JDBC:DatabaseMetaData.getCatalogs() */ databases in account

--5th dbeaver query:
show /* JDBC:DatabaseMetaData.getSchemas() */ schemas in database "DB"

--6th dbeaver query:
USE DATABASE DB;

--7th dbeaver query:
SELECT CURRENT_DATABASE(), CURRENT_SCHEMA();
`

Is there any options to disable/prevent  DBEAVER from executing ANY queries on its own ?
If I execute 4 queries, I only want to see 4 queries executed in snowflake.

Thanks ~

serge-rider:


Matvey16:
I think there is unexpected behaviour here. If the option to not read metedata is on filtering by value works incorrectly. 
By clicking `filter by value` in column header you get a pop up that reading from server may take a long time. But in the opened window showed values would always be from the loaded data and not from server even if you click on checkbox to `read from server`

Matvey16:
Verified

PR #34258, opened 2024-06-05
dbeaver/dbeaver#20668 - Can't save ERD diagram

### LINKED ISSUE #20668: Can't save ERD diagram
### Description

Saving an ERD diagram based on a MySQL database results in the error message "Cant't determine diagram file". It is possible though to save the diagram into a new file.

### DBeaver Version

23.1.2

### Operating System

Ubuntu 22.04

### Database and driver

MySQL

### Steps to reproduce

Create a ERD diagram based on some tables. Add some virtual foreign key constraints. Then try to save the file.

### Additional context

_No response_

serjiokov:
Save procedure implementation for external diagram resource.


serjiokov:
How to test:
1) Create a diagram and save outside of the workspace using the "Save as..." from menu.
2) Modify elements ( just a change position of entities) 
3) Try to save.

plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/editor/ERDEditorStandalone.java
```diff
@@ -125,17 +126,25 @@ public void doSave(IProgressMonitor monitor)
...
+            if (file != null) {
+                file.setContents(
+                    new ByteArrayInputStream(diagramState.getBytes(StandardCharsets.UTF_8)),
+                    true,
+                    true,
+                    monitor);
+                getCommandStack().markSaveLocation();
+            } else {
+                // file is out of workspace
+                if (editorInput instanceof IURIEditorInput uriInput) {
```
ShadelessFox: This can be merged into `} else if {`

plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/editor/ERDEditorStandalone.java
```diff
@@ -125,17 +126,25 @@ public void doSave(IProgressMonitor monitor)
...
-                throw new DBException("Can't determine diagram file");
+            IEditorInput editorInput = getEditorInput();
+            final IFile file = EditorUtils.getFileFromInput(editorInput);
+            if (file != null) {
+                file.setContents(
+                    new ByteArrayInputStream(diagramState.getBytes(StandardCharsets.UTF_8)),
+                    true,
+                    true,
+                    monitor);
+                getCommandStack().markSaveLocation();
```
ShadelessFox: This can be moved after the `if` chain to avoid duplicating code

ShadelessFox (commented):
Small code style issues

ShadelessFox:
Kinda sucks that we don't have a utility method for writing a file that will do all this checks. Bet we have multiple other issues like this one.

plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/editor/ERDEditorStandalone.java
serjiokov: Generally, this block can not be after if because in the first case a file is IFile under EFS, in the second case File - java.io 
both have different API for write content.

ShadelessFox: I'm talking solely about the `getCommandStack().markSaveLocation();` statement. It's present in both cases, and the third one throws an exception (hence unreachable), so it can be safely moved below

serjiokov: ok, get it.

E1izabeth (approved):

ShadelessFox (approved):

MashaKorax:
verified

PR #34270, opened 2024-06-05
dbeaver/pro#2878 - DBeaver leaks org.eclipse.swt.graphics.Resource

serjiokov:
instances

A specific flag was added for resource leak detection.


ShadelessFox (commented):
I thought we had discussed adding these parameters only in the dev environment, not in the product itself?

serjiokov:
Should not be in prod

PR #34274, opened 2024-06-05
dbeaver/pro#2902 EAP limitations

ShadelessFox:


plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/actions/datasource/DataSourceHandler.java
```diff
@@ -89,6 +94,24 @@ public static void connectToDataSource(
                 return;
             }
 
+            if (DBWorkbench.getPlatform().getApplication().isEarlyAccessProgramActive()) {
```
serge-rider: The idea was to limit total amount of connections, not just active ones.
We must check it in new connection create handler.
We also must check that only one project is loaded and restrict creation of more than one project.

Also we cannot use getAllDataSources function (it triggers all projects loading).

plugins/org.jkiss.dbeaver.ui.statistics/src/org/jkiss/dbeaver/ui/statistics/UIStatisticsActivator.java
```diff
@@ -53,21 +51,6 @@ public static void setTrackingEnabled(boolean enabled) {
             return;
         }
         setPreferenceValue(PREF_FEATURE_TRACKING_ENABLED, enabled);
-        if (tracker != null) {
```
serge-rider, line 56: Why delete this code?

serge-rider, line 56: isTrackingEnabled must return true if we are in EAP mode

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.statistics/src/org/jkiss/dbeaver/ui/statistics/UIStatisticsActivator.java
ShadelessFox: Unused.

ShadelessFox: The application instance is not available at this point. During the application startup, the preference is set to true, effectively emulating the desired behavior.

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/actions/datasource/DataSourceHandler.java
ShadelessFox: We haven't finished discussing the desired approach yet. This is the first iteration, and I think it's ok to keep it like this for now

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPApplication.java
```diff
@@ -136,4 +136,24 @@ default boolean isCommunity() {
         return false;
     }
 
+    /**
+     * Whether this product can participate in the Early Access Program but no license is present
+     */
+    default boolean isEarlyAccessProgramCandidate() {
```
serge-rider: WE don't use it in CE

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPApplication.java
```diff
@@ -136,4 +136,24 @@ default boolean isCommunity() {
...
+        return false;
+    }
+
+    /**
+     * Whether this product is part of the Early Access Program
+     */
+    default boolean isEarlyAccessProgramActive() {
```
serge-rider: Maybe in CE it should be isStatisticsCollectionRequired? Or move all stuff related to it into PRO

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.statistics/src/org/jkiss/dbeaver/ui/statistics/PrefPageUsageStatistics.java
```diff
@@ -76,6 +77,12 @@ protected Control createPreferenceContent(@NotNull Composite parent) {
             NLS.bind(StatisticCollectionMessages.statistic_collection_pref_content_opensource_link, LINK_GIHUB_REPO),
             SelectionListener.widgetSelectedAdapter(selectionEvent -> ShellUtils.launchProgram(LINK_GIHUB_REPO)));
 
+        if (DBWorkbench.getPlatform().getApplication().isStatisticsCollectionRequired()) {
+            checkSendUsageStatistics.setEnabled(false);
```
serge-rider, line 81: true?

ShadelessFox: It's selected down below in the `performDefaults` method

PR #34278, opened 2024-06-06
#29307-add OcientMetaModel

### LINKED ISSUE #29307: Ocient - DDL generating incorrectly
### Description

When using an Ocient database, attempting to export any table with values of `BIGINT` and `SMALLINT` turns the values into `LONG` and `SHORT` respectively 
image
image

### DBeaver Version

24.0.3

### Operating System

Ubuntu 20.04.5 LTS

### Database and driver

1. Database name and version: Ocient
2. Driver name and version: Ocient JDBC 3.1.4

### Steps to reproduce

1. Create a table with a BIGINT value or a SMALLINT value
2. Generate DDL

### Additional context

This happens because of the way our driver reports metadata into DBeaver. To fix this I plan on extending `GenericMetaModel.java` into the Ocient model. I believe this is all that needs to be done but please correct me if I'm wrong.

mjsterckx:
Generating DDL for Ocient tables and views resulted in SQL that could not be executed on an Ocient system. This PR adds `OcientMetaModel` with a custom implementation for the `getObjectDDL` function, using Ocient's `export table/view` commands.


E1izabeth (approved):

E1izabeth:
Thank you for contribution

ShadelessFox (approved):

LonwoLonwo:
Hello @mjsterckx 

Thanks for your contribution.

I want to clarify something. By default, DBeaver generates table DDL. With information of columns, keys, etc. In your implementation in case of the reading DDL from database - there is no super method as an option, as I can see. I think, it is better to leave an ability to use standard DBeaver logic for table DDL creation for exceptional cases.

mjsterckx:
Hi @LonwoLonwo 

To make sure I get it right: you mean that if `getObjectDDL()` returns an exception, we catch that exception in `getTableDDL`/`getViewDDL` and call `super` as a fallback option? If so, I can add that shortly.

mjsterckx:
I've added those changes to the PR. Please let me know if that was not what you meant, or if there's anything else that should change.

plugins/org.jkiss.dbeaver.ext.ocient/src/org/jkiss/dbeaver/ext/ocient/model/OcientTable.java
```diff
@@ -0,0 +1,44 @@
...
+		GenericStructContainer container,
+        @Nullable String tableName,
+        @Nullable String tableType,
+        @Nullable JDBCResultSet dbResult)
+    {
+        super(container, tableName, tableType, dbResult);
+    }
+
+	@Override
+    @Property(hidden = true, editable = true, updatable = true, order = -1)
```
LonwoLonwo:
```suggestion
    @Property(hidden = true, order = -1)
```

LonwoLonwo: Table DDL is not editable and updatable.

LonwoLonwo:
Yeah, thanks, that was that.
I added a small new extra note.

plugins/org.jkiss.dbeaver.ext.ocient/src/org/jkiss/dbeaver/ext/ocient/model/OcientMetaModel.java
```diff
@@ -0,0 +1,120 @@
...
+        catch (SQLException e) {
+            throw new DBException(e, dataSource);
+        }
+    }
+
+    public String getViewDDL(DBRProgressMonitor monitor,
+            GenericView sourceObject, Map<String, Object> options) throws DBException {
+    	String schemaName = sourceObject.getContainer().getName();
+        String viewName = sourceObject.getName();
+        String viewDDLSQL = String.format("export view %s.%s", schemaName, viewName);
```
serge-rider: It is better to use DBUtils.getFullQualifiedName instead of `%s.%s`  concatenation.
getFullQualifiedName quotes schema and table names if needed (in case of special characters in identifiers).

plugins/org.jkiss.dbeaver.ext.ocient/src/org/jkiss/dbeaver/ext/ocient/model/OcientMetaModel.java
```diff
@@ -0,0 +1,120 @@
...
+        } catch (DBException e) {
+        	return super.getViewDDL(monitor, sourceObject, options);
+        }
+    }
+
+    public String getTableDDL(DBRProgressMonitor monitor,
+            GenericTableBase sourceObject, Map<String, Object> options) throws DBException {
+    	String schemaName = sourceObject.getContainer().getName();
+        String tableName = sourceObject.getName();
+        String tableDDLSQL = String.format("export table %s.%s", schemaName, tableName);
```
serge-rider: The same

plugins/org.jkiss.dbeaver.ext.ocient/src/org/jkiss/dbeaver/ext/ocient/model/OcientView.java
```diff
@@ -0,0 +1,44 @@
...
+	public OcientView(
+		GenericStructContainer container,
+        @Nullable String viewName,
+        @Nullable String viewType,
+        @Nullable JDBCResultSet dbResult)
+    {
+        super(container, viewName, viewType, dbResult);
+    }
+
+	@Override
```
serge-rider: Please do not use TAB characters (1 tab = 4 spaces)

serge-rider (changes requested):

mjsterckx:
Hi @serge-rider , do the most recent changes look sufficient to address your comments?

E1izabeth:

MashaKorax:
Hello @mjsterckx . For now, we have no test environment for testing your PR. Could you share  ocient instance with me for testing?

mjsterckx:
@MashaKorax I will discuss with my team on how to approach this and I'll get back to you.

mjsterckx:
@MashaKorax Someone from my company has emailed Serge, but please let me know if there's a different point of contact/email address we can or should use.

MashaKorax:
Hello @mjsterckx . I asked Serge about your email. As far as I understand, you do not have the opportunity to provide a test instance. In this case, we can offer you an early access build so that you can check the fix yourself. So, what do you think?

mjsterckx:
@MashaKorax That would be much appreciated, thank you! I'll loop in my product manager and then I'd be happy to verify this asap.

PR #34281, opened 2024-06-06
dbeaver/dbeaver#34280 Show info about current key, top-left icon change

### LINKED ISSUE #34280: Data editor: show info about current key
We should show information about current table key which will be used to save data changes.
Currently we show only read-only icon to mark non-editable resultsets (which do not have a key).

serge-rider:


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/internal/ResultSetMessages.properties
```diff
@@ -113,6 +113,9 @@ controls_resultset_ref_menu_references=<Table References>
 controls_resultset_ref_menu_no_associations=<No Foreign Keys>
 controls_resultset_ref_menu_associations=<Table Foreign Keys>
 
+controls_resultset_results_read_only = All columns are read-only
+controls_resultset_results_edit_key=Results edit key: {0}
```
E1izabeth, line 117:
```suggestion
controls_resultset_results_edit_key = Results edit key: {0}
```

uslss:
verified

PR #34285, opened 2024-06-06
#34055 Don't force default database, when user name is not empty

### LINKED ISSUE #34055: Setting default database on connect based on username
### Description

While trying to connect to a given PostgreSQL server which has one database with the same name as its user (also its owner), still doesn't set that database as default, as implied it would do in the connection setup.
Upon connection, the default database is always set to 'postgres' instead.

The 'Database' field says it will use the username to set the default one, if left emtpy:
image

Leaving those fields empty prompts the user for both the username and password. Then the expected result is DBeaver setting the default database as the same username informed on connect.

This feature would help on the workflow, as the user wouldn't need to go to the database list and set the desired one as default.
We have lots of sample databases that are created and deleted frequently, so setting one profile for each database only to be changed soon after is a chore on itself.

So, I wanted to know if what I described is in fact the intended behavior of the empty database field (therefore a bug?), or if it has another purpose that I maybe didn't understand correctly.

Thanks!

### DBeaver Version

Community Edition 24.0.5.202405191503

### Operating System

Windows 10

### Database and driver

- PostgreSQL 16 server
- PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Now it works according to the documentation - if database field is empty, then we take user name and try to connect to the database with the same name and set it as default. If there is no such database, it's expected that connection attempt fail.


ShadelessFox (approved):

serjiokov (approved):

uslss:
verified

PR #34302, opened 2024-06-07
CB-5129. Implemented add and delete logic for user team api

DenisSinelnikov:


plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMAdminController.java
```diff
@@ -66,6 +68,8 @@ void createUser(
     void invalidateAllTokens() throws DBException;
 
     void setUserTeams(String userId, String[] teamIds, String grantorId) throws DBException;
+    void addUserTeams(String userId, String[] teamIds, String grantorId) throws DBException;
```
ShadelessFox: Please add annotations are reformat code

ShadelessFox (approved):

PR #34329, opened 2024-06-11
dbeaver/pro#2941 Fix new table creation

E1izabeth:
Should not be reproducible.
Please test several different databases.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorHandlerObjectCreateBase.java
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorHandlerObjectCreateBase.java
@@ -222,6 +222,9 @@ public abstract class NavigatorHandlerObjectCreateBase extends NavigatorHandlerO
                 return Status.CANCEL_STATUS;//GeneralUtils.makeErrorStatus("Null object returned");
             }
             monitor.beginTask("Save " + newObject.getClass().getSimpleName(), 3);
+            var closure = new Object() {
+                DBNDatabaseNode node = null;
+            };
             try {
                 if (parentObject instanceof DBSObject) {
                     if ((objectMaker.getMakerOptions(((DBSObject) parentObject).getDataSource()) & DBEObjectMaker.FEATURE_SAVE_IMMEDIATELY) != 0) {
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorHandlerObjectCreateBase.java
```diff
@@ -222,6 +222,9 @@ protected IStatus run(DBRProgressMonitor monitor) {
                 return Status.CANCEL_STATUS;//GeneralUtils.makeErrorStatus("Null object returned");
             }
             monitor.beginTask("Save " + newObject.getClass().getSimpleName(), 3);
+            var closure = new Object() {
+                DBNDatabaseNode node = null;
+            };
```
ShadelessFox: Why not the common pattern with a single-element array?
```java
DBNDatabaseNode[] node = {null};
```

Your approach will result in the creation of a new inner anonymous class file.

E1izabeth: I don't like this pattern, it looks bad. Of course, if you insist, I can change it,

E1izabeth: Also that's what IDE suggest :)

ShadelessFox: Any approach introducing a heap-allocated object with a reference will work. The array version is very common and doesn't introduce overhead like an anonymous class.

E1izabeth: OK

serge-rider: I'd move it in the class variable, why use closure or array?

serge-rider (changes requested):

PR #34340, opened 2024-06-12
dbeaver/dbeaver#22313 Fix smart mode

### LINKED ISSUE #22313: Smart Commit mode
### Description

When using the smart commit mode in new versions when we commit, in the next query it asks to commit again.

I attach images for your review.

### DBeaver Version

DBeaver Community 23.3.1

### Operating System

Windows 11 Pro

### Database and driver

Database DB2 i Series

### Steps to reproduce

1. I make a modification to a record.
2. I save the modification and commit.
3. I do a "Select * from" and it shows me as if I had modified the registry again.

SmartCommitMode1
SmartCommitMode2

### Additional context

_No response_

HocKu7:
Now "Return to auto-commit on transaction end" is true for dev and test environment


Matvey16:
Enabling `return to auto-commit` in connection types won't change anything if smart-commit in connection type is turned off by default. If specific connection settings are disabled it won't matter because smart commit just doesn't work. But turning it on manually from a toolbar turnes on specific connection properties that would override type properties anyway.

E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Verified

Matvey16:
Verified

PR #34352, opened 2024-06-13
dbeaver/dbeaver#23070 Print the driver that isn't downloaded

### LINKED ISSUE #23070: Show the URL attempted when connection not working
I'm using DBeaver behind a restrictive proxy, and when it tries to download drivers, it gets blocked and just simply shows the error that it couldn't connect.

If it can at least show me what URL it was trying then I can submit that to the proxy bypass to get it to work.

The alternative to download the driver actually did not work. I was able to get the driver installed, but yet, it still is stuck on some internet connection, and again, I have no idea what it's trying to connect to.

HocKu7:
Now we show information about the failed downloading and, particularly, the URL to jar that we didn't manage to download
image


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ThreadExceptionsContext.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+public class ThreadExceptionsContext {
+
+    private static final Map<Thread, List<Exception>> THREAD_TO_EXCEPTIONS = new ConcurrentHashMap<>();
+    private static final Map<Thread, Thread> THREAD_BINDING_MAP = new ConcurrentHashMap<>();
+    //private static final Logger log = LoggerFactory.getLogger(ThreadExceptionsContext.class);
```
E1izabeth:
```suggestion
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ThreadExceptionsContext.java
```diff
@@ -0,0 +1,91 @@
...
+
+    public static List<Exception> getListExceptionsForCurrentThread() {
+        Thread threadForRegistry = getThreadForRegistry();
+        List<Exception> exceptionList = THREAD_TO_EXCEPTIONS.get(threadForRegistry);
+        if(exceptionList == null){
```
E1izabeth:
```suggestion
        if (exceptionList == null) {
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ThreadExceptionsContext.java
```diff
@@ -0,0 +1,91 @@
...
+        return new ArrayList<>(exceptionList);
+    }
+
+    public static Set<Exception> getSetExceptionsForCurrentThread() {
+        return new HashSet<>(getListExceptionsForCurrentThread());
+    }
+
+    public static void registerExceptionForCurrentThread(Exception exception) {
+
+        Thread threadForRegistry = getThreadForRegistry();
```
E1izabeth:
```suggestion
    public static void registerExceptionForCurrentThread(Exception exception) {
        Thread threadForRegistry = getThreadForRegistry();
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ThreadExceptionsContext.java
```diff
@@ -0,0 +1,91 @@
...
+            //log.warn(String.format("Thread %s is already in the binding map for ThreadExceptionContext. Overriding...",
+            //    currentThread.getName()));
+        }
+        THREAD_BINDING_MAP.put(currentThread, callerThread);
+    }
+
+
+    public static void realiseResources() {
+
+        Thread currentThread = Thread.currentThread();
```
E1izabeth:
```suggestion
    public static void realiseResources() {
        Thread currentThread = Thread.currentThread();
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -579,6 +579,36 @@ public static IStatus makeExceptionStatus(int severity, Throwable ex) {
         return makeExceptionStatus(severity, ex, false);
     }
 
+    public static IStatus transformExceptionsToStatus(@NotNull Set<Exception> exceptions) {
+
+        if (exceptions.isEmpty()) {
+            return new Status(IStatus.ERROR, (Class<?>) null, "Empty exceptions list");
+        }
+        Set<String> excaptionMessageSet = new HashSet<>();
```
E1izabeth:
```suggestion
        Set<String> exceptionMessageSet = new HashSet<>();
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ThreadExceptionsContext.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * If the thread which you are registering is a different thread that gonna read exception list
+     *
+     * @param callerThread a key where you will be able to find exceptions for current thread
+     */
+    public static void bindCurrentContextFor(Thread callerThread) {
+        Thread currentThread = Thread.currentThread();
+        //todo if it will need, there can be used a set for threads
+        if (THREAD_BINDING_MAP.get(currentThread) != null) {
+            log.warn(String.format("Thread %s is already in the binding map for ThreadExceptionContext. Overriding...",
```
E1izabeth: Do we really need to log it? User might see his message, but it's hard to understand, what this message is about, without context of this code

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/driver/DriverLibraryMavenArtifact.java
```diff
@@ -339,7 +339,8 @@ protected MavenArtifactVersion resolveLocalVersion(DBRProgressMonitor monitor, b
         }
         MavenArtifactVersion version = getArtifactVersion(monitor);
         if (version == null) {
-            throw new IOException("Maven artifact '" + path + "' not found");
+            String versionMessageError = preferredVersion != null ? String.format(":%s", preferredVersion) : "";
+            throw new IOException(String.format("Maven artifact %s %s not found", path, versionMessageError));
```
E1izabeth:
```suggestion
            throw new IOException(String.format("Maven artifact %s %s is not found", path, versionMessageError));
```

ShadelessFox:
I don't understand the whole logic behind this PR. What does it do? Stores a list of exceptions per thread?

E1izabeth:
Yes, as far as I understand, it accumulated exceptions per thread between the first download attempt and the end of attempting. And I agree, that it's doubtful

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ThreadExceptionsContext.java
HocKu7: I suppose there could be a concurrency issue that we can't see right now, so I prefer to see any such bugs related to concurrency in the log. I'm thinking of the log for developers, not for users, isn't it?

E1izabeth: Yes, but sometimes users read logs, if some bug occurred. 
If it's an exclusive situation, then I'm fine with logging it. I thought it will be called each time.

HocKu7:
Generally speaking, the main goal of the PR is to show the URL of the driver that we didn't manage to reach. For this, I created a ThreadExceptionContextx that allows us to collect exceptions through different threads. Otherwise, there is a need a complicated refactoring

ShadelessFox:
Maybe we need refactoring. This solution is quite complex and really hard to follow/debug.

ShadelessFox:
Also, on your screenshot, there are two mentions of the same file but in different repositories, but the top message clearly says that DBeaver can't access the last one. Is this expected?

HocKu7:
We discussed that with @serge-rider. He would like to have some variation of the Exception Context.  About the screenshot - yes, it corresponds to the current logic of the attempt to download the driver. It tries to download the driver from each repository(maven, sonatype), and then if it gets an exception, it checks internet connectivity to maeven.org and gets an exception again. So, there are three exceptions

serge-rider:
I don't like the approach with thread local variables.
It is better to pass ErrorContext to WebUtils and other methods as a parameter

ShadelessFox:
Why don't you use the existing mechanism of suppressed exceptions? When catching multiple exceptions, collect them and add them as suppressed to a parent exception

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/WebUtils.java
```diff
@@ -56,6 +58,11 @@ public static URLConnection openConnection(String urlString, DBPAuthInfo authInf
         return openURLConnection(urlString, authInfo, referrer, 1);
     }
 
+    @NotNull
+    public static URLConnection openConnection(String urlString, DBPAuthInfo authInfo, String referrer, DBRProgressMonitor monitor) throws IOException {
```
serge-rider: By convention monitor is always the first parameter

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/driver/DriverDependenciesTree.java
```diff
@@ -192,16 +200,24 @@ private void grayOutInstalledArtifact(DBPDriverDependencies.DependencyNode node,
...
 
-    public boolean handleDownloadError(DBException e) {
+    public boolean handleDownloadError(DBException causeException) {
         try {
             checkNetworkAccessible();
         } catch (DBException dbException) {
-            DBWorkbench.getPlatformUI().showError("Download error",
-                "Network error", dbException);
-            return false;
+            if(causeException instanceof DBExceptionWithHistory exceptionWithHistory){
```
serge-rider: Code style

serge-rider (changes requested):

HocKu7:
> Why don't you use the existing mechanism of suppressed exceptions? When catching multiple exceptions, collect them and add them as suppressed to a parent exception

Decided that approach with monitor more suitable for the task

Matvey16:
`Exception during to connect to` while understabdable from technical standpoint, is not a correct sentance. Can we change it?

Matvey16:
Verified

PR #34384, opened 2024-06-18
#34371 Add color for functions in sql editor

### LINKED ISSUE #34371: Color "SQL datatype color" is applied for some sql keywords other than datatypes
### DBeaver Version

Community Edition 24.1.0

### Operating System

Windows 10

### Database and driver
PostgreSQL 16
PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

1. copy-paste this code to a new script:
select 1::int;
select min(oid) from pg_class;

2. change color of datatype as shown in gif

Expected: word "int" has changed the color, word "min" has not changed the color (because it is not a datatype, it is an sql keyword).
Actual: both words changed the color

### Additional context
Another case:
```
    select 
        row_number() over(order by p."Id") as "Id",
        p."MyColumn"::int::bool as "SomeColumnName123",
        upper(p."Code"),
    from  public."Product" as p;
```

изображение

E1izabeth:
The new color is the same as color for data types by default, however now user can change it.

Example script for PostgreSQL for testing:
`select tan(90), "3"::int from "Somteable" s`
(Note: the connection should be established)

image


ShadelessFox (approved):

Matvey16:
Verified

PR #34393, opened 2024-06-19
#29942 Fix dollar quote rule and postgre escape string rule

### LINKED ISSUE #29942: SQL Still problematic
### Description

Running just a single statement in in DBeaver is problematic.

Even though:
SQL Processing Statement delimiter is **;**
Ignore native delimiter is **unchecked**
Blank line is statement delimiter is set to **Never**
Remove trailing query delimiter is **unchecked**

placing my cursor instaide of a complete statement and pressing "Execute SQL Query" will occationally try to run every statement in the SQL Editor with disasterous results.

I had thought that switching:
Blank line is statement delimiter: from the default of **Smart**, to **Never** had fixed the issue, but apparently all doing that accomplished was to reduce the occurance not eliminate it entirely.

The only way to ensure that only a single statement is run is to select the entire statement.

This is a marked regression from the behavior of previous versions,
hopefully this can be corrected  soon.

### DBeaver Version

Community Edition 24.0.4

### Operating System

Windows 10

### Database and driver

PostgreSQL 16

### Steps to reproduce

Have an SQL Editor window with many sql statements

place the cursor (symbolized by {c} in the example below) inside a single statement
press the orange arrow (Execute SQL Query)
sometimes it will run the specified query
other times it will run **all** queries in that editor

ex:
select * from foo;
select yaba, daba, doo {c} from flintstones;
insert into baz set fld1 = 'something' where fld2 > 10;
delete from important_table where entered < '2023-01-01';

since the cursor {c} is in the select from flinstones statement that's the **only** statement that should be run.

### Additional context

_No response_

E1izabeth:
If user uses dollar strings, then the settings should be set like this
image

With this settings now we correctly handle such cases (see foldings)
image

Script for testing
```
select
    format($x$%1$s$x$, 'foo'); 

select
    'bar';

select
    'baz';

DO $x$
BEGIN
	RAISE NOTICE '%', 'Test';
END $x$;


select 'zoo' 
from x
left join y on a=b
left join z on right(c,d)=0
where p>q

```


uslss:
verified

PR #34395, opened 2024-06-19
#23377 Add `right` and `left` to the list of non-reserved keywords

### LINKED ISSUE #23377: Syntax highlighting stops working after functions contains multiple argument
### Description

image

As you can see here, everything after `RIGHT` function failed to highlight properly.

But if I change to a different function like `TRIM`, then everything works fine:
image

What makes things more weird is, if I remove the second argument of RIGHT, then the highlight works fine although the syntax is wrong:
image

### DBeaver Version

Community Version 24.0.1.202403241413

### Operating System

Windows 10

### Database and driver

PostgreSQL 9.4.26
PostgreSQL 9.4.26 (Greenplum Database 6.22.2)

### Steps to reproduce

In the join statement, use some functions like CAST(), RIGHT() or LEFT() with 2 or more arguments.

### Additional context

_No response_

E1izabeth:
Issue should be fixed


uslss:
verified

PR #34401, opened 2024-06-20
dbeaver/pro#2651 Import project to TE

E1izabeth:
image

image


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/LocalProjectImpl.java
```diff
@@ -264,4 +265,11 @@ public void recoverProjectDescription() throws IOException {
             EMPTY_PROJECT_TEMPLATE.replace("${project-name}", project.getName()));
     }
 
+    @NotNull
+    public Map<String, Map<String, Object>> getAllResourceProperties() {
+        this.loadMetadata();
+        synchronized (resourcesSync) {
+            return new HashMap<>(this.resourceProperties);
```
serge-rider: Isn't it TreeMap?

serge-rider (changes requested):

PR #34405, opened 2024-06-20
dbeaver/pro#2946 Fix find qm session

HocKu7:
Add a few tries for fetching QM history so the exception about QM session info should be gone


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/controls/querylog/QueryLogViewer.java
```diff
@@ -1224,7 +1226,14 @@ public List<QMEvent> evaluate(DBRProgressMonitor monitor) throws InvocationTarge
 
                 String qmSessionId = null;
                 if (DBWorkbench.getPlatform().getApplication() instanceof QMSessionProvider provider) {
+                    int tries = 0;
                     qmSessionId = provider.getQmSessionId();
+                    //github#2946
+                    while (qmSessionId == null && tries < RETRIES_QM_WAITING) {
+                        Thread.sleep(WAITING_QM_SESSION_SECONDS_PER_TRY * 1000);
```
serge-rider: Let's wait after first fail. Why wait for 5 seconds in non-TE products?

Also let's make 60 attempts after 1 second.

serge-rider (changes requested):

PR #34448, opened 2024-06-24
#22795 Transaction status update fix

### LINKED ISSUE #22795: Rollback timer flickers after the last update
### Description

The rollback timer at the bottom right used to stay stable but now it is flickering two times every second. Is this an intentional choice? Attaching the screen recording.

https://github.com/dbeaver/dbeaver/assets/114302087/bcb7d6c3-9f07-4b94-a545-f9395f6ee890

### DBeaver Version

Version 23.3.4.202402041841

### Operating System

MacOS Sonoma 14.3.1

### Database and driver

Server: Snowflake 8.5.2
Driver: Snowflake 3.14.4

### Steps to reproduce

Wait after your first query and put Dbeaver in the background without rolling your mouse pointer over it. Make it so that you would be able to see the Rollback timer in the background. After a while, you'll see that it will flicker like in the screen recording.

### Additional context

_No response_

ShadelessFox:
I did not find a better solution except for creating a separate service that will listen for part activation and update the status bar accordingly, but I decided to keep it simple.

If we ever decide to move this indicator from the status bar to the toolbar, we can easily use the approach described above.

STR:
1. You must have two connections: one with a disconnect timeout set and the second without
2. Open SQL editors for both of them
3. Go away for some time so the timer appears

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
@@ -253,6 +253,7 @@ public class SQLEditor extends SQLEditorBase implements
     private Boolean isDisableFetchResultSet = null;
     private boolean datasourceChanged;
     private TransactionStatusUpdateJob transactionStatusUpdateJob;
+    private AbstractPartListener partListener;
 
     private final ArrayList<SQLEditorAddIn> addIns = new ArrayList<>();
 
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -2165,6 +2166,22 @@ protected IStatus run(DBRProgressMonitor monitor) {
 
         transactionStatusUpdateJob = new TransactionStatusUpdateJob();
         transactionStatusUpdateJob.schedule();
+        partListener = new AbstractPartListener() {
```
serge-rider: Looks a bit over-complicated.
Maybe we'd better check that editor is currently active in the TransactionStatusUpdateJob ?

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -304,6 +303,10 @@ public void widgetDisposed(DisposeEvent e) {
         }
     };
 
+    static {
+        new TransactionStatusUpdateJob().schedule();
```
serge-rider: Let's create it in SQLEditor.createPartControl

serge-rider (changes requested):

Matvey16:
Verified

PR #34449, opened 2024-06-24
dbeaver/dbeaver#22679 Foreign key dialog UIX

### LINKED ISSUE #22679: Foreign keys screen is super confusing
### Description

This is an UX bug report on the **Define foreign keys** dialog box, which is just so confusing. Also, on the **Define virtual foreign keys** dialog box, which is a tad worse:

image

It was only very recently that I figured out how I was supposed to operate it. That after many attempts, across many months. Also, I could not find any sort of manual for these screens.

image

For almost an year, I just stared at the **OK** button stubbornly grayed out. No matter what I did, it just refused to get enabled.
virtual foreign keys.

image

I even hypothesized on whether "virtual foreign keys" was some sort of experimental yet-to-be-implemented kind of feature.

image

It was only after some frustation-driven random clicking aimlessly around, that this sphynx of a screen decided to share its secrets with my poor soul

image

It was only then, that it dawned upon me that these cells...

image

...secretly are combo boxes:

image

After filling these cells, the  **OK** button finally enables:

image

Yay!

But this is just **Sub-issue 1**. There's more...

**Sub-issue 2** The titles ...

image

image

...are not consistent with other dialog boxes...

image

image

**Sub-issue 3** This **Choose ...** button...

image

...opens the **Select reference table catalog/schema** dialog box:

image

It's got this *Show connected databases only* checkbox... 

image

...which is redundant with this little guy, up here...

image

**Sub-issue 4** Speaking of which... There you can find these **Filter table containers** and **Filter tables, procedures, etc.** options...

image

Their names are complete lies.

* "Filter table containers" actually filters schemes, not table containers. *(Remember, 'tables containers' = 'schemes' + 'packages')*
* "Filter tables, procedures, etc." actually filters packages.

I wonder whats even the point of talking about "tables" and "procedures" in a screen that is not supposed to show either.

There is another screen in DBeaver with a similar issue:

* https://github.com/dbeaver/dbeaver/issues/22675#issue-2113564276

**Sub-issue 5** Back to the **Define foreign keys** dialog box. There's this *Reference table:* list...

image

This list definitely deserves a filtering feature.

**Sub-issue 6** The tables in this list don't have any context menu. There should be at least be a "Copy" option.

**Sub-issue 7** When you have that list focused and type a letter on the keyboard, it advances to the first item whose name starts with that letter.

If you hit more keys, you would expect that it would keep refining the search by the second, third, fouth... letters. But no... It seeks by the first letter, once again.

This is super frustating.

Sub-issues 5 and 7 combined make creating virtual keys withing a BBoM scheme a huge pain. You have to pick the reference table you are looking for very diligently, which is annoying.

**Sub-issue 8** Finally, there's this **Unique Key:** combo box...

image

...which does make sense if you *already know* how the dialog operates, but is rather daunting the first time you see it.

If you are stuck with the disabled **OK** button caused by the **Columns:** table not beign filled, as I previously was, you may think there's something wrong with this "unique key" thing. Even when it's just fine.

That's some pure cognitive overload you just would not want.

To be honest, I can't devise a better place for it to be. Perhaps, transform the *Reference table:* list into a tree...?

Maybe? I dunno `¯\_(ツ)_/¯`

### DBeaver Version

DBeaver Community Edition 23.3.0.202312031432

### Operating System

Windows 11 (21H2/10.0.22000.2652)

### Database and driver

`oracle.jdbc.OracleDriver`

### Steps to reproduce

*(...)*

### Additional context

*(...)*

serge-rider:


uslss:
[We discussed this with the team and decided to do the following:](https://github.com/dbeaver/dbeaver/issues/22679#issuecomment-2123112557)

- [x] User doesn’t understand what to do because there is no columns => Add hint to the tables ‘Click here to choose source column’
- [x] Delete duplicate dialog names after slash **|** 
- [x] Write everywhere ‘Show connected databases only’ (near the checkbox and in the Context menu)
- [x] Filter table container — rename to “Filter catalogs and schemes” **- not renamed**
- [x] Add a filter to the Reference tables **-no changes**
- [x] Add a context menu for the table **-no changes**

serge-rider:
Fixed

uslss:
List of reference tables is empty if another table container was chosen

https://github.com/dbeaver/dbeaver/assets/31996417/c9fdc8f6-936e-4c98-b76b-0cdba92e611f

serge-rider:
Fixed

uslss:
Still reproducible. no errors are in the logs

uslss:
verified

PR #34453, opened 2024-06-25
Altibase: Support for DB objects that were not supported in DB browser such as Job, Database Link, Directory, and Library

### LINKED ISSUE #34041: Altibase: support non-schema objects: Directory, Job, Library, Database Link
Hello, 

I'd like to support non-schema objects at Altibase DBMS such as Directory, Job, Library, Database Link.

Thanks,

zennken:
Hello,

I am happy to ask a pull-request for #34041

https://github.com/ALTIBASE/Documents/blob/master/Manuals/Altibase_7.3/eng/Administrator

- Job: Altibase provides the JOB object, which is the addition of an execution schedule to a stored procedure. The stored procedure to be executed, the point in time of execution, the interval after which it is to be executed and etc. can be set when creating the JOB object.
- Database link: Database Link unites disparate data sources on interconnected servers to produce a single unified result, even if the data are stored in different kinds of data servers that are physically far apart from one another.
- Directory: Stored procedures are able to control files, which allow them to read from and write to text files in the file system managed by the operating system. Thanks to this functionality, the user can perform various kinds of tasks using stored procedures such as leaving messages in files, reporting the results of files or reading data from files for insertion into tables. The directory object is used to manage information about the directories accessed by stored procedures.

https://github.com/ALTIBASE/Documents/blob/master/Manuals/Altibase_7.3/eng/External%20Procedures%20Manual.md
- Library (External Library Objects): It is necessary for user-defined functions or, in other words, the dynamic library file in which the external procedure resides to be made identifiable by the Altibase server. For this purpose, Altibase provides the external library object, which is a database object, and SQL statements for its management.

New or modified code has been checked with Check Style to ensure compliance.

Thanks,

image


E1izabeth:
Thank you for contribution. We will review your PR and let you know.

zennken:
> Thank you for contribution. We will review your PR and let you know.

If you need test scenario for these objects, please let me know.

Thanks for your help :)

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.altibase/src/org/jkiss/dbeaver/ext/altibase/model/AltibaseDbLink.java
```diff
@@ -0,0 +1,149 @@
...
+        linkId = JDBCUtils.safeGetInt(resultSet, "LINK_ID");
+        userMode = JDBCUtils.safeGetInt(resultSet, "USER_MODE");
+        remoteUserId = JDBCUtils.safeGetString(resultSet, "REMOTE_USER_ID");
+        linkType = JDBCUtils.safeGetInt(resultSet, "LINK_TYPE");
+        targetName = JDBCUtils.safeGetString(resultSet, "TARGET_NAME");
+        created = JDBCUtils.safeGetTimestamp(resultSet, "CREATED");
+        lastDdlTime = JDBCUtils.safeGetTimestamp(resultSet, "LAST_DDL_TIME");
+    }
+
+    @NotNull
```
ShadelessFox: Nullability annotations are not required for primitive types, such as int, long, float, etc.

plugins/org.jkiss.dbeaver.ext.altibase/src/org/jkiss/dbeaver/ext/altibase/model/AltibaseDbLink.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @Override
+    public String getObjectDefinitionText(DBRProgressMonitor monitor, Map<String, Object> options) throws DBException {
+        if (CommonUtils.isEmpty(ddl)) {
+            options.put("SCHEMA", isPublic() ? "PUBLIC" : this.getParentObject().getName());
+            ddl = ((AltibaseMetaModel) getDataSource().getMetaModel()).getDbLinkDDL(monitor, this, options);
+        }
+
+        return (CommonUtils.isEmpty(ddl)) ? "" : ddl + ";";
```
ShadelessFox: Why don't you want to append `;` in the `if` above?

plugins/org.jkiss.dbeaver.ext.altibase/src/org/jkiss/dbeaver/ext/altibase/model/AltibaseDbLink.java
```diff
@@ -0,0 +1,149 @@
...
+import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
+import org.jkiss.dbeaver.model.struct.DBSObject;
+import org.jkiss.utils.CommonUtils;
+
+import java.sql.Timestamp;
+import java.util.Map;
+
+public class AltibaseDbLink extends AltibaseObject<GenericStructContainer> implements DBPScriptObject, DBPRefreshableObject {
+
+    protected String ddl;
```
ShadelessFox: Why is it `protected`? I can't see any usages outside this class. Same for other objects.

plugins/org.jkiss.dbeaver.ext.altibase/src/org/jkiss/dbeaver/ext/altibase/model/AltibaseDbLink.java
zennken: Thanks, I'll remove the useless annotations from AltibaseDbLink, AltibaseDirectory, AltibaseJob, and AltibaseLibrary classes.

plugins/org.jkiss.dbeaver.ext.altibase/src/org/jkiss/dbeaver/ext/altibase/model/AltibaseDbLink.java
zennken: Oops..no need to use protected. I'll change them. Thanks :)

plugins/org.jkiss.dbeaver.ext.altibase/src/org/jkiss/dbeaver/ext/altibase/model/AltibaseDbLink.java
zennken: This takes into account cases where creating DDL fails due to exceptional circumstances while fetching from DB. Although it is very unlikely, I implemented it this way because I thought there would be no need to add ';' to the empty DDL string if this case occurred.

ShadelessFox: I suggest directly appending the semicolon to the buffer the first time you read it to avoid redundant copies introduced by concatenation.

zennken: Okay, since it's actually very unlikely to happen, I'll move the semicolon inside the if statement as you recommended.

Thanks for your review :)

ShadelessFox (approved):

E1izabeth:
By the way, we passed your PR to our QA team and waiting for their approval.

zennken:
> By the way, we passed your PR to our QA team and waiting for their approval.

I was curious about the progress. Thanks for kind notice. :)

zennken:
Hello,

Could you please update me on the progress of QA? I'm worried that this PR has escaped from the work items :)

E1izabeth:
Hi! Unfortunately, they haven't checked this PR yet. It's waiting in the list :)

zennken:
> Hi! Unfortunately, they haven't checked this PR yet. It's waiting in the list :)

Thanks for your check. As long as it's on the list, that's OK :)

uslss:
verified

PR #34470, opened 2024-06-25
CB-5165. Add parameter to api - autocommit

DenisSinelnikov:


alexander-skoblikov (changes requested):
property is not saved in json configuration

PR #34510, opened 2024-06-26
dbeaver/pro#2630 Fix duplicated server output

E1izabeth:


uslss:
Output is now only shown for script execution. If I execute `USE db` only - no output is shown

https://github.com/dbeaver/dbeaver/assets/31996417/f0b90abf-a074-4650-9d3a-765150528eaf

uslss:
verified

PR #34535, opened 2024-06-27
Dbeaver/pro#2952 rsv statis columns

serge-rider:


Matvey16:
1. If there is no data in the table the hint would be shown with any cursor position (there are no border to an area where you can hover to see hint)

https://github.com/dbeaver/dbeaver/assets/82543000/a7a377e3-5fe8-487e-8033-8df283161b65

2. Every cell is uneditable if there is no key. The only way to edit data (even without saving) is to add virtual key manually on properties tab. This is a regression because previous behavior allows editing data and would propose to create a virtual key after clicking save.


https://github.com/dbeaver/dbeaver/assets/82543000/f42f1d0f-aac4-4d93-a7a3-caae8541bd62

serge-rider:
Fixed

Matvey16:
Verified

PR #34572, opened 2024-06-29
dbeaver/pro#3007 Use default SSH connect timeout (10S)

serge-rider:


plugins/org.jkiss.dbeaver.net.ssh.sshj/src/org/jkiss/dbeaver/model/net/ssh/SSHJSessionController.java
```diff
@@ -64,6 +66,7 @@ protected SSHClient createNewSession(
         client.setConnectTimeout(connectTimeout);
         client.getConnection().getKeepAlive().setKeepAliveInterval(keepAliveInterval);
         client.getTransport().getConfig().setLoggerFactory(new FilterLoggerFactory());
+        client.getTransport().setTimeoutMs(10000);
```
ShadelessFox:
```suggestion
        client.getTransport().setTimeoutMs(connectTimeout);
```
Or is it not the same thing?

serge-rider: Oops. It was a line added during debugging, we don't need it (setConnectTimeout does all the job)

MashaKorax:
verified

PR #34580, opened 2024-06-30
#34579 Expand navigator tree on connect. Change default + show loader…

### LINKED ISSUE #34579: Visualize connection establish in navigator trees
Currently connection establish is visualized only when user expands navigator item explicitly.

When connection is triggered from SQL editor activation there is no any visualization (connect job is visualized in background tasks but it is hard to mention).

serge-rider:
… before connect


serge-rider:
We probably need to refactor DBP event model to handle beforeConnect and afterConnect events.

The best solution would be to handle these events in navigator and show load spinners instead of expanding nodes explicitly.

PR #34593, opened 2024-07-01
dbeaver/pro#2764 Add idea import plugin

HocKu7:


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -156,7 +156,9 @@ protected void initializeRemoteInstance(@NotNull DBRProgressMonitor monitor) thr
                 activeDatabaseName = url.substring(divPos + 1, lastPos);
             }
         }
-        if (CommonUtils.isEmpty(activeDatabaseName) && CommonUtils.isEmpty(configuration.getUserName())) {
+        if (CommonUtils.isEmpty(activeDatabaseName) && !CommonUtils.isEmpty(configuration.getUserName())) {
```
serge-rider: Why?

plugins/org.jkiss.dbeaver.ext.oracle/plugin.xml
```diff
@@ -358,7 +358,7 @@
                     id="oracle_thin"
                     label="Oracle"
                     class="oracle.jdbc.OracleDriver"
-                    sampleURL="jdbc:oracle:thin:@{host}[:{port}]/{database}"
+                    sampleURL="jdbc:oracle:thin:@{host}[:{port}][:][/]{database}"
```
serge-rider: Wrong URL. Please don't change any drivers config

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
HocKu7: There was a mistake. If a database name is empty, we must put a username in the database, as I understand. Just faced it during the task, but it could be a separate ticket

PR #34616, opened 2024-07-02
#34613 Fix too long name of simple structure compare report

### LINKED ISSUE #34613: Simple Structure Compare fails when the generated file name is too long.
### Description

When you compare several database (8 in our case), each with around 30 character names, it cant save the generated comparison file, it gives the error: 

`"C:\Users\myfirstname.mylastname.myorg\Downloads\compareXXXXXX-report.html (The filename, directory name, or volume label syntax is incorrect)"`

where XXXXX is a long list of our database names which gives a path of 335 characters.

The limit in windows 10 for file names is 255 chars
The limit in windows 10 for the path is 260 chars. 

In this case, I would recommend that dbeaver either allow users to edit the file name, or truncate the file name.

It is possible in windows 10 to enable long path names, but this requires regedit which is blocked on all corporate PCs, and it didnt solve the problem for me.

### DBeaver Version

community edition 24.1.1

### Operating System

windows 10

### Database and driver

N/a

### Steps to reproduce

On a windows 10 Machine:
1) select 7 databases to compare, each with names > 30 chars.
2) right lick on one of the selected databases and choose "compare/migrate"->"Simple Structure Compare"
3) hit next until it tries to save the file.
4) observe error saving file.

### Additional context

_No response_

E1izabeth:
If more than 3 objects selected, we won't append object name to report

image


ShadelessFox (approved):

Destrolaric (approved):

MashaKorax:
verified

PR #34668, opened 2024-07-05
#34667 Use local monitor in data read (if metadata read is disabled)

### LINKED ISSUE #34667: [MS SQL Server] Aditional queries are executed even when the property to not do it is checked
### Description

A property to not read metadata in SQL Editor doesn't work in some cases in SQL Server. For example when a basic select is executed on the database that is selected in hotbar (metedata won't be rad as expected, when select is executed on another database with a full table name)

### DBeaver Version

24.1.3

### Operating System

Any

### Database and driver

SQL Server

### Steps to reproduce

1. Check `Do not read table information in SQL and data editors` in metadata settings
2. Open SQL Editor and select any database as active
3. Perform a select on the table in this database

**Expected:** Only one query is executed
**Actual:** A few other queries apart from the only desired select can be seen in Query manager

### Additional context

_No response_

serge-rider:


uslss:
verified

PR #34700, opened 2024-07-08
dbeaver/pro#3021 Don't disable Compress option if Append on filename conflict has bee chosen

E1izabeth:
Now if you have Append option chosen and check Compress checkbox, Append will be disabled and selector will select Autofix name option
image


ShadelessFox (approved):

Destrolaric (approved):

Matvey16:
Selector changes to autofix regardless of initial choice when compress is checked


https://github.com/dbeaver/dbeaver/assets/82543000/54c697cd-8d43-41a1-9f25-9463717eab43

E1izabeth:
It's intentional, we can't left selector on Append option

Matvey16:
Verified

PR #34702, opened 2024-07-08
Fix of LCK10-J and MET05-J violations that lead to errors

KonstantinRight:
Hi, I found two vulnerabilities in code and fixed them.
The first one is [LCK10-J](https://wiki.sei.cmu.edu/confluence/display/java/LCK10-J.+Use+a+correct+form+of+the+double-checked+locking+idiom). You're supposed to apply volatile keyword to a double-checked locking variable, otherwise the pattern [doesn't work](https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html).
The second one is [MET05-J](https://wiki.sei.cmu.edu/confluence/display/java/MET05-J.+Ensure+that+constructors+do+not+call+overridable+methods). This is a recommendation itself, but in the case of the code I fixed it led to an error: isBinaryContents() in TextOpen constructor was called before the binary field in TextOpenFile was initialized.


plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/TextWithOpen.java
```diff
@@ -51,14 +51,16 @@ public class TextWithOpen extends Composite {
...
     private final boolean multiFS;
+    private final boolean binary;
 
     public TextWithOpen(Composite parent, boolean multiFS) {
-        this(parent, multiFS, false);
+        this(parent, multiFS, false, false);
     }
     
-    public TextWithOpen(Composite parent, boolean multiFS, boolean secured) {
+    public TextWithOpen(Composite parent, boolean multiFS, boolean secured, boolean binary) {
```
ShadelessFox, line 62: To avoid breaking existing code that utilizes this API, I'd suggest making an overloaded version of the constructor.

ShadelessFox (commented):
Thanks for the contribution

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/TextWithOpen.java
E1izabeth: @KonstantinRight there are some comments that need to be addressed.

KonstantinRight: Yeah, it seems reasonable. I've added an overloaded constructor just like you said.

E1izabeth (approved):

E1izabeth:

MashaKorax:
verified

PR #34706, opened 2024-07-08
dbeaver/dbeaver#34657 PostgreSQL metadata read fixes. Duplicate conte…

### LINKED ISSUE #34657: Dbeaver don't finish "Discover atribute {field}" - Redshift
### Description

Even on simple queries like:
select * from stl_query where query=887227603
it stays forever on this message
image
I'm getting only the first 200 rows for big queries, and in this particular case, only one row.

Other clients get that info almost instantaneously.

The error log I get from Window->show view->error log
[Error log.log](https://github.com/user-attachments/files/16100305/Error.log.log)

### DBeaver Version

24.1.1

### Operating System

macOS 14

### Database and driver

Redshift
com.amazon.redshift:redshift-jdbc42:RELEASE 2.1.0.29

### Steps to reproduce

It happens randomly, makes me restart Dbeaver

### Additional context

_No response_

serge-rider:
…xt fix


Matvey16:
Verified

PR #34708, opened 2024-07-08
#2998 Disconnect from data source when closing all associated editors

### LINKED ISSUE #2998: How to automatically close connections when Script window is closed ?
Hello,
When I'm using a MySql script, connecting to the desired serveur and database, after running the script (a select query) if I close the script window, the connection to the database is keeped.
How to autoatically realase the connection when window is closed ?
Some client side items don't work if there's another connection to the database.
Thanks

ShadelessFox:
image


E1izabeth (approved):

Destrolaric (approved):

uslss:
verified

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/ApplicationWorkbenchWindowAdvisor.java
```diff
@@ -468,6 +472,48 @@ private void closeEmptyEditors() {
...
+                    }
+                }
+            }
+        }
+
+        return false;
+    }
+
+    private boolean isEditorForDataSource(@NotNull IEditorReference ref, @NotNull DBPDataSourceContainer container) {
+        try {
```
serge-rider, line 510: It checks all editors including non-activated.
Is it expected?

ShadelessFox: Yes - otherwise, it might result in unwanted disconnects.

Let's say you have three tabs:
```
[ Album (deactivated) ] [ Artist (deactivated) ] [ Invoice (activated, selected) ]
```
And you want to close the "Invoice" tab. Since it's the only activated editor, the connection can be closed. After that, the "Artist" tab is activated, and since the connection is closed, it must be opened again. Hence, this method considers ALL editors, whether they're activated or not

serge-rider: Then this feature doesn't make much sense. Quite often there are multiple SQL editors open and most of them are not active. In this case connection will never close.

ShadelessFox: The idea was to close a connection when there are no more editors left regarding whether they're active or not - this makes sense in my head with an explanation in the form of an example above.

ShadelessFox: We can schedule an overview of this feature in the following UX discussion next week

PR #34712, opened 2024-07-09
dbeaver/dbeaver#29934 Use default database as user name

### LINKED ISSUE #29934: Teradata connection requires Database/Schema override
When creating a connection to a Teradata database the field Database/schema (DATABASE=) is required by DBeaver but not by Teradata as it is only used to override the default database defined for the database user. It causes some confusion for newcomers.
> Specifies a default database for use after logon, to override the default database defined for the database user.
https://teradata-docs.s3.amazonaws.com/doc/connectivity/jdbc/reference/current/jdbcug_chapter_2.html#URL_DATABASE

The field Database/schema to be optional.
I guess changing it here: https://github.com/dbeaver/dbeaver/blob/28ecc6e2f484ee8c8038c0fd736d2c80d98b92cb/plugins/org.jkiss.dbeaver.ext.generic/plugin.xml#L584
Perhaps using defaultDatabase somehow?
Setting it to the username of the user connecting?
**Additional context**

serge-rider:


uslss:
verified

PR #34745, opened 2024-07-10
dbeaver/pro#3012 Task scheduling error when exporting CSV to AWS S3

serjiokov:
Throw a custom exception in case of incorrect path definition


serjiokov:
image

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamTransferConsumer.java
```diff
@@ -772,19 +772,20 @@ private Path makeOutputFile(@NotNull DBRProgressMonitor monitor, @Nullable Strin
...
             log.error("Error resolving output folder", e);
-            dir = Path.of(outputFolder);
+            throw new DBCException(e.getMessage(), e);
         }
         if (!Files.exists(dir)) {
             try {
                 Files.createDirectories(dir);
             } catch (IOException e) {
                 log.error("Error creating output folder", e);
+                throw new DBCException(e.getMessage(), e);
```
serge-rider: You can just pass IOException, there is not specific handler here

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamTransferConsumer.java
```diff
@@ -772,19 +772,20 @@ private Path makeOutputFile(@NotNull DBRProgressMonitor monitor, @Nullable Strin
...
     @NotNull
-    private Path makeOutputFile(@NotNull DBRProgressMonitor monitor, @Nullable String suffix, @NotNull String outputFolder) {
+    private Path makeOutputFile(@NotNull DBRProgressMonitor monitor, @Nullable String suffix, @NotNull String outputFolder) throws DBCException {
         Path dir;
         try {
             dir = DBFUtils.resolvePathFromString(monitor, getProject(), outputFolder);
         } catch (Exception e) {
             log.error("Error resolving output folder", e);
-            dir = Path.of(outputFolder);
+            throw new DBCException(e.getMessage(), e);
```
serge-rider: What the point?
resolvePathFromString already throws DBException.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamTransferConsumer.java
serjiokov: Inside the method resolvePathFromString have use `Path.of()` that 
Throws InvalidPathException - if the path string cannot be converted to a Path

PR #34761, opened 2024-07-10
CB-5134. Added filter for connections when connected

DenisSinelnikov:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNDatabaseNode.java
```diff
@@ -848,6 +851,19 @@ public void setNodeFilter(DBXTreeItem meta, DBSObjectFilter filter, boolean save
             } else {
                 log.error("Cannot detect child node type - can't save filter configuration");
             }
+        } else if (this.getParentNode() instanceof DBNDataSource) {
+            Class<?> childrenClass = this.getChildrenOrFolderClass(meta);
+            if (childrenClass != null) {
+                dataSource.setObjectFilter(
```
ShadelessFox: Why is it inside a method that loads children? It shouldn't modify the configuration 🤔

DenisSinelnikov: removed

alexander-skoblikov (approved):

PR #34770, opened 2024-07-11
Dbeaver/pro#3073 fix checkbox appearing on linux

Destrolaric:
Closes dbeaver/pro#3073
Checkbox should not appear on linux distros

### DIFF
--- a/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageDatabaseUserInterface.java
+++ b/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageDatabaseUserInterface.java
@@ -195,7 +195,7 @@ public class PrefPageDatabaseUserInterface extends AbstractPrefPage implements I
             ));
         }
 
-        if (isStandalone) {
+        if (isStandalone && !RuntimeUtils.isLinux()) {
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageDatabaseUserInterface.java
```diff
@@ -256,7 +256,9 @@ private void setSettings() {
     protected void performDefaults() {
         DBPPreferenceStore store = DBWorkbench.getPlatform().getPreferenceStore();
         if (isStandalone) {
-            useEmbeddedBrowserAuth.setSelection(store.getDefaultBoolean(DBeaverPreferences.UI_USE_EMBEDDED_AUTH));
+            if (!RuntimeUtils.isLinux()) {
```
ShadelessFox, line 259: What about macOS?

Destrolaric: It works for it

ShadelessFox (approved):

Matvey16:
Verified

PR #34780, opened 2024-07-11
dbeaver/pro#3074 SQL completion fix: ignore lazy DBS objects (avoid d…

serge-rider:
…atabase connection instantiation)


E1izabeth (approved):

Matvey16:
Verified

PR #34783, opened 2024-07-11
#34396 Disable index reading

### LINKED ISSUE #34396: [Redshift] Support indexes
Redshift JDBC Driver 2.1.0.29 supports indexes. DBeaver should support them too.
We need to investigate indexes in redshift and fix them or disable.
If we can't do anything with it, we need to disable it.

ShadelessFox:
This PR disables metadata reading for indexes


Matvey16:
Verified

PR #34784, opened 2024-07-11
#22831 Fix problem markers update

### LINKED ISSUE #22831: SQL script execution does not recognise changes in SQL script
### Description

When I open the SQL Editor and create an SQL script with an error in it, it (rightfully) shows the place/nature of the error. However even after I fix it, it keeps pointing to the same place and reporting the same error, even quoting part of the code that is simply not there anymore.

This happened with parenthesis errors, and now with trying to call st_point with 3 arguments; I altered the line to `SET origin=st_pointz(c.x, c.y, c.z)  FROM (VALUES` yet the error popup quotes

> SQL Error [25P02]: ERROR: current transaction is aborted, commands ignored until end of transaction block
  ERROR: current transaction is aborted, commands ignored until end of transaction block
  ERROR: current transaction is aborted, commands ignored until end of transaction block
    ERROR: function st_point(numeric, numeric, numeric) does not exist
  Hint: No function matches the given name and argument types. You might need to add explicit type casts.
  Position: 55
    ERROR: function st_point(numeric, numeric, numeric) does not exist
  Hint: No function matches the given name and argument types. You might need to add explicit type casts.
  Position: 55

Workarounds are re-opening the script, restarting DBeaver or turning the PSQL parser off and on.

### DBeaver Version

Community 23.3.4

### Operating System

Ubuntu 22.04

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
I'm not sure how to reproduce it. It happened to me only once.
This is a partially blind fix.


serjiokov (approved):

uslss:
Also cannot reproduce.
Checked that script editing still works as expected

PR #34792, opened 2024-07-12
#34791 Idle connections timeout setting get/set fixes

### LINKED ISSUE #34791: Bug: idle connections timeout set to 0 by default for new connections
### Description

Currently when we create new connection then "Close idle connections interval" is always zero.
1. But it must be the default from the connection type.
2. If we don't change value in certain connection then value must be taken from connection type always (i.e. we can change it for connection type and it will affect all connections of this type)

### DBeaver Version

24.1

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

serge-rider:


Matvey16:
Verified

PR #34793, opened 2024-07-12
#34378 Don't show proposals inside literals in INSERT INTO query

### LINKED ISSUE #34378: Autocomplete appears and suggests object names inside single quoted values
### Description

It may appear somwhere else too, but the most obvious example of this bug can be found when the user generates sql for insert and then tries to enter values in SQL Editor. Autocomplete starts suggesting table or container names after any symbol is typed

### DBeaver Version

24.1.1

### Operating System

Any

### Database and driver

Any

### Steps to reproduce

1. Generated insert SQL for any table with varchar column
2. Copy it to SQL editor
3. Start typing values inside single-quoutes

**Expected:** Nothing is suggested
**Actual:** Object names appear in autocomplete

### Additional context

https://github.com/dbeaver/dbeaver/assets/82543000/12532e2c-3e7a-49e3-b94f-28941094a786

E1izabeth:


ShadelessFox (approved):

serjiokov (approved):

Matvey16:
Verified

Issue #341, opened 2016-04-06
Oracle - Cannot generate explain plan as non admin user due to insufficient privileges (but TOAD can)


steromano87:
If you try to get an explain plain using a non admin user, DBeaver raises an "Insufficient privileges" error and does not output anything:

![image](https://cloud.githubusercontent.com/assets/8759452/14313312/219eacf0-fbf1-11e5-9fb8-5fda0b9faa1f.png)

However, if I try to get the eplain plan for the same query using the same user in TOAD, the explain plan is calculated normally:

![image](https://cloud.githubusercontent.com/assets/8759452/14313336/59bea75c-fbf1-11e5-8817-1d502a71ef1a.png)

Here are listed all the user permissions from USER_ROLE_PRIVS (USER_SYS_PRIVS and USER_TAB_PRIVS are empty for the given user):

| GRANTED_ROLE | ADMIN_OPTION | DEFAULT_ROLE | OS_GRANTED |
| -------------- | -------------- | -------------- | ------------ |
| UTE_READ     | NO           | YES          | NO         |

serge-rider:
Can't reproduce that.
Please post error with stacktrace

steromano87:
Hi Serge,

here you can find the output of the log:
!SESSION 2016-04-08 11:02:35.426 -----------------------------------------------
eclipse.buildId=unknown
java.version=1.8.0_25
java.vendor=Oracle Corporation
BootLoader constants: OS=win32, ARCH=x86, WS=win32, NL=it
Command-line arguments:  -os win32 -ws win32 -arch x86

!ENTRY org.jkiss.dbeaver.model 4 0 2016-04-08 11:16:19.817
!MESSAGE org.jkiss.dbeaver.model.exec.DBCException: Errore SQL [1031] [42000]: ORA-01031: privilegi insufficienti

!SUBENTRY 1 org.jkiss.dbeaver.model 4 0 2016-04-08 11:16:19.817
!MESSAGE DBCException: Errore SQL [1031] [42000]: ORA-01031: privilegi insufficienti
!STACK 0
org.jkiss.dbeaver.model.exec.DBCException: Errore SQL [1031] [42000]: ORA-01031: privilegi insufficienti

	at org.jkiss.dbeaver.ext.oracle.model.plan.OraclePlanAnalyser.explain(OraclePlanAnalyser.java:121)
[...]

PR #34802, opened 2024-07-13
#21827 remove delete query before plan query

PzaThief:
For Oracle plan queries, dbeaver drops plans with the same statement ID just in case.
However, some users only have read-only permissions due to the DBA's access controls and cannot use delete queries due to insufficient permissions. (yes, my story...., and it seems like there are some have already experienced. #341 )
The PLAN_TABLE is a GLOBAL TEMPORARY TABLE, not shared with other session and emptied at the end of the session automatically.
Even if same statement id was present before in same session, Oracle overwrite the statement and plan so it will be safe still.
So, I propose to remove the delete statement in the explain code.


E1izabeth (approved):

kseniaguzeeva:
Checked

serge-rider:
I'd suggest to surround it in try/catch and print debug log if delete fails.

Explains are still pollute plan tables, it is better to clean them up.

E1izabeth:
@PzaThief 
Thanks for your contribution! Please check the review comment https://github.com/dbeaver/dbeaver/pull/34802#issuecomment-2469839898

E1izabeth:

PR #34817, opened 2024-07-15
dbeaver/pro#3306 [SUB] Fix data filters in Neo4j

serjiokov:
Insert attributes by ordinal position


serjiokov:
Fixed issue with ordering document type attributes.

PR #34823, opened 2024-07-15
dbeaver/pro#34731 Fix dollar quotes handling on execution from the cursor position

E1izabeth:
Please test Vertica, Snowflake and PostgreSQL queries with $$. I noticed that Snowflake has some bug on settings switching or initial state initialization, so, if something won't work for it, please compare with older version of DBeaver first and report a new bug.
Also please test the setting defining the way, how we handle/highlight double-dollar parts for these dialects.


serjiokov (approved):

uslss:
verified

PR #34826, opened 2024-07-15
3085 sql completion performance

serge-rider:


E1izabeth:
Please check New completion engine too

E1izabeth:
How to test:

0. Open Query Manager
1. Type `select * from `
2. Continue typing the name of the table, that doesn't exist in the database, for example, `qpeieqiqri`
4. Observe, that every new key pressed cause DBeaver to send query to the database

If you can't reproduce, try some database without schemas.

E1izabeth:
On typing in SQLite
image

E1izabeth (approved):

Matvey16:
Verified

PR #34832, opened 2024-07-16
dbeaver/dbeaver#34831 Fix postgres database name if it was empty

### LINKED ISSUE #34831: Setting database name the same as user if it is empty
### Description

Expected: When we are trying to connect to the Postgresql database without explicit naming of the database, the name of the database will be the same as the username.
Actual: When we are trying to connect to the Postgresql database without explicit name of the database, the name of the database will be set to 'postgres'

### DBeaver Version

last devel

### Operating System

_No response_

### Database and driver

PostgreSQL

### Steps to reproduce

_No response_

### Additional context

_No response_

HocKu7:


Matvey16:
Verified

PR #34860, opened 2024-07-17
Dbeaver/pro#2886 add bigquery dashboards

HocKu7:
Added a predefined dashboard for BigQuery
image


plugins/org.jkiss.dbeaver.ext.bigquery/plugin.xml
```diff
@@ -76,4 +76,17 @@
         </authModel>
     </extension>
 
+
+    <extension point="org.jkiss.dbeaver.dashboard">
+        <dashboard id="bigquery-dash"
+                   label="API metrics"
+                   description="BigQuery API monitoring charts"
+                   defaultView="browser"
```
serjiokov: Would be nice to have i18n here

plugins/org.jkiss.dbeaver.ext.bigquery/plugin.xml
```diff
@@ -76,4 +76,17 @@
         </authModel>
     </extension>
 
+
+    <extension point="org.jkiss.dbeaver.dashboard">
+        <dashboard id="bigquery-dash"
```
serge-rider: Please move it to PRO

serge-rider (changes requested):

E1izabeth (approved):

serjiokov (approved):

plugins/org.jkiss.dbeaver.model.dashboard/src/org/jkiss/dbeaver/model/dashboard/registry/DashboardRegistry.java
```diff
@@ -82,6 +83,10 @@ private DashboardRegistry(IExtensionRegistry registry) {
         DBDashboardContext staticContext = new DBDashboardContext();
         for (DashboardProviderDescriptor dp : dashboardProviders.values()) {
             for (DashboardItemConfiguration dashboard : dp.getInstance().loadStaticDashboards(dp)) {
+                //fixme after resolving macOs issue
```
serge-rider: Super-ugly hardcode.

If we want to disable some dashboard for certain OSes then we need to use OS mapping. 
See driver files mapping xml.

serge-rider (changes requested):

ShadelessFox (approved):

PR #34905, opened 2024-07-19
#34731 Boundaries of the query in smart mode

### LINKED ISSUE #34731: Boundaries of the query are detected incorrectly if a query contains blank line and the query followed by ; (with option: Blank line statement delimiter = Smart)
### Description

Boundaries of the query are detected incorrectly if a query contains blank line and the query followed by ; (with option: Blank line statement delimiter = Smart).

### DBeaver Version

Community Edition 24.1.3.202407042322

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

- set option: Blank line is statement delimiter = Smart
- use this code:
```
select * 
from (values(random())
        
    ) as s(v)
where s.v is not null;
```
- set caret at where-clause
- execute the query with Ctrl + Enter

### Additional context

Notice that it works correctly without symbol ; somehow so it is possible to detect boundaries in spite of an empty line in between a query text

E1izabeth:
Please test NOT only the case in the issue.


HocKu7 (approved):

serjiokov (approved):

Matvey16:
Verified

PR #34968, opened 2024-07-24
dbeaver/pro#3065 Don't create a new query processor if now new tab requested explicitly

E1izabeth:
```
select pg_sleep(15);
select * from "Employee" e;
```
1. Run pg_sleep with `Ctrl+Enter`
2. Run any other query
3. This message should appear
image

____________________
Please test using of all possible combinations of these buttons
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
@@ -2747,9 +2747,9 @@ public class SQLEditor extends SQLEditorBase implements
                     }
                 }
                 // Just create a new query processor
-                if (!foundSuitableTab) {
+                if (!foundSuitableTab && newTab) {
                     // If we already have useless multi-tabbed processor, but we want single-tabbed, then get rid of the useless one  
-                    if (curQueryProcessor instanceof MultiTabsQueryProcessor 
+                    if (!useTabPerQuery(queries.size() == 1) && curQueryProcessor instanceof MultiTabsQueryProcessor
                         && curQueryProcessor.getResultContainers().size() == 1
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -2747,9 +2747,9 @@ public boolean processQueries(@NotNull final List<SQLScriptElement> queries, fin
                     }
                 }
                 // Just create a new query processor
-                if (!foundSuitableTab) {
+                if (!foundSuitableTab && newTab) {
                     // If we already have useless multi-tabbed processor, but we want single-tabbed, then get rid of the useless one  
-                    if (curQueryProcessor instanceof MultiTabsQueryProcessor 
+                    if (!useTabPerQuery(queries.size() == 1) && curQueryProcessor instanceof MultiTabsQueryProcessor
```
E1izabeth, line 2753: It used to be here, but was accidentally removed with refactoring

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -2747,9 +2747,9 @@ public boolean processQueries(@NotNull final List<SQLScriptElement> queries, fin
                     }
                 }
                 // Just create a new query processor
-                if (!foundSuitableTab) {
+                if (!foundSuitableTab && newTab) {
```
E1izabeth, line 2751: We should create new processor only if it was explicitly requested (newTab == true)

Destrolaric (approved):

HocKu7 (approved):

uslss:
verified

PR #34973, opened 2024-07-24
dbeaver/dbeaver#34819 Fix generate ddl context menu

### LINKED ISSUE #34819: Generate SQL context menu not activating
### Description

When opening table detail, right-click on the empty grid space the "Generate SQL" menu item shows up in the context menu does not activate

### DBeaver Version

24.1.2.202407071610

### Operating System

Windows 10

### Database and driver

Firebird v3.0.10, Jaybird JCA/JDBC driver 4.0.0

### Steps to reproduce

https://github.com/user-attachments/assets/d8d5afcf-5bae-4a10-854a-de54984cc282

### Additional context

_No response_

HocKu7:
Now Generate DDL appears properly in the window
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorContributor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/generator/SQLGeneratorContributor.java
@@ -32,6 +32,7 @@ import org.jkiss.dbeaver.model.DBPScriptObject;
 import org.jkiss.dbeaver.model.DBUtils;
 import org.jkiss.dbeaver.model.data.DBDAttributeBinding;
 import org.jkiss.dbeaver.model.exec.DBCExecutionContext;
+import org.jkiss.dbeaver.model.navigator.DBNDatabaseFolder;
 import org.jkiss.dbeaver.model.sql.generator.SQLGenerator;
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -798,15 +799,19 @@ public static URI makeURIFromFilePath(@NotNull String path) throws URISyntaxExce
...
+    public static <T> T adapt(Object sourceObject, Class<T> adapter, boolean allowActivation) {
+        return adapt(sourceObject, adapter, allowActivation, adapter::isInstance);
+    }
+
     /////////////////////////////////////////////////////////////////////////
     // Adapters
     // Copy-pasted from org.eclipse.core.runtime.Adapters to support Eclipse Mars (#46667)
 
-    public static <T> T adapt(Object sourceObject, Class<T> adapter, boolean allowActivation) {
+    public static <T> T adapt(Object sourceObject, Class<T> adapter, boolean allowActivation, Predicate<Object> isInstance) {
```
E1izabeth: Could we call this parameter in a different way? Maybe castPredicate, castTest, ... (I suppose you can figure it out better, than me). Currently it's not clear what should I pass to the parameter with this name only be the method contract.

E1izabeth (approved):

serjiokov (approved):

Matvey16:
Verified

PR #34985, opened 2024-07-25
#34179 Prohibit execution of mutating statements if connection is read-only

### LINKED ISSUE #34179: I was able to delete a row in a DB that was read only.
### Description

Due to data security issues, I cannot give more information. But, I had the client DB in read only.
I was able without an issue to remove data from the DB with "DELETE FROM..."_

### DBeaver Version

community Version 23.2.0.202309041200

### Operating System

Windows 11

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
image


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLQuery.java
```diff
@@ -510,6 +515,30 @@ public boolean isDropTableDangerous() {
...
+            return false;
+        }
+        if (statement instanceof Select) {
+            SelectBody selectBody = ((Select) statement).getSelectBody();
+            if (selectBody instanceof PlainSelect plainSelectBody) {
+                return plainSelectBody.isForUpdate() || plainSelectBody.getIntoTables() != null;
+            }
+            return false;
+        } else {
+            return true;
```
LonwoLonwo, line 531: Just to check - is it expected to have "true" on non-select statement in isModifyingSelect method?

E1izabeth: I think, I misunderstood this method.
P.S. I haven't changed it, just moved and renamed.

serjiokov (approved):

LonwoLonwo (approved):

uslss:
verified

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLQuery.java
```diff
@@ -510,6 +510,21 @@ public boolean isDropTableDangerous() {
...
```
serge-rider, line 531: No sure we can use this approach. There are many non-SELECT queries which are not modifying. E.g. USE, SHOW, WITH, etc. Every dialect may have extra allowed statements.
Let's restrict only INSERT, DELETE, UPDATE, CREATE, DROP, ALTER for now.

serge-rider (changes requested):

uslss:
verified

Issue #34963, opened 2024-07-23
Advanced Copy > As Markdown -> does not use the column alias in paste result


m3c-ode:
**Is your feature request related to a problem? Please describe.**
I wonder if it's a settings or not. but tried looking, and didn't find anything.
I would like the Copy and Paste feature to use the columns names given in the alias. 
It would make for some much more readable results.

**Describe the solution you'd like**
When giving alias to column names, would like them used in the paste result

**Describe alternatives you've considered**

**Additional context**

MashaKorax:
Hello @m3c-ode . Actually, Advanced Copy > As Markdown with column alias works for me. Please provide more details to reproduce the issue

https://github.com/user-attachments/assets/a5935936-5ae2-49a7-bcd9-1fb2e01f803c

dstivanov:
@MashaKorax I think that https://github.com/dbeaver/dbeaver/pull/35007 will solve this, isn't it ? 
cc @m3c-ode

m3c-ode:
It seems to work on a simple query as you display @MashaKorax 

But when I have a query with JOIN, it doesn't seem to be consistent.

![image](https://github.com/user-attachments/assets/25796259-31eb-481a-9e4f-b46d0d63d200)
![image](https://github.com/user-attachments/assets/1c9d8ef7-7d40-4730-8142-8c2e49fb23bb)
![image](https://github.com/user-attachments/assets/c29e8a27-2e6f-4f73-955f-4dc604bf9975)

MashaKorax:
I reproduced the issue.
It is fixed by PR https://github.com/dbeaver/dbeaver/pull/35007

Issue #34050, opened 2024-05-23
When advanced copy to json, key is not the other name of the as in the sql statement, but the original field name


Saber-CC:
### Description

On the query result page, select a result, right-click -> Advanced Copy -> Copy as json to the clipboard, key is not the other name of the as in the sql statement, but the original field name

### DBeaver Version

24.0.5.202405191503

### Operating System

13.6.6 (22G630)

### Database and driver

MySQL

### Steps to reproduce

right-click -> Advanced Copy -> Copy as json to the clipboard

uslss:
@Saber-CC unfortunately cannot reproduce the issue. Could you add some screenshots or video with the issue?

thanks. And sorry for the late response

Saber-CC:
> @Saber-CC unfortunately cannot reproduce the issue. Could you add some screenshots or video with the issue?不幸的是无法重现该问题。您能否添加一些有关该问题的屏幕截图或视频？
> 
> thanks. And sorry for the late response谢谢。抱歉回复晚了

I upgraded to 24.1.0 and still have this problem. More specifically, in a mysql query, the alias as of the sql statement does not correspond to the key output into json, which is still the name in the table.Here's the video.

https://github.com/dbeaver/dbeaver/assets/49194769/0917e9fa-8896-48f1-afe5-a1d3a7f29013

uslss:
Unfortunately, still cannot reproduce the issue.
Could you try if it is reproducible for you against the recent version?
![image](https://github.com/dbeaver/dbeaver/assets/31996417/c7b22714-db49-4e18-a2ca-58f639f2a17f)

Saber-CC:
[...]

PR #35007, opened 2024-07-28
Use Column Aliases When Doing "Advanced Copy"

### LINKED ISSUE #18946: Query result: column aliases ignored when copied as CSV
### Description

Trying to copy the result of a query with more than 1 result column alias, those aliases are ignored and the original column names are pasted instead 

### DBeaver Version

Community Edition 22.3.3

### Operating System

Windows 11

### Database and driver

MariaDB 10.5.16

### Steps to reproduce

execute query "SELECT t.ID AS Alias1, t.Description AS Alias2 FROM mytable t"
select all the results clicking the top-left cell in the results section (note that the aliases are correctly displayed there)
right-click "Advanced Copy"->Copy as CSV
Paste in Excel/LibreOffice Calc/NotePad: the first row will be "ID" - "Description" instead of "Alias1" - "Alias2"

### Additional context

_No response_

dstivanov:
* As it was reported in #18946 the users where not seeing the columns aliases when doing the "Advanced copy" of the results;
* The problem was reported for MySQL and MariaDB, I also observed it personally with Postgres;
* The example of the problem is as follows:
    * Having a query like 
```sql
SELECT
	f.author as 'Author Name'
from
	family f
```
produces the following results in the UI: 
Screenshot 2024-07-29 at 1 30 34
* When the users selects a few results and selected `Advanced copy` -> `Copy as CSV` (or `Copy as TXT`, etc.) the results are as follows:
```csv
"author"
"Griffiths-Jones SR, Mifsud W, Gardner PP"
"Griffiths-Jones SR, Mifsud W"
```
and the alias of the `author` column is not used.
* This MR fixes this behavior for all the Advanced copies.
Cheers!


E1izabeth:
Thank you for contribution!

E1izabeth (approved):

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/AttributeMetaDataProxy.java
```diff
@@ -81,6 +81,11 @@ public DBSObject getSource() {
     @NotNull
     public String getLabel()
     {
+        DBSAttributeBase proxyAttribute = getProxyAttribute();
+        if (proxyAttribute instanceof DBCAttributeMetaData) {
+            return ((DBCAttributeMetaData) proxyAttribute).getLabel();
```
ShadelessFox: Just a nitpick: since we use Java 17, you can do the following:

```suggestion
        if (proxyAttribute instanceof DBCAttributeMetaData attributeMetaData) {
            return attributeMetaData.getLabel();
```

ShadelessFox:
We passed the PR to our QA team for further testing.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/AttributeMetaDataProxy.java
dstivanov: Addressed in the recent version. Also applied on the code that I used for the fix. @ShadelessFox Resolve at will.

E1izabeth:
@MashaKorax 
Please check #34050 and #34963 cases also.

MashaKorax:
verified

PR #35013, opened 2024-07-29
#34516 Option enablement fix; close idle connections with embedded drivers

### LINKED ISSUE #34516: Idle timeout setting from connection types does not affect connections
### Description

Two related issues:

1. When creating a new database connection, the default setting for "Close idle connection after (seconds)" is active and set to 0. This overrides any setting from the connection type and should not be the default
image

3. Idle timeout settings from the connection type do not work

### DBeaver Version

24.1.1

### Operating System

macOS 14.5

### Database and driver

MariaDB 3.4.0

### Steps to reproduce

1. Create a new MariaDB connection
2. Go to the Connection settings | Initialization tab
4. The "Close idle connection after (seconds)" box is checked and set to 0
5. Uncheck the box
6. Assign the a connection type to the connection
7. Set the idle timeout for the connection to a short duration (10 seconds)
8. Connect to the database. The connection will not close after the idle time expires

### Additional context

_No response_

ShadelessFox:
This PR fixes two issues mentioned in the ticket.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/jobs/DataSourceMonitorJob.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/jobs/DataSourceMonitorJob.java
@@ -21,7 +21,6 @@ import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.IJobChangeEvent;
 import org.eclipse.core.runtime.jobs.JobChangeAdapter;
 import org.jkiss.code.NotNull;
-import org.jkiss.dbeaver.DBException;
 import org.jkiss.dbeaver.Log;
 import org.jkiss.dbeaver.ModelPreferences;
 import org.jkiss.dbeaver.model.DBPDataSource;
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/jobs/DataSourceMonitorJob.java
```diff
@@ -315,9 +313,6 @@ public void scheduleMonitor() {
     }
 
     public static long getDisconnectTimeoutSeconds(@NotNull DBPDataSourceContainer container) {
-        if (container.getDriver().isEmbedded() && !DBWorkbench.getPlatform().getApplication().isMultiuser()) {
-            return 0;
-        }
```
E1izabeth, line 320: Is this removal intentional?

ShadelessFox: Yes. It addresses https://github.com/dbeaver/dbeaver/issues/34516#issuecomment-2244771049; c24eef8575a705a87980c70413e67032a3ed2ea1

E1izabeth (approved):

LonwoLonwo (approved):

Matvey16:
Verified

PR #35046, opened 2024-08-01
HANA connection dialog: convert GENERIC + host=<host>:443 to CLOUD (#35045)

### LINKED ISSUE #35045: improve connection dialog for HANA Cloud
HANA cloud UI offers an action to get the SQL endpoint: 
image

This returns a host:port string like `abcd1234-5678-abcd-1234-5678abcd1234.hana.prod-eu10.hanacloud.ondemand.com:443`

When creating a new HANA connection and pasting that value in the *Host* field, based on the `:443` port suffix, the port should be stripped of and the *Edition* should be automatically changed to `Cloud`.

*Note*: see PR below

kai-morich:


E1izabeth:
Thank you for contribution!

plugins/org.jkiss.dbeaver.ext.hana.ui/src/org/jkiss/dbeaver/ext/hana/ui/views/HANAConnectionPage.java
```diff
@@ -290,21 +290,34 @@ private void editionUpdated() {
...
+            ((GridData) layoutData).exclude = !visible;
         }
     }
 
+    private void hostUpdated() {
+        if (CommonUtils.isEmpty(hostText.getText())) {
+            return;
+        }
+        String host = hostText.getText().trim();
+        if (edition == HANAEdition.GENERIC && CommonUtils.isEmpty(portText.getText()) && host.endsWith(":443")) {
```
serge-rider: Let's move ":443" into constants (HTTPS_PORT_SUFFIX). 
So condition will be host.length() - HTTPS_PORT_SUFFIX.length()

plugins/org.jkiss.dbeaver.ext.hana.ui/src/org/jkiss/dbeaver/ext/hana/ui/views/HANAConnectionPage.java
```diff
@@ -182,7 +182,7 @@ public void loadSettings() {
         super.loadSettings();
         DBPConnectionConfiguration connectionInfo = site.getActiveDataSource().getConnectionConfiguration();
         edition = HANAEdition.fromName(connectionInfo.getProviderProperty(PROV_PROP_EDITION));
-        portValue = CommonUtils.toString(connectionInfo.getHostPort(), site.getDriver().getDefaultPort());
+        portValue = CommonUtils.toString(connectionInfo.getHostPort()/*, site.getDriver().getDefaultPort()*/);
```
serge-rider, line 186: We don't need default port anymore?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.hana.ui/src/org/jkiss/dbeaver/ext/hana/ui/views/HANAConnectionPage.java
kai-morich: All editions except GENERIC have there own rule to calculate the port from other fields or have a fixed port. 
GENERIC is not really an own edition, but a fallback that allows pasting already know host+port. The preset port was wrong in most cases as an on-premise HANA has an instance number in the range from 0 to 99 and the actual port depends on that number.

plugins/org.jkiss.dbeaver.ext.hana.ui/src/org/jkiss/dbeaver/ext/hana/ui/views/HANAConnectionPage.java
kai-morich: done

E1izabeth:
We passed your PR to our QA team for testing

MashaKorax:
verified

PR #35084, opened 2024-08-06
dbeaver/pro#3135 Oracle session query performance + connection page fix

serge-rider:


uslss:
after fix no sql query and execution plan are shown:
image
image

serge-rider:
Odd. 
I added a small fix but not sure that it fix results.
Anyway I can see results:

image

uslss:
verified
image

PR #35096, opened 2024-08-06
#34906 Set affinity to text for sqlite, if column data type is unknown

### LINKED ISSUE #34906: Some text fields show up as type "NA" and DBeaver can't find text in these fields
### Description

I'm trying to edit my Adobe Lightroom catalog file (SQLite) to change the location where it thinks my photos are. When I use "Search" -> "DB Full TexT" to search for the string "thunder", the search returns no results, despite my being able to see this string just fine in some fields. Further investigation reveals that DBeaver reports the data type for these fields as "NA", which I assume has something to do with the inability to search for text in them. Can anyone help me here?

### DBeaver Version

Enterprise Version 24.1.0.202406100905

### Operating System

macOS 14.5

### Database and driver

SQLite (Pro)

### Steps to reproduce

I've created a tiny version of the original database with just one table in it. Download it from:

https://www.dropbox.com/scl/fi/hwtr77qw78gbkm5jxn7pg/test.lrcat?rlkey=emz9anxkt1oc13vv5oos6600o&st=w4zl7ltm&dl=0

Open a SQLite connection to the file and do a full-text search for "thunder". It returns no results, even though you can see the string in the field "absolutePath"

### Additional context

_No response_

E1izabeth:
Full text search should work now.


plugins/org.jkiss.dbeaver.ext.sqlite/src/org/jkiss/dbeaver/ext/sqlite/model/SQLiteDataSource.java
```diff
@@ -87,7 +87,8 @@ public DBSDataType getLocalDataType(String typeName) {
         } else if (typeName.startsWith("REAL") || typeName.startsWith("FLOA") || typeName.startsWith("DOUB")) {
             affinity = SQLiteAffinity.REAL;
         } else {
-            affinity = SQLiteAffinity.NUMERIC;
+            // If type is unknown, let's assume it's a text. Otherwise, search and data editor doesn't work right.
```
serge-rider, line 93: Thus it never can be integer numeric. Add extra if for "INT".

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.sqlite/src/org/jkiss/dbeaver/ext/sqlite/model/SQLiteDataSource.java
```diff
@@ -86,8 +86,14 @@ public DBSDataType getLocalDataType(String typeName) {
             affinity = SQLiteAffinity.BLOB;
         } else if (typeName.startsWith("REAL") || typeName.startsWith("FLOA") || typeName.startsWith("DOUB")) {
             affinity = SQLiteAffinity.REAL;
-        } else {
+        } else if (typeName.toUpperCase().contains(SQLConstants.DATA_TYPE_INT) || typeName.toUpperCase().contains("NUMERIC") ||
```
ShadelessFox: Maybe cache the result of `typeName.toUpperCase()`?

E1izabeth: it's already in upper case, actually, so no needed here
fixed

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.ext.sqlite/src/org/jkiss/dbeaver/ext/sqlite/model/SQLiteDataSource.java
```diff
@@ -86,8 +86,13 @@ public DBSDataType getLocalDataType(String typeName) {
             affinity = SQLiteAffinity.BLOB;
         } else if (typeName.startsWith("REAL") || typeName.startsWith("FLOA") || typeName.startsWith("DOUB")) {
             affinity = SQLiteAffinity.REAL;
-        } else {
+        } else if (typeName.contains(SQLConstants.DATA_TYPE_INT) || typeName.contains("NUMERIC") || typeName.contains("DECIMAL") ||
+            typeName.contains("DATE") || typeName.contains("TIME") || typeName.contains("BOOL")
```
serge-rider: Date-time fields must be strings (they are strings now)

serge-rider (changes requested):

uslss:
verified

PR #35099, opened 2024-08-06
#35072 Create new tab if all tabs is pinned

### LINKED ISSUE #35072: Pinned Tab is being replaced by new query
### Description

Expected behaviour: 
Pinning a tab, then running another query would open result for new query in a seperate tab. 

Current behaviour: 
It no longer does this, after pinning the tab the new query replaces the previously pinned tab.

### DBeaver Version

Community Version 24.1.4.202408041450

### Operating System

macOS Sonoma 14.5

### Database and driver

Mysql 8.2.0 Driver

### Steps to reproduce

Run first query:
```select 'hi there' data from dual;```

Pin the result windows

Run second query:
```select 'hi again' data from dual;```

Pinned tab is replaced, no new tab is opened.

### Additional context

Was able to replicate this against database on local machine and remote via socks tunnel

### LINKED ISSUE #35135: "Execute SQL script" does not always work first time after upgrade to Version 24.1.4.202408041450
### Description

This is happening many times a day since the upgrade, but I have not been able to identify the exact sequence of events to duplicate it.  I write long scripts made up of Temp Table creation queries (each preceded by a DROP of that same Temp Table), which cascade one to the next.  So, I will often highlight a single (DROP + query that creates a Temp Table) script and will click the "Execute SQL script" icon.  Estimating about 1 out of every 3 or 4 times I do this, after waiting longer than I expected, I'll look at the screen and see the area is still highlighted, but there are no right pointing arrowheads on the left edge of the editor.  I'll click the "Execute SQL script" icon again, and execution will start.  I go through this process a hundred or more times every day, so it is very ingrained - highlight query then click Execute.

I hesitated to post since I have not been able to reliably duplicate it, and have been checking ISSUES to see any similar problems.  But since it has been happening so much and only after the most recent upgrade, I wanted to submit it anyway.  I will continue trying to identify some sequence of events that cause it.

### DBeaver Version

Version 24.1.4.202408041450

### Operating System

Windows 10 Enterprise  Version:22H2  Build:19045.4651

### Database and driver

Driver: Snowflake

### Steps to reproduce

I have not been able to reliably duplicate the steps, but it is happening many time every day.

### Additional context

_No response_

E1izabeth:


ShadelessFox (approved):

Matvey16:
Verified

PR #35110, opened 2024-08-07
#34969 Fix potential NPE

### LINKED ISSUE #34969: NPEx on refresh in DB Navigator
### Description

On refreshing in _Database Navigator_ pane I got dialog with:
```
Error refreshing node

Reason:
Cannot invoke "org.jkiss.dbeaver.model.impl.jdbc.JDBCRemoteInstance.initializeMetaContext(org.jkiss.dbeaver.model.runtime.DBRProgressMonitor)" because the return value of "org.jkiss.dbeaver.model.impl.jdbc.JDBCDataSource.getDefaultInstance()" is null
```

### DBeaver Version

Community 24.1.3.202407211303

### Operating System

GNU/Linux fc40

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

I guess it happened at
https://github.com/dbeaver/dbeaver/blob/0d326aa112032931ec134156d89caa757fa64c13/plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCDataSource.java#L481

But here
https://github.com/dbeaver/dbeaver/blob/0d326aa112032931ec134156d89caa757fa64c13/plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCDataSource.java#L440-L441
the possible `null` is considered.

E1izabeth:
Can't reproduce, but maybe it will work.
Let's test that connection is OK, metadata editing is OK.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCDataSource.java
+++ b/plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCDataSource.java
@@ -444,8 +444,7 @@ public abstract class JDBCDataSource extends AbstractDataSource
     }
 
     @Override
-    public void shutdown(@NotNull DBRProgressMonitor monitor)
-    {
+    public void shutdown(@NotNull DBRProgressMonitor monitor) {
         for (JDBCRemoteInstance instance : getAvailableInstances()) {
             Object exclusiveLock = instance.getExclusiveLock().acquireExclusiveLock();
             try {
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCDataSource.java
```diff
@@ -475,10 +474,12 @@ public void shutdown(@NotNull DBRProgressMonitor monitor)
     }
 
     @Override
-    public void initialize(@NotNull DBRProgressMonitor monitor)
-        throws DBException
-    {
-        getDefaultInstance().initializeMetaContext(monitor);
+    public void initialize(@NotNull DBRProgressMonitor monitor) throws DBException {
```
serge-rider, line 478: It changes default behavior. Are we sure?

E1izabeth: getAvailableInstances just returns getDefaultInstance copied to a list.
I suppose it doesn't change the behavior. If you're afraid it may broke smth, then I can add just null check, ok

ShadelessFox (approved):

Matvey16:
Verified

PR #35131, opened 2024-08-08
dbeaver/pro#3058 add vector datatype

Destrolaric:
Closes dbeaver/pro#3058


serge-rider:
Please verify for regression in all databases which support collections

Matvey16:
Inline edit of the elements of the array looks like it is allowed but any edit is auto-reverted

https://github.com/user-attachments/assets/9191dd0d-92e7-45a8-bc3e-542b8397d86a

Destrolaric:
@Matvey16 Might something more DBeaver wide, just tested on Mongo

PR #35141, opened 2024-08-09
#29309 Constraint columns ordering fix

### LINKED ISSUE #29309: DBeaver not reading SQLite composite foreign keys correctly
### Description

Since the release of DBeaver-Community 23.3.2 DBeaver does not read composite foreign keys in a SQLite database correctly - it worked perfectly before until version 23.3.1.
The problem persist until the latest version 24.0.3

The problem is that DBeaver seems to mixe up the columns referencing two table if it is defined with a composite foreign key. 

In the screenshot below one can see that the integer value **42** and the text string **RAHal2** turn up at the wrong place in the  command line generated by DBeaver (as indicated with the red arrow). Consequently, this results in an error message, since **RAHal2** is not put in parentheses and there interpreted as column - but in fact it is the field value of column  **ID_FSTkurz** (vise versa with **42** and **ID_Bef**)
> SQL Error [1]: [SQLITE_ERROR] SQL error or missing database (no such column: RAHal2)

Screenshot 2024-05-02 at 11 35 09

The composite screenshot with parts of the DDLs of both tables provides information on table definition with focus on primary and foreign keys.

Screenshot 2024-05-02 at 11 54 47

I aware that it might not be best practice to use composite primary/foreign keys, but for my purposes I deem it necessary.

### DBeaver Version

Community Edition 24.0.3.202404211624

### Operating System

macOS Sonoma 14.4.1

### Database and driver

1. SQLite
2. Driver: SQLite JDBC 3.45.3.0 (org.xerial:sqlite-jdbc:RELEASE 3.45.3.0)

### Steps to reproduce

_No response_

### Additional context

[error.log](https://github.com/dbeaver/dbeaver/files/15186820/error.log)

serge-rider:


uslss:
verified

PR #35163, opened 2024-08-12
#34865 Fix SQL output for SQL scripts. Fix output panel opening

### LINKED ISSUE #34865: Broken output when use Execute SQL script
### Description

After upgrading to 24.1.2, output in the Output window became broken. It may or may not display anything in the window. For example, this code on PostgreSQL may or may not produce a row in Output window:

DO $$ 
DECLARE
    v   TEXT = 'test';
BEGIN
   RAISE NOTICE '%', v;
END;$$;

My colleague on Windows could not reproduce the issue. I downgraded to 24.1.1 where the issue does not exist.

### DBeaver Version

24.1.2 for Apple Silicon

### Operating System

macOS 14.5

### Database and driver

Database is PostgreSQL 14.12
PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

Connect to the database and execute the following script, and look at the Output window:

DO $$ 
DECLARE
    v   TEXT = 'test';
BEGIN
   RAISE NOTICE '%', v;
END;$$;

### Additional context

_No response_

### LINKED ISSUE #35135: "Execute SQL script" does not always work first time after upgrade to Version 24.1.4.202408041450

serge-rider:


uslss:
verified. Output work checked for pg, mysql, sql server, oracle

PR #35168, opened 2024-08-12
dbeaver/pro#3181 Fix parsing of fully qualified column type name in create table statement

E1izabeth:
Errors should not appear in the log on table ddl opening


uslss:
verified

PR #35170, opened 2024-08-12
#22843 Improve error handling for project renaming

### LINKED ISSUE #22843: Renaming a project fails silently when the name includes an invalid caracther
### Description

When renaming a project, if the name contains an invalid character (like `/`) the action fails silently, simply not renaming the project. Recording below to illustrate the issue:

https://github.com/dbeaver/dbeaver/assets/26719984/c4d104f5-c811-40ab-bca7-69cca5a4ef21

I believe a user would expect a behavior similar to the Project Create Wizard, where it shows an error message and the 'confirm' button is disabled:

image

---
PS: Not sure if I should have filed this under bug or feature request as, technically, it's working fine. However, since an error message should be the expected behavior, I think it fits better under a bug report. Apologies if I misclassified the issue.

### DBeaver Version

Community Edition 23.3.4

### Operating System

macOS Sonoma 14.1

### Database and driver

_No response_

### Steps to reproduce

1. Rename a project with a name containing an invalid character (like `/`)
2. Observe that the name has not changed

### Additional context

_No response_

ivgag:
- Added validation for project renaming: the name cannot start with a dot or contain any characters that match the pattern (?U)[^/:'\"\\\\<>|?*]+.
- Display an error message to the user if project rename validation fails.

image

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
@@ -891,6 +891,13 @@ public class GeneralUtils {
     }
 
     public static void validateResourceName(String name) throws DBException {
+        if (!DBWorkbench.isDistributed() && !DBWorkbench.getPlatform().getApplication().isMultiuser()) {
+            return;
+        }
+        validateResourceNameUnconditionally(name);
+    }
+
+    public static void validateResourceNameUnconditionally(String name) throws DBException {
         if (name.startsWith(".")) {
             throw new DBException("Resource name '" + name + "' can't start with dot");
         }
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -891,9 +891,6 @@ public static UUID getMixedEndianUUIDFromBytes(byte[] bytes) {
     }
 
     public static void validateResourceName(String name) throws DBException {
-        if (!DBWorkbench.isDistributed() && !DBWorkbench.getPlatform().getApplication().isMultiuser()) {
```
serge-rider, line 894: Cannot just remove it. This function is used in many too places.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
ivgag: The function was used in only one place, so I moved the condition there.

E1izabeth (approved):

ShadelessFox (approved):

uslss:
verified

PR #35186, opened 2024-08-13
dbeaver/dbeaver#34297  add gis types to cache

### LINKED ISSUE #34297: Add support for "geometry" column type
The MariaDb database supports column type ["Geometry"](https://mariadb.com/kb/en/geometry-types/), which is not available in the "Data Type" selectbox.

Snímek obrazovky 2024-06-07 v 12 45 38

It will be nice to have the value in the selectbox values to know, that it is the correct value :)

It is possible to just fill in the "geometry" string the field and hit enter.

Destrolaric:
Closes dbeaver/dbeaver#34297


Matvey16:
Verified

PR #35190, opened 2024-08-13
#22948 Rename "Persist" buttons to "Execute" for clarity

### LINKED ISSUE #22948: "Persist" and "Save" mean the same thing (?)
In the English version, it looks like the words "Save" and "Persist" are used interchangeably throughout various menus, options, buttons, dialog boxes, etc. leading to a less than awesome UI experience.

I think using the same verb should help give a more consistent user interface to user.
I would personally prefer "persist", but "save" is fine as well.

**Additional context**
If I am wrong in assuming that "save" and "persist" are the same thing, can you kindly add an FAQ documentation describing the difference? Would be helpful.

ivgag:
Renamed "Persist" buttons to "Execute" for clarity in the following windows:

- Configuration of columns for exporting data from one table to another
- Table creation/modification window

image
image


E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
For consistency I suggest also changing presist to execute here, when we show a preview of a query to change data in the table

https://github.com/user-attachments/assets/bc64dca9-264d-4534-b0f3-25da4fa1ed36

E1izabeth:
Agree

PR #35198, opened 2024-08-14
[CUBRID] Remove 'test sql' . so that it is not used as a transaction and used 'isvalid()'.

### LINKED ISSUE #35120: [CUBRID] Issue with about transaction(log) in 'Keep Alive' when Autocommit is false.
If you set 'Keep Alive' and 'autocommit' to false, When not used for a long period of time, transaction(log) accumulate occurs. (Because every 'SQL Editor's connection' sends a 'Keep Alive')

Similar or identical issues are expected to occur in other databases.
(when database overriding 'testsql')

To this end, we offer the following suggestions:

1. Request to modify isConnectionAlive inside the JDBCUtils class.
Even if testSQL is defined, We want to execute 'conntion.isValid' first.

2. Let's create a PR by overriding 'checkContextAlive' just for Cubrid. This creates duplicate code.

3. We want to use 'select 1' , but Let's create a PR by delete 'testsql' defind again for Cubrid . No duplicate code is not created.

Would like advice.
Best Regard.

hwany7seo:
Issue : https://github.com/dbeaver/dbeaver/issues/35120
Description : Removed 'test sql(ping-sql)' to use 'isvalid()'.


E1izabeth:
Thank you for contribution!

ShadelessFox (approved):

E1izabeth (approved):

uslss:
@hwany7seo could you provide some info about what should be tested here?

also, please resolve conflicts

thanks in advance

hwany7seo:
@uslss 
The isvalid function only works on cubrid server and jdbc 10.2.1.8849 or higher.
However, if you do not apply the next commit, a query will be used and problems will occur when 'autocommit' is false.

In CUBRID Server and JDBC 10.2.1 or later
1. Set up Keep Alive in DBeaver.
2. Verify isvalid is executed in JDBCUtils.java code in DBeaver 
Or
You can check whether queries are executed through the log on the installed CUBRID server.
e.g.) 
cd $CUBRID
tail -f log/broker/sql_log/*

And the conflict  was resolved.

uslss:
@hwany7seo please also merge devel, dor now branch is out-of-date. thanks

hwany7seo:
@uslss 
Sorry, i made a mistake. 
Updated to the latest branch.

E1izabeth:
@hwany7seo please pull latest changes from devel again :)
Alternatively, but not mandatory: you can allow us editing this PR so we can do it ourselves.

E1izabeth:
P.S. you can just press this button in github web interface to pull latest changes from devel
image

uslss:
verified

PR #35210, opened 2024-08-14
#34692 Fix color rendering for shapes

### LINKED ISSUE #34692: GIS viewer "color" property
Sometimes it's hard to quick analyze geo data using GIS viewer in case you try to view various geo data (points, linstrings, polygons, etc.) at the same time. I'd like to be able to manually separate geo data in GIS viewer for convenient work with them.
2023-12-12 18_52_35-DBeaver GIS viewer - Leaflet

DBeaver CE version 21.2.3 and older allowed to display geometry objects from same column with various colors by specifying hex code in column named "color".
2024-07-08 10_05_30-DBeaver 24 1 2 - Script-266 sql
2023-12-12 18_38_04-DBeaver GIS viewer - Leaflet
In bugfix https://github.com/dbeaver/dbeaver/pull/14377 "colorizing" was considered a bug and now all geometies from same column displayed in default blue. "color" column is ignored in DBeaver CE version 21.2.4 and later.

I'd like to bring back the ability to set value for "color" column for colorize geo data as I please.

I can separate geo data in different columns to get different colors, but it's too inconvinient to work with many various geo data and I can't choose color that's I need.
2024-07-08 10_50_03-DBeaver 24 1 2 - Script-266 sql
2024-07-08 10_49_31-DBeaver 24 1 2 - Script-266 sql

**Additional context**
Related discussion: https://github.com/orgs/dbeaver/discussions/22157

ivgag:
- Added functionality to render shapes in their assigned color if a 'color' column is present in the result set.
- If no 'color' column is available, shapes are rendered based on their index in the row.

The first case
image

The second case
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.data.gis.view/src/org/jkiss/dbeaver/ui/gis/GeometryDataUtils.java
+++ b/plugins/org.jkiss.dbeaver.data.gis.view/src/org/jkiss/dbeaver/ui/gis/GeometryDataUtils.java
@@ -86,7 +86,22 @@ public class GeometryDataUtils {
         return result;
     }
 
-    public static void setGeometryProperties(@NotNull IResultSetController controller, @NotNull GeomAttrs geomAttrs, @NotNull DBGeometry geometry, @NotNull RGB geometryColor, @NotNull ResultSetRow row) {
+    /**
+     * Set geometry properties for the given geometry.
+     *
+     * @param controller result set controller to get the model.
+     * @param geomAttrs  geometry attributes.
+     * @param geometry   geometry to set properties.
+     * @param index      index of the geometry in the result set.
+     * @param row        row of the result set.
+     */
+    public static void setGeometryProperties(
+        @NotNull IResultSetController controller,
+        @NotNull GeomAttrs geomAttrs,
+        @NotNull DBGeometry geometry,
+        int index,
+        @NotNull ResultSetRow row
+    ) {
         final ResultSetModel model = controller.getModel();
         final Map<String, String> info = new LinkedHashMap<>();
         for (DBDAttributeBinding binding : geomAttrs.descAttrs) {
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.data.gis.view/src/org/jkiss/dbeaver/ui/gis/GeometryDataUtils.java
```diff
@@ -97,7 +112,9 @@ public static void setGeometryProperties(@NotNull IResultSetController controlle
         }
         final Map<String, Object> properties = new LinkedHashMap<>();
         properties.put("id", DBUtils.getObjectFullName(geomAttrs.geomAttr, DBPEvaluationContext.UI));
-        properties.put("color", String.format("#%02x%02x%02x", geometryColor.red, geometryColor.green, geometryColor.blue));
+        properties.put("color",
+            info.getOrDefault("color", rgbToHex(makeGeometryColor(index)))
+        );
```
ShadelessFox: Just a note that the default line width is 140 columns

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.data.gis.view/src/org/jkiss/dbeaver/ui/gis/GeometryDataUtils.java
```diff
@@ -124,14 +141,18 @@ public static void setGeometryProperties(@NotNull IResultSetController controlle
...
-    public static RGB makeGeometryColor(int index) {
+    private static RGB makeGeometryColor(int index) {
         if (index == 0) {
             return Display.getCurrent().getSystemColor(SWT.COLOR_BLUE).getRGB();
         } else {
             return UIColors.getColor(index).getRGB();
         }
     }
 
+    private static String rgbToHex(RGB rgb) {
```
ShadelessFox, line 151: We tend to specify `@NotNull` and `@Nullable` annotations throughout the codebase to simplify debugging and minimize risks of NPEs thanks to IDE inspections. Even if a method is private, it'd still require the programmer to read through it to figure out which arguments are nullable and whether it itself might return a null or not.

ivgag: Perhaps we could agree that everything should be non-nullable by default and only use `@NotNull` where necessary. What do you think?

ShadelessFox: Unfortunately, this can't be enforced by default. Perhaps if we migrated to another library, such as the Checker Framework, it would be possible.

E1izabeth (approved):

DariaMarkaryan:
Verified

PR #35233, opened 2024-08-16
dbeaver/pro#3215 Restrict local filesystem links open

serge-rider:


uslss:
verified

andrej1919:
Hello, I was using 'file:///path/some_file.pdf' links extensively in order to store links to local copy of important files in several databases. Would you be so kind as to add exception and allow links to local file opening? Or alternatively, this could be an option with default restriction but user could allow this behavior if he knows what is doing.

PR #35235, opened 2024-08-16
#34632 Add keywords for preference search

### LINKED ISSUE #34632: Add search over option's name in "Preferences" window
I'm always frustrated when I need to find some option that I remember the name of but don't remember its location in hierarchy of a tree in Preferences window.
In this case user has to click each item in a tree and search it manually.
The more settings are added in the settings, the more difficult it is to find the desired option.

You already have an excellent solution to fix the issue: search over Preferences.
But for some reason you don't make use of it fully: search over the certain option name do not give any results.

I suggest to search over text of all available options on a form.

Ideally, that newly added options should be searchable without adding its text manually to a search base to avoid the case when the option was added to a form but not available for searching.

ivgag:
Added keywords for the preference search: reopen, open, Hippie, autosave.

Unfortunately, at the moment, pages don't have their own lists of tags. We have tag groups, and several pages can be linked to the same group, so it's challenging to expand these groups as it leads to unrelated pages appearing in the search results. However, reworking the tag groups right now doesn't make sense because we will automate collecting keywords.


E1izabeth:
Please amend commit message and add # before number. This way commit will be linked with issue automatically

uslss:
verified

PR #35270, opened 2024-08-20
dbeaver/dbeaver#34909 Fix active catalog name

### LINKED ISSUE #34909: Incorrect database is chosen when changing connection
### Description

When I choose another connection from a list of connections, application doesn't choose "active catalog/schema" of a chosen connection, instead it remains the same.
Finally, I end up with connecting to wrong database.

### DBeaver Version

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

- create 5 databases and 5 connections, each connection - connects to its own database (pg1 = db1, pg2 = db2, etc..)
- in proprerties of each connection, set option "Show all databases" = true
- close and open the app, so that postgres databases are grouped in one sub-list of items "PostgreSQL" when pressing a small triangle with connection list 
- being connected to pg1 connection, try to change current connection to pg2, how I do it in the video

expected: "Active catalog/schema" is set to "public@db2"
actual: "Active catalog/schema" is set to "public@db1"

https://github.com/user-attachments/assets/5bc5a255-2d5e-4219-91b2-0f67b4aa13d9

HocKu7:


plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/actions/datasource/SelectActiveSchemaHandler.java
```diff
@@ -183,11 +183,17 @@ public void propertyChanged(Object source, int propId) {
                 if (contextDefaults != null) {
                     DBSCatalog defaultCatalog = contextDefaults.getDefaultCatalog();
                     DBSSchema defaultSchema = contextDefaults.getDefaultSchema();
+                    String databaseNameFromConnectionConfiguration = executionContext.getDataSource().getContainer()
```
serge-rider: Looks like conceptual behavior change.
Why take database name from configuration if it differs from actual active catalog?

serge-rider (changes requested):

uslss:
verified

PR #35273, opened 2024-08-20
#34192 Nested query condition generated fix

### LINKED ISSUE #34192: DuckDB filtering on columns with spaces doesn't work
### Description

When I have this sample query in Dbeaver with DuckDB :memory: setting:

`SELECT 'test' AS "test value"`

And I want to filter on this column, via:
image

I get this error:

`SQL Error: java.sql.SQLException: Parser Error: syntax error at or near "value"`

Probably because this SQL is generated automatically:

```
SELECT * FROM (
SELECT 'test' AS "test value"
) z_q WHERE test value = 'test'
```

If I copy this SQL to a new tab and change it as follows (enclose column name with double quotes), it does work:

```
SELECT * FROM (
SELECT 'test' AS "test value"
) z_q WHERE "test value" = 'test'
```

The request is to have filter queries and group by panel queries automatically enclosed with double quotes, as per the DuckDB standard: https://duckdb.org/docs/sql/keywords_and_identifiers.html

A possibly related issue might be Grouping on columns with a `/` (slash) in the column name. However, if you want I can make a separate ticket for this.

Perhaps this bug/ request is related to: #23020 and/ or #24122 

### DBeaver Version

Community edition Version 24.0.1.202403241413

### Operating System

Windows 11 Pro x64

### Database and driver

duckdb_jdbc-0.10.3.jar with `:memory:` as Path

### Steps to reproduce

Create :memory: connection to DuckDB
Execute MWE query as described above
Attempt to filter on this value, as per the printscreen

### Additional context

_No response_

serge-rider:


uslss:
verified

PR #35294, opened 2024-08-21
dbeaver/pro#3252 Show warning instead of error if table used in DDL doesn't exist

E1izabeth:
For example, here should be only warnings, no errors and no red highlighting
image

```
CREATE TABLE ArtWithRowid2 (
    [ArtistId] INTEGER NOT NULL PRIMARY KEY ,
    [Name] NVARCHAR(120),
    [something] int  REFERENCES Artist (address)
);

ALTER TABLE Album ADD CONSTRAINT dfk FOREIGN KEY (AlbumId) REFERENCES Artist (address) NOT VALID; 

```


ShadelessFox (approved):

DariaMarkaryan:
Verified

PR #35298, opened 2024-08-21
Update JDBCTable.java to add `AS` before table alias

jovezhong:
A very simple change
* Currently when a user clicks the `Data` tab for a table in DBeaver, it will generate a SQL like `SELECT x.* FROM my_table x`
* This PR will add `AS` so the SQL will be `SELECT x.* FROM my_table AS x`
* The reason for making such changes is that some databases enforce the `AS` keyword for making alias, sometimes the organizations also apply some SQL auditing or block SQL without meeting the standard. Having the `AS` will make the SQL a bit easier to read.

Not a big deal, but very low risk.

Reference

https://github.com/brooklyn-data/co/blob/main/sql_style_guide.md
Always use the `as` keyword when aliasing columns, expressions, and tables.
```sql
/* Good */
select count(*) as customers_count
from customers

/* Bad */
select count(*) customers_count
from customers
```

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTable.java
+++ b/plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTable.java
@@ -162,7 +162,7 @@ public abstract class JDBCTable<DATASOURCE extends DBPDataSource, CONTAINER exte
         appendSelectSource(monitor, query, tableAlias, rowIdAttribute);
         query.append(" FROM ").append(getTableName());
         if (tableAlias != null) {
-            query.append(" ").append(tableAlias); //$NON-NLS-1$
+            query.append(" AS ").append(tableAlias); //$NON-NLS-1$
         }
         appendExtraSelectParameters(query);
         SQLUtils.appendQueryConditions(dataSource, query, tableAlias, dataFilter);


jovezhong:
Can you merge it? Very simple code change.

E1izabeth:
Thank you for contribution! We'll review your PR and let you know, if some changes are needed.

ShadelessFox (approved):
While the `AS` keyword is defined in the standard, some databases may not expect it. Let's hope we won't encounter such databases.

E1izabeth (approved):

jovezhong:
> While the `AS` keyword is defined in the standard, some databases may not expect it. Let's hope we won't encounter such databases.

Unfortunately in the database our team is working on (https://github.com/timeplus-io/proton) `select * from a b` will fail. Has to be `select * from a as b` This could be improved later but we decided to enforce the `AS` to make query analayzer a bit easier such as `select .. from a left latest join b`  https://docs.timeplus.com/joins

I have rebased the PR. It will be great to merge this. Thank you🙏

ShadelessFox:
It would be nicer to have something like `org.jkiss.dbeaver.model.sql.SQLDialect#supportsAsInAliases`. That property can also be exposed in the generic driver to allow changing behavior for custom drivers without digging into the code.

Matvey16:
That change seem dangerous. There are databases that on the countrary don't support this syntax for aliases, so making it default for everything doesn't look like a good idea

jovezhong:
All make sense. Since DBeaver has been generating such query for a long time without too many complains, I guess many databases support the alias without `AS` keyword.

I agree it'll be great if each SQLDialect can set this setting and the framework code as `JDBCTable.java` can generate SQL with or without `AS` 
> It would be nicer to have something like org.jkiss.dbeaver.model.sql.SQLDialect#supportsAsInAliases. That property can also be exposed in the generic driver to allow changing behavior for custom drivers without digging into the code.

benjGam (commented):
Yup

E1izabeth:
This PR looks stale. So, I'm closing it.
We can't merge this change as is, because AS in from clause is not supported by some databases, for example, Oracle
image

PR #35316, opened 2024-08-22
#34996 Fix negative date/time handling for AWS Aurora/MySQL

### LINKED ISSUE #34996: DBeaver's Time formats cannot handle negative times
### Description

When calculating the difference between two times, it is possible to produce a negative result. For example, in MySQL you can use the `TIMEDIFF` function to calculate the first timestamp minus the second one. If the first timestamp is less than the second, the result is a negative amount of time, but DBeaver renders it as a positive number.

DBeaver renders the result of this query as 00:00:05:
```sql
SELECT timediff('2024-07-26 00:00:45', '2024-07-26 00:00:50');
```
However, the actual result is -00:00:05 because 00:00:45 is less than 00:00:50. You can see the actual result if you run the subtraction on the MySQL command line, or if you use `TIME_TO_SEC`:
```sql
SELECT time_to_sec(timediff('2024-07-26 00:00:45', '2024-07-26 00:00:50'));
```
which yields a result of -5 (and correctly renders as -5).

In contrast, in the following queries the first timestamp is larger than the second:
```sql
SELECT timediff('2024-07-26 00:00:45', '2024-07-26 00:00:00');
SELECT time_to_sec(timediff('2024-07-26 00:00:45', '2024-07-26 00:00:00'));
```
DBeaver renders their results correctly as 00:00:45 and 45, respectively.

I'm guessing this might be a limitation of Java's time formatting library, because when I went to Window>Preferences>Editors>Data Editor>Data Formats to try to update the time format and then went to [Data format patterns documentation](https://dbeaver.com/docs/dbeaver/Managing-Data-Formats/#time), there was no mention of a way to display a negative time either there or in the Java [docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/text/SimpleDateFormat.html) it refers to.

### DBeaver Version

Community Edition 24.1.2.202407071610

### Operating System

Linux Ubuntu 22.04.4 LTS

### Database and driver

Database: AWS Aurora emulting MySQL 8.0: 8.0.mysql_aurora.3.05.2
Driver: MySQL (current, for MySQL 8 and later)

### Steps to reproduce

Run the following query:
```sql
SELECT timediff('2024-07-26 00:00:45', '2024-07-26 00:00:00');
```
Note the result. The correct result is "-00:00:05". A positive number, "00:00:05", is incorrect. You can verify by surrounding the `timediff()` with `time_to_sec()` and noting whether it produces a positive or negative number of seconds:
```sql
SELECT time_to_sec(timediff('2024-07-26 00:00:45', '2024-07-26 00:00:00'));
```

This query might also be useful for testing:
```sql
SELECT time('-00:00:05');
```
which you an also verify using TIME_TO_SEC:
```sql
SELECT TIME_TO_SEC(time('-00:00:05'));
```
Note that `SELECT '-00:00:05';` without the cast to TIME, prints a negative time, but that's probably because it's rendering as a string, not as a time, so it's probably not a valid test.

### Additional context

_No response_

ShadelessFox:
This PR affects MySQL databases but shouldn't cause any trouble. Only works when date/time formatting is **turned off**.


E1izabeth (approved):

Matvey16:
Verified

Matvey16:
Verified

PR #35318, opened 2024-08-22
dbeaver/pro#3075 Fix encrypted window warning appears

HocKu7:
There is no repeatedly warning windows appearing when you are trying to open an encrypted project in CE


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNProjectConstant.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.navigator;
+
+public class DBNProjectConstant {
```
ShadelessFox:
```suggestion
public class DBNProjectConstants {
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNProjectConstant.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public static final String USER_DECLINE_DECRYPTION = "userDeclineProjectCredentialsDecryption";
```
ShadelessFox:
```suggestion
    public static final String PROP_USER_DECLINE_DECRYPTION = "userDeclineProjectCredentialsDecryption";
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNProject.java
```diff
@@ -175,6 +175,7 @@ protected DBNNode[] readChildNodes(DBRProgressMonitor monitor) throws DBExceptio
             project.ensureOpen();
         }
 
+        project.setProjectProperty(DBNProjectConstant.USER_DECLINE_DECRYPTION, false);
```
ShadelessFox: You don't need to set it. It will be implicitly `false`.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNProjectConstant.java
ShadelessFox: Check my comment in DBNProject. This way, you don't even need to expose that constant. Store it inside the serializer class.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNProject.java
HocKu7: I need to set it to false in case the user would like to open the project manually after declining. That is why I need to set it

ShadelessFox: Ah, so it resets it. I see. Okay.

E1izabeth (approved):

E1izabeth:
@HocKu7 please resolve conflicts

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/DatabaseNavigatorTree.java
```diff
@@ -369,6 +369,12 @@ protected void handleTreeExpand(TreeEvent event) {
                 // Disable redraw during expand (its blinking)
                 getTree().setRedraw(false);
                 try {
+                    if (event.getSource() instanceof Tree tree
+                        && tree.getItemCount() == 1
+                        && tree.getItem(0).getData() instanceof DBNProject dbnProject) {
```
ShadelessFox:
```suggestion
                    if (event.item != null && event.item.getData() instanceof DBNProject dbnProject) {
```

ShadelessFox (approved):

MashaKorax:
verified

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPProject.java
```diff
@@ -163,4 +163,6 @@ default SMSession getWorkspaceSession() {
      */
     void dispose();
 
+    void setRuntimeProperty(String key, String value);
+    String getRuntimeProperty(String key);
```
ShadelessFox: Please specify annotations and fix formatting

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPProject.java
```diff
@@ -163,4 +163,6 @@ default SMSession getWorkspaceSession() {
      */
     void dispose();
 
+    void setRuntimeProperty(String key, String value);
```
serge-rider: Please 
- declare getter before setter
- write some comment for getter (what is runtime property)

serge-rider (changes requested):

PR #35319, opened 2024-08-22
dbeaver/pro#3164 Improve create and drop queries handling

E1izabeth:
- Fix error from issue
- Handle CREATE TEMPORARY TABLE statement


uslss:
verified

PR #35326, opened 2024-08-23
#34797 Fix completion synchronization with parsing

### LINKED ISSUE #34797: New autocompletion replaces words that it is not supposed to
### Description

I don't know all the cases, but in some the proposed word would be added while deleting words that come before it. In my case it happened with `AND` keyword. THis also happens regardless if the property to replace initial word is checked or not.

### DBeaver Version

24.1.3

### Operating System

Any

### Database and driver

PostgreSQL

### Steps to reproduce

1. Start typing a query with a `where` clause
2. Type a condition and then type `and`
3. Call autocomplete after a space and select one of the proposals

**Expected:** Text is added after the caret
**Actual:** text is added inside the `and` keyword

### Additional context

https://github.com/user-attachments/assets/413b25d3-b53a-4f57-b0f4-3aba5012e6e6

### LINKED ISSUE #34845: New autocomplete mode sync issue
### Description

By the way, in this video I do not call suggestions by Ctrl + Space.
It appears only on Space symbol automatically.

https://github.com/user-attachments/assets/1d75ed2e-fdc8-482c-87f4-85003a600933

Please also check this query on dvdrental database (Ctrl + Space after "on"):

```
select * from actor as a
join film_actor as fa on 
```

### DBeaver Version

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.2

E1izabeth:
Fixes #34845 and #34797

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/SQLBackgroundParsingJob.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/SQLBackgroundParsingJob.java
@@ -54,6 +54,8 @@ import java.util.ArrayDeque;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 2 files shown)


uslss:
verified

PR #35347, opened 2024-08-26
dbeaver/pro#3015 Resource leaks fix (driver selector leaks whenever y…

serge-rider:
…ou filter by database name)


uslss:
verified

PR #35350, opened 2024-08-26
dbeaver/pro#3192 Flipped splash on macOS 15 fix

ShadelessFox:


E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/RuntimeUtils.java
```diff
@@ -418,12 +415,21 @@ public static boolean isOSArchAMD64() {
...
+     *
+     * <p><b>Note:</b> this method was only tested to be working on macOS. Please test on other platforms.
+     *
+     * @return version of the operating system
      */
-    public static boolean isMacOsSomona() {
-        return isMacOS() && System.getProperty("os.version").startsWith("14");
+    @NotNull
+    public static Version getOSVersion() {
+        return new Version(System.getProperty("os.version"));
```
serge-rider: new Version may throw exception. This may block app start.

serge-rider: e.g. is it possible that OS version is 11.alpha ?

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -251,8 +251,9 @@ public Object start(IApplicationContext context) {
         // Write version info
         writeWorkspaceInfo();
 
-        // Update splash. Do it AFTER platform startup because platform may initiate some splash shell interactions
-        if (!RuntimeUtils.isMacOsSomona()) {
+        // https://github.com/eclipse-platform/eclipse.platform.swt/issues/772
+        if (!RuntimeUtils.isMacOS() || !RuntimeUtils.isOSVersionAtLeast(14, 0, 0)) {
```
serge-rider, line 255: It will break splash for Linuxes with version higher than 14

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/RuntimeUtils.java
ShadelessFox: As far as I am concerned, it always contains a string in the form of `major.minor[.micro]`.

I added a disclaimer about its reliability on different OSes.

This is why I wanted to use a more trustworthy source of that information, such as SWT Cocoa's OS class but we can't have a bundle dependency on it.

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
ShadelessFox: No? It seems fine. It's a negation of `RuntimeUtils.isMacOS() && RuntimeUtils.isOSVersionAtLeast(14, 0, 0)`

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/RuntimeUtils.java
serge-rider: We cannot be 100% sure. Lets handle exception at least

PR #35367, opened 2024-08-27
dbeaver/pro#3273 New autocompletion offset syncronisation

E1izabeth:
We need to work with Positions under syncExec to not interfere with eclipse reading and updating them


Matvey16:
Verified

PR #35385, opened 2024-08-29
dbeaver/pro#3257 Syntax tree to model mapping refactoring

E1izabeth:
- Now, we don't throw exceptions if something is not correctly parsed (see case from the issue).
- If we haven't parsed some part of the query, we build an incomplete model (not sure how to test it)
- If the table is not found, we don't highlight columns and don't show errors for them - only for the table.


uslss:
verified

PR #35393, opened 2024-08-29
#35040 Fix refreshing schema

### LINKED ISSUE #35040: SQL command use database or use schema does not change active datatabase in Dbeaver 24.1.3
### Description

Regression bug
In Snowflake SQL you have commands like __use database__ or __use schema__ to switch to a certain database or a certain schema. We used to use an old version of Dbeaver (23.1.0) and in this version, when you do __use database__, it switches the database for the session __and__ it switches the active database for DBeaver, which is shown in the menu bar. 

Now we have DBeaver version 24.1.3 and the behaviour changed. A __use database__ still switches the database for the session, but not the active database for DBeaver. This means, the SQLs you type and execute will use the database set for the session (with __use database__). However, it is very confusing to see another database (which is the default database) at the top of the menu bar as the active database. You are never sure which database you are working in.

When you use F4 to see the definition of what's under the cursor, it's using the database shown in the menu bar and __not__ the one set for the session with __use database__.

### DBeaver Version

24.1.3

### Operating System

Windows 10

### Database and driver

Snowflake

### Steps to reproduce

_No response_

### Additional context

_No response_

ivgag:
- Refresh schema only after the last query in the script is finished or an error occurs.
- Refresh schema regardless of isExtraMetadataReadEnabled status.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/SqlJobResult.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/SqlJobResult.java
@@ -17,6 +17,7 @@
 package org.jkiss.dbeaver.model.sql;
 
 public enum SqlJobResult {
+    NOT_RUN,
     SUCCESS,
     PARTIAL_SUCCESS,
     FAILURE,
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/SqlJobResult.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/SqlJobResult.java
@@ -16,23 +16,8 @@
  */
 package org.jkiss.dbeaver.model.sql;
 
-import org.jkiss.dbeaver.model.exec.DBCSession;
-import org.jkiss.dbeaver.model.exec.DBCStatistics;
-
-/**
- * SQLQueryListener
- *
- * @author Serge Rider
- */
-public interface SQLQueryListener
-{
-    void onStartScript();
-
-    void onStartQuery(DBCSession session, SQLQuery query);
-
-    void onEndQuery(DBCSession session, SQLQueryResult result, DBCStatistics statistics);
-
-    void onEndScript(DBCStatistics statistics, boolean hasErrors);
-
-    void onEndSqlJob(DBCSession session, SqlJobResult result);
+public enum SqlJobResult {
+    SUCCESS,
+    PARTIAL_SUCCESS,
+    FAILURE,
 }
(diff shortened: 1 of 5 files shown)


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -4881,10 +4883,7 @@ private void refreshContextDefaults(DBCSession session, SQLQueryResult result) {
             final DBCExecutionContext executionContext = getExecutionContext();
             if (executionContext != null && session != null) {
                 // Refresh active object
-                if ((result == null || !result.hasError()) &&
```
serge-rider, line 4884: Why remove this condition? It mustn't refresh context if query faled.

ivgag: The condition should be removed to cover a case when we have a script and the last query in the script failed.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/SqlJobResult.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.sql;
+
+public enum SqlJobResult {
+    NOT_RUN,
+    SUCCESS,
+    PARTIAL_SUCCESS,
+    FAILURE,
```
ShadelessFox: Just `SUCCESS` and `PARTIAL_SUCCESS` are used. Do we really need to differentiate between these four states? Maybe it can be reduced down to just two: either success or not?

ivgag: Which status should we use if a script was partially executed and then encountered an error?

uslss:
verified

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/SqlJobResult.java
ShadelessFox: But you don't handle these. Just the ones mentioned are used.

ivgag: You’re right! I haven’t come up with better names to reduce it to just two options yet, and I prefer not to use the SUCCESS status for cases where errors occur. I’ve removed `NOT_RUN` to streamline things a bit.

ShadelessFox: And still, the difference between `SUCCESS` and `PARTIAL_SUCCESS` is not clear - they both are handled the same way. It seems that you want to use a `boolean` flag to indicate whether the execution was successful or not.

ivgag: Yes, they are handled the same way, but it makes the code more readable. I’m open to suggestions for naming that boolean to maintain a consistent level of abstraction between SQLQueryJob and SQLQueryListener.

PR #35413, opened 2024-08-30
dbeaver/pro#3198 Refactor `DBUtils#getAttributeValue`

ShadelessFox:
Fixed:
- Complex collection expanding (arrays of arrays, struct of arrays, arrays of structs, etc)
- The first collection element is now showing instead of the inline collection presentation

Known issues/to be done in the following PRs:
- The value panel sometimes shows actual collection objects instead of the deepest value
- Can't expand a collection of booleans by clicking on the expander inside the cell
- Can't expand a collection/struct in the record mode in some cases


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -722,78 +724,56 @@ public static Object getAttributeValue(
             return null;
         }
 
+        Deque<DBDAttributeBinding> pendingAttributes = new ArrayDeque<>();
+        Deque<Integer> pendingIndices = new ArrayDeque<>();
```
ShadelessFox, line 728: This does not introduce a substantial performance penalty but simplifies the code. Thanks to the cache, boxed integers won't be allocated if indices are in the range of 0 to 127.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -722,78 +724,56 @@ public static Object getAttributeValue(
...
 
+        Deque<DBDAttributeBinding> pendingAttributes = new ArrayDeque<>(depth);
+        for (int i = 0; i < depth; i++) {
+            pendingAttributes.offer(Objects.requireNonNull(attribute.getParent(depth - i - 1)));
+        }
+
+        Deque<Integer> pendingIndices = new ArrayDeque<>(nestedIndexes != null ? nestedIndexes.length : 0);
+        for (int i = 0; nestedIndexes != null && i < nestedIndexes.length; i++) {
+            pendingIndices.offer(nestedIndexes[i]);
+        }
```
ShadelessFox: This does not introduce a substantial performance penalty but simplifies the code. Thanks to the cache, boxed integers won't be allocated if indices are in the range of 0 to 127.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
serge-rider: Can we use just array instead of deque? WE always know the size.

ShadelessFox: Yes, but this way I can go with manual indexing.

I'm using deque because it's simpler to work with without badly hitting the performance.

PR #35434, opened 2024-09-02
#34747 Show a confirmation when trying to save SQL console

### LINKED ISSUE #34747: [Question] How can I make console unsavable?
I have the urge to press ctrl+s ever now and then because of muscle memory but when I open a console tab I explicitly don't want it to be saved. Especially automatically becoming a script when I opened it as a console.

It would be nice to have an option to disable saving on console. Even nicer if it also disable the pesky "*" indicator of unsaved changes, obviously it has unsaved changes they were never meant to be saved! Also I think it's one of the biggest triggers of me unconsciously clicking ctrl+s.

Another option could be to show a window asking If one really wants to transform a console to a script. The unpleasant need to close it every 5 seconds eventually would reverse pavlov my brain into not clicking ctrl+s in dbeaver I think.  

I just want a nice folder with meaningful scripts that I can rely on not a bunch of 'script-1, script-2, script-3, script-4, script-5..., script-21937' where I can never find what I need...

ShadelessFox:
A new confirmation was added that will be shown whenever the user attempts to save an SQL console:

image

If the remembered choice is "No", the dirty mark will not appear in the tab name.

Can be configured in the preferences: **User Interface -> Confirmations**.


E1izabeth:
I would expect to see the checkbox on the one line with question. This way
image

ShadelessFox:
Well, that's what Eclipse's `MessageDialogWithToggle` looks like.

E1izabeth:
OK

E1izabeth (approved):

HocKu7 (approved):

MashaKorax:
Nothing happens when press Yes to save console as a script

https://github.com/user-attachments/assets/0be851e8-7269-4592-b95a-b98cd9309b62

Matvey16:
Verified

PR #35438, opened 2024-09-02
#35380 Fix query processor selection, when processor has multiple tabs and some of them are unpinned

### LINKED ISSUE #35380: Pinned tabs are overwritten if statistics tab is present
### Description

The pinned tab overwrite issue rears its ugly head again.

If the "Show Statistics Tab" setting is set to "Always" or otherwise appears, it too must be pinned or the pinned tab is overwritten.

### DBeaver Version

24.1.5

### Operating System

macOS

### Database and driver

_No response_

### Steps to reproduce

1. Set "Show statistics tab" to "Always"
2. Run a query. Two tabs will be generated: One for the query results, one for the statistics
3. Pin the results tab
4. Run a different query. The results tab is overwritten, even though it is pinned.
5. Pin the statistics tab
6. Run the first query. A new results tab is created

### Additional context

Related to #35072 #35173  #35203  #35238

E1izabeth:
Please check this issue itself and all other mentioned issues, that we fixed before.


HocKu7 (approved):

ivgag (approved):

Matvey16:
If you pin a tab then both the table and the statistics tab won't disappear after running another query, but unpinning this tab later won't change the bahaviour and both the table and statistics would stay until closed manually, I don't think it is an expected behaviour.

E1izabeth:
@Matvey16 could you provide specific steps to reproduce what you described? For now I can't find any difference with 23.3.0 version. Probably, I miss some specific case

E1izabeth:
https://github.com/user-attachments/assets/6140e0e1-85ba-4be7-92df-16e0f3e2e27f

Matvey16:
Verified

PR #35452, opened 2024-09-03
dbeaver/pro#3300 Pick the default router if chosen one is not available

ShadelessFox:


E1izabeth (approved):

arhayka:
Verified: blank screen in devel, normal ERD in this branch

Issue #35392, opened 2024-08-29
DB CE 24.1.5 Win32.x86_64 SWTError: No more handles


cowbert:
### Description
This is on the portable/ZIP Windows version of CE 24.1.5

With multiple windows, but fewer than 10 SQL Editor tabs per window, switching tabs intermittently throws org.eclipse.swt.SWTError: No more handles.

Here is a stacktrace from the log for the initial exception. This then drives Unhandled Loop Exceptions throughout the rest of the application requiring application shutdown. 

We're also using the following SQLEditor config lines in org.jkiss.dbeaver.core.prefs:
```
SQLEditor.ContentAssistant.auto.activation.enable=false
SQLEditor.ContentAssistant.auto.keystrokes.activation=false
SQLEditor.ContentAssistant.autocompletion.tab=false
SQLEditor.ContentAssistant.insert.single.proposal=false
SQLEditor.Highlighting.advanced.enable=false
SQLEditor.autoSaveOnClose=true
SQLEditor.format.activeQuery=false
```

Stacktrace sample:
```
!ENTRY org.eclipse.e4.ui.workbench 4 0 2024-08-29 10:38:07.674
!MESSAGE No more handles
!STACK 0
org.eclipse.swt.SWTError: No more handles
	at org.eclipse.swt.SWT.error(SWT.java:4948)
	at org.eclipse.swt.SWT.error(SWT.java:4837)
	at org.eclipse.swt.SWT.error(SWT.java:4808)
	at org.eclipse.swt.widgets.Widget.error(Widget.java:488)
	at org.eclipse.swt.widgets.Control.createHandle(Control.java:674)
	at org.eclipse.swt.widgets.Scrollable.createHandle(Scrollable.java:145)
	at org.eclipse.swt.widgets.Composite.createHandle(Composite.java:298)
	at org.eclipse.swt.widgets.Control.createWidget(Control.java:700)
	at org.eclipse.swt.widgets.Scrollable.createWidget(Scrollable.java:160)
	at org.eclipse.swt.widgets.Control.<init>(Control.java:121)
	at org.eclipse.swt.widgets.Scrollable.<init>(Scrollable.java:85)
	at org.eclipse.swt.widgets.Composite.<init>(Composite.java:103)
	at org.eclipse.swt.widgets.Canvas.<init>(Canvas.java:81)
	at org.jkiss.dbeaver.ui.controls.VerticalFolder.<init>(VerticalFolder.java:43)
[...]

Issue #18197, opened 2022-11-11
DB CE 22.2.4 Linux - Unhandled event loop exception: No more handles


mj162:
### Description

**1. Expected Behaviour**
1. Install from dbeaver-ce-22.2.4-linux.gtk.x86_64-nojdk.tar.gz
2. Launch dbeaver using existing java version "11.0.9" 2020-10-20 LTS
3. File > Open
4. Able to load source and continue working on it

**2. Observed Behaviour**
1. Install from dbeaver-ce-22.2.4-linux.gtk.x86_64-nojdk.tar.gz
2. Launch dbeaver using existing java version "11.0.9" 2020-10-20 LTS
3. File > Open
Alert: Problem Occurred: Unhandled event loop exception: No more handles (or.eclipse.ui, org.jkiss.dbeaver.model)

***3. Java Paths and Arguments***
```bash
JVM terminated. Exit code=127
/media/shared/opt/jdk11/x86_64/bin/java
-XX:+IgnoreUnrecognizedVMOptions
--add-modules=ALL-SYSTEM
--add-opens=java.base/java.nio=ALL-UNNAMED
-Dosgi.requiredJavaVersion=11
-Xms64m
-Xmx1024m
-jar /tmp/tmpfs/media/shared/opt/dbeaver/dbeaver-ce-22.2.4-linux.gtk.x86_64-nojdk//plugins/org.eclipse.equinox.launcher_1.6.400.v20210924-0641.jar
-os linux
-ws gtk
-arch x86_64
-showsplash
-launcher /tmp/tmpfs/media/shared/opt/dbeaver/dbeaver-ce-22.2.4-linux.gtk.x86_64-nojdk/dbeaver
-name Dbeaver
--launcher.library /tmp/tmpfs/media/shared/opt/dbeaver/dbeaver-ce-22.2.4-linux.gtk.x86_64-nojdk//plugins/org.eclipse.equinox.launcher.gtk.linux.x86_64_1.2.600.v20220720-1916/eclipse_11701.so
-startup /tmp/tmpfs/media/shared/opt/dbeaver/dbeaver-ce-22.2.4-linux.gtk.x86_64-nojdk//plugins/org.eclipse.equinox.launcher_1.6.400.v20210924-0641.jar
--launcher.overrideVmargs
-exitdata 1820050
-vm /media/shared/opt/jdk11/x86_64/bin/java
-vmargs
-XX:+IgnoreUnrecognizedVMOptions
[...]

PR #35482, opened 2024-09-05
Add resource leak detection

bt90:
Register a resource leak detector via [Resource.setNonDisposeHandler(Consumer<Error>)](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fswt%2Fgraphics%2FResource.html)

Any fonts, images, etc. that aren't disposed of will trigger a warning in the log with a stack trace showing where they were originally created.


bt90:
This should help in finding the culprit for issues such as #35392 or #18197.

ask9:
This one also: #35493
Here is a full list of similar issues:
https://github.com/dbeaver/dbeaver/issues?q=is%3Aissue+is%3Aopen+No+more+handles+sort%3Aupdated-desc

E1izabeth:
Thank you for contribution

E1izabeth:
As far as I know, a similar behavior can be reached by adding the `-Dorg.eclipse.swt.graphics.Resource.reportNonDisposed=true` flag when starting DBeaver.

bt90:
This would also result in a Consumer being registered, but with the drawback that it doesn't use any kind of logging framework and will only print to stderr.

The overhead of having this enabled by default is limited to creating `Error` instances for each resource. I doubt that this has any measurable effect in production, while giving users a clear log message to create an issue for.

ShadelessFox (approved):

E1izabeth (approved):

MashaKorax:
verified

PR #35486, opened 2024-09-05
#34246 Move object editor toolbar to the left

### LINKED ISSUE #34246: Can the refresh button be placed in the same place
Double click on the table name to view the refresh of the table property bar in the bottom right corner, but the refresh of the data bar is in the bottom left corner

### data bar
image

### property bar
image

ShadelessFox:
image

Must be tested in all object editors, such as folders, entities, DDL, ERD, and maybe some other places.


Matvey16:
Verified

PR #35503, opened 2024-09-06
dbeaver#34917 Remove splash for console using

### LINKED ISSUE #34917: Splash screen overlaps error message and doesn't disappear until closing the error message which is not visible
### Description

If some error is shown due to opening a script, the splash window can overlap the error message and block the user.
It is very confusing for a user because he could miss the error message and wait for the splash screen to close (infinitely 🤣).

### DBeaver Version

### Operating System

Windows 10

### Steps to reproduce

- open dbeaver(!)
- use command line (cmd.exe) to run dbeaver with passing an argument (file path to a script, which actually not exists)

expected: error message is shown to a user
actual: error message is behind the splash screen, so user is unable to see the message and react

By the way, I'm not sure if splash screen should be shown at all in this case.
For example, SSMS doesn't show splash screen if main window has already opened.

Here on the video, first of all I show that the Dbeaver has already running before executing a command in cmd.exe.

https://github.com/user-attachments/assets/6d7bbc5a-e2a3-4149-8da0-71b26a4185f4

HocKu7:
The issue appears when we try to start another instance of DBeaver with an error path, so an error appears in the first instance, and the second one gets blocked. Now, It will not block the second one

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/rpc/DBeaverInstanceServer.java
+++ b/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/rpc/DBeaverInstanceServer.java
@@ -164,7 +164,7 @@ public class DBeaverInstanceServer implements IInstanceController {
 
         final IWorkbenchWindow window = UIUtils.getActiveWorkbenchWindow();
         final Shell shell = window.getShell();
-        UIUtils.syncExec(() -> {
+        UIUtils.asyncExec(() -> {
             for (String filePath : fileNames) {
                 File file = new File(filePath);
                 if (file.exists()) {


plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/DBeaverLauncher.java
```diff
@@ -2262,7 +2262,8 @@ private InputStream getStream(URL location) throws IOException {
     private void handleSplash(URL[] defaultPath) {
         // run without splash if we are initializing or nosplash
         // was specified (splashdown = true)
-        if (initialize || splashDown || bridge == null) {
+        boolean isRunningFromConsole = System.console() != null;
```
ShadelessFox: Doesn't this mean the splash will be hidden in the development environment as well?

plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/DBeaverLauncher.java
```diff
@@ -2262,7 +2262,8 @@ private InputStream getStream(URL location) throws IOException {
     private void handleSplash(URL[] defaultPath) {
         // run without splash if we are initializing or nosplash
         // was specified (splashdown = true)
-        if (initialize || splashDown || bridge == null) {
+        boolean isRunningFromConsole = new ConsoleChecker().isRunningFromConsole();
```
ShadelessFox: Maybe make the method static?

ShadelessFox: This function may potentially throw linkage errors. To be extra safe, please wrap it with try-catch so in case something goes wrong, the product will start normally regardless.

ShadelessFox (approved):

ShadelessFox:
Explanation of what's happening: there's a running instance of DBeaver; the user starts DBeaver via the CLI and passes a script filename. The newly created instance of DBeaver finds an existing instance of DBeaver and uses RMI to call the `IInstanceController#openExternalFiles` method. This method tries to open a file in blocking mode and shows an error if the file cannot be opened for whatever reasons, in blocking mode as well. All this happens when there's still a splash screen present: it won't be closed because the remote method is still executing because it's blocked by the `UIUtils#syncExec`. The fix is to perform UI operations asynchronously.

E1izabeth (approved):

Matvey16:
Verified

PR #35522, opened 2024-09-09
dbeaver#35147 Fix incorrect generated `SELECT ... WHERE ... IN` query for system tables

### LINKED ISSUE #35147: "Generate SQL" -> SELECT .. WHERE .. IN: generates incorrect query for system tables
### Description

"Generate SQL" generates incorrect query: there is no column name and values in IN-operator.

- Execute this query:
```
select *
from pg_stat_activity

```

https://github.com/user-attachments/assets/b837f7f7-2818-48cd-9e2c-531933bd9636

### DBeaver Version

Community Edition 24.1.5.202408012322

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

_No response_

### Additional context

_No response_

tlm365:
Closes #35147.

** After change

https://github.com/user-attachments/assets/2e92e737-675e-4fe6-8f3b-1e76193a4ab2


E1izabeth:
Thank you for contribution. We'll review your PR and let you know if some changes are required

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/generator/SQLGeneratorSelectManyFromData.java
```diff
@@ -24,48 +25,40 @@
...
 import java.util.List;
 
 public class SQLGeneratorSelectManyFromData extends SQLGeneratorResultSet {
 
     @Override
-    public void generateSQL(DBRProgressMonitor monitor, StringBuilder sql, IResultSetController object) {
-        Collection<DBDAttributeBinding> keyAttributes = getKeyAttributes(monitor, object);
+    public void generateSQL(DBRProgressMonitor monitor, StringBuilder sql, IResultSetController object) throws DBException {
+        List<DBDAttributeBinding> whereAttr = getSelectedAttributes();
+        if (whereAttr.size() > 1) {
```
HocKu7: It is supposed to work with multiple attributes as well. In that case, try to put them together with an AND statement. Like `a IN (...) and b in (...)`

HocKu7 (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/generator/SQLGeneratorSelectManyFromData.java
tlm365: @HocKu7 Thanks for reviewing, it's updated.

HocKu7 (approved):

Matvey16:
It was originally intended to use primary key (if it exists) values in this query. This fix affects that functionality and overrides the use of primary key if it is not explicitly selected. I suggest to prioritize the primary keys values if it exists regardless of selection

ask9:
@Matvey16 Sorry. It seems to be, it was my misunderstanding of the feature.
I expected to get a correct query with where-condition based on the values of the selected cells and their columns even if there is no primary key in the result set.
Now I'm not sure if this should be another feature or it is possible to combine these two behaviours somehow...
Sometimes, it would be useful to get custom WHERE-condition based on the selected cell in spite of the existing primary key on the table.

E1izabeth:
@tlm365 Please check this comment: https://github.com/dbeaver/dbeaver/pull/35522#issuecomment-2431567597, and let me know about your plan for this PR. Will you fix it?
Now it's stale and has conflicts with the `devel` branch

tlm365:
Sorry everyone for the late update,

> @tlm365 Please check this comment: [#35522 (comment)](https://github.com/dbeaver/dbeaver/pull/35522#issuecomment-2431567597), and let me know about your plan for this PR. Will you fix it? Now it's stale and has conflicts with the `devel` branch

@E1izabeth Yes, I am willing to finish this PR. Just want to confirm the solution.
(1) Use WHERE on (primary key) and **ignore** the selected column if the primary key exists; otherwise, use WHERE on the selected column.
(2) Use WHERE on both (primary key and selected column) if the primary key exists; otherwise, use WHERE on the selected column.

Which one, (1) or (2), is our intended target?

arhayka:
We discussed this question with the development team and decided to do the following:
1/We want to add a new generator that will generate select by cells
2/Add for the existing generator that it generates selects by unique keys

tlm365:
> We discussed this question with the development team and decided to do the following: 1/We want to add a new generator that will generate select by cells 2/Add for the existing generator that it generates selects by unique keys

@arhayka So sorry, I couldn't implement the above solution. I'll close this PR to wait for someone take it on.

PR #35615, opened 2024-09-16
#34420 Fix booelan settings

### LINKED ISSUE #34420: `Booleans` settings UX is too complex
### Description

1. We want to move this settings to some other page about UI (it's hard to find them now)
image
2. It's not clear how to use all these switches. State column confuses. Label contents in text representation should
3. Label contents disappear if you switch to Icons and then back. Page becomes invalid - see error in the top, Apply button disabled.

https://github.com/dbeaver/dbeaver/assets/28875055/26ecb8f4-e78f-4df6-8320-24d7e08169ae

### DBeaver Version

DBeaver 24.1.0

ivgag:


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageMiscellaneous.java
```diff
@@ -30,7 +30,15 @@
...
-import org.eclipse.swt.widgets.*;
+import org.eclipse.swt.widgets.Button;
+import org.eclipse.swt.widgets.Composite;
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.swt.widgets.Group;
+import org.eclipse.swt.widgets.Label;
+import org.eclipse.swt.widgets.MenuItem;
+import org.eclipse.swt.widgets.Text;
+import org.eclipse.swt.widgets.ToolBar;
+import org.eclipse.swt.widgets.ToolItem;
```
E1izabeth: I believe these changes in import was not intentional. Please return it back.

ShadelessFox (approved):

Matvey16:
This bug still persists if an option with icons is saved. After that changing to text is impossible without manually changing every option
STR
1. Change the display to icons
2. Apply setting and close
3. Try changing to text
4. 
https://github.com/user-attachments/assets/f4ea07c8-67b1-4244-8d75-0a3e2b8d31a2

Matvey16:
But now opening settings overrides previous settings
1. Change the text for boolean values
2. Open preferences again and press apply without changing anything


https://github.com/user-attachments/assets/24dc08db-2be8-43b5-969f-235acbe4e652

Matvey16:
Verified

PR #35665, opened 2024-09-20
dbeaver/pro#3352 Make completion job to be user job, set cancel status on job cancel

E1izabeth:


ShadelessFox (approved):
FYI, there's an API for managing multiple jobs at once: `org.eclipse.core.runtime.jobs.JobGroup`. Maybe it could be useful.

Matvey16:
Verified

PR #35707, opened 2024-09-25
#35207 Fix highlighting update on disconnect

### LINKED ISSUE #35207: Highlighting issues after reconnect
With a relatively complex query against a redshift database, DBeaver 24.13 (Community edition) highlights all of the column names in my tables as red. Even though I qualify the table name properly. This happens in the ON clause in a join, in simple selection of the columns in a table, inside a wit tn as(select....) construct. In other words pretty much anywhere.

select
	t.tkt_num_id tnum,
	tc2.cpn_num cnum,
	t.base_fare_amt amnt
from
	tkt t
join tkt_cpn tc2 on
	tc2.tkt_num_id = t.tkt_num_id

In the example pasted the columns tkt_num_id and base_fare_amt are both in the tkt table (aliased as t)
Also cpn_num is in the tkt_cpn table (aliased as tc2). However they are both shown in red as can be seen in the attached screenshot.

Screenshot 2024-08-14 094947

Is there something that I can do to help DBeaver get this right? Or is it that my query is too complicated? There are 7 tables involved in the actual query, some of them multiple times (with different aliases, of course)

E1izabeth:
How to reproduce and test:
1. Create a script for PostgreSQL containing, when there is no table Employee in the active scheme
```
ALTER TABLE "Employee"
    ALTER COLUMN employeename DROP DEFAULT,
    ADD COLUMN code uuid default null REFERENCES "Elizabeth"."actor_uuid"(code7_uuid),
    ALTER COLUMN employeename TYPE timestamp with time zone
    USING
        timestamp with time zone 'epoch' + foo_timestamp * interval '1 second',
    ALTER COLUMN foo_timestamp SET DEFAULT now()
;
```
2. Connect to the database and see that  "Employee" is highlighted in red
3. Disconnect from the database - highlighting should be updated and red highlighting should disappear for this query

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
@@ -3074,13 +3074,11 @@ public class SQLEditor extends SQLEditorBase implements
     }
 
     @Override
-    public void beforeConnect()
-    {
+    public void beforeConnect() {
     }
 
     @Override
-    public void beforeDisconnect()
-    {
+    public void beforeDisconnect() {
         closeAllJobs();
     }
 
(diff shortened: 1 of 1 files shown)


uslss:
verified

PR #35721, opened 2024-09-26
#35706 Active project delete fix

### LINKED ISSUE #35706: Restriction on deleting an active project has disappeared
### Description

Previously it was impossible to delete an active project, its context menu didn't have this item. Most likely it was broken after refactoring merge.
Steps to reproduce:
1. Open the Projects view
2. Call the context menu for the active project
Expected: No Delete menu Item
Actual: Delete item exists in the context menu, and we can successfully delete active project, after that no active project is selected

### DBeaver Version

24.2.2

### Operating System

win11

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

serge-rider:


plugins/org.jkiss.dbeaver.model.rcp/src/org/jkiss/dbeaver/model/rcp/DesktopWorkspaceImpl.java
```diff
@@ -318,14 +318,17 @@ public void deleteProject(@NotNull DBPProject project, boolean deleteContents) t
         if (!(project instanceof RCPProject rcpProject)) {
             throw new DBException("Project '" + project.getName() + "' is not an RCP project");
         }
+        if (project == activeProject) {
```
E1izabeth, line 321: But user will see the Delete button in the context menu still, as far as I understand.

Matvey16:
Verified

plugins/org.jkiss.dbeaver.model.rcp/src/org/jkiss/dbeaver/model/rcp/DesktopWorkspaceImpl.java
serge-rider: There is another verification for that

PR #35726, opened 2024-09-26
dbeaver/dbeaver#35725 Export driver files

### LINKED ISSUE #35725: Export driver files
When driver is a Maven artifact its dependencies may reside in tens of different folders and it is hard to locate these files on disk.

Sometimes we need to get JDBC driver jar files in a single folder (e.g. to examine. compare, upload to 3rd party software, etc).
Let's add driver files export button in the driver edit dialog.

serge-rider:


Matvey16:
In case of more than one artifact, Dbeaver will export files only from the artifacts with the same GroupID.

In case of oracle all files are exported

dbeaver_9Rh7c0UdFD


In case of postgres only two files are exported

dbeaver_qBWmbFhmte

Matvey16:
Verified

PR #35739, opened 2024-09-27
#35530 Fix SQL completion doesn't show schemas for H2 untill you type prefix

### LINKED ISSUE #35530: SQL completion doesn't show schemas for H2 untill you type prefix
### Description

At least in H2 databases.
Type `SELECT * FROM |` and call completion - schemas and tables should be shown, but now only tables come to the list of proposals.

### DBeaver Version

24.2

E1izabeth:
H2
image

PostgreSQL
image


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
```diff
@@ -270,7 +270,7 @@ private void runAnalyzer() throws DBException {
                     } else {
                         rootObject = dataSource;
                     }
-                    if (!(rootObject instanceof DBPDataSource)) {
+                    if (!(rootObject instanceof DBPDataSource) || rootObject == dataSource) {
```
serge-rider: Isn't it the same as if (true) ?

E1izabeth: Maybe yes

uslss:
verified

PR #35751, opened 2024-09-30
dbeaver/pro#3281 Resolve object aliases during semantic analysis

E1izabeth:
For example, for Oracle `customers` should not be classified as error here
```sql
CREATE SYNONYM customers FOR EMPLOYEE;
SELECT COUNT(*) FROM customers;
```


HocKu7 (approved):

ShadelessFox (approved):

uslss:
verified

Issue #35760, opened 2024-10-01
DBeaver l10n: Brazilian Portuguese new translation strings and revisions


brlarini:
Hello.

I've translated some more of the missing strings in pt_BR files and done some revisions.
I'll create a pull request and link it here.

Thanks!

MashaKorax:
verified

PR #35761, opened 2024-10-01
dbeaver#35760 Brazilian Portuguese translation strings added, plus some revisions on existing ones

brlarini:
Pull request associated with issue #35760
Thanks!


brlarini:
Wait... seems like not all files were included...
I'll review it as soon as I can, sorry.

PR #35769, opened 2024-10-01
dbeaver/pro#3102 hostname could be a optional

HocKu7:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/access/DBAAuthModel.java
```diff
@@ -97,4 +97,12 @@ void refreshCredentials(
         @NotNull CREDENTIALS credentials
     ) throws DBException;
 
+    /**
+     * Getting necessity of the hostname field
+     *
+     * @return necessity of the hostname field
+     */
+    default boolean isModelRequireHost() {
```
ivgag:
```suggestion
    default boolean isHostRequired() {
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/access/DBAAuthModel.java
```diff
@@ -97,4 +97,12 @@ void refreshCredentials(
         @NotNull CREDENTIALS credentials
     ) throws DBException;
 
+    /**
+     * Getting necessity of the hostname field
+     *
+     * @return necessity of the hostname field
+     */
+    default boolean isHostRequired() {
+        return true;
```
serge-rider: Host is not required by many JDBC drivers. In Generic driver we can verify this by presence of {host} item in sample URL

plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/views/MySQLConnectionPage.java
```diff
@@ -220,11 +223,18 @@ private void updateUrl() {
     public boolean isComplete() {
         if (isCustomURL()) {
             return !CommonUtils.isEmpty(urlText.getText());
+        } else {
+            boolean hostAndPortIsPresent = hostText != null &&
+                !CommonUtils.isEmpty(hostText.getText()) &&
+                (portText == null || !CommonUtils.isEmpty(portText.getText()));
+            DBPAuthModelDescriptor selectedAuthModel = authModelSelector.getSelectedAuthModel();
```
serge-rider: Let's move this verification (isHostRequiredByAuthModel) to some utils class to avoid copy-paste.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageWithAuth.java
```diff
@@ -38,7 +38,7 @@ public abstract class ConnectionPageWithAuth extends ConnectionPageAbstract {
 
     private static final Log log = Log.getLog(DataSourceProviderRegistry.class);
 
-    private AuthModelSelector authModelSelector;
+    protected AuthModelSelector authModelSelector;
```
serge-rider: Let's use getter but direct access

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/access/DBAuthUtils.java
```diff
@@ -84,4 +88,21 @@ public static boolean promptAndChangePasswordForCurrentUser(
         }
         return false;
     }
+
+    public static boolean isPageCompleteByAuthModel(
```
serge-rider: Let's rename it to isHostPresent.
Btw port is optional for almost all database driver. Why we check it as required?

serge-rider (changes requested):

PR #35772, opened 2024-10-01
#14474 Fix postgres permissions editor

### LINKED ISSUE #14474: Checkboxes for permissions disappear in table properties

ivgag:


E1izabeth:
We need to test Redshift also

Matvey16:
Verified

Issue #35430, opened 2024-09-02
INFORMATION_SCHEMA views aren't recognized in SQL Server


Tix2garr:
### Description

When trying to use INFORMATION_SCHEMA in SQL Server, the views are marked as not found.  Such as:

`SELECT *
FROM INFORMATION_SCHEMA.ROUTINES`

I get a message _Table INFORMATION_SCHEMA.ROUTINES not found_, like:

SELECT *
_Table INFORMATION_SCHEMA.ROUTINES not found_
FROM INFORMATION_SCHEMA._ROUTINES_

If I just have

SELECT *
FROM INFORMATION_SCHEMA

and don't type fast enough, I get 
_Table INFORMATION_SCHEMA not found_

The new analysis is nice, and it's getting better!, but still has some issues.  <wry smile>

[dbeaver-diagnostic-info-1725285105290.zip](https://github.com/user-attachments/files/16838713/dbeaver-diagnostic-info-1725285105290.zip)

### DBeaver Version

Community edition 24.2.0

### Operating System

Windows 10

### Database and driver

Microsoft JDBC Driver for SQL Server 
12.8.0.jre11

### Steps to reproduce

1. Connect to SQL Server
2. Type in 

SELECT *
FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE

and wait for the error.  <wry smile>

E1izabeth:
Hi, @Tix2garr 
Thank you for bug report. For now I can't reproduce the issue. However, I see a lot of errors saying that connection is broken. Could you check if this bug is stable reproducible for you to check if it's related to the connection state?

uslss:
Thanks for the report.
The issue is reproducible for me after reconnecting and the script reopen

https://github.com/user-attachments/assets/13e43a85-fb15-437a-8d37-9ad2e1e671c0

Issue #35313, opened 2024-08-22
Miss Warning In SQL Editor


mehmetkurban:
### Description

Everting works. But SQL editor giving me table all cons colums not found. My script works well and returning me succsess result. But i dont understand why sql editor shows error. Is there a bug here? 
![17243242085998099003552667939204](https://github.com/user-attachments/assets/c706595f-a280-4340-a746-2ee21494c87b)

### DBeaver Version

24.1.5

### Operating System

W10 Enterprise 

### Database and driver

Oracle
Oracle ojdbc11-23.2.0.0

### Steps to reproduce

![17243236389573843568635365780093](https://github.com/user-attachments/assets/eacd26cb-5221-4e48-a924-c1dbd9853c24)

E1izabeth:
Does active catalog/schema selector shows the schema where these tables are located?
![image](https://github.com/user-attachments/assets/22d49731-77a2-4841-9a2c-a115ea866b95)

mehmetkurban:
Its active. Not N/A. 
The ALL_CONSTRAINTS table is a system view in Oracle databases. It provides information about all the constraints in the database, such as primary keys, foreign keys, unique keys, and check constraints. This table is particularly useful for database administrators and developers to query and manage table constraints.


Oracle Database: The ALL_CONSTRAINTS table is part of the Oracle database system and is provided by Oracle.

[![IMG_20240822_160151.jpg](https://github.com/user-attachments/assets/2731db8e-3e35-4504-9abf-5ce74c1d5efa)

One more point SQL Editor not shows my whole squences and giving me Warning. But i can take succsses seqence id when i run script. 

If you want, you can try miss warning.
I am sharing the script:
SELECT

    a.owner AS schema_name,

    a.table_name AS table_name,

    col.column_name AS foreign_key_column,

    --a.constraint_name AS foreign_key_name,

    a.delete_rule AS delete_rule,

    c.owner AS referenced_schema,

    c.table_name AS referenced_table

    --c.constraint_name AS referenced_constraint

FROM

    all_constraints a

JOIN

    all_constraints c ON a.r_constraint_name = c.constraint_name

JOIN

    all_cons_columns col ON a.constraint_name = col.constraint_name AND a.owner = col.owner

WHERE

    a.constraint_type = 'R'

ORDER BY

    a.owner, a.table_name;

E1izabeth:
Unfortunately, I can't reproduce it
[...]

PR #35797, opened 2024-10-03
dbeaver/pro#3281 Support for synonyms in semantic analyses

E1izabeth:
image

```sql
SELECT * FROM elizabeth.yuuy ; -- synonym for employee table 

SELECT * FROM "PUBLIC".customers ; -- public synonym for table employee

SELECT * FROM "ELIZABETH".employee e WHERE rowid > 3; -- real table

 -- synonyms are not allowed for ALTER tables
ALTER TABLE "ELIZABETH".yuuy ADD COLUMN x int


-- queries to check synonyms in metadata
SELECT S.* FROM ALL_SYNONYMS S WHERE S.SYNONYM_NAME = 'CUSTOMERS';
SELECT S.* FROM ALL_SYNONYMS S WHERE S.table_owner = 'ELIZABETH';
```


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBSObjectContainer.java
```diff
@@ -93,4 +95,7 @@ public interface DBSObjectContainer extends DBSObject
      */
     void cacheStructure(@NotNull DBRProgressMonitor monitor, int scope) throws DBException;
 
+    default List<DBSObjectContainer> getImportedScopes(@NotNull DBRProgressMonitor monitor) throws DBException {
```
serge-rider: Let's move it to separate interface DBS
DBSVisibiityScopeProvider {
  List<DBSStructContainer> getPublicContainers()
}
It should be implemented by datasource or catalog.

Almost all databases have INFORMATION_SCHEMA which is always visible. 
In PostgreSQL (and forks) we have searchPath (which we already read). It is dynamic and can be changed in runtime.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
```diff
@@ -109,6 +110,11 @@ public class PostgreSchema implements
         dataTypeCache = new PostgreDataTypeCache();
     }
 
+    @Override
+    public List<DBSObjectContainer> getPublicContainers(@NotNull DBRProgressMonitor monitor) throws DBException {
+        return List.of(this.database.getSchema(monitor, PostgreConstants.INFO_SCHEMA_NAME));
```
serge-rider: I think this logic must be somewhere in AbstractDataSource because info schema is SQL standard (it even present in some NoSQL databases). We can add something like isInformationSchemaSupported in DBPDataSourceInfo.

In PostgreSQL pg_catalog is also always visible.

serge-rider: In Oracle SYS and SYSTEM are public schemas.
In DB2 SYS and IBMSYS and maybe something else too.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreSchema.java
E1izabeth: I separated the public schemas support from this PR. Will handle it in scope of #[35430](https://github.com/dbeaver/dbeaver/issues/35430) and https://github.com/dbeaver/dbeaver/issues/35313#issuecomment-2314762724

ShadelessFox (approved):

Matvey16:
Verified

PR #35799, opened 2024-10-03
dbeaver/dbeaver#35080 Additional JVM arguments search locations

### LINKED ISSUE #35080: Add config.ini location which is not tied to a specific version
### Description

The wiki page https://github.com/dbeaver/dbeaver/wiki/How-to-set-a-variable-if-dbeaver.ini-is-read-only says to look up the directory based upon `Help -> Installation Information -> Configuration -> org.osgi.framework.storage`.

However, as you can see in that screenshot, this spits out a directory inside install-data which is nested inside a version-specific folder. I tried this on both Windows and Linux and while it doesn't print a directory inside `.eclipse` now (it's in the proper workspace location: https://github.com/dbeaver/dbeaver/wiki/Workspace-location)

I'm sure you can understand why I'd rather not put a config file into a version-specific directory - can we just use the workspace6 directory for config.ini?

As another note - and maybe for another issue - it would be nice if there was a documentation page - https://github.com/dbeaver/dbeaver/wiki/Configuration-files-in-DBeaver would fit well - which mentioned the other files that you can use to configure - most notably `workspace6/.metadata/.plugins/org.eclipse.core.runtime./.settings/org.jkiss.dbeaver.core.prefs`  - ideally this can be contrasted w/ the config.ini - I believe there are some fields that overlap and others that do not? For example I couldn't get config.ini to recognize the `ui.show.tip.of.the.day.on.startup` config but I believe you can put `-Djavax.net.ssl.trustStore` into the eclipse configs and into config.ini (where you omit the `-D`).

Note that https://github.com/dbeaver/dbeaver/wiki/Configuration-files-in-DBeaver doesn't even mention config.ini which I consider to be a gap that should be fixed.

My use case is to persist these configs into a dotfiles repo so I can easily restore my preferences.

### DBeaver Version

Community Edition 24.1.3

### Operating System

Windows 10

### Database and driver

N/A

### Steps to reproduce

_No response_

### Additional context

_No response_

ShadelessFox:


Matvey16:
Verified

sebalaise:
Hello @serge-rider ,
Why the product/debug directory was removed om this release_24.2.3 version ?

Issue #35848, opened 2024-10-07
Right-click on the empty place of the Metadata Editor throws NPE


kseniaguzeeva:
### Description

Right-click on the empty place of the Metadata Editor throws NPE

### DBeaver Version

Version 24.2.2.202409300848

### Operating System

Win11

### Database and driver

PostgreSQL and MySQL have been tested

### Steps to reproduce

1. Open a database/schema/table in the Metadata Editor
2. Don't place a focus on an object and right-click on an empty place of the Metadata Editor

AR: NPE appears in the log
ER: the context menu is opened withput any error
Workaround: select an object, then click on an empty place

https://github.com/user-attachments/assets/4df2de8c-07e5-4ce5-aa90-831bdbd05cfc

[NPE_CE.log](https://github.com/user-attachments/files/17280300/NPE_CE.log)

kseniaguzeeva:
Verified

PR #35850, opened 2024-10-07
#35848 Fix storesQuotedCase return value for MySQLDialect

### LINKED ISSUE #35812: Table not found in SQL editor when using backticks
### Description

While writing a SQL query with table name between backticks, the name turn red with error "Table [name] not found". If I remove backticks the table is found with no problem.

Before updating to version 24.2.1 it didn't show any error, but I don't know what version I was using then.

### DBeaver Version

24.2.1

### Operating System

PopOS 22.04

### Database and driver

MySQL driver

### Steps to reproduce

1. Connect to MySQL DB
2. Create table `my_table`
3. Open SQL editor
4. Write ```SELECT * FROM `my_table`; ```
5. ``` `my_table` ``` show error

### Additional context

_No response_

E1izabeth:
MySQL driver returns `true` for `metaData.storesUpperCaseQuotedIdentifiers()` and `metaData.storesMixedCaseQuotedIdentifiers()`. However, in fact, execution works as it stores MIXED case for quoted identifiers. So. let's override it in dialect.


MashaKorax:
verified

Issue #35475, opened 2024-09-04
Disconnect connection and re-execute the connection without explicit connection leads to wrong state of execution (query is actually executing but seems to be as non executing)


ask9:
### Description

It confuses a user and many controls don't work ("Cancel active query", maybe something else).

https://github.com/user-attachments/assets/ebd2510a-8b49-48a9-b599-59bbb20c2f9b

use code:
```

create or replace procedure proc1()
language plpgsql
as
$$
begin
	select pg_sleep(10000);
end;
$$
;

call proc1();

```

use code:
```

SELECT * FROM pg_stat_activity
WHERE state = 'active' ;


select pg_terminate_backend(9999) ;
```

### DBeaver Version

Community Edition 24.1.5-202408122322

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.4

PR #35857, opened 2024-10-07
dbeaver/dbeaver#35475 Fix sql script execution

### LINKED ISSUE #35882: Execution of the script on disconnected state
### Description

```
language plpgsql
as
$$
begin
	select pg_sleep(10000);
end;
$$
;
```
image

### DBeaver Version

DBeaver 24.2.2

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

HocKu7:
Now SQLDialect should be taken from dataSourceContener, not dataSource. In that case, query will be parsed with correct dialect, even without dataSource


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -675,6 +675,27 @@ public boolean isActiveTask() {
...
 
+    @NotNull
+    @Override
+    public SQLDialect getSQLDialect() {
+        if (dataSourceContainer == null) {
+            return BasicSQLDialect.INSTANCE;
+        }
+        SQLDialectMetadata scriptDialect = dataSourceContainer.getScriptDialect();
+        try {
+            return scriptDialect.createInstance();
```
E1izabeth: I believe we shouldn't do it in SQLEditor. Dialect should be initialized (initialize() call) then, which happens only on connection establishment.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/SQLBackgroundParsingJob.java
```diff
@@ -488,7 +487,7 @@ public Interval runWithResult() {
                 // for now just cover the region of interest
                 {
                     NodesIterator<QueuedRegionInfo> it = this.queuedForReparse.nodesIteratorAt(0);
-                    workOffset = (it.getCurrValue() != null || it.next()) ? it.getCurrOffset() : 0;
+                    workOffset = it.getCurrOffset();
```
E1izabeth: It doesn't seem right. Now you are failing to correctly obtain offset of the region which starts at the offset grater that zero, because `it.getCurrValue()` will be null in this case, so we need `it.next()` to move iterator further from the requested position when the region doesn't start exactly at the zero offset.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLParserContext.java
```diff
@@ -74,7 +94,14 @@ public IDocument getDocument() {
     }
 
     public SQLDialect getDialect() {
-        return SQLUtils.getDialectFromDataSource(getDataSource());
+        if (dataSourceContainer != null) {
```
serge-rider, line 109: It must take dialect from DBPDataSource first.
And then from container (in datasource it is more correct because it loads meta info from actual database).

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRuleManager.java
```diff
@@ -89,12 +89,20 @@ public TPRule[] getRulesByType(@NotNull SQLTokenType requiredType) {
...
 
     public void loadRules() {
-        loadRules(null, false);
+        loadRules((DBPDataSourceContainer) null, false);
     }
 
     public void loadRules(@Nullable DBPDataSource dataSource, boolean minimalRules) {
+
+        if (dataSource == null) {
+            loadRules((DBPDataSourceContainer) null, minimalRules);
```
serge-rider, line 97: Maybe it must use container anyway?

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLParserContext.java
```diff
@@ -48,6 +55,19 @@ public class SQLParserContext {
 
     public SQLParserContext(@Nullable DBPDataSource dataSource, @NotNull SQLSyntaxManager syntaxManager, @NotNull SQLRuleManager ruleManager, @NotNull IDocument document) {
         this.dataSource = dataSource;
+        if (dataSource != null) {
```
serge-rider, line 62: Maybe we should pass only container here. And parsercontext will get datasource from it? Does it make sense to have 2 variables?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -675,6 +675,27 @@ public boolean isActiveTask() {
...
+        } catch (DBException e) {
+            log.warn(String.format("Can't create sql dialect for %s:%s", scriptDialect.getId(), scriptDialect.getLabel()));
+        }
+        return BasicSQLDialect.INSTANCE;
+    }
+
+    @NotNull
+    @Override
+    protected SQLParserContext getParserContext(SQLSyntaxManager syntaxManager, SQLRuleManager ruleManager, IDocument document) {
+        return new SQLParserContext(dataSourceContainer, syntaxManager, ruleManager, document != null ? document : new Document());
```
E1izabeth: Could you point me out to the place where we call this method? I can't find it.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLParserContext.java
```diff
@@ -74,7 +94,14 @@ public IDocument getDocument() {
     }
 
     public SQLDialect getDialect() {
-        return SQLUtils.getDialectFromDataSource(getDataSource());
+        if (dataSourceContainer != null) {
+            try {
+                return dataSourceContainer.getScriptDialect().createInstance();
```
E1izabeth: I believe we shouldn't create dialect instance this way. Described in another comment to a similar place.

E1izabeth (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -675,6 +675,27 @@ public boolean isActiveTask() {
         return getTotalQueryRunning() > 0;
     }
 
+    @NotNull
+    @Override
+    public SQLDialect getSQLDialect() {
```
E1izabeth, line 679: Also I believe it's not right to return instances that require two-stage initialization only after the first stage from getter

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/data/SQLQueryDataContainer.java
```diff
@@ -100,8 +100,8 @@ public DBCStatistics readData(
             SQLSyntaxManager syntaxManager = new SQLSyntaxManager();
             syntaxManager.init(dataSource.getSQLDialect(), dataSource.getContainer().getPreferenceStore());
             SQLRuleManager ruleManager = new SQLRuleManager(syntaxManager);
-            ruleManager.loadRules(dataSource, false);
-            SQLParserContext parserContext = new SQLParserContext(getDataSource(), syntaxManager, ruleManager, new Document(query.getText()));
+            ruleManager.loadRules(getDataSourceContainer(), false);
+            SQLParserContext parserContext = new SQLParserContext(getDataSourceContainer(), syntaxManager, ruleManager, new Document(query.getText()));
```
E1izabeth: I don't understand this approach at all. Why do you change dataSource to dataSourceContainer everywhere? Is it only for cases where we don't have dataSource?

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
HocKu7: You are right, but the connection will be established after the query parsing. In that case, we can't parse database-specific syntax. So, we need to know the dialect before a connection

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/data/SQLQueryDataContainer.java
HocKu7: Yes, probably it seems more right because we can extract dialect by dataSourceConnection without even connecting

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLParserContext.java
serge-rider: At least we can cache it.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLScriptParser.java
```diff
@@ -748,7 +748,7 @@ public static SQLScriptElement extractActiveQuery(@NotNull SQLParserContext cont
 
     public static List<SQLQueryParameter> parseParametersAndVariables(SQLParserContext context, String selectedQueryText) {
         SQLParserContext ctx = new SQLParserContext(
-                context.getDataSource(),
+                context.getDataSource() != null ? context.getDataSource().getContainer() : null,
```
serge-rider: Let's keep both constructors with datasource and with container

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLScriptParser.java
```diff
@@ -1011,11 +1011,11 @@ public static List<SQLScriptElement> parseScript(DBPDataSource dataSource, Strin
         SQLSyntaxManager syntaxManager = new SQLSyntaxManager();
         syntaxManager.init(dataSource.getSQLDialect(), dataSource.getContainer().getPreferenceStore());
         SQLRuleManager ruleManager = new SQLRuleManager(syntaxManager);
-        ruleManager.loadRules(dataSource, false);
+        ruleManager.loadRules(dataSource.getContainer(), false);
```
serge-rider: Let's keep old function with datasource (it must call loadRules with container)

serge-rider: This will also reduce number of changes/refactoring

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLParserContext.java
```diff
@@ -46,16 +54,33 @@ public class SQLParserContext {
     @Nullable
     private DBPPreferenceStore preferenceStore;
 
+    @Nullable
+    private SQLDialect cacheSqlDialect = null;
```
E1izabeth:
```suggestion
    private SQLDialect cachedSqlDialect = null;
```

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRuleManager.java
```diff
@@ -89,12 +89,19 @@ public TPRule[] getRulesByType(@NotNull SQLTokenType requiredType) {
...
 
     public void loadRules() {
-        loadRules(null, false);
+        loadRules((DBPDataSourceContainer) null, false);
     }
 
     public void loadRules(@Nullable DBPDataSource dataSource, boolean minimalRules) {
+        if (dataSource == null) {
+            loadRules((DBPDataSourceContainer) null, minimalRules);
+        } else{
```
E1izabeth:
```suggestion
        } else {
```

serge-rider:
What I don't like: we now rely only on dialect from container. Although dialect from datasource is more correct and complete.
I.e. we should use container.getDriver().getScriptDialect() ONLY if we don't have live datasource.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorBase.java
```diff
@@ -149,6 +149,7 @@ public abstract class SQLEditorBase extends BaseTextEditor implements DBPContext
     private SQLCompletionContext completionContext;
     private SQLOccurrencesHighlighter occurrencesHighlighter;
     private SQLSymbolInserter sqlSymbolInserter;
+    protected DBPDataSourceContainer dataSourceContainer;
```
serge-rider: We shouldn't move it here. Container is usually calculated dynamically in sub classes.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLParserContext.java
```diff
@@ -74,7 +103,24 @@ public IDocument getDocument() {
     }
 
     public SQLDialect getDialect() {
```
E1izabeth, line 105: I still don't like that the method with such a complex thing as dialect instantiation is called like just a getter

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRuleManager.java
```diff
@@ -89,12 +89,19 @@ public TPRule[] getRulesByType(@NotNull SQLTokenType requiredType) {
...
     public void loadRules() {
-        loadRules(null, false);
+        loadRules((DBPDataSourceContainer) null, false);
     }
 
     public void loadRules(@Nullable DBPDataSource dataSource, boolean minimalRules) {
+        if (dataSource == null) {
+            loadRules((DBPDataSourceContainer) null, minimalRules);
+        } else {
+            loadRules(dataSource.getContainer(), minimalRules);
```
E1izabeth, line 99: Now you get data source container from data source and then in the loadRules method you get data source from that container. Why not passing datasource directly then?

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLParserContext.java
E1izabeth: I agree with this comment

HocKu7: it doesn't work in that way. This change will bring a lot of changes in other places so we already decided to not to do that

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLParserContext.java
HocKu7: Probably it should be part of refactoring, but I'm not sure that it related to current fix, isn't?

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRuleManager.java
HocKu7: it brings a lot of changes

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -675,6 +674,23 @@ public boolean isActiveTask() {
...
+        if (dataSourceContainer == null) {
+            return BasicSQLDialect.INSTANCE;
+        }
+        SQLDialectMetadata scriptDialect = dataSourceContainer.getScriptDialect();
```
serge-rider, line 683: Here we also must check dialect from actual datasource first.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLParserContext.java
E1izabeth: You change the whole logic of this method. Since you're doing it, it's a part of this PR

PR #35873, opened 2024-10-08
#35475 Fix negative curJobRunning value

### LINKED ISSUE #35475: Disconnect connection and re-execute the connection without explicit connection leads to wrong state of execution (query is actually executing but seems to be as non executing)

E1izabeth:
### How to test

1. Create procedure with pg_sleep(10000)
```
language plpgsql
AS 
$$
begin
	select pg_sleep(10000);
end;
$$
;
```
2. Introduce query `call proc1();` to the script
3. Select this query and execute
4. Press disconnect and agree to stop query execution
5. Select this query and execute
6. Check the state of the `Execute-->Cancel active query` context menu item. It should be enabled while the query execution task is running (you can check it in Background Tasks view)


HocKu7 (approved):

ShadelessFox (approved):

MashaKorax:
verified

PR #35879, opened 2024-10-09
dbeaver/pro#3078 Clickhouse structs and arrays. Show value as is when…

serge-rider:
… complex transform disabled.


adityaranjan2005 (commented):
Changes are required

PR #35883, opened 2024-10-09
#35638 Turn off by default `Show values` setting in autocompletion

### LINKED ISSUE #35638: Dbeaver not authorized to
### Description

Just wanted to bring this up again - having this kind of 'extra queries by default that will definitely crash your server if it's very large ' is exceedingly dangerous should NOT be on by default...

https://github.com/dbeaver/dbeaver/issues/12087

### DBeaver Version

Community 23.0.4.202305071722

### Operating System

Windows 11

### Database and driver

mysql

### Steps to reproduce

Trying to run an sql command like update items  SET tagnames=concat(tagnames, ", adult") WHERE id=17939088

Causes the query SELECT DISTINCT tagnames\nFROM apm.items\nORDER BY 1 to be sent by dbeaver without permission and without any logging in the query log.  As there are ~20m rows and this is not an indexed column, this crashes the site.

Quite a few folks have had this problem as evidenced in the original ticket.

Thanks for an otherwise great program :)

### Additional context

_No response_

E1izabeth:
Turn off by default the `Show values` setting, which turns on the proposals that require SELECT DISTINCT queries
image


ShadelessFox (approved):

Matvey16:
Verified

PR #35885, opened 2024-10-09
CB-5424 QM cleanup job

alexander-skoblikov:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/runtime/PeriodicSystemJob.java
```diff
@@ -0,0 +1,60 @@
...
+    @NotNull
+    protected final DBPPlatform platform;
+    private final long periodMs;
+
+    public PeriodicSystemJob(@NotNull String name, @NotNull DBPPlatform platform, long periodMs) {
+        super(name);
+        this.platform = platform;
+        this.periodMs = periodMs;
+
+        setUser(false);
```
serge-rider: I think it shouldn't be system. It is good to see progress indicator in desktop UI

serge-rider (changes requested):

serge-rider (approved):

PR #35918, opened 2024-10-11
CB-5744. Refactor after review

DenisSinelnikov:


plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMAdminController.java
```diff
@@ -219,4 +219,6 @@ void deleteObjectPermissions(
     @NotNull
     List<SMTeamMemberInfo> getTeamMembersInfo(@NotNull String teamId) throws DBException;
 
+    @Nullable
+    String[] getEnableFeatures();
```
alexander-skoblikov, line 223: do we really need this method?

Issue #19542, opened 2023-03-30
Project, Navigator etc. windows do not autoscroll


RichardSCarchitect:
### Description

When you drag an item (for example a script) up or down the Project window, if you reach the top or bottom of the window you cannot scroll to other items that are there but not displayed

### DBeaver Version

Community Edition 22.3.5

### Operating System

Windows 10 Enterprise

### Database and driver

Any

### Steps to reproduce

Have a project with lots of scripts divided between several folders
Open the Project window
Open one folder near the top, click and drag one script to move to another folder near the bottom.  The target folder is not displayed because it's too far down
As you drag to the bottom of the window, the window does not scroll down to show the target folder

E1izabeth:
Thank you for bug report.

micheal-icp:
@E1izabeth

It's been well over a year since this was reported. Is it possible we could revisit this and get the auto-scrolling in the project and navigator windows implemented? It is such a big pain when you are having to move scripts around within a big nest of script folders and projects.

E1izabeth:
@micheal-icp 
You can vote for the issue with 👍 under the issue description to promote it to a higher priority.

micheal-icp:
@E1izabeth 

I thumbed it up, but I had no idea we had to upvote bugs. I thought that was for feature requests. This ticket was flagged as a bug, not feature request.

arhayka:
This was fixed in version 24.3.0

PR #35937, opened 2024-10-13
dbeaver/pro#3432 add wait for votes description to readme

arhayka:


E1izabeth (approved):

ask9:
Can you please clarify one moment.
Does voting apply to both bugs and feature requests?
Looking at such messages, users do not usually vote for bugs.
I was holding the same opinion.


https://github.com/dbeaver/dbeaver/issues/19542#issuecomment-2321296249

ShadelessFox (approved):

arhayka:
Hi @ask9!
This is mostly about feature requests. However, some tickets can be ambiguous. What users see as a feature request might appear to developers as a bug since it involves something that should work but needs fixing. In these cases, a bug may gather a significant number of votes.

PR #35946, opened 2024-10-14
dbeaver/pro#3458 txn isolation read fix

serge-rider:


DariaMarkaryan:
[dbeaver-thread-dump.txt](https://github.com/user-attachments/files/17376991/dbeaver-thread-dump.txt)

PR #35953, opened 2024-10-15
dbeaver/pro#3403 Add single file selection for ff drivers

HocKu7:


plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/GenericConstants.java
```diff
@@ -73,6 +73,7 @@ public class GenericConstants {
     public static final String PARAM_SUPPORTS_MULTI_INSERT = "supports-multi-insert";
     public static final String PARAM_SUPPORTS_EMBEDDED_DATABASE_CREATION = "supports-embedded-database-creation";
     public static final String PARAM_IDENTIFIER_QUOTES = "identifier-quotes";
+    public static final String PARAM_SUPPORT_DATABASE_FILE_EXTENSION = "supports-database-file-extension";
```
serge-rider: Let's rename it into `database-file-extensions`, it is a coma separated list

serge-rider (changes requested):

PR #35971, opened 2024-10-16
#35469 Show warning instead of error if drop table statement contains if not exists

### LINKED ISSUE #35469: Corretly executed code is displayed with an error (exclamation left of the query) and highlighted with red color
### Description

https://github.com/user-attachments/assets/903f88bb-7677-42f9-8de9-9c2cee1643d8

- use code:
```
drop table if exists hello;
```

Please also check all other Postgres statements with `if exists`-clause.

### DBeaver Version

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.4

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
image


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/model/ddl/SQLQueryTableDropModel.java
```diff
@@ -94,7 +94,13 @@ public SQLQueryDataContext getGivenDataContext() {
     @Override
     protected void applyContext(@NotNull SQLQueryDataContext dataContext, @NotNull SQLQueryRecognitionContext recognitionContext) {
         this.dataContext = dataContext;
+        if (ifExists) {
+            recognitionContext.setTreatErrorAsWarnings(true);
```
ShadelessFox, line 98: I wonder if there might be a situation when this can overwrite the flag at an appropriate time. What I mean by that is that if the caller explicitly sets to `true`, let's say, it will be set to `false` when this method exists, and it might not be the desired behavior.

So I was thinking, what if you make it a stack so you can push and pop different states?

MashaKorax:
Rendering issue.
Str:
1) execute drop statement
2) add ` if exists` after that

dbeaver_PvF4pTbAQ7

https://github.com/user-attachments/assets/2aa2a0f1-d5e1-4081-9560-c6219fb83da1

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/model/ddl/SQLQueryTableDropModel.java
E1izabeth: I don't see how such a situation may happen

E1izabeth:
We don't have a mechanism to aggregate markers of a different type (execution and semantic analysis markers) to show only one of them. It's quite a lot of work to do which doesn't fit to the current issue. Let's create a separate ticket about that.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/model/ddl/SQLQueryTableDropModel.java
E1izabeth: Yes, potentially it could be implemented like a stack

MashaKorax:
verified https://github.com/dbeaver/dbeaver/issues/36048

PR #36016, opened 2024-10-21
dbeaver/dbeaver#36014 fix NPE (in SQLite it's fine for a column to have no data type)

### LINKED ISSUE #36014: [SQLite] NPE when create a column without data type
### Description

STR
In SQLite create a column with epthy data type

https://github.com/user-attachments/assets/a87cc8d3-2950-486b-bef4-cc27eb10c37e

### DBeaver Version

DBeaver 24.2.4

### Operating System

windows 10

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

flambchap:


E1izabeth (approved):

ShadelessFox (approved):

kseniaguzeeva:
Separate bug will be created:
Impossible to delete a column when a column without a data type exists. The reason is in wrong syntax (it should be null instead of n/a):

```
CREATE TEMPORARY TABLE temp AS
SELECT
  Column1,
  Column3
FROM test;

DROP TABLE test;

CREATE TABLE test (
	Column1 INTEGER NOT NULL,
	Column3 N/A
);

INSERT INTO test
 (Column1,
  Column3)
SELECT
  Column1,
  Column3
FROM temp;

DROP TABLE temp;
```

flambchap:
Yeah, null or nothing at all, for example. Thanks @kseniaguzeeva!

PR #36024, opened 2024-10-22
dbeaver/pro#3443 Network profiles credentials save fix

serge-rider:


MashaKorax:
A bag with disappeared username still reproduced.

MashaKorax:
verified

PR #36040, opened 2024-10-23
dbeaver/pro#3498 Navigator filters fixes

serge-rider:


Matvey16:
Second case from the ticket is not fixed. Reproduced on Derby connection

**Case 2**
1. Open the schema context menu (DB2 z/os)
2. Filter -> Hide *schema*
Actual result: the setting does not apply
3. Filter -> Clean filter
Actual result: an error appears
`null
org.jkiss.dbeaver.model
Error
Wed Oct 23 10:38:11 CEST 2024
Cannot detect child node type - can't save filter configuration
`

Matvey16:
Case 3 also reproducebale. The filter is applied but the error pop up appears

image

Matvey16:
In the case of exclusion of every schema there is no way to remove filter without reconnection.

1. Open filter from connection context menu
2. Make filter exclude all schemas (e.g. type `%`)
3. Save filter
4. Open filter again

**Actual result:** Filters are empty
**Expected:** already applied filters are shown


https://github.com/user-attachments/assets/c47923a1-c36b-4292-a42f-87f394cc3b8b

Matvey16:
Filter windeow is not opened if it is called from a schema context menu (works only from connection context menu).
It happens with every object (tables, views etc.) in every driver. Filter window appears only when called from a folder context menu, but not from one of the objects.

https://github.com/user-attachments/assets/4d527d93-ce6a-4972-9e9c-7caee5b66f7c

Matvey16:
Filter opened from child nodes now opens but can't be saved

https://github.com/user-attachments/assets/934a4633-4dd8-4e06-b09f-5d452720e68d

Matvey16:
Also this bug still persists https://github.com/dbeaver/dbeaver/pull/36040#issuecomment-2437314409

Matvey16:
Verified

PR #36056, opened 2024-10-24
Dbeaver/dbeaver#35882 fix sql script execution

### LINKED ISSUE #35882: Execution of the script on disconnected state

HocKu7:


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLParserContext.java
```diff
@@ -74,7 +103,24 @@ public IDocument getDocument() {
     }
 
     public SQLDialect getDialect() {
```
E1izabeth, line 105: By the name, it looks like a getter, but it instantiates an uninitialized script dialect. Getters should not do such things.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorBase.java
```diff
@@ -1309,6 +1323,12 @@ protected boolean isNavigationTarget(Annotation annotation) {
         return super.isNavigationTarget(annotation);
     }
 
+    @Nullable
+    @Override
+    public DBPDataSourceContainer getDataSourceContainer() {
+        return null;
```
E1izabeth: Maybe I missed something, but I don't understand the logic here. You call this method from other methods of this class, however, it always returns null. What's the point of doing that?

HocKu7: It's overridden in SQLEditor

MashaKorax:
verified

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorBase.java
serge-rider: Logic is still incorrect. Why SQLEditorBase become DBPDataSourceContainerProvider ?
It mustn't return null, at least it should return dataSource.

serge-rider (changes requested):
https://github.com/dbeaver/dbeaver/pull/36056#discussion_r1818833826

kseniaguzeeva:
Verified

PR #36081, opened 2024-10-28
#36080 Improve clarity in JSON values export options

### LINKED ISSUE #36080: Improve clarity in JSON values export options
In this PR [#35948](https://github.com/dbeaver/dbeaver/pull/35948), a new ability was introduced to export JSON values as embedded JSON. The problem is that the property name can be confusing, and it’s unclear what will happen if it’s set to ‘false’.
image

Rename the property to “Export JSON values as” and add options: “String” and “Embedded JSON”.

ivgag:
image


ShadelessFox (changes requested):
This PR breaks backward compatibility with the now-removed option. Every saved task that has this option previously enabled will now behave differently.

arhayka:
> This PR breaks backward compatibility with the now-removed option. Every saved task that has this option previously enabled will now behave differently.

This option was added during the current sprint in the following ticket: https://github.com/dbeaver/pro/issues/3111.
Since it didn’t exist before, no tasks were previously saved with it.

ShadelessFox (approved):

arhayka (approved):

HocKu7 (approved):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterJSON.java
```diff
@@ -54,22 +54,25 @@ public class DataExporterJSON extends StreamExporterAbstract implements IDocumen
 
     public static final String PROP_FORMAT_DATE_ISO = "formatDateISO";
     public static final String PROP_PRINT_TABLE_NAME = "printTableName";
-    public static final String PROP_EXPORT_JSON_AS_STRING = "exportJsonAsString";
+    public static final String PROP_EXPORT_JSON_VALUES = "exportJsonValues";
+    public static final String PROP_EXPORT_JSON_VALUES_AS_STRING = "string";
+    public static final String PROP_EXPORT_JSON_VALUES_AS_JSON = "embedded JSON";
```
serge-rider: let's make it just `json`. It is not good to have s[paces in property values

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterJSON.java
ivgag: image

kseniaguzeeva:
Tested

PR #36097, opened 2024-10-29
Dbeaver/dbeaver#35411 fix tooltip length

### LINKED ISSUE #35411: PC crashes when tooltip text is too long
### Description

When hovering over a cell in a table and the tooltip text would be too long, it takes the whole screen and is truncated. This causes the entire PC to crash.

[dbeaver_tooltip_crash.log](https://github.com/user-attachments/files/16819122/dbeaver_tooltip_crash.log)

### DBeaver Version

24.1.5

### Operating System

Ubuntu 24.04

### Database and driver

Mysql 8.2.0

### Steps to reproduce

1. Open a table
2. open either the Properties or Data view
3. hover over a cell that has a lot of text content
4. Text is larger than the screen
5. result: crash

### Additional context

Tested back to version 23.3.5, still reproducible

HocKu7:


Matvey16:
Verified

PR #36098, opened 2024-10-29
#36091 PG database connection fix

### LINKED ISSUE #36091: Non public schema missing in PostgreSQL connections since 24.2.3
### Description

The recent dBeaver update has caused some problematic behavior in regards to being able to see custom schema.
Prior to this update expanding the schema tree looked like this:
dBeaver_all_schema_20241021_131506

Since the update, the exact same connection object with the exact same setting now looks like this:
dBeaver_new_schema_20241021_131506

The difference appears to be what database is listed in the Connection Settings ->Main->Database field.  Whichever database is listed in that field displays all of the schema.  The rest of the databases accessed via that connection only list the information_schema, pg_catalog, and public schema.  

Most of my PostgreSQL connections contain multiple databases, anywhere from two to ten databases.  Since upgrading to the current version of dBeaver I have to reset the connection to point to a different database, save the connection, and then reconnect.  Unfortunately this will now hide all of the non default schema of the previous database.

Is there anyway to revert to the previous behavior of showing all of the schema of all of the databases accessed via a connection?

Thanks.

### DBeaver Version

Community Edition 24.2.3

### Operating System

Windows 10

### Database and driver

PostgreSQL 15 / 16

### Steps to reproduce

1. have a PostgreSQL connection object
2. let there be multiple databases on that PostgreSQL instance
3. let there by multiple custom schema in each database
4. construct a dBeaver connection object with one of the databases as the value in Connection Settings->Main->Server->Database
5. check the Show all databases checkbox in the same location
6. establish the connection
7. notice that all of the schema are listed for the database from step 4 and only the default schema are shown for the rest of the databases.

### Additional context

_No response_

serge-rider:


Matvey16:
Verified

PR #36114, opened 2024-10-30
dbeaver/pro#3465 pg connection stress test

alexander-skoblikov:


plugins/org.jkiss.dbeaver.ext.test/src/org/jkiss/dbeaver/ext/test/handlers/HandlerConnectionStressTest.java
```diff
@@ -0,0 +1,145 @@
...
+        return null;
+    }
+
+    private void stressTest(DBPDataSource dataSource, int maxConnections) {
+        var contexts = new ArrayList<DBCExecutionContext>();
+        try {
+            IntStream.range(0, maxConnections).parallel()
+                .forEach(i -> {
+                        try {
+                            contexts.add(dataSource.getDefaultInstance()
```
ShadelessFox, line 86: Since it's `parallel` and the collection is not thread-safe, it might result in race conditions. I'd recommend doing all steps inside a single parallel stream instead of introducing an intermediate step.

PR #36125, opened 2024-10-31
Dbeaver/pro#3535 add export table name

HocKu7:


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/processor/FailedExportFileCleanerProcessor.java
```diff
@@ -0,0 +1,64 @@
...
+                             @NotNull Throwable error,
+                             @NotNull StreamTransferConsumer consumer,
+                             @Nullable DBTTask task,
+                             @NotNull Map<String, Object> settings
+    ) throws DBException {
+        for (Path outputFile : consumer.getOutputFiles()) {
+            try {
+                Files.deleteIfExists(outputFile);
+            } catch (IOException e) {
+                log.log(Level.WARNING, String.format("An error occurred while attempting to delete the output file: %s", outputFile.toAbsolutePath()));
```
ivgag, line 60: java.util.logging.Logger#warning(java.lang.String)
```suggestion
                log.warning("An error occurred while attempting to delete the output file: " + outputFile.toAbsolutePath());
```
Also, the info about the error is missed

ivgag (changes requested):

PR #36138, opened 2024-11-01
Dbeaver/dbeaver#35882 fix sql script execution

### LINKED ISSUE #35882: Execution of the script on disconnected state

HocKu7:


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorBase.java
```diff
@@ -1309,6 +1318,12 @@ protected boolean isNavigationTarget(Annotation annotation) {
         return super.isNavigationTarget(annotation);
     }
 
+    @Nullable
+    protected DBPDataSourceContainer getDataSourceContainerForSyntaxRuleReloading() {
```
ivgag, line 1326:
```suggestion
    protected DBPDataSourceContainer getDataSourceContainer() {
```
Is it possible to use this container for something else?

ivgag (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorBase.java
HocKu7: It is possible, but we got a bug with a simple 'get' name and decided to name it differently so it looks safe but awkward; I agree

MashaKorax:
verified

Issue #3232, opened 2018-03-28
dbeaver lost id keys !! & ltree bug!!


221V:
i use postgresql and have table like this:

```
CREATE TABLE "accounts" (
  "id" bigserial NOT NULL,
  "password" varchar(255),
  "other_etc" varchar(255),
...
```

i click "export table data", "sql", and get next rows:
```
INSERT INTO public.accounts (password,other_etc, ...
```
where are "id"s ???

i lost accounts on product !! (f*ck!! thanks for this!! )

------

other f*ck bug is in export data from table with using ltree
is we have table with
```
...
"upline" ltree,
...
```
we get export
```
... ,10000.10001.10003.10005, ...
```
but needs next valid data:
```
... ,'10000.10001.10003.10005', ...
```

thank you again!! (((

PR #36153, opened 2024-11-04
dbeaver/dbeaver#3232 Allow empty value in database tasks

HocKu7:
Now, the configuration with such an empty parameter will be saved correctly and will be displayed after restarting the application.
image


ShadelessFox (approved):

ivgag (approved):

Matvey16:
Verified

PR #36175, opened 2024-11-06
dbeaver/pro#3214 Improve SQL Editor commands enablement

ShadelessFox:
The truth table for command enablement depending on the active presentation:

image


HocKu7 (approved):

plugins/org.jkiss.dbeaver.ui.editors.sql/plugin.xml
```diff
@@ -576,7 +576,10 @@
             <enabledWhen>
                 <with variable="activeEditor">
                     <adapt type="org.jkiss.dbeaver.ui.editors.sql.SQLEditor">
-                        <test property="org.jkiss.dbeaver.ui.editors.sql.canExplain"/>
+                        <and>
+                            <test property="org.jkiss.dbeaver.ui.editors.sql.canExecute" value="script"/>
```
serge-rider, line 581: Why explain is enabled for script mode only?

ShadelessFox: The SQL editor is the only "presentation" that can work with multiple statements at the same time. This check ensures the "Explain Plan" command is not available in VQB or AI Chat.
I'd say it's a hacky approach since it's based on assumptions, but I didn't want to introduce a new property and instead fixed the existing one, reusing it in other places as well.

PR #36176, opened 2024-11-06
#36166 Preserve the order of row color mappings

### LINKED ISSUE #36166: Row colorization priorities seem random
### Description

How to sort the row colors? Two columns in a single row are trying to color the same row, how to choose which one ends up being the dominant color? I have the same table twice, but the behavior in both tables on this is different.

### DBeaver Version

Community Edition 24.2.4.202411031831

### Operating System

Windows 11

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

ShadelessFox:
Now the latest color of the latest attribute should be used as the final color.


HocKu7 (approved):

ChaoticEntertainment:
What would be the latest attribute? Would it be the latest one as initially defined, or would it use the table's custom column order? Custom column order would be preferred.

PR #36186, opened 2024-11-07
#36178 Condition expressions + content keys fix

### LINKED ISSUE #36178: Data changes can't be saved using Value Panel
### Description

When editing values in the Value panel in DBeaver, changes are not saved. This issue affects text fields and JSON fields. Previously, changes made in the Value panel could be saved directly, but now they require editing in the cell itself for changes to be applied. Additionally, after modifying a value in the Value panel, the Save icon remains unavailable, indicating that DBeaver doesn't recognize the change.
image

### DBeaver Version

24.2.4

### Operating System

Windows 11 23H2

### Database and driver

1. Database: PostgreSQL 15.8
2. Driver: postgres-jdbc

### Steps to reproduce

1. Open DBeaver and connect to the relevant database.
2. Navigate to the table with the fields (e.g., text or JSON) that need to be edited.
3. Open the Data Editor for this table.
4. Click on a field, open the Value panel, and make a change (e.g., modify text content or JSON).
5. Notice that the Save icon is not enabled, even though a modification was made in the Value panel.
6. Attempt to save the change (if applicable, commit the transaction if auto-commit is disabled).

### Additional context

_No response_

serge-rider:


MashaKorax:
verified

PR #36195, opened 2024-11-07
dbeaver/dbeaver#35808 Rename driver properties to Default properties

### LINKED ISSUE #35808: Rename `Driver properties` to `Default properties` in Connection settings
The name Driver properties may be confusing in the Connection configuration dialog. Let's rename it.
image

To do: 
- Rename Driver properties —> Default properties
- Translate this to all the languages (only the label)
- Add a hint : 
Here are the default values for all connections. These will be applied if the properties are not defined in the Driver properties within the Connection settings.

HocKu7:


plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/internal/UIConnectionMessages.properties
```diff
@@ -28,8 +28,8 @@ controls_client_homes_panel_label_product_name = Product Name
 controls_client_homes_panel_label_product_version = Product Version
 controls_client_homes_panel_link_message = See <a>how to configure local clients</a> in our wiki
 
-dialog_setting_connection_driver_properties_description = Driver properties
-dialog_setting_connection_driver_properties_title = Driver properties
+dialog_setting_connection_driver_properties_description = Here are the default values for all connections. These will be applied if the properties are not defined in the Driver properties within the Connection settings.
+dialog_setting_connection_driver_properties_title = Default properties
```
serge-rider: It looks like not what we wanted.
Default properties are on driver edit dialog. In the connection edit dialog we want to keep everything as is.

serge-rider (changes requested):

MashaKorax:
What about Cloudbeaver?
chrome_ZwLjwyX8OF

MashaKorax:
In Dbeaver  - > Connection configuration dialog - still named Driver properties

MashaKorax:
dbeaver_uboPridUyz

MashaKorax:
verified

Issue #18941, opened 2023-02-03
Export data window crashes occasionally


michal-bartak:
### Description

It happens from time o time. When choosing Export data feature, export window opens and throws exception.
It's not reproducible in 100% cases. Sometimes it works. Siometimes it works after DBeaver restart to throw exception with another try.

![image](https://user-images.githubusercontent.com/7868445/216619843-2a82a2ea-da5d-448c-a664-05d6481385ff.png)

Here are snipets of db log (complete log is too big to include)
```
!ENTRY org.eclipse.ui 4 0 2023-02-03 14:56:30.017
!MESSAGE Unhandled event loop exception
!STACK 0
java.lang.StackOverflowError
	at org.eclipse.core.runtime.Assert.isNotNull(Assert.java:76)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:43)
	at org.eclipse.core.commands.Command.fireCommandChanged(Command.java:522)
	at org.eclipse.core.commands.Command.lambda$0(Command.java:1002)
	at org.eclipse.core.commands.AbstractHandler.fireHandlerChanged(AbstractHandler.java:77)
	at org.eclipse.e4.core.commands.internal.HandlerServiceHandler.fireHandlerChanged(HandlerServiceHandler.java:180)
	at org.eclipse.ui.internal.handlers.E4HandlerProxy.handlerChanged(E4HandlerProxy.java:115)
	at org.eclipse.core.commands.AbstractHandler.fireHandlerChanged(AbstractHandler.java:77)
	at org.eclipse.ui.internal.handlers.HandlerProxy.lambda$1(HandlerProxy.java:372)
```

```
!ENTRY org.jkiss.dbeaver.model 4 0 2023-02-03 14:56:30.065
!MESSAGE Event loop exception
!SUBENTRY 1 org.jkiss.dbeaver.model 4 0 2023-02-03 14:56:30.065
!MESSAGE StackOverflowError
!STACK 0
java.lang.StackOverflowError
	at org.eclipse.core.runtime.Assert.isNotNull(Assert.java:76)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:43)
	at org.eclipse.core.commands.Command.fireCommandChanged(Command.java:522)
	at org.eclipse.core.commands.Command.lambda$0(Command.java:1002)
	at org.eclipse.core.commands.AbstractHandler.fireHandlerChanged(AbstractHandler.java:77)
[...]

PR #36199, opened 2024-11-08
dbeaver/dbeaver#18491#11778-sqlite-fix-multiPK

### LINKED ISSUE #11778: Column cannot be dropped if primary key exists
#### System information: 
- Operating system (distribution) and version win10
- DBeaver version 21.0.1

- Database name and version SQLite

If table has primary key, column from it cannot be dropped - error appears

1. Create table
```sql
CREATE TABLE "employees"
(
    [EmployeeId] INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL,
    [LastName] NVARCHAR(20)  NOT NULL,
    [FirstName] NVARCHAR(20)  NOT NULL,
    [Title] NVARCHAR(30),
    [ReportsTo] INTEGER,
    [BirthDate] DATETIME,
    [HireDate] DATETIME,
    [Address] NVARCHAR(70),
    [City] NVARCHAR(40),
    [State] NVARCHAR(40),
    [Country] NVARCHAR(40),
    [PostalCode] NVARCHAR(10),
    [Phone] NVARCHAR(24),
    [Fax] NVARCHAR(24),
    [Email] NVARCHAR(60),
    FOREIGN KEY ([ReportsTo]) REFERENCES "employees" ([EmployeeId]) 
		ON DELETE NO ACTION ON UPDATE NO ACTION
);

CREATE INDEX [IFK_EmployeeReportsTo] ON "employees" ([ReportsTo]);

```
2. Try to delete e.g. Fax column using UI

**Actual result**
image

In case of deleting column with related PK or/and FK:
image

### LINKED ISSUE #18491: The SQL generated for SQLite creates the primary key twice on save (persist)
### Description

When modifying the SQLite table with dbeaver's graphical interface for an existing table with a primary key, the SQL generated when saving generates PRIMARY KEY constraint twice. 

### DBeaver Version

Community Edition 22.2.5.202211202223

### Operating System

Windows 11 Pro

### Database and driver

SQLite JDBC driver (couldn't find version) https://github.com/xerial/sqlite-jdbc

### Steps to reproduce

- Create a table with a primary key in SQLite using dbeaver graphical interface
- Save it and without closing the tab (not sure if bug persists after closing the tab) create a new foreign key. 
- Now try saving again. This time, the SQL generated would be 
```SQL
-- Create foreign key message_receiver_FK

CREATE TEMPORARY TABLE temp AS
SELECT *
FROM messages;

DROP TABLE messages;

CREATE TABLE messages (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    sender_id INTEGER,
    receiver_id INTEGER,
    CONSTRAINT MESSAGES_PK PRIMARY KEY (id),
    CONSTRAINT messages_FK FOREIGN KEY (sender_id) REFERENCES users(id) ON DELETE SET NULL ON UPDATE CASCADE,
    CONSTRAINT message_receiver_FK FOREIGN KEY (receiver_id) REFERENCES users(id) ON DELETE SET NULL ON UPDATE CASCADE
);

INSERT INTO messages
SELECT *
FROM temp;

DROP TABLE temp;
```

As you can see, the constraint `MESSAGES_PK` is declaring PRIMARY KEY twice which would result in an error when clicking persist.

### Additional context

_No response_

### LINKED ISSUE #21653: [SQLITE_ERROR] SQL error or missing database (table "Frames" has more than one primary key)
### Description

When adding a foreign key to a table the following code is generated that has both the `PRIMARY KEY` on the `ID` keyword and the a constraint for the same `ID`

```CREATE TEMPORARY TABLE temp AS
SELECT *
FROM Frames;

DROP TABLE Frames;

CREATE TABLE Frames (
	ID INTEGER PRIMARY KEY AUTOINCREMENT,
	Designation TEXT,
	Description TEXT,
	Corners INTEGER,
	FrameType_ID INTEGER,
	CONSTRAINT FRAMES_PK PRIMARY KEY (ID),
	CONSTRAINT Frames_FK FOREIGN KEY (FrameType_ID) REFERENCES FrameTypes(ID)
);

INSERT INTO Frames
SELECT *
FROM temp;

DROP TABLE temp;
```

**Error log**

```!ENTRY org.jkiss.dbeaver.model 4 0 2023-10-25 16:00:43.796
!MESSAGE [SQLITE_ERROR] SQL error or missing database (table "Frames" has more than one primary key)
!STACK 0
org.sqlite.SQLiteException: [SQLITE_ERROR] SQL error or missing database (table "Frames" has more than one primary key)
	at org.sqlite.core.DB.newSQLException(DB.java:1179)
	at org.sqlite.core.DB.newSQLException(DB.java:1190)
	at org.sqlite.core.DB.throwex(DB.java:1150)
	at org.sqlite.core.NativeDB.prepare_utf8(Native Method)
	at org.sqlite.core.NativeDB.prepare(NativeDB.java:132)
	at org.sqlite.core.DB.prepare(DB.java:264)
	at org.sqlite.jdbc3.JDBC3Statement.lambda$execute$0(JDBC3Statement.java:53)
	at org.sqlite.jdbc3.JDBC3Statement.withConnectionTimeout(JDBC3Statement.java:459)
	at org.sqlite.jdbc3.JDBC3Statement.execute(JDBC3Statement.java:42)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCStatementImpl.execute(JDBCStatementImpl.java:330)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCStatementImpl.executeStatement(JDBCStatementImpl.java:131)
	at org.jkiss.dbeaver.model.exec.DBExecUtils.executePersistAction(DBExecUtils.java:358)
	at org.jkiss.dbeaver.model.impl.edit.AbstractObjectManager.executePersistAction(AbstractObjectManager.java:40)
	at org.jkiss.dbeaver.model.impl.edit.AbstractCommandContext.executeCommands(AbstractCommandContext.java:206)
	at org.jkiss.dbeaver.model.impl.edit.AbstractCommandContext.saveChanges(AbstractCommandContext.java:122)
	at org.jkiss.dbeaver.ui.editors.entity.EntityEditor.lambda$1(EntityEditor.java:384)
	at org.jkiss.dbeaver.model.exec.DBExecUtils.tryExecuteRecover(DBExecUtils.java:190)
	at org.jkiss.dbeaver.ui.editors.entity.EntityEditor.saveCommandContext(EntityEditor.java:382)
	at org.jkiss.dbeaver.ui.editors.entity.EntityEditor$SaveJob.run(EntityEditor.java:1253)
``` 

### DBeaver Version

Community Edition 23.2.3

### Operating System

Windows 10

### Database and driver

Database: 
SQLite

Driver: 
sqlite-jdbc-3.43.3.1
slf4j-api-1.7.36

### Steps to reproduce

1. Create a reference table (FrameTypes)
`CREATE TABLE FrameTypes (
	ID INTEGER PRIMARY KEY AUTOINCREMENT,
	Designation TEXT(50),
	Description TEXT(250),
	Ordinal INTEGER,
	IsValid INTEGER DEFAULT (1)
);`
2. Add an ID (int) as the primary key
3. Create another table (Frames)
4. Add an ID (int) as the primary key
5. Add another column to be used as the foreign key of type int
`CREATE TABLE Frames (
	ID INTEGER PRIMARY KEY AUTOINCREMENT
, Designation TEXT(100), Description TEXT(250), Corners INTEGER, FrameType_ID INTEGER);`
7. Save the changes
8. Create a foreign key using the two ID's
9. Save the changes again
10. The error occurs.

### Additional context

_No response_

eusebe-cda:
Fix issues #18491 & #11778 : modifying a sqlite table with autoincrement column generate DDL declaring PK twice


eusebe-cda:
This pull request would also fix this third issue : #21653

To prevent the generation of a separete DDL instruction for the constraint when the PK is auto-incremented in SQLite, I suggest treating the constraint as "inherited", even though this is not really the case...

E1izabeth:
Thanks for contribution. We will review your PR and let you know

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.sqlite/src/org/jkiss/dbeaver/ext/sqlite/model/SQLiteUniqueKey.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+
+public class SQLiteUniqueKey extends GenericUniqueKey implements DBPInheritedObject {
+
+    public SQLiteUniqueKey(GenericTableBase table, String name, @Nullable String remarks, DBSEntityConstraintType constraintType, boolean persisted) {
+        super(table, name, remarks, constraintType, persisted);
+    }
+
+    @Override
+    public boolean isInherited() {
```
ShadelessFox: Wouldn't it make more sense to override `SQLTableManager#isIncludeIndexInDDL` and put this logic there? This feels hacky, indeed.

eusebe-cda: Thanks for your review. The SQLTableManager#isIncludeIndexInDDL method cannot be overridden to correct this issue: in the SQLTableManager#getTableDDL the private method skipObject is used instead of isIncludeIndexInDDL to check if the primary key must be added or not...

Another solution could be to override SQLTableManager#getTableDDL to modify the DBPScriptObject.OPTION_SKIP_UNIQUE_KEYS option before generate the DLL, what do you think?

ShadelessFox: You're referring to this part in the code, right?
https://github.com/dbeaver/dbeaver/blob/c40f4d5e4162e573f055ab3a72dd129517ae9034/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableManager.java#L263-L275

You can add a new method similar to the one that checks an index for inclusion but for PKs. It's totally fine.

eusebe-cda: > You're referring to this part in the code, right?
> 
> https://github.com/dbeaver/dbeaver/blob/c40f4d5e4162e573f055ab3a72dd129517ae9034/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableManager.java#L263-L275

Yes

> You can add a new method similar to the one that checks an index for inclusion but for PKs. It's totally fine.

Ok, I'll try that, I wanted to avoid modifying generic classes to limit the impacts...

eusebe-cda: > You can add a new method similar to the one that checks an index for inclusion but for PKs. It's totally fine.

That's what I did in the last commit...

plugins/org.jkiss.dbeaver.ext.sqlite/src/org/jkiss/dbeaver/ext/sqlite/edit/SQLiteTableManager.java
```diff
@@ -78,6 +81,16 @@ public void renameObject(@NotNull DBECommandContext commandContext, @NotNull Gen
         processObjectRename(commandContext, object, options, newName);
     }
 
+    @Override
+    protected boolean isIncludeConstraintInDDL(DBRProgressMonitor monitor, DBSEntityConstraint constraint) {
+        if (constraint.getConstraintType() == DBSEntityConstraintType.PRIMARY_KEY && constraint instanceof GenericUniqueKey) {
+            if (((GenericUniqueKey)constraint).getAttributeReferences(monitor).size() == 1 && ((GenericUniqueKey)constraint).getAttributeReferences(monitor).get(0).getAttribute().isAutoIncrement()) {
```
ShadelessFox: Just a nitpick:
```suggestion
        if (constraint.getConstraintType() == DBSEntityConstraintType.PRIMARY_KEY && constraint instanceof GenericUniqueKey key) {
            if (key.getAttributeReferences(monitor).size() == 1 && key.getAttributeReferences(monitor).get(0).getAttribute().isAutoIncrement()) {
```
Additionally, I'd extract the result of `getAttributeReferences` into a separate variable to improve readability.

eusebe-cda: Thanks for review and advice, and sorry for the code, I'm a very casual java developer ;)
I just pushed a new version which takes into this comments account.

ShadelessFox (approved):

ShadelessFox:
Thanks for the contribution! Passing the PR to our QA team now.

uslss:
#11778 - ok
#18941 - ok
#21653 - ok
verified

PR #36211, opened 2024-11-09
#26711 FK create fix: set proper data type for new column

### LINKED ISSUE #26711: Create fk working incorrect
### Description

there is no way to select a unique key for the connection.

In versions 23* it worked

### DBeaver Version

community edition 24.0.3

### Operating System

windows 11 64bit

### Database and driver

Postrgresql 14

### Steps to reproduce

1.create table with pk and unique index
CREATE TABLE public.newtable_1 (
	c int4 NOT NULL,
	c2 varchar NOT NULL,
	CONSTRAINT newtable_1_pk PRIMARY KEY (c)
);
CREATE UNIQUE INDEX newtable_1_c2_idx ON public.newtable_1 USING btree (c2);

2.create table2
CREATE TABLE public.newtable_2 (
	ff varchar NULL
);

try to create a foreign key on a table using a unique index.

image

There is no option to select a unique index for a connection

### Additional context

_No response_

serge-rider:


kseniaguzeeva:
SQLite: when I create a new column for FK, the error appears at first. Separate ticker will be created
Oracle/Firebird don't support unique indexes for FK and throw the error on attempt to use them.

PR #36214, opened 2024-11-10
#36212 data editor cell hints

### LINKED ISSUE #36212: Data view: show cell hints
It is convenient to show hints for cell values depending on values themselves.
Hint can be rendered with disabled color to the right of value itself.
Hints can be enabled and disabled for all columns or for a particular columns.

1. Array size
2. Binary data size
3. Referenced tables links
4. Dictionary titles (PRO)
It is convenient to see FK referenced dictionary value inlined in grid mode. 
Currently we can see just IDs, value panel helps but it show only one value. When work with grid as with a spreadsheet it is very inconvenient.

serge-rider:


Matvey16:
Hints change color if `colorize data types` is enabled

Matvey16:
Verified

AndyWu2015:
Sorry to bother you, I just want to know if i can close the effect that cell border will show when horver. I don't think it is necessary.
And I don't find prefs to disable this. I also found it is hard coded in   https://github.com/dbeaver/dbeaver/pull/36214/commits/b1a501dc7cafcc5b453a9519384f6d61322973d3

PR #36218, opened 2024-11-11
dbeaver/pro#3507 Remove DynamoDB and Firestore driver teaser

E1izabeth:
This 
image
 should not be shown in any DBeaver editions


MashaKorax:
The same issue with Firestore in Le. Do we want remove it in separate ticket?
dbeaver_DRmKW5Ii0I

MashaKorax:
verified

PR #36229, opened 2024-11-12
dbeaver/pro#3517 Support different begin-end blocks for BigQuery

E1izabeth:
Now it's correctly extracted. You can check it by folding
image

### For QA team
Please check different types of queries, including LOOPS, IFs, BEGIN END, CASE in different combinations
See docs https://cloud.google.com/bigquery/docs/reference/standard-sql/procedural-language#loops


uslss:
verified

PR #36231, opened 2024-11-13
Dbeaver/pro#3644 connection cancel

serge-rider:


DariaMarkaryan:
Verified with auto tests

mr-anton-t:
Verified with e2e tests

PR #36238, opened 2024-11-13
#36208 Fix invalid thread access on editing data in multiple results in one tab mode

### LINKED ISSUE #36208: Exception `Invalid thread access` occurs on changing data in Result tab when `Multiple results in a single tab` = true
### Description

The same happens on updating/deleting the existing rows as well (on pressing Save button).

https://github.com/user-attachments/assets/8e653724-454f-4196-b145-e72a6ac24087

### DBeaver Version

Enterprise Edition 24.2.4-202410172322

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.4

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Should not happen anymore


MashaKorax:
verified

PR #36249, opened 2024-11-15
Fix typo

anisalibegic:
Found a typo in one of the error messages so I fixed it.


ShadelessFox:
Thanks for the contribution. Unfortunately, we don't accept minor fixes like the proposed one.

You still can help us make DBeaver better by:
- Fixing a bug https://github.com/dbeaver/dbeaver/issues?q=is%3Aissue+is%3Aopen+label%3Abug
- Localizing it to your native language or improving existing localization https://github.com/dbeaver/dbeaver/wiki/Localization

PR #36252, opened 2024-11-15
dbeaver/pro#3620 Introduce JUL handler

ivgag:
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/JULHandler.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/JULHandler.java
@@ -1,0 +1,70 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2024 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver;
+
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+
+/**
+ * Implementation of {@link Handler} that redirects JUL logging to DBeaver log.
+ */
+public class JULHandler extends Handler {
+    private final Log log;
+
+    public JULHandler(Log log) {
+        this.log = log;
+    }
+
+    @Override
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/JULHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public void publish(LogRecord record) {
+        log.log(
+                new Status(
+                        getSeverity(record.getLevel()),
+                        record.getLoggerName(),
+                        record.getMessage(),
+                        record.getThrown()
+                )
+        );
```
ShadelessFox:
```suggestion
        log.log(new Status(
            getSeverity(record.getLevel()),
            record.getLoggerName(),
            record.getMessage(),
            record.getThrown()
        ));
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/JULHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public void flush() {
+
```
ShadelessFox:
```suggestion
        // no op
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/JULHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @Override
+    public void close() throws SecurityException {
+
```
ShadelessFox:
```suggestion
        // no op
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/JULHandler.java
```diff
@@ -0,0 +1,66 @@
...
+package org.jkiss.dbeaver;
+
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+
+import java.util.logging.Handler;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+
+public class JULHandler extends Handler {
```
ShadelessFox, line 29: Please write a short Javadoc describing what this handler is used for.

ShadelessFox:
I don't like that you're using INFO logging level in the driver. They bloat the Error Log view which is used for tracking issues like warnings and errors. Can it be fixed within a separate ticket?

ShadelessFox (approved):

E1izabeth (approved):

ivgag:
> I don't like that you're using INFO logging level in the driver. They bloat the Error Log view which is used for tracking issues like warnings and errors. Can it be fixed within a separate ticket?

So, if you think that INFO logs bloat the Error Log view, which is meant for tracking warnings and errors, perhaps INFO logs shouldn’t be included in the Error Log? I don’t see a reason to avoid using the INFO logging level in the driver.

E1izabeth:
Error Log view has Filters actually 
image

Matvey16:
Verified

serge-rider:
Anyway, I agree that spamming info messages in Error Log view is not a good idea.
Logging above looks more like debug log.

Matvey16:
Verified

E1izabeth (approved):

PR #36262, opened 2024-11-18
#36260 Fix table meta lookup statement when `Server side object filters` setting is off

### LINKED ISSUE #36260: Error refreshing node
### Description

When clicking refresh on the table view page, an error message will appear.
image

I checked the SQL statement that reported the error. 

`/* ApplicationName=DBeaver 24.2.5 - Metadata */ SHOW FULL TABLES FROM iyuu = 'cn_client'`
maybe replace '=' with 'like',it will be work

### DBeaver Version

CE 24.2.5.202411171748

### Operating System

Windows11 23H2

### Database and driver

mysql 8.0 & mysql 5.7
driver: mysql8

### Steps to reproduce

1. open table view
2. show table properties
3. click refresh button
4. you can see

### Additional context

_No response_

E1izabeth:
Now we should send the correct query with LIKE op instead of =


ShadelessFox (commented):
I'm not sure how this fixes the issue. Can you elaborate, please?

E1izabeth:
`show full tables from` statement doesn't work with = operator
It works this way `show full tables from someName like 'abcd'`, but not `show full tables from somename = 'abcd'`

We correctly handled it previously, but broke in refactoring in https://github.com/dbeaver/dbeaver/pull/20368

ShadelessFox (approved):

uslss:
verified

PR #36273, opened 2024-11-18
#36022 Add BOOLEAN to the predefined types for Oracle

### LINKED ISSUE #36022: Setting a boolean value in the inline data editor fails with binding not supported (Oracle DB)
### Description

I have the following table in an Oracle:
```
CREATE TABLE USERS (
	ID NUMBER(38,0),
	EMAIL VARCHAR2(4000),
	PASSWORD VARCHAR2(4000),
	ENABLED BOOLEAN DEFAULT 0
);
```

I can insert/update a boolean for ENABLED easily when using the SQL text editor:
```
insert into users (email, password, enabled) values ('a@test.com', 'asdf', false);
update users set enabled = true where email = 'a@test.com';
```

The same fails when I want to insert or update such a row with the data editor:
image

It seems that the column is not recognized as boolean at all, since normally (Postgres) I can just click a checkbox to change boolean values.

### DBeaver Version

24.2.2.202410061621 Community Edition

### Operating System

macOS 15 Sequoia

### Database and driver

Database name and version taken from Docker Hub: `image: gvenzl/oracle-free:23.5-slim-faststart`

DBeaver DB driver: 
image

### Steps to reproduce

1. Start the database with `docker compose up` (Docker compose file below)
```
services:

  oracle:
    image: gvenzl/oracle-free:23.5-slim-faststart
    environment:
      ORACLE_DATABASE: app
      ORACLE_PASSWORD: password
      APP_USER: root
      APP_USER_PASSWORD: password
    ports:
      - "1521:1521"
```

2. Connect to it with DBeaver: connection string = `jdbc:oracle:thin:@//localhost:1521/app` (I kept all default settings for other connection parameters) 
image

3. create table 
```
CREATE TABLE USERS (
	ID NUMBER(38,0),
	EMAIL VARCHAR2(4000),
	PASSWORD VARCHAR2(4000),
	ENABLED BOOLEAN DEFAULT 0
);
```

4. Go to the data editor and try to insert a row

It fails saying:

```
Error synchronizing data with database

Reason:
Parameter [ENABLED] value [true] binding not supported
```

### Additional context

_No response_

E1izabeth:
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleDataType.java
+++ b/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleDataType.java
@@ -63,25 +63,40 @@ public class OracleDataType extends OracleObject<DBSObject>
         final int precision;
         final int minScale;
         final int maxScale;
-        private TypeDesc(DBPDataKind dataKind, int valueType, int precision, int minScale, int maxScale)
-        {
+        final int serverAtLeastMajor;
+        final int serverAtLeastMinor;
+
+        private TypeDesc(@NotNull DBPDataKind dataKind, int valueType, int precision, int minScale, int maxScale) {
+            this(dataKind, valueType, precision, minScale, maxScale, -1, -1);
+        }
+
+        private TypeDesc(
+            @NotNull DBPDataKind dataKind,
+            int valueType,
+            int precision,
+            int minScale,
+            int maxScale,
+            int serverAtLeastMajor,
+            int serverAtLeastMinor
+        ) {
             this.dataKind = dataKind;
             this.valueType = valueType;
             this.precision = precision;
             this.minScale = minScale;
             this.maxScale = maxScale;
+            this.serverAtLeastMajor = serverAtLeastMajor;
+            this.serverAtLeastMinor = serverAtLeastMinor;
         }
     }
 
     static final Map<String, TypeDesc> PREDEFINED_TYPES = new HashMap<>();
-    static final Map<Integer, TypeDesc> PREDEFINED_TYPE_IDS = new HashMap<>();
     static  {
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleDataType.java
```diff
@@ -81,6 +81,7 @@ private TypeDesc(DBPDataKind dataKind, int valueType, int precision, int minScal
         PREDEFINED_TYPES.put("BINARY_DOUBLE", new TypeDesc(DBPDataKind.NUMERIC, Types.DOUBLE, 38, 127, -84));
         PREDEFINED_TYPES.put("BINARY_FLOAT", new TypeDesc(DBPDataKind.NUMERIC, Types.FLOAT, 38, 127, -84));
         PREDEFINED_TYPES.put("BLOB", new TypeDesc(DBPDataKind.CONTENT, Types.BLOB, 0, 0, 0));
+        PREDEFINED_TYPES.put("BOOLEAN", new TypeDesc(DBPDataKind.BOOLEAN, Types.BOOLEAN, 0, 0, 0));
```
serge-rider: Boolean was added in Oracle 23c only.
Let's make this conditional

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleDataType.java
E1izabeth: I'll need to change the static initialization to instance in this case. I decided not to do it, because it doesn't break something on older versions. However, if you think it's matter, OK

ShadelessFox (approved):

Matvey16:
Verified

PR #36282, opened 2024-11-19
#35944 Array parenthesis customization

### LINKED ISSUE #35944: Wrong parentheses type in nested data export/sql generation
### Description

This happened after some fixes for nested data representation, maybe this one - https://github.com/dbeaver/pro/issues/3313.
Test on this table
```sql
CREATE TABLE default.TEST
(
    `Column1` Array(Int64)
)
ENGINE = Log;
INSERT INTO `default`.TEST
(Column1)
VALUES([1]);
```
the correct Insert sql for arrays in Clickhouse is
```sql
INSERT INTO `default`.TEST
(Column1)
VALUES([1]);
```
but in exported file (export the table to SQL format) or in generated sql (in context menu of the table Generate SQL -> INSERT) we get the wrong type of parentheses, so if we'll this script to the sql editor and execute, then synthax error will be thrown
```sql
INSERT INTO `default`.TEST 
(Column1) 
VALUES
({1});
```
Don't sure if the parentheses must be the same type in sql export and other like txt, dbunit, scv, but take a look at these formats too. Also can't add a value to an array from the data editor if the table is empty, but maybe these problems are not connected.

### DBeaver Version

24.2.3

### Operating System

win10

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

serge-rider:


plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/data/JDBCCollection.java
```diff
@@ -130,7 +130,7 @@ public String toString() {
     }
 
     @NotNull
-    public String makeArrayString() {
+    public String makeArrayString(char ... brackets) {
```
ShadelessFox, line 133: Why is it a variadic-length parameters? Pass two parameters: for the left bracket and the right one.

DariaMarkaryan:
verified

PR #36285, opened 2024-11-19
dbeaver/dbeaver#36237 Fix rollback on data transfer fail

### LINKED ISSUE #36237: Incorrect transaction commit/rollback during data import
### Description

It seems as transaction work incorrectly during insert as in some cases queries are commited when they are not supposed to. The truncation of the table is also done in separate transaction (if in a transaction at all)

### DBeaver Version

24.2.5

### Operating System

Any

### Database and driver

PostgreSQL, SQLite

### Steps to reproduce

1. Create a csv file that has incompatible data on second on a third row (e.g. string in a correcponding numeric column)
2. Open import wizard for an empty existing table
3. In data load settings check `Use transactions` and set a commit after 10000 rows
4. Start the import
5. On an expected error with incorrect data click **Stop**

**Expected**: Nothing is imported into table because of the rollback
**Actual:** First few rows before the row with an error are imported

### Additional context

image

HocKu7:


Matvey16:
Verified

PR #36295, opened 2024-11-20
dbeaver/pro#3667 Add value handler for timestamp

ShadelessFox:


plugins/org.jkiss.dbeaver.ext.sqlite/src/org/jkiss/dbeaver/ext/sqlite/model/data/SQLiteValueHandlerProvider.java
```diff
@@ -39,6 +41,9 @@ public DBDValueHandler getValueHandler(DBPDataSource dataSource, DBDFormatSettin
         if (dataKind == DBPDataKind.BINARY) {
             return JDBCContentValueHandler.INSTANCE;
         }
+        if (dataKind == DBPDataKind.DATETIME) {
```
serge-rider: Not sure about that..

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.sqlite/src/org/jkiss/dbeaver/ext/sqlite/model/data/SQLiteValueHandlerProvider.java
ShadelessFox: Why?

PR #36296, opened 2024-11-20
#36162 Restore the “Show all databases” checkbox for Azure SQL

### LINKED ISSUE #36162: Why DBeaver 24.2.1 made this change? Azure SQL: “show all databases” option was removed
Hi,

In DBeaver 24.2.1 it can be seen that Azure SQL: “show all databases” option was removed. My question is why? We need that option to be able to see schemas in our Azure Synapse, and I believe some other people can have the same problem. We had to downgrade all our software because of that. We have DBeaver Enterprise licenses.

Thanks in advance,

Unai

ivgag:
Restore the “Show all databases” checkbox for Azure SQL and provide a detailed error message for cross-database queries.

image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/SQLServerUtils.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/SQLServerUtils.java
@@ -48,6 +48,7 @@ import org.jkiss.utils.CommonUtils;
 
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 /**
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/SQLServerUtils.java
```diff
@@ -403,4 +407,14 @@ public static JDBCPreparedStatement prepareTableStatisticLoadStatement(@NotNull
         return dbStat;
     }
 
+    public static DBException mapException(DBException e) {
```
ShadelessFox: Annotations pls :(

ShadelessFox (approved):

uslss:
verified

PR #36302, opened 2024-11-20
dbeaver/pro#3664 Disable `Open separate connection` setting on SQL Editor and Metadata pages for embedded databases

E1izabeth:
image


uslss:
verified

PR #36303, opened 2024-11-20
#36298 [CUBRID] Foreign Key Column Modification

### LINKED ISSUE #36298: [CUBRID] Foreign Key Column Modification
### Description

In the current implementation, modifications to foreign key columns' comments are not possible.

And what we will fix here is:
1. Allow modifications to the comment using the COMMENT ON COLUMN syntax.
2. Prevent any modifications other than comment changes to foreign key columns.
3. Display a warning dialog when users attempt to modify foreign key columns in the table view, indicating that only the comment can be edited.

### DBeaver Version

DBeaver version 24.2.1

### Operating System

Window 10

### Database and driver

1. Database Name and Version: CUBRID 11.2
2. Driver Name and Version: CUBRID JDBC Driver 11.2

### Steps to reproduce

1. Connect to Cubrid database
2. Click on Table folder
3. Select one of Cubrid table that contains foreign key
4. Click on column that is a foreign key
5. Try to modify comment and click save

### Additional context

_No response_

longhaseng52:
In the current implementation, modifications to foreign key columns' comments are not possible.

And here is what we have fixed:
1. Allow modifications to the comment using the COMMENT ON COLUMN syntax.
2. Prevent any modifications other than comment changes to foreign key columns.
3. Display a warning dialog when users attempt to modify foreign key columns in the table view, indicating that only the comment can be edited.


ShadelessFox (commented):
You can turn off property editing based on certain conditions. I think this is better than showing a message every time a user alters a key.

See where the `editableExpr` of `@Property` is used.

longhaseng52:
> You can turn off property editing based on certain conditions. I think this is better than showing a message every time a user alters a key.
> 
> See where the `editableExpr` of `@Property` is used.

Thank you for your suggestion, I have updated as your comment.

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridTableColumn.java
```diff
@@ -31,16 +31,19 @@
...
     final static String [] customTypes = {"ENUM", "JSON"};
+    public boolean isForeignKey;
 
     public CubridTableColumn(
             @NotNull GenericTableBase table,
             @Nullable String columnName,
             @Nullable String dataType,
-            @Nullable Boolean autoIncrement,
+            @Nullable boolean autoIncrement,
+            @Nullable boolean isForeignKey,
```
ShadelessFox: You may not annotate primitives with nullability annotations. They can't be null.
```suggestion
            boolean autoIncrement,
            boolean isForeignKey,
```

ShadelessFox (commented):
Looks good. Did you test if this works?

longhaseng52:
> Looks good. Did you test if this works?

Yes, I have tested on my local and it works fine.

ShadelessFox (approved):

E1izabeth (approved):

E1izabeth:
Thanks for the contribution! I passed this PR to our QA team for testing

kseniaguzeeva:
Unfortunately, it's impossible to see table properties - the error appears (see enc. log).

https://github.com/user-attachments/assets/e5a0dc8a-8bab-4a2e-b632-49eb6ef74efb

[1.log](https://github.com/user-attachments/files/18126385/1.log)

longhaseng52:
> Unfortunately, it's impossible to see table properties - the error appears (see enc. log).
> 
>  dbeaver_DL5k1d5Jq5.mp4 
> [1.log](https://github.com/user-attachments/files/18126385/1.log)

Hi @kseniaguzeeva

Does this error appear on every table? And could you please provide the details of the CUBRID server version you are using?
Thank you.

kseniaguzeeva:
Server version: CUBRID 11.0.13.0378
I tried tables with a key and without including the example from [the documentation](https://www.cubrid.org/manual/en/9.3.0/sql/schema/table.html):
```
-- creating two tables where one is referring to the other
CREATE TABLE a_tbl (
    id INT NOT NULL DEFAULT 0 PRIMARY KEY,
    phone VARCHAR(10)
);

CREATE TABLE b_tbl (
    ID INT NOT NULL,
    name VARCHAR (10) NOT NULL,
    CONSTRAINT pk_id PRIMARY KEY (id),
    CONSTRAINT fk_id FOREIGN KEY (id) REFERENCES a_tbl (id)
    ON DELETE CASCADE ON UPDATE RESTRICT
);

INSERT INTO a_tbl VALUES (1,'111-1111'), (2,'222-2222'), (3, '333-3333');
INSERT INTO b_tbl VALUES (1,'George'),(2,'Laura'), (3,'Max');
SELECT a.id, b.id, a.phone, b.name FROM a_tbl a, b_tbl b WHERE a.id = b.id;
```

longhaseng52:
> Server version: CUBRID 11.0.13.0378 I tried tables with a key and without including the example from [the documentation](https://www.cubrid.org/manual/en/9.3.0/sql/schema/table.html):
> 
> ```
> -- creating two tables where one is referring to the other
> CREATE TABLE a_tbl (
>     id INT NOT NULL DEFAULT 0 PRIMARY KEY,
>     phone VARCHAR(10)
> );
> 
> CREATE TABLE b_tbl (
>     ID INT NOT NULL,
>     name VARCHAR (10) NOT NULL,
>     CONSTRAINT pk_id PRIMARY KEY (id),
>     CONSTRAINT fk_id FOREIGN KEY (id) REFERENCES a_tbl (id)
>     ON DELETE CASCADE ON UPDATE RESTRICT
> );
> 
> INSERT INTO a_tbl VALUES (1,'111-1111'), (2,'222-2222'), (3, '333-3333');
> INSERT INTO b_tbl VALUES (1,'George'),(2,'Laura'), (3,'Max');
> SELECT a.id, b.id, a.phone, b.name FROM a_tbl a, b_tbl b WHERE a.id = b.id;
> ```

Thank you. let me check this and get back to you soon.

longhaseng52:
Hi @kseniaguzeeva 

I have fixed the error. It works fine on my local now. 
Let me know if there is still any error. Thank you.

kseniaguzeeva:
Comment to FK can be added, edited but can't be deleted. Do you prefer to fix it here or separately? You can use the tables above for test 

https://github.com/user-attachments/assets/13883911-a4b9-4d99-a08c-a88f562218b0

longhaseng52:
> Comment to FK can be added, edited but can't be deleted. Do you prefer to fix it here or separately? You can use the tables above for test
> 
>  dbeaver_9QbO9Bk1z0.mp4

Hi @kseniaguzeeva 

I prefer to fix it here in this PR. Thank you

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridUser.java
```diff
@@ -199,7 +199,21 @@ protected GenericTableColumn fetchChild(
                     }
                 }
             }
-            return new CubridTableColumn(table, columnName, showDataType == null ? dataType : showDataType, autoIncrement, dbResult);
+            boolean isForeignKey = false;
```
serge-rider: It looks like a very expensive way to read metadata. It executes separate SQL query for each and every table column. If you have 1000 columns in a certain table schema then it will execute 1000 additional queries just to read this schema meta.

Let's rewrite the algorithm. Perfect way is to read all columns in a single SQL query (which is constructed in prepareChildrenStatement in TableCache). Currently we use JDBC database metadata API for Cubrid - we can write our custom query instead.
Alternatively we can file a separate SQL query AFTER all table columns were read from JDBC API, e.g. in overloaded loadChildren function. It will read extra properties for all table/schema columns.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridUser.java
longhaseng52: @serge-rider,
I have optimized metadata retrieval by combining both queries into a single query to improve performance. Thank you

longhaseng52:
@serge-rider
I've pushed another commit corresponding to the updated query.

kseniaguzeeva:
verified

PR #36379, opened 2024-11-26
Dbeaver/pro#3714 fix warning spaming in logs

HocKu7:


E1izabeth (approved):

ShadelessFox (approved):
Kinda dirty, but okay

PR #36397, opened 2024-11-28
 #36367 Fixed Row colour issue when querying multiple tables.

### LINKED ISSUE #36367: "Set row color" is not working for non single result tab
### Description

STR
Execute sqript like 
```
Select * From Artist;
Select * From Album;
```

Open context menu for tab Album -> View/Format --> Set row color
 Actual result: rows are NOT colored in the results tab for the second tab. Rows colored only for first tab

https://github.com/user-attachments/assets/f74b29e1-455e-4313-97f5-75869c16b9a0

### DBeaver Version

24.3.0

### Operating System

mac 15.1

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

coderkill:
This fix extends to coloring more than 2 tables.

Logic:
By passing DBVEntity as parameter we are using that entity to identify the table. The previous one just used the first table as the virtual entity and failed to color it (This is because the model is linked to the first table and not the others).  The new logic keeps the old logic but colors the correct virtual entity instead of the model's virtual entity (which is only the 1st table)

Proof:
image


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -707,6 +707,33 @@ public void setData(@NotNull List<Object[]> rows) {
...
+                }
+            }
+        }
+    }
+
+    public void updateColorMapping(DBVEntity virtualEntity, boolean reset) {
+        colorMapping.clear();
+        if (virtualEntity == null) {
+            return;
+        }
```
ShadelessFox: Redundant because `virtualEntity` is never `null` according to the call site

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -707,6 +707,33 @@ public void setData(@NotNull List<Object[]> rows) {
...
+                            colorMapping.computeIfAbsent(binding, k -> new ArrayList<>());
+                    cmList.add(new AttributeColorSettings(co));
+                } else {
+                    log.debug("Attribute '" + co.getAttributeName() + "' not found in bindings. Skip colors.");
+                }
+            }
+        }
+    }
+
+    public void updateColorMapping(DBVEntity virtualEntity, boolean reset) {
```
ShadelessFox: Please add nullability annotations

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
coderkill: Done!

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
coderkill: Done!

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
ShadelessFox: Add the `@NotNull` annotation as well please

coderkill: Already added, do have a look

ShadelessFox: Pardon, in the method above.

coderkill: Done! Do have a look

ShadelessFox (approved):

ShadelessFox:
Thanks for the contribution! Passing the PR to our QA team now.

coderkill:
Hi,

is there any update on the testing for this PR?

ShadelessFox:
Our QA team is busy right now; it's _that_ time before the major release. They will eventually get their hands on your PR, don't worry :^)

Matvey16:
Verified current changes.
But can you also extend this fix to this feature
image

It currently has the same bug so we can't close the issue with half of the fix

coderkill:
Noted, Will make the change

coderkill:
@Matvey16 I have pushed my changes. Now it should work for the use case you have described

coderkill:
@Matvey16 @E1izabeth @ShadelessFox any 

update on the above?

E1izabeth:
@coderkill, your PR is waiting while the QA team tests it again

Matvey16:
Verified

coderkill:
@E1izabeth @ShadelessFox  Can you please have this merged?

PR #36404, opened 2024-11-29
Dbeaver/pro#3360 Fix datasets opening

E1izabeth:


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -2292,6 +2277,35 @@ protected void doSetInput(IEditorInput editorInput) throws CoreException {
         editorImage = new Image(Display.getCurrent(), baseEditorImage, SWT.IMAGE_COPY);
     }
 
+    @NotNull
+    private Runnable getEditorInputInitializer(@NotNull IEditorInput editorInput) {
```
ShadelessFox: Let's not make it return a `Runnable`. Instead, make it do what its `run` method does, and at the call site wrap it into lambda.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -2258,25 +2261,7 @@ protected void doSetInput(IEditorInput editorInput) throws CoreException {
...
-            }
-
-            {
-                DBPDataSourceContainer dataSource = EditorUtils.getInputDataSource(finalEditorInput);
-                SQLEditorFeatures.SQL_EDITOR_OPEN.use(Map.of(
-                    "driver", dataSource == null ? "" : dataSource.getDriver().getPreconfiguredId()
-                ));
-            }
-        };
+        Runnable inputinitializer = getEditorInputInitializer(editorInput);
```
ShadelessFox: A continuation of the previous comment
```suggestion
        Runnable inputinitializer = () -> getEditorInputInitializer(editorInput);
```
But be careful as you might want to capture just the variable.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
E1izabeth: I need Runnable for UIExecutionQueue.queueExec

ShadelessFox: Right, but is it the method's responsibility to make you a `Runnable`? I don't think so

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -3268,17 +3279,19 @@ public void handleDataSourceEvent(final DBPEvent event) {
...
-                : execContext.getContextDefaults();
-            if (ctxDefault != null
-                && (event.getObject() == ctxDefault.getDefaultCatalog() || event.getObject() == ctxDefault.getDefaultSchema())
-            ) {
-                contextChanged = true;
+            if (execContext != null) {
+                DBCExecutionContextDefaults<?, ?> ctxDefault = execContext.getContextDefaults();
+                if (ctxDefault != null && eventObject != null) {
+                    boolean defaultChanged = eventObject == ctxDefault.getDefaultCatalog() || eventObject == ctxDefault.getDefaultSchema();
+                    if (lastExecutionContext != executionContext || (lastExecutionContext == executionContext && defaultChanged)) {
```
ShadelessFox: Can be simplified
```suggestion
                    if (lastExecutionContext != executionContext || defaultChanged)) {
```

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -2292,6 +2287,23 @@ protected void doSetInput(IEditorInput editorInput) throws CoreException {
         editorImage = new Image(Display.getCurrent(), baseEditorImage, SWT.IMAGE_COPY);
     }
 
+    private void accomplishEditorInputInitialization(@NotNull IEditorInput editorInput) {
+        DBPDataSourceContainer oldDataSource = SQLEditor.this.getDataSourceContainer();
```
ShadelessFox, line 2287: Why do you use explicit `this` everywhere?

E1izabeth: It doesn't work without `this` here.
Also I like the general approach of clearly specifying that the object comes from the current class scope

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
```diff
@@ -111,6 +111,10 @@ void setDefaultCatalog(@NotNull DBRProgressMonitor monitor, @NotNull PostgreData
                     disconnect();
                     setOwnerInstance(catalog);
                     connect(monitor, null, null, null, false);
+                    // Catalog has been changed. Get the new one and propagate isolated flag
```
serge-rider: Don't understand this logic. We already have new instance (it is catalog). Why read it again why set isolated flag?

E1izabeth: It's the same situation as here https://github.com/dbeaver/dbeaver/pull/36404/files#diff-39b703e635479b90005852d1883628ef40d44433da8992522bd6bea810ea6ffbR125-R128
After disconnecting and connecting, the catalog is a different object, and it loses the isolatestContext property value.

PR #36416, opened 2024-11-30
#6086 - Added mysql export settings option to not export routines/procedures

### LINKED ISSUE #6086: How to remove --routines option from mysqldump?
#### System information: 
- Operating system (distribution) and version: Arch Linux Kernel 5.1.8-arch1-1-ARCH
- DBeaver version 6.0.4

- Database name and version: Server MySQL 5.7 - Client MariaDB 10.3

#### Describe your question:
When dumping a database the client always adds the `--routines` flag to `mysqldump`.
This is problematic with the above client server combination, see https://jira.mariadb.org/browse/MDEV-17429

I found the corresponding code here: https://github.com/dbeaver/dbeaver/blob/6ebe0bcc6d531508bdc5dffaf16a50a7a3bade75/plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/tools/MySQLExportWizard.java#L161-L163

How can I remove the option from the command?

pajonas:
This PR adds an settings option to not export the routines/procedures as suggested in #6086. 

image


E1izabeth (approved):

E1izabeth:
Thanks for the contribution!

ShadelessFox (approved):

pajonas:
@ShadelessFox @E1izabeth I fixed a wrong translation edit in the chinese version. I previously changed the original no_data text instead of adding a no_routines text. I fixed that with the last commit

E1izabeth (approved):

Matvey16:
Verified

PR #36431, opened 2024-12-02
#22148 Change expired password in MySQL

### LINKED ISSUE #22148: Allow change password/expired password handling for Mysql/MariaDB
A user is unable to change their own password and is not prompted to change their password if it is expired when connecting to MySQL or MariaDB

Similar to what is mentioned in https://github.com/dbeaver/dbeaver/wiki/Change-current-user-password, it would be great if DBeaver could add MySQL/MariaDB to that list.

ShadelessFox:
Shows the default "change password" dialog. This change only affects **MySQL**. I was unable to create such a user in MariaDB.

image

Sample user for testing:
```sql
CREATE USER 'expire_daily'@'%' IDENTIFIED WITH 'mysql_native_password' BY 'expire_daily' PASSWORD EXPIRE INTERVAL 1 DAY;

ALTER USER 'expire_daily'@'%' PASSWORD EXPIRE;
```

Update: MariaDB doesn't seem to provide any driver property required to catch password expiration errors early.


E1izabeth (approved):

MashaKorax:
verified

PR #36438, opened 2024-12-03
dbeaver#36415 Composite types in custom SQL

### LINKED ISSUE #36415: Composite types are handled incorrectly (the solution is presented)
### Description

https://github.com/user-attachments/assets/6eb2ce3c-1024-495b-9024-68444127503b

Actual query:
```
SELECT "f.title", COUNT(*) FROM (
select f
from film as f
) src
GROUP BY "f.title"
```

Expected query:
```
SELECT (src.f).title, COUNT(*)
FROM (
    select f
    from film as f
) src
GROUP BY (src.f).title
```

### DBeaver Version

Enterprise Edition 24.3.0-202411230022

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.4

### Steps to reproduce

_No response_

### Additional context

_No response_

serge-rider:
Added support for the composite types to the grouping panel
image

Added ability to reorder columns in the grouping panel
image

Fixed Filter by Value and Grouping panel for the nested composite types
image


uslss:
dbeaver: verified
cloudbeaver: checked that there is no regression on CB side (works as it worked before)

PR #36446, opened 2024-12-03
dbeaver/dbeaver#35989 - Syntax error in postgres roles

### LINKED ISSUE #35989: [postgres] Syntax error in role settings
### Description

DDL of settings configured for particular role (shown in Source code in Role properties) is not single quoted, resulting in syntax error

See screenshot.
image

Proper syntax is:
```
ALTER ROLE app_settlement SET statement_timeout='0s';
```

BTW I found too many empty rows between role and permissions

### DBeaver Version

CE 24.2.2

### Operating System

Fedora 40

### Database and driver

Database: Postgres 16.3
Driver Postgres

### Steps to reproduce

1. Create role
2. Set role settings ie `ALTER ROLE your_role_name SET statement_timeout='5s';
3. refresh role properties tab
4. look at source code

### Additional context

_No response_

HocKu7:


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDialect.java
```diff
@@ -1258,4 +1258,12 @@ public EnumSet<ProjectionAliasVisibilityScope> getProjectionAliasVisibilityScope
             ProjectionAliasVisibilityScope.ORDER_BY
         );
     }
+
+    @Override
+    public boolean isQuotedString(String string) {
```
E1izabeth: https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-STRINGS

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreRole.java
```diff
@@ -364,6 +365,23 @@ private void loadExtraConfigParameters(@NotNull DBRProgressMonitor monitor) thro
         }
     }
 
+    private String quoteParamIfNeed(String parameter) {
```
ShadelessFox, line 368: Maybe it would be better to extract just the value here, without quotes, and store both the name and the value inside `PostgreRoleSetting`?

HocKu7: But anyway, we need the logic about quoting. Your idea to move that into `PostgreRoleSetting`?

ShadelessFox: No, process it here but store key/value (UNQUOTED) separately in `PostgreRoleSetting`

ShadelessFox (approved):

Matvey16:
Verified

PR #36459, opened 2024-12-04
#36452 Generate unique sheet name

### LINKED ISSUE #36452: Export to Excel Issue
### Description

When trying to export tables with more than 1M records to Excel, it only exports the first 1M records, but there are more records in the table.

### DBeaver Version

DBeaver Community 24.3

### Operating System

macOS Sequoia 15.1.1

### Database and driver

database:bigquery  driver: com.simba.googlebigquery.jdbc42.Driver

### Steps to reproduce

image

image

image

image

Export only first 1.048.575 rows .

I changed Max row on sheet limit but the result did not change. I tried also EA version but it didn't work.

### Additional context

_No response_

ShadelessFox:
Try exporting any table and set **Max row on sheet** to any value below the total row count in that table.


HocKu7 (approved):

E1izabeth (approved):

MashaKorax:
verified

PR #36461, opened 2024-12-04
#35950 Show default privilege grantor in DDL

### LINKED ISSUE #35950: PostgreSQL - Missing Grantor in Default-Privileges
### Description

With #14804 and https://github.com/dbeaver/dbeaver/pull/21553 a new feature was introduced, which allows you to see (and edit) default-privileges directly via DBeaver.

The grantor of these default-privileges is missing. In PostgreSQL the grantor of default-privileges (the role which was used while creating the default-privileges) is essential for them. These default-privileges will only be used on new Objects which will be created by the grantor and will not be global on the complete schema.

A mapping consists of the grantor, grantee and the schema.
For example (we have default-privileges set by the grantor to the grantee on a schema): if a table is created inside the schema by another role than the grantor, default-privileges for the grantee will not be assigned. This will only work if the grantor creates a table in the schema.

So if you habe multiple default-privileges in on one Schema, DBeaver will show them as duplicates and you don't really know, whats the difference between them and when they will be used.

For example I have two default-privileges on a schema called `test` which where granted by two different users (test1, test2). I will be shown (in DDL) like this:

```sql
ALTER DEFAULT PRIVILEGES IN SCHEMA test GRANT SELECT ON TABLES TO test_select;
ALTER DEFAULT PRIVILEGES IN SCHEMA test GRANT SELECT ON TABLES TO test_select;
```

This also causes a little inconvenience when you're using the privileges menu inside DBeaver. You can only see if the role (you are logged in with) assigned default-privileges to a role you are inspecting. But you cannot see if there are other default-privileges already assigned to this role.

If you inspect the source of a role with the option to show permissions, you cannot see the default-privileges that were assigned to this role. You can only see this on schemas. Maybe it would make sense to have it on both places, because they are related.

### DBeaver Version

Enterprise 24.2.0

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

ShadelessFox:
java_OPcWVvYgGZ

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/privileges/RolePermissionsEditor.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/ui/editors/privileges/RolePermissionsEditor.java
@@ -104,6 +104,7 @@ public class RolePermissionsEditor extends PostgresPermissionsEditor<DBSObject> 
             PostgreDefaultPrivilege defaultPrivilege = new PostgreDefaultPrivilege(
                 owner,
                 role.getRoleReference(),
+                null,
                 List.of(createGrant(owner, role, privilegeType))
             );
             defaultPrivilege.setUnderKind(defineKind(folder.getChildrenClass()));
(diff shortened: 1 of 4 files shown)


ShadelessFox:
@ridicolos can you please take a look and say if that is what your request was about? I'm unfamiliar with privileges in PostgreSQL, so I might have understood the request incorrectly.

E1izabeth (approved):

HocKu7 (approved):

uslss:
verified

ridicolos:
@ShadelessFox this looks great! Thanks for the work 🚀

PR #36472, opened 2024-12-05
#14042 Add keywords for Metadata preference page

### LINKED ISSUE #14042: Preferences filter text does not work for all settings
#### System information: 
- Ubuntu 21.04
- DBeaver 21.2.1.202109181446
- Additional extensions: none

Preferences "type filter text" field does not work for filter settings. #10979 may have been related

1. Go to Window > Preferences
2. In "type filter text" field on top left, enter "DDL", or "case", or "case-sensitive", etc
3. Some results may come up, but
4. Under Connections > Metadata is a setting "Use case-sensitive names in DDL statements", and
5. This setting never shows up, regardless of which words you put in the filter test

There is nothing in the error log

E1izabeth:
image


ShadelessFox (commented):
Missing keyword reference in `org.jkiss.dbeaver.ui.app.eclipse`

ShadelessFox (approved):

HocKu7 (approved):

Matvey16:
Verified

PR #36474, opened 2024-12-05
dbeaver/pro#3625 AS before table alias in FROM clause

E1izabeth:
Specify dialect property about supporting as before table alias in from and use it in query generation for data editor.

Oracle and Influx AS should not be used for the generated query when you open the Data Editor; for other databases, it should.


ShadelessFox (approved):

HocKu7 (approved):

Matvey16:
Verified

PR #36502, opened 2024-12-09
dbeaver/pro#3365 Add new driver for Salesforce Data Cloud

ShadelessFox:
java_XUfwUS9egl


Matvey16:
Data Cloud driver disappeared from CloudBeaver
msedge_c7sCYDbPsU


How it was before
msedge_4nIW8PgHVV

PR #36528, opened 2024-12-11
#35614 Replacing new line characters to paragraph symbol in CSV generator

### LINKED ISSUE #35614: Line break character handling on CSV export
This was submitted before but obvious miscommunication between the reporter and the responders led to it being left unresolved. Shame.

https://github.com/dbeaver/dbeaver/issues/8285

When a string value contains a linebreak symbol of any kind, it utterly breaks Excel when you try exporting a CSV and then opening it in Excel.

As is:
Step one. Export CSV. Your settings do not seem to matter.
Step two. Open CSV in Excel.
Step Three. Watch your table get disfigured because the importer cannot handle a linebreak symbol and starts a new line partway through the column list causing columns to be displaced. If you have 54 columns and column 33 has a linebreak in it, column 34 becomes column 1 of a new line which is no good terrible horrible if it occurs more than once per file.
EDIT: It also notably breaks the breaking up of the rest of the original lines into separate columns at least as far as Excel's CSV importer is concerned, which makes it impossible to fix in Excel itself unless you write an excessively meticulous macro.

To be:
Add an option to excise or replace all "unprintable signs" as Microsoft Office understands them (the button that looks like a reverse capital P with an extra line, the carriage return) with a chosen symbol when exporting CSVs, focusing on line breaks you cannot remove with REPLACE in your SQL query.

Not all of us can edit the table data to fix incorrect data entries, and manually fixing all occurences is intractable if your CSV is more than a hundred lines long.

IvanGorshechnikov:
Screenshot 2024-12-17 at 17 09 47


ShadelessFox (approved):

E1izabeth (approved):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterCSV.java
```diff
@@ -98,6 +100,8 @@ public void init(IStreamDataExporterSite site) throws DBException
         Map<String, Object> properties = site.getProperties();
         this.delimiter = StreamTransferUtils.getDelimiterString(properties, PROP_DELIMITER);
         this.rowDelimiter = StreamTransferUtils.getDelimiterString(properties, PROP_ROW_DELIMITER);
+//        this.lineFeedEscapeString = CommonUtils.toString(properties.get(PROP_LINE_FEED_ESCAPE_STRING), null);
```
ShadelessFox:
```suggestion
```

IvanGorshechnikov: It's not the last commit. I already deleted it

kseniaguzeeva:
1. Would it be better to change name on "Line break character"?
2. We have 3 variants, but the dropdown shows 2 only. Should we show the empty value in the dropdown too? Also when we remove values, the key remain bold

https://github.com/user-attachments/assets/1af7cd70-1573-4ead-9695-e6e00aa1e1b4

3. we can add our own values instead of the values. Should we restrict to do it?
dbeaver_BIr7TZPtpX

IvanGorshechnikov:
1. Yes, I think name "Line break character" is better
2. I add third variant of empty value. About bold keys it is another bug with fileds, where default value is empty string. Similar situation with "NULL string"
3. User should have variant to add custom value

E1izabeth (approved):

ShadelessFox (approved):

kseniaguzeeva:
Is it possible to add only 1 empty value in the dropdown dialog box?

aa8e5QHpGH

kseniaguzeeva:
Verified.

Note: when I enter custom `\r\n` to "Line break character", Notepad and Visual Studio Code display the CSV file by a different way comparing to DBeaver and XLSX

XLSX

soffice bin_b8G0StgpfJ

Visual Studio Code

Code_Y5RVOMeCTJ

The setting

dbeaver_vKSlLhKZHS

PR #36534, opened 2024-12-12
#36532 Fix NPE on opening Users editor in MySQL

### LINKED ISSUE #36532: Load catalog privileges has encountered a problem
### Description

image

This error just started appearing in version 24.3.0 CE. It appears after I've not used DBeaver for a while and then come back to it. Please let me know if any additional logs are required

### DBeaver Version

24.3.0 CE

### Operating System

macOS 15.1.1

### Database and driver

MariaDB

### Steps to reproduce

Occurs after letting DBeaver sit idle for a period of time. I am unsure how long

### Additional context

_No response_

E1izabeth:
I can't reproduce. It looks like a floating bug. Let's try to fix by log.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/editors/MySQLUserEditorGeneral.java
+++ b/plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/editors/MySQLUserEditorGeneral.java
@@ -35,6 +35,7 @@ import org.jkiss.dbeaver.ext.mysql.ui.internal.MySQLUIMessages;
 import org.jkiss.dbeaver.model.edit.DBECommand;
 import org.jkiss.dbeaver.model.edit.DBECommandContext;
 import org.jkiss.dbeaver.model.edit.DBECommandReflector;
+import org.jkiss.dbeaver.model.exec.DBCExecutionContext;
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/editors/MySQLUserEditorGeneral.java
```diff
@@ -169,33 +170,40 @@ public void dispose()
...
-                                iterator.remove();
-                            }
+                        var dbObj = getDatabaseObject();
+                        if (dbObj == null) {
+                            isLoaded = false;
+                            return null;
                         }
+                        List<MySQLPrivilege> privList = dbObj.getDataSource().getPrivilegesByKind(monitor, MySQLPrivilege.Kind.ADMIN);
+                        // Remove proxy (it is not singleton)
+                        privList.removeIf(priv -> priv.getName().equalsIgnoreCase("proxy"));
```
ShadelessFox: I really dislike that we modify the returned collection. It's bad we return mutable collections from APIs in the first place.

Let's be more explicit and filter elements inside of a Stream.

ShadelessFox (approved):

HocKu7 (approved):

Matvey16:
Verified

PR #36538, opened 2024-12-12
dbeaver/pro#2667 Remove unindenting after delimiter when add new line

IvanGorshechnikov:


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/indent/SQLIndenter.java
```diff
@@ -140,8 +150,12 @@ private String getLeadingWhitespace(int offset) {
 
             int indentLength = nonWS - lineOffset;
             StringBuilder indent = createIndent();
-            if (indentLength >= indent.length() && scanner.endsWithDelimiter(lineOffset, lineOffset + line.getLength())) {
-                nonWS -= indent.length();
+            if (indentLength >= indent.length() && scanner.endsWithDelimiter(lineOffset, lineOffset + line.getLength() - 1)) {
+                SQLParserContext context = new SQLParserContext((DBPDataSource) null, syntaxManager, new SQLRuleManager(syntaxManager), document);
```
E1izabeth: Let's create a field of SQLParserContext type and initialize it in the constructor. No need to create a new instance each time the method is called.

PR #36560, opened 2024-12-16
#35445 Change table alias to first character of its name

### LINKED ISSUE #35445: Use the table's first letter as the alias instead of "X" when generating queries from the "Click to open query in editor" button
When using the "Click to open query in editor" button while viewing filters on a table in DBeaver, the generated query always assigns the alias "X" to the table. This can be frustrating, especially when working with multiple tables, as it is not very intuitive and requires extra effort to change the alias to something more meaningful.

image

I would like DBeaver to automatically use the first letter of the table's name as the alias in the generated query. For example, if the table name is "public.table," the query should look like this:
```sql
SELECT t.* FROM public.table t
```

instead of:

```sql
SELECT x.* FROM public.table x
```

An alternative solution could be allowing users to configure a default alias pattern in the settings. However, in my opinion, automatically using the first letter of the table's name would be the most intuitive and efficient solution.

IvanGorshechnikov:
image


plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTable.java
```diff
@@ -148,8 +148,10 @@ public DBCStatistics readData(
         // Some criteria doesn't work without alias
         // (e.g. structured attributes in Oracle or composite types in PostgreSQL requires table alias)
         String tableAlias = null;
+        String tableFullName = getTableName();
         if (needAliasInSelect(dataFilter, rowIdAttribute, dataSource)) {
-            tableAlias = DEFAULT_TABLE_ALIAS;
+            String[] parts = tableFullName.split("\\.");
```
E1izabeth: Better use dialect.getStructSeparator() for dot. I have never seen a database where it's not a dot, but we have this in SQL Dialect (maybe for some reason)

E1izabeth (approved):

ShadelessFox (approved):

uslss:
verified

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTable.java
```diff
@@ -148,8 +148,10 @@ public DBCStatistics readData(
         // Some criteria doesn't work without alias
         // (e.g. structured attributes in Oracle or composite types in PostgreSQL requires table alias)
         String tableAlias = null;
+        String tableFullName = getTableName();
         if (needAliasInSelect(dataFilter, rowIdAttribute, dataSource)) {
-            tableAlias = DEFAULT_TABLE_ALIAS;
+            List<String> parts = CommonUtils.splitString(tableFullName, (dataSource.getSQLDialect().getStructSeparator()));
```
serge-rider: Use SQLUtils.generateEntityAlias

serge-rider (changes requested):

Matvey16:
Verified

PR #36571, opened 2024-12-17
#36073 Add vertical scrolling in tip of the day

### LINKED ISSUE #36073: The "Tip of the day" text is more than the available area and there's no way to scroll
### Description

On 10/26, when I opened DBeaver, I was reading the Tip of the day and saw that the final visible line wrapped and the was at least one more line of text - I could see the tips on the top of tall characters.  There is no scroll bar, and I tried highlighting the area and pulled down while highlighting.  I could see that the next (hidden) row of text had gotten highlighted, but the text did not scroll, whihc is what I was hoping.

The visible text is:
In SQL Editor, you can call Content Assist (Ctrl+Space) after the asterisk in a query like "SELECT * FROM tableName" or like "INSERT INTO tableName (*)" (brackets are important), and the

Including the next row that I was to highlight, the complete text is:
In SQL Editor, you can call Content Assist (Ctrl+Space) after the asterisk in a query like "SELECT * FROM tableName" or like "INSERT INTO tableName (*)" (brackets are important), and the **_asterisk will be replaced with a list of all table columns._**

### DBeaver Version

Community Version 24.2.3.202410201725

### Operating System

Windows 11 Enterprise Version 23H2

### Database and driver

Snowflake

### Steps to reproduce

1. Start DBeaver, the "Tip of the day" should display.
2. If the Tip for "today" does not overflow past the end of the visible area, hit <Back> or <Next> to cycle through other Tips until one is found that does overflow

### Additional context

_No response_

IvanGorshechnikov:
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.ui.tipoftheday/src/org/jkiss/dbeaver/ext/ui/tipoftheday/ShowTipOfTheDayDialog.java
+++ b/plugins/org.jkiss.dbeaver.ext.ui.tipoftheday/src/org/jkiss/dbeaver/ext/ui/tipoftheday/ShowTipOfTheDayDialog.java
@@ -26,16 +26,14 @@ import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.graphics.FontData;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

ShadelessFox (approved):

uslss:
verified

Issue #35267, opened 2024-08-20
Display breadcrumbs at top of Database Navigator pane


johnrix:
**Is your feature request related to a problem? Please describe.**
I maintain connections to multiple DB servers, each with multiple instances of the same database (dev, test, prod etc.), each DB instance contains many tables. When viewing the Database Navigator pane, it is not obvious to me which database I am looking at if the database name has scrolled above the top of the pane, which is most of the time. I have to scroll up every time in order to confirm which database I am now looking at.

**Describe the solution you'd like**
Displaying the ancestors above the top-most visible node in the navigator pane would be very helpful to combat this issue, much akin to the way the Explorer pane in VSCode does so in recent versions. 

**Describe alternatives you've considered**
Presently, I maintain an open tab for each database instance in the main work area (not sure the correct name of this pane) and switch to those tabs to access a specific table that I need. This means wasted tab space in the work area though, which fills up quickly with multiple table tabs also.

**Additional context**
Example from VSCode Explorer:
![image](https://github.com/user-attachments/assets/26378850-a5c8-4223-9eff-484e7ab453ce)

E1izabeth:
Thank you for the idea.

BTW, currently we have them in the top right corner of Entity editor.
![image](https://github.com/user-attachments/assets/b0dee64d-a7cc-40da-b1b3-56db37d1143e)

johnrix:
I check that every time I switch tabs in fact, since once again, I need to be sure I'm looking at the table for the correct DB. I will often have the same table open across multiple databases. Perhaps this is the subject of a separate feature request, but the ability to colour-code and/or group the tabs in the entity editor by database would be very helpful also.

arhayka:
We discussed this issue with the team and decided to implement breadcrumbs in the status bar at the bottom of all windows:
<img width="400" alt="Screenshot 2024-09-25 at 13 03 44" src="https://github.com/user-attachments/assets/e6c8eb8f-5d2f-4d42-9f33-e40acd388a64">

It should be clickable and update on focus when the user selects something in Navigator.

It seems this will help better understand the current context.

This control could be used:
<img width="797" alt="Screenshot 2024-09-25 at 12 52 39" src="https://github.com/user-attachments/assets/b60410b6-c6f5-4e49-8646-81d763d92037">

PR #36572, opened 2024-12-17
#35267 breadcrumbs redesign

serge-rider:


plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/EntityEditorContributor.java
```diff
@@ -0,0 +1,53 @@
...
+        entityStatusItem = new BreadcrumbsContributionItem();
+    }
+
+    private class BreadcrumbsContributionItem extends ContributionItem {
+
+        @Override
+        public void fill(Composite parent) {
+            if (activeEditor != null) {
+
+            }
```
E1izabeth: Is it expected to be empty?

PR #36583, opened 2024-12-18
dbeaver/dbeaver#35945 Fix escaping slash for mysql

### LINKED ISSUE #35945: Add backslash to exported data for mariadb driver, if field contains backslashes
### Description

In #10697 you have fixed that behaviour for MySQL driver. However as I can see I have got same problem using MariaDB driver.

`subjects` is a longtext field containing JSON data
obraz

When I export data to SQL format, I will get following query:

```sql
INSERT INTO table (subjects) VALUES	('["{\"subjectId\":3,\"levelId\":2,\"isOur\":true}"]')
```

which is wrong (I think) because running such query gives me following result

obraz

So the correct query should contain double backslashes in such case.

### DBeaver Version

Community Edition 24.2.2

### Operating System

Ubuntu 20.04

### Database and driver

MariaDB - MySQL (org.mariadb.jdbc.Driver)

### Steps to reproduce

_No response_

### Additional context

_No response_

HocKu7:


ShadelessFox (approved):

test/org.jkiss.dbeaver.ext.mysql.test/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDialectTest.java
```diff
@@ -16,19 +16,51 @@
...
-        Assert.assertEquals("`a ' b`", dialect.getQuotedIdentifier("a ' b", false, true));
-        Assert.assertEquals("`a `` b`", dialect.getQuotedIdentifier("a ` b", false, true));
-        Assert.assertEquals("`a b`", dialect.getQuotedIdentifier("a b", false, true));
+    @Test
+    public void escapeString_whenSimpleStringPasses_thenSameStringReturn(){
+        // GIVEN
+        String expected = "There is simple string without any quotes and slashes";
+        // WHEN
+        String actual = dialect.escapeString(expected);
+        // THEN
```
E1izabeth, line 43: I find such comments redundant

E1izabeth (approved):

Matvey16:
Verified

PR #36586, opened 2024-12-18
Add support for Timeplus streaming database

### LINKED ISSUE #36650: Add support for Timeplus database
Timeplus is a simple, powerful, and cost-efficient stream processing platform. 
Timeplus Proton (https://github.com/timeplus-io/proton)
Timeplus Enterprise (https://timeplus.com/product).

jovezhong:
Timeplus is a simple, powerful, and cost-efficient stream processing platform. This PR added support for both Timeplus Proton (https://github.com/timeplus-io/proton) and Timeplus Enterprise (https://timeplus.com/product). 

To test it:

Start a docker instance of Timeplus Enterprise via 
```bash
docker run --name tpe2.5.11 -p 8000:8000 -p 8463:8463  -p 7587:7587 timeplus/timeplus-enterprise:2.5.11
```
Then you can access http://localhost:8000 and create the first user, say admin, with password `changeme`.

Download the JDBC jar from https://github.com/timeplus-io/timeplus-native-jdbc/releases/tag/v2.0.7

You can create a connection for localhost and 7587 port (we also support streaming SQL using 8463 port, but the SQL will be long running and DBeaver will consider it hang)

-----

PS. the code is largely based on the ClickHouse database support, with necessary SQL keywords or data type names changes.


E1izabeth:
Thanks for the contribution! We will review your PR and let you know if some changes are required

ShadelessFox (changes requested):
I believe this is basically a copy-paste of the Clickhouse bundle. It's terrible as maintaining features becomes really hard.

What makes Timeplus different from ClickHouse? Why is there a need for a separate bundle with a lot of copy-pasted code? Does it have APIs incompatible with the latter?

jovezhong:
Hi @ShadelessFox , thanks for review comments. Timeplus is a fork of ClickHouse code with the following changes/differences:
* the default query mode is streaming. For example "select count() from sth" in ClickHouse will return a number, say 10. In Timeplus, it will show 0 first, then 2, 5, 7.. Depending on how many new events inserted to the table/stream. The aggregation is incremental, so it won't rescan old data. To get the OLAP style query, need to run "select count() from table(sth)", or ""select count() from sth settings query_mode='table'", or use the non-default ports https://docs.timeplus.com/proton-ports
* data types and function names are different, for example "Nullable(String)" would be "nullable(string)" in Timeplus, also toString is to_string in Timeplus. We made a choice to consolidate the functions name style in ClickHouse, such as JSON_QUERY, toString.. all functions in Timeplus is in the to_int style. This also requires many changes
* ClickHouse official JDBC driver is HTTP based. We also provide a TCP based JDBC driver for higher performance.

You're certainly right about the maintenance effort. Not sure whether there is any chance to make the clickhouse plugin open to override, to minimize the copy/paste code.

ShadelessFox:
What's the status of this PR @jovezhong?

My main concern is the amount of code duplication. Please minimize it by reusing code. You can alter the code in the ClickHouse bundle if needed.

jovezhong:
Sorry, I haven't looked into this PR for a while. Yes, maybe a more elegant way is to start from latest dbeaver codebase and try to make the clickhouse plugin configurable or extensible, so that the Timeplus syntax can be injected to overwrite those SQL generations or API calls. A few options
- one is to add a checkbox in the clickhouse plugin. If this is selected, then use the timeplus JDBC driver and use timeplus datatype, function names. This may introduce a set of if-else in the clickhouse plugin code
- the other option is to create a new plugin to extend the clickhouse plugin, but not sure whether this will create too many extensions

Well, another option is to make the current PR work and deliver this as v1 baseline. In next few months we can work together on the v2 implementation to minimize the code duplication by choosing one of the options above.

plugins/org.jkiss.dbeaver.ext.timeplus/src/org/jkiss/dbeaver/ext/timeplus/TimeplusTypeParser.java
```diff
@@ -0,0 +1,238 @@
...
+import org.jkiss.dbeaver.model.struct.DBSTypedObject;
+import org.jkiss.utils.CommonUtils;
+import org.jkiss.utils.Pair;
+
+import java.io.StringReader;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class TimeplusTypeParser {
```
E1izabeth, line 49: This and the following parts of code look exactly like for Clickhouse. Please avoid copy-paste and duplication. You can inherit from clickhouse classes.

We can't accept it in the current state.

E1izabeth (changes requested):

jovezhong:
Thank all for the comments. I am going to close this PR. We are working on a new feature to open a port to use ClickHouse protocol and data type or function names. Hopefully that will enable our users to leverage existing clickhouse plugin of DBeaver to run queries. Alternatively, I will check how to extend the classes in clickhouse plugin without duplicating the code.

PR #36627, opened 2024-12-24
#35614 Replace line breaks in csv exporter

### LINKED ISSUE #35614: Line break character handling on CSV export

IvanGorshechnikov:
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterCSV.java
+++ b/plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterCSV.java
@@ -43,6 +43,8 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Date;
 import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * CSV Exporter
@@ -60,6 +62,7 @@ public class DataExporterCSV extends StreamExporterAbstract implements IAppendab
     private static final String PROP_NULL_STRING = "nullString";
     private static final String PROP_FORMAT_NUMBERS = "formatNumbers";
     private static final String PROP_LINE_FEED_ESCAPE_STRING = "lineFeedEscapeString";
+    private final Pattern LINE_BREAK_REGEX = Pattern.compile("\\r\\n|\\r|\\n");
 
     private static final String DEF_QUOTE_CHAR = "\"";
     private boolean formatNumbers;
@@ -286,7 +289,10 @@ public class DataExporterCSV extends StreamExporterAbstract implements IAppendab
         final boolean hasQuotes = useQuotes && value.indexOf(quoteChar) != -1;
 
         if (CommonUtils.isNotEmpty(lineFeedEscapeString)) {
-            value = value.replaceAll("\\r\\n|\\r|\\n", lineFeedEscapeString);
+            Matcher matcher = LINE_BREAK_REGEX.matcher(value);
+            if (matcher.find()) {
+                value = matcher.replaceAll(lineFeedEscapeString);
+            }
         }
 
         if (quoteStrategy == QuoteStrategy.ALL || (useQuotes && value.isEmpty())) {
--- a/plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterCSV.java
+++ b/plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterCSV.java
@@ -103,7 +103,7 @@ public class DataExporterCSV extends StreamExporterAbstract implements IAppendab
         if (ROW_DELIMITER_DEFAULT.equalsIgnoreCase(this.rowDelimiter.trim())) {
             this.rowDelimiter = GeneralUtils.getDefaultLineSeparator();
         }
-        this.lineFeedEscapeString = CommonUtils.toString(properties.get(PROP_LINE_FEED_ESCAPE_STRING), null)
+        this.lineFeedEscapeString = CommonUtils.toString(properties.get(PROP_LINE_FEED_ESCAPE_STRING), "")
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterCSV.java
```diff
@@ -101,6 +103,10 @@ public void init(IStreamDataExporterSite site) throws DBException
         if (ROW_DELIMITER_DEFAULT.equalsIgnoreCase(this.rowDelimiter.trim())) {
             this.rowDelimiter = GeneralUtils.getDefaultLineSeparator();
         }
+        this.lineFeedEscapeString = CommonUtils.toString(properties.get(PROP_LINE_FEED_ESCAPE_STRING), null)
```
serge-rider: NPE?
CommonUtils.toString(properties.get(PROP_LINE_FEED_ESCAPE_STRING), null) return null if property is missing

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterCSV.java
```diff
@@ -101,6 +103,10 @@ public void init(IStreamDataExporterSite site) throws DBException
         if (ROW_DELIMITER_DEFAULT.equalsIgnoreCase(this.rowDelimiter.trim())) {
             this.rowDelimiter = GeneralUtils.getDefaultLineSeparator();
         }
+        this.lineFeedEscapeString = CommonUtils.toString(properties.get(PROP_LINE_FEED_ESCAPE_STRING), "")
+            .replace("\\t", "\t")
+            .replace("\\n", "\n")
+            .replace("\\r", "\r");
```
ShadelessFox, line 112: Maybe, just maybe, we want to allow users to use, e.g., literal `\t` by doubling the slash `\\t`?

I have the following regex that does that, but I don't know of an easy way to convert it to a symbol those escapes correspond to: `(?<!\\)\\[rnt]` (matches `\r`, `\n`, and `\t`)

IvanGorshechnikov: I think it's not necessary to do this because in other fields on this page, similar actions are taking place to replace \\\t with the \t symbol how I do it in current varioant. The replacement prevent to do different behaviors in one place.

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterCSV.java
```diff
@@ -279,6 +285,10 @@ private void writeCellValue(String value, boolean quote)
         // check for needed quote
         final boolean hasQuotes = useQuotes && value.indexOf(quoteChar) != -1;
 
+        if (CommonUtils.isNotEmpty(lineFeedEscapeString)) {
```
serge-rider, line 291: I actually have doubts about performance.
Now we will compile and match regex for every single cell. It means millions of regex compilations + matches for big files.
Although there may be zero matches.
It may slow down data export significantly.
Let's precompile pattern + verify that cell has at least one \n character.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterCSV.java
```diff
@@ -60,6 +62,7 @@ public class DataExporterCSV extends StreamExporterAbstract implements IAppendab
     private static final String PROP_NULL_STRING = "nullString";
     private static final String PROP_FORMAT_NUMBERS = "formatNumbers";
     private static final String PROP_LINE_FEED_ESCAPE_STRING = "lineFeedEscapeString";
+    private final Pattern LINE_BREAK_REGEX = Pattern.compile("\\r\\n|\\r|\\n");
```
ShadelessFox:
```suggestion
    private static final Pattern LINE_BREAK_REGEX = Pattern.compile("\\r\\n|\\r|\\n");
```

ShadelessFox (approved):

Matvey16:
Parameter name should not be in bold if it uses the default value


https://github.com/user-attachments/assets/71f55a22-deba-46b6-85bf-ac4290116167

IvanGorshechnikov:
The problem with the bold font remaining is a bug related to those fields where an empty row is the default value. The problem is reproduced with the "NULL string" field

I think it's better to solve this in a separate ticket

Matvey16:
Verified

PR #36645, opened 2024-12-27
dbeaver/pro#3661 Add option to compact JSON/XML value

IvanGorshechnikov:
image


plugins/org.jkiss.dbeaver.ui.editors.json/src/org/jkiss/dbeaver/ui/data/managers/stream/JSONPanelEditor.java
```diff
@@ -55,5 +56,20 @@ protected String getFileExtension() {
...
+    @Override
+    public String minify(String value) {
+        JsonElement jsonElement;
+        try {
+            jsonElement = JsonParser.parseString(value);
+        } catch (JsonSyntaxException ex) {
+            return value;
+        }
 
+        Gson gson = new GsonBuilder()
```
serge-rider: Reuse gson from JSONFormattingStrategy, use some factory method to create formatted/non-formatted gson parser.

plugins/org.jkiss.dbeaver.ui.editors.xml/src/org/jkiss/dbeaver/ui/data/managers/stream/XMLPanelEditor.java
```diff
@@ -54,4 +54,9 @@ protected String getFileFolderName() {
     protected String getFileExtension() {
         return ".xml";
     }
+
+    @Override
+    public String minify(String value) {
+        return value.replaceAll(">\\s+<", "> <");
```
serge-rider: Very bad idea. It will corrupt xml in case of CDATA, for example.
You need to parse XML and then perform transformation to remove formatting.
E.g. https://stackoverflow.com/questions/8476318/how-to-unformat-xml-file , https://www.baeldung.com/java-pretty-print-xml

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.xml/src/org/jkiss/dbeaver/ui/data/managers/stream/XMLPanelEditor.java
```diff
@@ -54,4 +69,53 @@ protected String getFileFolderName() {
...
+        }
+    }
+
+    private static void removeWhitespaceNodes(Node node) {
+        NodeList childNodes = node.getChildNodes();
+        for (int i = childNodes.getLength() - 1; i >= 0; i--) {
+            Node child = childNodes.item(i);
+
+            if (child.getNodeType() == Node.TEXT_NODE) {
+                if (child.getNodeValue().trim().isEmpty()) {
```
serge-rider: Use isBlank instead

plugins/org.jkiss.dbeaver.ui.editors.xml/src/org/jkiss/dbeaver/ui/data/managers/stream/XMLPanelEditor.java
```diff
@@ -54,4 +69,53 @@ protected String getFileFolderName() {
...
+                    node.removeChild(child);
+                }
+            } else if (child.getNodeType() == Node.CDATA_SECTION_NODE) {
+                CDATASection cdata = (CDATASection) child;
+                child.setNodeValue(cdata.getNodeValue().trim());
```
serge-rider: No need to trim cdata

serge-rider (changes requested):

Matvey16:
Verified

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/AbstractTextPanelEditor.java
```diff
@@ -191,6 +192,9 @@ public void run() {
             final Action afAction = new AutoFormatAction();
             afAction.setChecked(getPanelSettings().getBoolean(PREF_TEXT_EDITOR_AUTO_FORMAT));
             manager.add(afAction);
+            final Action msAction = new SaveMinifyValue();
```
serge-rider: Just realized that we need it only for JSON and XML.
For other formats we mustn't show this option.

serge-rider (changes requested):

Matvey16:
Verified

PR #36646, opened 2024-12-27
dbeaver/pro#3903 Use native catalog read query

serge-rider:


Matvey16:
Schemas are duplicated

image

serge-rider:
Fixed

Matvey16:
Verified

PR #36697, opened 2025-01-06
#36693 Fix tableReferences rule in sql grammar

### LINKED ISSUE #36693: Auto Completion Not Working Properly for select with table hints
### Description

auto completion not working properly, after new update it is not working. after typing half of table name still not show references. or after typing table name not show columns. and when typing variable names not show available variables. (new completion engine)

even after change it to combine proposals, it's not working properly. (don't know if it is the old one)

so, auto completion is broke and not work like before.

### DBeaver Version

community - 24.3.2.202501051917

### Operating System

Windows 10

### Database and driver

MS SQL Server 2022
com.microsoft.sqlserver.jdbc.SQLServerDriver

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Autocomplete in `|` for the following query should work now
```
select c.id, c.Customer
from dbo.Customers c with(nolock)
where c.|
```


uslss:
verified

PR #36702, opened 2025-01-06
#35132 Add bookmarks in context menu

### LINKED ISSUE #35132: Bookmark creation poorly discoverable
DBeaver allows creating bookmarks for table and other objects, which is very useful. Unfortunately, finding how to create a bookmark for an object is difficult, as evidenced by [the first question in there](https://github.com/dbeaver/dbeaver/issues/10459), even though it can be done from the _Edit_ menu or [using the Ctrl+Alt+Shift+D keyboard shortcut, as documented in the _General User Guide_](https://dbeaver.com/docs/dbeaver/Bookmarks/).

The _Project - General_ view should offer to bookmark one or all of the opened objects.

There are other options which are not exclusive:

1. Add the menu entry to a sensible menu (I would imagine _Navigation_)
2. Allow bookmarking from the contextual menus of objects in the Database Navigator
3. Allow bookmarking from the contextual menu of object tabs
4. Allow bookmarking from the breadcrumbs (top right) contextual menu (or with a button in the tab)

IvanGorshechnikov:
The option 'Add bookmark' was added to the object tab context menu

image


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/actions/common/AddBookmarkHandler.java
```diff
@@ -76,6 +72,14 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
         return null;
     }
 
+    public void createBookmarkDialog(DBNDatabaseNode node, Shell activeShell) throws DBException {
```
serge-rider: Make it static

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/e4/DBeaverStackRenderer.java
```diff
@@ -271,6 +273,21 @@ public void widgetSelected(SelectionEvent e) {
                         DBWorkbench.getPlatformUI().copyTextToClipboard(DBUtils.getObjectFullName(object, DBPEvaluationContext.UI), false);
                     }
                 });
+
+                final MenuItem addBookmarkItem = new MenuItem(menu, SWT.NONE);
```
serge-rider: 1. This item has to be only for EntityEditor.
2. Do not add yet another "Add Bookmark" string. Use command `command.org.jkiss.dbeaver.core.navigator.bookmark.add` label. See ActionUtils.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/e4/DBeaverStackRenderer.java
```diff
@@ -271,6 +275,22 @@ public void widgetSelected(SelectionEvent e) {
                         DBWorkbench.getPlatformUI().copyTextToClipboard(DBUtils.getObjectFullName(object, DBPEvaluationContext.UI), false);
                     }
                 });
+                if (workbenchPart instanceof EntityEditor) {
+                    final MenuItem addBookmarkItem = new MenuItem(menu, SWT.NONE);
+                    addBookmarkItem.setText(NLS.bind(ADD_BOOKMARK_COMMAND_NAME, label));
```
serge-rider: 1. Why do we need ADD_BOOKMARK_COMMAND_NAME as a static variable?
2. Add just command ID to CoreCommands.
3. What NLS.bind does here? 
4. Let's also add icon and shortcut info in menu item.

serge-rider (changes requested):

MashaKorax:
verified

Issue #35193, opened 2024-08-13
DBeaver returns wrong date


grag11:
### Description

Hello 
The problem #21619 is still actual:

<img width="734" alt="image" src="https://github.com/user-attachments/assets/352145ea-93a5-4e43-88e7-f4276d44b834">

I think the problem is actual only for servers with UTC timezone because with another server (same driver, same dbeaver) it returns correct results:
<img width="723" alt="image" src="https://github.com/user-attachments/assets/5f3e3631-10ee-467d-8fa2-2e6146112ea3">

### DBeaver Version

CE 24.1.4.202408041450

### Operating System

macOS Sonoma 14.5 (23F79)

### Database and driver

com.clickhouse:clickhouse-jdbc 0.6.1

### Steps to reproduce

```
select 
	serverTimeZone() as server_tz
	,timeZone() as client_tz
	,makeDate(2024, 08, 13) as dt
	,toDateTime(dt) as to_dttm
	,toDate(to_dttm) as to_date_dttm
```

arhayka:
Could you try to specify the UTC timezone in Preferences —> User interface:
![image](https://github.com/user-attachments/assets/54f26e0c-d308-48bc-a8b0-e6b8c3b2d383)

grag11:
@arhayka Now it's ok! Thank you!
<img width="690" alt="image" src="https://github.com/user-attachments/assets/cbdb36d7-896d-4b44-831d-7246c7ce6c44">
But why does it not work properly with "Default" option?

arhayka:
Thank you for your question. The Default timezone isn't equal to the database timezone. Moreover, different databases may have different timezones.

We need to discuss this to see if we can use the database timezone instead of the timezone currently used as Default.

grag11:
Thank you for the answer!

Issue #21619, opened 2023-10-21
Date and DateTime32 are off one day


aadant:
### Description

```
create database if not exists test;
CREATE TABLE test.test_date
(
    `dt_date32` Date32 NOT NULL,
    `dt_date` Date NOT NULL
)
ENGINE = MergeTree
ORDER BY dt_date32;

insert into test.test_date values('2023-10-17','2023-10-17');

SELECT
    dt_date,
    toString(dt_date),
    dt_date32,
    toString(dt_date32)
FROM test.test_date;


Expected : 

┌────dt_date─┬─toString(dt_date)─┬──dt_date32─┬─toString(dt_date32)─┐
│ 2023-10-17 │ 2023-10-17        │ 2023-10-17 │ 2023-10-17          │
└────────────┴───────────────────┴────────────┴─────────────────────┘

1 row in set. Elapsed: 0.001 sec.
```

Running the same query in DBeaver returns 

```
dt_date   |toString(dt_date)|dt_date32 |toString(dt_date32)|
----------+-----------------+----------+-------------------+
2023-10-16|2023-10-17       |2023-10-16|2023-10-17         |
```

### DBeaver Version

Community Edition 23.03.3

### Operating System

macOS Ventura 3.6

### Database and driver

1. ClickHouse 23.3.8 (also tested with 22.3)
2. ClickHouse JDBC drivers (default, 0.4.2 also 0.5.0)

### Steps to reproduce

Connect to a 23.3.8 instance 
```
create database if not exists test;
use test;
CREATE TABLE test.test_date
(
    `dt_date32` Date32 NOT NULL,
    `dt_date` Date NOT NULL
)
ENGINE = MergeTree
ORDER BY dt_date32;

insert into test.test_date values('2023-10-17','2023-10-17');

SELECT
    dt_date,
    toString(dt_date),
    dt_date32,
    toString(dt_date32)
FROM test.test_date;


Expected : 

┌────dt_date─┬─toString(dt_date)─┬──dt_date32─┬─toString(dt_date32)─┐
│ 2023-10-17 │ 2023-10-17        │ 2023-10-17 │ 2023-10-17          │
└────────────┴───────────────────┴────────────┴─────────────────────┘

1 row in set. Elapsed: 0.001 sec.
```

Running the same query in DBeaver returns 

```
dt_date   |toString(dt_date)|dt_date32 |toString(dt_date32)|
----------+-----------------+----------+-------------------+
[...]

PR #36703, opened 2025-01-06
#35193 Fix date value in Clickhouse

IvanGorshechnikov:
image

Case from https://github.com/dbeaver/dbeaver/issues/21619

image

### DIFF
--- a/plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/datatype/DataTypeProviderRegistry.java
+++ b/plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/datatype/DataTypeProviderRegistry.java
@@ -81,13 +81,20 @@ public class DataTypeProviderRegistry implements DBDRegistry
     // DataType providers
 
     @Nullable
-    public DBDValueHandlerProvider getValueHandlerProvider(@NotNull DBPDataSource dataSource, @NotNull DBSTypedObject typedObject)
-    {
+    public DBDValueHandlerProvider getValueHandlerProvider(@NotNull DBPDataSource dataSource, @NotNull DBSTypedObject typedObject) {
+        ValueHandlerDescriptor genericValueHandler = null;
         // First try to find type provider for specific datasource type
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/datatype/DataTypeProviderRegistry.java
```diff
@@ -81,15 +81,22 @@ public void dispose()
...
 
     @Nullable
-    public DBDValueHandlerProvider getValueHandlerProvider(@NotNull DBPDataSource dataSource, @NotNull DBSTypedObject typedObject)
-    {
+    public DBDValueHandlerProvider getValueHandlerProvider(@NotNull DBPDataSource dataSource, @NotNull DBSTypedObject typedObject) {
+        ValueHandlerDescriptor genericValueHandler = null;
         // First try to find type provider for specific datasource type
         for (ValueHandlerDescriptor dtProvider : dataTypeProviders) {
-            if (!dtProvider.isGlobal() && dtProvider.supportsDataSource(dataSource) && dtProvider.supportsType(typedObject)) {
+            if (dtProvider.getId().contains("generic")) {
```
serge-rider: Weird logic.
Registry knows nothing about "generic"  data type provider. It corrupts abstractions.

Why generic is a special case?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/datatype/DataTypeProviderRegistry.java
IvanGorshechnikov: This is a special case because it is the parent of ClickHouse and all other db.

In previus logic, we refer to the first provider we meet that supports the given datasource and data type, and all the time we do not refer to the ClickhouseProvider, but to the generic (its parent)

In order not to create a third for loop, I just put it through the following condition

I think that it should be something like:
1) We try to find provider for exactly that datasource
2) Look at parents
3) Check global providers

PR #36710, opened 2025-01-07
dbeaver/pro#3930 Association hints

### LINKED ISSUE #35193: DBeaver returns wrong date

serge-rider:


Matvey16:
Verified

PR #36716, opened 2025-01-07
#35194 Properly close single tab result containers on new multitabs execution

### LINKED ISSUE #35194: Alt + X keeps showing queries in single tab mode after changing back to multiple tab mode
### Description

This issue is about changing between tab modes: single mode to multiple mode and vice versa.
If the tab with Results is not closed, the next query executions will take into account the mode the Results tab had, not the current mode from settings.
Seems like this can be related: #35187 but I'm not sure

### DBeaver Version

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

https://github.com/user-attachments/assets/63ea918c-c038-4b89-9ca2-dc6a7e4b8500

### Additional context

_No response_

E1izabeth:
See https://github.com/dbeaver/dbeaver/issues/35194#issuecomment-2512443624. Now it should be fixed


Matvey16:
Verified

PR #36722, opened 2025-01-07
#35671 Add preference to disable dictionary view

### LINKED ISSUE #35671: Provide means to disable the dictionary view when editing a column that is an association
When using the grid view and a cell that is a foreign key is selected, the value panel will be filled with possible values. The values seem to be retrieved using a background task/query that can have a serious impact if the target the association has a large number of rows due to the type of query used.
A production server was severely impacted when dbeaver executed the following query in a table with a few millions of rows, in this case due to the order by desc.
`SELECT id dbvrvalue, id FROM tzc_corerepository."transaction" WHERE id < $1 ORDER BY id DESC`

A option to disable this feature when the column in question is an association.

E1izabeth:
Now you can set `Maximum amount of elements on a dictionary panel page` to 0 or negative to turn off dictionary view

image


ShadelessFox (approved):

uslss:
verified

PR #36726, opened 2025-01-08
#36712 New zoom commands. Stackoverflow error fix

### LINKED ISSUE #36712: StackOverflow when trying to execute script with opened tabs with coloring
### Description

There is a StackOverglow error if the user tries to execute script again, after resultsets in current tabs were modified with coloring.

To reproduce this bug, it is required to call a warning that asks to close crrent tabs (it causes the error). So it must be enabled and there needs to be more than on result tab.

### DBeaver Version

24.3.3

### Operating System

Win11

### Database and driver

SQLite

### Steps to reproduce

1. Enable confirmation `close extra result tabs` in preferences
2. Execute a script with two or more selects
3. In any of result tabs set any row color modification
4. Try executing the script again

**Expected:** Confirmation to close current tabs appear and script is executed
**Actual:** StackOverflow

**Workaround:** Disable `close extra result tabs` confirmation

### Additional context

https://github.com/user-attachments/assets/ae89df61-a68f-4a1a-baec-a4615aa4140d

serge-rider:


Matvey16:
Verified

PR #36730, opened 2025-01-08
dbeaver/pro#3932 Don't expose packages for completion proposals

E1izabeth:
If you write `select * from |` in Oracle with the default SYS schema, the following should not happen in Query manager anymore
image


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/completion/SQLQueryCompletionContext.java
```diff
@@ -865,6 +865,8 @@ private <T extends DBSObject> void collectObjectsRecursively(
                             if (alreadyReferencedObjects.add(child) && score > 0) {
                                 accumulator.add(completionItemFabric.produce(score, childName, (T) child));
                             }
+                        } else if (child instanceof DBSPackage) {
```
serge-rider, line 868: We can just call cacheStructure once to avoid multiple re-entrances?

1. Generally we must call it before any attempts to do recursive scanning.
2. Do we really need recursive scanning for object containers?

PR #36731, opened 2025-01-08
#23045 Replacing escape char

### LINKED ISSUE #23045: The "Generate Script" feature does not correctly escape the escape characters
### Description

Hello,

When using the Grid editor and asking to generate the matching DML script, escape characters (at least back-slash) are not rendered escaped themselves - the suggested script is eiter unrunnable, or not DMLing data as wished.

Basic example : 
image
image
image

The correct code that should have been generated is : 
`INSERT INTO TECH_TMP.TEST_NO_ESCAPE (TEXT_FIELD)
	VALUES ('Text with an escape ''\\'' to be re-escaped ! ');`

### DBeaver Version

CE Version 24.0.0.202403031740

### Operating System

Win11 23H2 (Build 22631.3155)

### Database and driver

Snowflake JDBC 3.13.29

### Steps to reproduce

_No response_

### Additional context

_No response_

IvanGorshechnikov:
image


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/BasicSQLDialect.java
```diff
@@ -213,7 +213,7 @@ public DBPIdentifierCase storesQuotedCase() {
     @NotNull
     @Override
     public String escapeString(String string) {
-        return string.replace("'", "''");
+        return string.replace("'", "''").replaceAll("\\\\(?![_%?])", "\\\\\\\\");
```
serge-rider: It is definitely not a base logic.
I thought it is Snowflake specific?

This regex breaks mind. What does it do? If we need to escape back slash then we should do it much simpler.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/BasicSQLDialect.java
```diff
@@ -213,6 +213,9 @@ public DBPIdentifierCase storesQuotedCase() {
     @NotNull
     @Override
     public String escapeString(String string) {
+        if (isEscapeBackslash()) {
+            return string.replace("'", "''").replace("\\", "\\\\");
```
serge-rider: Do not copy-paste.

if (isEscapeBackslash()) {
            string = string.replace("\\", "\\\\");
}

serge-rider (changes requested):

MashaKorax:
verified

PR #36732, opened 2025-01-08
#35624 Fix visualization of infinity for date in postgres

### LINKED ISSUE #35624: Incorrect Infinity date display
### Description

select 'Infinity'::date, '-Infinity'::date

Shows: 292278994-08-17, 292269055-12-03
Should show: infinity, -infinity

It is basically the exact same problem that is described in https://github.com/dbeaver/dbeaver/issues/10978 and fixed for timestamp and timestamptz. Apparently, it was not fixed for date.

### DBeaver Version

Community Edition 24.0.5

### Operating System

Windows 11

### Database and driver

PostgreSQL 14.13

### Steps to reproduce

_No response_

### Additional context

_No response_

IvanGorshechnikov:
image


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreDateTimeValueHandler.java
```diff
@@ -64,6 +64,11 @@ public Object getValueFromObject(@NotNull DBCSession session, @NotNull DBSTypedO
         return super.getValueFromObject(session, type, object, copy, validateValue);
     }
 
+    @Override
+    protected boolean isReadDateAsObject() {
+        return true;
```
serge-rider, line 69: This changes entire logic of datetime values reading in PostgreSQL. Why?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreDateTimeValueHandler.java
serge-rider: We had it for Clickhouse only because it has a very specific date/time handling.

IvanGorshechnikov: For Postgres we need it to process cases with infinity date. But I agree that I shouldn't change logic for all datetime values

I change `isReadDateAsObject` logic for PostgreDateTimeValueHandler so it returns true only for infinity cases

uslss:
verified

PR #36734, opened 2025-01-08
#35222 Fix aliases names for generated SQL for Join

### LINKED ISSUE #35222: Generate SQL with JOIN from the diagram
Every time I need to create a script from ER Diagrama with more than one table, selecting the tables and clicking with the right button, going to the Generate SQL, JOIN option, I ended up having to create a prefix for the table names, because the full name comes and JOINS are not created, but tables separated by ',' are created.
```sql
SELECT
  pedido.*,
  pedido_item.*
FROM
  pedido pedido,
  pedido_item pedido_item
WHERE 
  pedido.codigo = pedido_item.codigo
```
When creating a script from ER Diagram with more than one table, selecting the tables and right-clicking, going to the Generate SQL, JOIN option. Come with option with table prefix and links with LEFT JOIN or choose whether to use JOIN or ','

When creating a script from ER Diagram with more than one table, selecting the tables and right-clicking, going to the Generate SQL, JOIN option. Come with option with table prefix and links with LEFT JOIN or choose whether to use JOIN or ','

**Additional context**
For the table prefix, I noticed that the selected settings are not being respected.
2024-08-15_09h56_05
2024-08-15_10h31_35
2024-08-15_10h34_43

IvanGorshechnikov:
1) "JOIN" was renamed to "SIMPLE JOIN"
2) Aliases became compact
3) Added another option "INNER JOIN" to generate query with joins
image

image


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/generator/SQLGeneratorJoin.java
```diff
@@ -66,6 +73,12 @@ public void run(DBRProgressMonitor monitor) throws InvocationTargetException {
         result = sql.toString();
     }
 
+    private String generateTableAlias(DBSEntity entity) {
```
serge-rider: Why keep `aliases` a a class variable? It is used only in one function.
As well as dialect.

Just pass them as parameters.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/generator/SQLGeneratorJoin.java
```diff
@@ -66,6 +71,12 @@ public void run(DBRProgressMonitor monitor) throws InvocationTargetException {
         result = sql.toString();
     }
 
+    private String generateTableAlias(DBSEntity entity, List<String> aliases, SQLDialect sqlDialect) {
```
serge-rider: We don't need this function at all. You can do everything in `run`

serge-rider (changes requested):

uslss:
verified

uslss:
incorrect query is generated for Simle Join - select is used twice and no alias is shown
image

uslss:
verified

PR #36782, opened 2025-01-13
dbeaver/pro#3504 Add a more noticeable indication of active edit mode

ShadelessFox:


Matvey16:
This breaks the border in Dark Theme

https://github.com/user-attachments/assets/a1442c74-f8c6-4caa-b75c-cc79282d7483

Matvey16:
Verified

PR #36805, opened 2025-01-14
#36711 Fix completion proposals sorting

### LINKED ISSUE #36711: Autcomplete results are in a bad order
### Description

I've recently updated my DBeaver to 24.3.2.202501051917 (flatpak) and there is a new autocomplete when compared to the previous version I was using (I don't know which one, but I update my apps frequently).

I'm connected to a DB that has several tables with `product` in their name and one `products` table.
When I type `SELECT * FROM products`, the autocomplete popup shows up with a bunch of tables. The `products` table is a exact match for what I typed, but its entry in the list is the second to last item.

Before this auto complete update, doing the same would get the table with exact match as the first entry in the auto complete list.

### DBeaver Version

Community Edition 24.3.2.202501051917

### Operating System

Pop!_OS 22.04 x86_64

### Database and driver

MySQL 5

### Steps to reproduce

1. Have a DB with several table where the full name of one is `products` and the others have names with `product_` in it, like `product_options`.
2. Type `SELECT * FROM products`.
3. Expected: The `products` table as the first item in the auto complete list. Actual: The `products` table is not the first item in the auto complete list.

### Additional context

_No response_

E1izabeth:
image
image

image


ShadelessFox (approved):

HocKu7 (approved):

uslss:
verified

PR #36816, opened 2025-01-15
dbeaver/pro#2348 Oracle user management

IvanGorshechnikov:


plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleUser.java
```diff
@@ -204,4 +217,24 @@ public boolean isPropertyCached(OracleUser object, Object propertyId)
...
+    public boolean isPersisted()
+    {
+        return persisted;
+    }
+
+    @Override
+    public void setPersisted(boolean persisted)
+    {
+        this.persisted = persisted;
+        DBUtils.fireObjectUpdate(this);
```
serge-rider: Why here?

IvanGorshechnikov: Took this method from MySQLUser
But now I realised that it is not needed here and removed it

PR #36817, opened 2025-01-15
dbeaver/pro#3544 Fix oracle tablespace ddl generation

HocKu7:
Now, ddl for Oracle tablespace is fetching from the database, not building on the application side

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleTablespace.java
+++ b/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleTablespace.java
@@ -20,9 +20,10 @@ import org.jkiss.code.NotNull;
 import org.jkiss.code.Nullable;
 import org.jkiss.dbeaver.DBDatabaseException;
 import org.jkiss.dbeaver.DBException;
-import org.jkiss.dbeaver.model.DBPScriptObject;
+import org.jkiss.dbeaver.Log;
 import org.jkiss.dbeaver.model.DBPObjectStatistics;
 import org.jkiss.dbeaver.model.DBPRefreshableObject;
+import org.jkiss.dbeaver.model.DBPScriptObject;
 import org.jkiss.dbeaver.model.DBUtils;
 import org.jkiss.dbeaver.model.exec.jdbc.JDBCPreparedStatement;
 import org.jkiss.dbeaver.model.exec.jdbc.JDBCResultSet;
@@ -50,6 +51,8 @@ import java.util.Map;
  */
 public class OracleTablespace extends OracleGlobalObject implements DBPRefreshableObject, DBPObjectStatistics, DBPScriptObject
 {
+
+    private static final Log log = Log.getLog(OracleTablespace.class);
 
     public enum Status {
         ONLINE,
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleUtils.java
```diff
@@ -189,6 +161,56 @@ public static String getDDL(
...
+            }
+            try (JDBCResultSet dbResult = dbStat.executeQuery()) {
+                if (dbResult.next()) {
+                    Object ddlValue = dbResult.getObject(1);
+                    if (ddlValue instanceof Clob) {
+                        StringWriter buf = new StringWriter();
+                        try (Reader clobReader = ((Clob) ddlValue).getCharacterStream()) {
+                            IOUtils.copyText(clobReader, buf);
+                        } catch (IOException e) {
+                            e.printStackTrace(new PrintWriter(buf, true));
```
serge-rider: don't do that

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleUtils.java
```diff
@@ -112,37 +113,8 @@ public static String getDDL(
...
-                        }
-                    } else {
-                        log.warn("No DDL for " + objectType + " '" + objectFullName + "'");
-                        return "-- EMPTY DDL";
-                    }
-                }
-            }
-            ddl = ddl.trim();
+            String ddl = fetchDDL(session, objectType, object.getName(), schema);
+            if (ddl == null) return "-- EMPTY DDL";
```
serge-rider, line 116: null handles mustn't be in util function. UI must decide.

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleTablespace.java
```diff
@@ -114,6 +117,7 @@ public enum Retention {
 
     final FileCache fileCache = new FileCache();
     final SegmentCache segmentCache = new SegmentCache();
+    String ddlStringHolder = null;
```
serge-rider: 1. Do not declare packagign visible files
2. Why cache it? We do not cache DDL intentionally.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleUtils.java
HocKu7: It's already a few components that based on this logic. Don't think that we need refactoring in this ticket

plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleTablespace.java
HocKu7: Actually, we cache it for views (OracleView, PostgreView, etc.). I don't know why we need to do query every time when windows is opened. Looks good for caching it. Don't know the disadvantages of it

Matvey16:
Verified

serge-rider:
I realized that you deleted code which can be used for new tablespace creation (when tablespace is not persisted).
Probably we should keep it

PR #36826, opened 2025-01-16
fix Driver path and class name for Machbase in plugin.xml

grey-shim:
Hi, this is Machbase 
Our jdbc driver download path and the driver class name has been changed, but it is not reflected in dbeaver, so We are reflecting the changes and sending a pull reqeust.

We are very happy to be able to contribute to a great open source like dbeaver. 
We would be grateful if you could review and accept this request.


ShadelessFox:
Hi @grey-shim,

I tried downloading the jar file from the newly provided URL (https://dl.machbase.com/dist/machbase.jar), but my machine couldn't reach it. Is it expected to be accessible worldwide?

ShadelessFox:
I just found out you guys also have a Maven artifact located here: https://mvnrepository.com/artifact/com.machbase/machjdbc.

But it hasn't been updated in a while. Is version **8.0.2**, released on Aug 14, 2023, the newest version? Maybe we can use it instead? It's always preferable to use a repository to allow picking a particular version of an artifact instead of using the latest one.

grey-shim:
> I just found out you guys also have a Maven artifact located here: https://mvnrepository.com/artifact/com.machbase/machjdbc.
> 
> But it hasn't been updated in a while. Is version **8.0.2**, released on Aug 14, 2023, the newest version? Maybe we can use it instead? It's always preferable to use a repository to allow picking a particular version of an artifact instead of using the latest one.

Yes!, version 8.0.2 is the latest version. Please change the path with maven artifact path. 
Thank you.

ShadelessFox:
In such a case, can you please update the PR to use the Maven artifact instead?
```xml
<file type="jar" path="maven:/com.machbase:machjdbc:RELEASE"/>
```

grey-shim:
> In such a case, can you please update the PR to use the Maven artifact instead?
> 
> ```
> <file type="jar" path="maven:/com.machbase:machjdbc:RELEASE"/>
> ```

Hi @ShadelessFox, I changed jdbc path.

ShadelessFox (approved):

E1izabeth (approved):

E1izabeth:
Thanks for the contribution! We passed your PR to our QA team for testing

Matvey16:
Verified

PR #36841, opened 2025-01-16
#36812 Fix field order

### LINKED ISSUE #36812: Order of fields in results view pane (shown after an update query execution) has changed
### Description

Some relatively important fields in the results pane (that shows up after execution of an update query) are ordered differently than they used to be.

Previously, information that seems important to be readily visible - Updated Rows, Execute Time, Start Time, and Finish Time - all appeared before the Query. Now, they show up after the Query. When the query is longer (a usual case I would argue), it requires scrolling to see them.

The previous order was more convenient when the details of the query weren't important but you wanted to know how many rows were updated or how long it took.

### DBeaver Version

Community Edition Version 24.3.2.202501051917

### Operating System

macOS Monterey 12.7

### Database and driver

PostgreSQL 16.2
PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

Execute any query that updates rows in a database table.
View results pane - notice that "Updated Rows" is near the bottom of the fields listed.

### Additional context

_No response_

paulaajwm:
What I've Done
- Reordered columns in the result set for single SQL statement execution to prioritize key information like "Updated Rows" and "Execute Time" over "Query Text."
- Added comments in the code to clarify the reason behind the column order changes.
Why These Changes Were Made
- The reordering ensures that the most relevant fields are displayed first, improving usability and readability for users reviewing execution results.
Linked Issue
Resolves #36812


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/execute/SQLQueryJob.java
```diff
@@ -831,7 +831,7 @@ private boolean isShowExecutionResult() {
             default:
                 return false;
         }
-        
+        co
```
E1izabeth:
```suggestion
```

E1izabeth (changes requested):
Thank you for contribution.
I believe this change was unintentional. Please revert this

E1izabeth (approved):

ShadelessFox (approved):

ShadelessFox:
I think this change is good. Thanks for the contribution! Passing the PR to our QA team now.

Matvey16:
Verified

PR #36847, opened 2025-01-17
dbeaver/dbeaver-infra#188 check if workspace is read-only to avoid multiple log transmission

Destrolaric:
Closes dbeaver/dbeaver-infra#188


plugins/org.jkiss.dbeaver.ui.statistics/src/org/jkiss/dbeaver/ui/statistics/StatisticsTransmitter.java
```diff
@@ -77,6 +77,10 @@ private void sendStatistics(DBRProgressMonitor monitor, boolean sendActiveSessio
         try {
             String appSessionId = DBWorkbench.getPlatform().getApplication().getApplicationRunId();
             Path activityLogsFolder = FeatureStatisticsCollector.getActivityLogsFolder();
+            if (activityLogsFolder.toFile().exists() && !activityLogsFolder.toFile().canWrite()) {
```
alexander-skoblikov: do not use io api, use nio

alexander-skoblikov (changes requested):

plugins/org.jkiss.dbeaver.ui.statistics/src/org/jkiss/dbeaver/ui/statistics/StatisticsTransmitter.java
Destrolaric: done

alexander-skoblikov (approved):

PR #36867, opened 2025-01-20
dbeaver/dbeaver#35943 Now any statement can be sent to mysql server

### LINKED ISSUE #35943: Extend "Only SELECT statements could produce execution plan"
On newer versions of MySQL/MariaDB one can get EXPLAIN details for UPDATE/DELETE clauses.
DBeaver says "Only SELECT statements could produce execution plan".

DBeaver should not interfere and send "EXPLAIN Query" (or even better in MariaDB "EXPLAIN EXTENDED Query") to the server and let the server decide if the query is acceptable or not.

As an alternative you can add UPDATE/DELETE statement on the whitelist for EXPLAIN.

**Additional context**
In MySQL you can issue EXPLAIN for UPDATE starting MySQL 5.6 (but not in 5.5) and for SELECT/DELETE/INSERT/REPLACE/UPDATE.
Given the relation w/ MariaDB I would allow it starting from any MariaDB 10.x

HocKu7:
According to [documentation](https://dev.mysql.com/doc/refman/8.4/en/explain.html), it is now possible to use quite more statements to a explain plan (actually, 'with' wasn't there). I Don't see any reason to throw our exception based on the MySQL server version for a specific statement. The exception from the server is pretty self-explained. Also, maybe MySQL and mariaDB forks will be affected by such logic (if restrict something on our side)
Exception from the server if you try to do a explain plan for keyword that isn't supported by db
image


E1izabeth (commented):
The error message is not user-friendly if the database doesn't support an explain plan for such statements. However, I don't see the way to know this from metadata.

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/plan/MySQLPlanAnalyser.java
```diff
@@ -52,12 +49,6 @@ public MySQLPlanAnalyser(MySQLDataSource dataSource) {
     }
 
     public MySQLPlanAbstract explain(JDBCSession session, String query) throws DBCException {
-        final SQLDialect dialect = SQLUtils.getDialectFromObject(dataSource);
-        final String plainQuery = SQLUtils.stripComments(dialect, query).toUpperCase();
-        final String firstKeyword = SQLUtils.getFirstKeyword(dialect, plainQuery);
-        if (!"SELECT".equalsIgnoreCase(firstKeyword) && !"WITH".equalsIgnoreCase(firstKeyword)) {
-            throw new DBCException(MySQLMessages.exception_only_select_could_produce_execution_plan);
```
E1izabeth, line 59: Please remove this message from properties files

ShadelessFox (approved):

E1izabeth (approved):

kseniaguzeeva:
verified (MySQL 8)

PR #36870, opened 2025-01-21
3984 external files open

serge-rider:


Matvey16:
Parquet files can't be opened from cloud explorer


https://github.com/user-attachments/assets/2be4ab7c-e931-44c9-9257-17ae94fa58c7

PR #36873, opened 2025-01-21
dbeaver/pro#2581 add information schema for databricks

labanovichttps:


HocKu7 (approved):

plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/model/DatabricksMetaModel.java
```diff
@@ -76,6 +78,38 @@ public GenericSchema createSchemaImpl(
...
+        GenericDataSource dataSource,
+        GenericCatalog catalog
+    ) throws DBException {
+        List<GenericSchema> schemas = super.loadSchemas(session, dataSource, catalog);
+        Set<String> schemaNames = schemas.stream()
+            .map(GenericSchema::getName)
+            .collect(Collectors.toSet());
+
+        try (JDBCPreparedStatement dbStat = session.prepareStatement(
+            "select schema_name from " + DBUtils.getQuotedIdentifier(catalog) + ".information_schema.schemata"
```
E1izabeth: As far as I understand, we can just read all schemas from schemata table without calling super method. It might improve performance as in this case we will send only one query to the database.
I would call super method only in case reading of schemas from schemata failed like it's done in NetezzaMetaModelExt::loadSchemas

plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/model/DatabricksMetaModel.java
```diff
@@ -76,6 +78,38 @@ public GenericSchema createSchemaImpl(
         return new DatabricksSchema(dataSource, catalog, schemaName);
     }
 
+    @Override
+    public List<GenericSchema> loadSchemas(
+        JDBCSession session,
+        GenericDataSource dataSource,
+        GenericCatalog catalog
+    ) throws DBException {
+        List<GenericSchema> schemas = super.loadSchemas(session, dataSource, catalog);
```
serge-rider: Over engineering.
If the problem is only in missing information_schema then just add it in Java, not need to use extra selects or fix anything else.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/model/DatabricksMetaModel.java
labanovichttps: There's a problem that information_schema may be absent in some catalogs. If we add it manually in java there will be an exception

plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/model/DatabricksMetaModel.java
labanovichttps: On the test environment i tried to read all schemas only from "schemata", but i found that severals schemas were absent in it and it was available only from load them from the driver.

Matvey16:
verified

PR #36877, opened 2025-01-21
dbeaver/dbeaver#36853 Remove "work" as a keyword in postgresql

### LINKED ISSUE #36853: Keyword Dialect selection
I prefer to enable the SQL editor feature to capitalize keywords. However, I commonly use the table name "work", which is an SQL 92 reserved word. Apparently DBeaver casts a wide net and capitalizes it, and keeping it lower case is trialsome. However, "work" and many others are no longer reserved words in Postgresql or many other databases.

It would be helpful to pick in the configuration the SQL standard and/or database dialect to govern keyword identification.

labanovichttps:


HocKu7 (approved):

E1izabeth (approved):

Matvey16:
Verified

PR #36878, opened 2025-01-21
dbeaver/dbeaver#36324 Do not allow to save the value in data editor i…

### LINKED ISSUE #36324: Even if the values ​​are the same when pasting, they are modified.
### Description

When you copy data and paste it into a cell with the same value, the cell is marked as modified and is updated when you save it.

### DBeaver Version

Community Edition 24.2.5.202411171748

### Operating System

Windows 10

### Database and driver

Database: Postgresql

### Steps to reproduce

1. open query results
2. move to result cell
3. ctrl + c and ctrl + v

### Additional context

_No response_

HocKu7:
…f it wasn't changed

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
@@ -519,7 +519,13 @@ public class ResultSetModel implements DBDResultSetModel {
         row.values[rootIndex] = valueToEdit;
 
         if (updateChanges && row.getState() == ResultSetRow.STATE_NORMAL) {
-            changesCount++;
+            if (!(oldHistoricValue instanceof DBDValue) &&
+                !(valueToEdit instanceof DBDValue) &&
+                Objects.equals(oldHistoricValue, valueToEdit)) {
+                changesCount = Math.min(changesCount - 1, 0);
+            } else {
+                changesCount++;
+            }
         }
 
         return true;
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -519,7 +519,13 @@ boolean updateCellValue(
         row.values[rootIndex] = valueToEdit;
 
         if (updateChanges && row.getState() == ResultSetRow.STATE_NORMAL) {
-            changesCount++;
+            if (!(oldHistoricValue instanceof DBDValue) &&
+                !(valueToEdit instanceof DBDValue) &&
+                Objects.equals(oldHistoricValue, valueToEdit)) {
+                changesCount = Math.min(changesCount - 1, 0);
```
ShadelessFox: I'm a bit skeptical about that `min`. Do you know when it might already be `0`? The only case I can think about is when `oldHistoricValue` is `null`. So maybe it's better to check whether it's not `null`?

HocKu7: Do you think such checking could lead to a bug? I think it code just more reliable.

ShadelessFox: That leads to more questions than answers

serge-rider:
Very weird.
- We cant compare all values with equals
- Why passing `needToUpdateIfSameValue` - it must be always false

PR #36913, opened 2025-01-24
CB-6157 add provider properties for oracle url connection

yagudin10:


serge-rider (commented):
A bit strange fix.
The idea was to enable all properties (except host/port/database fields) by default.
It also must apply to any new properties.

PR #36921, opened 2025-01-27
dbeaver/dbeaver#36915 Incremental search fix

### LINKED ISSUE #36915: Find Replace Not Working?
Just installed Version 24.3.3.202501191633 and the Find / Replace feature seems to be disabled.  When I press ctrl+f twice for the replace dialog box, the Replace With text box is grayed out and disabled, so there is no way to enter the replace text.  The only button that is active is Find, and when I click Find from this dialog, it says the string not found, even though it is present and it does find the text using the little corner drop down search box.

serge-rider:


MashaKorax:
Find cannot return to previous value


https://github.com/user-attachments/assets/5d25cdc7-ff42-42f1-a9c2-65f71e62af26

PR #36925, opened 2025-01-27
dbeaver/pro#3464 Fix filter for database navigator internal objects

labanovichttps:


Matvey16:
Columns are also sorted by the filter


https://github.com/user-attachments/assets/84706f37-1feb-464f-a137-7097aabc6742

Matvey16:
Verified

PR #36926, opened 2025-01-27
#36924 Return Bookmarks view

### LINKED ISSUE #36924: SQL Editor Bookmarks View Error - Could not create the view: org.eclipse.ui.views.BookmarkView
### Description

The SQL Editor Bookmarks view no longer appears to work after updating to 24.3.3 (confirmed working in 24.3.2).

Image

The option to view the pane is also missing from Show View:
Image

Has this feature been removed?

### DBeaver Version

24.3.3

### Operating System

Windows 10 22H2

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
image


uslss:
verified

PR #36927, opened 2025-01-27
dbeaver/pro#3587 QM filters panel

ShadelessFox:


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/views/qm/QueryManagerViewFilter.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.ui.views.qm;
+
+import org.eclipse.swt.widgets.Composite;
+import org.jkiss.code.NotNull;
+import org.jkiss.dbeaver.ui.controls.querylog.QueryLogViewer;
+
+public interface QueryManagerViewFilter {
```
E1izabeth, line 23: I'm not sure there will ever be more than one implementation of this interface. Do you have some other possible implementations in your mind?

ShadelessFox: The implementation is unavailable in CE, so it must be wired somehow. I haven't found a suitable interface among existing ones (besides `IObjectPropertyConfigurator`, but it offers too many things I'm not interested in)

E1izabeth (approved):

PR #36962, opened 2025-01-30
dbeaver/pro#3899 Unpin SSH/SSL tabs; enable/disable unpinned tabs

ShadelessFox:
What's changed:
- SSH and SSL tabs are no longer pinned.
- All network tabs can be _turned off_ - this action won't remove their configuration, and they will persist between restarts. Closing the tab will turn it off and reset its configuration.


Matvey16:
SSH configuration is removed along with the active tab if Dbeaver is retarted with the checkbox to use ssh disabled

Matvey16:
SSH config is not deleted and is still used after deleting the ssh tab from the connection

Matvey16:
Verified

PR #36986, opened 2025-02-01
dbeaver/dbeaver#35139 sqlite guid convesion to UUID

### LINKED ISSUE #35139: SQLite BLOB is not formatted as human-readable UniqueIdentifier string
### Description

I have an SQLite database with a Guid field stored as BLOB:
```
CREATE TABLE [DBVersion]
(
	[Id]         Guid          NOT NULL,
	[AppVersion] nvarchar(255)     NULL,

	CONSTRAINT [PK_DBVersion] PRIMARY KEY ([Id])
);
```
Some other SQLite viewers can display them in a human-readable formatting. [SQLite Expert Personal](https://www.sqliteexpert.com/download.html) do this automatically:
image
Latest DB Browser for SQLite [has this option](https://github.com/sqlitebrowser/sqlitebrowser/issues/2342) in a built-in display format list:
image
But in latest DBeaver Community 24.1.4 I can't see any suitable tranformation:
image
image
According to another issue UUID transformation [has already been added](https://github.com/dbeaver/dbeaver/issues/8960#issuecomment-678347253) but I don't see it. 
Technically I can transform value like that:
`SELECT substr(hex("column"), 7, 2) || substr(hex("column"), 5, 2) || substr(hex("column"), 3, 2) || substr(hex("column"), 1, 2) || '-' || substr(hex("column"), 11, 2) || substr(hex("column"), 9, 2) || '-' || substr(hex("column"), 15, 2) || substr(hex("column"), 13, 2) || '-' || substr(hex("column"), 17, 4) || '-' || substr(hex("column"), 21, 12) FROM table - this solution came [from here](https://neosmart.net/blog/2018/converting-a-binary-blob-guid-to-text-in-sql)`
But it's very uncomfortable to rewrite this code for each column name.
Does DBeaver Community Edition has this functionality or is it planned to be implemented?

My sample database - [entities.zip](https://github.com/user-attachments/files/16558999/entities.zip)

### DBeaver Version

Community Edition 24.1.4

### Operating System

Ubuntu 22.04 on WSL2 or Windows 10 22H2 x64 Pro

### Database and driver

1. Database name and version: SQLite 3.46.0
2. Driver name and version: jdbc:sqlite

### Steps to reproduce

1) Open my attached sample database. 
2) DBVersion.Id field is not in a human-readable format.

### Additional context

_No response_

eusebe-cda:
Consider GUID type as BLOB affinity to allow conversion to UUID

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/transformers/UUIDAttributeTransformer.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/transformers/UUIDAttributeTransformer.java
@@ -85,6 +85,8 @@ public class UUIDAttributeTransformer implements DBDAttributeTransformer {
                 Object cachedValue = ((DBDContentCached) value).getCachedValue();
                 if (cachedValue instanceof byte[]) {
                     bytes = (byte[]) cachedValue;
+                } else if (cachedValue instanceof String) {
+                    bytes = ((String) cachedValue).getBytes();
                 }
             }
             if (bytes != null) {
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/transformers/UUIDAttributeTransformer.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/transformers/UUIDAttributeTransformer.java
@@ -85,8 +85,8 @@ public class UUIDAttributeTransformer implements DBDAttributeTransformer {
                 Object cachedValue = ((DBDContentCached) value).getCachedValue();
                 if (cachedValue instanceof byte[]) {
                     bytes = (byte[]) cachedValue;
-                } else if (cachedValue instanceof String) {
-                    bytes = ((String) cachedValue).getBytes();
+                } else if (cachedValue instanceof String string) {
+                    bytes = string.getBytes();
                 }
             }
             if (bytes != null) {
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/transformers/UUIDAttributeTransformer.java
```diff
@@ -85,6 +85,8 @@ public String getValueDisplayString(@NotNull DBSTypedObject column, @Nullable Ob
                 Object cachedValue = ((DBDContentCached) value).getCachedValue();
                 if (cachedValue instanceof byte[]) {
                     bytes = (byte[]) cachedValue;
+                } else if (cachedValue instanceof String) {
+                    bytes = ((String) cachedValue).getBytes();
```
ShadelessFox:
```suggestion
                } else if (cachedValue instanceof String string) {
                    bytes = string.getBytes();
```

eusebe-cda: I was hesitant to use this syntax but didn't because it wasn't used elsewhere in the function. Is it therefore preferable to apply it all over?

current code :
```java
            if (value instanceof byte[]) {
                bytes = (byte[]) value;
            } else if (value instanceof DBDContentCached) {
                Object cachedValue = ((DBDContentCached) value).getCachedValue();
                if (cachedValue instanceof byte[]) {
                    bytes = (byte[]) cachedValue;
                } else if (cachedValue instanceof String) {
                    bytes = ((String) cachedValue).getBytes();
                }
            }
```

proposal :
```java
            if (value instanceof byte[] byteValue) {
                bytes = byteValue;
            } else if (value instanceof DBDContentCached contentCached) {
                Object cachedValue = contentCached.getCachedValue();
                if (cachedValue instanceof byte[] byteValue) {
                    bytes = byteValue;
                } else if (cachedValue instanceof String string) {
                    bytes = string.getBytes();
                }
            }
```

ShadelessFox: Just apply the proposed change. We don't refactor existing code if it's not needed.

eusebe-cda: ok, thanks for the feedback

ShadelessFox (approved):

E1izabeth (approved):
Thanks! Passing to our QA team for testing

uslss:
verified

PR #37007, opened 2025-02-03
dbeaver/pro#3879 Fix rename script and export data

labanovichttps:


Matvey16:
This leads to an NPE when using that variable

```
java.lang.NullPointerException: Cannot invoke "org.jkiss.dbeaver.model.DBPDataSourceContainer.getName()" because "this.dataSourceContainer" is null
	at org.jkiss.dbeaver.tools.transfer.database.DatabaseTransferProducer.getObjectFullName(DatabaseTransferProducer.java:122)
	at org.jkiss.dbeaver.tools.transfer.DataTransferJob.transferData(DataTransferJob.java:123)
	at org.jkiss.dbeaver.tools.transfer.DataTransferJob.run(DataTransferJob.java:100)
	at org.jkiss.dbeaver.model.runtime.AbstractJob.run(AbstractJob.java:119)
```

Matvey16:
Verified

PR #37014, opened 2025-02-03
dbeaver/pro#37003 Add state for filter type

### LINKED ISSUE #37003: Filter feature in DB browser is not working anymore
### Description

The search tool only filters connections name now ?? 

Image

Image

Image

### DBeaver Version

CE 24.3.4.202502021521

### Operating System

Win11 24H2 26100.2894

### Database and driver

Snowflake JDBC 3.22.0

### Steps to reproduce

_No response_

### Additional context

_No response_

labanovichttps:


E1izabeth (commented):
I would expect the `Restore filter` to restore the filter type, too. Why do we need a separate setting for that?

labanovichttps:
> I would expect the `Restore filter` to restore the filter type, too. Why do we need a separate setting for that?

Reverted commit, currently save also user's input filter and type

Matvey16:
Verified

PR #37020, opened 2025-02-03
dbeaver/dbeaver#37005 Add fallback for databricks schemas loading

### LINKED ISSUE #37005: Version 24.3.4.202502021521 database navigator does not show schemas in Azure  Databricks
### Description

Version Version 24.2.4.202411031831 showed Databricks catalog, then folders for each schema, then folders for tables and views etc. by default.  24.3.4 shows only catalogs and then all objects bunched together regardless of schema.  I thought it may be a connection setting or filter but I could not find any ways to show schemas and for each schema the objects it contains in database navigator.  Not sure if it's a bug or of there's a new preference setting that I am not aware of.  Appreciate bug fix or info on how to show schemas otherwise I'll stick with 24.2!  Thanks!

### DBeaver Version

Community Edition 24.3.4

### Operating System

Windows 11 enterprise

### Database and driver

Azure databricks default driver

### Steps to reproduce

Connect to a Azure Databricks catalog, open it in database explorer

### Additional context

_No response_

labanovichttps:
image


plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/model/DatabricksMetaModel.java
```diff
@@ -95,7 +95,8 @@ public List<GenericSchema> loadSchemas(
                 }
             }
         } catch (SQLException e) {
-            log.error("Cannot load schemas", e);
+            log.error("Cannot load schemas with query", e);
```
serge-rider: No need to log as error. It is expected so it is debug message

serge-rider (changes requested):

E1izabeth (approved):

Matvey16:
Verified

PR #37022, opened 2025-02-03
dbeaver/pro#4079 Remove duplicated proposals when combinin proposals from both completion engines

E1izabeth:
No duplicated strings to replace should be proposed anymore.
Proposals from legacy completion engine come last.
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/CompletionProposalBase.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/CompletionProposalBase.java
@@ -19,6 +19,7 @@ package org.jkiss.dbeaver.model.sql.completion;
 public abstract class CompletionProposalBase {
 
     protected abstract int getReplacementOffset();
+    
     protected abstract String getReplacementString();
 
     @Override
@@ -33,9 +34,10 @@ public abstract class CompletionProposalBase {
 
     @Override
     public int hashCode() {
-        int hashCode = 181846194;
-        hashCode = hashCode * -1521134295 + this.getReplacementString().hashCode();
-        hashCode = hashCode * -1521134295 + Integer.hashCode(this.getReplacementOffset());
+        int hashCode = 7;
+        hashCode = hashCode * 31 + this.getReplacementString().hashCode();
+        hashCode = hashCode * 31 + Integer.hashCode(this.getReplacementOffset());
+        jdk.internal.util.ArraysSupport.hashCode
         return hashCode;
     }
 }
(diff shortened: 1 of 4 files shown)


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/CompletionProposalBase.java
```diff
@@ -0,0 +1,41 @@
...
+            return false;
+        }
+    }
+
+    @Override
+    public int hashCode() {
+        int hashCode = 181846194;
+        hashCode = hashCode * -1521134295 + this.getReplacementString().hashCode();
+        hashCode = hashCode * -1521134295 + Integer.hashCode(this.getReplacementOffset());
+        return hashCode;
```
ShadelessFox: Why such arbitrary values? It's `31` most of the time. Just seems weird to me. `Integer.hashCode` yields the same value

E1izabeth: Why not? This is what Visual Studio generates automatically for hashCode, and I like it more than what IDEA propose

ShadelessFox: Why are you using Visual Studio for Java 🤔
Either way, JDK uses `31` as the multiplier value<sup>*</sup>. We all use IJ, which automatically generates consistent `hashCode` implementations. Now you're introducing a new style, which is bad IMO.

<sup>*</sup> It can be seen here `jdk.internal.util.ArraysSupport#hashCode(java.lang.Object[], int, int, int)`

E1izabeth: OK, it's not a big deal to change it.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/CompletionProposalBase.java
```diff
@@ -33,9 +34,10 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        int hashCode = 181846194;
-        hashCode = hashCode * -1521134295 + this.getReplacementString().hashCode();
-        hashCode = hashCode * -1521134295 + Integer.hashCode(this.getReplacementOffset());
+        int hashCode = 7;
+        hashCode = hashCode * 31 + this.getReplacementString().hashCode();
+        hashCode = hashCode * 31 + Integer.hashCode(this.getReplacementOffset());
+        jdk.internal.util.ArraysSupport.hashCode
```
ShadelessFox: Probably this line is not intentional
```suggestion
```

E1izabeth: yes, already deleted

Matvey16:
Verified

ShadelessFox (approved):

PR #37099, opened 2025-02-07
CB-6041 api for internal database information

yagudin10:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBInternalDatabaseInformationProvider.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Internal database information provider.
+ * Contains information about database used for internal needs.
+ */
+public interface DBInternalDatabaseInformationProvider {
+
+    /**
+     * Internal database info.
+     */
+    @NotNull
```
serge-rider, line 31: We need separate info about driver name, URL, database name, database version.
DBInternalDatabaseInformationProvider  -> DBPConnectionInformation

serge-rider (dismissed):

Issue #3957, opened 2018-08-09
commit result enhancement


fayca121:
Hello,

I think it would be useful to have a synchronised tansactions result on the tool bar when you execute **commit;** in the sql editor (no need to do another commit manually by clicking on commit button).

![2018-08-09 15_40_57-dbeaver 5 1 5 201808091355 - _cbiop_ script-33](https://user-images.githubusercontent.com/547548/43906320-e8d328e6-9bea-11e8-9fc4-2444506bc9dd.png)

PR #37103, opened 2025-02-07
dbeaver/dbeaver#3957 sort db level objects

labanovichttps:
How we sort indexes, triggers and others db entities in ascending order. Previous behaviour was incorrect and didn't sort them. 
image
incorrect


Matvey16:
Columns are still sorted alphabetically


https://github.com/user-attachments/assets/b5bdead7-3b4f-4f46-b295-dcfb85ebf0bc

Matvey16:
Verified

PR #37115, opened 2025-02-08
dbeaver/#36632 clear correct preference file for reset ui

### LINKED ISSUE #36632: Font size is not reset by reset settings
### Description

STR
Go to Colors and Fonts - DBeaver fonts - Main font 
Change font size
Go to Help - Reset settings - Applay and close

AR -after restart,ing the font size is not reseted
ER - after restarting, the font size is  reseted

### DBeaver Version

24.3.2

### Operating System

mac 15.1

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

mikiubo:


E1izabeth (approved):
Thanks for the contribution!

ShadelessFox (commented):
Have you tested if this fix actually does something? I'm not even sure this code gets executed at all.

mikiubo:
> Have you tested if this fix actually does something? I'm not even sure this code gets executed at all.

I checked in which file the font configuration is stored and found that the file is different from the one already present in th code.
I tested it on my PC:
Help -> Reset Settings -> Apply and Restart.
Why do you think that code is never executed? 

Here you can see the stack trace:
image


@ShadelessFox

ShadelessFox:
Because we have another mechanism that is supposed to do the same:

```
org.jkiss.dbeaver.core.DBeaverActivator#processResetSettings
```

I believe it exists because the "original" approach doesn't work on some platforms (most likely macOS). I have no idea why it was left, though.

ShadelessFox:
Closing in favor of https://github.com/dbeaver/dbeaver/pull/37298. The mentioned mechanism was broken for some time now.

PR #37119, opened 2025-02-09
add fallback for postgresql sequence last_val

f-m-t-t-m:
Add fallback for last_value if value in `pg_sequences`  is null.


E1izabeth:
Thanks for the contribution. Here is the guidelines for contributors https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code
Please create an issue describing the case you've fixed, especially how to reproduce it.
Also, commit messages must start with the issue number.

PR #37164, opened 2025-02-13
CB-6223 enable flat files driver by default

HocKu7:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBConstants.java
```diff
@@ -154,6 +154,7 @@ public class DBConstants {
     public static final String DRIVER_PARAM_DISABLE_NETWORK_PARAMETERS = "disableNetworkParameters";
     public static final String DRIVER_PARAM_ENABLE_NETWORK_PARAMETERS = "enableNetworkParameters";
     public static final String PARAM_DATABASE_FILE_EXTENSIONS = "database-file-extensions";
+    public static final String PARAM_EXCLUDE_AS_EMBEDDED = "exclude-as-embedded";
```
serge-rider: let's name it safe-embedded-driver

serge-rider (changes requested):

serge-rider (approved):

PR #37169, opened 2025-02-13
dbeaver/pro#3881 Data transfer interruption fixes

### LINKED ISSUE #36526: DBeave hangs if connection is lost during export
### Description

If for some reason user gets disconnected from the database the export will continue indefinately and even can't be cancelled manually

### DBeaver Version

24.3.1

### Operating System

Win11

### Database and driver

PostgreSQL

### Steps to reproduce

1. Start an export from a big table that would take at least a minute
2. During export when rows are being fetched diconnect your machine from the internet
3. Wait for 15+ seconds
4. Reconect to the internet
5. Try to cancel the export or wait

**Expected:** Export is stopped
**Actual:** Export continues and won't end with any result until DBeaver is killed

### Additional context

Depending on how you kill DBeaver there are two results here. If you kill it in task manager everything will work as expected afterwards. But if you right-click on a DBeaver icon in task panel and click close window, dbeaver will appear to be closed, but the task will still hang in task manager and will continue to lock the workspace

serge-rider:


Matvey16:
Some kind of desinchronization (lock) happens here.
STR:
1. Start the export
2. Disconnect the internet
3. Reconnect
4. Abort the export

Export wizard is closed, but there is an endless export task running in the background that locks the connection from certain actions (e.g. chaging transaction mode)

image

Matvey16:
Bug with leftover backgroud job is still repoducible

PR #37191, opened 2025-02-14
#37190 [CUBRID] Fix Error When Invoking Method on a Null Value

### LINKED ISSUE #37190: [CUBRID] Error When Invoking Method on a Null Value
### Description

The error occurs because the getName() method returns null, and calling .toLowerCase() on a null value leads to an error as shown in the screenshot below. 
Image

### DBeaver Version

DBeaver version 24.3.4

### Operating System

Window 10

### Database and driver

- Database Name and Version: CUBRID 11.4
- Driver Name and Version: CUBRID JDBC Driver 11.2

### Steps to reproduce

1. Connect to Cubrid database
2. Right Click on Tables folder
3. Select Create New Table
4. Delete the text in table field

### Additional context

_No response_

longhaseng52:
The error occurs because the getName() method returns null, and calling .toLowerCase() on a null value leads to an error.


ShadelessFox (commented):
Is `toLowerCase` mandatory? Don't we want to show the original names instead, without altering them in any way?

longhaseng52:
@ShadelessFox
Yes, it is necessary to use` toLowerCase`. 
Our database objects are always stored in lowercase, even if entered in uppercase. Since CUBRID treats all objects as lowercase, we must convert them to lowercase for accurate comparisons and other operations.

ShadelessFox:
> Our database objects are always stored in lowercase

If object names are stored and retrieved in lowercase, why is there a need to convert names to lowercase regardless?

E1izabeth (commented):
I don't have questions about these concrete PR. But yes, the idea of the previous PR that brought this lover case conversion is unclear to me.

E1izabeth (dismissed):

longhaseng52:
@ShadelessFox 
The reason we convert name to lowercase, this is because it is automatically transformed to uppercase due to `<parameter name="stored-unquoted-case" value="UPPER"/>` in plugin.xml, which is responsible for converting queries to uppercase in the query editor. Therefore, I need to convert them back to lowercase for every comparison (filtering, searching objects, etc.).

You can refer to the highlighted line below, where the name is transformed to uppercase.

image

ShadelessFox:
It seems like you can change the parameter to `LOWER`, no?

longhaseng52:
@ShadelessFox 
Even changing the parameter to lowercase doesn't help because function `DBObjectNameCaseTransformer.transformName()` at line 249 will still convert it to uppercase anyway.

ShadelessFox:
Look for `stored-unquoted-case` driver property.

longhaseng52:
@ShadelessFox 
We want to keep the `stored-unquoted-case` driver property set to `UPPER` because we want the SQL editor's default case (lowercase) to be converted to uppercase for reserved keywords.

And this `stored-unquoted-case` property also affects database object names, so the only solution I can think of is to convert them to `toLowerCase()`. 

Do you have any suggestions or alternative ways to handle this?

ShadelessFox:
To refactor our model and add missing functionality. But it's way out of the scope of this PR, so I'll approve it and let it go.

ShadelessFox:
Hold on: the `DBPNamedObject#getName` is marked as not null, but we somehow end up with null values being returned from it. I assume the name is assigned during object construction from a result set. Can we check for null there and take appropriate actions?

longhaseng52:
The null value occurs when you edit the name in the UI. For example, when you create a table using the UI and then delete the default name, the `getName()` function will return `null`, leading to a `NullPointerException` error.

ShadelessFox:
This totally could be a bug somewhere in DBeaver's source code.

ShadelessFox (dismissed):

ShadelessFox:
Thanks for the contribution. Passing the PR to our QA team now.

MashaKorax:
verified

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridProcedure.java
```diff
@@ -67,7 +67,7 @@ public CubridProcedure(@NotNull GenericStructContainer container, DBSProcedureTy
     @Override
     @Property(viewable = true, order = 1)
     public String getName() {
-        return super.getName().toLowerCase();
+        return super.getName() != null ? super.getName().toLowerCase() : null;
```
serge-rider: Generally I don't like the idea to do toLowerCase everywhere.
Why can't we just convert object names into lower case on metadata read?
In CubridMetaModel.loadProcedures etc.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridProcedure.java
longhaseng52: @serge-rider 
Even we convert object names into lower case on metadata read, it still affects by the `stored-unquoted-case` driver property that set to UPPER. I have no option but to set it everywhere like this.

serge-rider: But what if identifier was quoted? There is no check for that.
It is not clear why driver returns identifier in one case and then you convert it in uppercase on every getName invocation (besides other side-effects it is really slow, getName can be invoked thousand times per second and it should be fast, toUpperCase is expensive and produces new Java object on each call)

longhaseng52: @serge-rider 
Since CUBRID normalizes both quoted and unquoted identifiers to lowercase, any identifier created in uppercase is automatically converted and stored in lowercase. This ensures consistency when retrieving and displaying object names.

I just realized that applying `toLowerCase()` in `getName()` may impact performance. so I will be better to apply `toLowerCase()` during metadata retrieval or in the constructor instead.
Thank you.

longhaseng52: Hello @serge-rider,
Could you review my updated commit? I would like to know if this approach is acceptable. Thank you.

serge-rider (dismissed):

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridProcedure.java
serge-rider: Yes, this approach is much better than the previous one.

MashaKorax:
@longhaseng52  please take a look at this error

https://github.com/user-attachments/assets/6f4759d8-e1dd-450b-9bf2-7b214141a035

```
java.lang.IllegalArgumentException: Can't write property 'Table Name' value
	at org.jkiss.dbeaver.runtime.properties.PropertySourceEditable.updatePropertyValue(PropertySourceEditable.java:204)
	at org.jkiss.dbeaver.runtime.properties.PropertySourceEditable.setPropertyValue(PropertySourceEditable.java:99)
	at org.jkiss.dbeaver.runtime.properties.PropertySourceAbstract.setPropertyValue(PropertySourceAbstract.java:280)
	at org.jkiss.dbeaver.ui.controls.CustomFormEditor.updatePropertyValue(CustomFormEditor.java:224)
	at org.jkiss.dbeaver.ui.controls.CustomFormEditor.lambda$1(CustomFormEditor.java:190)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:217)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:91)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4321)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1208)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1232)
	at org.eclipse.swt.widgets.Widget.sendEvent(Widget.java:1213)
	at org.eclipse.swt.widgets.Text.wmCommandChild(Text.java:3077)
	at org.eclipse.swt.widgets.Control.WM_COMMAND(Control.java:4909)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4760)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5089)
	at org.eclipse.swt.internal.win32.OS.CallWindowProc(Native Method)
	at org.eclipse.swt.widgets.Text.callWindowProc(Text.java:296)
	at org.eclipse.swt.widgets.Control.windowProc(Control.java:4859)
	at org.eclipse.swt.widgets.Text.windowProc(Text.java:2655)
	at org.eclipse.swt.widgets.Display.windowProc(Display.java:5089)
	at org.eclipse.swt.internal.win32.OS.DispatchMessage(Native Method)
	at org.eclipse.swt.widgets.Display.readAndDispatch(Display.java:3705)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine$5.run(PartRenderingEngine.java:1151)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.e4.ui.internal.workbench.swt.PartRenderingEngine.run(PartRenderingEngine.java:1042)
	at org.eclipse.e4.ui.internal.workbench.E4Workbench.createAndRunUI(E4Workbench.java:153)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:668)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:576)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:173)
	at org.jkiss.dbeaver.ui.app.standalone.DBeaverApplication.start(DBeaverApplication.java:312)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:208)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.runApplication(EclipseAppLauncher.java:149)
	at org.eclipse.core.runtime.internal.adaptor.EclipseAppLauncher.start(EclipseAppLauncher.java:115)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:467)
	at org.eclipse.core.runtime.adaptor.EclipseStarter.run(EclipseStarter.java:298)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.jkiss.dbeaver.launcher.DBeaverLauncher.invokeFramework(DBeaverLauncher.java:823)
	at org.jkiss.dbeaver.launcher.DBeaverLauncher.basicRun(DBeaverLauncher.java:646)
	at org.jkiss.dbeaver.launcher.DBeaverLauncher.run(DBeaverLauncher.java:1604)
Caused by: java.lang.NullPointerException: Cannot invoke "String.toLowerCase()" because "name" is null
	at org.jkiss.dbeaver.ext.cubrid.model.CubridTable.setName(CubridTable.java:80)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
	at org.jkiss.dbeaver.runtime.properties.ObjectPropertyDescriptor.writeValue(ObjectPropertyDescriptor.java:564)
	at org.jkiss.dbeaver.runtime.properties.PropertySourceEditable.updatePropertyValue(PropertySourceEditable.java:190)
	... 40 more

```

longhaseng52:
Hi @MashaKorax
I have fixed this issue. Thanks.

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridView.java
```diff
@@ -37,7 +37,7 @@ public CubridView(
             @Nullable String tableName,
             @Nullable String tableType,
             @Nullable JDBCResultSet dbResult) {
-        super(container, tableName, tableType, dbResult);
+        super(container, tableName.toLowerCase(), tableType, dbResult);
```
ShadelessFox: `tableName` might be null.

ShadelessFox (commented):
We've previously discussed that you shouldn't alter names anywhere in the model. And instead, convert them to lowercase when you fetch them from the database. Please address this in the future.

MashaKorax:
verified

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridView.java
longhaseng52: @ShadelessFox 
You're right — my apologies for the oversight.
I’ve corrected the issue as per our previous discussion.
Thank you.

ShadelessFox (approved):

PR #37219, opened 2025-02-17
#37089 Semantic completion proposals sorting

### LINKED ISSUE #37089: AutoComplete infer Join criteria from foreign keys
This is something I depend on heavily in MSSQL/Redgate and in others like DbForge that I find to be essentially the only blocker I have to switching full time to DBeaver.

This is in MSSQL Server, using the DBeaver Community edition.

When I write a query involving a join between two tables I would like to have the AutoComplete automatically suggest join 'ON' conditions based on the keys the two tables share.

e.g. (in bold)
SELECT * FROM dbo.Taxi t INNER JOIN dbo.TaxiPassenger tp ON _**t.Id = tp.TaxiId**_

I searched here for previous questions about this and only found one match that seemed to imply that it exists but it doesn't for me: it just gives me a bland list of all fields on both tables.

E1izabeth:
1. Join condition is always at the top
image
2. If Search inside names is OFF, then alphabetical sorting works only inside proposal types (i.e., we show tables first and sort them alphabetically, then schemas and sort them, and so on)
image

3. If Search inside names is ON, the fuzzy score has more priority than alphabetical order when prefix is typed
image


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/completion/SQLQueryCompletionItemKind.java
```diff
@@ -21,59 +21,62 @@ public enum SQLQueryCompletionItemKind {
...
 
 
     public final boolean isTableName;
     public final boolean isColumnName;
-
+    public final int sortOrder;
 
     SQLQueryCompletionItemKind() {
-        this(false, false);
+        this(false, false, Integer.MAX_VALUE);
```
ShadelessFox, line 73: I don't think we need that constructor anymore

ShadelessFox (dismissed):

Matvey16:
Verified

ShadelessFox (approved):

PR #37221, opened 2025-02-17
dbeaver/pro#4078 Hide `New Constraint from Selection` for managers

E1izabeth:
`New Constraint from Selection` should not be visible for managers, but should be available for devs and admins


kseniaguzeeva:
verified (admin, developer have, manager doesn't have)

PR #37249, opened 2025-02-19
dbeaver/dbeaver#37194 Check constraints read fix

### LINKED ISSUE #37194: dbeaver shows each constrain doubled or sometimes tripled
### Description

dbeaver shows each constrain doubled or sometimes tripled the first time it is viewed. If the screen is refereshed with the Refresh button, the problem goes away.

### DBeaver Version

24.3.4.202502021521

### Operating System

Windows 10

### Database and driver

MySQL 8

### Steps to reproduce

Create constraints.
Close the program.
Start the program and then go look at the constraints again.

### Additional context

Here is an example of what it looks like when them doubled. Sometimes they are tripled.

Image

serge-rider:


Matvey16:
Verified

greekcntr:
This was not fixed in the latest release, version 25. Here is an example with the constraints quadrupled.
image

PR #37250, opened 2025-02-19
dbeaver/pro#4002 Fix refreshing table comment in ERD

IvanGorshechnikov:
Previously, when selecting to show comments in ERD, comments for columns were displayed immediately, but for tables, you had to press ‘Refresh’. There was a similar issue with the ‘Fully qualified name’ option

Now, when this option is enabled, the refresh occurs immediately for table names and their columns


plugins/org.jkiss.dbeaver.erd.ui/src/org/jkiss/dbeaver/erd/ui/editor/ERDEditorPart.java
```diff
@@ -1155,7 +1155,7 @@ public void run()
         {
             getDiagram().setAttributeStyle(style, !isChecked());
             refreshEntityAndAttributes();
-            refreshDiagram(true, false);
+            refreshDiagram(true, true);
```
E1izabeth, line 1158: That's probably not part of the current issue, but we lose user's layout on refreshing the diagram (Keep layout checked)

E1izabeth (approved):

ShadelessFox (approved):

uslss:
verified

PR #37299, opened 2025-02-24
dbeaver/pro#4204 lock truststore

alexander-skoblikov:
closes dbeaver/pro#4204


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/lock/FileLockInfo.java
```diff
@@ -0,0 +1,105 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.fs.lock;
+
+import org.jkiss.code.NotNull;
+import org.jkiss.code.Nullable;
+
+public class FileLockInfo {
```
ShadelessFox, line 22: `record`? Please?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/lock/FileLock.java
```diff
@@ -0,0 +1,64 @@
...
+import org.jkiss.dbeaver.Log;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * Resource Manager resource lock
+ */
+public class FileLock implements AutoCloseable {
```
ShadelessFox, line 29: This class it not exposed anywhere in the API. Perhaps it should be package-private?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/lock/FileLockInfo.java
ShadelessFox, line 22: This class is not exposed anywhere in the API as well

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/lock/FileLockController.java
```diff
@@ -0,0 +1,251 @@
...
+    private final Path lockFolderPath;
+    private final String applicationId;
+    private final long maxLockTime;
+
+    public FileLockController(@NotNull String applicationId) throws DBException {
+        this(applicationId, DEFAULT_MAX_LOCK_TIME);
+    }
+
+    // for tests
+    public FileLockController(@NotNull String applicationId, long maxLockTime) throws DBException {
```
ShadelessFox, line 57: What the hell? So we're now polluting the storage during tests. Can we please not and instead make it properly testable? By the way, I don't see any tests.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/lock/FileLockController.java
```diff
@@ -0,0 +1,251 @@
...
+     * or force intercept lock, if the operation will take too long and
+     * exceeds the maximum available locking time {@link #maxLockTime} or the lock is invalid {@link #awaitUnlock)}.
+     *
+     * @param lockFileName - file to be locked
+     * @param operationName - executed operation name
+     * @return - lock
+     */
+    @NotNull
+    public FileLock lock(@NotNull String lockFileName, @NotNull String operationName) throws DBException {
+        synchronized (FileLockController.class) {
```
ShadelessFox, line 76: Who on Earth `synchronize` on a class? No, just no, don't ever do that

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/lock/FileLockController.java
```diff
@@ -0,0 +1,251 @@
...
+            try {
+                FileLockInfo lockInfo = new FileLockInfo.Builder(UUID.randomUUID().toString())
+                    .setApplicationId(applicationId)
+                    .setOperationName(operationName)
+                    .setOperationStartTime(System.currentTimeMillis())
+                    .build();
```
ShadelessFox, line 82: What's the point of this builder if it's only used here and nowhere else?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/lock/FileLockController.java
```diff
@@ -0,0 +1,251 @@
...
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.Duration;
+import java.util.UUID;
+
+/**
+ * File based resource locks
+ */
+public class FileLockController {
```
ShadelessFox, line 38: What does it even do? I see that it creates files with random names somewhere inside the workspace folder. It's now how lock files work. They prevent a directory from being deleted. Moreover, it doesn't even work, e.g., on Windows.
I believe it serves a different purpose, but I don't quite see it. Is it supposed to be used across multiple services?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/lock/FileLock.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Unlock resource and remove .lock file
+     */
+    public void unlock() {
+        try {
+            Files.deleteIfExists(lockFilePath);
+        } catch (IOException e) {
+            log.error("Failed to unlock file: " + lockFilePath, e);
+            if (Files.exists(lockFilePath)) {
+                // file still locket, try to unlock again
```
flambchap, line 47: image

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/lock/FileLockController.java
alexander-skoblikov: > Who on Earth synchronize on a class

me

this is necessary in situations where several lock controllers access the same file

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/lock/FileLockController.java
alexander-skoblikov: > So we're now polluting the storage during tests

nope, because for tests used test workspace which is automatically cleared

>  I don't see any tests.

tests still in cloudbeaver repo, to reduce the code in the pr

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/lock/FileLockController.java
alexander-skoblikov: it creates a special lock in the workspace, and as long as this file exists (or the timeout is not exceeded) - other processes/requests to the file will wait for the lock to be released (only if the file was accessed via the lock controller - direct io operations are still available)
this is independent of the operating system, as it is custom logic, and is not related to nio FileLock

PR #37326, opened 2025-02-25
dbeaver/pro#4220 set file crc fix for downloaded driver

yagudin10:
closes dbeaver/pro#4220


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/connection/DBPDriverLibrary.java
```diff
@@ -98,6 +98,11 @@ public static FileType getFileTypeByFileName(String fileName) {
     @Nullable
     Path getLocalFile();
 
+    /**
+     * Returns CRC of library file.
+     */
+    long getFileCRC();
```
ShadelessFox, line 104: CRC32 is 32-bit. Can we please update `DriverDescriptor#calculateFileCRC` so it returns `int`, please? Pretty please?

serge-rider: Disagreed. We can change algorithm to use 64bit crc an we won't need to change API.
t is a bad idea to link API to a particular algorithm

ShadelessFox: I'm 99% sure this API was designed around the current implementation of `DriverDescriptor#calculateFileCRC`. CRC is the name of the family of algorithms. If the API is supposed to be algorithm-agnostic, it should've been called `getFileChecksum` instead, not `getFileCRC`.

alexander-skoblikov (approved):

PR #37337, opened 2025-02-26
dbeaver/pro#4006 Fix scroll active selection

labanovichttps:
start editing and scroll, then edit one more time 
image


ShadelessFox (dismissed):

ShadelessFox:
Does it also scroll when closing the editor?

labanovichttps:
> Does it also scroll when closing the editor?

https://github.com/user-attachments/assets/91c408a5-2788-4e48-ade1-e2fe3cd23665

@ShadelessFox yes

ShadelessFox:
I mean when pressing the Escape button or applying changes using Enter.

labanovichttps:
> I mean when pressing the Escape button or applying changes using Enter.

Added one more listener

ShadelessFox (approved):
Not sure if this is the ideal approach, but I believe it's acceptable

E1izabeth (approved):

Matvey16:
Verified

PR #37346, opened 2025-02-26
dbeaver/pro#3994 Fix connection dialog icon draw

ShadelessFox:
Please test against Windows, Linux, and macOS against both 100% and 200% scaling (HiDPI).


E1izabeth (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified

PR #37347, opened 2025-02-26
dbeaver/pro#3935 Fix skip statement before the command

labanovichttps:
Shouldn't throw any exception if statement goes before without delimiter 
image


E1izabeth (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified

PR #37361, opened 2025-02-27
dbeaver/dbeaver#35474 Fix cancelling of retry cancelled query

### LINKED ISSUE #35474: Feature "Cancel active query" doesn't work after "Retry", only kill session manually works to stop the query (or disconnect the connection)
### Description

Notice that the query doesn't stop after the second click to "Cancel active query" so I have to kill it manually with a query from another tab.
The button to stop the query on a tab "Background task" also doesn't work (inactive).

*Please also check the behaviour on other databases, not only postgres.

https://github.com/user-attachments/assets/a3320bd7-f88a-484d-9871-6f5f2e1be506

Use code 1:
```
language plpgsql
as
$$
begin
	select pg_sleep(10000);
end;
$$
;

call proc1();
```

Use code 2:

```
SELECT * FROM pg_stat_activity
WHERE state = 'active' ;


select pg_terminate_backend(9999) ;
```

### DBeaver Version

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.4

### Steps to reproduce

_No response_

### Additional context

_No response_

IvanGorshechnikov:


E1izabeth (dismissed):

ShadelessFox (dismissed):

Matvey16:
Verified

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/execute/SQLQueryJob.java
```diff
@@ -267,7 +266,8 @@ protected IStatus run(DBRProgressMonitor monitor)
                                 break;
                             case RETRY:
                                 // just make it again
-                                continue;
+                                this.schedule(100);
```
serge-rider, line 273: Not clear process flow.
Retry must re-execute only the failed query, not entire script

PR #37362, opened 2025-02-27
dbeaver/pro#4255 Fix sql script parser unit tests

labanovichttps:


test/org.jkiss.dbeaver.test.platform/src/org/jkiss/util/SQLEditorTestUtil.java
```diff
@@ -0,0 +1,74 @@
...
+package org.jkiss.util;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public final class SQLEditorTestUtil {
+
+    public static final String CURSOR = "<-|";
```
E1izabeth, line 26: I'd use just `|`, but it's a question of taste.

E1izabeth (approved):

IvanGorshechnikov (approved):

test/org.jkiss.dbeaver.test.platform/src/org/jkiss/util/SQLEditorTestUtil.java
labanovichttps: @E1izabeth I wanted to use it too, but if u wanna use **concat** in ur queries it will be incorrect to use `|` as a cursor

IvanGorshechnikov: I think that `<-|` is a better option because it is easier to see in the text with your eyes

serge-rider (approved):

Issue #37034, opened 2025-02-04
Select multiple columns in data grid


rhuesken:
### Description

I often use the map to display the geometry columns from the selected rows. I have a query that contains multiple geometry columns, e.g. the line and the start and endpoints

Postgis query example:

`select i.*,  ST_StartPoint(i.geomz_wgs) start_point, ST_EndPoint(i.geomz_wgs) end_point 
from   pros_index_tracks i
where  ...`

When I click a line (gemetry cell in the data grid), it is displayed on the map.

![Image](https://github.com/user-attachments/assets/bebe60e0-a2d2-468c-9280-94439277d0ed)

When I use CTRL-click on another cell to select a start point, it is also displayed on the map.
However, much more columns are selected and the grid is shifted, the left most columns are displayed now.

![Image](https://github.com/user-attachments/assets/24f12bec-96e8-440b-ba89-c3b571bc02be)

This is annoying, since it used to be working. Can you please have a look and fix this?

### DBeaver Version

Community edition Version 24.3.4.202502021521

### Operating System

Windows

uslss:
thanks for the report

PR #37392, opened 2025-03-03
dbeaver/dbeaver#37034 Fix selecting multiple GIS columns

IvanGorshechnikov:


ShadelessFox (dismissed):

labanovichttps (dismissed):

E1izabeth (dismissed):

Matvey16:
Still reproducible

Matvey16:
https://github.com/user-attachments/assets/aea64e55-b952-4fdc-ade2-57b17abf4da3

labanovichttps (approved):

plugins/org.jkiss.dbeaver.data.gis.view/src/org/jkiss/dbeaver/ui/gis/panel/GISLeafletViewer.java
```diff
@@ -126,7 +126,7 @@ public GISLeafletViewer(Composite parent, @NotNull DBDAttributeBinding[] binding
         CSSUtils.setCSSClass(composite, DBStyles.COLORED_BY_CONNECTION_TYPE);
         browserCreating = true;
         try {
-            browser = new Browser(composite, SWT.NONE);
+            browser = new Browser(composite, SWT.WEBKIT);
```
E1izabeth, line 129: Why? As far as I remember, it may break browser creation on some OSs

ShadelessFox: Yeah, I'm not fond of this fix either

Destrolaric: Really bad idea - IE is deprecated and no longer supported by Microsoft whatsoever. There is currently an effort from Eclipse to finally make EDGE the default config, so we should avoid using IE at all times.

Destrolaric (changes requested):
Forcing users to use deprecated and insecure browser is a bad idea

plugins/org.jkiss.dbeaver.data.gis.view/src/org/jkiss/dbeaver/ui/gis/panel/GISLeafletViewer.java
IvanGorshechnikov: Yes, I understand. Unfortunately, this bug is reproduced only with the Edge browser, and I can't find any solution except switching to IE for Windows. We can also do anything with that and wait until Eclipse developers fix it.

PR #37406, opened 2025-03-03
#36881 Fix formatting of user-defined functions

### LINKED ISSUE #36881: Formatter has too aggressive line-breaking behaviour when cast to decimal() vs. varchar()
### Description

If I have a query like:
`select generate_series(1,10,1)::decimal(10,2);`

And I use the format SQL function, it makes it a very long chunk of code and unnecessarily added line break between the 2 arguments in decimal()

Image

It seems only happen to the data format with multiple arguments. If I instead casting it to varchar(), everything seems fine, despite I put a very long argument in it:

Image

I am not sure if this has something to do with the line length. Unless there is a setting in the preference for the preferred line length, the line breaking choice for the formatter seems too aggressive. To be honest, I don't even think the generate_series() needs to be broken into 3 lines either...
 

### DBeaver Version

Community Edition 24.3.3.202501191633

### Operating System

Windows 11

### Database and driver

PostgreSQL 9.4.26

### Steps to reproduce

1. Write a query with casting that contains multiple arguments like `select generate_series(1,10,1)::decimal(10,2);`
2. Use Ctrl+Shift+F or select **formatting SQL**
3. See the result

### Additional context

_No response_

E1izabeth:
After formatting:
image


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/format/SQLFormatterConfiguration.java
```diff
@@ -110,6 +112,12 @@ public boolean isFunction(String name) {
         return syntaxManager.getDialect().getFunctions().contains(name.toUpperCase(Locale.ENGLISH));
     }
 
+    public boolean isIdentifier(String name) {
+        SQLDialect dialect = syntaxManager.getDialect();
+        boolean isQuoted = dialect.isQuotedIdentifier(name);
+        return name.chars().mapToObj(c -> (char) c).allMatch(c -> dialect.validIdentifierPart(c, isQuoted));
```
ShadelessFox: Please don't box, that pipeline step is completely redundant

```suggestion
        return name.chars().allMatch(c -> dialect.validIdentifierPart((char) c, isQuoted));
```

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/format/SQLFormatterConfiguration.java
```diff
@@ -110,6 +112,12 @@ public boolean isFunction(String name) {
         return syntaxManager.getDialect().getFunctions().contains(name.toUpperCase(Locale.ENGLISH));
     }
 
+    public boolean isIdentifier(String name) {
```
ShadelessFox: Annotations, as well

ShadelessFox (approved):

IvanGorshechnikov (approved):

MashaKorax:
verified

PR #37419, opened 2025-03-04
dbeaver/dbeaver#36382 delay popups

### LINKED ISSUE #36382: Delay popup for password until after update check
### Description

Today if you have a db connection without a saved password it pops up when opening dbeaver if you had any scripts open that are connected to that db.

It also looks for updates when starting, so most of the time my team and I get frustrated when trying to type the password we get a new popup and focus comes up to "Update now" (sometimes we are using Dbeaver on a client machine and want to simply run a query to fix something and then we end up updating the whole program).

Id like to request a delay on the popup for password until after update check or the other way around (first preferred).

### DBeaver Version

Community Edition 24.2.5.202411171748

### Operating System

Windows 11 Pro

### Database and driver

Not relevant.

### Steps to reproduce

1. Open a script to a database without saved credentials
2. Close DBeaver with that script open
3. Wait for an update (🙃)
4. Open DBeaver again and wait a moment, you'll see both popups show up (first the password and then the update)

### Additional context

_No response_

labanovichttps:
`open ~/Library/DBeaverData/workspace6/.metadata/.plugins/org.eclipse.core.runtime/.settings/   ` 
add/replace to your /.metadata/.plugins/org.eclipse.core.runtime/.settings/org.jkiss.dbeaver.core.prefs
sample.database.canceled=false
tipOfTheDayInitializer.notFirstRun=true
ui.show.tip.of.the.day.on.startup=

Open a script to a database without saved credentials
Close DBeaver with that script open
Wait for an update/tip of the day/Sample database
Open DBeaver and check that any popups won't open until u don't close the Auth popup. 
See


ShadelessFox (changes requested):
Can we have a single utility method instead of copying the same code in every place we want such a functionality? You can easily grab a list of all `Shell`s currently visible. You can determine their modality as well. That's all you need.

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -2528,6 +2529,44 @@ public static String getCatalogSchemaTerms(@Nullable DBPDataSourceContainer data
...
+     * @param popupAction  the action to be executed when the main shell is active; must not be null.
+     * @param initialDelay the initial delay in milliseconds before the job is first scheduled.
+     * @param retryDelay   the delay in milliseconds before the job is rescheduled if the main shell is not active.
+     * @param jobName      the name of the UIJob for identification purposes; must not be null.
+     */
+    public static void scheduleDelayedPopup(
+        @NotNull Shell mainShell,
+        @NotNull Runnable popupAction,
+        long initialDelay,
+        long retryDelay,
```
ShadelessFox: I doubt we actually need to have these configured externally. Do we?

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -2528,6 +2529,44 @@ public static String getCatalogSchemaTerms(@Nullable DBPDataSourceContainer data
...
+        @NotNull String jobName
+    ) {
+        new UIJob(mainShell.getDisplay(), jobName) {
```
ShadelessFox: Call `setSystem(true)` so it won't appear in the progress monitor

ShadelessFox (dismissed):

E1izabeth (dismissed):

uslss:
Tip of the day isn't shown on start even if it is set to tipOfTheDayInitializer.notFirstRun=true 
Also nothing happens when I'll try to open tips from main menu
image

ShadelessFox (approved):

E1izabeth (approved):

PR #37421, opened 2025-03-04
dbeaver/dbeaver#37332 Fix showing function comments in DDL in Postgres

### LINKED ISSUE #37332: Incorrect display of the "\" symbol in the comment block (function editing window)
### Description

Incorrect display of the "\\" symbol in the comment block (function editing window)

### DBeaver Version

CE 24.3.5.202502162012

### Operating System

Win 10 x64

### Database and driver

PostgreSQL

### Steps to reproduce

Creating a new function, we specify a comment in which there is a "\\" character.

```pgsql
CREATE OR REPLACE FUNCTION public.test()
 RETURNS jsonb
 LANGUAGE plpgsql
AS $function$
  begin        
    return '{"test":"01.01.2025"}';
  end;
$function$
;

COMMENT ON FUNCTION public.test() IS 'Name: test
#return: 
{"JSON_SCHEMA":{
  "example":"{\"test\":\"01.01.2025\"}",
  "type":"object",
  "properties":{
    "test":{
      "type":"string",
      "pattern":"\d\d\.\d\d\.\d\d\d\d"
    }
  }
}}';

```

the "\\" symbol doubles when displayed 

Image

and the number doubles with each subsequent save

Image

The comment is saved correctly in the database.

```sql
select obj_description('public.test()'::regprocedure,'pg_proc');    
```

```text
Name: test
#return: 
{"JSON_SCHEMA":{
  "example":"{\"test\":\"01.01.2025\"}",
  "type":"object",
  "properties":{
    "test":{
      "type":"string",
      "pattern":"\d\d\.\d\d\.\d\d\d\d"
    }
  }
}}
```

Image

### Additional context

_No response_

IvanGorshechnikov:
Steps to reproduce:
1. Create function with comment
```
CREATE OR REPLACE FUNCTION public.test()
 RETURNS jsonb
 LANGUAGE plpgsql
AS $function$
  begin        
    return '{"test":"01.01.2025"}';
  end;
$function$
;

COMMENT ON FUNCTION public.test() IS 'Name: test
#return: 
{"JSON_SCHEMA":{
  "example":"{\"test\":\"01.01.2025\"}",
  "type":"object",
  "properties":{
    "test":{
      "type":"string",
      "pattern":"\d\d\.\d\d\.\d\d\d\d"
    }
  }
}}';
```
2. Open it in DBeaver and click "Show comments" 
3. Make any change, for example, add extra line
4. Save
5. Refresh it, and it should be doubled without that PR


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreProcedure.java
```diff
@@ -461,7 +461,8 @@ public String getObjectDefinitionText(DBRProgressMonitor monitor, Map<String, Ob
             procDDL += ";\n";
 
             if (CommonUtils.getOption(options, DBPScriptObject.OPTION_INCLUDE_COMMENTS) && !CommonUtils.isEmpty(getDescription())) {
-                procDDL += "\nCOMMENT ON " + getProcedureTypeName() + " " + getFullQualifiedSignature() + " IS " + SQLUtils.quoteString(this, getDescription()) + ";\n";
+                procDDL += "\nCOMMENT ON " + getProcedureTypeName() + " " + getFullQualifiedSignature() + " IS "
```
ShadelessFox: I'm honestly not sure about this fix. Maybe the original approach is incorrect to begin with?

E1izabeth: I agree. It's unclear why we're replacing it here but not elsewhere. I believe this means that this bug can appear in other places. We need a general approach to solve this.

ShadelessFox (approved):

E1izabeth (approved):

MashaKorax:
Does not correctly escape the escape characters
iScreen Shoter - DBeaver Community - 250310141725

ShadelessFox:
The behavior now depends on https://www.postgresql.org/docs/current/runtime-config-compatible.html#GUC-STANDARD-CONFORMING-STRINGS

MashaKorax:
verified

PR #37426, opened 2025-03-04
CB-6089 write file stream directly to http response

HocKu7:


yagudin10 (dismissed):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/IDataTransferConsumer.java
```diff
@@ -40,13 +41,21 @@ class TransferParameters {
         public boolean isBinary;
         public boolean isHTML;
         public Date startTimestamp;
+        public boolean writeDirectToStream;
```
serge-rider: Redundant variable. Stream is enough.
streamForDirectWriting -> exportToStream

serge-rider (dismissed):

yagudin10 (approved):

PR #37439, opened 2025-03-05
dbeaver/dbeaver#22652 Fix dollar-quoted format string

### LINKED ISSUE #22652: A query containing dollar-quoted format string causes SQL editor to run other commands unintentionally (Postgres)
### Discussed in https://github.com/orgs/dbeaver/discussions/22543

<sup>Originally posted by **philosaur** December 28, 2023</sup>
### Description

Running a query (Execute SQL query) that contains a dollar-quoted format string causes DBeaver to run that query as well as all subsequent queries in the same editor--as if "Execute SQL script" had been run instead.

My last version of DBeaver was from probably 6 moths ago, and this behavior began when I upgraded to the Christmas Day version.

### DBeaver Version

Community Edition, vers. 23.3.1.202312241705

### Operating System

Windows 10

### Database and driver

Postgres, (PostgreSQL JDBC Driver)

### Steps to reproduce

1. Click "Execute SQL Script" or press F5 (by default) with this SQL:
```
SELECT format($$%1$s$$, 'foo');

SELECT 'bar';

SELECT 'baz1';

SELECT 'baz2';

SELECT 'baz3';
```
2. DBeaver produce five separate output tabs with output from each respective command, instead of running just the first command.  
3. Switching the dollar quotes to single quotes causes it to only run the first query, as expected.

IvanGorshechnikov:
In this PR, the identification of named tokens of dollar quotes (e.g., $s$) has been changed. https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING
Added a check to see if there is a paired token for this named token. Previously broke on this query SELECT format($$%1$s$$, ‘foo’); It was identifying $s$ as a separate token, so it was incorrectly highlighting and scoping the query boundaries


E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/rules/SQLDollarQuoteRule.java
```diff
@@ -56,16 +58,16 @@ public TPToken getSuccessToken() {
...
 
+    @NotNull
     @Override
-    public TPToken evaluate(TPCharacterScanner scanner, boolean resume) {
+    public TPToken evaluate(@NotNull TPCharacterScanner scanner, boolean resume) {
         String start = this.tryReadDollarQuote(scanner);
         if (start != null) {
-            if ((start.length() == 2 && this.fullyConsumeUnnamed) || (start.length() > 2 && this.fullyConsumeNamed)) {
-                int c = scanner.read();
+            if ((start.length() == 2 && this.fullyConsumeUnnamed) || (start.length() > 2 && this.fullyConsumeNamed)) {int c = scanner.read();
```
ShadelessFox:
```suggestion
            if ((start.length() == 2 && this.fullyConsumeUnnamed) || (start.length() > 2 && this.fullyConsumeNamed)) {
                int c = scanner.read();
```

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/rules/SQLDollarQuoteRule.java
```diff
@@ -129,10 +139,59 @@ private String tryReadDollarQuote(TPCharacterScanner scanner) {
...
+
+
+        while (c != TPCharacterScanner.EOF) {
+            buffer.append((char) c);
+            c = scanner.read();
+            readCount++;
+            if (buffer.length() > matchLength) {
+                buffer.deleteCharAt(0);
+            }
+            if (buffer.toString().equals(quoteName)) {
```
ShadelessFox: Consider using `compareTo` instead to avoid extra allocations

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/rules/SQLDollarQuoteRule.java
```diff
@@ -129,10 +139,59 @@ private String tryReadDollarQuote(TPCharacterScanner scanner) {
...
+        while (scanner.getOffset() > 0) {
+            scanner.unread();
+            readCount++;
+        }
+
+        while (readCount-- > 0) {
+            prefixBuffer.append((char) scanner.read());
+        }
+        prefixBuffer.delete(prefixBuffer.length() - quoteName.length(), prefixBuffer.length());
+        return prefixBuffer.toString().contains(quoteName);
```
ShadelessFox: Consider using `indexOf` to avoid extra allocations

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/rules/SQLDollarQuoteRule.java
```diff
@@ -129,10 +139,59 @@ private String tryReadDollarQuote(TPCharacterScanner scanner) {
...
+        StringBuilder prefixBuffer = new StringBuilder();
+        int readCount = 0;
+
+        while (scanner.getOffset() > 0) {
+            scanner.unread();
+            readCount++;
+        }
+
+        while (readCount-- > 0) {
+            prefixBuffer.append((char) scanner.read());
```
ShadelessFox: Consider performing post-increment in the body to mirror the loop above

ShadelessFox (approved):

E1izabeth (approved):

Matvey16:
Reproducible in a case where `$s$` can be foud twice

https://github.com/user-attachments/assets/3e93c9b3-e5aa-4546-8439-1b28ad79b92d

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/rules/SQLDollarQuoteRule.java
```diff
@@ -129,10 +140,54 @@ private String tryReadDollarQuote(TPCharacterScanner scanner) {
...
+    }
+
+    /**
+     * Checks if there is an OPEN quote $name$ before the current offset.
+     * After checking, restores the position.
+     */
+    private boolean hasOpeningDollarQuoteBefore(@NotNull TPCharacterScanner scanner, @NotNull String quoteName) {
+        StringBuilder prefixBuffer = new StringBuilder();
+        int offset = scanner.getOffset();
+        unread(scanner, offset);
```
E1izabeth, line 178: It will greatly influence performance on parsing and highlighting of big scripts

E1izabeth (changes requested):

E1izabeth:
Please consider switching these two settings 

image

PR #37465, opened 2025-03-07
#36904 Enhance execution error result set view

### LINKED ISSUE #36904: Make "Go To Error" work on double click of error message
In SSMS, if an error occurs with a query, I can double click the error and it will highlight where the error occurred

Image

DBeaver has similar functionality where if you get an error you can click the "Go To The Error" target reticle button and that will move the cursor to the point of the statement with the error (I prefer highlighting as it's more easily noticeable but it's sufficient as is I suppose and highlighting could pose difficulties)

Perhaps a nitpick on my part, but could we make it so that if you double click the error and/or error icon it will do the same thing as when you click the "Go To The Error" button.  The reason for this is that:

1) I guess I'm just used to SSMS where I see the error and I want to double click it to go to the error
2) The "Go To The Error" icon is small whereas the error and error icon are obviously very large so I would find it quicker/more efficient to just double click that
3) I wouldn't think it would be that hard of a change

Just clicking the "Go To The Error" button.

Thanks for the consideration!

E1izabeth:
image


ShadelessFox (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified

PR #37479, opened 2025-03-10
dbeaver/pro#4285 Fix xml/xlsx data importing for db tasks

labanovichttps:


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/StreamProducerSettings.java
```diff
@@ -130,10 +130,14 @@ public void updateProducerSettingsFromStream(
         @NotNull DataTransferSettings dataTransferSettings
     ) {
         try {
+            var processor = dataTransferSettings.getProcessor();
+            if (processor == null) {
+                throw new DBException("Processor for data import/export is null.");
```
E1izabeth: As far as I can see, we don't catch this exception, so the message will be shown to the user. Let's make it more friendly and explainable. I believe not many users know what is null.

labanovichttps: Changed to **Processor haven't chosen for data import/export (e.g., XML, CSV, etc.).**

E1izabeth (approved):

ShadelessFox (approved):

PR #37494, opened 2025-03-11
CB-6302 add ability to connect by url for hive

yagudin10:


plugins/org.jkiss.dbeaver.ext.hive/plugin.xml
```diff
@@ -36,6 +36,7 @@
                         description="Apache Hive JDBC"
```
LonwoLonwo:
```suggestion
                        description="Apache Hive JDBC (Legacy)"
```

LonwoLonwo (approved):

PR #37497, opened 2025-03-11
Don't highlight tokens as usual in sql control commands

E1izabeth:
image


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/SQLCommandModelRecognizer.java
```diff
@@ -103,8 +115,23 @@ public static SQLQueryModel recognizeCommand(
             scriptContext.getVariable(varEntry.name().toUpperCase(Locale.ENGLISH));
             cmdModel.addVariable(symbolEntry, variablesResolver == null ? "?" : variablesResolver.get(varEntry.name()));
         }
+    }
 
-        return new SQLQueryModel(fakeTree, cmdModel, symbolEntries, Collections.emptyList());
+    private static void registerCommandParameter(@NotNull String text, @NotNull Set<SQLQuerySymbolEntry> symbolEntries) {
```
labanovichttps, line 120: (minor) Add formatting for params

labanovichttps (approved):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/SQLCommandModelRecognizer.java
E1izabeth: The line is not too long to use such formatting

ShadelessFox (approved):

Matvey16:
Verified

PR #37500, opened 2025-03-11
Dynamic classpath

serge-rider:


Matvey16:
Can't connect with sspi

mstsc_KqdA94mVC8

mstsc_irEK6W4W21

PR #37505, opened 2025-03-11
dbeaver/dbeaver#37303 Refresh password in UI when test connection

### LINKED ISSUE #37303: [MariaDB] Expired password doesn't change on connection creation
25.0.0 EA
MariaDB 10.6.5

The expired password doesn't change in the connection setting if it was changed during Test Connection

https://github.com/user-attachments/assets/310ddfde-48f7-4599-8fd1-a6e8b3b3a510

_Originally posted by @uslss in https://github.com/dbeaver/dbeaver/issues/37275#issuecomment-2678788565_

This should be changed in the connection configuration when changed in the Test connection.

labanovichttps:


E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionWizard.java
```diff
@@ -265,6 +275,21 @@ public void testConnection() {
         }
     }
 
+    private void updatePasswordInComposite(
```
ShadelessFox: It's a dirty approach. I'm not sure...

serge-rider: Very dirty.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -1414,6 +1414,17 @@ public static void fireObjectUpdate(DBSObject object, @Nullable Map<String, Obje
         }
     }
 
+    /**
+     * Send event for reset datasource password
+     */
+    public static void firePasswordUpdate(DBSObject object, Object data) {
```
ShadelessFox: This signature and javadoc don't tell much about what this method does.

It needs:
- The data source the password is updated of
- The new password

However, it instead accepts an arbitrary `DBSObject` and an untyped `Object` containing some data.

Moreover, this event seems specific. I doubt we want to pollute our events model with redundant event types. Perhaps we can use `OBJECT_UPDATE` and pass a special object denoting a password change in `data`?

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionWizard.java
```diff
@@ -233,6 +234,10 @@ public void testConnection() {
                         throw new InterruptedException("cancel");
                     }
                 });
+                var changedUserPassword = targetDataSource.getActualConnectionConfiguration().getUserPassword();
+                if (changedUserPassword != null) {
+                    DBUtils.firePasswordUpdate(activeDataSource, changedUserPassword);
```
ShadelessFox: It should check if the password differs from the original password

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
labanovichttps: Add some fixes for javadoc, but `PASSWORD_UPDATE` seems like the nice decision,  cuz if changed to `OBJECT_UPDATE`, there will be unnecessary checks for catching

ShadelessFox: > PASSWORD_UPDATE seems like the nice decision, cuz if changed to OBJECT_UPDATE, there will be unnecessary checks for catching

It's a generic model. You are polluting it with things for dealing with your specific case.

labanovichttps: fix

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/DateTimeCustomValueHandler.java
```diff
@@ -126,4 +126,12 @@ public void refreshValueHandlerConfiguration(DBSTypedObject type) {
     @NotNull
     protected abstract String getFormatterId(DBSTypedObject column);
 
+    @NotNull
+    public Object getQuotedValue(@NotNull Object value) {
```
ShadelessFox: Why is it here?

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/AuthModelSelector.java
```diff
@@ -164,6 +167,25 @@ public void loadSettings(
         changeAuthModel();
     }
 
+    @Override
+    public void handleDataSourceEvent(DBPEvent event) {
+        if (event.getAction() == DBPEvent.Action.OBJECT_UPDATE) {
+            UIUtils.asyncExec(() -> {
+                if (event.getData() instanceof DBPConnectionConfiguration newConfig) {
```
ShadelessFox: Can be merged to just one `if`

```suggestion
        if (event.getAction() == DBPEvent.Action.OBJECT_UPDATE && event.getData() instanceof DBPConnectionConfiguration newConfig) {
            UIUtils.asyncExec(() -> {
```

ShadelessFox (dismissed):

E1izabeth (dismissed):

ShadelessFox (approved):

PR #37518, opened 2025-03-12
dbeaver/pro#3907 Fix NPE when the application stops

labanovichttps:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/fs/DBNFileSystem.java
```diff
@@ -154,6 +154,9 @@ protected DBNFileSystemRoot[] readChildNodes(
         @NotNull DBRProgressMonitor monitor,
         @Nullable DBNFileSystemRoot[] mergeWith
     ) throws DBException {
+        if (fileSystem == null) { // when app stops fileSystem is null
+            return new DBNFileSystemRoot[]{};
```
ShadelessFox:
```suggestion
            return new DBNFileSystemRoot[0];
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/fs/DBNFileSystem.java
```diff
@@ -154,6 +154,9 @@ protected DBNFileSystemRoot[] readChildNodes(
         @NotNull DBRProgressMonitor monitor,
         @Nullable DBNFileSystemRoot[] mergeWith
     ) throws DBException {
+        if (fileSystem == null) { // when app stops fileSystem is null
```
ShadelessFox: > // when app stops fileSystem is null

It's kinda assumed since it's set to `null` in `dispose`

labanovichttps: Yes, it is set to `null` when disposing calls, and when the app closes, it calls refresh cloud fileSystem, and it's already `null.`

ShadelessFox (approved):

E1izabeth (approved):

PR #37534, opened 2025-03-13
dbeaver/dbeaver#35447 Add button for multiline properties

### LINKED ISSUE #35447: Bigger field for a comment
No

Please make accessing a comment easier by making its window wider at least.
Here is what it looks like by default.

image

Even expanding this part of UI vertically, the comment window is very narrow. 
It could be helpful if it occupied the whole width of the tab it belongs to.
ObjectId field could be relocated to the first row, wrapping to the new line if there is not enough space.

* Adding the comment DDL to the content of the source code. Somewhere between function source code and privileges.
* A modal popup with a Comment editor

**Additional context**
We document our databases using object comments. Then we generate wiki-like documentation out of it. The current size of comment windows makes it impossible to read even the first sentence of the documentation.

Please note, it applies to all objects, not only functions.

labanovichttps:
image
image


plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/CustomFormEditor.java
```diff
@@ -301,17 +305,48 @@ public void widgetSelected(SelectionEvent e) {
...
+                                editor.getText()
+                            );
+                            if (newValue != null) {
+                                editor.setText(newValue);
+                            }
+                        }
+                    }
+                });
+                GridData gdComposite = new GridData(SWT.FILL, SWT.TOP, true, false);
+                gdComposite.heightHint = textHeight + 6;
```
E1izabeth, line 349: Maybe I would make the button height smaller, but I am not sure how exactly it would look (if it would be better)

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/CustomFormEditor.java
```diff
@@ -301,17 +305,48 @@ public void widgetSelected(SelectionEvent e) {
...
-                editor.setLayoutData(gd);
-                return editor;
+
+                int textHeight = UIUtils.getTextHeight(editor) * 3;
+                GridData gdEditor = new GridData(SWT.FILL, SWT.TOP, true, false);
+                gdEditor.heightHint = textHeight;
+                composite.setLayoutData(gdEditor);
+
+                Button editButton = new Button(composite, SWT.FLAT | SWT.PUSH);
+                editButton.setImage(DBeaverIcons.getImage(UIIcon.DOTS_BUTTON));
```
ShadelessFox, line 323: Why don't we want to use a proper "pencil" icon, let's say?

labanovichttps: image
I think it seems uglier than the dots icon. 
@serge-rider What do you think?

ShadelessFox: Remove the `SWT.FLAT` style.

labanovichttps: @ShadelessFox Think this could be a little bit confusing because the user already could edit text, and that `pencil` said that you can also edit it. 
image

ShadelessFox: For consistency reasons, `...` is nice. But its definition is too vague. It would be nicer to introduce a new icon as a replacement. Okay, let's keep dots for now.

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/CustomFormEditor.java
labanovichttps: image

ShadelessFox (dismissed):

Matvey16:
That makes the comment field uneditable (the field itself can be edited but DBeaver doesn't recognize any changes and save/revert buttons remain inactive)

Also now that field can't be scrolled

https://github.com/user-attachments/assets/b417cf40-5133-478d-a2b9-96bf5d244e7c

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/CustomFormEditor.java
```diff
@@ -301,16 +305,47 @@ public void widgetSelected(SelectionEvent e) {
...
-                gd.heightHint = (UIUtils.getTextHeight(editor) + editor.getBorderWidth()) * 2;
-                editor.setLayoutData(gd);
+
+                int textHeight = UIUtils.getTextHeight(editor) * 3;
+                GridData gdEditor = new GridData(SWT.FILL, SWT.TOP, true, false);
+                gdEditor.heightHint = textHeight;
+                composite.setLayoutData(gdEditor);
+
+                Button editButton = new Button(composite, SWT.FLAT | SWT.PUSH);
+                editButton.setImage(DBeaverIcons.getImage(UIIcon.DOTS_BUTTON));
```
ShadelessFox, line 323: While we are at it, please add a tooltip for this button so its intention is clear without requiring a click.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/CustomFormEditor.java
labanovichttps: fixed

ShadelessFox (dismissed):

E1izabeth (dismissed):

IvanGorshechnikov (approved):

Matvey16:
Verified

PR #37540, opened 2025-03-14
dbeaver/pro#4243 Respect TE server export and copy policies, add copy policy to desktop

E1izabeth:
The new policy for copy data operations is called "policy.data.copy.disabled". Works the same as "policy.data.export.disabled"


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/BasePolicyDataProvider.java
```diff
@@ -29,32 +30,37 @@
...
         // private constructor
     }
 
-    /**
-     * Return boolean value of policy data property
-     *
-     * @param propertyName - property name
-     * @return - boolean value
-     */
+    public static void setInstance(BasePolicyDataProvider instance) {
```
labanovichttps: missing `@NotNull/@Nullable` annotation

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/BasePolicyDataProvider.java
```diff
@@ -29,32 +30,37 @@
...
+public class BasePolicyDataProvider implements DBPPolicyDataProvider {
     private static final Log log = Log.getLog(BasePolicyDataProvider.class);
 
     private static final String DBEAVER_REGISTRY_POLICY_NODE = "Software\\DBeaver Corp\\DBeaver\\policy"; //$NON-NLS-1$
-    private static final BasePolicyDataProvider INSTANCE = new BasePolicyDataProvider();
+
+    public static final String POLICY_DATA_EXPORT = "policy.data.export.disabled"; //$NON-NLS-1$
+    public static final String POLICY_DATA_COPY = "policy.data.copy.disabled"; //$NON-NLS-1$
+
+    private static BasePolicyDataProvider INSTANCE = new BasePolicyDataProvider();
```
labanovichttps: `INSTANCE` -> `instance` cuz not the constant. 
+ Maybe mark `instance` as `volatile` and make `setInstance` synchronized?

E1izabeth: `setInstance` is called only once on a really early app start. I don't see a reason for adding synchronisation.

labanovichttps (dismissed):

PR #37563, opened 2025-03-17
Cb 5875 creating schemas only for pro

HocKu7:


plugins/org.jkiss.dbeaver.model.sql.jdbc/src/org/jkiss/dbeaver/model/sql/db/InternalDB.java
```diff
@@ -122,25 +122,29 @@ protected void initSchema(@NotNull DBRProgressMonitor monitor, @NotNull Connecti
...
+                ),
+                monitor1 -> connection,
+                schemaConfig.versionManager(),
+                dialect,
+                schemaConfig.schemaVersionActual(),
+                schemaConfig.schemaVersionObsolete(),
+                databaseConfig,
+                this
+            );
+            schemaManager.updateSchema(monitor);
```
serge-rider: Let's catch errors and rethrow with additional info (which particular schema update has failed)

serge-rider (dismissed):

alexander-skoblikov (approved):

PR #37586, opened 2025-03-19
dbeaver/pro#3116 Add cas datetime and date formatting for ycql

labanovichttps:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/DateTimeCustomValueHandler.java
```diff
@@ -126,4 +126,12 @@ public void refreshValueHandlerConfiguration(DBSTypedObject type) {
     @NotNull
     protected abstract String getFormatterId(DBSTypedObject column);
 
+    @NotNull
+    public Object getQuotedValue(@NotNull Object value) {
+        String strValue = value.toString();
+        if (!strValue.startsWith("'") && !strValue.endsWith("'")) {
+            value = "'" + strValue + "'";
+        }
+        return value;
```
E1izabeth: Let's use `SQLUtils.quoteString` instead or move this utility method to SQLUtils and implement it using `SQLUtils.quoteString` and `SQLUtils.isStringQuoted`

labanovichttps: There's no datasource or `DBSObject` for a call `SQLUtils.quoteString`

E1izabeth (commented):
I'm unsure it's OK that we have almost the same methods in `LibSqlUtils.quote`, `FFDriverUtils.quote`. Looks like duplication. I would refactor it and put some universal method to CommonUtils.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/DateTimeCustomValueHandler.java
IvanGorshechnikov: You can get a datasource, but to do so, the column object needs to be cast into a CasTableColumn (well, a column should always be a CasTableColumn) that already has a dataSource

IvanGorshechnikov: But I don't know is it a good solution of issue or not

E1izabeth:
Let's move this to a new class called `StringUtils` in `org.jkiss.utils`

PR #37609, opened 2025-03-20
dbeaver/pro#4319 global provider config type

alexander-skoblikov:
closes dbeaver/pro#4319


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/meta/Property.java
```diff
@@ -166,6 +166,13 @@
      */
     String[] features() default {};
 
+    /**
+     * List of required application features to enable property
```
ivgag, line 170:
```suggestion
     * Array of required application features to enable property

```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/PropertyDescriptor.java
```diff
@@ -312,6 +312,12 @@ public String[] getFeatures() {
         return allFeatures;
     }
 
+    @Nullable
+    @Override
+    public String[] getRequiredFeatures() {
+        return null;
```
ivgag, line 318: Would it be better to return an empty array?

ivgag (approved):

PR #37625, opened 2025-03-24
dbeaver/pro#4375 Fix NPE when execution context is missing on autocomplete context init

E1izabeth:
Bug should be fixed


IvanGorshechnikov (dismissed):

kseniaguzeeva:
Could you please fix here too if it's possible: NPE appears after WHERE while schema is loading for the SQL Editor. 
STR:
1. Connect to a relational database and write SELECT * FROM `table_name` WHERE and press Ctrl+space to see the column list
2. Disconnect from the database, connect to it again and press Ctrl+space after WHERE while the database is connection

AR: NPE in the log
ER: the tooltip "no proposals" appears


https://github.com/user-attachments/assets/860ab7e1-69d5-4eba-9ae3-937ab8f33357

[teradata_where.log](https://github.com/user-attachments/files/19445110/teradata_where.log)

kseniaguzeeva:
verified

PR #37626, opened 2025-03-24
CB-5774 add connection brute force protection

yagudin10:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/qm/meta/QMMDataSourceInfo.java
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.qm.meta;
+
+/**
+ * QM model data source info.
+ */
+public interface QMMDataSourceInfo {
```
ShadelessFox, line 25: Nullability annotations please :^)

plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMAdminController.java
```diff
@@ -225,4 +225,6 @@ void deleteObjectPermissions(
     @NotNull
     List<SMTeamMemberInfo> getTeamMembersInfo(@NotNull String teamId) throws DBException;
 
+    void blockUserByBruteForceProtection(@NotNull String userId) throws DBException;
```
alexander-skoblikov: check that it possible to remove this method from public model

HocKu7 (dismissed):

PR #37689, opened 2025-03-31
#35492 Fix script generated for mysql auto_incremet column with pk

### LINKED ISSUE #35492: Adding/removing auto-incrementing int column to MariaDB/MySQL generates bad code
### Description

Adding or removing an auto-incrementing integer column to a table which does not have one via the table view generates code which cannot run without being edited.

### DBeaver Version

24.2.0

### Operating System

macOS

### Database and driver

MariaDB 11

### Steps to reproduce

1. Create a table with a varchar column 
2. Edit the table via the GUI and add a new column like so:
image
3. Click OK. Notice the "Key" column for the new column is not populated
4. Click Save to get the following code:

```
ALTER TABLE mytable ADD id INT auto_increment NOT NULL;
ALTER TABLE mytable ADD CONSTRAINT `primary` PRIMARY KEY (id); 
```

5. Clicking Execute will generate an error:
```
SQL Error [1075] [42000]: (conn=107406) Incorrect table definition; there can be only one auto column and it must be defined as a key
  (conn=107406) Incorrect table definition; there can be only one auto column and it must be defined as a key
```

The correct sql to add such a column is:
```
ALTER TABLE mytable ADD id INT auto_increment NOT NULL PRIMARY KEY
```
6. Once the column is added to the table, delete it via the GUI
7. Answer "Yes" to both questions
8. Click Save to get the following code:
```
ALTER TABLE mytable DROP PRIMARY KEY;
ALTER TABLE mytable DROP COLUMN view_id;
```
9. Clicking Execute will generate an error:

```
SQL Error [1075] [42000]: (conn=107406) Incorrect table definition; there can be only one auto column and it must be defined as a key
  (conn=107406) Incorrect table definition; there can be only one auto column and it must be defined as a key
```

The correct way to drop such a column is to drop the column first, then the primary key:

```
ALTER TABLE mytable DROP COLUMN view_id;
ALTER TABLE mytable DROP PRIMARY KEY;
```

### Additional context

_No response_

E1izabeth:
Now create and delete operations for such columns are executed without error.


plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/edit/MySQLTableColumnManager.java
```diff
@@ -127,6 +120,18 @@ public StringBuilder getNestedDeclaration(@NotNull DBRProgressMonitor monitor, @
         {
             decl.append(" AUTO_INCREMENT"); //$NON-NLS-1$
         }
+        try {
+            if (
+                Optional.ofNullable(command.getObject().getParentObject().getConstraints(monitor))
+                .map(cc -> cc.stream().anyMatch(c -> c instanceof  MySQLTableConstraint tc &&
+                    MySQLConstraintManager.tryGetColumnOfPrimaryKeyConstraintForAutoincrementColumn(monitor, tc, false) == column))
+                .orElse(false)
```
ShadelessFox: I don't know how I feel about this

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/edit/MySQLConstraintManager.java
```diff
@@ -49,6 +52,56 @@ public DBSObjectCache<MySQLCatalog, MySQLTableConstraint> getObjectsCache(MySQLT
...
+                 ) {
+                     return column;
+                 }
+             }
+         }
+        return null;
+    }
+
+    private static boolean checkPersistedState(DBPPersistedObject persistedObject, boolean checkPersisted) {
+        return checkPersisted && persistedObject.isPersisted() || !checkPersisted && !persistedObject.isPersisted();
```
ShadelessFox: Is this the correct logic? If `checkPersisted` is `true`, we ensure that the object is **persisted**. If it's `false`, we ensure the object is **not persisted**.

If so, then it can be simplified:
```suggestion
        return checkPersisted == persistedObject.isPersisted();
```

E1izabeth: The logic behind this expression looks less straightforward for me, but OK

ShadelessFox (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified

PR #37691, opened 2025-04-01
Fix for auto-completion suggestions not working in begin-end blocks

davejinyoung:


arhayka:
It's been a while since there has been an update here. Please let me know if the PR is till actual.

PR #37700, opened 2025-04-02
Cb 6160 ldap user provisioning

HocKu7:
Closes https://github.com/dbeaver/pro/issues/5152


plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMStandardMeta.java
```diff
@@ -21,5 +21,7 @@ public interface SMStandardMeta {
 
     String META_FIRST_NAME = "firstName";
     String META_LAST_NAME = "lastName";
+    String META_USERNAME = "username";
+    String META_USE_USERNAME_FOR_LOGIN = "useUsernameForLogin";
```
alexander-skoblikov: do we really need this meta attr?

alexander-skoblikov (dismissed):

alexander-skoblikov (approved):

PR #37742, opened 2025-04-08
Dbeaver/dbeaver#37393 fix postgre connection open

### LINKED ISSUE #37393: Postresql: doesn't connect while query executing
### Description

Can't open new editor of current connection or editor to another DB while query executes.

### DBeaver Version

Community Edition 25.0.0

### Operating System

Windows 10 x64

### Database and driver

Postresql 15, default JDBC driver

### Steps to reproduce

Connect to first db and execute long time query like `select pg_sleep(10)`
While query runs open new sql editor or current connected db - editor doesn't connect until previous query ends.

Or

Connect to first db and execute long time query like `select pg_sleep(10)`
While query runs open new sql editor for another DB - editor doesn't connect until previous query ends.

Tested at parametr "Open separate connection for each editor": "Default", "Always"

Image

### Additional context

_No response_

labanovichttps:
https://github.com/user-attachments/assets/4331bd20-c2f0-45d2-b34a-0e87aada2142


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIAssistant.java
```diff
@@ -59,4 +59,14 @@ CommandResult command(
...
+
+
+    /**
+     *  Generates the description for DBSObject.
+     */
+    @NotNull
+    String describe(
+        @NotNull DBRProgressMonitor monitor,
+        @NotNull DAICommandRequest request
+    ) throws DBException;
```
E1izabeth: Why is it here?

E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #37745, opened 2025-04-08
Dbeaver/pro#4149 add navigator asc sorting

labanovichttps:
image

XLSX For testing
[testmasha.xlsx](https://github.com/user-attachments/files/19647073/testmasha.xlsx)


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIAssistant.java
```diff
@@ -59,4 +59,14 @@ CommandResult command(
      * Returns whether the AI assistant has a valid configuration.
      */
     boolean hasValidConfiguration() throws DBException;
+
+
+    /**
+     *  Generates the description for DBSObject.
+     */
+    @NotNull
+    String describe(
```
E1izabeth: Please remove irrelevant changes

E1izabeth (changes requested):

E1izabeth (commented):
We sort all nodes twice now under some conditions. Isn't it better to write one comparator?
Also there is a setting `ModelPreferences.NAVIGATOR_SORT_ALPHABETICALLY`. Shouldn't alphabetical sorting depend on it?

labanovichttps:
> We sort all nodes twice now under some conditions. Isn't it better to write one comparator? Also there is a setting `ModelPreferences.NAVIGATOR_SORT_ALPHABETICALLY`. Shouldn't alphabetical sorting depend on it?

Added sorting for `ModelPreferences.NAVIGATOR_SORT_ALPHABETICALLY.` 
I also added a check for only one sorting instead of twice.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNUtils.java
```diff
@@ -48,6 +46,8 @@
 public class DBNUtils {
 
     private static final Log log = Log.getLog(DBNUtils.class);
+    private static final Comparator<DBNNode> ALPHANUMERIC_COMPARATOR_COMPARATOR = (o1, o2) -> AlphanumericComparator.getInstance()
```
E1izabeth: Maybe just `ALPHANUMERIC_COMPARATOR`?
```suggestion
    private static final Comparator<DBNNode> ALPHANUMERIC_COMPARATOR = (o1, o2) -> AlphanumericComparator.getInstance()
```

E1izabeth (commented):
It would be nice to write a couple of unit tests for this logic

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNUtils.java
labanovichttps: fixed

ShadelessFox: `Comparator.comparing(DBNNode::getNodeDisplayName, AlphanumericComparator.getInstance())`

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNUtils.java
```diff
@@ -135,14 +134,17 @@ private static void sortNodes(DBNNode[] children)
                     Arrays.sort(children, NodeFolderComparator.INSTANCE);
                 } else if (prefStore.getBoolean(ModelPreferences.NAVIGATOR_SORT_ALPHABETICALLY) || isMergedEntity(firstChild)) {
                     if (!(firstChild instanceof DBNContainer)) {
-                        Arrays.sort(children, NodeNameComparator.INSTANCE);
+                        Arrays.sort(children, ALPHANUMERIC_COMPARATOR_COMPARATOR);
```
ShadelessFox: `NodeNameComparator` is not used anymore; please delete it

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNUtils.java
```diff
@@ -135,14 +134,17 @@ private static void sortNodes(DBNNode[] children)
...
                     }
                 } else if (prefStore.getBoolean(ModelPreferences.NAVIGATOR_SORT_FOLDERS_FIRST)) {
-                    Arrays.sort(children, NodeFolderComparator.INSTANCE);
+                    if (firstChild instanceof DBSFolder) {
+                        Arrays.sort(children, NodeFolderComparator.INSTANCE);
+                    } else if (!(firstChild instanceof DBNContainer)) {
+                        Arrays.sort(children, ALPHANUMERIC_COMPARATOR_COMPARATOR);
+                    }
```
ShadelessFox: I don't quite understand what this change does. Can you please elaborate?

labanovichttps: For example, opening a connection and seeing Databases/Administrator, etc. It applies NodeFolderComparator.INSTANCE as default, but if you open any schema or table it applies ALPHANUMERIC_COMPARATOR_COMPARATOR for comparing like this 
image

E1izabeth (dismissed):

IvanGorshechnikov (dismissed):

E1izabeth (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified

PR #37748, opened 2025-04-08
dbeaver/dbeaver#36576 Remove warn of parallel execution on separate task

### LINKED ISSUE #36576: Dbeaver run multiple queries at same time within same script
If we have different queries within the same script, if we execute, they will go serial (one by one), but if you want them to run in parallel, you need to open different scripts.

Would it be possible to execute all the queries in parallel? 

**Additional context**
Using postgres connector, I want to insert data into different tables, at the same time, and I want to run them in parallel. e.g:

INSERT INTO tpcds.web_returns 
SELECT * FROM tpcds_ext.foreign_web_returns;

INSERT INTO tpcds.web_sales 
SELECT * FROM tpcds_ext.foreign_web_sales;

IvanGorshechnikov:
Discussed here: https://dbeaver.slack.com/archives/C03RBCQ47UN/p1744016607141239 that we wanna remove warn of confirming massive parallel queries execution when you "Execute queries in separate tabs"


ShadelessFox (changes requested):
You must remove all related constants and localization as well.

ShadelessFox (approved):

E1izabeth (approved):

PR #37794, opened 2025-04-14
dbeaver/pro#5419 Fix connection to new JDBC driver in Clickhouse

IvanGorshechnikov:
### Please check that now is used client timezone and not server

Fixed two problems:
1. Not allowing empty password
image

2. Problem with "use_server_time_zone" and "use_time_zone" 
image


E1izabeth (approved):

ShadelessFox (approved):

uslss:
verified

PR #37805, opened 2025-04-15
dbeaver/pro#4380 Check collation support for Postgres

IvanGorshechnikov:
PostgreSQL support collations since 9.1 https://www.postgresql.org/docs/9.1/release-9-1.html
Redshift is based on 8.0.2

Here, I add a check not to load collation for early versions (< 9.1)


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDatabase.java
```diff
@@ -1137,6 +1142,9 @@ static class CollationCache extends PostgreDatabaseJDBCObjectCache<PostgreCollat
         @Override
         protected JDBCStatement prepareObjectsStatement(@NotNull JDBCSession session, @NotNull PostgreDatabase owner)
             throws SQLException {
+            if (!owner.getDataSource().isSupportsCollation()) {
+                return session.prepareStatement("SELECT 1");
+            }
```
E1izabeth: If we can avoid sending such dummy queries to the database, it's better to avoid it. I don't see the reason why we need it here

serge-rider: Agreed. We mustn't be here is isSupportsCollation == false

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -918,6 +928,10 @@ public boolean isSupportsEnumTable() {
         return supportsEnumTable;
     }
 
+    public boolean isSupportsCollation() {
```
ShadelessFox: Please hide collations in the navigator if they aren't supported

IvanGorshechnikov: Collations don't show in the navigator

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDatabase.java
IvanGorshechnikov: Agree. We shouldn't get into this part at all, there are checks earlier
I removed it

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
ShadelessFox: Remove `@Association` annotation on its getter then

E1izabeth (approved):

ShadelessFox (approved):

PR #37828, opened 2025-04-17
dbeaver/pro#4284 Make the "check for updates on startup" preference on by default

ShadelessFox:
The option for checking for updates was renamed:
image

A new preferences link was added:
java_jdxq9LR3Tk


plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/internal/UIConnectionMessages.properties
```diff
@@ -80,6 +80,7 @@ dialog_driver_manager_message_cant_delete_title = Cannot delete driver
 dialog_driver_manager_message_delete_driver_text = Are you sure you want to delete driver '
 dialog_driver_manager_message_delete_driver_title = Delete driver
 dialog_driver_manager_title = Driver Manager
+dialog_driver_manager_preferences_link = Global options can be changed in <a>{0}</a>
```
E1izabeth: Why do you call it options if it's a preferences? As I recall, we usually don't use "options" for that

IvanGorshechnikov (dismissed):

Matvey16:
`Close` button in driver manager can now move out of the border


https://github.com/user-attachments/assets/37b9738e-ecfe-4c9c-a18f-d8103804d119

Matvey16:
Verified

labanovichttps (approved):

Issue #36763, opened 2025-01-10
Improve highlighting update proces


E1izabeth:
### Description

Serge: it looks more like "incorrectly" than "too recently". Because we update highlighting two times per a single button press.
I press a button and add/delete a character (e.g. edit alias) - then query immediately blinks and highligting for the table name is removed. After a second it is back. Although I didn't make any semantic changes to the query.

### DBeaver Version

DBeaver 24.3.2

PR #37845, opened 2025-04-21
#36763 Fix query highlight blinking during text edits

mikiubo:


E1izabeth (changes requested):
It may cause highlighting to be incorrect on editing. That's why the comment above is written: "drop what was invalidated starting at the current position". Please explain the meaning of the introduced behavior modification to the code taking into account the above comments and the purposes of the method. And why it doesn't break state consistency (script cached script items offsets vs actual editor state).
Also please add the comment explaining the meaning of such behavior to your code.

mikiubo:
> It may cause highlighting to be incorrect on editing. That's why the comment above is written: "drop what was invalidated starting at the current position". Please explain the meaning of the introduced behavior modification to the code taking into account the above comments and the purposes of the method. And why it doesn't break state consistency (script cached script items offsets vs actual editor state). Also please add the comment explaining the meaning of such behavior to your code.

I’m not sure if you’ve already considered a possible solution for this: #36763.
From what I understand, you currently remove the highlight and then re-add it in SQLBackgroundParsingJob.

Initially, I tried implementing the TODO at line 202, but it seems that changing only the "current" item causes inconsistencies in the highlighting of the following items.

So I thought of this solution: instead of removing the "current" item (since we know it will be refreshed), we only invalidate the following items.
Let me know if you’re aware of any tests I could run to reproduce the inconsistency,
or if I misunderstood any important aspect of the behavior,
or if you have an alternative approach in mind.

Otherwise, if this approach looks correct and clear to you, I’ll go ahead and add the comment you requested.

Thanks!

E1izabeth:
We've tested this fix. It leads to incorrect coloring when editing, while blinking is still present. I'm not sure it's better than what we have, as it leads to an inconsistency we aimed to avoid.

https://github.com/user-attachments/assets/7b996505-9876-4911-9ad7-add2d9231b4c

The correct fix is to keep letters coloring consistent with their positions.

mikiubo:
yes you are right i think i will need a different approach i will close this pull request and try another way

PR #37851, opened 2025-04-21
dbeaver/dbeaver#37685 Add counter of children nodes in DatabaseNavigtorLabelProvider

### LINKED ISSUE #37685: Show total number of elements beside every object.
When I connect to a database it shows the list of objects inside,
Example:
  Tables
  Views,
  .
  .
  .
  Data Types
and I want to know the total number of elements inside of every object, like this:
  Tables (50)
  Views (8)
Using DBeaver 25.0.1 with W10 and Firebird 5 database.
Thanks.

isai-salazar-enc:
Add an item counter for each node in the navigation tree that has children

Screenshot 2025-04-21 at 10 04 53 a m


ShadelessFox:
I believe the counter should only appear on folder nodes. What do you think?

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/DatabaseNavigatorLabelProvider.java
```diff
@@ -120,6 +133,11 @@ public String getText(Object obj) {
         return text;
     }
 
+    private boolean isRelevantNodeType(DBNNode node) {
+        String nodeType = node.getNodeType();
+        return !"index".equalsIgnoreCase(nodeType);
```
ShadelessFox: What's "index"? Why is it hardcoded?

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/DatabaseNavigatorLabelProvider.java
```diff
@@ -102,6 +106,15 @@ public String getText(Object obj) {
             text = labelProvider.getText(obj);
         } else if (obj instanceof DBNNode dbnNode) {
             text = dbnNode.getNodeDisplayName();
+            if (dbnNode.hasChildren(false) && isRelevantNodeType(dbnNode)) {
+                int childCount = 0;
+                try {
+                    childCount = dbnNode.getChildren(new VoidProgressMonitor()).length;
+                } catch (DBException e) {
+                    log.error("Error fetching children for node: " + dbnNode.getNodeDisplayName());
+                }
```
ShadelessFox: This will cause the UI to freeze, as it will load all children upon expanding a node. I suggest showing a counter only for expanded nodes.

ShadelessFox (changes requested):

ShadelessFox:
This should also be configurable in the preferences. And, perhaps, use a gray font instead. See `org.jkiss.dbeaver.ui.navigator.database.StatisticsNavigatorNodeRenderer#drawObjectDescription`

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/DatabaseNavigatorLabelProvider.java
isai-salazar: It was part of a quick filter, but will be removed for the folder-only approach

isai-salazar-enc: It was a minor filter, will be removed for the folder-only approach

isai-salazar-enc:
The requested changes have been implemented:

- Show children only for expanded folders
- Gray font for counter
- Configuration in preferences

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/StatisticsNavigatorNodeRenderer.java
```diff
@@ -192,6 +200,17 @@ private void drawObjectDescription(@NotNull GC gc, @NotNull DBNDatabaseNode node
         }
     }
 
+    private void drawObjectChildrenCounter(@NotNull GC gc, @NotNull DBNDatabaseNode node, @NotNull Rectangle bounds) {
+        int childCount = 0;
+        try {
+            childCount = node.getChildren(new VoidProgressMonitor()).length;
```
ShadelessFox:
```suggestion
            childCount = node.getChildren(new LocalCacheProgressMonitor(new VoidProgressMonitor())).length;
```
So it's guaranteed that no expensive reads will be performed

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/StatisticsNavigatorNodeRenderer.java
```diff
@@ -192,6 +200,17 @@ private void drawObjectDescription(@NotNull GC gc, @NotNull DBNDatabaseNode node
...
+        } catch (DBException e) {
+            log.error("Error fetching children for node: " + node.getNodeDisplayName());
```
ShadelessFox: Logging could be omitted - otherwise, in the case of an error, log files will be flooded with the same errors. Moreover, you don't log the exception, so its value is worthless.

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/StatisticsNavigatorNodeRenderer.java
```diff
@@ -118,6 +120,12 @@ public void paintNodeDetails(DBNNode node, Tree tree, GC gc, Event event) {
             if (store.getBoolean(NavigatorPreferences.NAVIGATOR_SHOW_STATISTICS_INFO)) {
                 drawObjectStatistics(gc, databaseNode, item, event);
             }
+            if (node instanceof DBNDatabaseFolder && store.getBoolean(NavigatorPreferences.NAVIGATOR_SHOW_CHILD_COUNT)) {
+                TreeItem treeItem = (TreeItem) event.item;
+                if(treeItem.getExpanded()) {
```
ShadelessFox: I suggest using `DBNDatabaseNode#needsInitialization` instead

E1izabeth:
```suggestion
                if (treeItem.getExpanded()) {
```

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/database/StatisticsNavigatorNodeRenderer.java
```diff
@@ -192,6 +197,17 @@ private void drawObjectDescription(@NotNull GC gc, @NotNull DBNDatabaseNode node
         }
     }
 
+    private void drawObjectChildrenCounter(@NotNull GC gc, @NotNull DBNDatabaseNode node, @NotNull Rectangle bounds) {
+        int childCount = 0;
+        try {
+            childCount = node.getChildren(new LocalCacheProgressMonitor(new VoidProgressMonitor())).length;
+        } catch (DBException e) {
+            // Ignore
```
ShadelessFox: No need to show `(0)` if we can't for some reason read the children
```suggestion
            return;
```

ShadelessFox (approved):

E1izabeth (approved):

Matvey16:
Verified

PR #37858, opened 2025-04-22
#37727 Omit catalogs

### LINKED ISSUE #37727: All schemas appear under each schema in ClickHouse connection with latest driver
### Description

Hi DBeaver team,

After upgrading to DBeaver 25.0.1 and the ClickHouse driver 0.8.2, we're encountering an issue where each schema (database) in a ClickHouse connection displays all schemas and tables from every other schema. Effectively, it looks like the entire database structure is being duplicated under each schema node in the Database Navigator.

Environment
DBeaver: 25.0.1
ClickHouse JDBC Driver: 0.8.2
ClickHouse Server: 25.3.2.39

This was not the case before updating DBeaver and the driver.
We confirmed it occurs on all ClickHouse connections, not just specific ones.

Screenshots

Image
(Taken from older issue: https://github.com/dbeaver/dbeaver/issues/36719)

Workarounds Tried
✅ Navigator > Omit catalogs: This hides the duplicated structure and shows the expected schema-table hierarchy.
✅ Reverting to the ClickHouse 0.7.1 driver: Structure appears correct.

❌ No fix found using current default driver settings.

Expected Behavior
Each schema node in the navigator should show only the tables belonging to that schema, not tables from all other databases.

Notes
From what we've gathered, this may be related to the newer JDBC driver behavior where databaseTerm=catalog is now default.
It seems DBeaver is mapping both catalogs and schemas, leading to duplication. Based on other reports, this appears to be unintended.

Would appreciate any confirmation or ETA on a fix in a future version, since we need the newer version of the driver for JSON support.

Thanks a lot for your hard work!

### DBeaver Version

25.0.1

### Operating System

Windows 11

### Database and driver

1. ClickHouse - 25.3.2.39
2. ClickHouse JDBC Driver - 0.8.2

### Steps to reproduce

1. Open DBeaver 25.0.1.
2. Create or open a ClickHouse connection using the default ClickHouse JDBC driver 0.8.2 and in ClickHouse new version (25.3.2.39).
3. Expand the connection in the Database Navigator panel.
4. Observe the list of schemas (i.e., ClickHouse databases).
5. Expand any one of the schemas.
6. You will see all schemas duplicated under it — each containing all tables from every schema.

### Additional context

_No response_

ShadelessFox:
As seen in the legacy driver


IvanGorshechnikov (approved):

uslss:
verified

PR #37894, opened 2025-04-28
dbeaver/pro#3162 Implement a new semantic resolution algorithm to allow correct CTE handling

E1izabeth:
Implement a new semantic resolution algorithm based on the separation of processing stages and eliminate excessive dependencies to allow correct CTE handling.

This new algorithm is applied only for CTE, so autocompletion for other query types should not change! (May not test it so hard)

image

```
CREATE TABLE BinaryTree(
  ID INT,
  `left` INT NULL,
  `right` INT NULL,
  value VARCHAR(8) NULL
);

INSERT INTO BinaryTree
VALUES (1, 4, 3, NULL),
       (2, 1, NULL, 'A'),
       (3, 7, NULL, 'B'),
       (4, NULL, NULL, 'C'),
       (5, NULL, NULL, 'D'),
       (6, 8, 9, 'E'),
       (7, NULL, 5, 'F'),
       (8, NULL, NULL, 'G'),
       (9, NULL, NULL, 'H');


WITH RECURSIVE Descendents(ID, l, r, y) AS (
  SELECT ID, `left` as l, `right` as r, `value` as x FROM BinaryTree WHERE ID = 2      
  UNION ALL                                                       
  SELECT b.* FROM BinaryTree b, Descendents d  
  WHERE b.ID = d.l OR b.ID = d.r              
)       
SELECT ID, y  FROM Descendents
WHERE ID <> 2;
```

Another example:
```
with recursive all_data as (
    select *
    from (values(1, 1, 10),(1, 2, 3),(1, 3, 7),(1, 4, 1),(1, 5, 2),(2, 1, 6),(2, 2, 4),(2, 3, 5),(2, 4, 8),(2, 5, 1)) as s(gr, id, quantity)
),
ct AS (
    select *, c.quantity AS sm
    from all_data AS c
    WHERE id = 1
    UNION ALL
    select c2.*, c.sm + c2.quantity AS sm
    from ct AS c
    JOIN all_data AS c2 ON c2.id = c.id + 1 AND c.gr = c2.gr
)
SELECT *
FROM ct
ORDER BY gr, id
;
```


serge-rider (approved):

ShadelessFox (approved):
I have no idea what's going on, but I trust you :^)

uslss:
verified

PR #37898, opened 2025-04-29
dbeaver/pro#5579 Gtk separator size warning fix

ShadelessFox:
This PR resizes separator icons to be 6x20 and 20x6, one extra pixel. It seems like either Gtk or SWT has a minimal size for separators.


E1izabeth (approved):

Matvey16:
Still getting the error
image

ShadelessFox:
Guess it's specific to a particular DE. Closing the PR then.

Issue #36977, opened 2025-01-31
Some keywords are not completed properly


MashaKorax:
### Description

DESK, ORDER, LIMIT
`SELECT * from  Album order by Album.AlbumId dDESC`

https://github.com/user-attachments/assets/2e954413-9307-4617-9a24-a0680986cef2

### DBeaver Version

24.3.5

### Operating System

win 10

PR #37910, opened 2025-04-30
#37476 #37746 Fix columns suggestion at the end of the query when the part of the column name is already typed

### LINKED ISSUE #37476: Autocomplete glitches
### Description

I would like to report two annoying glitches, introduced recently (within last months).

Autocomplete enters fully qualified column name, in form schema.table.column, instead of column name alone. 

https://youtu.be/NeqlmlZZ4Kc

Autocomplete sometimes offers function names instead of columns. While I can imagine using functions in conditions, mostly those are not on the left side of equation. Anyway, let columns be listed first, then functions.

https://youtu.be/wFCKtArQWcI

Here are my settings:

Image

Note, that after disabling Use Global Search and Show Stored Procedures in column list, the DBeaver still offers functions instead of columns. 

Image

### DBeaver Version

CE 25.0.0

### Operating System

Fedora 41

### Database and driver

Postgres

### Steps to reproduce

See video in description

### Additional context

_No response_

### LINKED ISSUE #37746: Autocompletion for columns is not working as expected
### Description

Autocompletion for columns is not working as expected. Column names are suggested only when you start typing it and then delete. Instead of it I see functions.

### DBeaver Version

25.0.1

### Operating System

macOS 15.4

### Database and driver

com.mysql:mysql-connector-j:RELEASE [8.2.0]

### Steps to reproduce

1. Type any SELECT in format from the screen 
2. Try to type column name
3. Delete column name

### Additional context

Image
Image

E1izabeth:
Caused by #36977
Please check that #37476 #37746 #36977 are all fine.


IvanGorshechnikov (approved):

ShadelessFox (approved):

uslss:
verified

PR #37916, opened 2025-04-30
dbeaver/pro#5565 AI chat improvements

ShadelessFox:


E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/AIHttpUtils.java
```diff
@@ -29,15 +29,11 @@ private AIHttpUtils() {
...
-    public static URI resolve(String base, String... paths) throws DBException {
-        try {
-            URI uri = new URI(base);
-            for (String path : paths) {
-                uri = uri.resolve(path);
-            }
-            return uri;
-        } catch (URISyntaxException e) {
-            throw new DBException("Incorrect URI", e);
+    public static URI resolve(String base, String... paths) {
```
serge-rider: Why? 
Suppressing broken URI verification may break a lot of things.

ShadelessFox: It does not suppress anything. The old code is just as faulty as this one. Passing invalid URIs to our own APIs is a bug. Moreover, all URIs are hardcoded and should ideally be initialized in place.

PR #37924, opened 2025-05-02
#37895 Fix regex matching remove definer clause for mysql dump option

### LINKED ISSUE #37895: Remove Definer is not always working.
### Description

hi, when i dump a db, with 'remove definer' checked, it removes many (160 / 164ish)  but leaves in 4 instances.  Versions 25,  24.1.3 
the remaining look like DEFINER=`username`@`[66.44.111.0/255.255.224.0`] It doesnt remove the definers with the submask 255.255...

### DBeaver Version

25, 24

### Operating System

win 10

### Database and driver

com.mysql.cj.jdbc.Driver  mysql 8

### Steps to reproduce

dump db with 'remove definer' checked (in a task)
dump still has some definers

### Additional context

_No response_

E1izabeth:
Can test on MySQL database on db1, dump db `sakila`.
image

The `films_count` procedure should not have such a clause in the dump file.

This procedure was created with such DDL:
```
CREATE DEFINER = `root`@`[66.44.111.0/255.255.224.0]` PROCEDURE films_count()
BEGIN
  SELECT 'Number of films:', COUNT(*) FROM sakila.film;
END;
```


IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/tasks/MySQLDatabaseExportHandler.java
```diff
@@ -172,7 +172,7 @@ protected void startProcessHandler(
 
 
     static class DumpFilterJob extends DumpJob {
-        private final Pattern DEFINER_PATTER = Pattern.compile("DEFINER\\s*=\\s*`[^*]*`@`[0-9a-z\\-_\\.%]*`", Pattern.CASE_INSENSITIVE);
+        private final Pattern DEFINER_PATTER = Pattern.compile("DEFINER\\s*=\\s*`[^*]*`@`([0-9a-z\\-_\\.%]*|\\[[0-9a-z\\-_.%/]*])`", Pattern.CASE_INSENSITIVE);
```
ShadelessFox: I wonder if we actually care about the contents of the definer. Could just look for `` DEFINER\s*=\s*`.*?`@`.*?` ``, no?

E1izabeth: Actually, you're right. We don't care what's inside. I'll try

ShadelessFox (approved):

Matvey16:
Verified

PR #37931, opened 2025-05-02
dbeaver/pro#5596 Add force timeout for JSqlParser in SQLSemanticProcessor to make sure we'll handle if it hangs

### LINKED ISSUE #37028: DBeaver hangs if fetching bigger json
### Description

Starting with version 24.3.4 there is an issue with bigger scripts in a PostgreSQL db. If executed dbeaver hangs and i see the sandglass.
Earlier version

### DBeaver Version

24.3.4

### Operating System

MacOs

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Nothing should change, just the parser won't freeze endlessly on execution of the queries proposed as an example here #37028

For QA:
- check some basic features like query execution, vqb, panels
- check if it resolves #37028 (may not resolve, but need to check) _**- checked, no need to recheck with latest changes**_


IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/parser/SQLSemanticProcessor.java
```diff
@@ -62,12 +62,16 @@ public class SQLSemanticProcessor {
...
 
     private static final boolean ALLOW_COMPLEX_PARSING = false;
+    private static final int PARSE_FUTURE_TIMEOUT_MS = 10000; // 8000 is the default timeout for parser itself
 
     public static Statement parseQuery(@Nullable SQLDialect dialect, @NotNull String sql) throws DBCException {
+
+
         String sqlWithoutComments = dialect == null ? sql : SQLUtils.stripComments(dialect, sql);
-        CCJSqlParser parser = new CCJSqlParser(new StringProvider(sqlWithoutComments));
+        ExecutorService executorService =  Executors.newSingleThreadExecutor();
```
ShadelessFox: You're not terminating the executor. Judging by the docs, it will be disposed of once it becomes phantom reachable, but we can do that ourselves here. Moreover, we have a suitable `try` block just a line below :^)

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/parser/SQLSemanticProcessor.java
```diff
@@ -77,9 +81,20 @@ public static Statement parseQuery(@Nullable SQLDialect dialect, @NotNull String
...
+            try {
+                return future.get(PARSE_FUTURE_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            } catch (TimeoutException | InterruptedException interruptedEx) {
+                parser.interrupted = true;
+                future.cancel(true);
+                throw new DBCException("Failed to parse SQL query within reasonable time ", interruptedEx);
+            }
         } catch (Exception e) {
             throw new DBCException("Error parsing SQL query: " + e.getMessage(), e);
+        } finally {
```
ShadelessFox: Okay, you shut it down here.

ShadelessFox (dismissed):

Matvey16:
Verified

Giant json from the mentioned ticket can be executed without hanging

ShadelessFox (approved):

Matvey16:
Verified

PR #37933, opened 2025-05-02
dbeaver/pro#3161 Allow PostgreSQL identifiers start from underscore

### LINKED ISSUE #22079: PostgreSQL - Double quoted column name
When using autocompletion with at least PostgreSQL column names that start with a double underscore like __column are automatically quoted by DBeaver.

So I get something like:

`select "__column" from table`

Can I disable this behavior somewhere?

E1izabeth:
No quotes anymore
image

### For QA:
Fixes https://github.com/dbeaver/dbeaver/issues/22079 and https://github.com/dbeaver/pro/issues/3161
Check both


IvanGorshechnikov (approved):

ShadelessFox (approved):

uslss:
verified

PR #37946, opened 2025-05-05
#37939 Fix `create table` statement parsing

### LINKED ISSUE #37939: [Column definition in PostgreSQL] GENERATED not recognised as keyword
In PostgreSQL, the keyword GENERATED in identity primary key is not recognised as a keyword but raise an error as column not found.

Example of valid SQL:
```sql
CREATE TABLE IF NOT EXISTS mytable (
  id integer GENERATED ALWAYS AS IDENTITY PRIMARY KEY
)
```

It should be recognised as a keyword.

E1izabeth:
image


IvanGorshechnikov (approved):

ShadelessFox (approved):

uslss:
verified

PR #37965, opened 2025-05-06
dbeaver/pro#4343 Fix query to fetch PostgreSQL Foreign servers

E1izabeth:
Nothing should change from the user's perspective. We just changed the query that is sent to the database to fetch this info


ShadelessFox (approved):

uslss:
verified

PR #37967, opened 2025-05-06
dbeaver/pro#5587 Refactor and remove errors from logs

labanovichttps:


Matvey16:
Select is still deformed. Somehow the first part of it is in the end

image

Matvey16:
Redundant command in keys preferences

dbeaver_ztFwmgFy6v

PR #37987, opened 2025-05-08
dbeaver/pro#5541 Prohibit switch workspace with EAP license

IvanGorshechnikov:
If the user tries to run DBeaver with a non-default workspace with an EAP license, on startup, it shuts down and sees an  error message in the log "Workspace switching is not allowed for EAP"


plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -194,6 +194,16 @@ public Object start(IApplicationContext context) {
             return IApplication.EXIT_OK;
         }
 
+        try {
+            if (!isWorkspaceSwitchingAllowed() && !WORKSPACE_DIR_CURRENT.equals(RuntimeUtils.getLocalFileFromURL(instanceLoc.getURL())
+                .getAbsolutePath())) {
+                System.err.println("Workspace switching is not allowed for EAP. Exit " + GeneralUtils.getProductName() + ".");
```
ShadelessFox:
```suggestion
                System.err.println("Workspace switching is not allowed when participating in the early access program. Exiting " + GeneralUtils.getProductName() + ".");
```

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -194,6 +194,16 @@ public Object start(IApplicationContext context) {
...
+                return IApplication.EXIT_OK;
+            }
+        } catch (IOException e) {
```
ShadelessFox, line 204: Maybe log the exception at least?

ShadelessFox (approved):

E1izabeth (approved):

Matvey16:
Verified

PR #37994, opened 2025-05-08
#37353 Make multiline fields resizable

### LINKED ISSUE #37353: The comment field does not increase in size when the upper area of the function editing window is stretched
### Description

The comment field does not increase in size when the upper area of the function editing window is stretched

### DBeaver Version

CE 24.3.5.202502162012

### Operating System

Win 10 x64

### Database and driver

Postgresql

### Steps to reproduce

Image

### Additional context

_No response_

ShadelessFox:
https://github.com/user-attachments/assets/c5b00a58-3aa8-47e4-b7ed-503edd35489d


serge-rider:
Looks pretty cool. Need to test on MacOS and on Dark theme

Matvey16:
Verified

PR #38022, opened 2025-05-13
dbeaver/pro#5674 Fix CTE semantic recognition and validation

E1izabeth:
You can test the following query on our PostgreSQL 12 on db3, schema stats.

image


uslss:
verified

PR #38027, opened 2025-05-13
Dbeaver/pro#5545 apache poi version bump

labanovichttps:


serge-rider:
1. The must be a PR in https://github.com/dbeaver/eclipse-bundle-apache-poi
2. It doesn't make sense to just change a version of our artifact (this version doesn't exists yet).

PR #38042, opened 2025-05-14
dbeaver/dbeaver#37948 Fixed JDBC-collection boxed primitive arrays data type

### LINKED ISSUE #37948: Numeric array values are mistakenly quoted in the ClickHouse metadata
### Description

After update the driver version from 0.6.1 to 0.8.5 in results of export data or SQL generation of Int arrays, the array values with type of Int are erroneously quoted.
Generate Insert example:
Expected
```sql
INSERT INTO `default`.TEST
(Column1)
VALUES([235346, 0, 1, 22, 45748]);
```
 Actual
```sql
INSERT INTO `default`.TEST
(Column1)
VALUES(['235346','0','1','22','45748']);
```

### DBeaver Version

25.0.4

### Operating System

win10

### Database and driver

ClickHouse 0.8.5

### Steps to reproduce

1. create a table with an Int array column
```
CREATE TABLE default.TEST
(
    Column1 Array(Int64)
)
ENGINE = MergeTree
ORDER BY Column1
SETTINGS index_granularity = 8192;
```
2. Insert data  
```
INSERT INTO default.TEST
(Column1)
VALUES( [235346,0,1,22,45748]);
```
3. on the table's context menu select Generate SQL -> INSERT

### Additional context

_No response_

Nexus6v2:
Clickhouse supports nullable arrays, therefore arrays are always fetched by jdbc using boxed `Object[]` and not the primitive arrays.
It caused the `JDBCCollection`'s `dataType` and `dataKind` to be set to default `VARCHAR` and resulted in:
1. incorrect type representation in UI
2. incorrect SQL INSERT query generation (original ticket https://github.com/dbeaver/dbeaver/issues/37948).


IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/data/JDBCCollection.java
```diff
@@ -455,6 +455,17 @@ private static JDBCCollection makeCollectionFromJavaArray(
         return new JDBCCollection(session.getProgressMonitor(), elementType, elementValueHandler, contents);
     }
 
+    private static boolean isBoxedObjectArray(Object array, Class<?> type) {
```
labanovichttps: Add `@NotNull/@Nullable` annotations, and **a** as variable name is unclear name

ShadelessFox (commented):
Don't we also want to check for `instanceof` for boxed types, too? Like `array instanceof Integer[]`. And please, specify nullability annotations.

Nexus6v2:
@ShadelessFox 
> Don't we also want to check for instanceof for boxed types, too? Like array instanceof Integer[].

It looks to me that that some drivers might return boxed arrays like `Integer[]`.
This solution will cover such cases.

ShadelessFox (approved):

E1izabeth (approved):

DariaMarkaryan:
Nested Int arrays are still quoted in metadata/sql generation.
Test data:
```sql
CREATE TABLE default.test_nested_array
(
    id UInt64,
    texts Array(String),
    numbers Array(Float64),
    nested_arrays Array(Array(UInt8))
)
ENGINE = MergeTree
ORDER BY id
SETTINGS index_granularity = 8192;
INSERT INTO default.test_nested_array
(id, texts, numbers, nested_arrays)
VALUES(1, ['hello','world'], [3.14,2.71], [[1, 2],[3, 4, 5]]);
INSERT INTO default.test_nested_array
(id, texts, numbers, nested_arrays)
VALUES(2, ['ClickHouse','is','cool'], [1.0,2.0,3.0], [[10],[20, 30]]);
```
Result INSERT query:

Image

Nexus6v2:
@DariaMarkaryan I created new [ticket](https://github.com/dbeaver/dbeaver/issues/38193) to cover nested arrays' issue.

PR #38068, opened 2025-05-15
dbeaver/dbeaver#37968 Extended wildcard operators filtering support

### LINKED ISSUE #37968: ILIKE filtering in Postgres should add % around text
ILIKE filter in the data grid should add % % around the filter value, like the LIKE one.

Image

Nexus6v2:


E1izabeth (approved):

Matvey16:
Verified

PR #38076, opened 2025-05-15
dbeaver/dbeaver#37938 Added quoted-null-values csv export option

### LINKED ISSUE #37938: Export in CSV File null columns in quotes
I have a question, sorry for the wrong path to open request.

I'm Brazilian and need export null columns in a CSV file, but the null values need be in quotes.

Example: "RODRIGO";"27";"MALE";"";"";"";"BRAZIL";

;""; = null columns

How can i configure export menu to do this?

Sorry again, and thank you for help.

Image

Nexus6v2:
1. "All but nulls" option has been removed because it duplicated the "All" option.
2. A new option, "All including nulls," has been added. When selected, quotes are applied to all values, including nulls.


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterCSV.java
```diff
@@ -245,14 +245,13 @@ public void exportRow(DBCSession session, DBCResultSet resultSet, Object[] row)
                     if (!(row[i] instanceof Number)) {
                         quote = true;
                     }
-                } else if (quoteStrategy == QuoteStrategy.ALL_BUT_NULLS) {
-                    if (!DBUtils.isNullValue(row[i])) {
-                        quote = true;
-                    }
                 }
+
```
IvanGorshechnikov, line 249: Does it quote not-null values? I think for ALL_INCLUDING_NULLS strategy for not-null values quote would be false

Nexus6v2: Yes, it does

I checked all cases of it, + with NULL strings, works as expected

Matvey16:
Verified

PR #38084, opened 2025-05-15
dbeaver/pro#5484 AI for data transfer

ivgag:


plugins/org.jkiss.dbeaver.model.ai/META-INF/MANIFEST.MF
```diff
@@ -5,6 +5,7 @@ Bundle-Activator: org.jkiss.dbeaver.model.ai.AIActivator
 Bundle-Name: DBeaver AI Model
 Bundle-SymbolicName: org.jkiss.dbeaver.model.ai;singleton:=true
 Bundle-Version: 1.0.59.qualifier
+Bundle-ClassPath: lib/jsonschema-generator-4.38.0.jar,.
```
serge-rider: Do we really need it? Maybe it is better to move it to p2 dependency?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/openai/OpenAICompletionRequest.java
```diff
@@ -0,0 +1,167 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.theokanning.openai.completion.chat.ChatCompletionRequest;
+import com.theokanning.openai.completion.chat.ChatMessage;
+
+import java.util.List;
+import java.util.Map;
+
+public record OpenAICompletionRequest(
```
serge-rider: It is not used anywhere

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.ai/META-INF/MANIFEST.MF
ivgag: We don’t need it right now since we’re already sending an example of the expected JSON. However, using JSON schemas would be better in the long run — AI tends to follow them more reliably.

Anyway, deleted.

plugins/org.jkiss.dbeaver.model.ai/lib/jsonschema-generator-4.38.0.jar
ShadelessFox, line 1: Why is it here? Seems to be unused

plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/registry/DataTransferNodeConfiguratorDescriptor.java
```diff
@@ -47,9 +46,34 @@ public class DataTransferNodeConfiguratorDescriptor extends AbstractDescriptor
...
-            pageTypes.add(new DataTransferPageDescriptor(pageConfig));
+        // Own descriptors by their id
+        Map<String, DataTransferPageDescriptor> pagesById = new HashMap<>();
+        // Remember all replacement pairs to apply in a second pass
+        List<Map.Entry<String, DataTransferPageDescriptor>> replacements = new ArrayList<>();
+
+        for (IConfigurationElement pageCfg : ArrayUtils.safeArray(config.getChildren("page"))) {
+            String id = pageCfg.getAttribute("id");
+            if (id == null || id.isEmpty()) {
+                continue;                       // skip broken config
```
ShadelessFox: It would be nice to report such configs so they can be tracked down

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AITextUtils.java
```diff
@@ -46,6 +48,32 @@ private AITextUtils() {
...
+            if (code.endsWith(";")) {
+                code = code.substring(0, code.length() - 1);
+            }
+            return code;
+        }
+
+        return "";
+    }
+
+
```
ShadelessFox, line 76: Unused code

ivgag: it's used in `com.dbeaver.model.ai.impl.AIAssistantProImpl#resolveTargetType`

ShadelessFox (approved):

PR #38100, opened 2025-05-17
dbeaver/pro#5736 Change file extensions order

serge-rider:
https://dbeaver.com/wp-admin/post.php?post=507701&action=edit


Matvey16:
Verified

PR #38123, opened 2025-05-20
Broken autoincrement altering for mariadb

vkirst:


vkirst:
Non-proprietary creation.

PR #38161, opened 2025-05-22
#37951 Try fixing connection type coloring for data editor, entity editor and sql editor

### LINKED ISSUE #37951: Connection color no longer applied consistently in UI after update
### Description

In previous versions of DBeaver, the connection color (e.g., red for production, green for development) was applied across the entire interface, making it easy to visually distinguish which database connection was active. This was extremely helpful when working with multiple environments.

After the recent update, only a small part of the UI is now colored (e.g., just the tab bar), which makes it difficult to identify which connection is being used. This change introduces a risk of running queries on the wrong database (e.g., production instead of development).

### DBeaver Version

25.0.4

### Operating System

Windows 11

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
It should become more or less OK, but we can't reach the same result as before because Eclipse now overrides the SCC classes of the controls. So we set a class, then some UI event happens (focus, mouse move, any), and Eclipse just reapplies the class that it wants, overriding the CSS class that we set on the control creation.

image
image
image


Matvey16:
Verified

PR #38164, opened 2025-05-22
dbeaver/dbeaver#18874 Fix slashes in postgres

### LINKED ISSUE #18874: bug if db name includes slashes
### Description

Connection is to Postgres, DB name is 'some/db/name'. When I click on 'Schemas' I get the error like this 'SQL Error [99999]: Unable to parse URL jdbc:postgresql://localhost:5432/some/db/name'. 
The slashes are the problem because when I rename the DB works fine

### DBeaver Version

Community Edition Version 22.3.3.202301211053

### Operating System

Ubuntu 20.4

### Database and driver

PostgreSQL 14.4

### Steps to reproduce

_No response_

### Additional context

_No response_

labanovichttps:
image

STR: create a database with a name like qweq/qwe/qwe

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreDataSourceProvider.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreDataSourceProvider.java
@@ -40,6 +40,8 @@ import org.jkiss.utils.IOUtils;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.InputStreamReader;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.*;
 
 public class PostgreDataSourceProvider extends JDBCDataSourceProvider implements DBPNativeClientLocationManager {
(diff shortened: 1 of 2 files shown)


E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
```diff
@@ -1031,6 +1031,9 @@ public static String getDatabaseNameFromConfiguration(DBPConnectionConfiguration
                 activeDatabaseName = url.substring(divPos + 1, lastPos);
             }
         }
+        if (activeDatabaseName != null) {
```
IvanGorshechnikov: Use CommonUtils.isNotEmpty

IvanGorshechnikov (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreDataSourceProvider.java
```diff
@@ -72,6 +72,8 @@ public long getFeatures() {
     @Override
     public String getConnectionURL(DBPDriver driver, DBPConnectionConfiguration connectionInfo) {
         DBAAuthModel<?> authModel = connectionInfo.getAuthModel();
+        String databaseName = connectionInfo.getDatabaseName();
+        connectionInfo.setDatabaseName(databaseName.replace("/", "%2F"));
```
serge-rider: You cannot modify connection info in getter. 
It may lead to side effects. And you don't need to change it actually.
Just use fixed name for URL build

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
serge-rider: Weird. You should keep database name in the configuration as-is. PErform transforms only when passing it to the driver.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreDataSourceProvider.java
```diff
@@ -71,41 +71,55 @@ public long getFeatures() {
...
-//            if (CommonUtils.toBoolean(connectionInfo.getProperty(PostgreConstants.PROP_SSL_NON_VALIDATING))) {
-//                url.append("&sslfactory=org.postgresql.ssl.NonValidatingFactory");
-//            }
-//        }
+        //        if (CommonUtils.toBoolean(connectionInfo.getProperty(PostgreConstants.PROP_USE_SSL))) {
+        //            url.append("?ssl=true");
+        //            if (CommonUtils.toBoolean(connectionInfo.getProperty(PostgreConstants.PROP_SSL_NON_VALIDATING))) {
+        //                url.append("&sslfactory=org.postgresql.ssl.NonValidatingFactory");
+        //            }
+        //        }
```
ShadelessFox: Commented-out code

uslss:
verified

PR #38171, opened 2025-05-22
#37764 Fix completion autoactivation vs parsing job synchronization issues

### LINKED ISSUE #37764: "Analyzing query for proposals" stuck at 50%
### Description

This is essentially the same issue described in #37608 and discussed in #37612, but I'm opening a new ticket here because the problem persists and I'd like to provide additional information, namely the following:

1. "Long time" in my case means _10 minutes plus_ in some instances. Maybe it's possible to get it unstuck by performing some other action in the editing area, but it may well be that it's actually stuck permanently unless the "Analyzing" job is cancelled through GUI.
2. While this "Analyzing" job is running, the mouse cursor turns to hourglass when moved outside of the editing area. When this happens several times in a row, it's _very annoying_, and I'd argue that this particular point is the real issue in practice.
3. I haven't seen this happen in 24.x and earlier versions.
4. Unlike the original reporter, who was using Snowflake, I'm seeing this problem with PostgreSQL 16, which suggests it's not database-related.

I haven't found a way to reliably reproduce this behavior.

Of course, switching the autoactivation off, as suggested in the answer to the original issue, may or may not be an acceptable workaround.

For me, changing the behavior of the mouse cursor so that it remains an arrow when moved outside the editing area would at least mitigate the issue by eliminating the main annoyance.

### DBeaver Version

CE 25.0.2

### Operating System

Windows 11 Pro 24H2

### Database and driver

PostgreSQL 16.6, JDBC driver 42.7.2

E1izabeth:
I have managed to reproduce the issue when you have quite a long query (more than a screen), then somewhere in the middle trying to complete a column name by `alias. |`, then immediately press backspace and call completion with Ctrl+Space again.


IvanGorshechnikov (approved):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/semantics/SQLBackgroundParsingJob.java
```diff
@@ -664,7 +674,16 @@ private void accomplishWork(int parsedOffset, int parsedLength) {
                 log.debug("known is " + knownRegionStart + "-" + knownRegionEnd);
             }
             this.isRunning = false;
-            this.lastParsingFinishStamp.complete(System.currentTimeMillis());
+            this.signalAccomplished();
+        }
+    }
+    
+    private static int saturatedSum(int a, int b) {
```
ShadelessFox, line 681: How does saturated addition help mitigate the issue? Are we talking about scripts 2M characters long or more?

E1izabeth: We use Integer.MAX_VALUE to indicate that the cached information was invalidated for the fragment of text till its end, no matter where it is, so it might happen that we'll need to reparse the whole tail of the text.

Consider two moments of time:
A - when a certain input operation happens, and we decide to parse the text till the end,
B - when the background parsing actually happens.
There might be an unknown number of other input operations between these moments, changing text in an unpredictable way, so there is no point in remembering the text length from moment A.
So, the logic being fixed computes the actual text region to reparse, including transformation of this "parse till the end of text" information into the actual end of region position when the end of text is covered by the visible region.

ShadelessFox: So either `a` or `b` might be `Integer.MAX_VALUE`, and we perform saturated addition to handle these cases?

E1izabeth: Right

ShadelessFox: Okay, I guess

ShadelessFox (approved):

uslss:
verfied

PR #38186, opened 2025-05-23
dbeaver/dbeaver#37511 Added sql parser support for alter privileges

### LINKED ISSUE #37511: SQL editor flag `alter default privileges in schema` statement as invalid
### Description

The SQL editor flags alter database T-SQL statements as invalid with the error message "table reference expected":

Image

### DBeaver Version

Community Edition 25.0.0.202503021833

### Operating System

Windows 11

### Database and driver

PostgreSQL 17.2 with PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

_No response_

### Additional context

_No response_

Nexus6v2:
Covered all rules for `alter default privileges` defined in the [Postgres doc](https://www.postgresql.org/docs/current/sql-alterdefaultprivileges.html).


plugins/org.jkiss.dbeaver.model.lsm/grammar/SQLStandardParser.g4
```diff
@@ -100,6 +100,7 @@ intervalLiteral: INTERVAL sign? valueExpressionPrimary intervalQualifier;
 characterSetSpecification: qualifiedName;
 characterSetName: qualifiedName;
 schemaName: qualifiedName;
+schemaNameList: schemaName (COMMA schemaName)*;
```
E1izabeth: In lexer, it's called "Comma". Does it work in upper case too?

E1izabeth (commented):
Also keywords `GRANT`, `PRIVELEGES`, `TABLES`, etc. are missing in SQLStandardLexer.g4. Please add them, otherwise I doubt it will work

plugins/org.jkiss.dbeaver.model.lsm/grammar/SQLStandardParser.g4
Nexus6v2: Good point. I don't think I checked multiple args and didn't think it's case-sensitive.
Will fix!

plugins/org.jkiss.dbeaver.model.lsm/grammar/SQLStandardLexer.g4
```diff
@@ -404,7 +404,21 @@ WORK: W O R K ;
 WRITE: W R I T E ;
 YEAR: Y E A R ;
 ZONE: Z O N E ;
-
+PRIVILEGES: P R I V I L E G E S ;
```
E1izabeth, line 410: Also PUBLIC keyword is not defined as a token.

E1izabeth (commented):
Please add all new keywords to `nonReserved` as we discussed. I believe they're all non reserved in most (or all) databases

plugins/org.jkiss.dbeaver.model.lsm/grammar/SQLStandardLexer.g4
Nexus6v2: Good catch

Nexus6v2:
@E1izabeth 
I've checked the list of [Postgres' reserved keywords](https://www.postgresql.org/docs/current/sql-keywords-appendix.html) and added all new ones that are not in it to the according list in parser

E1izabeth:
What about PUBLIC keyword? :) I believe it should be non-reserved too, because we have a `public` schema/database in many databases

E1izabeth (commented):
What's about GRANT, REVOKE, FOR, TRUNCATE?

Nexus6v2:
@E1izabeth 
`GRANT, REVOKE, FOR, TRUNCATE` are reserved in all of ~top 10 dbs (`PostgreSQL, MySQL, MariaDB, SQL Server, Oracle, SQLite (except `FOR`), DB2, Firebird, Teradata, Snowflake`)

I decided to leave it out.

E1izabeth (dismissed):

IvanGorshechnikov (dismissed):

Matvey16:
Verified

serge-rider:
Are we sure that we want to add all PG-specific stuff in the default parser?

Nexus6v2:
@serge-rider I think so. 
As I understood, currently there are no dialect specific parsers.
Do you think this change requires us to make one?

PR #38196, opened 2025-05-26
#37602 Add the setting for column names completion

### LINKED ISSUE #37602: Setting for column names completion
If a table has an alias, we want to always use this alias for columns. If the table doesn't have an alias, it should be regulated by setting whether to prepend the table name.
Conflicting column names should be proposed as it works now - always with table names.

### LINKED ISSUE #37815: Aliases aren't added when autocomplete is used after * for conflicting column names
### Description

Aliases aren't added when autocomplete is used after *

### DBeaver Version

DBeaver CE 25.0.3 ea

### Operating System

Win11

### Database and driver

MariaDB and PostgreSQL have been tested

### Steps to reproduce

Precondition:

Image

STR:

1. Open SQL Editor and write select * table_name table_aliase (according settings in SQL Editor completion)
2. Place cursor after * and press Ctrl+space

AR: columns without aliases appear
ER: columns with aliases appear

https://github.com/user-attachments/assets/d049cf3d-717a-4c7d-8763-44353ae41772

### Additional context

_No response_

E1izabeth:
By default, the new setting is false, which means that if table don't have an alias, we don't want to prepend its name to the column. Implemented exactly as it's written in the linked issue.
image


IvanGorshechnikov (approved):

ShadelessFox (approved):

uslss:
verified

PR #38202, opened 2025-05-26
dbeaver/pro#5691 Fix query execution in Ai chat

labanovichttps:


IvanGorshechnikov (dismissed):

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -3036,9 +3036,29 @@ public boolean transformQueryWithParameters(SQLQuery query) {
...
+            && getDataSourceContainer().isConnected()
+            && getExecutionContext() != null
+        ) {
+            if (onFinish != null) {
+                onFinish.onTaskFinished(Status.OK_STATUS);
+            }
+        } else {
+            checkSession(status -> {
+                if (status.isOK() && getExecutionContext() == null) {
+                    status = new Status(IStatus.ERROR, SQLEditorActivator.PLUGIN_ID,
```
serge-rider: Why create new status? Can you just pass the original?

labanovichttps: I need to specify the problem in case where executionContext == null, because Status is immutable

serge-rider: Then use makeErrorStatus

serge-rider (changes requested):

PR #38210, opened 2025-05-27
dbeaver/pro#4119 license type stats

Destrolaric:
Closes dbeaver/pro#4119


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPApplication.java
```diff
@@ -66,6 +67,8 @@ default void beforeWorkspaceInitialization() {}
 
     boolean isDetachedProcess();
 
+    Map<String, String> getAdditionalApplicationProperties();
```
ShadelessFox, line 71: Please specify nullability annotations and add a javadoc

PR #38221, opened 2025-05-28
#37870 indexes do not show in the right order

### LINKED ISSUE #37870: indexes view do not show columns for composite index in the right order
### Description

When showing composite indexes on indexes list (table->properties->indexes) indexed columns do not respect index "Seq_in_index" or the right position/order inside the index.

Image

in this image you can clearly see that the index "orders_user_id_status_index" have:
(status, user_id)
but the index is declared (user_id, status) and this might confuse the developer!

Solution:
DBeaver should list the columns inside a composite index in the right order as declared in "Seq_in_index".

### DBeaver Version

Community Edition 25.0.3.202504201842

### Operating System

Win11 24H2

### Database and driver

DB: MySql 8
Driver: MySQL

### Steps to reproduce

1. create a table with a composite index (Ex. (user_id, status))
2. go to DBeaver and click on the table "Properties"
3. than click on the left menu "Indexes"
4. the index columns do not respect the right index columns order as declaration

### Additional context

_No response_

vkirst:
Fix sorting column list

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNUtils.java
@@ -36,7 +36,6 @@ import org.jkiss.dbeaver.model.struct.DBSObject;
 import org.jkiss.dbeaver.model.struct.DBSObjectContainer;
 import org.jkiss.dbeaver.model.struct.DBSWrapper;
 import org.jkiss.dbeaver.model.struct.rdb.DBSTableColumn;
-import org.jkiss.dbeaver.model.struct.rdb.DBSTableIndexColumn;
 import org.jkiss.dbeaver.runtime.DBWorkbench;
 import org.jkiss.utils.AlphanumericComparator;
 import org.jkiss.utils.ArrayUtils;
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNUtils.java
@@ -36,6 +36,7 @@ import org.jkiss.dbeaver.model.struct.DBSObject;
 import org.jkiss.dbeaver.model.struct.DBSObjectContainer;
 import org.jkiss.dbeaver.model.struct.DBSWrapper;
 import org.jkiss.dbeaver.model.struct.rdb.DBSTableColumn;
+import org.jkiss.dbeaver.model.struct.rdb.DBSTableIndexColumn;
 import org.jkiss.dbeaver.runtime.DBWorkbench;
 import org.jkiss.utils.AlphanumericComparator;
 import org.jkiss.utils.ArrayUtils;
@@ -139,6 +140,10 @@ public class DBNUtils {
         }
 
         if (firstChild instanceof DBNContainer) {
+            return;
+        }
+
+        if (firstChild instanceof DBNDatabaseItem item && item.getObject() instanceof DBSTableIndexColumn) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/controls/ViewerColumnController.java
```diff
@@ -884,7 +884,7 @@ public int compare(Viewer viewer, Object e1, Object e2) {
             if (e1 instanceof DBNDatabaseNode n1 && e2 instanceof DBNDatabaseNode n2) {
                 Object o1 = n1.getValueObject();
                 Object o2 = n2.getValueObject();
-                if (o1 instanceof DBPObjectWithOrdinalPosition c1 && o2 instanceof DBPObjectWithOrdinalPosition c2) {
+                if (o1 instanceof DBSTableIndexColumn c1 && o2 instanceof DBSTableIndexColumn c2) {
```
serge-rider: I think DBSEntityAttributeRef + DBPObjectWithOrdinalPosition is better.
Anyhow it is a bad idea to add this logic in the default comparator. User may forcibly reorder index columns by any other attribute.

We order table columns correctly (by ordinal position, as they come from database model). 
How index columns are different?

Also let's add index column index in the list of visible columns (currently we don't show it at all)

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/controls/ViewerColumnController.java
```diff
@@ -881,12 +879,9 @@ public int compare(Viewer viewer, Object e1, Object e2) {
...
             }
 
-            if (e1 instanceof DBNDatabaseNode n1 && e2 instanceof DBNDatabaseNode n2) {
-                Object o1 = n1.getValueObject();
-                Object o2 = n2.getValueObject();
-                if (o1 instanceof DBPObjectWithOrdinalPosition c1 && o2 instanceof DBPObjectWithOrdinalPosition c2) {
-                    return Integer.compare(c1.getOrdinalPosition(), c2.getOrdinalPosition());
-                }
+            Object input = viewer.getInput();
+            if (input instanceof Collection<?> list && !list.contains(e1) && !list.contains(e2)) {
```
ShadelessFox, line 885: I don't understand the meaning behind this logic. What does it do? Why do we operate on the viewer's input and completely ignore the arguments we are supposed to compare?

E1izabeth: I suppose it was Serge's suggestion to apply the comparator only for top-level nodes

ShadelessFox: Okay, I see. I would do this instead, so we don't assume the type of the input:
```java
if (viewer instanceof StructuredViewer sv && sv.getContentProvider() instanceof IStructuredContentProvider provider) {
    Object[] elements = provider.getElements(viewer.getInput());
    if (!ArrayUtils.contains(elements, e1) && !ArrayUtils.contains(elements, e2)) {
        return 0;
    }
}
```

ShadelessFox (dismissed):

IvanGorshechnikov (dismissed):

Matvey16:
Original bug still persists

dbeaver_pynUtMXovP

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNUtils.java
```diff
@@ -142,6 +143,10 @@ private static void sortNodes(DBNNode[] children) {
             return;
         }
 
+        if (firstChild instanceof DBNDatabaseItem item && item.getObject() instanceof DBSTableIndexColumn) {
```
serge-rider: Maybe DBSTableIndexColumn -> DBPObjectWithOrdinalPosition ?
We also don't want to order constrain columns

serge-rider (changes requested):

Matvey16:
Columns now are unaffected by alphabetical sorting preference


https://github.com/user-attachments/assets/970a64f2-7922-46e6-8683-c4831e3967b0

Matvey16:
Verified

PR #38246, opened 2025-05-30
dbeaver/dbeaver#38231 Fix CUBRID view DDL extraction truncating UNION ALL fragments

### LINKED ISSUE #38231: [CUBRID] View DDL ouput is incomplete when using UNION ALL
### Description

When a view is created using `UNION ALL`, the DDL shown in the _Properties > Definition_ tab in DBeaver does not display the full query. It only shows the first `SELECT` block, omitting the rest of the unioned statements.

For example, the following view is created based on three tables:
```sql
CREATE TABLE t_1(col_1 varchar(10));
CREATE TABLE t_2(col_1 varchar(10));
CREATE TABLE t_3(col_1 varchar(10));

CREATE VIEW v_1 AS 
SELECT col_1 FROM t_1 
UNION ALL 
SELECT col_1 FROM t_2 
UNION ALL 
SELECT col_1 FROM t_3;
```

However, the DDL displayed in the _Definition_ tab is incomplete:
```sql
CREATE OR REPLACE
VIEW public.v_1 AS
SELECT
    [public.t_1].[col_1]
FROM
    [public.t_1] [public.t_1];
```

The expected DDL should include the full `UNION ALL` sql:
```sql
CREATE OR REPLACE 
VIEW public.v_1 AS
SELECT
    [public.t_1].[col_1]
FROM
    [public.t_1] [public.t_1]
UNION ALL
SELECT
    [public.t_2].[col_1]
FROM
    [public.t_2] [public.t_2]
UNION ALL
SELECT
    [public.t_3].[col_1]
FROM
    [public.t_3] [public.t_3];
```

This issue makes it difficult to understand or reuse the full view definition from the UI. The DDL generator should be enhanced to properly handle views that use `UNION ALL`.

I would like to take on this task and submit a pull request soon.

### DBeaver Version

- DBeaver 25.0.5

### Operating System

- macOS Sequoia 15.5

### Database and driver

1. CUBRID 11.4.0.1781
2. cubrid-jdbc 11.3.0.0047

### Steps to reproduce

1. Connect to a CUBRID database
2. Create three tables
3. Create a view using `UNION ALL`
4. Open the view’s _Properties > Definition_ tab
5. Observe the DDL output — it only includes the first `SELECT`, not the full unioned query

### Additional context

Image

Srltas:
### Problem
When a CUBRID view is defined with multiple `UNION ALL` clauses, the _Properties > Definition_ tab shows only the first `SELECT` block.
`SHOW CREATE VIEW` actually returns one row per fragment, but the current meta-model stops after the first, so the assembled DDL is incomplete.

### What has been done
- Iterate over every row returned by `SHOW CREATE VIEW`.
- Collect non-empty `Create View` fragments and join them with `" union all "` to rebuild the full statement.
- Add null/empty checks and return the fallback `-- View definition not available` when needed.

### How to test
1. Create a view containing multiple `UNION ALL` parts, for example:
```sql
CREATE VIEW v_test AS
SELECT col FROM t1
UNION ALL
SELECT col FROM t2;
```
2. Open _Properties > Definition_ for that view in DBeaver.
    - Before: only the first `SELECT` is shown.
    - After: the full definition with all `UNION ALL` parts appers, properly formatted.
3. Verify that views with single-row definitions still render exactly as before.


E1izabeth (approved):
Thanks for the contribution!

ShadelessFox (commented):
So, each fragment is considered to be an individual select statement, right? Is it always the case that these fragments are glued using `UNION ALL`, or could it be something else? Is it documented somewhere?

Srltas:
> So, each fragment is considered to be an individual select statement, right? Is it always the case that these fragments are glued using `UNION ALL`, or could it be something else? Is it documented somewhere?

Yes, when creating a view using `UNION ALL`, each fragment is indeed stored as a separate SELECT statement.

Unfortunately, this behavior is not documented officially.
For reference, here's the CUBRID manual page for `SHOW CREATE VIEW`:
https://www.cubrid.org/manual/en/11.4/sql/query/show.html#show-create-view

Let me share the actual test results I observed using CSQL, CUBRID's CLI-based SQL interpreter.

**Simple View**
```
csql> show create view view_simple;

=== <Result of SELECT Command in Line 1> ===

  VIEW                  CREATE VIEW
============================================
  'public.view_simple'  'select [public.t_1].[col_1] from [public.t_1] [public.t_1]'
```

**UNION View**
```
csql> show create view view_union;

=== <Result of SELECT Command in Line 1> ===

  VIEW                  CREATE VIEW
============================================
  'public.view_union'   '(select [public.t_1].[col_1] from [public.t_1] [public.t_1] union select [public.t_2].[col_1]
from [public.t_2] [public.t_2])'
```

**UNION ALL**
```
csql> show create view view_union_all;

=== <Result of SELECT Command in Line 1> ===

  VIEW                  CREATE VIEW
============================================
  'public.view_union_all'  'select [public.t_1].[col_1] from [public.t_1] [public.t_1]'
  'public.view_union_all'  'select [public.t_2].[col_1] from [public.t_2] [public.t_2]'
```

As you can see, in the case of `UNION ALL`, the view is stored as multiple SELECT statements rather than a single compound query.

Additionally, I checked the implementation in the CUBRID Manager tool, which retrieves the view DDL. The logic matches what's implemented in DBeaver. Here's the relevant code from CUBRID Manager.

https://github.com/CUBRID/cubrid-manager/blob/f2cbd1a5ea41780bbb744fbc536b040347bd0677/com.cubrid.common.ui/src/com/cubrid/common/ui/common/navigator/GetSchemaDDLTask.java#L99

Hope this helps clarify how view DDLs are handled internally in CUBRID. Let me know if you need further details.

ShadelessFox:
Interesting, thanks for the explanation.

ShadelessFox (approved):

ShadelessFox:
Passing the PR to our QA team now.

MashaKorax:
verified

PR #38258, opened 2025-05-30
Add HPE sqlmx support 

### LINKED ISSUE #37995: I'd like to contribute changes to support HPE NonStop SQL/MX
**HPE NonStop SQL/MX driver does not create correct DDL as "generic" driver**

1. When I generate DDL, fractions from numeric data are lost  
2. Tables are index-organized and do not require a separate index on the primary key. It will be generated by the generic IndexCache.java

**I created solutions in a private copy and would like to provide my changes in the next release**
I made a few corrections in the source code and tested them in a private repo which I cloned.
As it turned out, I don;t need sql-specific code to get this to work, and I believe the changes will not harm other databases that are generic.

I also found two minor issues. 

1. NCHAR datatype was not detected. I changed this in JDBCDataSource.java : if typeName equals "NCHAR" it returns Types.NCHAR
2. SQLSourceViewer.java would display the same data 2x instead. It should clear its working buffer.  

**Alternatives**
I thought I might need specific code as used in e.g., Oracle or Postgresql. Later I might want to add specific code, but for now, the minimal support for SQL/MX is sufficient 

**Additional context**
I am new in this type of collaboration, I would like to cooperate with you and implement carefully.

fjongma:
Create driver definitions and icons for SQL/MX driver to allow basic support for this database.

The generic driver did not produce the correct code in some cases, notably when NUMERIC or DECIMAL data types were used. 

SQL/MX tables are index organized and therefore they do not require a separate unique index on the primary key. Added code to detect this situation.

Attached word document highlights the incorrect definitions in DBeaver vs the version that is in this pull request. 
[Dbeaver change description.docx](https://github.com/user-attachments/files/20518629/Dbeaver.change.description.docx)


plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -176,6 +176,21 @@
             <drivers managable="true">
 
                 <!-- Real drivers -->
+                
+                <driver
```
ShadelessFox, line 180: Please put the declaration in `org.jkiss.dbeaver.ext.sqlmx`

plugins/org.jkiss.dbeaver.ext.sqlmx/pom.xml
```diff
@@ -0,0 +1,14 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>org.jkiss.dbeaver</groupId>
+    <artifactId>plugins</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+     <relativePath>../</relativePath>
+  </parent>
+  <artifactId>org.jkiss.dbeaver.ext.sqlmx</artifactId>
+  <version>3.9.0-SNAPSHOT</version>
```
ShadelessFox, line 12:
```suggestion
  <version>1.0.0-SNAPSHOT</version>
```

plugins/org.jkiss.dbeaver.ext.sqlmx/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,15 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: DBeaver NonStop SQL/MX Model
+Bundle-SymbolicName: org.jkiss.dbeaver.ext.sqlmx;singleton:=true
+Bundle-Version: 3.9.0.qualifier
```
ShadelessFox, line 5:
```suggestion
Bundle-Version: 1.0.0.qualifier
```

plugins/org.jkiss.dbeaver.ext.sqlmx/src/org/jkiss/dbeaver/ext/sqlmx/model/SqlmxSQLDialect.java
```diff
@@ -0,0 +1,40 @@
...
+import org.jkiss.dbeaver.ext.generic.model.GenericSQLDialect;
+import org.jkiss.dbeaver.model.exec.jdbc.JDBCDatabaseMetaData;
+import org.jkiss.dbeaver.model.exec.jdbc.JDBCSession;
+import org.jkiss.dbeaver.model.impl.jdbc.JDBCDataSource;
+
+
+/**
+ * Sqlmx dialect
+ */
+public class SqlmxSQLDialect extends GenericSQLDialect {
```
ShadelessFox, line 29: Seems redundant

plugins/org.jkiss.dbeaver.ext.sqlmx/src/org/jkiss/dbeaver/ext/sqlmx/model/SqlmxDataSource.java
```diff
@@ -0,0 +1,51 @@
...
+        throws DBException {
+        super(monitor, container, metaModel, new SqlmxSQLDialect());
+//FJ        log.debug ("****** SqlmxDataSource called ");
+    }
+
+    @Override
+    protected boolean isPopulateClientAppName() {
+    	// SQL/MX drivr does not support this type of settingthe application name
+        return false;
+    }
```
ShadelessFox, line 48: It's automatically set to `false` for all data sources that inherit from `GenericDataSource`

https://github.com/dbeaver/dbeaver/blob/268867bb73665954d082ca2e6191b44c8ad5dac3/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericDataSource.java#L211-L213

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCDataSource.java
```diff
@@ -646,6 +646,11 @@ private static int getValueTypeByTypeName(@NotNull String typeName, int valueTyp
             if (SQLConstants.DATA_TYPE_TINYINT.equalsIgnoreCase(typeName)) {
                 return Types.TINYINT;
             }
+        } else if (valueType == Types.ROWID) {
```
ShadelessFox, line 649: I don't think it should be here. If you think that other drivers may benefit from it, then please remove database-specific comment.

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCDataSource.java
```diff
@@ -664,6 +669,7 @@ public static DBPDataKind getDataKind(@NotNull String typeName, int valueType)
             case Types.BOOLEAN:
                 return DBPDataKind.BOOLEAN;
             case Types.CHAR:
+            case Types.NCHAR:                                //FJ [HPE sqlmx] NCHAR missing
```
ShadelessFox, line 672: Same here

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableColumnManager.java
```diff
@@ -56,7 +56,22 @@ protected interface ColumnModifier<OBJECT_TYPE extends DBPObject> {
...
+    	String typeName = column.getTypeName();    
+        if (typeName.equals("NUMERIC SIGNED") || typeName.equals ("NUMERIC UNSIGNED") ) {
+            log.debug("****** SQLTableColumnManager changing typeName for col: "+ column.getName() + " to NUMERIC");
+        	typeName = "NUMERIC";
+        } else {
+        	if (typeName.equals("DECIMAL SIGNED") || typeName.equals ("DECIMAL UNSIGNED") ) {
+                log.debug("****** SQLTableColumnManager changing typeName for col: "+ column.getName() + " to DECIMAL");
+            	typeName = "DECIMAL";
+        	}
+        }
```
ShadelessFox, line 74: This one is for sure database-specific. Please remove it and provide your own custom manager for table columns.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
```diff
@@ -110,6 +110,9 @@ protected Map<String, Object> getSourceOptions() {
         if (editorInput instanceof IDatabaseEditorInput) {
             Collection<String> attributeNames = ((IDatabaseEditorInput)editorInput).getAttributeNames();
             options.put(DBPScriptObject.OPTION_DDL_SOURCE, true);
+            //FJ [HPE sqlmx]: must clear existing ddl in buffer, by forcing refresh
+            //    if not, the same DDL lines may appear twice when FK constraints are present.
+            options.put(DBPScriptObject.OPTION_REFRESH, true);
```
ShadelessFox, line 115: This will affect all viewers in all databases.

ShadelessFox (changes requested):
Thanks for the contribution. I'm sorry, but you cannot include database-specific hacks or workarounds in different parts of the application.

Besides that, please:
- Remove redundant log messages, code that does nothing and commented-out code
- Consider providing database-specific model classes that fix the unintended behavior where applicable.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableColumnManager.java
fjongma: Most SQL databases don't support only NUMERIC (or DECIMAL) datatypes which are signed. UNSIGNED is legacy from decades ago. (and NonStop SQL dates from mid 1980). DBeaver generic code then discards the scale, because it does not check for SIGNED or UNSIGNED. INTEGERs can be signed/unsigned but that is not addressed with this change.  All this code does is to define any numeric (signed or unsigned to NUMERIC, which does not break any code. I do not disagree with the remark on database specific code comment. I wanted to avoid writing more code than necessary.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
fjongma: I think that any generic database that separates FK definitions will experience this. I cannot verify, because Oracle and  Postgres use their specific code for this.  I put my initials and the driver in for tracking purposes, not because the driver requires it.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableColumnManager.java
ShadelessFox: I don't see any mentions of `SIGNED` and `UNSIGNED` keywords in the SQL 92 spec. Moreover, the databases I tested against (PostgreSQL, MySQL, Oracle) reported syntax errors when attempting to cast to either type.

The query I used for testing:
```sql
SELECT CAST(1 AS NUMERIC SIGNED);
```
Please point out if I'm wrong.

fjongma:
Thanks for the comments. It is my first pull request in this large project and I had to make a few guesses. I will address the comments, remove specific references to SQL/MX, because I don;t think I put anything.  You suggested I write a dedicated version of SQLTableColumnManager. While it is pretty "deep down" into the application I am lacking a good example of what is needed.  I will create a new pull request and see if that is acceptable. 

Procedure question: should I close with comment , or just comment and to continue this thread? Thx.

ShadelessFox:
You can continue working on this PR, but please convert it to a draft. If you need a new PR instead, you can close this one.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableColumnManager.java
fjongma: "Wrong" is not the issue. The SQL 92 Spec defines all numeric datatypes as signed. However, unsigned definitions are supported by some DBMSes as vendor-specific extensions. MySQL, for example, supports unsigned int, sqlmx does too. When I researched this, the most common question was "Why does SQL Server not support UNSIGNED datatype?". So it is not such a wild idea. 
My change makes sure that the type name as returned by the database, "NUMERIC SIGNED" is modified to NUMERIC which will generate the correct datatype definition, because scale and precision are added only if the type name equals "NUMERIC".  The DDL column definitions after this change contain the correct precision and scale instead of defaulting to having no scale.  

I agree with you that specific features or issues may best be resolved in dedicated database-specific code. Something I may explore time and resources permitting. This change, however seemed an easy solution to fix an annoying issue in a fine application.

fjongma:
Thanks for the suggestions and comments. I will close this PR, make changes to my local version and will provide an new PR at a later time.

Issue #5743, opened 2019-04-17
Prompt for parameters


Cha0sSoul:
Installed the new 6.0.2 version and now whenever I trie to run my previously fine working scripts I get a prompt to bind parameters:
![image](https://user-images.githubusercontent.com/49717958/56287928-4ffad180-6126-11e9-8032-5246f985afe5.png)
Previously this part was working fine:
![image](https://user-images.githubusercontent.com/49717958/56287973-6739bf00-6126-11e9-8087-44aa475213ec.png)
Now I get an error:
![image](https://user-images.githubusercontent.com/49717958/56287986-728cea80-6126-11e9-9d42-a77521b09743.png)
Can you please tell me what has changed, where is the problem now ?

ced75:
Hi,

I have exactly the same issue with PostgreSQL and the "$" character.
With PostgreSQL, it is possible to use double dollar as character string delimiter.
For example, the following query is syntaxically correct:
`SELECT $$A text$$, col1
FROM table1;`

But when I try to execute the query, DBeaver asks for bind parameters, since there is no bind parameters, because it automatically detects $ as a bind parameter.

Is it possible to choose its own character as bind parameter character? 
Thanks in advance for your help.

Regards,
Cedric

ced75:
Ok, I have found that it is possible to inactivate variables in the SQL settings.

serge-rider:
the problem was in incorrect parameter prefix config in SQL Server extension. Now resolved.

PR #38262, opened 2025-05-30
#5743 Add neo4j  fake driver

vkirst:


E1izabeth (dismissed):

E1izabeth (dismissed):

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -1507,6 +1488,15 @@
                         categories="nosql,timeseries,aws">
                     <notAvailableDriver title="%not.available.title" description="%not.available.description" message="%driver.timestream.reason"/>
                 </driver>
+                <driver
+                        id="neo4j-ce"
```
serge-rider: WE shouldn't change old driver ID.
Users may have this config already, it must keep working when they switch to PRO.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -1507,6 +1488,15 @@
                         categories="nosql,timeseries,aws">
                     <notAvailableDriver title="%not.available.title" description="%not.available.description" message="%driver.timestream.reason"/>
                 </driver>
+                <driver
```
serge-rider, line 1491: You cannot delete old driver. neo4j_jdbc is used in PRO products for replacement.
Keep old driver

serge-rider (changes requested):

E1izabeth (approved):

uslss:
verified

PR #38263, opened 2025-05-30
#37872 Don't use `AS` before table alias for legacy firebird

### LINKED ISSUE #37872: Disable table alias when opening a table from explorer
### Description

I'm using a legacy Firebird database.
Even when you disable the option in SQL Editor -> Code Completion -> Insert table aliases (in FROM clause)
Doesn't work when you open it from the explorer
Image

Image

It works as expected when you use the "Read data in SQL console" option.

### DBeaver Version

25.0.3

### Operating System

Windows 11

### Database and driver

Firebird 1.5.6
Driver class: org.firebirdsql.jdbc.FBDriver

[jaybird-full-2.2.15.jar.zip](https://github.com/user-attachments/files/19881746/jaybird-full-2.2.15.jar.zip)

### Steps to reproduce

1.- Configure the driver connection
2.- Open a database with that driver
3.- Open/view a table from the explorer (F4) 
4.- Go to Data tab

### Additional context

This is a legacy database, very old version, doesn't support the "AS" keyword as ALIAS.
==================================
Firebird 1.5.6       (Win32 Build)
==================================

E1izabeth:
We don't have such an old Firebird to test, but we can fix it blindly.
For QA: check that for **our** Firebird, the query in QM contains AS when you open the table from the navigator.


IvanGorshechnikov (approved):

labanovichttps (approved):

MashaKorax:
verified

PR #38297, opened 2025-06-03
#38265 Fix varchar mapping in data transfer for Clickhouse

### LINKED ISSUE #38265: Since dbeaver 25.0.4, csv upload to clickhouse fails as it tries to use unsupported data type Enum(50)
### Description

Since dbeaver 25.0.4, csv upload to clickhouse fails as it tries to use unsupported data type Enum(50).

I uploaded a simple csv that has 2 columns - destination is a clickhouse db using latest driver. Here's a sample CSV:
```
Name,Number
name1,99.5941
name2,99.4639
name3,99.6027
```

When I try to import it on 25.0.4 or 25.0.5, it fails with:

Can't create or update target table:
```
CREATE TABLE `local`.contract_slo_copy (
	BsvcName Enum(50),
	`SLA-SLO` Float64
) ENGINE = Log;
```

But with 25.0.3, it works perfectly and correctly uses VARCHAR(50).

### DBeaver Version

24.0.5

### Operating System

Macos 14.3 (23D56)

### Database and driver

Database is clickhouse
Driver in dbeaver is: clickhouse client 0.8.5

### Additional context

Fails on 25.0.4 and 25.0.5, but works on 25.0.3

Thanks in advance!

E1izabeth:
The provided example csv should successfully pass the import with table creation.
As stated in docs https://clickhouse.com/docs/sql-reference/data-types/string, Clickhouse ignores varchar and other string data type aliases anyway and uses String data type instead.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/ClickhouseConstants.java
+++ b/plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/ClickhouseConstants.java
@@ -30,4 +30,5 @@ public class ClickhouseConstants {
 
     public static final String DATA_TYPE_IPV4 = "ipv4";
     public static final String DATA_TYPE_IPV6 = "ipv6";
+    public static final String DATA_TYPE_STRING = "String";
 }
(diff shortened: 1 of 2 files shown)


MashaKorax:
verified

PR #38301, opened 2025-06-03
dbeaver/pro#5677 Update JSqlParser to 5.3 version

E1izabeth:


Matvey16:
Dbeaver builds with JSQLParser 5.1.0

E1izabeth:
Because it should be built with "custom p2" checkbox

PR #38303, opened 2025-06-03
Prefer shaded version of Salesforce Data Cloud driver

### LINKED ISSUE #38306: Salesforce datacloud shaded driver
WE should use shaded maven artifact

jschneidereit:
The default artifact for `com.salesforce.datacloud:jdbc` is currently shaded in an effort to fix some issues described in [forcedotcom/datacloud-jdbc#52](https://github.com/forcedotcom/datacloud-jdbc/pull/52). My hope is that this will direct DBeaver to download the shaded version of the Salesforce Data Cloud driver and eventually we should be able to revert our driver back to non-shaded for no-classifier and only shaded with the `shaded` classifier.


E1izabeth:
Thanks for the contribution! We passed Pull Request to our QA team for testing

uslss:
Cannot download the driver due to an error
image

PR #38305, opened 2025-06-04
Display the field's position after the field name

330079598:
When a database has many fields, it can be troublesome to find the corresponding value for a specific field on the Data page. By displaying the position of each field after its name, it becomes much easier to locate and query the values we need.

Here's how it looks in action.
image
image

To enable this feature, go to the following settings.
image


E1izabeth:
Thanks for the contribution.
Please take a look at the contributions guide https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code#commit-message-rules. Create or link the issue and use the issue number in the commit messages.

Issue #13727, opened 2021-08-30
Can't execute set,unset,echo commands if carriage is placed after last letter


DariaMarkaryan:
#### System information: 
- Win10, DBeaver 21.2.0

#### Steps to reproduce, if exist:
Type @set var=somevalue and press ctrl+enter
Expected: variable value is set
Actual: nothing happened. You need to move the carriage from the end of the line to anywhere in the middle and then if you pressed ctrl+enter the variable will be created

https://user-images.githubusercontent.com/43252954/131304644-08501fc7-1829-4c34-9744-6cb14bb3a919.mp4

uslss:
verified

uslss:
verified

DariaMarkaryan:
Reproduceable again. For **Alt+X** it works well, but for **Ctrl+Enter** exist the problem is described in the ticket

PR #38311, opened 2025-06-04
dbeaver/pro#5865 sql completion should not work inside commands

### LINKED ISSUE #5994: Auto format: ignore keywords in comments
I have “convert keyword case” selected in SQL Completion and keyword case “Upper” in SQL Formatting. This works great when I’m typing queries, however it also applies to comments. For example whenever I type the word “tool in a comment it gets converted to “TO”. Is there a way to disable auto format for comments?
  
#### System information: 
- Operating system (distribution) and version: Windows 7
- DBeaver version: 6.0.4

E1izabeth:
- Autocomplete and case converter should not trigger automatically when you write text in a DBeaver command (for example, `@ai`) or comment 
- The following case should not be reproducible anymore https://github.com/dbeaver/dbeaver/issues/5994#issuecomment-848869791
- The following bug should not appear: https://github.com/dbeaver/dbeaver/issues/13727


MashaKorax:
Postgres


https://github.com/user-attachments/assets/a3daf1d5-2602-4633-a80e-3741f197b8db

MashaKorax:
verified

PR #38330, opened 2025-06-05
 dbeaver/pro#4219 Refactor table mapping UI: replace "Configure" with "Customise" and improve target editor with dynamic combo box logic.

labanovichttps:


IvanGorshechnikov (dismissed):

E1izabeth (dismissed):

Matvey16:
Picking a table from browse button now doesn't change the target table 

https://github.com/user-attachments/assets/16a94a8b-45d2-4356-b2bb-71854ce0a508

E1izabeth:
I have fixed the issue.
But actually all the code around is over complicated. It needs refactoring, which this PR doesn't provide despite the PR is called "Refactor...".

PR #38338, opened 2025-06-08
Sqlmx resubmitted PR after comments in #38528

### LINKED ISSUE #37995: I'd like to contribute changes to support HPE NonStop SQL/MX

fjongma:
Tried the suggested change in `plugins/org.jkiss.dbeaver.ext.generic/plugin.xml` but the icons could not be found by Dbeaver. See my remarks in thread  #37995 .  FYI, I left my changes in `sqlmx plugin.xml` as to show what I have tried.

Stiil would liketo apply the changes to `SQLTableColumnManager` since I believe this change is valid for all generic drivers which may see the datatype NUMERIC SIGNED. Also mentioned in #37995 and #38528


plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -176,6 +176,20 @@
             <drivers managable="true">
 
                 <!-- Real drivers -->
+
```
serge-rider, line 179: It is bettter to move all changes into new extension sqlmx, it doesn't make sense to keep in in generic.

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -176,6 +176,20 @@
             <drivers managable="true">
 
                 <!-- Real drivers -->
+                
+                <driver
+                    id="sqlmx"
+                    label="NonStop SQL/MX"
+                    icon="icons/sqlmx_icon.png"
+                    iconBig="icons/sqlmx_icon_big.png"
```
serge-rider, line 184: Unfortunately icons looks bad. They are not transparent and not adapted to dark theme.
Is it possible to change it? If no then you can leave default icon.

plugins/org.jkiss.dbeaver.ext.sqlmx/src/org/jkiss/dbeaver/ext/sqlmx/model/SqlmxMetaModel.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * SqlmxMetaModel
+ */
+public class SqlmxMetaModel extends GenericMetaModel
+{
+//    private static final Log log = Log.getLog(SqlmxMetaModel.class);
```
serge-rider, line 52: Please remove commented code

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLTableColumnManager.java
```diff
@@ -56,7 +56,21 @@ protected interface ColumnModifier<OBJECT_TYPE extends DBPObject> {
     }
 
     protected final ColumnModifier<OBJECT_TYPE> DataTypeModifier = (monitor, column, sql, command) -> {
-        final String typeName = column.getTypeName();
+    	// Some databases support SQL Extensions that allow signed and unsigned numeric definitions.
+    	// They may therefore contain additional information in the typeName field (as shown below. In Java, the NUMERIC
+    	// and DECIMAL data types are always signed. If the typeName value passes unchanged, the precision and scale 
+    	// information will be lost and the DDL column information will become incorrect.
+    	// final String typeName = column.getTypeName();
+    	String typeName = column.getTypeName();
```
serge-rider, line 64: 1. These changes most be moved into sqlmx plugin (in extended manager, for an example see PostgreTableColumnManager).
2. Do not use abessive logging with asterisks.
3. Move type name into constants.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLSourceViewer.java
```diff
@@ -110,6 +110,9 @@ protected Map<String, Object> getSourceOptions() {
         if (editorInput instanceof IDatabaseEditorInput) {
             Collection<String> attributeNames = ((IDatabaseEditorInput)editorInput).getAttributeNames();
             options.put(DBPScriptObject.OPTION_DDL_SOURCE, true);
+            // Must clear existing ddl in buffer, by forcing refresh
+            // if not, the same DDL lines may appear twice when FK constraints are present.
+            options.put(DBPScriptObject.OPTION_REFRESH, true);
```
serge-rider, line 115: Unrelated change and it affects default product behavior. Probably can be done in a separate ticket/PR.

serge-rider (changes requested):

IvanGorshechnikov:
Hi @fjongma 

Could you resolve the comments below (comments in code, images, and others)?
Without these fixes, we couldn't merge your PR.

fjongma:
Unfortunately, I am no longer working for HPE and have no resources left to address the suggestions to create a new extension as was suggested. I'm sorry I did not close this request any sooner. Thanks for your suggestions to improve my changes.

PR #38349, opened 2025-06-09
dbeaver/pro#5538 api for ai chat conversations

yagudin10:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/OpenAIClient.java
```diff
@@ -111,6 +112,10 @@ public Flow.Publisher<ChatCompletionChunk> createChatCompletionStream(
                         try {
                             ChatCompletionChunk chunk = MAPPER.readValue(data, ChatCompletionChunk.class);
                             publisher.submit(chunk);
+                            if (i[0] > 5) {
```
serge-rider: I thought we do it generally for all AI engines in AIAssistant invocation.
Let's add contant at least.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIMessage.java
```diff
@@ -39,11 +42,13 @@ public class AIMessage {
      * Creates AI message
      */
     public AIMessage(
+        @NotNull UUID uuid,
```
serge-rider: IT conflicts with https://github.com/dbeaver/pro/issues/5835
We must sync.
In this ticket we didn't plan to change any model.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/OpenAIClient.java
yagudin10: this will be deleted
added this for handling error events at the front-end side

Issue #37180, opened 2025-02-14
Clickhouse database tree display error


hyl6688:
### Description

![Image](https://github.com/user-attachments/assets/82a0da5c-0260-4ded-afa6-6ed073075e63)

### DBeaver Version

24.3.4

ShadelessFox:
Can't reproduce. Please provide more information about the database; provide a screenshot of the "test connection" results.

ewanwalk:
I am also seeing this issue only on "ClickHouse" drivers not the legacy ones. This is new with some recent versions

![Image](https://github.com/user-attachments/assets/4c50ff96-7437-4be1-a0da-32384a13b493)

krisnej:
We have the same issue in our team since a couple of weeks.
![Image](https://github.com/user-attachments/assets/89a6b1e2-7d31-4e1c-b111-ce7cc29e4439)

irustam:
I have this issue too

![Image](https://github.com/user-attachments/assets/5341ad3b-78a4-4ff0-9f02-940954c38f00)

ShadelessFox:
Thanks for the bug report

lucasguo:
The same problem. 
I think this is also the reason why autocomplete function is failed, and full of error in the editor that tell me table is not found.

![Image](https://github.com/user-attachments/assets/bca4e2e3-6d3a-4d71-a821-5b734cd39090)

However, these are working well under version 24.

lucasguo:
The problem can be resolved by re-downloading clickhouse driver.
Right click connection -> Edit Connection -> Driver Settings -> Libraries -> Download/Update -> Download

![Image](https://github.com/user-attachments/assets/4008e6f6-2443-41e2-8438-db6356096dc9)

irustam:
> The problem can be resolved by re-downloading clickhouse driver. Right click connection -> Edit Connection -> Driver Settings -> Libraries -> Download/Update -> Download
> 
> ![Image](https://github.com/user-attachments/assets/4008e6f6-2443-41e2-8438-db6356096dc9)

Thank you! It helped.

yousanghz:
I have the same issue. 

server: ClickHouse 23.8.16.1.

driver: ClickHouse JDBC Driver 0.8.6

E1izabeth:
Please try latest DBeaver version.

krisnej:
> Please try latest DBeaver version.

I still have this issue with DBeaver version 25.1.0 and Driver 0.8.5

[...]

PR #38373, opened 2025-06-12
dbeaver/pro#5958 Omit catalogs for ClickHouse driver versions > 0.8

ShadelessFox:
This fixes the issue with missing objects in the navigator for driver versions > 0.8. Funnily enough, 0.8 claims to be 0.7, so the fix doesn't work for that particular driver version.

The issue with SSL can't be reproduced, and I cannot fix it, unfortunately. I assume it has something to do with connection configuration.


E1izabeth (approved):

IvanGorshechnikov (approved):

E1izabeth:
Does it fix #37180? Please check

uslss:
verified for 0.8.5
before fix:
image


after fix:

image

chuntungho:
This issue also occurs in Clickhouse (Legacy)

PR #38380, opened 2025-06-13
#38313 Fix CTE contexts for expressions, fix outline icon for CTE source node

### LINKED ISSUE #38313: SQL Parser doesn't recognize CTE names and columns in condition expressions subqueries
### Description

This toy example highlights the problem:

```
with
test_1 as (
	select 1 as aaa
)
,
test_2 as (
	select 2 as bbb
)
select 
	aaa 
from 
	test_1 
where 
	aaa in (select bbb from test_2) -- errors appear here
;
```

DBeaver reports two errors: 

Column bbb not found.
Table test_2 not found.

### DBeaver Version

Community Version 25.0.5.202505181758

### Operating System

Windows 10

### Database and driver

PostgreSQL 14.15 on aarch64-unknown-linux-gnu, compiled by aarch64-unknown-linux-gnu-gcc (GCC) 9.5.0, 64-bit

### Steps to reproduce

Execute the toy example above.

### Additional context

Recently, I've noticed a consistent stutter when working with long chains of CTEs, about one second of lag for every 3-5 seconds of typing.

I suspect that it's the result of an overzealous SQL parser finding bugs where there are none.

E1izabeth:
Validation error should not happen

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/context/SQLQueryRowsSourceContext.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/context/SQLQueryRowsSourceContext.java
@@ -222,6 +222,11 @@ public class SQLQueryRowsSourceContext {
         });
     }
 
+    @NotNull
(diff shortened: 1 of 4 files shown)


ShadelessFox (approved):

IvanGorshechnikov (approved):

uslss:
verified

PR #38393, opened 2025-06-16
dbeaver/dbeaver#37016 Fix generated formatted script

### LINKED ISSUE #37016: Virtual key updates generate incorrect code and send different code when saved from data grid
### Description

When updating a table which has no keys using the data grid Dbeaver will generate an incorrect update query and will send a different query to the server when the save button is pressed.

### DBeaver Version

24.3.4

### Operating System

macOS 15.3

### Database and driver

MariaDB 11.1 JDBC 3.5.1

### Steps to reproduce

1. Create a table and populate it like so:
```
create table updatetest (
col1 varchar(10),
col2 varchar(10),
col3 varchar(10),
col4 varchar(10));

insert into  updatetest (col1, col2, col3, col4)
values ('A','B',null, null),
('A',null,'C','D'),
('null','B',null,'D');
```

2. Check the contents of the table and open a data grid:

`select * from updatetest;`

Image

3. Open the Query Manager in DBeaver
4. On row 1, change the [NULL] in col3 to 'C' and view the generated query. If asked, select to use all of the columns as the virtual key:

```
--  Auto-generated SQL script #202502031038
UPDATE updatetest
	SET col3='C'
	WHERE col1='A' AND col2='B' AND col3 IS NULL AND col4 IS NULL;
```

This query looks correct and will update the value in row1 col3 if executed. Execute or save the query and note the query in the Query Manager matches what was shown in the generated query

5. On row 2, change the [NULL] in col2 to 'B' and view the generated query:

```
--  Auto-generated SQL script #202502031040
UPDATE updatetest
	SET col2='B'
	WHERE col1='A' AND col2 IS NULL AND col3=NULL AND col4='C';
```

Note that `col3=NULL` is incorrect SQL (it should be IS NULL) and it is incorrect logically as row 2 has a 'C' for col3. and 'D' for col4.

6. Click execute and notice the query in the Query Manager *does not* match the generated query which we clicked execute on. This query is correct

```
UPDATE updatetest
	SET col2='B'
	WHERE col1='A' AND col2 IS NULL AND col3='C' AND col4='D'
```
7. Similarly, update col1 on row 3 and see what happens in both the generated query:

```
--  Auto-generated SQL script #202502031044
UPDATE updatetest
	SET col1='A'
	WHERE col1 IS NULL AND col2=NULL AND col3 IS NULL AND col4='B';
```

This is also wrong as col4 in row 3 is 'D' and not 'B' . But, again, the actually executed query is correct:

```
UPDATE updatetest
	SET col1='A'
	WHERE col1 IS NULL AND col2='B' AND col3 IS NULL AND col4='D'
```

### Additional context

This is a pretty severe bug in my opinion as the generated query could be saved, not checked for validity, and run at a later time, potentially causing data loss

IvanGorshechnikov:


E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteBatchImpl.java
```diff
@@ -193,7 +193,12 @@ DBCStatistics processBatch(@NotNull DBCSession session, @Nullable List<DBEPersis
                             }
                         }
                     } else {
-                        String queryString = formatQueryParameters(session, statement.getQueryString(), handlers, rowValues);
+                        String queryString;
+                        if (statement instanceof DBCParameterizedStatement parameterizedStatement) {
+                            queryString = parameterizedStatement.getFormattedQuery();
```
serge-rider: parameterizedStatement.getFormattedQuery() uses a silly algorithm. It is not always correct.
Now we switch to it always. Are you sure it is a good solution? Probably DBCParameterizedStatement  should be used for virtual keys only

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteBatchImpl.java
```diff
@@ -227,7 +224,11 @@ DBCStatistics processBatch(@NotNull DBCSession session, @Nullable List<DBEPersis
     }
 
     protected int getNextUsedParamIndex(Object[] attributeValues, int paramIndex) {
-        return paramIndex + 1;
+        paramIndex++;
+        while (paramIndex < attributeValues.length && attributeValues[paramIndex] instanceof DBDNull) {
```
serge-rider, line 229: Very weird logic. It will stop at first non-null?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/DBDNull.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    @Override
+    public void release() {
+
+    }
+
+    @Override
+    public String toString() {
+        return "";
```
serge-rider: "NULL"?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/data/ExecuteBatchImpl.java
serge-rider: PErhaps we need some comment here

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetPersister.java
```diff
@@ -895,7 +903,11 @@ private Throwable executeStatements(DBCSession session) {
                                 attributes[i] = statement.updateAttributes.get(i).getValue();
                             }
                             for (int i = 0; i < statement.keyAttributes.size(); i++) {
-                                attributes[statement.updateAttributes.size() + i] = statement.keyAttributes.get(i).getValue();
+                                if (DBUtils.isNullValue(statement.keyAttributes.get(i).getValue())) {
+                                    attributes[statement.updateAttributes.size() + i] = DBDNull.INSTANCE;
+                                } else {
+                                    attributes[statement.updateAttributes.size() + i] = statement.keyAttributes.get(i).getValue();
+                                }
```
E1izabeth: It would be nice to move this part to a separate method, because it's duplicated

serge-rider: +1

PR #38396, opened 2025-06-16
dbeaver/pro#5547 Add force restart while accepting EAP

IvanGorshechnikov:


plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/ActionUtils.java
```diff
@@ -496,4 +497,26 @@ public static String getLabelWithIndexMnemonic(@NotNull String label, int index)
             return label;
         }
     }
+
+    public static boolean restartForced() throws Throwable {
```
ShadelessFox: Well.. you return a `boolean` but also throw exceptions. Please stick with just one thing. I suggest the former.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/ActionUtils.java
ShadelessFox: Handle method invocation errors, log, and return `false`.

serge-rider: Also this method never called in CE

serge-rider (changes requested):

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/ApplicationWorkbenchAdvisor.java
```diff
@@ -178,7 +178,8 @@ public class ApplicationWorkbenchAdvisor extends IDEWorkbenchAdvisor {
             ApplicationWorkbenchWindowAdvisor.TREE_AND_TABLE_FONT_FOR_VIEWS
         )
     ); 
-    
+
+    private static boolean forcedRestart = false;
```
ShadelessFox: Not used?

IvanGorshechnikov: What do you mean?

E1izabeth: This field is unused. At least I can't find any references in PR changes to it. If it's so, let's remove this change.

IvanGorshechnikov: Removed

E1izabeth (dismissed):

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPApplication.java
```diff
@@ -151,6 +151,13 @@ default boolean isStatisticsCollectionRequired() {
         return false;
     }
 
+    /**
+     * Determines whether the application has been marked for a forced restart.
+     */
+    default boolean isForcedRestart() {
```
serge-rider: We don't need this function in interface. Move flag and getter/setter into ApplicationWorkbenchAdvisor.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/ApplicationWorkbenchAdvisor.java
```diff
@@ -177,7 +177,8 @@ public class ApplicationWorkbenchAdvisor extends IDEWorkbenchAdvisor {
             ApplicationWorkbenchWindowAdvisor.PART_TITLE_FONT,
             ApplicationWorkbenchWindowAdvisor.TREE_AND_TABLE_FONT_FOR_VIEWS
         )
-    ); 
+    );
+    public static boolean isForcedRestart = false;
```
ShadelessFox: Probably it shouldn't be public

PR #38401, opened 2025-06-17
dbeaver/pro#5567 Fix ai smart metadata description

labanovichttps:


plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/DatabaseEditorUtils.java
```diff
@@ -102,6 +103,12 @@ public static void contributeStandardEditorActions(IWorkbenchSite workbenchSite,
             EditorsMessages.database_editor_command_revert_tip,
             true));
         contributionManager.add(new ToolbarSeparatorContribution(true));
+
+        contributionManager.add(new Separator());
+        contributionManager.add(ActionUtils.makeCommandContribution(
```
serge-rider: We cannot add PRO commands in CE.
Moreover for all DB editors. It must be only in entity editors.  There must be another way to contribute it in PRO.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -2562,4 +2562,24 @@ public static boolean isUniqueIndexForAttributes(
         }
         return false;
     }
+
+    /**
+     * Retrieves the schema name associated with the provided database object.
+     */
+    @Nullable
+    public static String getObjectSchema(@NotNull DBSObject dbsObject) {
```
serge-rider: 1. Why only schema? WE must check catalogs too.
2. Move it to DBStructUtils

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIOperationType.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.ai;
+
+public enum AIOperationType {
```
serge-rider: We cannot have it CE model. Why do we need it at all?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIOperationType.java
labanovichttps: Because when I try to describe any object, it adds the default system prompt, and AI cannot consume the wrong context

E1izabeth: I don't see where it's used in this PR -_-

serge-rider: We still cannot use such a constants

serge-rider (changes requested):

IvanGorshechnikov (dismissed):

PR #38421, opened 2025-06-19
dbeaver/dbeaver#34851 Switched new connection default type to non-predefined

### LINKED ISSUE #34851: Can't delete new custom Connection Type until application restart
### Description

The button Delete is disabled for the newly created connection type, until you restart the DBeaver

https://github.com/user-attachments/assets/f4bcfe89-3bb2-4b8e-9c4d-0268fec38181

### DBeaver Version

24.1.3

### Operating System

win11

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

Nexus6v2:
Because the default new connection type was set to predefined, remove-connection button was disabled until the restart.


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/connection/DBPConnectionType.java
```diff
@@ -91,7 +91,10 @@ public class DBPConnectionType implements DBPDataSourcePermissionOwner {
             null); //$NON-NLS-1$ //$NON-NLS-3$
 
         SYSTEM_TYPES = new DBPConnectionType[] { DEV, TEST, PROD };
-        DEFAULT_TYPE = DEV;
+
+        DBPConnectionType defaultType = new DBPConnectionType(DEV);
```
IvanGorshechnikov, line 95: What is the goal of recreating DEV connection type?

Nexus6v2: To change the `predefined` field in `defaultType` without changing `DEV` type.
Basically copy on update.

IvanGorshechnikov (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/connection/DBPConnectionType.java
```diff
@@ -91,7 +91,10 @@ public class DBPConnectionType implements DBPDataSourcePermissionOwner {
...
+        defaultType.predefined = false;
```
serge-rider, line 96: It makes default type (DEV) not predefined. Thus it is possible to delete it. What is the idea? What if user will try to delete DEV?

Nexus6v2: It copies all the properties from DEV type and sets the predefined to false.
The object itself is different than DEV, that's why I used constructor instead of just setting `DEV.predifined = false`

Matvey16:
Verified

PR #38427, opened 2025-06-19
dbeaver/pro#6020 Fix lm stage mode

IvanGorshechnikov:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPApplication.java
```diff
@@ -67,6 +67,8 @@ default void beforeWorkspaceInitialization() {}
 
     boolean isDetachedProcess();
 
+    default boolean isStageMode() {return false;};
```
serge-rider: Let's move to pro

serge-rider (changes requested):

PR #38428, opened 2025-06-19
dbeaver/pro#6048 Fix postgres url

labanovichttps:


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
```diff
@@ -1025,7 +1025,7 @@ public static String getDatabaseNameFromConfiguration(DBPConnectionConfiguration
             }
         } else {
             String url = configuration.getUrl();
-            Pattern pattern = Pattern.compile("jdbc:postgresql://[^/]+/([^?#]+)");
+            Pattern pattern = Pattern.compile(".*://[^/]+/([^?#]+)");
```
E1izabeth: `.*` is greedy, are you sure?

E1izabeth: Also use string boundary anchors

labanovichttps: I thought about jdbc and odbc and maybe `Pattern pattern = Pattern.compile("^(?:jdbc|odbc).*://[^/]+/([^?#]+)");`
Is better solution.

E1izabeth: Unsure. However, it won't solve the greedy issue anyway

E1izabeth (approved):

PR #38434, opened 2025-06-20
dbeaver/pro#5399 Add new completion engine to cloudbeaver

HocKu7:
Closes https://github.com/dbeaver/pro/issues/5399


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/completion/SQLQueryCompletionAnalyzer.java
```diff
@@ -192,8 +195,22 @@ public Integer getActualContextOffset() {
         return this.result.get().getFirst();
     }
 
-    @Nullable
+    @NotNull
     protected DBPImage prepareProposalImage(@NotNull SQLQueryCompletionItem item) {
-        return null;
+        return switch (item.getKind()) {
+            case SCHEMA, CATALOG, UNKNOWN ->  DBValueFormatting.getObjectImage(item.getObject());
+            case RESERVED -> null; //todo
```
E1izabeth: We can't leave it without an image. Removing the value would be a regression, wouldn't it?
P.S. There was `UIIcon.SQL_TEXT`

E1izabeth (changes requested):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/completion/ISqlCompletionProposal.java
```diff
@@ -0,0 +1,44 @@
...
+import org.jkiss.dbeaver.model.DBPImage;
+import org.jkiss.dbeaver.model.DBPKeywordType;
+
+/**
+ * Unified interface for SQL completion proposal models.
+ * This interface is used to unify the handling of completion proposals
+ * from both the legacy and the new completion engines.
+ * It allows processing of completion proposals regardless of their implementation.
+ */
+
```
E1izabeth: Accidental empty line

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/completion/SQLQueryCompletionAnalyzer.java
E1izabeth: I'm sure we shouldn't move UI-related stuff to a non-UI-related class.
You can move this to some common utility class and left it for DBeaver as is, but make for Cloudbeaver whatever you want

E1izabeth (dismissed):

PR #38442, opened 2025-06-23
#17627 #38336 Fix connection by URL

### LINKED ISSUE #17627: jdbc:db2://server:port/database:sslConnection=true; results in No suitable driver was found for specified JDBC URL
### Description

Typing :sslConnection=true at the end of JDBC URL connection and there is not possible to select a driver.

### DBeaver Version

Community Edition 22.2.0.202209040629

### Operating System

Windows 10

### Database and driver

1. Database name and version: DB2 v11.5.7.0
2. Driver name version: Db2 v11.4 M7 FP0 version 4.31.10

### Steps to reproduce

1. From **Database | New Connection from JDBC URL**.
2. In **JDBC URL** field type in:` jdbc:db2://localhost:60227/test1` and in Drivers filed it appears driver selection. It works as expected.
image
3. Now at the end of **JDBC URL** field add `:sslConnection=true;` so connecting string becomes: `jdbc:db2://localhost:60227/test1:sslConnection=true;` but now suddenly error appears "No suitable driver was found for specified JDBC URL". This is the problem. I expect in Drivers field I could still select e.g. DB2 LUW driver (just like image above).
image

### Additional context

_No response_

### LINKED ISSUE #38336: Connection with URL still uses portnumber
### Description

When using a URL connection DBeaver actually still uses the disabled (grayed out) portnumber. So when the setting is like in the screen cap below it will try to connect with port 7432 and not with the one in the URL (in this example 6432). 

Image

It should of course use the URL and not the disabled portnumber. 

### DBeaver Version

25.1.0.202506011722

### Operating System

Win11

### Database and driver

1. Database Postgres 16.2
2. Postgres JDBC Driver 17.0

### Steps to reproduce

**Reproduction steps**

1. Simply copy an existing connection. 
2. Set connection to URL (if that is not already done)
3. Change the port in the URL (see green arrow in the screencap below)

You will see that the disabled portnumber field still shows the old portnumber (see red arrow in screencap below). 

Image

When connecting to the DB it will use that disabled portnumber and not the portnumber in de URL. 

**Desired behaviour**

When connecting with an URL it should only use the URL and not the disabled portnumber. 

### Additional context

n.a.

E1izabeth:
Fixes #17627
Fixes #38336


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -2562,4 +2570,89 @@ public static boolean isUniqueIndexForAttributes(
         }
         return false;
     }
+
```
serge-rider, line 2565: Move it to DBWUtils. 
DBUtils is too huge already

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -2562,4 +2570,89 @@ public static boolean isUniqueIndexForAttributes(
...
+                    if (urlConfiguration == null) {
+                        urlConfiguration = DatabaseURL.extractConfigurationFromUrl(DatabaseURL.GENERIC_URL_TEMPLATE, activeUrl);
+                        if (urlConfiguration != null) {
+                            metaURL = DatabaseURL.parseSampleURL(DatabaseURL.GENERIC_URL_TEMPLATE);
+                        }
+                    }
+                    if (urlConfiguration != null) {
+                        urlConnectivityParams = getExplicitConnectivityParameters(urlConfiguration);
+                    }
+                    if (urlConnectivityParams == null) {
```
serge-rider: Add some description in method.
Which explain the logic and flow.

Also - it looks that this function is used in PostgreSQL only. Maybe move it there?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
```diff
@@ -1015,6 +1015,9 @@ private static Number parseNumber(@NotNull String str) {
         }
     }
 
+    /**
+     * TODO consider using {@link org.jkiss.dbeaver.model.DBUtils#getConnectivityParameters(DBPConnectionConfiguration, DBPDriver)}.
```
ShadelessFox: Is this a deprecation warning, a suggestion, or a todo?

E1izabeth: It's a todo

E1izabeth: I'm going to figure out use cases when this method is used, describe them and then replace this method with a new one. But it would be a separate issue

ShadelessFox: I mean it's a Javadoc for some reason. If it's a TODO, make it a regular comment instead

ShadelessFox (approved):

PR #38453, opened 2025-06-23
#34424 Change active object panel text for result set viewer for db entities

### LINKED ISSUE #34424: [User experience story] SQL Editor with query from Data Editor
User was trying to save a query, that is executed to refresh the table data with custom filters and orderings.
What they tried to do:
1. Put the focus to the filter text field and press `Ctrl+S` - didn't help
2. Then they open main menu and look for keywords like scripts / saved queries / etc
image
3. Found something relevant here. But none of the buttons gave the desired result - "to move the query built in data editor to sql script".
image
Also user expected to see opened scripts history for Recent SQL Script button like this:
image

4. Then user created new script, wrote the query by themself, saved, then closed.
5. After that, user tried to open saved script again. They pressed `Open script button` and haven't found the script there _(because we show only scripts for selected connection here, but checkbox to show all scripts seems was still not obvious enough)_
image
6. After a long searching user has found Project Resources, wondered what the "Project" is and was happy, because they've finally found the script.
_this thing:_
image

_____________________________________________________________________________
Also here is the idea of how not to loose your script under the Scripts folder, if you have toooo many of them for different connections. It's like our navigator, but with scripts instead of database objects under the database node.
My Queries > Connection Name > DB Name > Query Name
image

E1izabeth:
Shows query text for queries and "Show SQL" for tables/views/etc.
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetFilterPanel.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetFilterPanel.java
@@ -451,8 +451,8 @@ class ResultSetFilterPanel extends Composite implements IContentProposalProvider
     private String getActiveSourceQuery() {
         String displayName;
         DBSDataContainer dataContainer = viewer.getDataContainer();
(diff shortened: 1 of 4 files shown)


ShadelessFox (approved):

uslss:
verified

PR #38458, opened 2025-06-24
dbeaver/dbeaver#38457 Support additional variables in virt columns

### LINKED ISSUE #38457: Support additional variables for virtual columns manipulation
Currently we can only refer another column names in virtual column expressions.
Moreover we cannot refer columns with spaces or other special characters in their names.

Let's
- Add variables `row` which will work as a map "column name->column value"
- Add variable `table` with fields `name`, `schema`, `container`

So we could use expressions like this:
```
  row[`Column with spaces in name`] + ' ' + table.name
```

serge-rider:


Matvey16:
`row` expression doesn't support any quoting except `"` regardless of the database, also it doesn't support column names without quotes, even if it doen't have them in the original DDL

Matvey16:
Verified

PR #38459, opened 2025-06-24
dbeaver/dbeaver#38450 Connections driver change fix

### LINKED ISSUE #38450: "Change Connection(s) Driver" wizard doesn't work from Projects panel
### Description

While converting a number of my database connections, I initially thought that the driver-changing wizard accessed via `Database > Change Connection(s) Driver` was completely broken. I could select a database connection in my Projects panel, and then open the "Change Connection(s) Driver" wizard window, but then no matter what database driver I selected, neither the "Next" nor "Finish" buttons would become enabled.

Only while doing additional testing for this bug report did I discover that only database connections selected in your "Database Navigator" panel allow the wizard to work properly. If you select one from there and open the wizard, then the "Finish" button is enabled and allows the wizard to do the driver migration.

The reason I believe this is a bug is two-fold: A) With a connection selected in the Projects panel, the wizard opens normally as though it is going to allow you to migrate the driver, with no error messages shown. B) The "Change Connection(s) Driver" menu option is configured to only appear in the menu when a database connection is selected, and this works for both ones in the Projects and Database Navigator panels. Selecting anything else in those panels or in any other panels causes the menu item to not appear.

### DBeaver Version

25.1.1

### Operating System

Windows Server 2019

### Database and driver

N/A

### Steps to reproduce

1. In your Projects panel, expand a project and expand its Connections.
2. Select a connection from the Connections list.
3. Select `Database > Change Connection(s) Driver` from the main app menu.
4. Select any driver to migrate to, and observe that the "Next" and "Finish" buttons do not activate/enable.

### Additional context

_No response_

serge-rider:


uslss:
verified

PR #38460, opened 2025-06-24
dbeaver/dbeaver#38456 Move misplaced comma before comment in DDL

### LINKED ISSUE #38456: 25.1.1 Postgres DDL generation puts comments before comma
### Description

25.1.1 is generating `col-def -- comment ,` so the comma is part of the comment
25.1.0 and earlier generate `col-def , -- comment`

### DBeaver Version

CE 25.1.1

### Operating System

_No response_

### Database and driver

PostgreSQL 14.9

### Steps to reproduce

_No response_

### Additional context

_No response_

vkirst:
Fix DDL generation: incorrect comma placement after inline column comment


uslss:
verified

PR #38466, opened 2025-06-24
#38189 Show MariaDB native clients only for MariaDB connections

### LINKED ISSUE #38189: I can't make database dump on mysql database
### Description

when i try to make a dump of my database (mysql), a erro happen on executation
Image
as you can see, dbeever try to execute a mariadb-dump, but my database is mysql.

### DBeaver Version

Community Edition 25.0.5 (snap)

### Operating System

Ubuntu 24.04.2 (WSL)

### Database and driver

MySQL 8.0.42 - Connector J - 8.2.0

### Steps to reproduce

I connect to a database;
Select database dump tools on it;
Select dump options on next windows; (every option as default)
Click on next an start.
The error occurs.

### Additional context

_No response_

ShadelessFox:
The MariaDB client must only appear on MariaDB connections. MySQL clients will still appear on MariaDB connections as well.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
@@ -311,6 +311,7 @@
             <nativeClients>
                 <client id="mariadb_client" label="MariaDB Binaries">
                     <dist os="win32" targetPath="clients/mariadb/win" remotePath="repo:/drivers/mariadb/client/win" resourcePath="clients/mariadb/win">
+                        <driver id="mariaDB"/>
                         <file type="exec" name="mariadb.exe"/>
                         <file type="exec" name="mariadb-dump.exe"/>
                     </dist>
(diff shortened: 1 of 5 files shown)


plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
```diff
@@ -311,6 +311,7 @@
             <nativeClients>
                 <client id="mariadb_client" label="MariaDB Binaries">
                     <dist os="win32" targetPath="clients/mariadb/win" remotePath="repo:/drivers/mariadb/client/win" resourcePath="clients/mariadb/win">
+                        <driver id="mariaDB"/>
```
E1izabeth, line 314: Don't we need to add similar line for MySQL?

ShadelessFox: No; it doesn't work the opposite way

ShadelessFox: MySQL is kinda complicated because it has about five different drivers. I wasn't sure if that would be ideal.

IvanGorshechnikov (approved):

E1izabeth (approved):

Matvey16:
Verified

Matvey16:
Verified

gabrielrsep:
Hello guys, i don't know if now is problem with my setup, but i update to 25.1.2 and the error persists

PR #38479, opened 2025-06-25
dbeaver/pro#3596 Ignore case for column and cte aliases, fix aliases visibility scopes for BigQuery

E1izabeth:
1. Aliases can be used in a case-insensitive manner everywhere in a query (for all databases).
2. In BigQuery, the column alias is not visible in WHERE, but is visible in other projection clauses (GROUP BY, HAVING, ORDER BY). Query validation should reflect it correctly. If the query can't be executed, the validation error should be shown.


labanovichttps (approved):

IvanGorshechnikov (approved):

uslss:
verified

PR #38520, opened 2025-06-30
dbeaver/pro#5987 profiles enhancements

ShadelessFox:
Fixes the first and second bullet points. The third one seems expected.

java_h6Pw1Yc6a8


IvanGorshechnikov (approved):

Matvey16:
Verified

PR #38542, opened 2025-07-02
dbeaver/dbeaver#38008 Configure session settings in connection proper…

### LINKED ISSUE #38008: Running ClickHouse script seems to runs each line in its own transaction\session
### Description

Hi,
I'm running a script that uses temp tables.

As an example:

create temporary table xpto as (select 1);
select * from xpto;

When I run this using ALT+X (after a CTRL+A) I always get the error:

SQL Error [22000]: Code: 60. DB::Exception: Unknown table expression identifier 'xpto' in scope SELECT * FROM xpto. (UNKNOWN_TABLE) (version 25.3.3.42 (official build)) 

It seems that the 1st line in the query is run, and then the 2nd is run in a different context, unaware of the temp table that was created.

### DBeaver Version

25.0.4

### Operating System

Windows 11

### Database and driver

Using 
Database: Clickhouse version 25.3.3
Driver: com.clickhouse | clickhouse-jdbc | 0.8.5

### Steps to reproduce

Connect to clickhouse database
Copy and paste:

create temporary table xpto as (select 1);
select * from xpto;

press: CTRL+A followed by ALT+X

### Additional context

_No response_

vkirst:


plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseDataSource.java
```diff
@@ -162,6 +165,11 @@ private void initSSL(DBRProgressMonitor monitor, Properties properties, DBWHandl
         }
     }
 
+    private void configureSession(@NotNull Properties properties) {
+        properties.put(ClickhouseConstants.CLICKHOUSE_SETTING_SESSION_ID, "sess_" + UUID.randomUUID());
+//        properties.put(ClickhouseConstants.CLICKHOUSE_SETTING_SESSION_TIMEOUT, "60");
```
ShadelessFox: No commented-out code please

ShadelessFox (changes requested):

E1izabeth (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #38561, opened 2025-07-04
#10235 Colorize editor tabs according to their connection's color

### LINKED ISSUE #10235: Colored tabs or Tab division by connection.
It is too hard to indistinguishable many tabs.

Colored tabs / Tab division by connection.

I saw this requests at past, but it's been a long time since then.
Is there a solution now?

https://dbeaver.io/forum/viewtopic.php?f=3&t=916
https://github.com/dbeaver/dbeaver/issues/300
https://github.com/dbeaver/dbeaver/issues/1777

ShadelessFox:
image

Doesn't work against the "Classic" theme.


plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/DatabaseEditorInput.java
```diff
@@ -192,6 +196,17 @@ public String getDefaultFolderId()
...
+    @Nullable
+    @Override
+    public Color getConnectionColor() {
+        DBPDataSourceContainer container = getDataSourceContainer();
+        if (container != null) {
+            return UIUtils.getConnectionColor(container.getConnectionConfiguration());
+        } else {
+            return null;
+        }
+    }
```
E1izabeth, line 208: It looks exactly the same as https://github.com/dbeaver/dbeaver-ee/pull/3728/files#diff-784bb51d2aed996b869beadc48d1afdc9ecee2fcc8401f6691fb70848d66f007R105-R114
Can we avoid duplication?

ShadelessFox: Either by providing a default implementation right in the interface, which is not ideal, as not all implementations can rely on it, or by introducing a utility function to which those two implementations will delegate. Do you think it's worth it?

E1izabeth: I thought about moving it up, but if it can't be applied to all implementations, then OK, let it be as it's not.

E1izabeth (approved):

Matvey16:
Verified

PR #38586, opened 2025-07-08
dbeaver/pro#5520 Fallback to minimal DDL generation

ivgag:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/registry/AIDdlGeneratorRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+import org.eclipse.core.runtime.IConfigurationElement;
+import org.eclipse.core.runtime.Platform;
+import org.jkiss.dbeaver.DBException;
+import org.jkiss.dbeaver.model.ai.AIDdlGenerator;
+import org.jkiss.dbeaver.registry.RegistryConstants;
+
+/**
+ * Registry for AI-based DDL generators.
+ */
+public class AIDdlGeneratorRegistry
```
serge-rider: Let's rename Ddl into DatabaseSchema. ddl is quite confusing.

serge-rider (changes requested):

PR #38593, opened 2025-07-09
dbeaver/pro#6172 Trim profile name

labanovichttps:
before 
image

and after 
image


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageNetworkHandler.java
```diff
@@ -160,7 +160,10 @@ public void refreshConfiguration(@Nullable DBWNetworkProfile profile) {
                 configuratorEnableState = ControlEnableState.disable(configuratorPlaceholder);
             }
 
-            profileProvidedHint.setText(NLS.bind("Using configuration from profile ''<a href=\"#\">{0}</a>''", profile.getProfileName()));
+            profileProvidedHint.setText(NLS.bind(
+                "Using configuration from profile ''<a href=\"#\">{0}</a>''",
+                profile.getProfileName().trim()
```
E1izabeth: Shouldn't we also cut the profile name if it's too long?

ShadelessFox: I don't think so, the UI adapts nicely anyway

ShadelessFox: Again, you shouldn't trim it here; I believe you're the only one who has ever encountered it.

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageSettings.java
```diff
@@ -467,7 +467,7 @@ private void updateHandlerItem(@NotNull List<IDialogPage> allPages) {
     private void updateProfileItem() {
         String profileName = getActiveDataSource().getConnectionConfiguration().getConfigProfileName();
         if (CommonUtils.isNotEmpty(profileName)) {
-            profileItem.setText(NLS.bind("Profile ''{0}''", profileName));
+            profileItem.setText(NLS.bind("Profile ''{0}''", profileName.trim()));
```
ShadelessFox: And here...

ShadelessFox (commented):
My personal idea was to trim the name when you **create** a new profile. Eliminate the issue before it even becomes an issue.

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/preferences/PrefPageProjectNetworkProfiles.java
```diff
@@ -144,9 +144,9 @@ protected DBWNetworkProfile createNewProfile(@Nullable DBWNetworkProfile sourceP
                 getShell(),
                 UIConnectionMessages.pref_page_network_profiles_tool_create_dialog_profile_name,
                 profileName
-            );
+            ).trim();
```
ShadelessFox: It can and will return `null` if the user pressed "Cancel".

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/preferences/PrefPageProjectNetworkProfiles.java
labanovichttps: fixed

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/preferences/PrefPageProjectNetworkProfiles.java
```diff
@@ -144,12 +144,16 @@ protected DBWNetworkProfile createNewProfile(@Nullable DBWNetworkProfile sourceP
...
-            ).trim();
+            );
 
             if (CommonUtils.isEmpty(profileName)) {
                 return null;
             }
 
+            if (CommonUtils.isNotEmpty(profileName)) {
+                profileName = profileName.trim();
+            }
```
ShadelessFox: `CommonUtils.isEmpty` already checks for null, so you can remove the if entirely.

```suggestion
            profileName = profileName.trim();
```

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/preferences/PrefPageProjectNetworkProfiles.java
```diff
@@ -146,10 +146,14 @@ protected DBWNetworkProfile createNewProfile(@Nullable DBWNetworkProfile sourceP
                 profileName
             );
 
-            if (CommonUtils.isEmptyTrimmed(profileName)) {
+            if (CommonUtils.isEmpty(profileName)) {
```
ShadelessFox: Bring `isEmptyTrimmed` back - otherwise, profiles with spaces only will pass through and get created

ShadelessFox (changes requested):

ShadelessFox (approved):

E1izabeth (approved):

Matvey16:
Verified

PR #38594, opened 2025-07-09
dbeaver/pro#6100 Refactoring postgre isPgObject

labanovichttps:
Try to make some things with postgres, redshift and Edb


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
```diff
@@ -1058,8 +1050,48 @@ private static int getLastNonDatabaseCharPos(int divPos, String url) {
                 break;
             }
         }
-        if (lastPos < 0) lastPos = url.length();
+        if (lastPos < 0)
+            lastPos = url.length();
```
E1izabeth:
```suggestion
        if (lastPos < 0) {
            lastPos = url.length();
        }
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -931,4 +931,12 @@ public boolean isSupportsEnumTable() {
     public boolean isSupportsReltypeColumn() {
         return supportsReltypeColumn;
     }
+
+    public boolean isPGObject(Object object) {
```
E1izabeth:
```suggestion
    public boolean isPGObject(@Nullable Object object) {
```

E1izabeth: What's the point of moving it to PostgreDataSource class? I don't see any reason

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
```diff
@@ -1058,8 +1050,48 @@ private static int getLastNonDatabaseCharPos(int divPos, String url) {
...
         return lastPos;
     }
 
-}
+    @Nullable
+    public static PostgreDataSource getPostgreDataSource(@Nullable Object object) {
```
ShadelessFox: DBSObject should be enough. Everything else is not needed

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
ShadelessFox: Correct - it seems pointless, as most PG extensions rely on `PostgreServerExtension`.

ShadelessFox (changes requested):
Please move `isPgObject` to the `PostgreServerExtension` interface. Most extensions don't extend the data source but implement the extension interface instead.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreServerExtension.java
```diff
@@ -251,4 +252,12 @@ public interface PostgreServerExtension {
      */
     boolean supportsNativeClient();
 
+    default boolean isPGObject(Object object) {
```
ShadelessFox: Annotations and javadoc :^)

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
```diff
@@ -1058,8 +1050,48 @@ private static int getLastNonDatabaseCharPos(int divPos, String url) {
...
         }
-        if (lastPos < 0) lastPos = url.length();
+        if (lastPos < 0)
+            lastPos = url.length();
         return lastPos;
     }
 
-}
+    @Nullable
+    public static PostgreServerExtension getPostgreServerExtension(@Nullable Object object) {
```
ShadelessFox: Why. Just pass `DBSObject`, it should be enough

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreGeometryValueHandler.java
```diff
@@ -106,6 +106,8 @@ public Class<?> getValueObjectType(@NotNull DBSTypedObject attribute) {
 
     @Override
     public Object getValueFromObject(@NotNull DBCSession session, @NotNull DBSTypedObject type, Object object, boolean copy, boolean validateValue) throws DBCException {
+        PostgreServerExtension postgreDataSource = PostgreUtils.getPostgreServerExtension(object);
+        boolean isPgObject = postgreDataSource != null && postgreDataSource.isPGObject(object);
```
ShadelessFox: You should preserve `PostgreUtils.isPGObject` but make it also accept a `DBSObject` and delegate it to the extension's `isPgObject` method

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreServerExtension.java
ShadelessFox: I suggest making it `@NotNull` and do an early return in `PostgreUtils`

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreServerExtension.java
```diff
@@ -251,4 +253,16 @@ public interface PostgreServerExtension {
      */
     boolean supportsNativeClient();
 
+
+    /**
+     * Determines if the provided object is a PostgreSQL-specific object (PGObject) like com.amazon.redshift.util.RedshiftObject.
+     */
+    default boolean isPGObject(@Nullable Object object) {
```
serge-rider: LEt's move it to PostgreServerExtensionBase

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
```diff
@@ -223,8 +215,8 @@ public static long[] getIdVector(Object pgObject) {
         }
     }
 
-    public static int[] getIntVector(Object pgObject) {
-        Object pgVector = extractPGObjectValue(pgObject);
+    public static int[] getIntVector(@Nullable Object pgObject, @Nullable DBPDataSource dataSource) {
```
serge-rider: How can dataSource be null?
It shouldn't because objects exist only within a live connection.,

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/PostgreServerEdb.java
```diff
@@ -86,5 +87,14 @@ protected boolean addExtraCondition(@NotNull JDBCSession session, @NotNull Strin
...
         }
     }
+
+    @Override
+    public  boolean isPGObject(Object object) {
+        if (object == null) {
+            return false;
+        }
+        String className = object.getClass().getName();
+        return className.equals(PostgreConstants.EDB_OBJECT_CLASS);
```
serge-rider: Let's move EDB_OBJECT_CLASS into EDBConstants. The same about Redshift.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreServerExtension.java
labanovichttps: We cannot, because `public PostgreServerExtension getServerType()`

serge-rider: I mean do not use default implementation in interface but move it in PostgreServerExtensionBase

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
```diff
@@ -120,21 +121,12 @@ public static <T extends PostgreAttribute> T getAttributeByNum(Collection<T> att
...
-    }
-
-    public static Object extractPGObjectValue(Object pgObject) {
+    @Nullable
+    public static Object extractPGObjectValue(@Nullable Object pgObject, @Nullable DBPDataSource dataSource) {
         if (pgObject == null) {
             return null;
         }
-        if (!isPGObject(pgObject)) {
+        if (!isPgObject(dataSource, pgObject)) {
```
E1izabeth, line 129:
```suggestion
        if (!isPgObject(dataSource, pgObject)) {
```

ShadelessFox: What?

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
```diff
@@ -120,21 +121,12 @@ public static <T extends PostgreAttribute> T getAttributeByNum(Collection<T> att
...
             return pgObject;
         }
```
E1izabeth, line 131: I don't see a reason for creating a utility method here, especially when it checks for ins
```java
 if (dataSource instanceof PostgreDataSource postgreDataSource
     && !postgreDataSource.getServerType().isPGObject(pgObject)
 ) {
            return pgObject;
 }
```

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreServerExtension.java
```diff
@@ -251,4 +252,8 @@ public interface PostgreServerExtension {
      */
     boolean supportsNativeClient();
 
+    /**
+     * Determines if the provided object is a PostgreSQL-specific object (PGObject) like com.amazon.redshift.util.RedshiftObject.
+     */
+    boolean isPGObject(@Nullable Object object);
```
ShadelessFox: I still believe the `object` must be marked as `@NotNull`. We don't call it directly anywhere but from the utility function that can do that check itself. This will simplify all existing implementations.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/PostgreServerEdb.java
```diff
@@ -86,5 +87,14 @@ protected boolean addExtraCondition(@NotNull JDBCSession session, @NotNull Strin
             return false;
         }
     }
+
+    @Override
+    public  boolean isPGObject(Object object) {
```
ShadelessFox: Annotations

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/impls/redshift/PostgreServerRedshift.java
```diff
@@ -492,4 +494,14 @@ public boolean supportsAlterTableForViewRename() {
     public boolean supportsNativeClient() {
         return false;
     }
+
+
+    @Override
+    public boolean isPGObject(Object object) {
```
ShadelessFox: Annotations

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreServerExtension.java
ShadelessFox: I just checked that the utility function already performs the null check. So please just mark it here as `@NotNull`.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
labanovichttps: pgObject will never be a postgreDataSource, `pgObject instanceof PostgreDataSource postgreDataSource` is always falls

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreServerExtension.java
```diff
@@ -251,4 +251,8 @@ public interface PostgreServerExtension {
      */
     boolean supportsNativeClient();
 
+    /**
+     * Determines if the provided object is a PostgreSQL-specific object (PGObject) like com.amazon.redshift.util.RedshiftObject.
```
ShadelessFox:
```suggestion
     * Determines if the provided object is a PostgreSQL-specific object (PGObject) like {@code com.amazon.redshift.util.RedshiftObject}.
```

labanovichttps: fixed

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
E1izabeth: Don't know, probably accidentally

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
E1izabeth: Of course, I mean dataSource check that you use in utility method.
I don't like adding utility method that just check for instanceof and call another method. I think such utility method is redundand.

E1izabeth: Corrected my suggestion

E1izabeth (dismissed):

Matvey16:
Incorrect display of NULL values in geometry columns
dbeaver_GhxskArzDA
dbeaver_D7cKb3xe4B

ShadelessFox (approved):

Matvey16:
Verified

PR #38599, opened 2025-07-09
dbeaver/pro#6131 Update connection, sort, and data transfer icons

ShadelessFox:
Please note that 4 is not implemented


E1izabeth (approved):

Matvey16:
Verified

Issue #36021, opened 2024-10-21
Unable to download Ingres driver


Spritekin:
### Description

Opened the driver manager, selected Ingres driver, clicked in the Libraries tab then the Download/Update button. 

The download dialog always says "Ingres driver files are missing". And in the list of missing files it displays nothing. Clicking in the Download button throws no error and closed the dialog but the driver is not downloaded and is still marked as unavailable or N/A. 
<img width="691" alt="Screenshot 2024-10-22 at 10 15 10 AM" src="https://github.com/user-attachments/assets/23cc22bb-4cc0-4425-ba43-fe82bd5cc5c8">

<img width="702" alt="Screenshot 2024-10-22 at 10 14 53 AM" src="https://github.com/user-attachments/assets/c74b9930-dbb2-430d-923e-1c0ae380ccb7">

Both images above are in MacOS. In windows it is the same behaviour except it displays a red "unavailable" miniicon in the drivers page:
<img width="616" alt="Screenshot 2024-10-22 at 9 47 32 AM" src="https://github.com/user-attachments/assets/74b655d7-e4be-4587-ac27-db3d4b6ab0b3">

Trying to download again still shows the "Ingres driver files are missing" message. Trying to use an ingres connection doesn't work because the driver is not available and takes me to the download dialog with the "Ingres driver files are missing" message.

Tested in Windows and MacOS, the problem appears both.

### DBeaver Version

24.2.3.202410201725

### Operating System

MacOS and Windows

### Database and driver

Ingres, any version, I'm not even trying to connect, just download the driver.

### Steps to reproduce

1. Make sure you don't have the Ingres driver installed.
2. Open the driver manager.
3. Select the Ingres driver in the list.
4. Double click or click the "Edit..." button.
5. In the Edit driver Ingres dialog, select the "Libraries" tab.
6. Click the "Download/Update" button.
7. You can see the "Driver setting" dialog and the "Ingres driver files are missing". No files are displayed in the driver file list. Same problem appears in Windows or MacOS.

<img width="691" alt="Screenshot 2024-10-22 at 10 15 10 AM" src="https://github.com/user-attachments/assets/ca07f270-31b3-4e09-8f62-e814c50e052c">

<img width="517" alt="Screenshot 2024-10-22 at 10 25 02 AM" src="https://github.com/user-attachments/assets/69209051-4e3d-49b4-8715-39a9a00c2a8a">

[...]

PR #38605, opened 2025-07-09
Fix #38604 Ingres iijdbc download  - org.jkiss.dbeaver.generic

### LINKED ISSUE #38604: Instructions for downloading Ingres driver do not work, meta data incorrect
### Description

Since the fix for https://github.com/dbeaver/dbeaver/issues/36021 the instructions to click the version column and select a version do not work. Workaround is to install older versions from 2024, for example the last good working version is https://github.com/dbeaver/dbeaver/releases/tag/24.2.3

I believe this is due to an incorrect path, which hard codes the jar file path to an old version [10.0-4.0.6](https://central.sonatype.com/artifact/com.ingres.jdbc/iijdbc/10.0-4.0.6)	from 2010-11-26

Ideal would be higher folder, which would allow new drivers (which is what version 24.2.3 and earlier did/do).

Also the meta data is incorrect.

❌  bad version https://github.com/dbeaver/dbeaver/releases/tag/24.2.4  and later
✅  good version https://github.com/dbeaver/dbeaver/releases/tag/24.2.3  and earlier

### DBeaver Version

dbeaver-ce-25.1.2-win32.win32.x86_64

### Operating System

Windows 11

### Database and driver

                <driver
                    id="ingres"
                    label="Ingres"
                    class="com.ingres.jdbc.IngresDriver"
                    sampleURL="jdbc:ingres://{host}:{port}/{database}"
                    defaultPort="II7"
                    webURL="http://www.ingres.com/products/ingres-database/"
                    description="Ingres Database">
                    <file type="jar" path="drivers/ingres/iijdbc.jar" url="http://dbeaver.jkiss.org/files/jdbc/drivers/ingres/iijdbc.jar"/>
                    <file type="license" path="drivers/ingres/LICENSE.txt" url="http://dbeaver.jkiss.org/files/jdbc/drivers/ingres/LICENSE.txt"/>
                </driver>

### Steps to reproduce

1. Create connection
2. select Ingres
3. enter in details, e.g. database/schema demodb
4. try and progress, which triggers download dialog

Links are incorrect.

### Additional context

My guess is https://github.com/dbeaver/dbeaver/releases/tag/24.2.4 is the first build with this problem, https://github.com/dbeaver/dbeaver/releases/tag/24.2.0 likely does not have jdbc download problem (but does have incorrect metadata).

Workaround install (older version, before bug) https://github.com/dbeaver/dbeaver/releases/tag/24.2.3

clach04:
Corrected Maven download path, along with other metadata to be current.


clach04:
I suspect the code 

https://github.com/clach04/dbeaver/blob/f156f8f43767c49abdffd4a8a44cd36f77d00743/plugins/org.jkiss.dbeaver.ext.generic/plugin.xml#L411

should be updated to:

```xml
<file type="jar" path="maven:/com.ingres.jdbc:iijdbc:RELEASE" bundle="!drivers.ingres"/>
```
which was the original code before the fix for #36021

I'm not sure how to test this, my environment seems to be caching the old content. Even after wiping out `%APPDATA%\DBeaverData`

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -403,14 +403,13 @@
...
-                    webURL="http://www.ingres.com/products/ingres-database/"
+                    webURL="https://www.actian.com/databases/ingres/"
                     description="Ingres Database"
                     supportedConfigurationTypes="MANUAL,URL"
                     iconBig="icons/ingres_icon_big.png"
                     categories="sql">
-                    <!-- We use direct jar path because Ingres metadata is broken #36021 -->
-                    <file type="jar" path="https://repo1.maven.org/maven2/com/ingres/jdbc/iijdbc/10.0-4.0.6/iijdbc-10.0-4.0.6.jar" bundle="!drivers.ingres"/>
-                    <file type="license" path="licenses/external/lgpl-3.0.txt"/>
+                    <file type="jar" path="https://repo1.maven.org/maven2/com/ingres/jdbc/iijdbc/" bundle="!drivers.ingres"/>
```
E1izabeth: The path here should point to a specific version or jar file like it was before. This won't work this way

E1izabeth (changes requested):

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
clach04: Do you have more information on this? Maybe a doc link? There are zero fixed jar file paths in the existing plugin file (other than the one Ingres had since Oct 2024), but there are about 50 maven: URLs . There are 16 that have version specifications that appear to be hints, rather than fixed ones. E.g.

https://github.com/dbeaver/dbeaver/blob/e46afd1132c087f6941f37c8aaf01da641994de4/plugins/org.jkiss.dbeaver.ext.generic/plugin.xml#L1244

That seems to more closely match the comment I made yesterday in https://github.com/dbeaver/dbeaver/pull/38605#issuecomment-3054324052

I'm thinking it needs to be:

```xml
<file type="jar" path="maven:/com.ingres.jdbc:iijdbc:RELEASE[12.0-4.4.4]" bundle="!drivers.ingres"/>
```
 I'm wondering if the bundle parameter should be removed:
```xml
<file type="jar" path="maven:/com.ingres.jdbc:iijdbc:RELEASE[12.0-4.4.4]"/>
```

clach04:
@serge-rider any thoughts on this? I'm wondering if the bundle part of the file element/line is needed. I don't 100% understand all of the implications here.

E1izabeth:
Passed to QA team for testing

uslss:
verified

PR #38628, opened 2025-07-12
dbeaver/pro#6247 OpenAI client redesign

serge-rider:


ShadelessFox (dismissed):

ivgag:
Have you considered using [openai-java](https://github.com/openai/openai-java)? It supports JSON schemas for structured output and functions, validation, and other useful stuff.

ivgag (changes requested):
We need the ability to generate JSON schemas for functions and structured outputs. As far as I know, Gson doesn’t support this, whereas Jackson does. Therefore, I’d prefer to keep using Jackson for AI serialization/deserialization operations to avoid mixing libraries.

PR #38636, opened 2025-07-13
dbeaver/pro#3820 Refactor semantic analysis

### LINKED ISSUE #34538: Query Runs Perfectly But Syntax Error Coloring
### Description

I am attaching a screenshot where the query runs perfectly. However, DBeaver highlights some parts in red. Any ideas why? 
Screenshot 2024-06-27 193349

### DBeaver Version

Version 24.1.1.202406231636

### Operating System

Windows 11 Enterprise

### Database and driver

Server: Snowflake 8.20.0
Driver: Snowflake 3.14.4

### Steps to reproduce

_No response_

### Additional context

_No response_

### LINKED ISSUE #34919: Problem marker is shown for schema name of a function call
### Description

This is another case when some correct code without error is colored with "SQL semantic error foreground".
This issue is similar to this: #34741

### DBeaver Version

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

You can use dvdrental database (it contains the table "actor").

use code:

```
CREATE OR REPLACE FUNCTION public."get_json"("in_actor_id" integer)
 RETURNS json
 LANGUAGE sql
 STABLE
AS $function$
    select json_build_object('id', actor_id)
    from public.actor as a
    ;
$function$
;

select public."get_json"(a.actor_id) 
from actor as a

```

here is how it is shown (notice the red word "public"): 

изображение

### LINKED ISSUE #35465: Wrong highlighting of function alias
`jsonb_array_elements` aliases, e.g., `grade` is incorrectly colored as red.

```sql
    SELECT
      *
    FROM
      jsonb_array_elements('[77, 85, 92, 95, 74, 79, 80, 97]') AS grade
    WHERE
      CAST(grade AS INTEGER) > 85;
```

_Originally posted by @theresa-de-ocampo in https://github.com/dbeaver/dbeaver/issues/34741#issuecomment-2310387451_

### LINKED ISSUE #35996: Problem marker "column not found" is shown for correct query with composite type from table alias
### Description

I'm not sure about the proper name of this feature, but I will name it later as a "composite type from table alias" until I find the correct name.
As I know, this feature works for Postgres, but not for some other types of databases.
I like how you show it in the application.
The issue is it doesn't supported well for some existing features.
The first issue is that the application doesn't treat alias as a correct column name in a select-clause of the query and shows problem marker for such a query:
```
select a
from actor as a
```
There is no column "a" in this table, it is true.
"a" is a table alias here, representing the whole line of the underlying table.
I suggest to do not show this problem marker if the alias name is presented.

https://github.com/user-attachments/assets/9906b2a5-00b7-4efa-b235-27c330350e7d

### DBeaver Version

Enterprise Edition 24.2.4-202410172322

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.4

### Steps to reproduce

_No response_

### Additional context

_No response_

### LINKED ISSUE #37911: SQL Editor Warning: `Table or subquery not found` for schema prefixed PostgreSQL function call
### Description

Functions with the full schema prefixed name shows warnings in the SQL editor:
```sql
SELECT pg_stat_statements_info();
SELECT public.pg_stat_statements_info();  -- WARNING
```

Image

### DBeaver Version

Community Edition 25.0.3

### Operating System

Windows 11

### Database and driver

PostgreSQL 17.4
PostgreSQL driver 42.7.5

### Steps to reproduce

Enter in SQL editor:

```sql
SELECT pg_stat_statements_info();
SELECT public.pg_stat_statements_info();  -- WARNING
```

### Additional context

_No response_

### LINKED ISSUE #38707: Formatting Disabled when adding Database to Function
When I add the database to the beginning of a function (i.e. change 
`if(d.loan_status_code is null, 0, gettotaldebtamt_v2('total', d.defer_bal_amt, d.curr_bal_amt)) as total_amt` 
to 
`if(d.loan_status_code is null, 0, prod.gettotaldebtamt_v2('total', d.defer_bal_amt, d.curr_bal_amt)) as total_amt` )
then I loose all color coding (see below boxed in red) and formatting. 

I want to NOT loose all color coding and formatting when I add the database to the beginning of a function. 

**Additional context**

_Before Adding Database_
Image

_After Adding Database_
Image

Image

E1izabeth:
### Main changes
- Refactored the fully qualified names model
- Use the new resolution algorithm everywhere (not only in CTE)
- Support LATERAL JOIN in CTE

- Support functions (added separate option for validation) - turned off by default for now. It makes sense to test for PostgreSQL. I checked that it doesn't work well for BigQuery, Snowflake and probably some other databases due to bugs in metadata model

image
image

-  Outline now has highlighting of columns, tables, and error nodes

image

- Outline now has columns under `*` node for `alias.*` expression (but not for `*` intentionally, because there might be too many columns)

image

### For QA:
- Check autocomplete
- Check queries validation, highlighting
- Check highlighting when no connection
- Check the procedures reading in Redshift (that they appear in the navigator)


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/SQLQueryComplexName.java
```diff
@@ -0,0 +1,113 @@
...
+import org.jkiss.dbeaver.model.stm.STMTreeNode;
+import org.jkiss.dbeaver.model.stm.STMUtils;
+
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Describes qualified or compound name of the database entity
+ */
+public class SQLQueryComplexName {
```
ShadelessFox, line 32: A perfect candidate for record-ification. It's read-only and non-extensible.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/SQLQuerySemanticUtils.java
```diff
@@ -0,0 +1,396 @@
...
+import org.jkiss.dbeaver.model.struct.*;
+import org.jkiss.dbeaver.model.struct.rdb.*;
+import org.jkiss.utils.Pair;
+
+import java.lang.reflect.AccessFlag;
+import java.util.*;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class SQLQuerySemanticUtils {
```
ShadelessFox, line 44: Add a private constructor

ShadelessFox (commented):
It's pretty much impossible to review every line of this PR, so here are at least a couple of small nitpicks

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/SQLQueryComplexName.java
E1izabeth: Record can't have logic inside a public primary constructor 🤔

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/SQLQuerySemanticUtils.java
E1izabeth: OK, thanks

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/SQLQueryComplexName.java
ShadelessFox: Use factory methods. `static SQLQueryComplexName of(...)`

E1izabeth: I's prefer to left it as class instead of using factory methods

E1izabeth: Also, the record's canonical constructor can't be private. In this case, it will allow the creation of invalid instances

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/context/SQLQueryConnectionContext.java
```diff
@@ -83,82 +66,48 @@ public static DBSObject expandAliases(@NotNull DBRProgressMonitor monitor, @Null
...
-        return obj instanceof DBSTable table ? table : (obj instanceof DBSView view ? view : null);
+        List<? extends DBSObject> objs = this.findRealObjectsImpl(monitor, tableName);
+        if (objs.isEmpty()) {
+            return Collections.emptyList();
+        } else {
+            List<DBSEntity> results = new ArrayList<>(objs.size());
+            for (int i  = 0; i < objs.size(); i++) {
+                DBSObject obj = expandAliases(monitor, objs.get(i));
+                DBSEntity entity = obj instanceof DBSTable table ? table : (obj instanceof DBSView view ? view : null);
+                results.add(entity);
```
labanovichttps: Maybe it's better to check if `entity` is not null before adding?

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/model/ddl/SQLQueryObjectDataModel.java
```diff
@@ -59,10 +57,10 @@ public class SQLQueryObjectDataModel extends SQLQueryRowsSourceModel implements
     private SQLQuerySymbolOrigin objectNameOrigin = null;
 
     public SQLQueryObjectDataModel(
-            @NotNull STMTreeNode syntaxNode,
-            @NotNull SQLQueryQualifiedName name,
-            @NotNull DBSObjectType objectTypes,
-            @NotNull Set<DBSObjectType> objectContainerTypes
+        @NotNull STMTreeNode syntaxNode,
+        @NotNull SQLQueryComplexName name,
```
labanovichttps, line 61: maybe use fullname for this field?

labanovichttps (dismissed):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/model/ddl/SQLQueryObjectDataModel.java
E1izabeth: It's the name of an object that might be represented as just a single identifier or as a fully qualified name. The common `name` is more suitable here

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/context/SQLQueryConnectionContext.java
E1izabeth: good point

ShadelessFox (dismissed):

uslss:
verified

PR #38646, opened 2025-07-14
#38644 Colorize SQL editor tabs according to the connection's color

### LINKED ISSUE #38644: Extend tab colorization to SQL editors
In #10235, we added color coding for tabs according to the connection's color. Let's extend this to SQL editor tabs as well.

ShadelessFox:
image


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/e4/DBeaverCTabFolderRenderer.java
```diff
@@ -128,6 +127,27 @@ private static IEditorInput getEditorInput(@NotNull MPart part) {
         return null;
     }
 
+    @Nullable
+    private static Color getConnectionColor(@NotNull IEditorPart editorPart) {
```
serge-rider, line 131: It is better to use DBPDataSourceContainerProvider but DBPContextProvider (as DBPContextProvider works only with connected databases)

serge-rider (changes requested):

Matvey16:
Verified

PR #38663, opened 2025-07-16
dbeaver/pro#6139 Added handling for boolean array editor values

Nexus6v2:
Added handler condition allowing to edit boolean arrays in Clickhouse

image


E1izabeth (approved):

E1izabeth:
As far as I understand, these settings don't work for boolean arrays, right?
image

labanovichttps (approved):

Matvey16:
Verified

PR #38673, opened 2025-07-16
dbeaver/pro#6289 Fix NPE for incomplete create view without view name

E1izabeth:


E1izabeth:
We decided that it's not a hotfix

PR #38685, opened 2025-07-17
dbeaver/pro#6304 Do not load datasources if project not loaded

serge-rider:


plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/EditorUtils.java
```diff
@@ -322,13 +328,17 @@ public static String[] getInputContextDefaults(DBPDataSourceContainer dataSource
 
     @Nullable
     public static DBPDataSourceContainer getFileDataSource(IFile file) {
+
+    }
```
ShadelessFox: Empty method with no return - this won't compile

ShadelessFox (changes requested):

ShadelessFox (approved):

Matvey16:
Verified

PR #38692, opened 2025-07-17
Dbeaver/pro#5894 audio recording and speech recognition

vkirst:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/speech/TranscriptResult.java
```diff
@@ -0,0 +1,23 @@
...
+ * patents in process, and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this material
+ * is strictly forbidden unless prior written permission is obtained
+ * from DBeaver Corp.
+ */
+package org.jkiss.dbeaver.model.ai.speech;
+
+public record TranscriptResult(String text, Usage usage) {
+
+    public record Usage(String type, int seconds) { }
```
ShadelessFox: What's `type`? And please use `java.time.Duration duration` instead of `int seconds`

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/speech/TranscriptResult.java
```diff
@@ -0,0 +1,23 @@
...
+ * herein are proprietary to DBeaver Corp and its suppliers
+ * and may be covered by U.S. and Foreign Patents,
+ * patents in process, and are protected by trade secret or copyright law.
+ * Dissemination of this information or reproduction of this material
+ * is strictly forbidden unless prior written permission is obtained
+ * from DBeaver Corp.
+ */
+package org.jkiss.dbeaver.model.ai.speech;
+
+public record TranscriptResult(String text, Usage usage) {
```
ShadelessFox: Can we inline `usage`?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AIEngine.java
```diff
@@ -75,4 +77,9 @@ Flow.Publisher<AIEngineResponseChunk> requestCompletionStream(
     boolean hasValidConfiguration() throws DBException;
 
     boolean isLoggingEnabled() throws DBException;
+
+    boolean isSuportTranscription();
+
+    @NotNull
+    TranscriptResult createSpeechTranscription(@NotNull Path audioFile) throws DBException;
```
ShadelessFox: This is a potentially time-consuming operation. Please supply a `DBRProgressMonitor` there.

Just a nitpick, but having the API accept a `Path` quite limits its potential usage. Can it instead accept an `InputStream`?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AIEngine.java
```diff
@@ -75,4 +77,9 @@ Flow.Publisher<AIEngineResponseChunk> requestCompletionStream(
     boolean hasValidConfiguration() throws DBException;
 
     boolean isLoggingEnabled() throws DBException;
+
+    boolean isSuportTranscription();
```
ShadelessFox:
```suggestion
    boolean supportsTranscription();
```

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/speech/TranscriptResult.java
vkirst: I’d keep it separate — it mirrors the OpenAI JSON response and will be easier to evolve for other providers.
```json
{
    "text": "bla-bla",
    "usage": {
        "type": "duration",
        "seconds": 3
    }
}
```

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AIEngine.java
vkirst: Actually, the caller already wraps this method in a Job with a DBRProgressMonitor. Would that be enough?

ShadelessFox: A monitor opens up many possibilities. You can report the upload process and the transcription process. So it's worth sticking one there regardless.

PR #38748, opened 2025-07-22
dbeaver/pro#6219 Add UserAgent param to the databricks

labanovichttps:


plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/DatabricksDataSource.java
```diff
@@ -51,4 +58,16 @@ private boolean isLegacyDriver() {
...
 
+
+    @Override
+    protected Connection openConnection(
+        @NotNull DBRProgressMonitor monitor,
+        @Nullable JDBCExecutionContext context,
+        @NotNull DBPConnectionConfiguration connectionInfo,
+        @NotNull String purpose
+    ) throws DBCException {
+        String userAgent = GeneralUtils.getProductName().replace(" ", "+") + "/" + GeneralUtils.getProductVersion();
```
ShadelessFox: > `.replace(" ", "+")`

What does it do? Should it be URL-escaped?

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/DatabricksDataSource.java
```diff
@@ -51,4 +60,19 @@ private boolean isLegacyDriver() {
         return CommonUtils.equalObjects(DatabricksConstants.DRIVER_CLASS_LEGACY, getContainer().getDriver().getDriverClassName());
     }
 
+
+    @Override
+    protected Connection openConnection(
```
E1izabeth: I think it's better to add this to `getAllConnectionProperties` method and use `DBUtils.getClientApplicationName`.

labanovichttps: getAllConnProps is a nice idea, thanks, but getClientAplicationName gives an incorrect formatting name 
image

E1izabeth (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified

plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/DatabricksDataSource.java
```diff
@@ -51,4 +59,19 @@ private boolean isLegacyDriver() {
...
+    @NotNull
+    @Override
+    protected Properties getAllConnectionProperties(
+        @NotNull DBRProgressMonitor monitor,
+        JDBCExecutionContext context,
+        String purpose,
+        DBPConnectionConfiguration connectionInfo
+    ) throws DBCException {
+        String userAgent = URLEncoder.encode(
+            GeneralUtils.getProductName() + "/" + GeneralUtils.getProductVersion(),
```
arhayka, line 71: It seems that here the company name is missing. There will be something like this: 
- DBeaver Ultimate/25.1.4
- CloudBeaver Enterprise/25.1.4

And according to Databricks docs, we need to pass something like this:
- DBeaver DBeaver Ultimate/25.1.4
- DBeaver CloudBeaver Enterprise/25.1.4

PR #38750, opened 2025-07-22
5679 cb execute gql requests via cli

alexander-skoblikov:
closes https://github.com/dbeaver/pro/issues/5679


plugins/org.jkiss.dbeaver.model.cli/src/org/jkiss/dbeaver/model/cli/CommandLineContext.java
```diff
@@ -0,0 +1,75 @@
...
+    public void setContextParameter(@NotNull String name, @NotNull Object value) {
+        contextParameter.put(name, value);
+    }
+
+
+    public void addResult(@NotNull Object result) {
+        this.results.add(result);
+    }
+
+    @Nullable
```
ShadelessFox: Not nullable

plugins/org.jkiss.dbeaver.model.cli/src/org/jkiss/dbeaver/model/cli/CommandLineContext.java
```diff
@@ -0,0 +1,75 @@
...
+    public void addCloseHandler(@NotNull Runnable closeHandler) {
+        closeHandlers.add(closeHandler);
+    }
+
+    @Override
+    public void close() {
+        for (Runnable closeHandler : closeHandlers) {
+            try {
+                closeHandler.run();
+            } catch (Exception e) {
```
ShadelessFox, line 70: How on earth could it throw? Maybe unhandled, yes, yet questionable

plugins/org.jkiss.dbeaver.model.cli/src/org/jkiss/dbeaver/model/cli/ICommandLineParameterHandler.java
```diff
@@ -14,12 +14,14 @@
...
+package org.jkiss.dbeaver.model.cli;
 
 import org.apache.commons.cli.CommandLine;
+import org.jkiss.code.NotNull;
+import org.jkiss.code.Nullable;
 
 public interface ICommandLineParameterHandler {
 
-    void handleParameter(CommandLine commandLine, String name, String value);
+    void handleParameter(@NotNull CommandLine commandLine, @NotNull String name, @Nullable String value, CommandLineContext context);
```
ShadelessFox: Missing nullability annotation

plugins/org.jkiss.dbeaver.model.cli/src/org/jkiss/dbeaver/model/cli/registry/CommandLineParameterDescriptor.java
```diff
@@ -75,5 +78,9 @@ public String getLongName() {
     public String getName() {
         return name;
     }
+
+    public boolean isContextInitializer() {
```
ShadelessFox, line 86: What does this flag mean?

plugins/org.jkiss.dbeaver.model.cli/src/org/jkiss/dbeaver/model/cli/CommandLineContext.java
alexander-skoblikov: This is for a potential RuntimeException,  in my opinion it is better to add extra handler and close the context than to find strange bugs after

plugins/org.jkiss.dbeaver.model.cli/src/org/jkiss/dbeaver/model/cli/registry/CommandLineParameterDescriptor.java
alexander-skoblikov: means whether the handler should be processed first to fill `CommandLineContext`

PR #38756, opened 2025-07-23
dbeaver/pro#6257 Show a warning when using an incompatible OS version

ShadelessFox:
The following warning will be shown against the following OS versions:
- Windows 7
- Windows 8
- Windows 8.1
- Windows Server 2012

java_jQiVJX9UZZ


plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/DBeaverLauncher.java
```diff
@@ -311,6 +314,13 @@ public void updateSplash() {
     private static final String DBEAVER_CONFIG_FILE = "global-settings.ini";
     private static final String DBEAVER_PROP_LANGUAGE = "nl";
 
+    // List of incompatible Windows versions for those which we want to show a warning dialog
+    private static final Set<String> INCOMPATIBLE_WINDOWS_VERSIONS = Set.of(
```
E1izabeth, line 318: We have a similar list `LEGACY_WINDOWS_VERSIONS`. Maybe we can reuse it separately adding the difference

Another suggestion is to add at least `Windows Server 2012 R2` because George added it for some reason. So, assume it's a valid value, too.

ShadelessFox: > We have a similar list LEGACY_WINDOWS_VERSIONS. Maybe we can reuse it separately adding the difference

It includes `Windows Server 2019`, so I can't use it. I wish I could rely on the build number (as SWT does), but we don't have a way to access the Win32 API (no JNA, no SWT, no FFM).

> Another suggestion is to add at least Windows Server 2012 R2 because George added it for some reason. So, assume it's a valid value, too.

The code uses `startsWith`, so it will also match `Windows Server 2012 R2`.

E1izabeth (approved):

IvanGorshechnikov (approved):

Matvey16:
Can't see the error on Windows 7, other errors appear earlier

https://github.com/user-attachments/assets/6b53d626-6847-43cf-b064-ec56949ed176

ShadelessFox:
Expected. We can't do anything about it on Windows 7.

Matvey16:
Verified

PR #38758, opened 2025-07-23
dbeaver/pro#5848 api for renaming projects

yagudin10:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/app/BaseProjectImpl.java
```diff
@@ -437,6 +453,16 @@ public void refreshProject(DBRProgressMonitor monitor) {
 
     }
 
+    @Override
+    public void updateProject(@Nullable String newName, @Nullable String description) throws DBException {
+        if (CommonUtils.isNotEmpty(newName)) {
```
serge-rider, line 467: Then how to remove custom name?
How to remove description?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/app/BaseProjectImpl.java
yagudin10: why do we need to clear name?
admin can set it manually if he wants

about description, you can remove it by setting the value as an empty string rn
anyway, the task is about renaming projects, we can think about description in another task

Nexus6v2 (dismissed):

PR #38771, opened 2025-07-25
dbeaver/pro#6115 Added AI chat query confirmation config

Nexus6v2:


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -2832,6 +2832,47 @@ public List<SQLScriptElement> getSelectedQueries() {
...
+        final DBCTransactionManager txnManager = Objects.requireNonNull(
+            DBUtils.getTransactionManager(getExecutionContext())
+        );
+        boolean autoCommit = dataSourceContainer.isDefaultAutoCommit();
+        boolean smartAutoCommit = isSmartAutoCommit();
+        DBRProgressMonitor monitor = new VoidProgressMonitor();
+
+        SQLQueryListener autocommitListener = new SQLEditorQueryListener(curQueryProcessor, false) {
+            @Override
+            public void onStartScript() {
```
Nexus6v2: I'm not sure, that's how it needs to be done. 
Doesn't work now at least

PR #38775, opened 2025-07-25
dbeaver/dbeaver#36784 Prevent duplicate user creation

### LINKED ISSUE #36784: Copy->Paste->Save of MySQL user does not work properly
### Description

1. When copying/pasting a user between databases/servers, "Save..." does not generate the `CREATE USER` command, so nothing happens (or gives error when choosing any schema privilege for it: `Can't find any matching row in the user table`).
2. When copying/pasting a user in the same DB, the new user has the same name. It should be added a numeric suffix to differentiate itself from other existing users.

### DBeaver Version

Community Edition v24.3.2.202501051917

### Operating System

Windows 11 64 bit

### Database and driver

JDBC MySQL v8.0.40

### Steps to reproduce

1. Copy a MySQL user
2. Paste it.
3. Save it.

### Additional context

_No response_

vkirst:


IvanGorshechnikov (dismissed):

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/config/MySQLUserManager.java
```diff
@@ -140,12 +140,22 @@ protected CommandCreateUser(MySQLUser user)
...
                 throw new DBException("Can't create user with empty name");
             }
-            if (CommonUtils.isEmpty(getObject().getHost())) {
+
+            final String host = getObject().getHost();
+            if (CommonUtils.isEmpty(host)) {
                 throw new DBException("Can't create user with empty host name");
             }
+
+            if (getObject().getDataSource().getUsers(monitor)
```
serge-rider, line 153: Why bot use MySQLDataSource.getUser?

plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/config/MySQLCommandChangeUser.java
```diff
@@ -92,6 +92,7 @@ public void afterExecute(DBCSession session, Throwable error)
                     {
                         if (error == null) {
                             getObject().setPersisted(true);
+                            getObject().getDataSource().resetUsers();
```
serge-rider, line 95: Not clear why to reset already read users here?

plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/config/MySQLUserManager.java
vkirst: Do you mean getUser(monitor, username)? But I need to find a user by both username and host.

plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/config/MySQLCommandChangeUser.java
vkirst: I reset the cache here because we’ve just added a user, so the cached users is outdated.
Otherwise, getUser(...) and getUsers(...) wouldn’t return the new user.

uslss:
Copied user disappears after refresh

https://github.com/user-attachments/assets/c90fb8d6-d95e-4be8-b2c9-d759c8a9cb07

E1izabeth (approved):

uslss:
verified

PR #38838, opened 2025-08-01
dbeaver/pro#5448 Handle and dispose of folder editor without an input

ShadelessFox:
If a node can't be loaded, the folder editor will automatically close. This change is a bit controversial, IMO, but it's better than having a broken editor that does nothing.


E1izabeth (dismissed):

Matvey16:
Helps only for folders, but the same behaviour still exists for files. Can't open any object from the navigator because of the error
```
java.lang.IllegalArgumentException: Invalid dbvfs URI: dbvfs://General/s3/aws/dbeaver-downloads/?%2Fdbeaver-downloads%2FTest%2FScript-1265432.sql
	at org.jkiss.dbeaver.model.fs.nio.EFSNIOFileSystem.getFileStoreFromUri(EFSNIOFileSystem.java:115)
	at org.jkiss.dbeaver.model.fs.nio.EFSNIOFileSystem.lambda$0(EFSNIOFileSystem.java:57)
	at org.jkiss.dbeaver.ui.UIUtils.lambda$15(UIUtils.java:1849)
	at org.jkiss.dbeaver.runtime.RunnableContextDelegate.lambda$0(RunnableContextDelegate.java:39)
	at org.eclipse.jface.operation.ModalContext$ModalContextThread.run(ModalContext.java:123)
```

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/e4/DBeaverCTabFolderRenderer.java
```diff
@@ -142,6 +146,10 @@ private static Color getConnectionColor(@NotNull CTabItem item) {
 
     @Nullable
     private static Color getConnectionColor(@NotNull MPart part) {
+        if (part.getTransientData().containsKey(PART_SKIP_KEY)) {
+            return null;
+        }
```
E1izabeth, line 152: Please, at least add a comment on why we need this

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/e4/DBeaverCTabFolderRenderer.java
```diff
@@ -155,7 +163,8 @@ private static Color getConnectionColor(@NotNull MPart part) {
 
             try {
                 return getConnectionColor(ref.getEditorInput());
-            } catch (PartInitException e) {
+            } catch (Exception e) {
+                part.getTransientData().put(PART_SKIP_KEY, Boolean.TRUE);
```
E1izabeth: and here

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/FolderEditor.java
```diff
@@ -65,6 +65,12 @@ public void createPartControl(Composite parent) {
 
         UIExecutionQueue.queueExec(() -> {
             final DBNNode navigatorNode = getEditorInput().getNavigatorNode();
+            if (navigatorNode == null) {
```
E1izabeth, line 68: How about other usages of getNavigatorNode method? I don't see it's expected that getNavigatorNode might return null somewhere else. I would not change the contract.

ShadelessFox: All other usages expect it to return `null` and handle such a scenario.

E1izabeth: image

E1izabeth: Also could you intruoduce `@Nullable` annotation then?

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/e4/DBeaverCTabFolderRenderer.java
ShadelessFox: Added comment

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/e4/DBeaverCTabFolderRenderer.java
ShadelessFox: Added comment where it's set

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/FolderEditor.java
ShadelessFox: Well, not every single one, I guess.

plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/NodeEditorInput.java
```diff
@@ -75,7 +77,8 @@ public boolean exists() {
 
     @Override
     public ImageDescriptor getImageDescriptor() {
-        return DBeaverIcons.getImageDescriptor(getNavigatorNode().getNodeIconDefault());
+        DBNNode node = getNavigatorNode();
+        return node == null ? null : DBeaverIcons.getImageDescriptor(node.getNodeIconDefault());
```
E1izabeth, line 81: Now we have three places with potential NPE for `getImageDescriptor()` call 🥲
image

E1izabeth: Maybe this NPE never happen, idk if it worth to worry about that, but that's what I see

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/NodeEditorInput.java
ShadelessFox: Well, the javadoc clearly says that it's allowed to return `null`. But judging by those calls, neither of them seems to be affected by my changes

labanovichttps (approved):

Matvey16:
Verified

PR #38839, opened 2025-08-01
dbeaver/pro#4302 Driver files edit fix in TE

serge-rider:


plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/driver/DriverEditDialog.java
```diff
@@ -964,6 +970,19 @@ private void syncDriverLibraries() throws DBException {
         }
     }
 
+    private boolean isAppInstallationFile(Path localFilePath) {
+        try {
+            Path appInstallPath = RuntimeUtils.getLocalPathFromURL(Platform.getInstallLocation().getURL());
+            if (localFilePath.toString().startsWith(appInstallPath.toString())) {
```
ShadelessFox: You can use `Path#startsWith`. Or is the string conversion intentional?

ShadelessFox (dismissed):

PR #38852, opened 2025-08-04
dbeaver/pro#6027 Fix workspaces load; prioritize `.workspaces` file

ShadelessFox:
DBeaver now prioritizes workspaces from the `.workspaces` file located under `DBeaverData`.

Please test different cases, e.g., when switching between different editions or having more than 10 workspaces.


plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -380,81 +379,94 @@ private boolean setIDEWorkspace(@NotNull Location instanceLoc) {
...
-            } else if (maxSize > 1) {
-                workspaces.set(recentWorkspaces.size() - 1, WORKSPACE_DIR_CURRENT);
+
+        if (!workspaces.contains(WORKSPACE_DIR_CURRENT)) {
+            if (workspaces.size() == RECENT_MAX_LENGTH) {
+                workspaces.set(workspaces.size() - 1, WORKSPACE_DIR_CURRENT);
+            } else {
+                workspaces.add(WORKSPACE_DIR_CURRENT);
             }
         }
```
flambchap: It seems like we can put the current (latest) workspace at the bottom of the file, while we expect it to be on the top (see `setIDEWorkspace `).

flambchap (changes requested):
I suggest extracting workspace-related logic to a dedicated class.

flambchap (approved):
lgtm

Matvey16:
Verified

PR #38857, opened 2025-08-05
dbeaver/pro#5805 Refactor foreign key loading in MySQLTable

labanovichttps:
https://github.com/user-attachments/assets/d5954dad-1dc1-4d02-a814-0f1574dd2c0e


serge-rider:
Did this change affected performance anyhow?
It looks like previously it did the same thing but inside JDBC driver

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLTable.java
```diff
@@ -469,153 +472,192 @@ void fetchAdditionalInfo(JDBCResultSet dbResult) {
...
+                    while (dbResult.next()) {
+                        String pkTableCatalog = JDBCUtils.safeGetString(dbResult, "pk_table_schema");
+                        String pkTableName = JDBCUtils.safeGetString(dbResult, "pk_table_name");
+                        String pkColumnName = JDBCUtils.safeGetString(dbResult, "pk_column_name");
+                        String fkTableCatalog = JDBCUtils.safeGetString(dbResult, "fk_table_schema");
+                        String fkTableName = JDBCUtils.safeGetString(dbResult, "fk_table_name");
+                        String fkColumnName = JDBCUtils.safeGetString(dbResult, "fk_column_name");
+                        int keySeq = JDBCUtils.safeGetInt(dbResult, "key_seq");
+                        String updateRule = JDBCUtils.safeGetString(dbResult, "UPDATE_RULE");
+                        String deleteRule = JDBCUtils.safeGetString(dbResult, "DELETE_RULE");
```
E1izabeth, line 529: Why not using JDBCConstants anymore?

uslss:
looks like it works faster...
verified

rubengonzalezasencio:
The changes made have caused poor performance on servers with hundreds of databases. In these cases, the query takes several minutes.

Anthony-Saresoft:
If this is the source of the problem, wouldn't it be advisable to reverse them?

P.S: @labanovichttps you broke the system hahaha

PR #38861, opened 2025-08-05
dbeaver/pro#5640 Add expandable composite with a border util

ShadelessFox:


E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -2507,4 +2508,30 @@ public static boolean isWindowVisible(@NotNull Display display, @NotNull Class<?
...
+     * Creates an {@link ExpandableComposite} that paints a separator on top.
+     */
+    @NotNull
+    public static ExpandableComposite createExpandableCompositeWithSeparator(
+        @NotNull Composite parent,
+        int style,
+        int expansionStyle
+    ) {
+        // We have to use an anonymous class because "textLabel" has protected access
+        return new ExpandableComposite(parent, style, expansionStyle) {{
```
serge-rider, line 2526: Why subclassing?
It pollutes codebase with inner classes and makes no sense.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
ShadelessFox: I have explained why it's need in the comment a line above. Having a standalone class is a bit too much IMO, as it doesn't expose any new functionality. I can make it package-private and leave the util function if you want

PR #38866, opened 2025-08-05
dbeaver/pro#4344 Make it clear that the library version can be changed

ShadelessFox:
|Dialog|Drop-down|
|-|-|
|java_qNFeNWSrYy|java_peWmWx6H7U|

I decided not to add the context menu because having a single entry there would be silly.


uslss:
verified

PR #38879, opened 2025-08-06
dbeaver/pro#6366 Log SELECT/INSERT queries in QM during data transfer

ShadelessFox:
Known issue: QM queries only show the last row, not the whole batch. A separate issue will be created to address that.

On export:
java_pw8jOUOFjZ

On import:
java_s8VWibWP6h


E1izabeth (dismissed):

Matvey16:
Checkbox selection is not saved after data transfer and needs to always be checked again


https://github.com/user-attachments/assets/1448946f-b0bd-413b-b937-cf562a7c72d3

Matvey16:
Verified

PR #38899, opened 2025-08-07
dbeaver/dbeaver#38631 Addeed all dbs checkbox for mysql

### LINKED ISSUE #38631: Consider to add to the "Databases" directory the option to show the unique connection or not
Hello

I have a server in my lan running with MySQL 8 Community edition.
Consider it has 4 databases created such as:

* alpha
* beta
* tango
* zulu

If possible create in peace a connection to anyone of them, let say `tango`

Thus when is accomplished the following steps

1. Select the connection
2. Expand the connection to establish the connection
3. Go to the `Databases` directory and expand it
4. Appears the database in bold according the connection configured from the beginning

**But** in the  `Databases` directory appears the other 3 databases too

Even more, is possible get access to any of them.

It would be problematic if for any reason by mistake (human error by being be tired in the end of the day) is done some undesired operation  in the other database

Therefore: Consider to add to the `Databases` directory the option to show the unique connection or not

Thanks for your understanding

Nexus6v2:
* Added filter for showing all/selected database for MySQL

image
image

* Includes support for url
image


plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/views/MySQLConnectionPage.java
```diff
@@ -220,6 +225,12 @@ public void loadSettings() {
                 CommonUtils.notEmpty(site.getDriver().getDefaultDatabase())
             ));
         }
+        if (showAllDatabases != null) {
+            showAllDatabases.setSelection(CommonUtils.getBoolean(
```
ShadelessFox, line 231: You use the preference as the default value if the connection doesn't have such a property. But it's only used as the state of the checkbox. The datasource itself only relies on the connection property, but doesn't fall back to the preference.

plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/config/MySQLPreferencesInitializer.java
```diff
@@ -0,0 +1,34 @@
...
+import org.jkiss.dbeaver.ext.mysql.ui.MySQLUIActivator;
+import org.jkiss.dbeaver.model.impl.preferences.BundlePreferenceStore;
+import org.jkiss.dbeaver.model.preferences.DBPPreferenceStore;
+import org.jkiss.dbeaver.utils.PrefUtils;
+
+public class MySQLPreferencesInitializer extends AbstractPreferenceInitializer {
+
+    @Override
+    public void initializeDefaultPreferences() {
+        DBPPreferenceStore store = new BundlePreferenceStore(MySQLUIActivator.getDefault().getBundle());
```
ShadelessFox: Why are we storing that prop in a preference store in the first place?

Nexus6v2: Yeah, that's an overkill for a variable in `providerProperties`
Also, I realized I should default it to `true` as it's the current behavior and cutting dbs from the list by default can be confusing

Nexus6v2: Removed store

plugins/org.jkiss.dbeaver.ext.mysql.ui/src/org/jkiss/dbeaver/ext/mysql/ui/views/MySQLConnectionPage.java
Nexus6v2: Added constant default

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLDataSource.java
```diff
@@ -864,6 +865,16 @@ public class CatalogCache extends JDBCObjectCache<MySQLDataSource, MySQLCatalog>
...
+            DBPConnectionConfiguration configuration = owner.getContainer().getConnectionConfiguration();
+            boolean showAllDatabases = CommonUtils.getBoolean(
+                configuration.getProviderProperty(MySQLConstants.PROP_SHOW_ALL_DBS),
+                MySQLConstants.PROP_SHOW_ALL_DBS_DEFAULT
+            );
+            String databaseName = getDatabaseName(configuration);
+            if (!showAllDatabases && CommonUtils.isNotEmpty(databaseName)) {
+                catalogFilters = new DBSObjectFilter();
+                catalogFilters.addInclude(databaseName);
+            }
```
ShadelessFox, line 877: Maybe you can instead filter the result set?

Nexus6v2: Why if we already have this logic in place?

ShadelessFox: Because we will fetch extra metadata that won't even be used. Moreover, this will show the "results are filtered" banner, no? It seems redundant.

Nexus6v2: No, "results are filtered" is not shown and this implementation works with filters
You also will see only the filters created manually in the filters config window

Nexus6v2: image
image
image
image
image

ShadelessFox: OK I guess

ShadelessFox (approved):

LonwoLonwo:
Is this option available for CloudBeaver?

Nexus6v2:
@LonwoLonwo The base change is in the model, so yes, it will be available in CB

ShadelessFox:
@Nexus6v2 it's not exposed via the xml though, so additional work is required to make it work in CB

Matvey16:
Verified

PR #38907, opened 2025-08-08
dbeaver/pro#5973 Exclude Eclipse browser preference page

ShadelessFox:
The page was moved before it could be removed. No need to move if it will be removed either way.


E1izabeth (approved):

Matvey16:
Verified

PR #38910, opened 2025-08-11
dbeaver/dbeaver#20379 Escape special regex characters in SQLUtils pattern matching logic

### LINKED ISSUE #20379: Column data filter does not work when values have plus character (`+`) prefix
### Description

`Filter by` column dialog returns no matches when value is prefixed with a `+` character. This is very frustrating when storing international phone numbers and trying to filter on them

For example, I have a cell containing values `+12345`. I can filter by `12345` or other substring that excludes the plus character, but as soon as I enter plus followed by anything, it returns no matches

### Broken use case: partial match including plus prefix character:
image

### Working use case: partial match on just prefix character:
image

### Working use case: partial match on substring excluding prefix character"
image

### DBeaver Version

Community Version 22.0.3.202204170718

### Operating System

macOS 13.3.1 (a) 

### Database and driver

Server: PostgreSQL 11.16
PostgreSQL 11.16 on x86_64-pc-linux-musl, compiled by gcc (Alpine 11.2.1_git20220219) 11.2.1 20220219, 64-bit

Driver: PostgreSQL JDBC Driver 42.2.5

### Steps to reproduce

1. Create DB with single value with a plus character prefix (e.g. `+123456`)
2. Try to filter this data based on the full value
Expected: Should return a match
Actual: Does not return a match

### Additional context

_No response_

labanovichttps:
image

```
CREATE TABLE phone_numbers (
    id int PRIMARY KEY,
    phone VARCHAR(20)
);

INSERT INTO phone_numbers (id, phone) VALUES (1, '+123456');
```


IvanGorshechnikov (approved):

Matvey16:
Verified

PR #38913, opened 2025-08-11
dbeaver/pro#6576 Fix complex name parts handling after falling back t…

E1izabeth:
Check that there is no exception in logs for 
```
SELECT
	*
FROM
	Employee e
WHERE
	e.EmployeeId > 0
	AND e.HireDate < '2025-01-01';
	
	
 select 'point(1 1)';
 
```


DariaMarkaryan:
Verified

PR #38914, opened 2025-08-11
dbeaver/pro#6390 Update Databricks meta model to improve catalog handling

labanovichttps:
STR in the ticket


DariaMarkaryan:
verified

PR #38948, opened 2025-08-14
Dbeaver/dbeaver#38591 fix mssql datatransfer

### LINKED ISSUE #38591: Export data table -> table with special colum names error
### Description

When exporting data from a table (by right-clicking on the table), the mapping between source and target columns breaks.

If the table has special column names containing \[ or ], the export/copy process tries to create new columns instead of offering mapping to existing columns. It seems to parse the column names incorrectly.

For example, the column is named:
rozdielAbs\[GJ]

But in the DDL and during export, the name becomes:
\[rozdielAbs\[GJ]]]

There is apparently an extra ] character.

### DBeaver Version

Comunity edition 25.1.1.202506221557

### Operating System

Windows 10

### Database and driver

1.  MS SQL Server / SQL Server
2.  Microsoft JDBC Driver 9.2 for SQL Server 9.2.0.0

### Steps to reproduce

1. right click on table
2. 'Export data' -> next
3. column mapping

Image

### Additional context

_No response_

labanovichttps:
STR in the ticket
simply try to export data from table with [] in the source table name, and in the target it shouldn't be `created`


E1izabeth (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified

PR #38950, opened 2025-08-15
dbeaver/pro#6610 Fix filtering of OpenAI models

ivgag:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/OpenAIModels.java
```diff
@@ -45,6 +40,9 @@ private OpenAIModels() {
         new AIModel("o1-pro", 200_000, Set.of(AIModelFeature.CHAT)),
         new AIModel("o1", 200_000, Set.of(AIModelFeature.CHAT)),
         new AIModel("o1-mini", 128_000, Set.of(AIModelFeature.CHAT)),
+        new AIModel("gpt-5", 400_000, Set.of(AIModelFeature.CHAT)),
```
serge-rider: Do we really need to hardcode these models?
Chat feature should be automatically discoverable from model ID.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotModels.java
```diff
@@ -44,22 +46,12 @@ private CopilotModels() {
...
-            return copilotModel.contextWindowSize();
-        }
-
-        AIModel openaiModel = OpenAIModels.KNOWN_MODELS.get(model);
-        if (openaiModel != null) {
-            return openaiModel.contextWindowSize();
-        }
-
-        return null;
+        return Optional.ofNullable(KNOWN_MODELS.get(model)).or(() -> OpenAIModels.getModelByName(model));
```
serge-rider, line 55: How come that Copilot uses OpenAI models?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/OpenAIProperties.java
```diff
@@ -45,6 +46,9 @@ public class OpenAIProperties implements OpenAIBaseProperties {
     @SerializedName("gpt.log.query")
     private Boolean loggingEnabled;
 
+    @SerializedName("gpt.streaming")
+    private Boolean streaming;
```
ShadelessFox: What about this?
```suggestion
    private boolean streaming = true;
```

plugins/org.jkiss.dbeaver.model.ai/plugin.xml
```diff
@@ -16,6 +16,7 @@
                 <property id="gpt.model.temperature" label="Temperature" type="integer" hint="Lower temperatures give more precise results"/>
                 <property id="gpt.log.query" label="Write AI queries to debug log" type="boolean" hint="Write GPT queries with metadata info in debug logs"/>
                 <property id="gpt.contextWindowSize" label="Context window size" type="integer" defaultValue="0" required="true"/>
+                <property id="gpt.streaming" label="Streaming is enabled" type="boolean" defaultValue="true" required="true"/>
```
ShadelessFox:
```suggestion
                <property id="gpt.streaming" label="Enable streaming" type="boolean" defaultValue="true" required="true"/>
```

ShadelessFox: Is it about the engine support for streaming, or the usage of streaming in the API?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotModels.java
```diff
@@ -44,22 +46,12 @@ private CopilotModels() {
         Function.identity()
     ));
 
-    @Nullable
-    public static Integer getContextWindowSize(@Nullable String model) {
+    @NotNull
+    public static Optional<AIModel> getModelByName(@Nullable String model) {
```
ShadelessFox, line 50: We don't usually use `Optional` in our codebase because we have nullability annotations. Can you please revert it back?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/dto/ChatCompletionRequest.java
```diff
@@ -71,7 +71,7 @@ public class ChatCompletionRequest {
      * The maximum number of tokens allowed for the generated answer. By default, the number of tokens the model can return will
      * be (4096 - prompt tokens).
      */
-    @SerializedName("max_tokens")
+    @SerializedName("max_completion_tokens")
```
ShadelessFox, line 74: This won't work well against old configuration files, no?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/IterablePublisher.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private final Iterable<T> source;
+
+
+    /**
+     * Publish all items from the given iterable using the provided executor.
+     */
+    public IterablePublisher(Iterable<T> source, Executor executor) {
+        this.source = Objects.requireNonNull(source, "source");
+        this.executor = Objects.requireNonNull(executor, "executor");
```
ShadelessFox: Use `@NotNull` instead of assertions

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/IterablePublisher.java
```diff
@@ -0,0 +1,179 @@
...
+            }
+        }
+
+        private void completeOnce() {
+            if (done.compareAndSet(false, true)) {
+                if (!cancelled.get()) {
+                    try {
+                        downstream.onComplete();
+                    } catch (Throwable throwable) {
+                        log.error("Error in subscriber onComplete", throwable);
```
ShadelessFox: `downstream.onError(t);`?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/IterablePublisher.java
ShadelessFox: Missing nullability throughout the class

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotModels.java
ivgag: Copilot uses OpenAI models

plugins/org.jkiss.dbeaver.model.ai/plugin.xml
ivgag: both

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotModels.java
ivgag: I don’t get why I shouldn't use optionals

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/dto/ChatCompletionRequest.java
ivgag: It's not a configuration file

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotModels.java
ShadelessFox: Judging by the usage, it's just plain redundant. Instead of `if`, you use lengthy `.ifPresentOrElse(then, otherwise)`.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/dto/ChatCompletionRequest.java
ShadelessFox: Didn't realize it's a DTO, my bad

plugins/org.jkiss.dbeaver.model.ai/plugin.xml
ShadelessFox: What about "Supports streaming"?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotModels.java
ivgag: I disagree

plugins/org.jkiss.dbeaver.model.ai/plugin.xml
ivgag: It depends on whether the user's account is verified on the openai platform. You can use streaming for the new models (o3, gpt-5) only after verification. Technically, all models support streaming, but it’s probably not available to the user yet

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/IterablePublisher.java
ivgag: From the API specification:

> If this method throws an exception, resulting behavior is undefined.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotModels.java
ShadelessFox: Yeah, sure, let's pollute the class file with two redundant lambdas that are also hard to read due to an extra indentation level.

ShadelessFox: And again, we utilize nullability annotations. Please stick to our code style documented here:

https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code

`Optional<T>` is a different matter of dealing with nulls introduced for a very particular reason. It doesn't fit here IMO.

plugins/org.jkiss.dbeaver.model.ai/plugin.xml
ShadelessFox: Okay... I can't think of a better name, so let's keep that one.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotModels.java
ivgag: And? The guidelines don't restrict its use. Every call to this method involves `if null then` checks anyway. Sure, you could write messy code anywhere, but I’d rather keep things clean and clear.

ShadelessFox: OK

ShadelessFox (dismissed):

PR #38956, opened 2025-08-15
#38352 Initially disable the OK button in the password dialog

### LINKED ISSUE #38352: Password expire dialog throws NPE with empty fields
### Description

Password expire dialog throws NPE with empty fields

### DBeaver Version

DBeaver CE 25.1.0

### Operating System

Win11

### Database and driver

MariaDB (version 10 or higher)

### Steps to reproduce

STR:
1. Login by a user to MariaDB (version 10 or higher)
2. Expire password for the user (`alter user 'test'@'%' password expire;`)
3. reconnect to the database
4. do not enter data to fields and press Ok

https://github.com/user-attachments/assets/c6649c5f-492a-465c-8b4a-c84d98531d38

[.log](https://github.com/user-attachments/files/20670460/default.log)
[dbeaver-debug.log](https://github.com/user-attachments/files/20670461/dbeaver-debug.log)

### Additional context

_No response_

ShadelessFox:


E1izabeth (approved):

IvanGorshechnikov (approved):

Matvey16:
I am not sure if this is the correct decision since blank password is valid in MariaDB and now it is impossible to set it like that

ShadelessFox:
It's clearly a bug, as it was possible to type anything in the password field and then erase it. You wouldn't be able to click the OK button.

E1izabeth:
But how to connect to db that has no password then?

Matvey16:
@E1izabeth Like this
image

It is valid in MariaDB. But I will create a separate ticket, because currently in devel OK button is inactive after you type anything and then delete it. We need to add an ability to save blank password separately

E1izabeth:
Won't we break the ability to connect to the database with the blank password?
If no, then OK

Matvey16:
It wont break the connection. But it definetely breaks the ability to change current password to blank from context menu here
image

ShadelessFox:
We need a new ticket that addresses that particular issue.

Matvey16:
Verified

PR #38972, opened 2025-08-18
dbeaver/pro#6264 Add afterReload listeners

vkirst:


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceProviderRegistry.java
```diff
@@ -761,5 +777,12 @@ public void saxEndElement(SAXReader reader, String namespaceURI, String localNam
         }
     }
 
+    public void addAfterReloadListener(@NotNull Runnable listener) {
+        afterReloadListeners.add(Objects.requireNonNull(listener));
+    }
+
+    public void removeAfterReloadListener(@NotNull Runnable listener) {
```
E1izabeth: Is it unused or am I missing something?

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceProviderRegistry.java
```diff
@@ -83,6 +84,9 @@ public synchronized static DataSourceProviderRegistry getInstance() {
     private final Map<String, DataSourceOriginProviderDescriptor> dataSourceOrigins = new LinkedHashMap<>();
     private final Map<String, DBPDriverSubstitutionDescriptor> driverSubstitutions = new HashMap<>();
 
+    private final List<Runnable> afterReloadListeners = new CopyOnWriteArrayList<>();
```
ShadelessFox: Please reuse `DBPRegistryListener` instead of adding a whole new set of listeners

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceProviderRegistry.java
ShadelessFox: I would assume it's for future use. Looks fine to me.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPRegistryListener.java
```diff
@@ -24,8 +24,16 @@
...
 public interface DBPRegistryListener {
 
-    void handleRegistryLoad(@NotNull DBPDataSourceRegistry registry);
+    default void handleRegistryLoad(@NotNull DBPDataSourceRegistry registry) {
 
-    void handleRegistryUnload(@NotNull DBPDataSourceRegistry registry);
+    }
+
+    default void handleRegistryUnload(@NotNull DBPDataSourceRegistry registry) {
+
```
ShadelessFox: Please add the following comment in the body for every default method:
```
// do nothing by default
```

ShadelessFox (approved):

PR #38974, opened 2025-08-18
dbeaver/pro#5806 Fix vqb save changes title

labanovichttps:
1. Open SQL Console
2. Switch to Builder
3. Make some changes
4. Disconnect from the database

image


labanovichttps:
It could also be fixed by overriding` getTitle()` method in VQBEditor part like 
```
    @Override
    public String getTitle() {
        String editorTitle = presentation.getSQLEditor().getTitle();
        if (CommonUtils.isNotEmpty(editorTitle)) {
            return editorTitle;
        }
        return super.getTitle();
    }
```

image

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/handlers/SaveChangesHandler.java
```diff
@@ -97,35 +97,33 @@ public void run()
...
                 Shell shell;
                 String saveableName;
-                if (saveable instanceof IWorkbenchPart) {
-                    shell = ((IWorkbenchPart) saveable).getSite().getShell();
-                    saveableName = ((IWorkbenchPart) saveable).getTitle();
+                if (saveable instanceof IWorkbenchPart workbenchPart) {
+                    shell = workbenchPart.getSite().getShell();
+                    saveableName = workbenchPart.getTitle();
+                    if (CommonUtils.isEmpty(saveableName)) {
+                        saveableName = workbenchPart.getTitleToolTip();
```
IvanGorshechnikov, line 104: Tooltip can be null or empty too? Or not?

labanovichttps: documentation:
`Returns: the workbench part title tool tip (not null)`

IvanGorshechnikov: I agree, but `String getTitle();` also has the same documentation: `Returns: the workbench part title tool tip (not null)`. So, can we trust the documentation?

labanovichttps: yes

IvanGorshechnikov:
I don't like this fix 
If we don't have Title, we use TitleTooltip. Why?
What if we don't have TitleTooltip?

labanovichttps:
> I don't like this fix If we don't have Title, we use TitleTooltip. Why? What if we don't have TitleTooltip?

For this case, we always have a tooltip

ShadelessFox (approved):

E1izabeth (approved):

PR #38975, opened 2025-08-18
dbeaver/dbeaver#38013 Check daylight to show timezone name in the status bar

### LINKED ISSUE #38013: Status line still shows standard time when daylight time in effect
### Description

The status bar shows the time zone abbreviation for standard time, even when daylight time is in effect; for example, this screenshot was taken on May 12, 2025, when Pacific Daylight Time (PDT) is in effect, but the status line shows Pacific Standard Time (PST):

Image

Hovering over the status line shows the tooltip “Pacific Standard Time.”

### DBeaver Version

Community Edition Version 25.0.4.202505041731

### Operating System

macOS Sequoia 15.4.1 (24E263)

### Database and driver

_No response_

### Steps to reproduce

Open DBeaver when daylight saving time is in effect and examine the status line.

### Additional context

It would seem that the hard-coded value of `false` here for the `daylight` argument is the problem:

https://github.com/dbeaver/dbeaver/blob/c60f27e49a1b16cee3cd5a9cfedfc2e0557e6cdf/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/ApplicationActionBarAdvisor.java#L357

IvanGorshechnikov:
For now, in Belgrade, it should show CEST 
image
If you change the date to winter, after a restart, it should show CET
image
Should work the same if you change the default timezone in the prefernces


labanovichttps (approved):

E1izabeth (approved):

uslss:
verified

PR #38991, opened 2025-08-19
dbeaver/pro#6342-clickhouse-maps-view removed maps from struct data kind

Nexus6v2:
Maps are processed as non-struct data kind (string values)
Having struct data kind prevented from viewing multiple values and changin to array was not possible due to driver returning the LinkedHashMap, which doesn't have `.toArray()` method (error was thrown from driver)
Upgrade of driver to 0.9.1 didn't solve the issue
image


IvanGorshechnikov (dismissed):

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseDataSource.java
```diff
@@ -263,7 +263,7 @@ public boolean isOmitCatalog() {
     public DBPDataKind resolveDataKind(@NotNull String typeName, int valueType) {
         if (typeName.startsWith("Array")) {
             return DBPDataKind.ARRAY;
-        } else if (ClickhouseTypeParser.isComplexType(typeName)) {
+        } else if (typeName.startsWith("Tuple")) {
```
serge-rider: Move Tuple and Array into ClickhouseConstats

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseDataSource.java
Nexus6v2: Done

Matvey16:
Verified

PR #38996, opened 2025-08-19
dbeaver/pro#6098 Add black/white text colors to the toolbar

labanovichttps:
image
image
image
image


IvanGorshechnikov (dismissed):

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageSettings.java
```diff
@@ -289,7 +289,9 @@ private void createProviderPage(Composite parent) {
 
                 var toolBar = new ToolBar(toolBarComposite, SWT.FLAT | SWT.RIGHT);
                 handlerItem = createHandlerItem(toolBar, allPages);
+                handlerItem.setForeground(UIStyles.getDefaultTextForeground());
                 profileItem = createProfileItem(toolBar);
+                profileItem.setForeground(UIStyles.getDefaultTextForeground());
```
E1izabeth: It's dirty. If we change the foreground, let's do it in one place. Or it should be the callee's responsibility.

ShadelessFox: It is. It's so strange to me that we praise the CSS but don't use it _at all_. The same thing could easily be done via CSS instead.

serge-rider:
We need generic solution.

Similar to this:
https://github.com/dbeaver/dbeaver/blob/devel/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/css/ConControlElementHandler.java#L49

It is still a bit of a hack but better than fixing some particular buttons.

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIStyles.java
```diff
@@ -277,4 +283,36 @@ private static float hueToRGB(float p, float q, float h) {
         }
         return p;
     }
+
+    /**
+     * Fix toolbar foreground color on macOS for toolbar managers
+     */
+    public static void fixToolBarForeground(Collection<ToolBarManager> toolbarManagers) {
```
ShadelessFox: Missing nullability annotations

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIStyles.java
```diff
@@ -277,4 +283,36 @@ private static float hueToRGB(float p, float q, float h) {
         }
         return p;
     }
+
+    /**
+     * Fix toolbar foreground color on macOS for toolbar managers
```
ShadelessFox: This javadoc is not descriptive at all

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIStyles.java
```diff
@@ -277,4 +283,36 @@ private static float hueToRGB(float p, float q, float h) {
...
+     */
+    public static void fixToolBarForeground(Collection<ToolBarManager> toolbarManagers) {
+        if (RuntimeUtils.isMacOS()) {
```
ShadelessFox: I suggest doing an early return so the code isn't that nested

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIStyles.java
```diff
@@ -277,4 +283,36 @@ private static float hueToRGB(float p, float q, float h) {
...
+                    for (ToolItem item : toolbar.getItems()) {
+                        item.setForeground(textColor);
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * Fix toolbar foreground color on macOS
```
ShadelessFox: Same issues here. Also, please reuse code

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIStyles.java
```diff
@@ -285,28 +285,22 @@ private static float hueToRGB(float p, float q, float h) {
...
             }
         }
     }
 
-    /**
-     * Fix toolbar foreground color on macOS
-     */
-    public static void fixToolBarForeground(ToolBar toolBar) {
-        if (RuntimeUtils.isMacOS() && toolBar != null && !toolBar.isDisposed()) {
+    public static void fixToolBarForeground(@NotNull ToolBar toolBar) {
```
ShadelessFox, line 302: If this method isn't called from outside, please mark it as `private`, as it no longer has the `RuntimeUtils.isMacOS()` check.

labanovichttps: It's called from outside

Matvey16:
Verified

PR #38998, opened 2025-08-20
dbeaver/dbeaver#38544 Fix mapping when have self link

### LINKED ISSUE #38544: ERD doesn't highlight FKs that reference tables in the public schema
### Description

Rather bizarre bug. The ERD will highlight columns that reference the selected column unless the table is in the public schema. The database being used is Postgres 14.9.

For example. Here I have 7 tables across 3 schemas. The public schema is in grey. When selecting the ID column in the user_profile table, it properly highlights the self referencing columns and the paths to the referencing tables but not the columns in those tables.

Image

It does correctly highlight every referencing column for tables not in the public schema as seen here with the value_map table. It correctly highlights columns for tables regardless of schema.

Image

### DBeaver Version

Community 25.1.0

### Operating System

Win 11

### Database and driver

Postgres 14.9

### Steps to reproduce

_No response_

### Additional context

SSH tunnel
No Extensions

IvanGorshechnikov:
image


E1izabeth (approved):

ShadelessFox (approved):

uslss:
verified

PR #39013, opened 2025-08-20
dbeaver/pro#6603 Added query preview to ai query confirmation dialogues

Nexus6v2:
Added query preview to ai query confirmation dialogues:

image
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/DesktopUI.java
+++ b/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/DesktopUI.java
@@ -31,11 +31,8 @@ import org.eclipse.swt.dnd.HTMLTransfer;
 import org.eclipse.swt.dnd.TextTransfer;
 import org.eclipse.swt.dnd.Transfer;
 import org.eclipse.swt.events.SelectionListener;
-import org.eclipse.swt.layout.FillLayout;
-import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Button;
-import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.swt.widgets.Shell;
 import org.eclipse.ui.IWorkbench;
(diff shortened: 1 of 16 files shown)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ui/DBPPlatformUI.java
```diff
@@ -82,6 +83,13 @@ public UserChoiceResponse(int choiceIndex, @Nullable Integer forAllChoiceIndex)
     boolean confirmAction(String title, String message, boolean isWarning);
     boolean confirmAction(@NotNull String title, @NotNull String message, @NotNull String buttonLabel, boolean isWarning);
 
+    boolean confirmScriptAction(
```
serge-rider: It is a bad idea to use SQL-specific classes in base platform UI.

It is better to add this function into UIServiceSQL. If UIServiceSQL is not installed then use confirmAction.

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/DesktopUI.java
```diff
@@ -287,6 +300,83 @@ public boolean confirmAction(@NotNull String title, @NotNull String message, @No
...
+            return super.confirmScriptAction(title, message, script, isWarning);
+        }
+        if (serviceSQL == null) {
+            serviceSQL = DBWorkbench.getService(UIServiceSQL.class);
+        }
+
+        final Reply[] reply = {null};
+        final boolean[] isExpanded = {false};
+
+        UIUtils.syncExec(() -> reply[0] = MessageBoxBuilder.builder(UIUtils.getActiveWorkbenchShell())
```
serge-rider: Looks too complicated. Let's just add extra button and show ViewSQLDialog when user clicks it.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/DesktopUI.java
Nexus6v2: That's sad. It looked pretty good on UI...
Also that is an exact approach we've discussed from the start

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ui/DBPPlatformUI.java
Nexus6v2: I can't use `UIServiceSQL` from `org.jkiss.dbeaver.model` where the confirmation logic is located
Unless I'm missing something

Nexus6v2: I updated the method to be less SQL-specific

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/DesktopUI.java
Nexus6v2: Replaced with `openSQLViewer`

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ui/DBPPlatformUI.java
Nexus6v2: Moved to `UIServiceSQL`

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/AIUtils.java
```diff
@@ -174,7 +180,29 @@ private static boolean isConfirmationNeeded(@NotNull String actionName) {
...
+
+    @NotNull
+    private static DataSourceContextProvider getContextProvider(@NotNull List<SQLScriptElement> script) {
+        SQLScriptElement scriptElement = script.stream().findFirst().orElse(null);
+        DBPDataSource dataSource = null;
+        if (scriptElement instanceof SQLQuery query) {
+            dataSource = query.getDataSource();
+        } else if (scriptElement instanceof SQLControlCommand command) {
+            dataSource = command.getDataSource();
+        }
```
E1izabeth: Isn't it better to pull up `getDataSource` method to the SQLScriptElement class?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ui/UIServiceSQL.java
```diff
@@ -93,8 +93,13 @@ Object createSQLPanel(Object site, Object parentControl, DBPContextProvider cont
...
 
-    void openResource(IResource element);
-
     boolean useIsolatedConnections(DBPContextProvider contextProvider);
 
+    boolean confirmQueryExecution(
+        @NotNull String title,
+        @NotNull String message,
+        @NotNull String queryText,
+        @NotNull DataSourceContextProvider contextProvider,
```
serge-rider: DBPContextProvider?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/AIUtils.java
Nexus6v2: Good point! Done

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/ui/UIServiceSQL.java
Nexus6v2: Done

E1izabeth (dismissed):

IvanGorshechnikov (dismissed):

PR #39015, opened 2025-08-20
dbeaver/pro#6655 gql cli improvements

alexander-skoblikov:
closes dbeaver/pro#6655


plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/DBeaverLauncher.java
```diff
@@ -758,7 +758,13 @@ private boolean processCommandLineAsClient(String[] args, Path dbeaverDataDir) t
             }
 
             if (output != null && !output.isEmpty()) {
-                output = output.replace("\\n", "\n");
+                output = output
```
serge-rider, line 762: ??
please add comments

serge-rider (changes requested):

PR #39023, opened 2025-08-21
dbeaver/pro#6318 Reset database drivers on configuration change

ivgag:


plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/driver/DriverEditDialog.java
```diff
@@ -628,6 +630,8 @@ private void addMavenArtifact() {
             libraries.addAll(fd.getArtifacts());
             changeLibContent();
         }
+
+        driver.resetDriverInstance();
```
serge-rider: Reset before save?
I think we can reset instance only if model changes (DriverDescriptor)

alexander-skoblikov (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/driver/DriverEditDialog.java
```diff
@@ -908,6 +914,10 @@ protected void okPressed() {
             return;
         }
 
+        if (oldDriver != null) {
+            driver.resetDriverInstance();
```
serge-rider: We reset it two times now. And the second is unconditional. Is it ok?

serge-rider (changes requested):

PR #39025, opened 2025-08-21
dbeaver/pro#6671 Make streaming for OpenAI models optional

ivgag:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/IterablePublisher.java
```diff
@@ -0,0 +1,178 @@
...
+import org.jkiss.dbeaver.Log;
+
+import java.util.Iterator;
+import java.util.concurrent.Executor;
+import java.util.concurrent.Flow;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+public final class IterablePublisher<T> implements Flow.Publisher<T> {
```
serge-rider, line 29: Is it used anywhere?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotCompletionEngine.java
```diff
@@ -99,7 +99,7 @@ public AIEngineResponse requestCompletion(
 
     @NotNull
     @Override
-    public void requestCompletionStream(
+    public void doRequestCompletionStream(
```
serge-rider: Let's not rename base method.
We call requestCompletionStream is one method only (in AIAssistantProImpl) and we can do all changes only there. So won't change a AI engines at all.

serge-rider (changes requested):

PR #39038, opened 2025-08-22
dbeaver/pro#6607 Added navigator filters tip, resized filter tables

Nexus6v2:
* Added a tip in the filters dialog explaining that filtered objects won't be visible anywhere
* Reduced tables size to take less space
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/internal/UINavigatorMessages.properties
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/internal/UINavigatorMessages.properties
@@ -44,6 +44,7 @@ dialog_filter_remove_button = Remove
 dialog_filter_save_label = Saved filter
 dialog_filter_name_label = Name
 dialog_filter_hint_text = You can use masks (%, _ and *) in filters
+dialog_filter_objects_scope_hint_text = Filtered objects won't be visible in any scope\n(e.g., Database Navigator, Table Editor, Auto-complete)
 
 actions_navigator_search_tip = Enter a part of object name here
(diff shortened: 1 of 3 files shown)


arhayka:
Looks great! Just a small correction:
- we don’t call it Table Editor — it’s Properties or Metadata Editor
- it’s important to highlight that this limitation also applies to AI
maybe something like this?

--
Filtered objects will not be visible in any context, 
including the Navigator, Metadata Editor, AI Chat, and Autocomplete

--
Filtered objects are hidden in Navigator, Metadata Editor, 
AI Chat, and Autocomplete

Nexus6v2:
@arhayka Thanks for a comment, I've updated the copy

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/internal/UINavigatorMessages.properties
```diff
@@ -44,6 +44,7 @@ dialog_filter_remove_button = Remove
 dialog_filter_save_label = Saved filter
 dialog_filter_name_label = Name
 dialog_filter_hint_text = You can use masks (%, _ and *) in filters
+dialog_filter_objects_scope_hint_text = Filtered objects will not be visible in any context, including\nNavigator, Metadata Editor, AI-chat and Autocomplete
```
E1izabeth: Not only "Autocomplete". All SQL Editor services

Nexus6v2: Updated

E1izabeth (approved):

uslss:
verified
image

PR #39042, opened 2025-08-22
dbeaver/dbeaver#37044 Fix history to script execute

### LINKED ISSUE #37044: When switching to a record by FK, the display of the value breaks
### Description

Image

### DBeaver Version

CE 24.3.4.202502021521

### Operating System

Windows 10 x64

### Database and driver

postgres-jdbc

### Steps to reproduce

1) execute a query that has a reference field by FK 
2) in the icon window, the associated row is normally discarded
3) Navigate to a row in the Linked Table. The transition is performed correctly
4) Click on the click button. This will be the first error, it returns not to the original query, but to the reference table without the disabled filter.
5) Execute the original query
6) On a column with an FK, the ability to switch to a connected row disappears. The values window displays only the ID, without the associated tables.

### Additional context

_No response_

IvanGorshechnikov:
STR:
1. Open Script (reopen if you executed something before)
2. Run as Execute SQL Script
3. Move to the referenced table
4. When you try to go back, it should go to wrong table (it is fixed)
5. Re-execute SQL Script, youn't have to go to referenced table (it is fixed too)


uslss:
verified

PR #39055, opened 2025-08-25
dbeaver/pro#6302 AI model refactoring

serge-rider:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotCompletionEngine.java
```diff
@@ -48,13 +49,20 @@ protected void onDispose(@NotNull CopilotClient disposedValue) {
             disposedValue.close();
         }
     };
+    private CopilotSessionToken sessionToken;
```
ivgag, line 50: always null

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/prompt/AIPromptGenerateSql.java
```diff
@@ -0,0 +1,89 @@
...
+import org.jkiss.code.NotNull;
+import org.jkiss.dbeaver.model.ai.AIConstants;
+import org.jkiss.dbeaver.model.ai.AISqlJoinRule;
+import org.jkiss.dbeaver.model.ai.impl.AIPromptUtils;
+import org.jkiss.dbeaver.model.logical.DBSLogicalDataSource;
+import org.jkiss.dbeaver.model.logical.DBSLogicalDataSourceSupplier;
+import org.jkiss.dbeaver.runtime.DBWorkbench;
+import org.jkiss.utils.CommonUtils;
+
+public class AIPromptGenerateSql extends AIPromptAbstract {
```
ivgag, line 28: The name doesn’t reflect the purpose of the prompt - it’s not just about generating SQL

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/impl/AIAssistantImpl.java
```diff
@@ -40,223 +43,67 @@ public class AIAssistantImpl implements AIAssistant {
...
-            .addContexts(AIPromptBuilder.describeContext(request.context().getDataSource()))
-            .addInstructions(AIPromptBuilder.createInstructionList(request.context().getDataSource()))
-            .addGoals(
-                "Translate natural language text to SQL."
-            )
-            .addOutputFormats(
-                "Place any explanation or comments before the SQL code block.",
-                "Provide the SQL query in a fenced Markdown code block."
+        try (AIEngine engine = createEngine()) {
+            String systemPrompt = systemGenerator.build();
```
ivgag, line 78: You pass AIPromptGenerator here just to call .build(). Wouldn’t it be possible to pass it as a String instead?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/prompt/AIPromptGenerateSql.java
serge-rider: Do you have any better name on mind?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/impl/AIAssistantImpl.java
serge-rider: Nope, we also use prompt ID to save it in chat history (in chat mode).
We also may wish to add extra parameters to prompts in future (e.g. info labels)

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotCompletionEngine.java
serge-rider: Old code, fixed

Matvey16:
Incorrect AI describe for schemas from non-default databse in PostgreSQL. You can see that the prompt is different, but there may be another cause
STR:
1. Expand PostgreSQL connection in the navigator
2. Expand any non-default database
3. Right-click on any schema and select `Describe database object`

Result in devel branch (table names are correct)

dbeaver_LGGObAQZBF

Result on this branch for a schema `public1` when there is a schema with a name `public` in default database (tables in the description are from the default database)

dbeaver_KFjV0n2lKe

Result in this branch for a schema with completely unique name (all tables in the description do not exist in the schema that is being described)

dbeaver_WiQANEHhSk





Also reproducible for tables

image

Matvey16:
Verified

PR #39056, opened 2025-08-25
feat: Add GeoJSON export option to Data Transfer

### LINKED ISSUE #39104: Feat: Add GeoJSON export option to Data Transfer
Currently, DBeaver users who work with spatial data often face the problem of not being able to export their geographic information directly into a widely recognized and interoperable format for GIS and web mapping applications. This forces users to either:
1.  Manually convert exported data (e.g., CSV with WKT/WKB) into GeoJSON using external tools or custom scripts.
2.  Use other database clients or GIS software for the initial data export, which adds an unnecessary step to their workflow and reduces DBeaver's utility for spatial data management.

This creates friction and reduces efficiency for users who need to move spatial data from their database into environments like QGIS, Leaflet, Mapbox, or other GeoJSON-consuming applications.

**Solution:**

I would like to add a new `GeoJSON` export option to DBeaver's existing Data Transfer framework. This solution would:
*   Introduce a new `DataExporterGeoJSON` class capable of exporting database rows as a GeoJSON `FeatureCollection`.
*   Automatically detect a geometry column (preferably by spatial data type, falling back to a column named 'geometry' case-insensitively).
*   Convert the database's spatial data (e.g., WKT, WKB) into valid GeoJSON geometry objects (Point, LineString, Polygon, etc.).
*   Export all other non-geometry columns as `properties` within each GeoJSON `Feature`.
*   Ensure proper JSON formatting, escaping, and UTF-8 encoding.
*   The generated GeoJSON file should be immediately usable in common GIS tools and web mapping libraries without further conversion.

**Alternatives:**
*   **Existing Export Formats:** While DBeaver offers CSV, JSON, XML, etc., none of these directly support the GeoJSON standard for spatial data. Exporting to a generic JSON and then manually post-processing it to conform to GeoJSON is inefficient and prone to errors.
*   **External Tools:** Relying on external GIS software (like QGIS, PostGIS `ST_AsGeoJSON`) to export data is a common workaround, but it means switching applications and doesn't leverage DBeaver's data transfer capabilities directly.
*   **WKT/WKB Export to CSV:** Exporting spatial data as Well-Known Text (WKT) or Well-Known Binary (WKB) strings in a CSV or text file requires an additional processing step to parse and convert these strings into the GeoJSON object structure, which is the exact problem this feature aims to solve.

**Additional context**
*   This feature is already under development in Pull Request [#39056](https://github.com/dbeaver/dbeaver/pull/39056).
*   The implementation uses `com.google.gson.stream.JsonWriter` for robust JSON construction and proper escaping.
*   Initial testing has shown successful export of `Point` geometries and associated properties.
*   This feature significantly enhances DBeaver's capabilities for users working with geographic information systems and spatial databases, making it a more complete tool for managing and transferring spatial data.

RohitYadav34980:
## ✨ Summary

This PR adds a new **GeoJSON** export option to DBeaver’s Data Transfer framework.

---

## 🔄 Changes

* Added a new exporter class: `DataExporterGeoJSON`
* Registered processor `stream.geojson` in `plugin.xml`
* Updated `MANIFEST.MF` dependencies
* Exports rows as [[GeoJSON](https://geojson.org/)]`FeatureCollection`

---

## 🛠 Implementation Details

* Detects a column named `geometry` (case-insensitive) and uses it as the `geometry` field.
* All other columns are exported as `properties`.
* If no geometry column is present, `"geometry": null` is written.
* Uses UTF-8 encoding and `JSONUtils` for safe string escaping.
* Exported output is compatible with GIS tools like QGIS, Leaflet, and Mapbox.

---

## 📄 Sample Output

```json
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {"type":"Point","coordinates":[77.5946,12.9716]},
      "properties": {
        "id": 1,
        "name": "Bangalore"
      }
    }
  ]
}
```

---

## ✅ Why This is Useful

GeoJSON is a widely used format in GIS and web mapping applications.
With this new export option, DBeaver users can directly export spatial data without needing external conversion tools.

---

## 🧪 Testing

* Export tested with a table containing `geometry`, `id`, and `name` columns.
* Generated GeoJSON validated in QGIS and online GeoJSON validator.
* Verified other exporters (CSV, JSON, SQL, etc.) remain unaffected.

---

## 📸 Screenshots
{084B6B62-FB45-47F8-A025-AFF494EBA4AE}

## Output file
https://drive.google.com/file/d/1TxLoqvDgC6I6tyQ_yHWomALjEdF4srrE/view?usp=sharing

Fixes #39104


plugins/org.jkiss.dbeaver.core/META-INF/MANIFEST.MF
```diff
@@ -51,7 +51,8 @@ Require-Bundle: org.osgi.service.event,
  org.jkiss.dbeaver.ui.editors.base;visibility:=reexport,
  org.jkiss.dbeaver.ui.editors.connection;visibility:=reexport,
  org.jkiss.dbeaver.ui.editors.sql;visibility:=reexport,
- org.jkiss.dbeaver.ui.navigator;visibility:=reexport
+ org.jkiss.dbeaver.ui.navigator;visibility:=reexport,
+ org.apache.commons.lang3
```
ShadelessFox: Unused bundle

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+        this.columns = List.of(attrs);
+        // Find geometry column index (case-insensitive)
+        for (int i = 0; i < columns.size(); i++) {
+            if ("geometry".equalsIgnoreCase(columns.get(i).getName())) {
+                geometryIndex = i;
+                break;
+            }
+        }
+        writer.write("{\"type\": \"FeatureCollection\", \"features\": [\n");
```
ShadelessFox: Please use Gson's JsonWriter instead of manually constructing JSON.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -0,0 +1,97 @@
...
+        this.geometryIndex = -1;
+        this.columns = null;
+    }
+
+    @Override
+    public void exportHeader(DBCSession session) throws IOException {
+        DBDAttributeBinding[] attrs = site.getAttributes();
+        if (attrs == null) {
+            throw new IllegalStateException("No attributes provided for export");
+        }
```
ShadelessFox: It's never null

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -0,0 +1,97 @@
...
+        this.columns = List.of(attrs);
+        // Find geometry column index (case-insensitive)
+        for (int i = 0; i < columns.size(); i++) {
+            if ("geometry".equalsIgnoreCase(columns.get(i).getName())) {
```
ShadelessFox: Why do we assume there's a column called `geometry` in the source table? What's the semantics behind this logic? Can we rely on its data type instead?

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -0,0 +1,97 @@
...
+    public void exportRow(DBCSession session, DBCResultSet resultSet, Object[] row) throws IOException {
+        if (!firstFeature) writer.write(",\n");
+        firstFeature = false;
+
+        writer.write("{\"type\": \"Feature\", ");
+
+        // Geometry
+        if (geometryIndex >= 0 && row[geometryIndex] != null) {
+            writer.write("\"geometry\": ");
+            writer.write(row[geometryIndex].toString());
```
ShadelessFox: Why do we assume the contents of that column represent geometry data? How can we be sure that its `toString` conversion produces valid data that won't bork the output JSON because you don't even escape it?

ShadelessFox (changes requested):
It would be nice to have actual tests for this new feature. See `org.jkiss.dbeaver.tools.transfer.DataExporterCSVTest`

And please update your PR according to our contribution guide: https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code#create-pull-request. Having an issue for a PR is mandatory.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
RohitYadav34980: To export geometry to json we need to use a function `ST_AsGeoJSON`. The datatype of the column is mostly `txt`.
so we cannot rely on its datatype.

ShadelessFox: Neither can we rely on its name.

RohitYadav34980: suggest me how we can fix it

ShadelessFox: The simplest way is to allow users to specify the column that contains geometry data, as a fallback if the proposed algorithm fails.
We could make assumptions based on a spec describing how tables are structured around geometry data, but I don't think there's one. Is it even standardized?

ShadelessFox:
Your issue claims things that aren't a thing in this PR. Are you still working on it?

> Convert the database's spatial data (e.g., WKT, WKB) into valid GeoJSON geometry objects (Point, LineString, Polygon, etc.).

You are not converting anything, just emitting raw values as strings that you don't sanitize in any way. That likely won't work well with WKB values. See how other exporters extract row values.

> Ensure proper JSON formatting, escaping, and UTF-8 encoding.

I am really concerned about that because you are constructing JSON from a raw string, assembled from parts that aren't sanitized. You could parse the geometry data as JSON to ensure it's valid and only then emit it.

> The implementation uses `com.google.gson.stream.JsonWriter` for robust JSON construction and proper escaping.

No, it doesn't. :^)

RohitYadav34980:
> Convert the database's spatial data (e.g., WKT, WKB) into valid GeoJSON geometry objects (Point, LineString, Polygon, etc.).

A PostGIS function [`ST_AsGeoJSON`](https://postgis.net/docs/ST_AsGeoJSON.html) exactly do this.

> > Ensure proper JSON formatting, escaping, and UTF-8 encoding.
> 
> I am really concerned about that because you are constructing JSON from a raw string, assembled from parts that aren't sanitized. You could parse the geometry data as JSON to ensure it's valid and only then emit it.

`ST_AsGeoJSON` return a `json` so no need to concern about that.
> 
> > The implementation uses `com.google.gson.stream.JsonWriter` for robust JSON construction and proper escaping.
> 
> No, it doesn't. :^)

It breaks the build. :(

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
RohitYadav34980: To use this feature we first need to use `ST_AsGeoJSON` function to export it as GeoJSON and user can can alias column with `geometry` in sql itself.

ShadelessFox: I don't understand what you are trying to say.

ShadelessFox:
> A PostGIS function [ST_AsGeoJSON](https://postgis.net/docs/ST_AsGeoJSON.html) exactly do this.

You don't use `ST_AsGeoJSON`.

> ST_AsGeoJSON return a json so no need to concern about that.

Again, you don't use `ST_AsGeoJSON`.

You assume things. You also assume that the database is PostgreSQL with PostGIS extension enabled. What about other databases?

In reality, geometry data can be stored in any format. It could be WKT, WKB, EWKB, or any other proprietary format. DBeaver can read most of these formats and represent them as `DBDGeometry`. Ideally, you need to obtain it from a value handler and convert it to a GeoJSON-compatible format.

> It breaks the build. :(

Because you need to add a dependency. See the `MANIFEST.MF`

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -0,0 +1,139 @@
...
+    @Override
+    public void exportHeader(DBCSession session) throws IOException {
+        DBDAttributeBinding[] attrs = site.getAttributes();
+        this.columns = List.of(attrs);
+        // Find geometry column index by type (case-insensitive, contains "geometry", "geojson", "geography", or "geom")
+        for (int i = 0; i < columns.size(); i++) {
+            String typeName = columns.get(i).getTypeName();
+            if (typeName != null) {
+                String typeLower = typeName.toLowerCase();
+                if (typeLower.contains("geometry") || typeLower.contains("geojson") || typeLower.contains("geography") || typeLower.contains("geom")) {
```
serge-rider: Please move types names in constants. Or even better pass them in export config (it may be a property geoDataTypes and have defaulkt value `geometry,geojson,geography,geometry`).

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -0,0 +1,139 @@
...
+                if (colName != null) {
+                    String n = colName.trim().toLowerCase();
+                    if (n.equals("geometry") || n.equals("geom") || n.equals("the_geom") || n.equals("geography") || n.equals("geojson") || n.equals("shape") || n.equals("wkb_geometry")) {
+                        geometryIndex = i;
+                        break;
+                    }
+                }
+            }
+        }
+        out.write("{\"type\": \"FeatureCollection\", \"features\": [\n");
```
serge-rider: GeoJson is a json. So maybe it is better to use JsonWriter that playing with strings?

As an example you can check TaskManagerImpl class where we use JsonWriter. It works in stream mode.

plugins/org.jkiss.dbeaver.data.transfer/plugin.xml
```diff
@@ -14,35 +14,35 @@
...
 
         <node type="consumer"
-            id="database_consumer"
-            class="org.jkiss.dbeaver.tools.transfer.database.DatabaseTransferConsumer"
-            icon="platform:/plugin/org.jkiss.dbeaver.model/icons/tree/table.svg"
-            label="%dataTransfer.consumer.database.name"
-            description="%dataTransfer.consumer.database.description"
-            settings="org.jkiss.dbeaver.tools.transfer.database.DatabaseConsumerSettings"
-            advanced="true">
+              id="database_consumer"
```
serge-rider: Please do not do unrelated changes. Let's keep original formatting.

serge-rider (changes requested):

serge-rider:
@ShadelessFox I think it is a simple way to export json as string representation and it works.
So we can keep as is in the first implementation. Later we ca improve it and use some database-specific adapters to do this export better.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -17,123 +17,258 @@
...
+        jsonWriter.beginObject();
+        jsonWriter.name("type").value("FeatureCollection");
+        jsonWriter.name("features");
+        jsonWriter.beginArray();
+
+        // Flush to ensure header is written immediately (needed for tests)
+        jsonWriter.flush();
+        if (writer != null) {
+            writer.flush();
+        }
```
ShadelessFox: You don't need that for tests. You are emitting an incorrect JSON. Moreover, there are no guarantees that `flush` will actually _flush_ all pending data to the stream. Just call `exportFooter` followed by `dispose` before checking the results.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -17,123 +17,258 @@
...
         // Geometry
-        out.write("      \"geometry\": ");
-        if (geometryIndex >= 0 && row[geometryIndex] != null) {
-            String geomStr = row[geometryIndex].toString();
-            if (geomStr.trim().startsWith("{") || geomStr.trim().startsWith("[")) {
-                out.write(geomStr);
+        jsonWriter.name("geometry");
+        Object geometryValue = (geometryIndex >= 0) ? row[geometryIndex] : null;
+        if (geometryValue instanceof DBGeometry) {
+            Object gisData = ((DBGeometry) geometryValue).getRawValue();
```
ShadelessFox: We're using Java 21, not Java 8. Same for other occurrences.


```suggestion
        if (geometryValue instanceof DBGeometry geometry) {
            Object gisData = geometry.getRawValue();
```

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -17,123 +17,258 @@
...
 
     @Override
     public void dispose() {
-        // No resources to clean up
+        try {
+            if (jsonWriter != null) {
+                jsonWriter.close();
+            }
+        } catch (IOException e) {
+            // Ignore
```
ShadelessFox: Please log the exception

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -0,0 +1,265 @@
...
+
+    // Gson instance configured for GeoJSON output
+    private static final Gson GEOJSON_GSON = new GsonBuilder()
+        .disableHtmlEscaping()
+        .create();
+
+    private PrintWriter writer;
+    private JsonWriter jsonWriter;
+    private List<DBDAttributeBinding> columns;
+    private boolean firstFeature = true;
```
ShadelessFox: Not used

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -0,0 +1,265 @@
...
+        this.site = site;
+        this.writer = site.getWriter();
+        try {
+            this.jsonWriter = GEOJSON_GSON.newJsonWriter(this.writer);
+        } catch (IOException e) {
+            log.error("Failed to initialize JSON writer", e);
+            throw new DBException("Failed to initialize JSON writer", e);
+        }
+        // Configure JsonWriter for the exact format expected by tests
+        this.jsonWriter.setIndent(""); // No indentation for compact output
```
ShadelessFox, line 79: I think it should be configurable. Could be implemented in a separate ticket.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -0,0 +1,265 @@
...
+                            if (parsed instanceof Map<?, ?> parsedMap) {
+                                writeMap((Map<String, Object>) parsedMap);
+                            } else {
+                                jsonWriter.nullValue();
+                            }
+                        } catch (Exception e) {
+                            log.error("Error parsing geometry JSON string: " + geoString, e);
+                            jsonWriter.nullValue();
+                        }
+                    }
```
ShadelessFox: Duplicate of `case String geoString`

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -0,0 +1,265 @@
...
+            var name = columns.get(i).getName();
+            var value = row[i];
+
+            jsonWriter.name(name);
+            switch (value) {
+                case null -> jsonWriter.nullValue();
+                case Number number -> jsonWriter.value(number);
+                case Boolean bool -> jsonWriter.value(bool);
+                default -> jsonWriter.value(value.toString());
+            }
```
ShadelessFox: Use `writeValue`

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterGeoJSON.java
```diff
@@ -0,0 +1,265 @@
...
+            case Map<?, ?> map -> writeMap((Map<String, Object>) map);
+            case String geoString -> {
+                try {
+                    var parsed = GEOJSON_GSON.fromJson(geoString, Map.class);
+                    if (parsed instanceof Map<?, ?> parsedMap) {
+                        writeMap((Map<String, Object>) parsedMap);
+                    } else {
+                        jsonWriter.nullValue();
+                    }
```
ShadelessFox: Why does this not handle anything that is not a `Map`? Seems like you can use `writeValue` here

ShadelessFox (commented):
Please add missing `@NotNull`/`@Nullable` annotations throughout the code

IvanGorshechnikov (approved):

RadAnton:
Testing results:

I tested exporting GeoJSON from several test tables containing different geometry types (Point, LineString, Polygon)
Each table contained at least the columns `id`, `name` and `geometry`.

Issue: The geometry field in all exported files is not a valid GeoJSON object.

It currently contains strings: either WKT ("POINT(30 50)") or JSON text ("{\"type\":\"Polygon\",\"coordinates\":[...]}").

Even using `ST_AsGeoJSON(geometry)` in the `SELECT`, DBeaver exports it as a string, not a proper JSON object.

As a result, the exported files cannot be directly loaded in GIS tools.

Export file example:
```
{
    "type": "FeatureCollection",
    "features": [
        {
            "type": "Feature",
            "geometry": "POLYGON ((30.3 59.93, 30.37 59.93, 30.37 59.95, 30.3 59.95, 30.3 59.93))",
            "properties": {
                "id": 1,
                "name": "Polygon 1"
            }
        }
    ]
}
```
```
{
    "type": "FeatureCollection",
    "features": [
        {
            "type": "Feature",
            "geometry": "{\"type\":\"Polygon\",\"coordinates\":[[[30.3,59.93],[30.37,59.93],[30.37,59.95],[30.3,59.95],[30.3,59.93]]]}",
            "properties": {
                "id": 1,
                "name": "Polygon 1"
            }
        }
    ]
}
```

arhayka:
Hi @RohitYadav34980, it looks like this feature is not working as expected. We can't merge this PR now.

PR #39077, opened 2025-08-27
dbeaver/pro#6705 Filter nulls in SQL name list processing

labanovichttps:


E1izabeth (changes requested):
That's not the right approach. Search utils can't and should not find by partial object names, so we should not handle nulls there at all. It should be handled earlier.
If we're talking about semantic analysis, there is a sign that the name is not fully or correctly typed `invalidPartsCount`.

E1izabeth:
This should be done like this https://github.com/dbeaver/dbeaver/pull/39078
Please take a look and test.

PR #39078, opened 2025-08-27
dbeaver/pro#6705 Handle partial table name in `create table` query model

E1izabeth:
the issue is not specific for Databricks
reproduce like `create table abc.`


labanovichttps (dismissed):

labanovichttps:
tested, for me it works fine

E1izabeth:
Also fixed that not only tables came in the following case - now only tables should come
image

Matvey16:
Verified

PR #39079, opened 2025-08-27
dbeaver/pro#6173 Fixed new scripts schema reset on reconnect

Nexus6v2:
When reconnecting in a new script with a non-default schema, now selected schema is persisted.
It was reproducible only in new scripts, was auto-fixing after the save. Now works fine as is.
1. Create new script
image
2. Select schema
image
3. Disconnect
image
4. Reconnect, schema is not reset to default
image


E1izabeth (approved):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -425,12 +425,12 @@ public DBPDataSourceContainer getDataSourceContainer() {
 
     @Override
     public boolean setDataSourceContainer(@Nullable DBPDataSourceContainer container) {
-        if (!datasourceChanged && curDataSource != null) {
-            datasourceChanged = true;
-        }
         if (container == dataSourceContainer) {
             return false;
         }
+        if (!datasourceChanged && curDataSource != null) {
```
serge-rider, line 431: Pleaser write a comment.
The logic is not clear (we always trigger container change even if it is the same container?)

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
Nexus6v2: We did that. 
Now we quit (line 429) if it's the same container, which to me seems more logical.
To be honest, I don't think any comment is needed here.

Matvey16:
Verified

PR #39082, opened 2025-08-27
dbeaver/dbeaver#37066 Refactor SQL parsing logic to support multiple queries detection and improve parser initialization handling

### LINKED ISSUE #37066: Running a SELECT query along with a CREATE TABLE/VIEW not respecting ResultSet fetch size
### Description

If I have a query that creates a temp table/view from a large table (millions of rows), then a select immediately after, DBeaver would try fetching all rows regardless of `ResultSet fetch size` in settings.
However, if I run the SELECT individually, it seems fine.

For example:
```
CREATE OR REPLACE TEMP VIEW v_large AS
	SELECT * FROM my_very_large_table;

SELECT * FROM v_large;
```

This would never complete as it fills the JVM memory quickly and JDBC would throw an error at the end.

Running them separately would not have any issue as it respects the `ResultSet fetch size` 
Or if I explicitly add `SELECT * FROM v_large LIMIT 100;` then it works fine.

### DBeaver Version

Community Version 24.3.4.202502021521

### Operating System

Windows 10

### Database and driver

PostgreSQL 9.4.26
PostgreSQL JDBC Driver 42.7.3

### Steps to reproduce

Select all of this in the editor and click run button:
```
CREATE OR REPLACE TEMP VIEW v_large AS
	SELECT * FROM my_very_large_table;

SELECT * FROM v_large;
```

Wait and see the JVM out of memory.

Image

### Additional context

_No response_

labanovichttps:
STR in the ticket


E1izabeth (approved):

uslss:
verified

PR #39084, opened 2025-08-27
#39017 Don't load associations and relations from completion logic

### LINKED ISSUE #39017: After last update, when i open dbeaver, it works "Analyzing query" very slow
### Description

After last update, when i open dbeaver, it works "Analyzing query" very slow and all mi PC works very slow too, honestly every time I download an update, ther are new issues, please fix it, is impossible use the program like that.

Thank you very much.

### DBeaver Version

25.1.5.202508171444

### Operating System

Windows 11

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

E1izabeth:
Now we don't put tables having FKs with other tables in the query if FKs were not loaded.
How to test:
1) SQLite sample database, write `select * from Playlist join |` - tables are in alphabetical order
image

2) Load FKs for Playlist or PlaylistTrack by expanding nodes in Navigator (or by any other way)
image

3) Call autocomplete again - now PlaylistTrack is at the top, because have FK with Playlist
image


Matvey16:
Verified

PR #39096, opened 2025-08-28
dbeaver/dbeaver#39092 Refactor MySQL foreign key query logic to improve readability and encapsulation

### LINKED ISSUE #39092: DBeaver version 25.2.1 slow on first reading of tables
### Description

The first table read, even if they contain no content, takes a minimum of 5 seconds and sometimes reaching 30 seconds.
This only happens on the first data read, but it's annoying because it didn't happen before.

Downgrading to version 25.1.4 resolves the issue where reading a table takes a millionth of a second

### DBeaver Version

25.2.1

### Operating System

Debian GNU/Linux 12 (bookworm)

### Database and driver

DriverName: Mariadb
DriverType: MySQL
Version: 10.5.10

### Steps to reproduce

Open a table for the first time. After that, it doesn't take long.

### Additional context

_No response_

labanovichttps:
Try to load fk and resultSet Navigate -> References


uslss:
verified

PR #39110, opened 2025-09-01
dbeaver/pro#4392 Fixed DND to/form Cloud Storage in TE

IvanGorshechnikov:
If you DND from Project Explorer to Cloud Storage, you should see confirmation, and then it should copy the script(s)
image

If you DND from Cloud Storage to Project Explorer, you should see an error message that this operation is not supported
image

The behaviour in TE with that fix should be the same as in UE
Supporting of pasting from Cloud Storage to Project Explorer, shold be fixed in https://github.com/dbeaver/pro/issues/4393


ShadelessFox (dismissed):

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/fs/DBNPathBase.java
```diff
@@ -243,7 +243,7 @@ public boolean supportsDrop(DBNNode otherNode) {
 
         if (getOwnerProject() instanceof DBFResourceAdapter rm) {
             // Drop supported only if both nodes are resource with the same handler and DROP feature is supported
-            return otherNode.getAdapter(Path.class) != null
+            return (otherNode.getAdapter(Path.class) != null || otherNode.getAdapter(InputStream.class) != null)
```
serge-rider: otherNode.getAdapter(InputStream.class) open a stream. which must be closed.

You can check for instanceof instead.

serge-rider (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/fs/DBNPathBase.java
ShadelessFox: Good point. It's pretty wild that it can happen in the first place, actually. A better alternative would be to return a supplier that, in turn, will open a new input stream.

IvanGorshechnikov: Fixed, rewrite it to instanceof DBNStreamData and supportStreamData check

labanovichttps (approved):

ShadelessFox (approved):

PR #39119, opened 2025-09-02
dbeaver/pro#6777 Enhance combo box value setting logic with type-specific handling and null checks

labanovichttps:


E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/pages/database/DatabaseConsumerPageMapping.java
```diff
@@ -893,8 +893,19 @@ protected Object doGetValue() {
...
             protected void doSetValue(Object value) {
-                if (combo != null && !combo.isDisposed()) {
-                    combo.setText(CommonUtils.toString(value));
+                if (combo == null || combo.isDisposed()) {
+                    return;
+                }
+                if (value == null) {
+                    combo.setText("");
+                } else {
+                    if (value instanceof DBPNamedObject dbpNamedObject) {
```
ShadelessFox, line 902: This is something a label provider should do. Is the value not constrained to a particular type? Can it really be either a `DBPNamedObject`, an `Enum`, or something else?

labanovichttps: It was taken from the `CustomComboBoxCellEditor`, it used to be before the custom editor

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/pages/database/DatabaseConsumerPageMapping.java
```diff
@@ -893,8 +893,19 @@ protected Object doGetValue() {
...
+                        combo.setText(dbpNamedObject.getName());
+                    } else if (value instanceof Enum anEnum) {
```
serge-rider: Why check for Enum? toString does the same

serge-rider (changes requested):

PR #39142, opened 2025-09-04
dbeaver/pro#6532 api for object settings

yagudin10:


plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMController.java
```diff
@@ -232,4 +232,12 @@ void deleteAllObjectPermissions(
     @NotNull
     String[] getTeamMembers(String teamId) throws DBException;
 
+    @NotNull
+    Map<String, Object> getObjectSettings(
+        @NotNull String objectId,
+        @NotNull SMObjectType objectType,
+        @Nullable String subjectId,
```
alexander-skoblikov: Can we automatically use current user id without additional parameter?

plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMController.java
```diff
@@ -232,4 +232,12 @@ void deleteAllObjectPermissions(
...
+        @Nullable String settingId
```
alexander-skoblikov, line 239: i think will be better return all settings for object

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/app/DBPProjectSettingsProvider.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.app;
+
+import org.jkiss.code.NotNull;
+import org.jkiss.code.Nullable;
+import org.jkiss.dbeaver.DBException;
+
+public interface DBPProjectSettingsProvider {
```
serge-rider: Why it is in dbeaver ce but never used?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMController.java
yagudin10: it will return all settings if settingId is not defined

plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMController.java
```diff
@@ -232,4 +232,23 @@ void deleteAllObjectPermissions(
...
     String[] getTeamMembers(String teamId) throws DBException;
 
+    @NotNull
+    Map<String, Object> getObjectSettings(
+        @NotNull String objectId,
+        @NotNull SMObjectType objectType,
+        @Nullable String settingId
+    ) throws DBException;
+
+    void addObjectSettings(
```
serge-rider: addObjectSettings -> setObjectSettings

serge-rider (changes requested):

PR #39146, opened 2025-09-04
dbeaver/pro#6229 execute sql via cli

alexander-skoblikov:
closes dbeaver/pro#6229


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
```diff
@@ -268,6 +268,16 @@ public static void setStatementFetchSize(DBCStatement dbStat, long firstRow, lon
...
+                if (maxRows > 0) {
+                    dbStat.setLimit(firstRow, maxRows);
+                }
+            } catch (Exception e) {
+                log.warn(e);
+            }
+        }
+        if (maxRows > 0) {
+            try {
+                dbStat.setLimit(firstRow, maxRows);
```
serge-rider: setlimit 2 times?

PR #39156, opened 2025-09-05
dbeaver/pro#6486 Strip SQL comments before checking view CREATE statement

labanovichttps:
image

```
create schema if not exists chp;

drop table if exists chp.accounts;
drop table if exists chp.orders;

create table chp.accounts (
  account_id int,
  status     varchar(20),
  created_at timestamp
);

create table chp.orders (
  order_id   int,
  account_id int,
  amount     decimal(12,2),
  created_at timestamp
);

drop materialized view if exists chp.mv_account_obj;

-- lalallalaa
create materialized view chp.mv_account_obj
auto refresh yes
as --some comment
select
  a.account_id,
  a.status,
  count(o.order_id)      as order_cnt,
  coalesce(sum(o.amount),0) as total_amount
from chp.accounts a
left join chp.orders o using (account_id)
group by a.account_id, a.status;
```


IvanGorshechnikov (approved):

ShadelessFox (approved):

Matvey16:
Verified

PR #39165, opened 2025-09-08
dbeaver/pro#6773 Replace database connection icons with svg

labanovichttps:
image
image
image
image
image
image
image
image
image


E1izabeth (dismissed):

ShadelessFox (dismissed):

ShadelessFox (approved):

Matvey16:
I don't think it's expected that the icon for DynamoDB top node in the AWS cloud is different from DynamoDB connections inside it and in the navigator

image

Matvey16:
Verified

PR #39186, opened 2025-09-11
dbeaver/dbeaver#39168 Respect AI disablement measures

### LINKED ISSUE #39168: DBeaver Community - switch to prevent installing extension
We want to prevent the installation of software / extensions by a standard user.
I set the value "-policy.software.install.disabled = true" in the ini-file, but that does not help
Also the hint on the website, that an installation to a path, where the user has no write-permission, does not help.
I installed the software to "c:\Program Files" on a Windows 11 64-bit client. A standard user has no write-permission to that path. Altough he can install extensions/plugins, they will be installed to "../../Users/xxx/AppData/Roaming/DBeaverData/install-data/"

Please add a switch (registry-key, value in the ini-file, etc) the will prevent the installation of extension/plugins. (For example hide the installation-function in the program)

ShadelessFox:
Test by disabling AI integration using the "Enable AI integration" under preferences, passing the `-Dai.disabled=true` VM argument, or setting the `DBEAVER_AI_DISABLED` environment variable.


E1izabeth (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified
Opened another ticket for incomplete AI disabling https://github.com/dbeaver/pro/issues/6909

PR #39194, opened 2025-09-12
dbeaver/pro#6802 RAG MVP

ivgag:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/OpenAIClient.java
```diff
@@ -168,6 +168,35 @@ public void createChatCompletionStream(
         );
     }
 
+    public List<float[]> embedTexts(
```
serge-rider: Let's move it to OpenAIClientPro. Along with all embeddings related dtop

serge-rider (changes requested):

PR #39197, opened 2025-09-12
dbeaver/dbeaver#36591 Check parent objects to validate matching

### LINKED ISSUE #36591: The table incorrectly selects the wrong ID after the join operation.
### Description

If you perform a join operation, it would look like this:

`SELECT x.id, y.id, y.name FROM tablex x 
LEFT JOIN tabley y ON y.id = x.y_id`

When you want to update the incoming records, the first index after the SELECT statement is placed in the WHERE condition, and the table reference is always the one that comes after FROM, which is tablex. This leads to the wrong record being modified during the update operation.

The correct approach should be that if a change is made in a specific column, for example, if we make a change in x.name, the reference table should be tablex, and the WHERE condition should refer to tablex.

If you perform an update operation, it works like the generated code.

image

if  I change select order  this time si true script;
image

### DBeaver Version

23.1.4

### Operating System

windows

### Database and driver

postgresql 15 

### Steps to reproduce

1. join 2 table 
2. run script with first id , second id 
3. change order  second id ,first id 
4. look at generated or persisting result 

### Additional context

-

IvanGorshechnikov:
image


E1izabeth (dismissed):

labanovichttps (dismissed):

Matvey16:
There is still an incorrect script if changes are made in columns from different tables, but in the same row of the resultset. We need to produce two different update statements, but they are merged into one

https://github.com/user-attachments/assets/9ac36ef2-aab1-4462-9996-cdc43be2d940

labanovichttps (approved):

Matvey16:
Verified

PR #39212, opened 2025-09-16
dbeaver/#38696 Support advanced properties via 'advProp.' prefix

### LINKED ISSUE #38696: PostgreSQL databases not shown by default #20854 - add CLI parameter
> **Is your feature request related to a problem? Please describe.** DBeaver only shows postgres database by default
> 
> **Describe the solution you'd like** Ideally the option to "Show all databases" would be selected by default (under: Connection settings | PostgreSQL).
> 
> **Additional context** This is the top voted issue with DBeaver on stack overflow https://stackoverflow.com/questions/54235029/dbeaver-can-only-see-default-postgresql-database-in-connection

According to @arhayka 's comment;

> Hello, [@robianmcd](https://github.com/robianmcd), This is a known issue. The "Show all databases" option was unchecked on purpose because it was causing too many other issues.

Is this still causing too many other issues? I also checked https://dbeaver.com/docs/dbeaver/Command-Line/#connection-parameters and i do not see arg for this option. At least could we add argument

vkirst:
- Introduce PREFIX_ADVANCED_PROP constant  
- Handle parameters with 'advProp.' prefix as "-con" provider properties
see https://dbeaver.com/docs/dbeaver/Command-Line/#connection-parameters
example: -con "advProp.@dbeaver-show-non-default-db@=true"


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceUtils.java
```diff
@@ -206,6 +207,10 @@ public static DBPDataSourceContainer getDataSourceBySpec(
                         Map<String, String> handlerPopMap = handlerProps.computeIfAbsent(handlerId, k -> new HashMap<>());
                         handlerPopMap.put(paramName, paramValue);
                         handled = true;
+                    } else if (paramName.length() > PREFIX_AVANCED_PROP.length() && paramName.startsWith(PREFIX_AVANCED_PROP)) {
```
ShadelessFox: Redundant check; `startsWith` already checks the length
```suggestion
                    } else if (paramName.startsWith(PREFIX_AVANCED_PROP)) {
```

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceUtils.java
```diff
@@ -206,6 +207,10 @@ public static DBPDataSourceContainer getDataSourceBySpec(
...
+                        paramName = paramName.substring(PREFIX_AVANCED_PROP.length());
+                        DBWorkbench.getPlatform().getPreferenceStore().setValue(paramName, paramValue);
```
ShadelessFox: Well, it seems like it's not an "advanced prop" but rather a preference value. An application's preference, to be precise. Can we rename it?

ShadelessFox (dismissed):
Please update the description of the PR as well

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceUtils.java
```diff
@@ -206,6 +207,12 @@ public static DBPDataSourceContainer getDataSourceBySpec(
                         Map<String, String> handlerPopMap = handlerProps.computeIfAbsent(handlerId, k -> new HashMap<>());
                         handlerPopMap.put(paramName, paramValue);
                         handled = true;
+                    } else if (paramName.startsWith(PREFIX_PREF_PROP)) {
+                        String suffix = paramName.substring(PREFIX_PREF_PROP.length());
+                        if (!suffix.isEmpty()) {
```
labanovichttps, line 213: replace via CommonUtils.isNotEmpty()

ShadelessFox: Uh, why? It also performs a null check, which is clearly redundant here.

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceUtils.java
```diff
@@ -206,6 +207,12 @@ public static DBPDataSourceContainer getDataSourceBySpec(
...
+                            DBWorkbench.getPlatform().getPreferenceStore().setValue(suffix, paramValue);
```
serge-rider: Nope, it should patch datasource preferences, not global

serge-rider (changes requested):

Matvey16:
Doesn't work. You can try it with those parameters. Everything is passed to DBeaver except `advProp.@dbeaver-show-non-default-db@` 

```
 -con "driver=postgres-jdbc|name=test_postgresql|folder="fff"|advProp.@dbeaver-show-non-default-db@=true|database=dvdrental|host=1.2.3.4|connect=false"
```

Also doesn't work for any of the different parameters from the same group like
```
"@dbeaver-show-non-default-db@": "true",
"@dbeaver-chosen-role@": "",
"@dbeaver-show-template-db@": "true",
"@dbeaver-show-unavailable-db@": "true",
"show-database-statistics": "true",
"@dbeaver-read-all-data-types-db@": "true",
"read-keys-with-columns": "false",
"replace-legacy-timezone": "false",
"@dbeaver-use-prepared-statements-db@": "false",
"postgresql.dd.plain.string": "false",
"postgresql.dd.tag.string": "false"
```

Matvey16:
Paramater gets ovewritten by the global settings

Matvey16:
Verified

PR #39215, opened 2025-09-16
dbeaver/pro#4376 `DBPObjectWithOrdinalPosition` extend `Comparable`

ShadelessFox:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBPObjectWithOrdinalPosition.java
```diff
@@ -20,7 +20,11 @@
 /**
  * Object with ordered position.
  */
-public interface DBPObjectWithOrdinalPosition {
+public interface DBPObjectWithOrdinalPosition extends Comparable<DBPObjectWithOrdinalPosition> {
     int getOrdinalPosition();
 
+    @Override
+    default int compareTo(DBPObjectWithOrdinalPosition o) {
```
serge-rider: @NotNull

serge-rider (changes requested):

serge-rider (approved):

PR #39221, opened 2025-09-17
dbeaver/pro#6923 Revert db icon png

labanovichttps:


IvanGorshechnikov:
What about database_icon_big.png?

IvanGorshechnikov (dismissed):

PR #39259, opened 2025-09-23
dbeaver/pro#2765 correct cause in case not found driver

SolovevAnton:
Corrected error message shown in case driver is not found


plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/driver/DriverEditHelpers.java
```diff
@@ -234,11 +236,20 @@ public static void showBadConfigDialog(final Shell shell, final String message,
...
         UIUtils.syncExec(runnable);
     }
 
+    @NotNull
+    private static DBException defineException(@NotNull DBException error) {
+        Throwable cause = error.getCause();
+        return error instanceof DBCConnectException
+            && cause instanceof DBException
+            ? (DBException) cause
+            : error;
```
ShadelessFox:
```suggestion
        Throwable cause = error.getCause();
        return error instanceof DBCConnectException 
            && error.getCause() instanceof DBException cause
            ? cause
            : error;
```

ShadelessFox (approved):

Matvey16:
Verified

PR #39260, opened 2025-09-23
dbeaver/pro#6812 queries logging checkbox set always enabled and remo…

SolovevAnton:
resolves dbeaver/pro#6812. 
Checkbox for "Log SELECT queries in Query Manager" is removed. Now logging is always active by default


ShadelessFox (approved):

uslss:
verified

PR #39261, opened 2025-09-23
dbeaver/pro#6263 Task execution timeout

ShadelessFox:
We decided to leave that option enabled only for **saved tasks**. If we really want to have it available for non-saved tasks, we need a separate ticket.

java_QtR3uVnaaJ

Localization files are borked, so don't mind them (or enable whitespace ignore)


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/RuntimeUtils.java
```diff
@@ -214,6 +216,32 @@ public static String formatExecutionTime(@NotNull Duration duration) {
         }
     }
 
+    @NotNull
+    public static String formatDuration(@NotNull Duration duration) {
```
serge-rider, line 220: Use RuntimeUtils.formatExecutionTime?

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/task/TaskRunJob.java
```diff
@@ -201,17 +203,25 @@ public void subTaskFinished(@Nullable DBTTask task, @Nullable Throwable error, @
...
-        if (task.getMaxExecutionTime() > 0
-            && taskStartTime > 0
-            && (System.currentTimeMillis() - taskStartTime) > (task.getMaxExecutionTime() * 1000L)) {
-            canceledByTimeOut = true;
-            cancel();
-            activeMonitor.getNestedMonitor().setCanceled(true);
-            if (isRunDirectly()) {
-                canceling();
-            }
+    public void checkCanceledByTimeout() {
```
labanovichttps: I think you need to rename the method or rewrite javadoc, is this method cancel or check and cancel

kseniaguzeeva:
the icon is active and opens the new dialog for data compare/schema compare/tools dialogs

dbeaver_0PYze8nimU

kseniaguzeeva:
- DBeaver UE/TE can't be built https://build.dev.dbeaver.infra/job/DBeaverDesktop/job/dbeaver-ee/7314/console
- DBeaver CE 25.2.3.202510060647: timeout button is active for tools dialog

dbeaver_F3M4ip05jj

dbeaver_YoldguogME

kseniaguzeeva:
1. Timeout doesn't work for now. STR: 
1.1 create a task and set a timeout
1.2 reopen the task and check the timeout
AR: the timeout isn't set
Note: tasks with set timeout from previous DBeaver version is executed with the timeout one time only in this PR, then it also disappears.

[dbeaver-debug.log](https://github.com/user-attachments/files/22740134/dbeaver-debug.log)

[.log](https://github.com/user-attachments/files/22740133/default.log)


https://github.com/user-attachments/assets/f3ba0e0f-8fa4-45c8-b6d7-9209936e2cc3


Note: new behavior: timeout button for not saved task is disabled for export/import and enabled for compare/dump


https://github.com/user-attachments/assets/9d2b1794-9103-451a-95d0-6857cfa5d4f8

ShadelessFox:
The latter is expected due to the implementation details; data transfer doesn't use tasks under the hood; compare/dump do use a temporary task that later gets saved.

kseniaguzeeva:
Checked.
Remarks: 
1. dump and schema compare dialogs opened from Navigator have active the timeout button, but the timeout is applied after saving the task (when it's just executed, the timeout isn't applied) - expected as discussed above

https://github.com/user-attachments/assets/18fe0102-42d7-454d-9db1-05335d700c9c

2. when a value is deleted from the timeout panel it's replaced by 1 

https://github.com/user-attachments/assets/6a99d24e-f85f-497e-bc67-8ef98ba697e3

3. `Ignore task error` setting in the composite task isn't applied to canceled due to a timeout task.

4. If a subtask in a composite task is canceled due to a timeout, tasks after it aren't executed.

kseniaguzeeva:
TE: 
1. Timeout doesn't work for tasks (not scheduled and scheduled)
[dbeaver-debug.log](https://github.com/user-attachments/files/23000023/dbeaver-debug.log)
[tm server.txt](https://github.com/user-attachments/files/23000028/tm.server.txt)

https://github.com/user-attachments/assets/e92f0159-37fc-4057-860b-4aed74369052

2. Is it possible to make timeout to work with the old version of the server (client 25.3 and server 25.2)?

Note about CE: timeout has been removed completely

kseniaguzeeva:
✅Removed from CE
✅Fixed in TE (works also with server 25.2.0)

PR #39271, opened 2025-09-24
dbeaver/pro#6932 dbeaver/pro#6905 Exclude JDT quick fix processors

ShadelessFox:
JDT quick fixes only work with the Java code model and use different interfaces, so it's pretty much impossible to use them there.


E1izabeth (approved):

uslss:
verified
image

PR #39274, opened 2025-09-25
dbeaver/pro#6931 Redshift connect fix

serge-rider:


uslss:
Redshift Serverless
image
Redshift
image

PR #39275, opened 2025-09-25
dbeaver/pro#6475 Enabled bigquery sessions for datagrid tx support

Nexus6v2:
For Bigquery:

* Fixes the error when trying to persist data grid changes in CB **with/without autocommit**:
image

* Fixes the error when trying to persist data grid changes in desktop **with manual commit**:
image

Doc: https://docs.simba.insightsoftware.com/hc/en-us/articles/30350148329101-EnableSession

Closes https://github.com/dbeaver/pro/issues/6475


plugins/org.jkiss.dbeaver.ext.bigquery/src/org/jkiss/dbeaver/ext/bigquery/model/BigQueryDataSource.java
```diff
@@ -53,7 +52,7 @@ protected String getConnectionURL(DBPConnectionConfiguration connectionInfo) {
             DBPDriver driver = getContainer().getDriver();
             return driver.getDataSourceProvider().getConnectionURL(driver, connectionInfo);
         }
-        return connectionURL;
+        return connectionURL + ";EnableSession=1;";
```
serge-rider, line 55: It cannot be anbled by default. We have this feature in desktop, add parameter in config

serge-rider (changes requested):

PR #39277, opened 2025-09-25
Dbeaver/this PR should be deleted 

SolovevAnton:
SHOULD BE DELETED. New PR: https://github.com/dbeaver/dbeaver/pull/39296 
added additional check, to make sure active DB and selected DB is the same instance


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -605,6 +600,19 @@ protected Connection openConnection(@NotNull DBRProgressMonitor monitor, @Nullab
...
+    @Nullable
+    private String getDbName(@Nullable JDBCRemoteInstance instance, @NotNull DBPConnectionConfiguration conConfig) {
+        return instance != null
+            ? instance.getName()
+            : PostgreUtils.getDatabaseNameFromConfiguration(conConfig);
+    }
+
+    private boolean shouldConnectWithNewConfig(@NotNull DBPConnectionConfiguration conConfig, @NotNull String dbName) {
+        return isReadDatabaseList(conConfig)
+            || !CommonUtils.isEmpty(conConfig.getBootstrap().getDefaultCatalogName())
```
ShadelessFox, line 612: What about the default schema?

PR #39282, opened 2025-09-25
Dbeaver/dbeaver#39228 fix model config uppercase modification

### LINKED ISSUE #39228: For model configuration items, no automatic modification in terms of case or prefixes should be performed.
### Description

> For model configuration items, no automatic modification in terms of case or prefixes should be performed.

Regarding the model settings, input transmission, they are all converted to lowercase, resulting in the error: `"type":"not_found_error","param":"invalid_model_name","code":"400"`. When testing manually through a curl request, the results are normal as long as the original case is maintained.

**Desired Behavior:**

The system should forward the "model" field provided by the user exactly as is, without any automatic modification of case or prefix.

### DBeaver Version

25.2.0.202509081303

### Operating System

macOS 15.5

### Database and driver

Independent of database, AI configuration

### Steps to reproduce

_No response_

### Additional context

_No response_

SolovevAnton:
fixes #39228 
changed getEffectiveModelName. Now returns model in original case, if its unknown


ShadelessFox (approved):

Matvey16:
Verified

PR #39298, opened 2025-09-29
dbeaver/pro#6816 Fix SQL Script task error reporting

ShadelessFox:
Only reproducible against the SQL Script task and when scheduled or run via CLI's `-runTask`


Matvey16:
Verified

PR #39299, opened 2025-09-29
dbeaver/pro#4569 change md5 hashing

HocKu7:
Closes https://github.com/dbeaver/pro/issues/4569


plugins/org.jkiss.dbeaver.model.sm/META-INF/MANIFEST.MF
```diff
@@ -14,6 +14,7 @@ Export-Package: org.jkiss.dbeaver.model.security,
  org.jkiss.dbeaver.model.secret
 Require-Bundle: org.eclipse.equinox.security,
  org.eclipse.core.runtime,
- org.jkiss.dbeaver.model;visibility:=reexport
+ org.jkiss.dbeaver.model;visibility:=reexport,
+ bcprov
```
serge-rider, line 18: Use import-package instead of bundle for BC.

plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/auth/Argon2IdHasher.java
```diff
@@ -0,0 +1,130 @@
...
+
+        Argon2BytesGenerator gen = new Argon2BytesGenerator();
+        gen.init(params);
+
+        byte[] out = new byte[outLen];
+        gen.generateBytes(pwd, out, 0, out.length);
+        return out;
+    }
+
+    private static Phc parsePhc(String phc) {
```
serge-rider, line 88: Use nullable annotations

serge-rider (changes requested):

PR #39320, opened 2025-10-02
Dbeaver/pro#6876 download drivers redesign

SolovevAnton:
closes dbeaver/pro#6876
to consider:
1. Dialog always opens with hidden options
2. Resizing is not remembered
3. Width is measured depending on the length of driver info and name

Recording 2025-10-06 102621

image


Matvey16:
Previously the window appeared centerd on the screen and not very wide (regardless of the size of the driver description)
dbeaver_MLruPYDO4A

Now if the description is too big the window will go over the screen borders


https://github.com/user-attachments/assets/ca3afd1f-c61b-46f8-b325-77d4ebd255a7

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/internal/UIConnectionMessages_ru.properties
```diff
@@ -222,6 +222,7 @@ dialog_driver_download_auto_page_download_specific_driver_files = Скачива
 dialog_driver_download_auto_page_driver_file_missing_text = Файлы драйвера {0} отсутствуют.\n\n
 dialog_driver_download_auto_page_force_download = Скачать принудительно / перезаписать
 dialog_driver_download_auto_page_force_download_tooltip = Скачать файлы принудительно. Скачивание будет произведено даже если файлы уже есть на диске
+dialog_driver_download_auto_page_show_details=Показать датели
```
IvanGorshechnikov:
```suggestion
dialog_driver_download_auto_page_show_details=Показать детали
```

Matvey16:
Verified

PR #39331, opened 2025-10-06
dbeaver/pro#7042: fix wording in AI popup

SolovevAnton:
closes dbeaver/pro#7042


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/HelpUtils.java
```diff
@@ -28,8 +28,14 @@ public class HelpUtils {
 
     public static final String GLOBAL_HELP_PREFIX = "https://dbeaver.com/docs/dbeaver/";
 
+    public static final String GITHUB_HELP_PREFIX = "https://github.com/dbeaver/dbeaver/wiki/";
```
serge-rider, line 31: Keep dbeaver.com as main source, it is much bigger

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/HelpUtils.java
SolovevAnton: so we leave link as it was: https://dbeaver.com/docs/dbeaver/AI-Smart-Assistance/ for CE as well. Correct?

SolovevAnton: decided to keep two basic links for different doc needs

Matvey16:
Verified

PR #39336, opened 2025-10-06
dbeaver/dbeaver#38274 Show action toolbar on hover

### LINKED ISSUE #38274: Where is the button for object comment
### Description

Not long ago, DBeaver introduced a button to open the object comment in a pop-up window.
It wasn't perfect; it opens a small window, forcing a user to expand it every time. But at least it gives the option to expand it vertically and horizontally as desired.

Since some of the recent versions (noticed in 25.0.5, still existing in 25.1.0)  this button is no longer available.
Now, the comment view is narrow with no option to expand it horizontally. Only vertical scaling is possible, being extremely clunky at the same time. See video.

https://youtu.be/oOMRM1gKiXM

I'm disappointed, especially because I was the one who asked for a more comfortable view of comments.  To me, it seems that some decisions are made by people who never work with the features they program. 

### DBeaver Version

CE 25.1.0

### Operating System

Fedora

### Database and driver

Postgresql

### Steps to reproduce

See video

### Additional context

_No response_

ShadelessFox:
https://github.com/user-attachments/assets/2647ed39-09c1-47c8-b0a6-9bac58502ddf


Matvey16:
Verified

PR #39347, opened 2025-10-07
dbeaver/pro#6509 CB Query confirmation

Nexus6v2:
Closes dbeaver/pro#6509


yagudin10 (dismissed):

plugins/org.jkiss.dbeaver.model/plugin.xml
```diff
@@ -28,12 +28,51 @@
...
     <extension-point id="org.jkiss.dbeaver.settings" name="DBeaver Settings" schema="schema/org.jkiss.dbeaver.settings.exsd"/>
+    <extension-point id="org.jkiss.dbeaver.confirmations" name="UI confirmations" schema="schema/org.jkiss.dbeaver.confirmations.exsd"/>
 
     <extension point="org.jkiss.dbeaver.expressions">
         <namespace id="math" class="java.lang.Math" description="Math functions"/>
         <namespace id="content" class="org.jkiss.dbeaver.registry.expressions.ContentExpressionFunctions" description="Content handling functions"/>
     </extension>
 
+    <extension point="org.jkiss.dbeaver.confirmations">
+        <confirmation
```
serge-rider: I believe these particular confirmations should be in `model.sql`

Descriptors may remain in model but I believe org.jkiss.dbeaver.registry is a better place.

serge-rider (dismissed):

plugins/org.jkiss.dbeaver.model/plugin.xml
Nexus6v2: Done, moved messages to `model.sql`
I left the descriptor in `model` because it's also used in bundles not dependent on `model.sql`.

yagudin10 (dismissed):

yagudin10 (approved):

PR #39349, opened 2025-10-07
dbeaver#38815 Add global preference page for project network profiles

### LINKED ISSUE #38815: Make SSH profiles available in general preferences
Currently the SSH profiles are only configurable when being in "add new connection" or "edit connection" windows.

However, I sometimes need to adjust a SSH profile without wanting to add/edit a connection. For example, if a customer leaves then I just want to delete a profile. Or sometimes I have SSH access some time earlier than DB credentials, so I'd like to prepare the SSH profile and later just use it for connections. Or sometimes I just want to get an overview of all my SSH profiles.

For these tasks it would be very helpful to not having to go through "creating a dummy connection" process or "editing some _unrelated_ connection" just to be able to list, edit or extend the SSH profiles.

As a user I totally expect the SSH profile managment to be available in the general "Preferences -> SSH" options.

ShadelessFox:
New preference page:

image

Context menu tweaks:

image


plugins/org.jkiss.dbeaver.core/plugin.xml
```diff
@@ -888,14 +901,14 @@
             <separator name="additions" visible="true"/>
             <command commandId="org.eclipse.ui.navigate.collapseAll"/>
             <command commandId="org.jkiss.dbeaver.core.navigator.linkeditor"/>
-            <command commandId="org.jkiss.dbeaver.core.navigator.view.configure"/>
+            <command commandId="org.eclipse.ui.file.properties"/>
```
serge-rider, line 904: Why? 
org.jkiss.dbeaver.core.navigator.view.configure seems to be smarter than just file properties

plugins/org.jkiss.dbeaver.ui.app.standalone/plugin.xml
```diff
@@ -106,6 +106,7 @@
         <image commandId="org.eclipse.ui.file.revert" icon="icons/eclipse/revert.png" disabledIcon="icons/eclipse/revert_d.png"/>
         <image commandId="org.eclipse.ui.file.import" icon="platform:/plugin/org.jkiss.dbeaver.ui/icons/file/import.svg"/>
         <image commandId="org.eclipse.ui.file.export" icon="platform:/plugin/org.jkiss.dbeaver.ui/icons/file/export.svg"/>
+        <image commandId="org.eclipse.ui.file.properties" icon="platform:/plugin/org.jkiss.dbeaver.ui/icons/configuration.svg"/>
```
serge-rider, line 109: Disagree. Properties is not the same as configure.
And we alredy have properties.svg

kseniaguzeeva:
TE:
- Long profile list is loaded twice in Preferences

https://github.com/user-attachments/assets/1d4dabe7-9a2d-47ac-8e87-21e0f8fcfcf9

- access for different roles - the same as in devel. Other ticket https://github.com/dbeaver/pro/issues/7127

MashaKorax:
verified

Issue #6813, opened 2019-09-24
Query still running after data fetch


geuss:
#### System information: 
- Operating system (distribution) and version: Windows 7
- DBeaver version: CE 6.2.1
- Java version : Default DBEAVER Java version
- Additional extensions: none

#### Connection specification:
- Database name and version: Teradata 15.00
- Driver name: Teradata JDBC Type 4 driver
- Do you use tunnels or proxies (SSH, SOCKS, etc)?  No
[dbeaver-debug.log](https://github.com/dbeaver/dbeaver/files/3646158/dbeaver-debug.log)

#### Include any warning/errors/backtraces from the logs

Hi!
From relase 6.1.5, sometimes, query still result in execution even if data was fetched:

![image](https://user-images.githubusercontent.com/26387091/65495359-ef32aa80-deb6-11e9-83cf-273437392c4c.png)

If I wait for a lot of time, the connection was closed:

![image](https://user-images.githubusercontent.com/26387091/65495523-2b660b00-deb7-11e9-8ac8-20e78994e489.png)

I use DBEAVER CE 6.2.1 with Teradata DBMS v.15.00.

geuss:
The issue is still present with version 6.2.2
![image](https://github.com/dbeaver/dbeaver/assets/26387091/fdf57925-5253-465b-ac97-2630808eae54)

geuss:
Hi! The issue is still present in version 6.2.4:

![image](https://github.com/dbeaver/dbeaver/assets/26387091/d63c0b84-ca27-4323-9c0f-347615563572)

I launched the query: "select * from mytable;"

Data was fetched and displayed in Result Viewer (all 200 records!), but query results always in execution (see "Background Task"). I can't execute others queries...:

![image](https://user-images.githubusercontent.com/26387091/68226469-6731db80-fff2-11e9-95ab-dc8944a6a6d0.png)

If I try to break execution from Background Task, after a lot of time... the connection was automatically closed.

It happens (almost always) with slow queries. The last version of Dbeaver where this bug was not present (or the problem is very infrequent) is the 6.1.5.

Thanks!

geuss:
The issue is still present with release 6.3.2. 

In this video you can see the problem with version > 6.1.5

Here, with the version 6.1.5 (that works...):


Thanks!!

geuss:
Sorry... I closed the issue by mistake. :-)

geuss:
Hi Serge, the error is still present with the 6.3.5 release.

[...]

PR #39353, opened 2025-10-08
dbeaver/dbeaver#6813 Prevent duplicate object name comment when generating view DDL

### LINKED ISSUE #35329: SQL Server: dbeaver adds a new comment line "--<object name> source" to the top of view (proc, function) definition and saves it every time
### Description

When the definition of a view, function or stored procedure is edited via Properties -> Definition, then a comment line `--<object name> source` is added to the top of the definition and persisted in the database.

With each new edit and saving, another same line is added.

Is there way to disable this behavior?

Thank you.

See the result of this below on an example of a view:
image

### DBeaver Version

Community Edition 24.1.5.202408181605

### Operating System

Windows 10

### Database and driver

SQL Server

### Steps to reproduce

1. Use SQL Server
2. Go to View - Properties - Definition
3. Make changes in the code
4. Save the view
5. Refresh the Definition with the button in the lower right corner
6. You will see that another line `-- <object name> source` appears at the top of the code
7. You can also check sys.sql_modules and see that all these texts are persisted in the database

### Additional context

_No response_

vkirst:


MashaKorax:
verified

PR #39354, opened 2025-10-08
dbeaver/dbeaver#39341 ResultSet fetch size set fix (revert)

### LINKED ISSUE #39341: ResultSet Fetch Size Defaulting To 1
### Description

I have always had "ResultSet fetch size" set to zero, to prevent result limits from being applied, but today, after upgrading to DBeaver Community Edition v25.2.2.202510052342, the "ResultSet fetch size" was changed to 1, which caused lost time, as I was getting unexpected results with no explanation, until I found the "ResultSet fetch size" was changed.

I can work around this by increasing the ResultSet fetch size in settings from zero to some higher number, but since I commonly return large result sets, I will always have to question whether I got everything or not.

Please correct this feature to once again allow zero to mean return all records, with no limit.

Thanks!

### DBeaver Version

Community Edition 25.2.2.202510052342

### Operating System

Windows 11 Pro

### Database and driver

PostrgreSQL 16.9
PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

1. Ensure the ResultSet fetch size is set to zero.
2. Open the database connection.
3. Run any query that should return more than one row.
4. Only 1 row is returned, and the Result Set size shown at the bottom of the Query Results window will be defaulted to 1 (I assume it is now seeing the zero from preferences zero as invalid, and defaulting the value to 1).

### Additional context

_No response_

serge-rider:


Matvey16:
Verified

PR #39362, opened 2025-10-09
Better support for EXTRA_ARGS

dqj1998:


ShadelessFox:
Thanks for the contribution. Sorry it took us so long to review your PR.

If I understand the problem correctly, it narrows down to the following line in the `dbeaver.ini` file:
```
-vm ../Eclipse/jre/Contents/Home/lib/libjli.dylib
```

It really should be put there during the packaging stage, not via the product file as we do it now:
https://github.com/dbeaver/dbeaver/blob/81c0067a62b7d0cd4f334111dfb6f54223da4f8b/product/community/DBeaver.product#L13

I see that your script also provides useful options for configuring JRE, as well as removing quarantine, but it won't be required if that line is removed from `dbeaver.ini`.

We'll address this issue soon.

PR #39374, opened 2025-10-10
dbeaver/pro#6614 ai chat warnings

IvanGorshechnikov:
Screenshots are in related ticket


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AIEngineRequest.java
```diff
@@ -30,17 +30,21 @@ public final class AIEngineRequest {
     @NotNull
     private final List<AIMessage> messages;
     private final List<AIFunctionDescriptor> functions = new ArrayList<>();
+    private final boolean wasPromptTruncated;
 
     public AIEngineRequest(
-        @NotNull List<AIMessage> messages
+        @NotNull List<AIMessage> messages,
+        boolean wasPromptTruncated
```
serge-rider: Let's just add setter for PromptTruncated property.
Adding all props in constructor is redundant

serge-rider (changes requested):

PR #39388, opened 2025-10-13
dbeaver/pro#7059 Data Compare improvements

ShadelessFox:


plugins/org.jkiss.dbeaver.tasks.native.ui/src/org/jkiss/dbeaver/tasks/ui/nativetool/AbstractNativeToolWizard.java
```diff
@@ -279,7 +279,9 @@ protected void onError() {
 
     @NotNull
     @Override
-    public TaskConfigurationWizardDialog createWizardDialog(@NotNull IWorkbenchWindow window, @Nullable IStructuredSelection selection) {
+    public TaskConfigurationWizardDialog createWizardDialog(@NotNull IWorkbenchWindow window, @NotNull IStructuredSelection selection,
```
serge-rider: One param per line

PR #39399, opened 2025-10-14
dbeaver/pro#5395 Fix semantic analysis not recognizing temp tables

ShadelessFox:
- [x] Please verify that comment recognition (`-- comment`, `# comment`) behaves as before
- [x] Please verify that semantic analysis takes roughly the same time to complete against the same queries as before

~~I enabled fallback to the structure assistant if an object couldn't be found. Theoretically, it should only affect the worst-case scenario when we couldn't find a cached object. In other cases, it would improve database-specific object recognition.~~ obsolete

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
@@ -278,7 +278,7 @@ public class SQLServerSchema implements DBSSchema, DBPSaveableObject, DBPQualifi
             return object;
         }
         // tempdb tables have special naming convention. See SQLServerUtils.stripTempdbTableName
-        if (database.getName().equalsIgnoreCase(SQLServerConstants.TEMPDB_DATABASE)) {
+        if (database.isTempDatabase()) {
             for (SQLServerTableTemp table : tableCache.getTypedObjects(monitor, this, SQLServerTableTemp.class)) {
                 if (table.getOriginalName().equalsIgnoreCase(childName)) {
                     return table;
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
@@ -272,7 +272,20 @@ public class SQLServerSchema implements DBSSchema, DBPSaveableObject, DBPQualifi
 
     @Override
     public SQLServerTableBase getChild(@NotNull DBRProgressMonitor monitor, @NotNull String childName) throws DBException {
-        return tableCache.getObject(monitor, this, childName);
+        SQLServerTableBase object = tableCache.getObject(monitor, this, childName);
+        if (object != null) {
+            return object;
+        }
+        // tempdb tables have special naming convention. See SQLServerUtils.stripTempdbTableName
+        if (database.getName().equalsIgnoreCase(SQLServerConstants.TEMPDB_DATABASE)) {
+            for (SQLServerTableBase table : tableCache.getAllObjects(monitor, this)) {
+                String name = SQLServerUtils.stripTempdbTableName(table.getName());
+                if (name != null && name.equalsIgnoreCase(childName)) {
+                    return table;
+                }
+            }
+        }
+        return null;
     }
 
     @NotNull
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
@@ -120,6 +120,7 @@ public class SQLServerSchema implements DBSSchema, DBPSaveableObject, DBPQualifi
         return database.getDataSource();
     }
 
+    @NotNull
     @Property(viewable = false, editable = true, order = 10)
     public SQLServerDatabase getDatabase() {
         return database;
@@ -278,9 +279,8 @@ public class SQLServerSchema implements DBSSchema, DBPSaveableObject, DBPQualifi
         }
         // tempdb tables have special naming convention. See SQLServerUtils.stripTempdbTableName
         if (database.getName().equalsIgnoreCase(SQLServerConstants.TEMPDB_DATABASE)) {
-            for (SQLServerTableBase table : tableCache.getAllObjects(monitor, this)) {
-                String name = SQLServerUtils.stripTempdbTableName(table.getName());
-                if (name != null && name.equalsIgnoreCase(childName)) {
+            for (SQLServerTableTemp table : tableCache.getTypedObjects(monitor, this, SQLServerTableTemp.class)) {
+                if (table.getOriginalName().equalsIgnoreCase(childName)) {
                     return table;
                 }
             }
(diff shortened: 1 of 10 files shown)


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/context/SQLQueryConnectionRealContext.java
```diff
@@ -82,7 +84,7 @@ protected List<? extends DBSObject> findRealObjectsImpl(@NotNull DBRProgressMoni
                 container,
                 this.executionContext,
                 objectName,
-                false,
+                true,
```
serge-rider: Using assistant is expensive (it executes quite complicated requests to inf schema).
Are we sure we need it in this case?

ShadelessFox: Yes. Unfortunately, it's required at least for that particular case. Temporary tables are weird in SQL Server; they're stored in a separate database called `tempdb` under mangled names. So yes, we must rely on the structure assistant there.

ShadelessFox: See https://github.com/dbeaver/dbeaver/blob/6aba25be486b532fe7f937a46811dede5d006908/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerStructureAssistant.java#L337

ged-yuko: @ShadelessFox if temporary tables are like globally visible, then it should be possible to supply (and cache) them for semantic analysis without assistant by implementing DBSVisibilityScopeProvider for SQLServer-specific schema object or datasource object and creating a special DBSObjectContainer logically containing these temporary table objects (if there is no existing schema object responsible for this, which can be used the same way how public schema handled for Oracle).

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
```diff
@@ -272,7 +272,20 @@ public List<SQLServerObject> getChildren(@NotNull DBRProgressMonitor monitor) th
 
     @Override
     public SQLServerTableBase getChild(@NotNull DBRProgressMonitor monitor, @NotNull String childName) throws DBException {
-        return tableCache.getObject(monitor, this, childName);
+        SQLServerTableBase object = tableCache.getObject(monitor, this, childName);
+        if (object != null) {
+            return object;
+        }
+        // tempdb tables have special naming convention. See SQLServerUtils.stripTempdbTableName
+        if (database.getName().equalsIgnoreCase(SQLServerConstants.TEMPDB_DATABASE)) {
```
serge-rider: Let's turn this expression into isTempDatabase in SQLServerDatabase. And calculate it in constructor.

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
```diff
@@ -272,7 +272,20 @@ public List<SQLServerObject> getChildren(@NotNull DBRProgressMonitor monitor) th
...
+            for (SQLServerTableBase table : tableCache.getAllObjects(monitor, this)) {
+                String name = SQLServerUtils.stripTempdbTableName(table.getName());
```
serge-rider: This is a super-slow way to find a table. Again we can cache stripped table names in SQLServerTable.
Or even create class `SQLServerTableTemp extends SQLServerTableBase`.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerSchema.java
```diff
@@ -272,7 +273,19 @@ public List<SQLServerObject> getChildren(@NotNull DBRProgressMonitor monitor) th
 
     @Override
     public SQLServerTableBase getChild(@NotNull DBRProgressMonitor monitor, @NotNull String childName) throws DBException {
-        return tableCache.getObject(monitor, this, childName);
+        SQLServerTableBase object = tableCache.getObject(monitor, this, childName);
+        if (object != null) {
+            return object;
+        }
+        // tempdb tables have special naming convention. See SQLServerUtils.stripTempdbTableName
+        if (database.isTempDatabase()) {
```
ged-yuko, line 281: Consider covering this case in getChildren(..) method as well. Assuming API consistency is about getChild(..) and getChildren(..) both operating on the same "logical collection". [The same way how `synonymsAsChildren` flag handled in `OracleSchema` for example](https://github.com/dbeaver/dbeaver/blob/0a9b46e1a6fac7324bc3fe49b5f62e1bbb640047/plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleSchema.java#L385C1-L419C6).

ShadelessFox: No, it's not a good idea. It will result in duplicates scattered around the navigator. The whole purpose of this piece of code is to find a table by its "unmangled" name, nothing more. We still want to show them under their original names in the navigator.

ged-yuko: Souldn't it be a different collection object then, than the observable in navigator in this case? And mightn't it lead to a strange behavior if the same table object will be obtainable by name X while it itself reporting having a name Y? Looks like API consistency violation to me personally. Either way, seems you know better if such an API behavior is clearly understandable according to your API contracts or not.

ged-yuko:
By the way. Shouldn't it _not_ prevent `#comment` from working in dialects where `#identifier` is not a valid table name?
Assuming `#identifier` may be valid only when `#` is _not_ used for comments in a certain dialect.
Should be easily distinguishable [here](https://github.com/dbeaver/dbeaver/blob/bce4443146fba2fea81f9ea250b22bd7b51c0e79/plugins/org.jkiss.dbeaver.model.lsm/src/org/jkiss/dbeaver/model/lsm/LSMAnalyzerParameters.java#L36) based on the SQLDialect properties.
Like allow hash-prefixed identifiers when `dialect.validIdentifierStart('#')` or when `dialect.getSingleLineComments()` does not contain `"#"`.
And then something along these lines:

```
    // ....

    private boolean allowHashPrefixedIdentifiers;

    public SQLStandardLexer(CharStream input, LSMAnalyzerParameters parameters) {
        this(input);
        this.allowHashPrefixedIdentifiers = parameters.allowHashPrefixedIdentifiers();
        // ....
    }

// ....

fragment Hashtag: '#'; // "fragment" to enforce it being treated only as part of other lexer rules and exclude possibility of its standalone classification, which influences performance

// ....

HashLineComment: { !allowHashPrefixedIdentifiers }? Hashtag ~ [\r\n]*;
DashLineComment: '--' ~ [\r\n]*;
LineComment : HashLineComment|DashLineComment;

// ....

fragment IdentifierBody:  IdentifierStart IdentifierPart*;
fragment IdentifierStart: SimpleLatinLetter|Underscore|({ allowHashPrefixedIdentifiers }? Hashtag);
fragment IdentifierPart:  SimpleLatinLetter|Underscore|Digit;

```

uslss:
verified

ShadelessFox:
@ged-yuko

> Shouldn't it not prevent #comment from working in dialects where #identifier is not a valid table name?

This is a good question. Ideally, it should not. It seems like it doesn't affect anything in its current form. It can be addressed in a separate issue.

I genuinely respect your desire to help. To be honest, this is the first time I have received a code review from a person who is not affiliated with DBeaver. Thank you :^)

PR #39411, opened 2025-10-15
Dbeaver/dbeaver#38490 advanced copy settings saved

### LINKED ISSUE #38490: Change default copy behavior
It's frustrating that it is so many steps to copy data with headers from the results grid.

I would like to be able to set somewhere the default copy behaviour in the results grid. Now Ctrl+C only copies the values, not including the headers.
Now I know about the Advanced Copy dialog (Ctrl+Shift+C) and accessing those settings later with Ctrl+Alt+Shift+C, but that is an annoying shortcut and the Advanced Copy dialog has to be configured every time after starting DBeaver again.

It would be great if you can add a setting that changes the Ctrl+C behavior to what the Advanced Copy dialog is set to by default: effectively Copy with headers.

This setting **must** persist after closing the application.
Preferably you could check/uncheck this option not only somewhere in the preferences, but also in the right click menu of the results grid.

This would be so very helpful for so many people in many usecases.

SolovevAnton:
closes Dbeaver/dbeaver#38490
Now when using copy.special.with.last.settings shortcut advanced settings dialog should not appear after restart.
changed mac default shortcuts of Advanced copy with prev settings to "CTRL+SHIFT+4"

### DIFF
--- a/plugins/org.jkiss.dbeaver.core/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.core/plugin.xml
@@ -924,9 +924,7 @@
         <key commandId="org.jkiss.dbeaver.ui.tools.menu" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" platform="cocoa" sequence="CTRL+V"/>
 
         <key commandId="org.jkiss.dbeaver.core.edit.copy.special" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" sequence="CTRL+SHIFT+C"/>
-        <key commandId="org.jkiss.dbeaver.core.edit.copy.special.with.last.settings" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" platform="win32"  sequence="CTRL+SHIFT+ALT+C"/>
-        <key commandId="org.jkiss.dbeaver.core.edit.copy.special.with.last.settings" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" platform="gtk" sequence="CTRL+SHIFT+ALT+C"/>
-        <key commandId="org.jkiss.dbeaver.core.edit.copy.special.with.last.settings" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" platform="cocoa" sequence="CTRL+SHIFT+4"/>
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.core/plugin.xml
```diff
@@ -923,8 +923,10 @@
...
         <key commandId="org.jkiss.dbeaver.ui.tools.menu" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" platform="cocoa" sequence="CTRL+V"/>
 
-        <key commandId="org.jkiss.dbeaver.core.edit.copy.special" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" sequence="CTRL+SHIFT+C"/>
-        <key commandId="org.jkiss.dbeaver.core.edit.copy.special.with.last.settings" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" sequence="CTRL+SHIFT+ALT+C"/>
+        <key commandId="org.jkiss.dbeaver.core.edit.copy.special" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" platform="win32" sequence="CTRL+SHIFT+C"/>
+        <key commandId="org.jkiss.dbeaver.core.edit.copy.special" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" platform="mac" sequence="CTRL+SHIFT+C"/>
+        <key commandId="org.jkiss.dbeaver.core.edit.copy.special.with.last.settings" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" platform="win32"  sequence="CTRL+SHIFT+ALT+C"/>
+        <key commandId="org.jkiss.dbeaver.core.edit.copy.special.with.last.settings" schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" platform="mac" sequence="CTRL+SHIFT+ALT+C"/>
```
labanovichttps: Are you sure that it'll work on a Mac? There's no key like ALT on Mac

SolovevAnton: ty, changed to CTRL+SHIFT+4

Matvey16:
Verified

PR #39412, opened 2025-10-15
dbeaver/dbeaver#11453 Add option to set search_path when switching schemas in the PostgreSQL connection settings

### LINKED ISSUE #11453: [postgresql] How to not change search_path when selecting database
#### System information: 
- Win/Linux
- 7.3.5

- Postgresql
- org.postgresql.Driver
- No

obrazek

How can I achieve that?

labanovichttps:
image


IvanGorshechnikov (dismissed):

ShadelessFox (changes requested):
This breaks old behavior, as the new option is **turned off** by default.

labanovichttps:
> This breaks old behavior, as the new option is **turned off** by default.

Yes, I did it on purpose, cuz we shouldn't change it at all

ShadelessFox:
> Yes, I did it on purpose, cuz we shouldn't change it at all

Was this change discussed? Just want to confirm it's indeed intentional.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
```diff
@@ -143,8 +143,16 @@ boolean changeDefaultSchema(DBRProgressMonitor monitor, PostgreSchema schema, bo
             return false;
         }
 
-        setSearchPath(monitor, schema);
-        setSearchPath(schema.getName());
+        boolean setSearchPath = CommonUtils.getBoolean(
```
serge-rider: It doesn't make sense. Moreover we already have such an option: 
image


The idea is to stop changing search_path when it is not needed. We definitely don't need yet another pointless config option.

WE should change search_path only if active schema is not in the path.

serge-rider (changes requested):

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
```diff
@@ -304,13 +297,16 @@ private void setUserInTheEndOfThePath(List<String> searchPath) {
     }
 
     private void setSearchPath(String path) {
-        searchPath.clear();
-        searchPath.add(path);
-        if (!path.equals(activeUser)) {
+        if (searchPath.contains(path)) {
```
serge-rider: I don't understand this logic

labanovichttps: just don't add a path to an existing one if we have it; also, do not clear it.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
```diff
@@ -143,8 +143,16 @@ boolean changeDefaultSchema(DBRProgressMonitor monitor, PostgreSchema schema, bo
...
-        setSearchPath(schema.getName());
+        boolean setSearchPath = CommonUtils.getBoolean(
+            getDataSource().getContainer().getConnectionConfiguration()
+                .getProviderProperty(PostgreConstants.PROP_SET_SEARCH_PATH_ON_SELECT),
+            false
+        );
+
+        if (setSearchPath) {
+            setSearchPath(monitor, schema);
+            setSearchPath(schema.getName());
```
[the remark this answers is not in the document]
labanovichttps: dude, both are nessesery

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
```diff
@@ -234,34 +235,32 @@ public List<String> getSearchPath() {
...
         }
 
+        newSearchPath.remove(defSchemaName);
+        newSearchPath.addFirst(defSchemaName);
+
+        if (activeUser != null && !newSearchPath.contains(activeUser)) {
+            newSearchPath.add(activeUser);
+        }
+
+        setUserInTheEndOfThePath(newSearchPath);
```
serge-rider: Why? Previously we added user only during connection init, why it is needed for every path change now?

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
```diff
@@ -234,34 +235,32 @@ public List<String> getSearchPath() {
...
         // Construct search path from current search path but put default schema first
         setSearchPath(monitor, schema.getName());
     }
 
     private void setSearchPath(DBRProgressMonitor monitor, String defSchemaName) throws DBCException {
-        List<String> newSearchPath = new ArrayList<>(getDefaultSearchPath());
-        int schemaIndex = newSearchPath.indexOf(defSchemaName);
-        /*if (schemaIndex == 0 || (schemaIndex == 1 && isUserFirstInPath(newSearchPath))) {
-            // Already default schema
+        LinkedList<String> newSearchPath = new LinkedList<>(getSearchPath());
```
serge-rider: Why linked list? It is a short list, ArrayList must be fine here

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
```diff
@@ -277,6 +276,20 @@ private void setSearchPath(DBRProgressMonitor monitor, String defSchemaName) thr
         }
     }
 
+    private void setSearchPath(String path) {
```
serge-rider: Add notnull annos

serge-rider (changes requested):

uslss:
verified

PR #39413, opened 2025-10-15
dbeaver/pro#6823 Add a short error message for OpenAI and Copilot

IvanGorshechnikov:
Please test all engines with different types of restrictions: wrong API key, wrong model name, restrictions to engine API

image

Please test both streaming and non-streaming mode. And it would nice to look it in CloudBeaver

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AbstractHttpAIClient.java
+++ b/plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AbstractHttpAIClient.java
@@ -16,13 +16,11 @@
  */
 package org.jkiss.dbeaver.model.ai.engine;
 
-import com.google.gson.JsonElement;
-import com.google.gson.JsonSyntaxException;
 import org.jkiss.code.NotNull;
 import org.jkiss.dbeaver.DBException;
 import org.jkiss.dbeaver.Log;
+import org.jkiss.dbeaver.model.ai.utils.AIHttpUtils;
 import org.jkiss.dbeaver.model.ai.utils.MonitoredHttpClient;
-import org.jkiss.dbeaver.model.data.json.JSONUtils;
 
 import java.net.http.HttpClient;
 
@@ -43,36 +41,8 @@ public abstract class AbstractHttpAIClient implements AutoCloseable {
     }
 
     @NotNull
-    protected DBException mapHttpError(int statusCode, @NotNull String body) {
-        return new DBException("AI request failed: " + statusCode + ", " + parseErrorMessage(body));
+    public DBException mapHttpError(int statusCode, @NotNull String body) {
+        log.debug("AI request failed: " + statusCode + ", " + body);
+        return new DBException("AI request failed: " + AIHttpUtils.parseErrorMessage(body));
     }
-
-    @NotNull
-    private static String parseErrorMessage(@NotNull String body) {
-        try {
-            JsonElement errorResponse = JSONUtils.GSON.fromJson(body, JsonElement.class);
-            if (errorResponse != null && errorResponse.isJsonObject()) {
-                if (errorResponse.getAsJsonObject().has("error")) {
-                    JsonElement errorElement = errorResponse.getAsJsonObject().get("error");
-                    if (errorElement.isJsonObject() && errorElement.getAsJsonObject().has("message")) {
-                        JsonElement messageElement = errorElement.getAsJsonObject().get("message");
-                        if (messageElement.isJsonPrimitive() && messageElement.getAsJsonPrimitive().isString()) {
-                            return messageElement.getAsString();
-                        }
-                    }
-                }
-                if (errorResponse.getAsJsonObject().has("message")) {
-                    JsonElement messageElement = errorResponse.getAsJsonObject().get("message");
-                    if (messageElement.isJsonPrimitive() && messageElement.getAsJsonPrimitive().isString()) {
-                        return messageElement.getAsString();
-                    }
-                }
-            }
-            return body;
-        } catch (JsonSyntaxException e) {
-            log.debug("Failed to parse error response: " + e.getMessage());
-            return body;
-        }
-    }
-
 }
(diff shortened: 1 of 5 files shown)


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/AIHttpUtils.java
```diff
@@ -77,4 +80,28 @@ public static DBException mapHttpError(@NotNull HttpResponse<String> response) {
         };
     }
 
+
+    @NotNull
+    public static String parseErrorMessage(@NotNull String body) {
```
ShadelessFox: No... it can't be generic. You're expecting it to be a JSON body, but it can be anything, really. It must be handled by the engine/client itself.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/OpenAIClient.java
```diff
@@ -143,10 +142,8 @@ public OAIResponsesResponse createChatCompletion(
         String body = response.body();
         if (response.statusCode() == 200) {
             return GSON.fromJson(body, OAIResponsesResponse.class);
-        } else if (response.statusCode() == 429) {
-            throw new TooManyRequestsException("Too many requests: " + body);
         } else {
-            throw new DBException("OpenAI request failed: " + response.statusCode() + ", body=" + body);
+            throw new DBException("OpenAI request failed: " + response.statusCode() + ", " + AIHttpUtils.parseErrorMessage(body));
```
ShadelessFox: Why are you embedding the parsed message here in the exception's name? We have a dedicated AI transport exception that already includes both status and body that can later be parsed and transformed by the AI chat.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/AIHttpUtils.java
serge-rider: As far as we saw it most engine report errors in OpenAI style. So it is more or less ok.
I'd just add some easy check before parse (e.g. check that body starts with { character).
If it is not json then return body.

Anyhow, let's rename parseErrorMessage into parseStandardJsonErrorMessage

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/AIHttpUtils.java
ShadelessFox: We've discussed and come up with a generally better approach for dealing with errors. Currently, each engine handles errors independently in several places, while in others, it's handled by the `MonitoredHttpClient`.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AbstractHttpAIClient.java
```diff
@@ -0,0 +1,78 @@
...
+        client.close();
+    }
+
+    @NotNull
+    protected DBException mapHttpError(int statusCode, @NotNull String body) {
+        return new DBException("AI request failed: " + statusCode + ", " + parseErrorMessage(body));
+    }
+
+    @NotNull
+    private static String parseErrorMessage(@NotNull String body) {
```
ShadelessFox: Why is it still here? The intention was to make `mapHttpError` abstract so each caller is responsible for handling the output. This particular function, `parseErrorMessage`, can be moved to a utility class and called in every implementation. This way, the intent (that the implementation is responsible for handling errors) is clear.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AbstractHttpAIClient.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Override
+    public void close() {
+        client.close();
+    }
+
+    @NotNull
+    protected DBException mapHttpError(int statusCode, @NotNull String body) {
+        return new DBException("AI request failed: " + statusCode + ", " + parseErrorMessage(body));
```
ShadelessFox: I  don't like that the message includes both the status code and the prefix "AI request failed". This can be logged, but the exception itself must only contain the "mapped" message. Alternatively, use a separate exception type that includes the "mapped" message as a separate property, which can be retrieved and displayed in the chat.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/MonitoredHttpClient.java
```diff
@@ -30,10 +30,22 @@
 import java.util.stream.Collectors;
 
 public class MonitoredHttpClient implements AutoCloseable {
+    @FunctionalInterface
```
ShadelessFox, line 36: Having a short javadoc describing the purpose of this class would be nice

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/MonitoredHttpClient.java
```diff
@@ -46,7 +58,7 @@ public HttpClient getHttpClient() {
      * The request is sent asynchronously and the method will block until the response is received.
      * The method will also check if the progress monitor is cancelled and cancel the request if it is.
      */
-    public HttpResponse<String> send(
+    public String send(
```
ShadelessFox, line 61:
```suggestion
    @NotNull
    public String send(
```

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/MonitoredHttpClient.java
```diff
@@ -109,4 +126,9 @@ public CompletableFuture<Void> sendAsync(
     public void close() {
         client.close();
     }
+
+    @NotNull
+    private static DBException defaultErrorMapper(int statusCode, @NotNull String body) {
```
ShadelessFox: This should be removed.

ShadelessFox: As well as the default constructor that utilizes it

ShadelessFox (changes requested):

ShadelessFox (approved):

MashaKorax:
CB
chrome_8jC5qC4LEu

PR #39414, opened 2025-10-15
dbeaver/dbeaver#38873 Add DuckDB execution context handling

### LINKED ISSUE #38873: Scripts losing catalog details on disconnect
### Description

Last selection of Catalog is lost after disconnecting from db connection.
Active Data Source is retained. 
Change has only appeared since upgrading to 25.1.3

### DBeaver Version

Community Edition 25.1.3.202507201434

### Operating System

Windows 10 Ent

### Database and driver

DuckDb 
Driver JAR 1.2.2.0  (local file)

### Steps to reproduce

Just had forced upgrade to 25.1.3  (sorry, not sure what version it was previously but was very recent)
We refresh the local databases very frequently, so disconnect to refresh.
On reconnecting the local Active Data Source selection is preserved but the Catalog selection is lost, so goes to N/A
Not a show stopper as can be reselected but is frustrating as this has never been required before.

Nothing obvious in the Release Notes, except about losing schema selection n 25.1.2 ???

### Additional context

All 100% local. No network connectivity. 
Same behaviour on multiple users, all with same forced upgrade

labanovichttps:
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.duckdb/src/org/jkiss/dbeaver/ext/duckdb/model/DuckDBExecutionContext.java
+++ b/plugins/org.jkiss.dbeaver.ext.duckdb/src/org/jkiss/dbeaver/ext/duckdb/model/DuckDBExecutionContext.java
@@ -1,0 +1,127 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2025 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.ext.duckdb.model;
+
+import org.jkiss.code.NotNull;
+import org.jkiss.dbeaver.DBException;
+import org.jkiss.dbeaver.ext.generic.model.GenericCatalog;
+import org.jkiss.dbeaver.ext.generic.model.GenericExecutionContext;
+import org.jkiss.dbeaver.ext.generic.model.GenericSchema;
+import org.jkiss.dbeaver.model.connection.DBPConnectionBootstrap;
+import org.jkiss.dbeaver.model.exec.DBCCachedContextDefaults;
+import org.jkiss.dbeaver.model.exec.DBCException;
+import org.jkiss.dbeaver.model.exec.DBCExecutionPurpose;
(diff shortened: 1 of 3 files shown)


IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ext.duckdb/src/org/jkiss/dbeaver/ext/duckdb/model/DuckDBExecutionContext.java
```diff
@@ -0,0 +1,127 @@
...
+            } catch (DBException ex) {
+                log.warn("Error setting active catalog", ex);
+            }
+        }
+
+        super.setDefaultCatalog(monitor, catalog, schema);
+    }
+
+    @Override
+    public void initDefaultsFrom(DBRProgressMonitor monitor, GenericExecutionContext context) {
```
ShadelessFox, line 110: Why is it required?

plugins/org.jkiss.dbeaver.ext.duckdb/src/org/jkiss/dbeaver/ext/duckdb/model/DuckDBExecutionContext.java
```diff
@@ -0,0 +1,127 @@
...
+
+        if (supportsCatalogChange() && CommonUtils.isEmpty(activeCatalogName)) {
+            try (JDBCSession session = openSession(monitor, DBCExecutionPurpose.META, "Query active catalog")) {
+                String currentCatalog = JDBCUtils.queryString(session, "SELECT current_catalog()");
+                if (!CommonUtils.isEmpty(currentCatalog) && !currentCatalog.equals(activeCatalogName)) {
+                    activeCatalogName = currentCatalog;
+                    changed = true;
+                }
+            } catch (Exception ignored) {
+                //ignored
```
ShadelessFox: Never consume exceptions. Log them

plugins/org.jkiss.dbeaver.ext.duckdb/src/org/jkiss/dbeaver/ext/duckdb/model/DuckDBExecutionContext.java
labanovichttps: because it restores previous catalog

plugins/org.jkiss.dbeaver.ext.duckdb/src/org/jkiss/dbeaver/ext/duckdb/model/DuckDBExecutionContext.java
labanovichttps: fixed

plugins/org.jkiss.dbeaver.ext.duckdb/src/org/jkiss/dbeaver/ext/duckdb/model/DuckDBExecutionContext.java
ShadelessFox: Seems like a dirty solution for a weird issue. It was never required before. Have you seen how other execution contexts are implemented?

uslss:
verified

PR #39415, opened 2025-10-15
6956 core redesign cli tools

alexander-skoblikov:
closes https://github.com/dbeaver/pro/issues/6956


plugins/org.jkiss.dbeaver.model.cli/src/org/jkiss/dbeaver/model/cli/model/option/OutputFileOption.java
```diff
@@ -0,0 +1,38 @@
+/*
```
serge-rider, line 1: OutputFileOption ?
And maybe move it to model/cli/options/file package?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -180,19 +181,17 @@ public DBPPreferenceStore getPreferenceStore() {
     public Object start(IApplicationContext context) {
         instance = this;
 
+        var args = Platform.getApplicationArgs();
```
serge-rider: Move it to preprocessCommandLineParameters function. Add a comment

serge-rider (changes requested):

PR #39416, opened 2025-10-15
dbeaver/pro#6364 add option to set default view for project

HocKu7:
Closes https://github.com/dbeaver/pro/issues/6364


plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMController.java
```diff
@@ -239,7 +239,8 @@ void deleteAllObjectPermissions(
     Map<String, Object> getObjectSettings(
         @NotNull String objectId,
         @NotNull SMObjectType objectType,
-        @Nullable String settingId
+        @Nullable String subjectId,
```
serge-rider, line 242: Why we pass subject id here? It must work for current user always.
And we need setObjectSettings for current user here.

plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/security/SMAdminController.java
```diff
@@ -233,4 +233,18 @@ void deleteObjectPermissions(
     @NotNull
     List<SMTeamMemberInfo> getTeamMembersInfo(@NotNull String teamId) throws DBException;
 
+    void setObjectSettings(
```
serge-rider, line 236: rename into setObjectSettingsForSubject and setObjectSettingsForSubject
This is pure admin API and can be used in admin tools only.

Regular users must use function without subject id.

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceNavigatorSettings.java
```diff
@@ -197,6 +211,28 @@ public boolean equals(Object obj) {
     private static final String DEFAULT_MERGE_SCHEMAS = "navigator.settings.default.hideSchemas";
     private static final String DEFAULT_HIDE_VIRTUAL_MODEL = "navigator.settings.default.hideVirtualModel";
 
+    //fixme (?) limit of 32 characters in database setting object key
+    private static final String DEFAULT_SHOW_SYSTEM_OBJECTS_SETTING = DEFAULT_SHOW_SYSTEM_OBJECTS.split("\\.")[3];
```
serge-rider, line 215: ???

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceNavigatorSettings.java
```diff
@@ -48,6 +49,19 @@ public class DataSourceNavigatorSettings implements DBNBrowseSettings {
         RegistryMessages.navigator_settings_preset_custom_view_description
     );
 
+    public void loadSettings(Map<String, Object> projectNavigatorSettings) {
```
serge-rider, line 52: I don't think that we need such a function here

serge-rider (changes requested):

PR #39437, opened 2025-10-17
dbeaver/dbeaver#38510 Add “Exclude ID-like columns” option to disable grouping for numeric ID field

### LINKED ISSUE #38510: number seprators for number columns, no separator for id columns using integers
Many people, even where its suboptimal will still have tables where the id is an integer column. This can be amongst many other columns in the same table that have numeric types where separators ',' are extremely handy however they're a fairly startling visual jolt in an id column setting. 

It would be abolutely awesome to be able to have separators for numeric columns where it contextually makes sense and not have them for ID/id columns where it feels very alien. 

I currently work without separators because seeing them in ids is uncomfortable but this means human err is far likelier than is ideal. 

**Additional context**
Thanks for your consideration. Although a convenience feature I feel the cognitive payoff here is worth the effort. Hopefully that resonates with others too.

vkirst:
Number formatting: add “Exclude ID-like columns” option and heuristics to disable grouping for numeric ID columns (id/user_id/etc); define in plugin.xml, set defaults, and apply in DataFormatterProfile.


plugins/org.jkiss.dbeaver.model/plugin.xml
```diff
@@ -39,6 +39,7 @@
             <propertyGroup label="%DateFormatter.number.general.label">
                 <property id="useGrouping" label="%DateFormatter.number.general.grouping.label" type="boolean" description="%DateFormatter.number.general.grouping.description"/>
                 <property id="groupingSize" label="%DateFormatter.number.general.groupingSize.label" type="integer" description="%DateFormatter.number.general.groupingSize.description"/>
+                <property id="excludeIdColumns" label="%DateFormatter.number.general.excludeIdColumns.label" type="boolean" description="%DateFormatter.number.general.excludeIdColumns.description"/>
```
ShadelessFox, line 42: Maybe we should instead make it a regex pattern that the user can supply themselves?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/registry/formatter/DataFormatterProfile.java
```diff
@@ -199,21 +205,80 @@ public DBDDataFormatter createFormatter(@NotNull String typeId, DBSTypedObject t
...
+            }
+        }
+
+        DBDDataFormatter formatter = descriptor.createFormatter();
         formatter.init(type, locale, formatterProps);
+
         return formatter;
     }
 
+    private boolean isIdLike(@NotNull DBSTypedObject type) {
```
ShadelessFox: I don't like this hardcoded detection logic. It also doesn't have any tests.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/registry/formatter/DataFormatterProfile.java
```diff
@@ -199,21 +203,93 @@ public DBDDataFormatter createFormatter(@NotNull String typeId, DBSTypedObject t
...
         Map<String, Object> formatterProps = new HashMap<>();
         if (defProps != null && !defProps.isEmpty()) {
             formatterProps.putAll(defProps);
         }
-        if (props != null && !props.isEmpty()) {
+        if (!props.isEmpty()) {
             formatterProps.putAll(props);
         }
+
+        if ("number".equalsIgnoreCase(typeId)) {
```
serge-rider: DBDDataFormatter.TYPE_NAME_NUMBER

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/registry/formatter/DataFormatterProfile.java
```diff
@@ -199,21 +203,93 @@ public DBDDataFormatter createFormatter(@NotNull String typeId, DBSTypedObject t
...
+            }
+        }
+
+        DBDDataFormatter formatter = descriptor.createFormatter();
         formatter.init(type, locale, formatterProps);
+
         return formatter;
     }
 
+    static boolean isPrimaryKey(DBSTableColumn column) {
```
serge-rider: NUllability annotations

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/registry/formatter/DataFormatterProfile.java
```diff
@@ -199,21 +203,93 @@ public DBDDataFormatter createFormatter(@NotNull String typeId, DBSTypedObject t
...
+        var monitor = new VoidProgressMonitor();
+        try {
+            return matchesColumnInRefs(
+                column,
+                referrers(column.getParentObject().getConstraints(monitor), DBSEntityConstraintType.PRIMARY_KEY),
+                monitor
+            );
+        } catch (DBException e) {
```
serge-rider: Do not ignore exception, use log.debug

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/registry/formatter/DataFormatterProfile.java
```diff
@@ -199,21 +203,93 @@ public DBDDataFormatter createFormatter(@NotNull String typeId, DBSTypedObject t
...
+        Stream<DBSEntityReferrer> referrers,
+        DBRProgressMonitor monitor
+    ) {
+        return referrers
+            .flatMap(ref -> safeRefs(ref, monitor))
+            .anyMatch(r -> r.getAttribute() == column);
+    }
+
+
+    private static Stream<? extends DBSEntityAttributeRef> safeRefs(DBSEntityReferrer ref, DBRProgressMonitor monitor) {
```
serge-rider: - Monitor always first parameter
- Move to DBStructUtils

serge-rider (changes requested):

IvanGorshechnikov (dismissed):

Matvey16:
Doesn't work for a PK in SQLite

https://github.com/user-attachments/assets/4d9d16f5-0143-4069-a348-5c02a34cfd28

Also, because now it should work for key values and not based on heuristics, the name and the description of this parameter must be changed, they are misleading for the user

Matvey16:
Verified

PR #39441, opened 2025-10-19
dbeaver/pro#7205 Fix redundant icons

serge-rider:


Matvey16:
Now secrets enterd for the connection are ignored (for every provider)

Example for AWS with invalid secret

Actual:

dbeaver_SNLcuvJjah

Expected:

dbeaver_yTT2DK9K45

Matvey16:
Verified

PR #39463, opened 2025-10-21
dbeaver/pro#7156 Add link parsing in AI Chat

IvanGorshechnikov:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AITextUtils.java
```diff
@@ -148,14 +152,96 @@ public static MessageChunk[] splitIntoChunks(@NotNull SQLDialect dialect, @NotNu
...
 
         return chunks.toArray(MessageChunk[]::new);
     }
 
+    private static void addTextWithLinks(
+        @NotNull List<MessageChunk> chunks,
+        @NotNull String text
+    ) {
+        final Pattern markdownLinkPattern = Pattern.compile("\\[([^]]+)]\\(([^)]+)\\)");
+        final Pattern urlPattern = Pattern.compile("\\b(https?://|ftp://)[^\\s<>\"{}|\\\\^`\\[\\]]+");
```
serge-rider: It is better to make patterns as constants.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AITextUtils.java
```diff
@@ -148,14 +152,96 @@ public static MessageChunk[] splitIntoChunks(@NotNull SQLDialect dialect, @NotNu
...
+            allMatches.add(new MatchInfo(
+                urlMatcher.start(),
+                urlMatcher.end(),
+                urlMatcher.group(),
+                urlMatcher.group(),
+                false
+            ));
+        }
+
+        allMatches.sort(Comparator.comparingInt(a -> a.start));
```
serge-rider, line 200: Why? It already should be in order.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AITextUtils.java
IvanGorshechnikov: No, they aren't sorted.
The sorting is necessary because the code collects matches from two separate regex patterns (markdown links and plain URLs) into one list. Without sorting, they would be ordered by pattern type rather than by position in the text.

PR #39466, opened 2025-10-22
dbeaver/pro#7116 Replace Crud rs icons

labanovichttps:
image


serge-rider:
I don't like circles. Red one looks like a brick (stop, no go, etc).
It doesn't look like "delete". We can use green plus and red cross for add/remove.
x2 should have the same color as "add". And it may look like copy and plus.

arhayka (approved):

uslss:
verified

PR #39470, opened 2025-10-22
Update InterSystems IRIS driver version spec

### LINKED ISSUE #39542: Update InterSystems IRIS driver version spec
This ticket was created based on this contribution: https://github.com/dbeaver/dbeaver/pull/39470

bdeboe:
The current RELEASE version spec for the InterSystems IRIS driver resulted in a legacy driver being picked (named 2018.xxx) after the year of release) that predated the semantic versioning numbers used these days. I've now added a simple selector to opt for the current major version (carving out room for the next) to avoid this.

Also fixing a few typos in the original submission ("intersystem" > "intersystems").


arhayka:
I created the ticket to track review and testing: https://github.com/dbeaver/dbeaver/issues/39542

IvanGorshechnikov (dismissed):

Destrolaric (approved):

Matvey16:
This makes the driver undownloadable

dbeaver_6QslddHsDJ

bdeboe:
Hi @Matvey16 , thanks for looking into this. I'm not sure how you could actually end up in that state. This PR was supposed to get rid of the `RELEASE` artifact version. Can you share the full stack trace or tell me how to reproduce this?

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -389,13 +389,13 @@
...
-                    description="InterSystems IRIS."
+                    description="InterSystems IRIS"
                     supportedConfigurationTypes="MANUAL,URL"
                     categories="sql,object">
 <!--                    <replace provider="generic" driver="cache"/>-->
-                    <file type="jar" path="maven:/com.intersystems:intersystems-jdbc:RELEASE" bundle="!drivers.intersystem"/>
+                    <file type="jar" path="maven:/com.intersystems:intersystems-jdbc:RELEASE[3.10,999]" bundle="!drivers.intersystems"/>
                     <file type="license" path="licenses/external/mit.txt"/>
-                    <file type="jar" path="drivers/intersystem" bundle="drivers.intersystem"/>
+                    <file type="jar" path="drivers/intersystems" bundle="drivers.intersystems"/>
```
Destrolaric: For now, I would recommend leaving
`intersystem`, the name is used for predownloaded drivers in the PRO version. 
Additionally, please try using 
`maven:/com.intersystems:intersystems-jdbc:RELEASE[3.10.5]` as a maven path

Destrolaric (dismissed):

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
bdeboe: @Destrolaric thanks for your review. 
I'd rather make sure the maven path leaves room for future releases. It just needs to steer clear of a bad submission in 2020 that didn't adhere to semantic versioning, which Maven doesn't allow us to take down. With your suggested range, we'd have to submit a PR each time there's a relevant update, and we don't want to give you extra work :-)

serge-rider: Let's just set version to 3.10.5? At least we know that it is a working version.
It is possible to change it manually.
In future we can update it to a higher recommended version.

This is how it work for all drivers.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
bdeboe: OK, I pinned to 3.10.5 rather than keep it open ended. 
We can indeed update it when there's a good reason to do so.

Destrolaric (approved):

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
```diff
@@ -389,13 +389,13 @@
...
-                    description="InterSystems IRIS."
+                    description="InterSystems IRIS"
                     supportedConfigurationTypes="MANUAL,URL"
                     categories="sql,object">
 <!--                    <replace provider="generic" driver="cache"/>-->
-                    <file type="jar" path="maven:/com.intersystems:intersystems-jdbc:RELEASE" bundle="!drivers.intersystem"/>
+                    <file type="jar" path="maven:/com.intersystems:intersystems-jdbc:RELEASE[3.10.5]" bundle="!drivers.intersystems"/>
                     <file type="license" path="licenses/external/mit.txt"/>
-                    <file type="jar" path="drivers/intersystem" bundle="drivers.intersystem"/>
+                    <file type="jar" path="drivers/intersystems" bundle="drivers.intersystems"/>
```
Destrolaric, line 398: Please avoid renaming `drivers/intersystem` and bundle `drivers.intersystem` (might lead to issues with backward compatibility and enterprise client)

Destrolaric (changes requested):

plugins/org.jkiss.dbeaver.ext.generic/plugin.xml
bdeboe: Hi @Destrolaric , wouldn't this just download the driver to the new location or are there other effects? I'm not sure what you mean with "enterprise clients". I work for the InterSystems enterprise :-), and we'd like to rectify this unfortunate typo, ahead of filing a PR for cloudbeaver as well. 
Thanks in advance for your response!

Matvey16:
Verified

PR #39485, opened 2025-10-23
dbeaver/dbeaver#38887: add force focus checkbox

### LINKED ISSUE #38887: disable focus stealing
Imagine you are working on something very important and suddenly dbeaver steals your focus to tell you that the connection to the database is disconnected. I don't think anybody cares about this, if I am not working inside the dbeaver, I don't want it to change the application I am currently working on. this is such an annoying thing that I uninstalled this program. please do not steal the focus.

SolovevAnton:
closes dbeaver/dbeaver#38887. Added checkbox to set force focus or disable it. should be false by default
when checked- must set focus on notifications no matter what narrator is chosen, otherwise should not.
image


plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/notifications/DatabaseNotificationSink.java
```diff
@@ -153,11 +152,14 @@ private void showPopup() {
...
-        String storedScreenReader = store.getString(ScreenReaderPreferences.PREF_SCREEN_READER_ACCESSIBILITY);
-        ScreenReader screenReader = ScreenReader.getScreenReader(storedScreenReader);
+        boolean isForceFocus = store.getBoolean(ScreenReaderPreferences.PREF_FORCE_FOCUS_ON_EDITOR);
         // set focus for all screen readers  
-        if (!screenReader.equals(ScreenReader.DEFAULT)) {
-            popup.setFocus();
+        if (isForceFocus) {
+            String storedScreenReader = store.getString(ScreenReaderPreferences.PREF_SCREEN_READER_ACCESSIBILITY);
+            ScreenReader screenReader = ScreenReader.getScreenReader(storedScreenReader);
+            if (!screenReader.equals(ScreenReader.DEFAULT)) {
```
IvanGorshechnikov: I think that we don't need this condition anymore
It will be strange behaviour if you check preferense of "Force focus" on Default Screen Reader and don't get it

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/CoreResources_ru.properties
```diff
@@ -447,6 +447,8 @@ pref_page_ui_general_boolean_color = Цвет
 pref_page_ui_general_boolean_color_use_theme_color = Использовать цвет темы
 pref_page_ui_general_boolean_invalid_values = Стили флажков содержат недопустимые значения
 
+pref_page_accessibility_force_focus_checkbox = Форсировать фокус на уведомлениях
```
IvanGorshechnikov: I think better would be "Переносить фокус на уведомления", "Форсировать" sounds strange

uslss:
verified

PR #39486, opened 2025-10-23
dbeaver/pro#6014 Fix splash update on Linux

ShadelessFox:
Currently, on Linux (KDE) against `devel`, the splash appears pitch black for several seconds.

Please test on **all** platforms: Windows, Linux, and macOS.


MashaKorax:
win - ok 
linux -ok
mac -ok

PR #39491, opened 2025-10-24
Dbeaver/pro#6973 fix snowflake functions

labanovichttps:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -1232,4 +1232,59 @@ public static String identifierToCanonicalForm(
...
+
+    /**
+     * <p>Removes parameter names and preserves type details and nesting.</p>
+     *
+     * <pre>{@code
+     * Input : "(x ARRAY, y OBJECT)"
+     * Output: "(ARRAY, OBJECT)"
+     * }</pre>
+     */
+    public static String extractTypesOnly(@Nullable String sig) {
```
serge-rider: extractProcedureParameterTypes

serge-rider (changes requested):

PR #39500, opened 2025-10-27
dbeaver/pro#7280 Fix external arrow not rendering correctly on macOS

ShadelessFox:
Before:
image

After:
image

Can be found in any connection's wizard under **Connection Settings -> Initialization**. 

Worth checking on all platforms just in case.


Matvey16:
Verified

PR #39504, opened 2025-10-28
dbeaver/pro#6446 Added connection and model features

Nexus6v2:
Added connection and model features for CB web properties

Closes https://github.com/dbeaver/pro/issues/6446


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIConstants.java
```diff
@@ -41,5 +41,8 @@ public class AIConstants {
     public static final String AI_JOIN_RULE = "ai.joinRule";
     public static final int DEFAULT_CONTEXT_WINDOW_SIZE = 16384;
     public static final String AI_USE_STREAM_MODE = "ai.useStreamMode";
+
+    public static final String AI_FEATURE_CONNECTION_PROPERTY = "connection";
```
serge-rider: I don't like these features. They are designed specifically for a particular web form, they don't have ny generic meaning.

Nexus6v2: I made it non-generic on purpose, to avoid it to be reused elsewhere.

We need very specific logic connected to these features. 
Otherwise, we just can't determine which fields are supposed to do what. On frontend we need both, on backend only "model".

serge-rider: And this makes this entire feature very questionable.
Perhaps we need to discuass it on tech meeting again.

Server API mustn't know about any forms or any front-end specific issues.

My current mood is - let's just keep this form static and avoid any new annotations/features.

Wroud: There should be only one abstraction on the model side:
1. fields required for connection
2. fields for model configuration

And this is not related in any way to the interface or forms, it's just a base meaning of how the engine can be configured, ui or other dependents might or might not use this separation for any reason, for example, to `connect` to the Engine first to load metadata (like available models)

If there is no such separation, this makes model UNUSABLE because you can't determine what data depends on what fields (available models depends on the connection status and connection fields)

**In simple words:
To load the model's information, you must first connect to the AI Model provider, which requires specifying specific fields. This is just how it works.**

yagudin10 (dismissed):

serge-rider (dismissed):

PR #39505, opened 2025-10-28
dbeaver/pro#7076 Export org.jkiss.dbeaver.ext.athena package

vkirst:


uslss:
cannot connect
image

serge-rider:
Fixed

uslss:
dbeaver - ok
image
cb - ok
image

PR #39535, opened 2025-10-31
dbeaver/pro#7336 Disable splash update on old MacOSes

serge-rider:


plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -269,7 +269,9 @@ public Object start(IApplicationContext context) {
         // It sets main windows name and images
         getDisplay();
 
-        updateSplashHandler();
+        if (!RuntimeUtils.isMacOS() || !RuntimeUtils.isOSVersionAtLeast(14, 0, 0)) {
```
ShadelessFox: Should be this. Can be read as "if not macOS, or macOS' version is at least 14, then update".
```suggestion
        if (!RuntimeUtils.isMacOS() || RuntimeUtils.isOSVersionAtLeast(14, 0, 0)) {
```

ShadelessFox (changes requested):

ShadelessFox (approved):

Matvey16:
Splash screen works fine on MacOS 12.6
Seems like still the same problem on MacOS 14

PR #39539, opened 2025-10-31
dbeaver/dbeaver#39496 Load table statistics when refreshing a single table

### LINKED ISSUE #39496: Table size not refreseshed
### Description

Size of table in menu si refreshed only, when I refresh all tables (click on "folder" Tables - press F5). But, when I select single table, right click on table - select option "Refresh" - size of table is missing. 

It looks it not working on MariaDB.

I tested also PostgreSQL, Vertica - here it works correctly. 

I refreshed two tables, here is result:

Image

### DBeaver Version

25.2.3

### Operating System

Windows

### Database and driver

MariaDB
Driver: org.mariadb.jdbc.Driver

org.mariadb.jdbc:mariadb-java-client:RELEASE [3.5.2]  -> mariadb-java-client-3.5.2.jar

DriverType: MySQL

### Steps to reproduce

_No response_

### Additional context

_No response_

vkirst:


IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLCatalog.java
```diff
@@ -626,10 +627,27 @@ protected MySQLTableBase fetchObject(@NotNull JDBCSession session, @NotNull MySQ
...
                 return new MySQLView(owner, dbResult);
             } else {
-                return new MySQLTable(owner, dbResult);
+                MySQLTable mySQLTable = new MySQLTable(owner, dbResult);
+                loadAdditionalInfoIfNeed(owner, mySQLTable);
+                return mySQLTable;
             }
         }
 
+        private void loadAdditionalInfoIfNeed(@NotNull MySQLCatalog owner, @NotNull MySQLTable mySQLTable) {
```
ShadelessFox: This is bad... So it refreshes the additional info of the old instance of `MySQLTable` that `mySQLTable` is supposed to replace? Shouldn't the new object replace the one in the cache? How does it work even?

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLCatalog.java
ShadelessFox: It also breaks the idea of the additional info being lazy to load.

vkirst: The new instance actually reloads the additional info
`mySQLTable.getAdditionalInfo(new VoidProgressMonitor());`
 — it doesn’t copy anything from the cached table.
The cached object is used only as an indicator that this table had statistics already loaded before.
If so, we trigger a fresh getAdditionalInfo() call for the new table instance, to ensure that the “Size” column in the navigator stays up-to-date after a single-table refresh.

So it doesn’t break the lazy-loading model — it only re-loads additional info when it was already loaded previously.

ShadelessFox: I don't get it. Won't the navigator detect that the object doesn't have its statistics cached, because it's a new object, and load it?

vkirst: No, this is exactly what the task is about

ShadelessFox: Then the issue is somewhere within the navigator itself. It should not have anything to do with the model, as the object is already fresh and new and clearly doesn't have any statistics.

ShadelessFox (approved):

IvanGorshechnikov (approved):

uslss:
verified

serge-rider:
Again, this shouldn't be done just for one driver.
It has to be a general fix

PR #39561, opened 2025-11-03
dbeaver/dbeaver#39373 Fix execute query handler (ctrl+enter)

### LINKED ISSUE #39373: 'ctrl + enter' in replace field doesn't work any more
### Description

A few version before I could do a 'ctrl + enter' in the replace input field and then all values will be replaced with the new one.
But now it fires a query instead.

to reproduce:
1. open a the SQL editor
2. hit 'CTRL + F'
3. hit 'CTRL + R'
4. add some values you want to change
5. press 'ctrl + enter' (as you can see in picture below)

Image

### DBeaver Version

25.2.2.202510052342

### Operating System

win 11

### Database and driver

doesn't invole a DB connection

### Steps to reproduce

_No response_

### Additional context

to reproduce:
1. open a the SQL editor
2. hit 'CTRL + F'
3. hit 'CTRL + R'
4. add some values you want to change
5. press 'ctrl + enter' (as you can see in picture below)

Image

serge-rider:


Matvey16:
Verified

PR #39563, opened 2025-11-04
dbeaver/dbeaver#39369 Legacy OpenAI client

### LINKED ISSUE #39369: Legacy OpenAI API support
### Description

Does the AI integration setting only support OpenAI Responses and not OpenAI Chat? When using a local model, the request URL is automatically appended with a "/responses" suffix, causing the request to fail.
I recall this issue did not exist in the previous version (before the problem of automatic conversion of uppercase letters to lowercase was fixed). Back then, the error message was "xxx model does not exist", whereas the current error message is "detail": "Not Found".

### DBeaver Version

25.2.2

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

serge-rider:


serge-rider:
image

PR #39569, opened 2025-11-04
#39566 Fix processor class check

### LINKED ISSUE #39566: "Write to the single file" disabled
### Description

I updated DBeaver to version 25.2.3.202510191611, and the option “Write to single file” is disabled. I’m exporting data from three queries and want to save it in a single .csv file. Any ideas on how to enable this? threads is set to 1.

### DBeaver Version

25.2.3.202510191611

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

ShadelessFox:


uslss:
verified

PR #39577, opened 2025-11-05
Fix execution time display for sub-10-second queries

### LINKED ISSUE #39556: Not able to see the execution time in milliseconds from Query Manager and Status bar
### Description

When I am running a SQL query, the execution time is in seconds instead of milliseconds. 
I had this  issue with older versions of DBeaver. I tried to change a lot of settings/property from the Preferences window without success.
My colleague has installed DBeaver also. He is connecting to the same DBs as me (same kind of connections) and he is seeing the execution time in milliseconds. 
I desinstalled completely DBeaver and reinstalled it without changing any settings or preferences. I still see the execution time in seconds instead of milliseconds.

Image

I really need to some help

### DBeaver Version

25.2.4

### Operating System

windows 11

### Database and driver

Database name: DB2 for LUW

### Steps to reproduce

1. Install DBeaver as is without changing any settings at the installation.
2. Create a DB2 for LUW connection
3. Run a simple query
4. As my older version 25.2.3 and previous versions, the execution time is in seconds instead of being in milliseconds.
5. See the screenshot added against the description (above)

### Additional context

_No response_

ritoban23:
### Fix: Show Execution Time in Milliseconds for Fast Queries

This PR fixes issue #39556 by improving the precision of execution time display in Query Manager and Status bar. Previously, queries faster than 1 second were shown as "0.0s" or "1.0s", losing millisecond detail. Now, times under 10 seconds are shown with one decimal place (e.g., "0.5s" for 500ms, "0.1s" for 50ms), restoring the expected behavior.

**Summary of changes:**
- Fixed `RuntimeUtils.formatExecutionTime()` to use `String.format("%.1fs", seconds + millis / 1000.0)` for sub-10-second durations.

Closes #39556.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/RuntimeUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/RuntimeUtils.java
@@ -212,7 +212,7 @@ public final class RuntimeUtils {
         } else if (seconds >= 10) {
             return String.format("%ds", seconds);
         } else {
-            return String.format("%d.%ds", seconds, millis / 100);
+            return String.format("%.1fs", seconds + millis / 1000.0);
         }
     }
 


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/RuntimeUtils.java
```diff
@@ -212,7 +212,7 @@ public static String formatExecutionTime(@NotNull Duration duration) {
         } else if (seconds >= 10) {
             return String.format("%ds", seconds);
         } else {
-            return String.format("%d.%ds", seconds, millis / 100);
+            return String.format("%.1fs", seconds + millis / 1000.0);
```
ShadelessFox, line 215: Why not just revert what d23a47481edd9f21f314707f841605113b8c681b did?


```suggestion
            return String.format("%d.%03ds", seconds, millis);
```

ShadelessFox:
You also need to update tests. I believe you haven't run them locally.

ShadelessFox:
There were no updates for the past two weeks. Closing as stale.

PR #39583, opened 2025-11-05
Dbeaver/pro#7349 redshift driver update to 2.2

vkirst:


plugins/org.jkiss.dbeaver.ext.postgresql/plugin.xml
```diff
@@ -515,9 +515,8 @@
                     <replace provider="generic" driver="aws_redshift"/>
                     <replace provider="generic" driver="aws_redshift_42"/>
 
-                    <file type="jar" path="maven:/com.amazon.redshift:redshift-jdbc42:RELEASE[2.1.0.8]" load-optional-dependencies="true" bundle="!drivers.redshift"/>
+                    <file type="jar" path="maven:/com.amazon.redshift:redshift-jdbc42:2.2.0" load-optional-dependencies="true" bundle="!drivers.redshift"/>
```
serge-rider: IT should be RELEASE[2.2.0]

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCExecutionContext.java
```diff
@@ -64,22 +67,30 @@ public class JDBCExecutionContext extends AbstractExecutionContext<JDBCDataSourc
     private volatile Boolean autoCommit;
     private volatile Integer transactionIsolationLevel;
     private transient volatile boolean txnIsolationLevelReadInProgress;
-    private final ReentrantLock queryExecutionLock;
+    private OwnershipLock statementLock = new OwnershipLock() {
```
serge-rider: Let's create NoOpLock and use a constant for it.

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/JDBCExecutionContext.java
```diff
@@ -64,22 +67,30 @@ public class JDBCExecutionContext extends AbstractExecutionContext<JDBCDataSourc
...
+    };
 
     public JDBCExecutionContext(@NotNull JDBCRemoteInstance instance, String purpose) {
         super(instance.getDataSource(), purpose);
         this.instance = instance;
         if (!instance.getDataSource().getContainer().getDriver().isThreadSafeDriver()) {
-            queryExecutionLock = new ReentrantLock();
-        } else {
-            queryExecutionLock = null;
+            statementLock = new OwnershipLock();
```
serge-rider: Let's call it SingleThreadLock

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTable.java
```diff
@@ -182,18 +182,23 @@ public DBCStatistics readData(
         statistics.setQueryText(sqlQuery);
 
         monitor.subTask(ModelMessages.model_jdbc_fetch_table_data);
+        DBCStatement dbStat = null;
```
serge-rider: Why changing code in JDBCTable?
We have zillion places where we execute statements.

serge-rider (changes requested):

PR #39585, opened 2025-11-05
dbeaver/pro#7048 Added support for embeddings to the Azure and Copilo…

ivgag:
…t agents


plugins/org.jkiss.dbeaver.ui.ai/src/org/jkiss/dbeaver/ui/ai/engine/copilot/CopilotConfigurator.java
```diff
@@ -250,7 +257,14 @@ private String acquireAccessToken(@NotNull DBRProgressMonitor monitor, @NotNull
...
-        saveSettings(copilotPropertiesCopy);
+        saveSettings((PROPERTIES) copilotPropertiesCopy);
         return Optional.of(copilotPropertiesCopy);
     }
+
+    protected <E extends CopilotCompletionEngine> E createEngine() {
+        CopilotProperties properties = new CopilotProperties();
+        properties.setToken(accessToken);
+
+        return (E) new CopilotCompletionEngine(properties);
```
ShadelessFox: Redundant generic bound

```suggestion
    @NotNull
    protected CopilotCompletionEngine E createEngine() {
        CopilotProperties properties = new CopilotProperties();
        properties.setToken(accessToken);

        return new CopilotCompletionEngine(properties);
```

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AIModelFeature.java
```diff
@@ -19,5 +19,5 @@
 public enum AIModelFeature {
     CHAT,
     STREAMING,
-    SPEECH_TO_TEXT,
+    SPEECH_TO_TEXT, EMBEDDING,
```
ShadelessFox:
```suggestion
    SPEECH_TO_TEXT,
    EMBEDDING
```

ShadelessFox (commented):
Missing nullability annotations and nitpicks

plugins/org.jkiss.dbeaver.model.ai/plugin.xml
```diff
@@ -19,7 +19,8 @@
             label="Github Copilot"
             class="org.jkiss.dbeaver.model.ai.engine.copilot.CopilotCompletionEngine"
             properties="org.jkiss.dbeaver.model.ai.engine.copilot.CopilotProperties"
-            supportsFunctions="true"/>
+            supportsFunctions="true"
+            fallbacks="copilot-pro"/>
```
serge-rider, line 23: Why do we need it? Also it is not fallback but `fallup`?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.ai/plugin.xml
ShadelessFox, line 23: Yeah... this seems weird. CE should not know about EE but the other way around. We already use `replaces` in multiple places. Can `copilot-pro` specify that it `replaces="copilot"`?

ivgag: It would be great. But when the OpenAI Pro engine was introduced, its config was placed under the openai-pro name instead of openai from the CE version. Then we faced a situation where we had to support the transition from the EE version to the CE version, so now CE knows about EE.
This has to be refactored, but not here.

PR #39605, opened 2025-11-07
dbeaver/dbeaver#39595: Make URL input field readonly copyable instead…

### LINKED ISSUE #39595: Enable URL input field to copy string
It is annoying to type the DB connection string in other applications when I already have the connection defined in DBeaver and could easily copy it from there.

Make the input field selectable so that the user can select and copy the string and paste it into any other application.

Image

Put a button next to the input field to copy the content to the clipboard.

**Additional context**
n/a

ged-yuko:
Closes https://github.com/dbeaver/dbeaver/issues/39595 by making URL input field readonly copyable in Host mode for Postgresql connection configuration

изображение


IvanGorshechnikov:
I don't like this solution because it is specific to Postgres. This issue affects all database connection pages. Could you make your solution less driver-specific?

ged-yuko:
@IvanGorshechnikov 
> I don't like this solution because it is specific to Postgres. This issue affects all database connection pages. Could you make your solution less driver-specific?

The logic of the mode switching is already resided in the ConnectionPageAbstract common super class. But, as far as i understand, the Host-vs-URL configuration mode switching supported in a database-specific basis across very few database (MySQL, SQLServer, GaussDB, Kingbase, Postgresql, DB2 and Generic here in CE). Each of the corresponding connection page implementations has its own definition of the urlText field, which needs to be configured. 
As such, please, note that the logic of the solution is not driver-specific as well already.

Would you like me to apply required configuration of the urlText field for all the other connection pages containg such a field?

And also there are DBeaver's non-community editions, i don't know if any connection pages there also have urlText field.

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageAbstract.java
```diff
@@ -445,9 +446,10 @@ private void updateConnectionModeControlsVisibility(@NotNull Collection<String>
             Set<Control> controls = propGroupMap.get(groupName);
             if (controls != null) {
                 for (Control control : controls) {
-                    control.setEnabled(enable);
-                    if (control instanceof Text text) {
+                    if (control instanceof Text text && Boolean.TRUE.equals(control.getData(CG_DISABLE_AS_READONLY))) {
```
serge-rider, line 449: As for me just marking it as non-editable is confusing.
URL and host fields look identical in dark theme. So it is not clear that URL is not supposed to be edited.

Also doing it just for PostgreSQL doesn't look like a good idea.

So a new idea needed here. E.g. adding icon "copy" to the right of text field.

But maybe it all doesn't make sense. After all it is just one extra click to switch to URL mode.

serge-rider (changes requested):

arhayka:
It’s been a while since the last update here. Unfortunately, we can’t merge this PR in its current form (please see @serge-rider's most recent comment for details).

If you’re able to continue working on it, feel free to reopen the PR.

PR #39614, opened 2025-11-10
dbeaver/pro#7248 Patch `java.io.tmpdir` if it can't be parsed

ShadelessFox:
Can be reproduced by following the instructions in the ticket.


Matvey16:
Still reproducible

PR #39615, opened 2025-11-10
dbeaver/pro#7429 Remove redundant databricks driver declaration

serge-rider:


Matvey16:
There is only one Databricks driver now
image

Matvey16:
Verified

PR #39626, opened 2025-11-11
dbeaver/dbeaver#39304 filter by owner

### LINKED ISSUE #39304: Incorrect reading of stored procedure code when the same name is in 2 different schemas
### Description

The test is in SAP IQ 16.1 using the JDBC driver.
When the same stored procedure name is in two different schemas, but they have different code, the code from the other schema appears instead of each schema with its corresponding code.
This bug occurs in versions 24.x and 25.x, but I have seen that it does not occur in versions 21.x and 23.x.

### DBeaver Version

24.x and 25.x

### Operating System

Windows 11

### Database and driver

Database: SAP IQ 16.1
Driver: JDBC jConnect driver (class com.sybase.jdbc4.jdbc.SybDriver)

### Steps to reproduce

View the code of one procedure in one schema and then the other procedure in another schema with the same name, but different code.

### Additional context

_No response_

Destrolaric:
Closes dbeaver/dbeaver#39304
Adds filter by owner


ShadelessFox (approved):

Matvey16:
Verified

PR #39629, opened 2025-11-11
Dbeaver/pro#7221 increase scope of dangerous drop queries

SolovevAnton:
closes dbeaver/pro#7221
_for testes_: please, be sure to test on MAC OS.

now will ask confirmation for any drop or delete/update without set query (dangerous queries).
in case of multiple queries will ask for each one continuously. 
If even one answer is "NO" will abort whole script execution.
in case both types present DROP and dangerous update will ask separately for both.

DROP statement recognition is only supported for:
-    "FUNCTION"
-     "INDEX"
-     "MATERIALIZED"
-     "SCHEMA"
-     "SEQUENCE"
-     "TABLE"
-     "TEMPORARY"
-     "VIEW"

**Dont ask me again checkbox**
If pressed cancel with "Dont ask me again" nothing happens, since dialog is considered cancelled.
If checked with OK or YES_TO_ALL button - will never ask user confirmation again

**Yes to all button:**
Button should not be present for single query execution
Yes to all button works separately for DROP and dangerous Update/Delete queries. 
For example if on drop queries user clicks 'yes to all" it will separately ask about dangerous update/delete queries and vise versa.

image

image


IvanGorshechnikov (dismissed):

Matvey16:
Now it is possible for the user to get into some kind of limbo, where they will try to execute a query but nothing will happen (until they specifically go to settings and turn on the confirmation again)

It looks like a very bad UX, especially since we always had the bahaviour where clicking `Cancel` that is now changed to `No` ignored the `Don't show again` checkbox. And it almost always works like this in other applications

STR
1. Execute an update without a where clause
2. Check `don't ask again` and click no
3. Try executing the query again

https://github.com/user-attachments/assets/84be0a8e-e919-403a-8937-807d6eba692d

IvanGorshechnikov (approved):

Destrolaric (approved):

Matvey16:
Verified

PR #39631, opened 2025-11-11
dbeaver/pro#7447 Use inline file name editor in import wizard

serge-rider:


DariaMarkaryan:
The Next button is not active if file name was inserted as text

https://github.com/user-attachments/assets/618b621d-bfc5-4814-bcbc-ebdb7eaca01e

serge-rider:
Fixed

DariaMarkaryan:
verified

PR #39635, opened 2025-11-11
dbeaver/pro#7443 Escape HTML entities from the input; tidy up the code

ShadelessFox:
See DB2 `gis.naumov.crime_scene`


plugins/org.jkiss.dbeaver.data.gis.view/src/org/jkiss/dbeaver/ui/gis/panel/GISLeafletViewer.java
```diff
@@ -96,7 +94,8 @@ public class GISLeafletViewer implements IGeometryValueEditor, DBPPreferenceList
     private volatile boolean browserCreating = false;
 
     private static final Gson gson = new GsonBuilder()
-            .registerTypeHierarchyAdapter(DBDContent.class, new DBDContentAdapter()).create();
+        .disableHtmlEscaping()
```
ShadelessFox, line 97: Just to avoid possible questions: it's redundant as we already escape HTML entities further below

uslss:
verified

PR #39642, opened 2025-11-12
Dbeaver/pro#7451 fix include not from workspace

SolovevAnton:
closes dbeaver/pro#7451: fix now scripts from outside the workspace works

Also now skip and skip all words correctly.
HOWEVER if skip or skip all is pressed script with error gets closed even if there was an error in it. related to: dbeaver/pro#7057


ShadelessFox (approved):

Matvey16:
Verified

PR #39671, opened 2025-11-16
dbeaver/pro# Core.net deps refactoring

serge-rider:


DariaMarkaryan:
Verified with auto tests

PR #39676, opened 2025-11-16
Dbeaver/pro#7442 http long polling fallback for ws event

vkirst:


plugins/org.jkiss.dbeaver.model.event/src/org/jkiss/dbeaver/model/websocket/WSConstants.java
```diff
@@ -21,6 +21,7 @@
  */
 public interface WSConstants {
     String WS_AUTH_HEADER = "SM-Auth-Token";
+    String WS_SESSION_HEADER = "X-CB-Session";
```
serge-rider: X-CB-Session -> X-LP-Session

serge-rider (changes requested):

alexander-skoblikov (approved):

PR #39690, opened 2025-11-17
#39344 Fix execution time formatting

### LINKED ISSUE #39344: Cannot see the query time in milliseconds
### Description

The query time is shown as for example 0.3 s instead of 0.321 s both underneath the query results and in the query manager

Image

Image

### DBeaver Version

Community Edition 25.2.2

### Operating System

ArchLinux

### Database and driver

PostgreSQL version 16.10.0
PostgreSQL JDBC 42.7.2

### Steps to reproduce

_No response_

### Additional context

_No response_

ShadelessFox:
Now duration that is less than 10 seconds should be formatted **without** trailing zeroes, like `1.2s`, `1.23s`


E1izabeth (dismissed):

uslss:
tasks that lasts less than 0,1s are shown as 0,0s

image

uslss:
verified
image

PR #39691, opened 2025-11-17
dbeaver/pro#6837 Refactor DataSourceUtils to simplify address text retrieval logic

labanovichttps:
image

### DIFF
--- a/plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceUtils.java
+++ b/plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceUtils.java
@@ -416,11 +416,7 @@ public class DataSourceUtils {
                 return hostText;
             }
 
-            if (CommonUtils.isNotEmpty(serverName)) {
-                return serverName;
-            }
-
-            return "";
+            return CommonUtils.notEmpty(serverName);
         } else {
             return cfg.getUrl();
         }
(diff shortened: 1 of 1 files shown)


Matvey16:
Verified

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceUtils.java
```diff
@@ -405,10 +408,19 @@ public static String getDataSourceAddressText(DBPDataSourceContainer dataSourceC
...
-            return hostText;
+            if (CommonUtils.isNotEmpty(hostText)) {
+                return hostText;
+            }
+
+            if (CommonUtils.isNotEmpty(serverName)) {
+                return serverName;
+            }
+
+            return "";
```
serge-rider: Behavior change - it was null before

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceUtils.java
labanovichttps: no, it returned empty string

labanovichttps: and method annotated as NotNull

PR #39694, opened 2025-11-17
dbeaver/dbeaver#37093 Decrease a driver version to one without depend…

### LINKED ISSUE #37093: H2GIS driver duplicate dependency version issue (JTS 1.19 and 1.20)
### Description

When using h2gis drivers, DBeaver download two versions of the same library JTS 1.19 and 1.20 that lead to method and class not found exception when doing jts stuff like Envelope computation.

Image

Is there a method to fix this dependency ? 1.19 should not be used by this version of h2gis.

@ebocher 

### DBeaver Version

24.3.4.202502021521

### Operating System

Ubuntu 24.10

### Database and driver

H2GIS 2.2.3

### Steps to reproduce

Create a H2GIS embedded connection;
download h2GIS drivers
look at dependency

### Additional context

_No response_

IvanGorshechnikov:
…ency problems

The issue in the driver was resolved in https://github.com/orbisgis/h2gis/pull/1407, and the driver should work correctly in the next version


E1izabeth (approved):

uslss:
verified

PR #39714, opened 2025-11-18
#37220 Add parentheses to the RHS expression

### LINKED ISSUE #37220: When filtering a result set with additional conditions, the conditions are not apply only to the result set but are instead added as conditions to the initial query without parentheses
### Description

**Description**

When filtering a result set with additional conditions, the conditions are not apply only to the result set but are instead added as conditions to the initial query without parentheses.

**Actual Result**
The additional conditions are added to the initial query without parentheses:

Image

**Expected Result**
The additional conditions are added to the initial query with parentheses.

### DBeaver Version

Community Edition 24.3.4.202502021521

### Operating System

Windows 10 Enterprise 22H2

### Database and driver

SQLite 3.48.0.0

### Steps to reproduce

1. Open DBeaver
2. Run a query similar to the one seen in the example above with conditions
3. Apply additional filters to the result set negating conditions in the initial query
4. Observe the result

### Additional context

_No response_

ShadelessFox:


IvanGorshechnikov (approved):

Matvey16:
Verified

PR #39731, opened 2025-11-19
dbeaver/pro#6925 Don't persist icons for custom drivers

ShadelessFox:
It's, well, pointless. We don't even offer a way to change the driver's icon. Originally introduced in #14966


IvanGorshechnikov (approved):

serge-rider:
I don't get it
I thought we are talking about icons for views/editors, not for drivers.

And do we really need to remove custom drivers icons after all?

ShadelessFox:
> I thought we are talking about icons for views/editors, not for drivers.

No, see https://github.com/dbeaver/pro/issues/6923 (this ticket refers to it). The issue you are talking about was fixed some time ago.

> And do we really need to remove custom drivers icons after all?

Yes, it seems completely pointless and doesn't do anything.

labanovichttps (approved):

PR #39748, opened 2025-11-20
#39344 Refactor duration formatting

### LINKED ISSUE #39344: Cannot see the query time in milliseconds

ShadelessFox:
Now progress reports just one digit of hundreds of milliseconds:

image

But in other places like Query Manager, up to three digits are shown:

image


uslss:
verified

PR #39757, opened 2025-11-21
dbeaver/pro#7482 additional statistics parameters

Destrolaric:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/utils/AIUtils.java
```diff
@@ -100,6 +100,17 @@ public static boolean isEligible(@Nullable DBPObject dbpObject) {
             || dbpObject instanceof DBSEntityConstraint;
     }
 
+    /**
+     * Retrieves all AI manager settings as a map of properties.
+     * @return a map containing all AI manager settings
+     */
+    public static Map<String, Object> getAIManagerSettings() {
```
serge-rider: No need

serge-rider (changes requested):

PR #39773, opened 2025-11-24
dbeaver/pro#7587 remove [] from cli output

alexander-skoblikov:
closes https://github.com/dbeaver/pro/issues/7587


plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/DBeaverLauncher.java
```diff
@@ -793,6 +793,9 @@ private CommandLineExecuteResult processCommandLineAsClient(String[] args, Path
             }
 
             if (output != null && !output.isEmpty()) {
+                if (output.startsWith("[") && output.endsWith("]")) {
```
LonwoLonwo, line 796: Feels like a good place for a unit test :)

ShadelessFox (approved):
We definitely in need of a refactoring that would bring a proper JSON parser instead of all this mess.

PR #39785, opened 2025-11-25
dbeaver/pro#7612 Fix metadata prefs not saving against a connection

ShadelessFox:
This only affects embedded databases, like SQLite. Doesn't seem like a regression.


IvanGorshechnikov (approved):

Matvey16:
Verfiied

PR #39794, opened 2025-11-26
dbeaver/pro#7599 Fix driver saving config

IvanGorshechnikov:


labanovichttps (dismissed):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/driver/DriverLoaderDescriptor.java
```diff
@@ -410,11 +410,6 @@ public boolean downloadDriverLibraries(@NotNull DBRProgressMonitor monitor, bool
             resolvedFiles.put(node.library, info);
             collectLibraryFiles(node, info);
         }
-        try {
-            driver.getProviderDescriptor().getRegistry().saveDrivers();
```
serge-rider, line 414: Why?

serge-rider (dismissed):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/driver/DriverLoaderDescriptor.java
IvanGorshechnikov: We shouldn't save driver after download his libraries. We shouldn't cache it.

IvanGorshechnikov: Will fixed in https://github.com/dbeaver/pro/issues/7637

ShadelessFox (approved):

PR #39799, opened 2025-11-26
dbeaver/pro#6123 add unauthorize request loging

HocKu7:
Closes https://github.com/dbeaver/pro/issues/6123


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/qm/QMConstants.java
```diff
@@ -25,6 +25,8 @@ public class QMConstants {
     public static final String QM_USERNAME_ATTR = "qm_username_id";
     public static final String QM_ANONYMOUS_DOMAIN = "anonymous";
     public static final String QM_ANONYMOUS_USER = "@anonymous@";
+    public static final String QM_UNAUTHORIZE_DOMAIN = "unauthorize";
```
Nexus6v2: I think `unauthorized` name would be better, both for user and domain

yagudin10 (approved):

alexander-skoblikov (approved):

PR #39807, opened 2025-11-27
dbeaver/pro#7176: fix DBUnit '"' escape to  &quot;

SolovevAnton:
closes dbeaver/pro#7176: now " inside cell values are replaced with &quote;

_for testers_: escaped and unescaped quotes.
for tables in Postgres only with **arrays** like: `(ARRAY['"special"','offer']);`
Quotes will be escaped by default. So the transfer to the DBUnit will be: `<ARRAYS ID="3" TAGS="{"\"special\"",offer}"/>`, however, it should still be parsed with no errors in the browser.
This is expected behavior. 
To change this and to get: `<ARRAYS ID="3" TAGS="{"special",offer}"/>` you shold change in data tranfer dialog Format Settings -> Advenced settings -> Value display format -> Database Native


IvanGorshechnikov (dismissed):

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/impl/AIEngineRequestFactory.java
```diff
@@ -141,7 +141,7 @@ protected void determineRequestTools(
         @NotNull AIPromptGenerator systemPromptGenerator,
         @NotNull AIEngineRequest request
     ) {
-        if (!engineDescriptor.isSupportsFunctions()) {
+        if (!engineDescriptor.isSupportsFunctions() || DBWorkbench.getPlatform().getApplication().isMultiuser()) {
```
ShadelessFox, line 144: Feels like this change doesn't belong to this PR

ShadelessFox (approved):

IvanGorshechnikov (approved):

PR #39838, opened 2025-12-01
dbeaver/pro#7523 Rework AI settings initialization

ivgag:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/registry/AISettingsInitializerDescriptor.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+package org.jkiss.dbeaver.model.ai.registry;
+
+import org.eclipse.core.runtime.IConfigurationElement;
+import org.jkiss.dbeaver.DBException;
+import org.jkiss.dbeaver.model.ai.internal.AISettingsInitializer;
+import org.jkiss.dbeaver.model.impl.AbstractDescriptor;
+import org.jkiss.dbeaver.registry.RegistryConstants;
+
+public class AISettingsInitializerDescriptor extends AbstractDescriptor {
```
serge-rider: Do we really need new registry? If we need just to differ CE/PRO then we can use AIAssistantRegistry/Descriptor

serge-rider (changes requested):

PR #39853, opened 2025-12-02
dbeaver/pro#7569 Refactor ordering data

IvanGorshechnikov:
Now, ordering mode and default ordering should apply in the "Data" page for the DB object. And it shouldn't work for custom queries


labanovichttps (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/order/OrderingUtils.java
```diff
@@ -0,0 +1,97 @@
...
+        if (!entity.getDataSource().getSQLDialect().supportsOrderByIndex()) {
+            return;
+        }
+        OrderingPolicy policy = OrderingPolicy.get(preferences);
+        if (policy == OrderingPolicy.DEFAULT) {
+            return;
+        }
+
+        try {
+            List<? extends DBSEntityAttribute> attrs = DBUtils.getBestTableIdentifier(new VoidProgressMonitor(), entity);
```
ShadelessFox: Pass a proper monitor instead

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/order/OrderingUtils.java
```diff
@@ -0,0 +1,97 @@
...
+            if (attrs.isEmpty()) {
+                return;
+            }
+            dataFilter.setOrder(String.join(",", attrs.stream().map(DBSEntityAttribute::getName).toList()) + " " + (
+                policy == OrderingPolicy.PRIMARY_KEY_DESC ? "DESC" : "ASC"));
+        } catch (Exception ignored) {
```
ShadelessFox: This try-catch block is only required for the `DBUtils.getBestTableIdentifier` call. Wrap just it

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -4090,29 +4090,27 @@ public void refresh()
...
-            for (DBDAttributeBinding binding : rowIdentifier.getAttributes()) {
-                DBDAttributeConstraint constraint = dataFilter.getConstraint(binding);
-                if (constraint != null) {
-                    constraint.setOrderPosition(dataFilter.getMaxOrderingPosition() + 1);
-                    constraint.setOrderDescending(policy == OrderingPolicy.PRIMARY_KEY_DESC);
-                }
-            }
+        final DBSDataContainer dataContainer = getDataContainer();
+        final DBDRowIdentifier rowIdentifier = model.getDefaultRowIdentifier();
+        final DBDDataFilter dataFilter = getDataFilter();
```
ShadelessFox:
```suggestion
        DBSDataContainer dataContainer = getDataContainer();
        DBDRowIdentifier rowIdentifier = model.getDefaultRowIdentifier();
        DBDDataFilter dataFilter = getDataFilter();
```

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/internal/DataEditorsPreferencesInitializer.java
```diff
@@ -56,8 +57,7 @@ public void initializeDefaultPreferences() {
         PrefUtils.setDefaultPreferenceValue(store, ResultSetPreferences.RESULT_SET_AUTOMATIC_ROW_COUNT, false);
         PrefUtils.setDefaultPreferenceValue(store, ResultSetPreferences.RESULT_SET_CANCEL_TIMEOUT, 5000);
         PrefUtils.setDefaultPreferenceValue(store, ResultSetPreferences.RESULT_SET_BINARY_EDITOR_TYPE, IValueController.EditType.EDITOR);
-        PrefUtils.setDefaultPreferenceValue(store, ResultSetPreferences.RESULT_SET_ORDERING_STRATEGY, ResultSetUtils.OrderingStrategy.SMART);
-        PrefUtils.setDefaultPreferenceValue(store, ResultSetPreferences.RESULT_SET_ORDERING_POLICY, ResultSetUtils.OrderingPolicy.DEFAULT);
+        PrefUtils.setDefaultPreferenceValue(store, ModelPreferences.RESULT_SET_ORDERING_STRATEGY, OrderingStrategy.SMART);
```
ShadelessFox: Why did you remove the other one?

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTable.java
```diff
@@ -176,6 +179,18 @@ public DBCStatistics readData(
         } catch (DBException e) {
             throw new DBCException("Can't generate query conditions", e, session.getExecutionContext());
         }
+
+        if (dataFilter != null && !dataFilter.hasOrdering()) {
```
ShadelessFox, line 184: I really dislike that a model object's behavior now depends on user preferences.

I'm pretty sure that the original idea was to extract relevant model parts (enum) and preferences so they can be reused in CloudBeaver. Folks over there have their own mechanism for query execution alongside ordering, so it's likely you don't need _this_ code to be here but instead in `ResultSetViewer`, as before.

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTable.java
```diff
@@ -176,6 +179,18 @@ public DBCStatistics readData(
...
+            final DBPPreferenceStore prefs = session.getDataSource().getContainer().getPreferenceStore();
+            final OrderingStrategy strategy = OrderingStrategy.get(prefs);
```
ShadelessFox:
```suggestion
            DBPPreferenceStore prefs = session.getDataSource().getContainer().getPreferenceStore();
            OrderingStrategy strategy = OrderingStrategy.get(prefs);
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/order/OrderingUtils.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public static boolean addOrderingOnClientSide(
+        @Nullable DBSDataContainer dataContainer,
+        @NotNull DBDDataFilter dataFilter,
+        @Nullable DBDRowIdentifier rowIdentifier,
+        @NotNull DBPPreferenceStore preferences
+    ) {
+        if (!(dataContainer instanceof DBSEntity)) {
+            return false;
+        }
```
ShadelessFox: What's this check for, again?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/order/OrderingUtils.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private OrderingUtils() {
+        // no instances
+    }
+
+    public static boolean addOrderingOnClientSide(
+        @Nullable DBSDataContainer dataContainer,
+        @NotNull DBDDataFilter dataFilter,
+        @Nullable DBDRowIdentifier rowIdentifier,
+        @NotNull DBPPreferenceStore preferences
```
ShadelessFox: I think you should provide an overload that accepts `OrderingPolicy`, avoiding even touching a preference store. Same for the other method.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/internal/DataEditorsPreferencesInitializer.java
IvanGorshechnikov: Move it to ModelPreferences

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/struct/JDBCTable.java
Nexus6v2:
```
I agree here, in current state it makes little sense to move preferenceStore-dependent code here, you should just leave it [in ResultSetViewer](https://github.com/dbeaver/dbeaver/pull/39853/files#diff-610159c9ab578295501a5d9cf3356a8ccec6f7f5c2dfe2e57435829dd56bfe38R4097).
But we need to expose the way to apply default filters from this class as well.
```

UPD: All good, this is needed to work in CB, and is fine

Nexus6v2 (approved):

uslss:
Default ordering is still aplied on SQL Editor 
image

uslss:
verified

PR #39857, opened 2025-12-03
dbeaver/dbeaver#39663 Handle 'use all columns as key' preference in resultset.

### LINKED ISSUE #39663: Bad warning / error when editing row in table w/o PK
### Description

Not sure if this is a configuration issue, usage issue, or bug.

I always get an error similar to "Column 'COL1' is read-only: No corresponding table column" when I try to edit a row in a table without a primary key.  

I actually **AM** able to edit data, because (I think) I have the "Use all table columns as a key upon saving" property checked on in the Data Editor / Behavior on missing key section.

The message probably shouldn't be shown to the user, because the table can be edited.

This is the table I am trying to edit:
`CREATE TABLE devdmgwrk/test1 (
col1 char(10),
col2 char(10),
col3 char(10)
)`

Here is an example of the message:
Image

This is the Data Editor / Behavior on missing key preference:
Image

### DBeaver Version

CE 25.2.4

### Operating System

Windows 11

### Database and driver

Database: DB2 for IBM i 7.4
Driver:      JT400 21.0.6

### Steps to reproduce

1. Open table in navigator
2. Click on Data tab
3. Double click on cell with data
4. Observe "Column 'COL1' is read-only: No corresponding table column"

### Additional context

_No response_

labanovichttps:
STR in the ticket


IvanGorshechnikov (approved):

SolovevAnton (approved):

uslss:
verified

PR #39865, opened 2025-12-03
dbeaver/pro#3872: fix large json rendering

SolovevAnton:
closes dbeaver/pro#3872 fix performance issue with removeWhiteSpaces.
removed whiteSpaces cleaning for values larger then 250 chars

_For testers_: please also test for oracle JSON option


plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/data/JDBCContentChars.java
```diff
@@ -198,7 +201,10 @@ public int hashCode() {
 
     @Override
     public String getDisplayString(@NotNull DBDDisplayFormat format) {
-        return data;
+        return data == null ? null :
+            (format == DBDDisplayFormat.EDIT || data.length() > MAX_LENGTH_REMOVE_WHITESPACES
+                ? data
+                : CommonUtils.compactWhiteSpaces(data));
```
IvanGorshechnikov:
```suggestion
        return data == null || format == DBDDisplayFormat.EDIT || data.length() > MAX_LENGTH_REMOVE_WHITESPACES
                ? data
                : CommonUtils.compactWhiteSpaces(data);
```

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/data/JDBCContentChars.java
```diff
@@ -42,6 +42,9 @@
  */
 public class JDBCContentChars extends JDBCContentAbstract implements DBDContentStorage, DBDContentCached {
 
+    private static final int MAX_LENGTH_REMOVE_WHITESPACES = 251;
```
ShadelessFox: Why 251, specifically?

SolovevAnton: because tool tip creator is check for 250 symbols to add "..." after value is truncated.
if it is 250 "..." are never added. A bit hacky, i know, but seems to be the easiest way

SolovevAnton: ps, cannot add direct mentioning of tooltip truncation constant, since it is in module not bundled to this one

ShadelessFox: Now I see. Can you clarify this in a comment, please? And change the constant to `250 + 1` for clarity; it will be optimized anyway.

ShadelessFox (approved):

IvanGorshechnikov (approved):

SolovevAnton:
comparison. Works only for very large jsons (tested on 100+ mb):
- old: 
https://github.com/user-attachments/assets/b29e8086-57e1-4942-859f-f52f2e6d9670

- new

https://github.com/user-attachments/assets/a2af3afc-866b-4e3f-8960-f0858c548da6

PR #39878, opened 2025-12-04
dbeaver/pro#6926 Add page with internal connection parameters

ShadelessFox:
java_vxSrpte7TA

This page is completely **read-only**; it only allows copying values but not editing them. It doesn't particularly help to solve the customer's case, though.

Moreover, an additional field "Provider ID" has been added to the driver's description:

image


uslss:
verified

PR #39879, opened 2025-12-04
#39706 Pick the highest index for a new file name

### LINKED ISSUE #39706: New scripts should always have highest number
It can happen that I have about 100 SQL scripts open. Every now and then I clean up the mess and remove some scripts which leads to gaps in the numbering.
When I now create a new script, it seems to look for the lowest number and the re-uses that again (here for example Script-33.sql). This is super confusing because later I look at the list of scripts and it shows like this

Script-1.sql
Script-2.sql
...
Script-33.sql (this was added after the clean up)
...
Script-110.sql
Script-111.sql

but Script-111.sql was not the one I added last but actually Script-33.sql.

Newly added scripts should always have the highest number.

I know that you can also display the column 'Modified' and sort by that to achieve what I am asking for but since that column is hidden per default it is not clear to the user that they can do that.
Most users are probably not even aware that they can add that column and sort by it and will accept the script numbering logic as a nuisance.

**Additional context**

ShadelessFox:


IvanGorshechnikov (approved):

plugins/org.jkiss.dbeaver.model.rcp/src/org/jkiss/dbeaver/utils/ResourceUtils.java
```diff
@@ -90,8 +96,45 @@ public static void syncFile(DBRProgressMonitor monitor, IResource localFile) {
...
+
+        IResource[] members;
+        try {
+            members = folder.members();
+        } catch (CoreException ignored) {
+            return getUniqueFileFallback(folder, fileName, fileExt);
+        }
+
+        // Look for all matching files and grab the one that has the highest index
+        var pattern = Pattern.compile("^%s-(\\d+).%s".formatted(Pattern.quote(fileName), Pattern.quote(fileExt)));
```
labanovichttps, line 115: put Pattern.compile("^%s-(\\d+).%s" in the field, cuz compile is very expensive operation

ShadelessFox: How am I supposed to put it in a field if it's evaluated from two arguments supplied at runtime?

labanovichttps: my bad

labanovichttps (approved):

uslss:
verified

PR #39880, opened 2025-12-04
dbeaver/pro#7661 fix datasource loading

HocKu7:
Closes https://github.com/dbeaver/pro/issues/7661


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceProviderRegistry.java
```diff
@@ -122,14 +122,20 @@ private void loadExtensions(IExtensionRegistry registry) {
...
-                if (CommonUtils.equalObjects(p1, p2)) return 0;
-                if (p1 == null) return -1;
-                if (p2 == null) return 1;
-                return 0;
-            });
+            Arrays.sort(
+                extElements,
+                Comparator.comparingInt(e -> {
+                    boolean hasParent = !CommonUtils.isEmpty(e.getAttribute(RegistryConstants.ATTR_PARENT));
+                    if (!hasParent) {
```
ShadelessFox: Double negation. I also suggest extracting `e.getAttribute(RegistryConstants.ATTR_PARENT)` to a variable to avoid extra lookup

HocKu7: Double negation is reasonable here. It makes it easy to read.

ShadelessFox: I don't see how it's clearler than

```suggestion
                    String parent = e.getAttribute(RegistryConstants.ATTR_PARENT);
                    if (CommonUtils.isEmpty(parent)) {
```

serge-rider (approved):

PR #39900, opened 2025-12-08
Dbeaver/pro#7643 fix streaming

Destrolaric:
Closes Dbeaver/pro#7643


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIConstants.java
```diff
@@ -40,6 +40,7 @@ public class AIConstants {
     public static final String AI_CONFIRM_OTHER = "ai.confirmation.other";
     public static final String AI_JOIN_RULE = "ai.joinRule";
     public static final int DEFAULT_CONTEXT_WINDOW_SIZE = 16384;
+    public static final String AI_USE_STREAM_MODE = "ai.useStreamMode";
```
IvanGorshechnikov: Why did you move it here?

IvanGorshechnikov (changes requested):

labanovichttps (dismissed):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIConstants.java
ShadelessFox: I moved it to `AIConstantsPro` intentionally. It's not used anywhere here in CE.

Destrolaric: My bad, artifact from revert

ShadelessFox (approved):

labanovichttps (approved):

PR #39904, opened 2025-12-08
Dbeaver/pro#7618 lsp server api

Nexus6v2:


features/org.jkiss.dbeaver.runtime.feature/feature.xml
```diff
@@ -46,6 +46,7 @@
     <plugin id="org.jkiss.dbeaver.model.sql.jdbc" version="0.0.0" />
     <plugin id="org.jkiss.dbeaver.model.lsm" version="0.0.0" />
     <plugin id="org.jkiss.dbeaver.model.ai" version="0.0.0" />
+    <plugin id="org.jkiss.dbeaver.model.lsp" version="0.0.0" />
```
serge-rider: It is not a part of dbeaver runtime.

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/DBLTextDocumentService.java
```diff
@@ -0,0 +1,343 @@
...
+    private void initContext(@NotNull String documentUri) {
+        ContextAwareDocument document = documentCache.get(documentUri);
+        if (document == null) {
+            log.warn(String.format("Unknown document %s, Skipping context init", documentUri));
+            return;
+        }
+
+        String projectId = URIUtils.extractProjectId(documentUri);
+        String resourcePath = URIUtils.extractResourcePath(documentUri);
+        DBPProject project = DBWorkbench.getPlatform().getWorkspace().getProject(projectId);
```
serge-rider: This won't work in cloud products. Global workspace doesn't have any projects.
You need to get workspace from user session.

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/DBLTextDocumentService.java
```diff
@@ -0,0 +1,343 @@
...
+        String dataSourceId;
+        if (project == null) {
+            log.error(String.format("Project %s not found. Proceeding without data source", projectId));
+            dataSourceId = null;
+        } else {
+            dataSourceId = String.valueOf(project.getResourceProperty(resourcePath, PROP_CONTEXT_DEFAULT_DATASOURCE));
```
serge-rider: Why do you need any properties within a project? It doesn't seem right

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/DBLTextDocumentService.java
```diff
@@ -0,0 +1,343 @@
...
+
+        log.debug("Initialized context for text document " + document);
+    }
+
+    @Nullable
+    private DBPDataSource resolveDataSource(@NotNull String projectId, @Nullable String dataSourceId) {
+        if (dataSourceId == null) {
+            return null;
+        }
+        return DBWorkbench.getPlatform().getWorkspace().getProjects().stream()
```
serge-rider: There is function findDataSourceById

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/DBLTextDocumentService.java
```diff
@@ -0,0 +1,343 @@
...
+        }
+
+        DBPDataSource dataSource = resolveDataSource(projectId, dataSourceId);
+        if (dataSource == null) {
+            log.warn(String.format("Unknown document data source %s, using basic SQL dialect", dataSourceId));
+        }
+        document.setDataSource(dataSource);
+
+        try {
+            document.setExecutionContext(DBUtils.getOrOpenDefaultContext(dataSource, false));
```
serge-rider: Where this execution context is closed? It is a very expensive resource and it must be closed when it no longer needed.

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/context/ContextAwareDocument.java
```diff
@@ -0,0 +1,113 @@
...
+import org.jkiss.dbeaver.model.DBPDataSource;
+import org.jkiss.dbeaver.model.exec.DBCExecutionContext;
+import org.jkiss.dbeaver.model.impl.sql.BasicSQLDialect;
+import org.jkiss.dbeaver.model.sql.SQLSyntaxManager;
+import org.jkiss.dbeaver.model.sql.parser.SQLRuleManager;
+import org.jkiss.dbeaver.runtime.DBWorkbench;
+
+public class ContextAwareDocument extends TextDocumentItem {
+    @Nullable
+    private DBPDataSource dataSource;
```
serge-rider: No need to keep datasource separately. You can get it from execution context

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/DocumentURI.java
```diff
@@ -0,0 +1,46 @@
...
+
+    public static void validateUri(@NotNull String uri) {
+        if (!URI_PATTERN.matcher(uri).matches()) {
+            throw new IllegalArgumentException(
+                String.format("Invalid URI format: %s, expected lsp://{projectId}/{resourcePath}", uri)
+            );
+        }
+    }
+
+    @NotNull
```
serge-rider, line 46: Let's create class DocumentURI and keep all parts there. Use regex instead of substrings to extract parts.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLIdentifierDetector.java
```diff
@@ -1,6 +1,6 @@
 /*
  * DBeaver - Universal Database Manager
- * Copyright (C) 2010-2024 DBeaver Corp and others
+ * Copyright (C) 2010-2025 DBeaver Corp and others
```
serge-rider, line 3: Remove unrelated changes

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/DBLTextDocumentService.java
Nexus6v2: `getOrOpenDefaultContext` opens a new execution context only in case there is no default present.
In case if default context already exists, it is returned.
I do this lazy init because features like completion won't work without context at all.

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/DBLTextDocumentService.java
Nexus6v2: Good point
I've added the ability to pass the workspace from outside in case it's needed

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/DBLTextDocumentService.java
Nexus6v2: As far as I understood, the default data source is stored as a resource property of a project under the `default-datasource` key.
Am I wrong here?

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/DBLTextDocumentService.java
Nexus6v2: As far as I understood, opening a default context should be fine.
If that's not true, my suggestion is to skip execution context init f we don't have the default one. But that won't allow us to use completion.

DariaMarkaryan:
Verified with desktop auto tests

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/utils/TextUtils.java
```diff
@@ -0,0 +1,64 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.lsp.utils;
+
+import org.eclipse.lsp4j.Position;
+import org.jkiss.code.NotNull;
+
+public class TextUtils {
```
serge-rider: Rename to LSPUtils. We already have TextUtils for sql

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/ProjectResourceUtils.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.jkiss.code.NotNull;
+import org.jkiss.code.Nullable;
+import org.jkiss.dbeaver.model.app.DBPProject;
+
+public class ProjectResourceUtils {
+    public static final String PROP_CONTEXT_DEFAULT_DATASOURCE = "default-datasource"; //$NON-NLS-1$
+
+    @Nullable
+    public static String getResourceDataSourceId(
```
serge-rider: Move to DBFUtils.
Actually we don't need utils method. Just move constant to DBConstants.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.lsp/src/org/jkiss/dbeaver/model/lsp/utils/TextUtils.java
Nexus6v2: Done

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/ProjectResourceUtils.java
Nexus6v2: Removed method, moved constant to DBConstants

PR #39920, opened 2025-12-10
dbeaver/dbeaver#39845 eclipse plugin install mixin

### LINKED ISSUE #39845: dbeaverc fails to install plugin due to added nosplash argument since 25.2.4
### Description

I'm attempting to automate installing plugins from the command line as shown in the [documentation](https://dbeaver.com/docs/dbeaver/Command-Line/#install-eclipse-extensions) but since 25.2.4 this is now failing with the error 'Unknown option -nosplash' [1764623006444.log](https://github.com/user-attachments/files/23865005/1764623006444.log). I'm not passing this argument and it looks like it's being added at runtime by the [launcher](https://github.com/dbeaver/dbeaver/blob/7e50f5ad22667f8c20a71a742e4c5c2f79bdd0b3/plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/DBeaverLauncher.java#L648). I've tried both zip and installer versions and this error happens on both and I've also had the error on 2 different machines.

### DBeaver Version

Community Edition 25.3.0

### Operating System

Windows 11

### Database and driver

_No response_

### Steps to reproduce

1. Open a terminal in the Dbeaver install directory
2. Run `.\dbeaverc.exe -application org.eclipse.equinox.p2.director -repository https://wakatime.com/eclipse/update-site/ -repository https://download.eclipse.org/releases/2025-06/ -installIU com.wakatime.eclipse.feature.feature.group -followReferences`
3. Installation fails due to 'Unknown option -nosplash'

### Additional context

I attempted to replicate this on a version I built from source from tag 25.3.0 but that didn't add the nosplash argument so this issue may be with release versions

alexander-skoblikov:
closes dbeaver/dbeaver#39845


MashaKorax:
verified

PR #39925, opened 2025-12-11
Dbeaver/pro#7715 fix nullability get source statement

SolovevAnton:
close dbeaver/pro#7715: fix all NULL returns from getSourceStatement method

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/exec/JDBCResultSetImpl.java
+++ b/plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/exec/JDBCResultSetImpl.java
@@ -75,7 +75,6 @@ public class JDBCResultSetImpl extends AbstractResultSet<JDBCSession, JDBCStatem
     }
 
     @NotNull
-    @NotNull
     public static JDBCResultSet makeResultSet(
         @NotNull JDBCSession session,
         @Nullable JDBCStatement statement,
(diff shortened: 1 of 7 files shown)


plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/exec/JDBCResultSetImpl.java
```diff
@@ -83,6 +74,17 @@ protected JDBCResultSetImpl(
             JDBCTrace.dumpResultSetOpen(this.original);
         }
     }
+
+    public static JDBCResultSet makeResultSet(
```
ShadelessFox, line 78:
```suggestion
    @NotNull
    public static JDBCResultSet makeResultSet(
```

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/exec/JDBCResultSetImpl.java
```diff
@@ -83,6 +74,17 @@ protected JDBCResultSetImpl(
...
+        @NotNull JDBCSession session,
+        @Nullable JDBCStatement statement,
+        @NotNull ResultSet original,
+        boolean disableLogging
+    ) throws SQLException {
+        JDBCStatement notNullStatement = Objects.requireNonNullElse(statement, new JDBCFakeStatementImpl(session, null, disableLogging));
```
ShadelessFox: It will create a new instance of `JDBCFakeStatementImpl` every time even if `statement != null`. Consider using a plain `if` check or `Objects.requireNonNullElseGet`

ShadelessFox (approved):

IvanGorshechnikov (approved):

PR #39927, opened 2025-12-11
dbeaver/pro#6977: pause command show results

SolovevAnton:
closes dbeaver/pro#6977 : for now still brocken on included scripts

TESTING:
main script
```
DROP TABLE IF EXISTS example;
CREATE TABLE example (id INT, name VARCHAR(20));
INSERT INTO example (id, name) VALUES (1, 'a'), (2, 'b'), (3, 'c');

@include sub_script.sql

select 1;
```

sub_script.sql
```
SELECT * FROM example;
@pause 'Pause here until clickkkkking'
```


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/execute/SQLQueryJob.java
```diff
@@ -428,6 +428,9 @@ private boolean executeSingleQuery(
         }
         if (element instanceof SQLControlCommand controlCommand) {
             try {
+                if (controlCommand.getCommandId().equalsIgnoreCase("pause")) {
```
E1izabeth: Looks like an abstractions leak

labanovichttps: I've replaced it via .equals(), and tested with Oracle and postgres, both commandId = `pause`

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/execute/SQLQueryJob.java
```diff
@@ -429,6 +429,9 @@ private boolean executeSingleQuery(
         }
         if (element instanceof SQLControlCommand controlCommand) {
             try {
+                if (controlCommand.getCommandId().equals("pause")) {
+                    controlCommand.setData(listener);
```
ShadelessFox, line 439: Why is it required to set command's data only for the `pause` command? In fact, we also have a shorthand `pau`; you should respect it as well. If this behavior is required for commands that interrupt the process, then perhaps we should add a new flag to indicate whether a command _can interrupt_? Then, `accept` would also become one.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -5465,6 +5465,10 @@ public void onEndSqlJob(DBCSession session, SqlJobResult result) {
                 extListener.onEndSqlJob(session, result);
             }
         }
+
+        public void redrawEditor() {
```
ShadelessFox, line 5549: Method's name does not reflect its behavior

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/execute/SQLQueryJob.java
labanovichttps: Add

IvanGorshechnikov (dismissed):

uslss:
error is thrown on "@pause" in the included script

https://github.com/user-attachments/assets/01be654f-9656-496b-9e8a-34ffec511c06

serge-rider:
Cannot reproduce.
What SQL settings do you have?

uslss:
[Script-412.sql](https://github.com/user-attachments/files/24611514/Script-412.sql)
[sub_script.sql](https://github.com/user-attachments/files/24611523/sub_script.sql)
reproducible in CE only

PR #39934, opened 2025-12-12
feat: macOS add minimize window command and handler

devolw:
Added minimize window to the “COMMAND+M” key combination for macOS

https://github.com/user-attachments/assets/ee770787-debb-4869-bc09-0eaf34069afd


ShadelessFox (approved):
Thanks for the contribution! One question, though: I'm not a macOS user, so it's pretty surprising to me that it's not an OS-level hotkey. Is this true?

devolw:
> Thanks for the contribution! One question, though: I'm not a macOS user, so it's pretty surprising to me that it's not an OS-level hotkey. Is this true?

Hello! On macOS, the Cmd+M shortcut is a system-wide key for minimizing the active window, but in DBeaver it doesn’t work by default. I’m not sure of the exact reason, but I assume it’s related to Eclipse’s behavior.
To enable the expected behavior, I added a Cmd+M key binding in plugin.xml and a handler that calls shell.setMinimized(true). This change only affects Cocoa builds and does not impact other platforms.

ShadelessFox:
Judging by the declarations in the `org.eclipse.ui` plugin, it's supposed to **maximize** the active part (e.g., editor). But on macOS, the M1+M (Cmd+M) shortcut is _unbound_ and instead replaced by _M4+M_ (Ctrl+M). I wonder if the declaration in the middle somehow breaks the system-wide shortcut...

```xml
      <key
            commandId="org.eclipse.ui.window.maximizePart"
            sequence="M1+M"
            schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" />
      <!-- Cmd+M is used for Minimize Window in cocoa -->
      <key
            platform="cocoa"
            sequence="M1+M"
            schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" />
      <key
            commandId="org.eclipse.ui.window.maximizePart"
            platform="cocoa"
            sequence="M4+M"
            schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" />
```

See https://github.com/eclipse-platform/eclipse.platform.ui/blob/e15e9bf606114f83531019524f3dca21ed351476/bundles/org.eclipse.ui/plugin.xml#L311-L319

So I guess your proposed fix is the best that can be done to address this issue, besides, well, contributing to Eclipse.

devolw:
Good day!

Mapping ```XML <key
            commandId="org.eclipse.ui.window.maximizePart"
            sequence="M1+M"
            schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" />```
&
Mapping ```XML <key 
            commandId="org.eclipse.ui.window.maximizePart"
            platform="cocoa"
            sequence="M4+M"
            schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" />```      
maximizePart expands the selected UI part (for example, an editor or a view) to the entire Eclipse workspace, temporarily hiding the other UI parts.
M1 + M — Linux/Windows;
M4 + M — macOS.

Mapping ```XML <key
            platform="cocoa"
            sequence="M1+M"
            schemeId="org.eclipse.ui.defaultAcceleratorConfiguration" />```
No action is assigned here, but apparently the mere presence of this mapping causes the standard shortcut to be intercepted, which is why the system action “Minimize window” does not work in Eclipse.
Basically, to restore the standard Cmd + M behavior, you just need to remove these lines from plugin.xml.
Screenshot 2025-12-15 at 4 33 31 AM

I’ll look into fixing this issue in Eclipse in my spare time.

Could you tell me if I need to edit the commit messages, and if so, which ones?

ShadelessFox:
> I’ll look into fixing this issue in Eclipse in my spare time.

Good! Thank you.

> Could you tell me if I need to edit the commit messages, and if so, which ones?

Per our contribution guide, every PR must have an associated issue. Since this one doesn't have one, it's okay to keep messages as-is. If you are willing to create an issue and link it, you'll need to rename all commits to be in a form of `#<issue-id> <message>`.

See https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code#commit-message-rules

Matvey16:
Verified

PR #39956, opened 2025-12-16
dbeaver/pro#7627: fix count dirty pattern

SolovevAnton:
closes dbeaver/pro#7627: now after esc pressed all changes are considered gone


plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -545,10 +543,8 @@ void resetCellValue(@NotNull DBDAttributeBinding attr, @NotNull ResultSetRow row
                 log.error(e);
             }
             row.clearChange(attr);
-            if (row.getState() == ResultSetRow.STATE_NORMAL) {
-                changesCount--;
-            }
         }
+        changedRows.remove(row);
```
serge-rider: What if we had multiple changed cells for that row?

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -910,7 +913,10 @@ public boolean hasData() {
     }
 
     public boolean isDirty() {
-        return changesCount != 0;
+        return !changedRows.isEmpty()
```
serge-rider: isDirty is called frequently. And new new algorithm is quite expensive.

serge-rider (changes requested):

Matvey16:
Verfiied

PR #39957, opened 2025-12-16
dbeaver/pro#7069 Add preference to enable query identification comments

labanovichttps:
image

Just start data export
ALSO enable setting DBeaver Preferences —> Connection —> Client identification


IvanGorshechnikov (dismissed):

SolovevAnton (dismissed):

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/exec/JDBCStatementImpl.java
```diff
@@ -146,6 +149,18 @@ public void addToBatch() throws DBCException {
         }
     }
 
+    private String addQueryIdentificationComment(@NotNull String sql) {
```
serge-rider: I don't think that it should one THAT low level.
We have several places where we produce metadata queries (like cache classes).
Also it adds comments to JDBC queries only.
Also the comment must be not jsut product name but something like "Query automatically generated by XXX"

serge-rider (changes requested):

Matvey16:
It works only for two specific queries. A `select` during export, and `create table` during import. Comment is not added to anything else

dbeaver_bxpcp41Fr3

dbeaver_HowSIAer5k

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/exec/JDBCConnectionImpl.java
```diff
@@ -258,15 +259,19 @@ public JDBCStatement createStatement()
     public JDBCPreparedStatement prepareStatement(String sql)
         throws SQLException
     {
-        return createPreparedStatementImpl(() -> getOriginal().prepareStatement(sql), sql);
+        sql = SQLUtils.addQueryIdentificationComment(this, sql);
+        final String finalSql = sql;
```
ShadelessFox: Inline `sql`

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/exec/JDBCConnectionImpl.java
```diff
@@ -258,15 +259,19 @@ public JDBCStatement createStatement()
...
     }
 
     @NotNull
     @Override
     public JDBCCallableStatement prepareCall(String sql)
         throws SQLException
     {
-        return createCallableStatementImpl(() -> getOriginal().prepareCall(sql), sql);
+        sql = SQLUtils.addQueryIdentificationComment(this, sql);
+        final String finalSql = sql;
```
ShadelessFox: Inline `sql`

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -774,6 +774,24 @@ public static String generateCommentLine(@Nullable DBPDataSource dataSource, @No
...
+            return query;
+        }
+        var prefStore = session.getDataSource().getContainer().getPreferenceStore();
+        if (!prefStore.getBoolean(org.jkiss.dbeaver.ModelPreferences.META_ADD_AUTO_QUERY_IDENTIFICATION_COMMENT)) {
+            return query;
+        }
+        if (session.getPurpose().isUser()) {
+            return query;
+        }
+        return generateCommentLine(session.getDataSource(), "Query automatically generated by " + GeneralUtils.getProductName()) + query;
```
ShadelessFox: This should likely be localized.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -774,6 +774,24 @@ public static String generateCommentLine(@Nullable DBPDataSource dataSource, @No
         return sb.toString();
     }
 
+    /**
+     * Appends query identification comment if it is enabled in preferences.
+     */
+    @NotNull
+    public static String addQueryIdentificationComment(@NotNull DBCSession session, @NotNull String query) {
+        if (query.contains("Query automatically generated by")) {
```
ShadelessFox: This feels like a half measure. Is it ever triggered?

Destrolaric (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -774,12 +776,14 @@ public static String generateCommentLine(@Nullable DBPDataSource dataSource, @No
         return sb.toString();
     }
 
+    private static final String QUERY_ID_COMMENT_MARK = " (auto-gen)";
```
ShadelessFox: I mean, you could totally rely on the localized message itself, no need for extra marks.

labanovichttps: fixed

ShadelessFox (dismissed):

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/CoreResources.properties
```diff
@@ -410,6 +410,8 @@ pref_page_database_general_separate_meta_connection_tip = Opening separate metad
 pref_page_database_general_server_side_object_filters = Server side object filters
 pref_page_database_general_server_side_object_filters_tip = Supported only by some datasources.\nModify metadata queries so only necessary objects will be read from the database.\nOtherwise, filtering will be applied on the client side.\nDisable this option if you want to see linked objects (e.g. from foreign keys).
 pref_page_database_label_client_application_name = Client Application Name
+pref_page_database_label_add_auto_query_identification_comment = Add identification comment before auto-generated queries
+pref_page_database_label_add_auto_query_identification_comment_tip = When enabled, DBeaver add -- {product} before internal META/UTIL queries.
```
Matvey16:
```suggestion
pref_page_database_label_add_auto_query_identification_comment_tip = When enabled, DBeaver will add -- {product} at the beginning of internal META/UTIL queries.
```

labanovichttps: fixed

Matvey16:
Verified

DB EE Windows

PR #39966, opened 2025-12-18
Dbeaver/pro#7616 ai chat files

IvanGorshechnikov:


plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/EditorUtils.java
```diff
@@ -648,13 +653,53 @@ public static boolean openFileEditors(
...
                 openedFiles.add(path);
             } else {
                 DBWorkbench.getPlatformUI().showError("Open file", "Can't open '" + path + "': file doesn't exist");
             }
         }
+        return filesByHandler;
+    }
+
+    @NotNull
+    public static Map<FileTypeHandlerDescriptor.Extension, List<Path>> getExtensionFiles(
```
ShadelessFox, line 669: This is a pretty wild signature. So it just groups input files based on their compatible extension handler? You don't really use that "grouping" anywhere, so I suggest introducing a pair-like type that encapsulates both the `Extension` and `Path` and just return a list of those.

plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/EditorUtils.java
```diff
@@ -622,21 +622,26 @@ public static List<Path> openExternalFiles(@NotNull String[] fileNames, @Nullabl
...
+    ) {
         log.debug("Open external file(s) [" + Arrays.toString(filePaths) + "]");
         List<Path> openedFiles = new ArrayList<>();
-        return openFileEditors(filePaths, currentContainer, openedFiles);
+        return openFileEditors(filePaths, currentContainer, openedFiles, databaseOnly);
     }
 
-    public static boolean openFileEditors(
+    @NotNull
+    public static Map<FileTypeHandlerDescriptor, List<Path>> getHandlerFiles(
```
ShadelessFox, line 641: Same here. I think you can even use `Pair`, should be fine

plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/EditorUtils.java
IvanGorshechnikov: I use it for checking ability of create and import

ShadelessFox (approved):

PR #39969, opened 2025-12-19
Add New StarRocks Plugin for External Catalog Support

### LINKED ISSUE #35880: Catalog support for Starrocks
Starrocks in version 2.3 add support for multiple catalogs.
https://docs.starrocks.io/docs/data_source/catalog/catalog_overview/

I'd like to see them in tree view.

None available.

chris-celerdata:
## Description
Resolves #35880. 
image

This PR introduces a new StarRocks plugin that supports external catalogs. The previous MySQL compatible plugin has been left for backwards compatibility; it is now named `StarRocks (Legacy)`. 

This plugin took heavy inspiration from the existing MySQL plugin.

## Design Decisions
The `StarRocksDatabase` and `StarRocksBaseTable` are functionally similar to `MySQLCatalog` and `MySQLBaseTable` respectively, so similar approaches have been taken. The primary change is that `StarRocksDataSource` returns `StarRocksCatalog` as the children. It also defines MySQL as the parent for reuse of the connection UI. 

I looked at other connectors and considered tree injection in order to add the `StarRocksCatalog` to the mysql tree, but I was unable to get this to work. Another option I considered that would be simpler was extending `GenericDataSource`, but StarRocks does not have a JDBC driver, preferring to use MySQL's. This will not fix the external catalog problem. 

## Additional Notes
Are there any tests necessary to write for new plugins?
In order to test this plugin, a local StarRocks instance can be created [with Docker](https://docs.starrocks.io/docs/quick_start/shared-nothing/). Then, connect at `localhost:9030`.


chris-celerdata:
Closing this because I did not read the commit message guidelines before making this. I will fix and then make a new one. Sorry for the inconvenience.

PR #39973, opened 2025-12-19
dbeaver/pro#7843 fix copilot streaming

Destrolaric:
Closes dbeaver/pro#7843


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/copilot/CopilotClient.java
```diff
@@ -179,6 +179,10 @@ public void createChatCompletionStream(
             request,
             line -> {
                 if (line.startsWith(DATA_EVENT)) {
+                    if (line.substring(6).trim().equals(DONE_EVENT)) {
```
ShadelessFox: `startsWith`?

Destrolaric: startsWith was before, this is latter part

ShadelessFox: I see. Then move
```java
String data = line.substring(6).trim();
```
to be ABOVE this check. This is a redundant operation

Destrolaric: Fixed

ShadelessFox (approved):

IvanGorshechnikov (approved):

PR #39975, opened 2025-12-19
dbeaver/dbeaver#35880 Add catalog support for StarRocks

### LINKED ISSUE #35880: Catalog support for Starrocks

chris-celerdata:
## Description
Resolves #35880. 
image

This PR introduces a new StarRocks plugin that supports external catalogs. The previous MySQL compatible plugin has been left for backwards compatibility; it is now named `StarRocks (Legacy)`. 

This plugin took heavy inspiration from the existing MySQL plugin.

## Design Decisions
The `StarRocksDatabase` and `StarRocksBaseTable` are functionally similar to `MySQLCatalog` and `MySQLBaseTable` respectively, so similar approaches have been taken. The primary change is that `StarRocksDataSource` returns `StarRocksCatalog` as the children. It also defines MySQL as the parent for reuse of the connection UI. 

I looked at other connectors and considered tree injection in order to add the `StarRocksCatalog` to the mysql tree, but I was unable to get this to work. Another option I considered that would be simpler was extending `GenericDataSource`, but StarRocks does not have a JDBC driver, preferring to use MySQL's. This will not fix the external catalog problem. 

## Additional Notes
Are there any tests necessary to write for new plugins?
In order to test this plugin, a local StarRocks instance can be created [with Docker](https://docs.starrocks.io/docs/quick_start/shared-nothing/). Then, connect at `localhost:9030`.


arhayka:
@chris-celerdata Thank you for the contribution! I’ve passed this PR to the development team for review.

plugins/org.jkiss.dbeaver.ext.starrocks/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,16 @@
...
+Bundle-Name: DBeaver StarRocks Support
+Bundle-SymbolicName: org.jkiss.dbeaver.ext.starrocks;singleton:=true
+Bundle-Version: 1.0.0.qualifier
+Bundle-Activator: org.jkiss.dbeaver.ext.starrocks.internal.StarRocksActivator
+Bundle-Vendor: DBeaver Corp
+Require-Bundle: org.eclipse.core.runtime,
+ org.jkiss.dbeaver.model,
+ org.jkiss.dbeaver.model.jdbc,
+ org.jkiss.dbeaver.model.sql,
+ org.jkiss.dbeaver.ext.mysql;bundle-version="1.0.0"
```
Destrolaric: Avoiud specifying direct version

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksCatalog.java
```diff
@@ -0,0 +1,164 @@
...
+
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Collection;
+
+/**
+ * StarRocks Catalog - represents a catalog (e.g., default_catalog, hive_catalog)
+ * Contains multiple databases (schemas)
+ */
+public class StarRocksCatalog implements DBSCatalog, DBPRefreshableObject {
```
Destrolaric: Can we inherit MySQL metadata classes to avoid code duplication, or are the changes too significant?

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksTableColumn.java
```diff
@@ -0,0 +1,45 @@
...
+ * StarRocks Table Column - used for both tables and views.
+ */
+public class StarRocksTableColumn extends JDBCTableColumn<StarRocksTableBase> {
+
+    public StarRocksTableColumn(StarRocksTableBase table, ResultSet dbResult) throws DBException {
+        super(table, true);
+        // Parse SHOW FULL COLUMNS result: Field | Type | Collation | Null | Key | Default | Extra | Privileges | Comment
+        setName(JDBCUtils.safeGetString(dbResult, "Field"));
+        setTypeName(JDBCUtils.safeGetString(dbResult, "Type"));
+        String nullableStr = JDBCUtils.safeGetString(dbResult, "Null");
```
Destrolaric: It's preferable to use constants in the safeGet. Additionally, please add `//$NON-NLS-1$` annotation where applicable.

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksExecutionContext.java
```diff
@@ -0,0 +1,243 @@
...
+ */
+public class StarRocksExecutionContext extends JDBCExecutionContext
+        implements DBCExecutionContextDefaults<StarRocksCatalog, StarRocksDatabase> {
+
+    private static final Log log = Log.getLog(StarRocksExecutionContext.class);
+
+    private String activeCatalogName;
+    private String activeDatabaseName;
+
+    StarRocksExecutionContext(@NotNull JDBCRemoteInstance instance, String purpose) {
```
Destrolaric: If applicable, it's preferable to add `@NotNull` or `@Nullable` annotations for methods and parameters

Destrolaric (changes requested):
Hi, thanks for the contribution. Could you please look at those issues

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksCatalog.java
chris-celerdata: I tried this by having `StarRocksDataSource` extend `MySQLDataSource` but there were Java type system constraints. `StarRocksCatalog` must be the top-level container. Then I tried to have `StarRocksDatabase` extend `MySQLCatalog`. However, `MySQLDataSource` (and extending classes) must return `MySQLCatalog` from `getCatalog`, `getChildren`, and `getChild`. `StarRocksCatalog` cannot extend `MySQLCatalog` because they are fundamentally different concepts (containing databases and tables respectively).

Destrolaric: Could you then think about inheriting from GenericCatalog and `GenericDataSource`, which should result in less code duplication? You could rely on generics for other metadata classes.

Destrolaric:
Thanks for the changes. I am going to invite the second reviewer. It might take some time. Thank you the changes.

plugins/org.jkiss.dbeaver.ext.starrocks/icons/starrocks_icon.png
ShadelessFox, line 1: Please consider using SVGs instead. Please note that `@2x.svg` variations are not required in that case.

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/internal/StarRocksActivator.java
```diff
@@ -0,0 +1,28 @@
...
+
+import org.eclipse.core.runtime.Plugin;
+import org.osgi.framework.BundleContext;
+
+public class StarRocksActivator extends Plugin {
+
+    private static StarRocksActivator plugin;
+
+    public StarRocksActivator() {
+    }
```
ShadelessFox: Redundant public constructor

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/internal/StarRocksActivator.java
```diff
@@ -0,0 +1,28 @@
...
+        plugin = this;
+    }
+
+    @Override
+    public void stop(BundleContext context) throws Exception {
+        plugin = null;
+        super.stop(context);
+    }
+
+    public static StarRocksActivator getDefault() {
```
ShadelessFox, line 42:
```suggestion
    @NotNull
    public static StarRocksActivator getDefault() {
```
Perhaps it's also worth wrapping `plugin` within `Objects.requireNonNull` to ensure the plugin has started.

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksCatalog.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Property(viewable = true, order = 2)
+    public String getType() {
+        return type;
+    }
+
+    @Property(viewable = true, order = 3)
+    public String getComment() {
+        return comment;
+    }
```
ShadelessFox: Please add nullability annotations. Since fields are not initialized in the constructor, you must use `@Nullable`.

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksCatalog.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @Nullable
+    public StarRocksDatabase getDatabase(@NotNull DBRProgressMonitor monitor, @NotNull String name) throws DBException {
+        return (StarRocksDatabase) getSchema(monitor, name);
+    }
+
+    @Nullable
+    public StarRocksDatabase getCachedDatabase(@NotNull String name) {
+        try {
+            Collection<GenericSchema> schemas = getSchemas(null);
```
ShadelessFox: Never pass `null` where a non-null value is expected; this call will throw a NPE if schemas aren't cached yet.

If you want this method to return only cached results, consider passing `new LocalCacheProgressMonitor(new VoidProgressMonitor())`. Not ideal, but still better than a ticking time bomb about to explode.

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksCatalog.java
```diff
@@ -0,0 +1,100 @@
...
+            if (schemas != null) {
+                for (GenericSchema schema : schemas) {
+                    if (schema.getName().equals(name)) {
```
ShadelessFox: This whole logic can be replaced with `getSchema(monitor, name)`

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksCatalog.java
```diff
@@ -0,0 +1,100 @@
...
+                        return (StarRocksDatabase) schema;
+                    }
+                }
+            }
+        } catch (DBException e) {
+            // Return null if schemas haven't been loaded yet
```
ShadelessFox: Please don't suppress exceptions; at least log them

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksDataSource.java
```diff
@@ -0,0 +1,165 @@
...
+    private static final String COL_COMMENT = "Comment"; //$NON-NLS-1$
+
+    /**
+     * catalog name -> [type, comment]
+     */
+    private final Map<String, CatalogMetadata> catalogMetadataCache = new ConcurrentHashMap<>();
+
+    public static class CatalogMetadata {
+        public final String type;
+        public final String comment;
```
ShadelessFox: Don't expose fields. Consider using `record` instead

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksDataSource.java
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    @Override
+    protected void initializeContextState(
+        @NotNull DBRProgressMonitor monitor,
+        @NotNull JDBCExecutionContext context,
+        JDBCExecutionContext initFrom
+    ) throws DBException {
+        if (initFrom != null) {
+            StarRocksExecutionContext starRocksContext = (StarRocksExecutionContext) context;
```
ShadelessFox: This variable can be extracted and reused in the `else` block

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksExecutionContext.java
```diff
@@ -0,0 +1,256 @@
...
+    @Override
+    public boolean supportsSchemaChange() {
+        return true;
+    }
+
+    @Override
+    public void setDefaultCatalog(
+            @NotNull DBRProgressMonitor monitor,
+            @Nullable StarRocksCatalog catalog,
+            @Nullable StarRocksDatabase schema
```
ShadelessFox: We use 4-space indentation


```suggestion
        @NotNull DBRProgressMonitor monitor,
        @Nullable StarRocksCatalog catalog,
        @Nullable StarRocksDatabase schema
```

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksExecutionContext.java
```diff
@@ -0,0 +1,256 @@
...
+        boolean catalogMatch = newCatalogName.equals(activeCatalogName);
+        boolean schemaMatch = schema == null || schema.getName().equals(activeDatabaseName);
+        return catalogMatch && schemaMatch;
+    }
+
+    private void fireSelectionChangeEvents(
+            @Nullable StarRocksCatalog oldCatalog,
+            @NotNull StarRocksCatalog newCatalog,
+            @Nullable StarRocksDatabase oldSchema,
+            @Nullable StarRocksDatabase newSchema
```
ShadelessFox: Same as the previous comment

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksMetaModel.java
```diff
@@ -0,0 +1,287 @@
...
+
+    @Override
+    public GenericTableColumn fetchTableColumn(
+        @NotNull JDBCSession session,
+        @NotNull GenericStructContainer owner,
+        @NotNull GenericTableBase table,
+        @NotNull JDBCResultSet dbResult
+    ) throws DBException {
+        if (table instanceof StarRocksTableBase) {
+            return new StarRocksTableColumn((StarRocksTableBase) table, dbResult);
```
ShadelessFox:
```suggestion
        if (table instanceof StarRocksTableBase starRocksTable) {
            return new StarRocksTableColumn(starRocksTable, dbResult);
```

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksMetaModel.java
```diff
@@ -0,0 +1,287 @@
...
+        @NotNull Map<String, Object> options
+    ) throws DBException {
+        if (sourceObject instanceof StarRocksTable) {
+            return ((StarRocksTable) sourceObject).getObjectDefinitionText(monitor, options);
+        }
+        if (sourceObject instanceof StarRocksView) {
+            return ((StarRocksView) sourceObject).getObjectDefinitionText(monitor, options);
+        }
+        if (sourceObject instanceof StarRocksMaterializedView) {
+            return ((StarRocksMaterializedView) sourceObject).getObjectDefinitionText(monitor, options);
```
ShadelessFox: Same as the previous comment

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksTableBase.java
```diff
@@ -0,0 +1,101 @@
...
+        return (StarRocksDataSource) super.getDataSource();
+    }
+
+    @Nullable
+    @Override
+    public GenericCatalog getCatalog() {
+        // Get catalog from the container hierarchy
+        GenericStructContainer container = getContainer();
+        if (container instanceof GenericSchema) {
+            return ((GenericSchema) container).getCatalog();
```
ShadelessFox: Same as the previous comment

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksTableBase.java
```diff
@@ -0,0 +1,101 @@
...
+        return container.getCatalog();
+    }
+
+    @Nullable
+    @Override
+    public GenericSchema getSchema() {
+        // The container should be the schema (StarRocksDatabase)
+        GenericStructContainer container = getContainer();
+        if (container instanceof GenericSchema) {
+            return (GenericSchema) container;
```
ShadelessFox: Same as the previous comment

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksTableColumn.java
```diff
@@ -0,0 +1,178 @@
...
+
+    private static final String COL_FIELD = "Field"; //$NON-NLS-1$
+    private static final String COL_TYPE = "Type"; //$NON-NLS-1$
+    private static final String COL_NULL = "Null"; //$NON-NLS-1$
+    private static final String COL_COMMENT = "Comment"; //$NON-NLS-1$
+    private static final String COL_DEFAULT = "Default"; //$NON-NLS-1$
+
+    private static int ordinalCounter = 0;
+
+    public StarRocksTableColumn(StarRocksTableBase table) {
```
ShadelessFox:
```suggestion
    public StarRocksTableColumn(@NotNull StarRocksTableBase table) {
```

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksTableColumn.java
```diff
@@ -0,0 +1,178 @@
...
+ */
+public class StarRocksTableColumn extends GenericTableColumn {
+
+    private static final String COL_FIELD = "Field"; //$NON-NLS-1$
+    private static final String COL_TYPE = "Type"; //$NON-NLS-1$
+    private static final String COL_NULL = "Null"; //$NON-NLS-1$
+    private static final String COL_COMMENT = "Comment"; //$NON-NLS-1$
+    private static final String COL_DEFAULT = "Default"; //$NON-NLS-1$
+
+    private static int ordinalCounter = 0;
```
ShadelessFox: Don't do that. Not only will it introduce race conditions, but it will also result in the first ordinal of the second table's column being `> 0`. Instead, pass the ordinal as a parameter.

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksTableColumn.java
```diff
@@ -0,0 +1,178 @@
...
+        );
+    }
+
+    @NotNull
+    @Override
+    public StarRocksDataSource getDataSource() {
+        return (StarRocksDataSource) super.getDataSource();
+    }
+
+    private static int mapSqlType(String typeName) {
```
ShadelessFox:
```suggestion
    private static int mapSqlType(@Nullable String typeName) {
```

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksTableColumn.java
```diff
@@ -0,0 +1,178 @@
...
+            case "DATE" -> Types.DATE;
+            case "DATETIME", "TIMESTAMP" -> Types.TIMESTAMP;
+            case "TIME" -> Types.TIME;
+            case "ARRAY" -> Types.ARRAY;
+            case "MAP", "STRUCT" -> Types.STRUCT;
+            default -> Types.OTHER;
+        };
+    }
+
+    private static long extractColumnSize(String typeName) {
```
ShadelessFox:
```suggestion
    private static long extractColumnSize(@Nullable String typeName) {
```

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksTableColumn.java
```diff
@@ -0,0 +1,178 @@
...
+                return Long.parseLong(sizeStr.trim());
+            } catch (NumberFormatException e) {
+                return 0;
+            }
+        }
+        return 0;
+    }
+
+    @Nullable
+    private static Integer extractScale(String typeName) {
```
ShadelessFox:
```suggestion
    private static Integer extractScale(@Nullable String typeName) {
```

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksTableColumn.java
```diff
@@ -0,0 +1,178 @@
...
+        int parenStart = typeName.indexOf('(');
+        int parenEnd = typeName.indexOf(')');
+        if (parenStart > 0 && parenEnd > parenStart) {
+            String sizeStr = typeName.substring(parenStart + 1, parenEnd);
+            int commaIndex = sizeStr.indexOf(',');
+            if (commaIndex > 0) {
+                try {
+                    return Integer.parseInt(sizeStr.substring(commaIndex + 1).trim());
+                } catch (NumberFormatException e) {
+                    return null;
```
ShadelessFox, line 177: Consider logging the exception and the input type name for troubleshoot purposes

plugins/org.jkiss.dbeaver.ext.starrocks/plugin.xml
```diff
@@ -0,0 +1,87 @@
...
+                                    </items>
+                                </folder>
+                            </items>
+                        </folder>
+                    </items>
+                </folder>
+            </tree>
+
+            <drivers managable="true">
+                <driver
```
ShadelessFox, line 48: If the new driver is compatible with the (now) legacy one, consider marking the new one as the replacement of the former.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksCatalog.java
chris-celerdata: My understanding is that the `StarRocksExecutionContext.getDefaultSchema()` will be called by the UI thread, so this function should be non-blocking. So it should only return cached results. Is this correct?

plugins/org.jkiss.dbeaver.ext.starrocks/plugin.xml
chris-celerdata: I am not fully certain how to do this, is this line: `<replace provider="mysql" driver="starRocks"/>` what you are talking about?

plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/model/StarRocksCatalog.java
ShadelessFox: Your understanding is correct. But calling `getSchemas` with `null` monitor won't make it magically non-blocking. The implementation doesn't handle `null` monitors; the parameter itself is also marked as `@NotNull`. Under certain conditions, it will result in an NPE.

The `LocalCacheProgressMonitor` implementation of a progress monitor is special: it has a flag that tells its consumers to use only cached values and not perform any IO tasks or other expensive operations. But since it wraps another progress monitor, and you don't have one, you can pass the dumb `VoidProgressMonitor` implementation that does nothing.

plugins/org.jkiss.dbeaver.ext.starrocks/plugin.xml
ShadelessFox: That's correct.

chris-celerdata:
I have implemented these fixes. I had a question: is it possible through the tree structure to exclude folders based on a condition?
image
In this image, databases in external catalogs do not support materialized views, but they still appear as a folder. The default catalog's databases do support them and I do want them to appear in the sidebar.

ShadelessFox:
> is it possible through the tree structure to exclude folders based on a condition?

Yes. You can specify the `visibleIf` attribute on a `folder` element. See

https://github.com/dbeaver/dbeaver/blob/027bb9e036ab53160334a23969cc0f7dea63dfc1/plugins/org.jkiss.dbeaver.ext.postgresql/plugin.xml#L170

Destrolaric (dismissed):

chris-celerdata:
Hi @ShadelessFox @Destrolaric, checking in on this PR. I believe I've addressed all the review feedback. Is there anything else needed from my side, or is this ready for another look when you get a chance? Happy to make any additional changes. Thanks!

Destrolaric (dismissed):

plugins/org.jkiss.dbeaver.ext.mysql/icons/starRocks_icon@2x.png
ShadelessFox, line 1: Why did you remove these? The `2x` variants aren't required for SVG icons, but you still have PNG icons.

chris-celerdata: I don't have PNG icons anymore. These icons were for the MySQL version. The only icon is now at: `plugins/org.jkiss.dbeaver.ext.starrocks/icons/starrocks_icon.svg`. Do the original icons need to stay for the legacy version?

Destrolaric: Hi, I would recommend using new icons even for the legacy driver, as it still relies on them.png icons

plugins/org.jkiss.dbeaver.ext.mysql/icons/starrocks_icon.svg
ShadelessFox, line 1: Please rename it to `starrocks_icon` (all lowercase)

ShadelessFox (dismissed):
Looks good :^)

Destrolaric (dismissed):

Matvey16:
I can see only one driver (no legacy) and it has a very low-res icon

Old

dbeaver_DkWxGGXARo

New

dbeaver_gxFgVYONGw

chris-celerdata:
I don't believe the legacy one should appear. I have added this line `<replace provider="mysql" driver="starRocks"/>` as recommended [here](https://github.com/dbeaver/dbeaver/pull/39975#discussion_r2723271802). Looking at other examples, it looks like this is expected behavior. For example, `duckdb_jdbc` in the generic driver. I have added a larger icon to fix the blurriness.

chris-celerdata:
Hi @ShadelessFox @Destrolaric @Matvey16, checking in again on this PR. Is there anything else needed for this PR? If you or the team encounter any bugs, please let me know and I'll fix them ASAP. Thank you!

ShadelessFox:
Hi @chris-celerdata,

> Is there anything else needed for this PR

It currently awaits testing :^)

Matvey16:
Some default databases are not shown with the new version, the internal catalog that appears after update is empty in the navigator

dbeaver_61vbP6ZYw6


Old version for the same database

dbeaver_pfYGkyA8cQ



Maybe that's because the new version does not have this feature. Is there a reason to remove it?

image

chris-celerdata:
There's no particular reason to not support it. It seems like it would require a `.ui` plugin (e.g. `org.jkiss.dbeaver.ext.starrocks.ui`) for a custom connection page. I wasn't sure if that level of complexity was expected. If it is, I can add it; otherwise, I can remove `isSystemSchema` check to show all databases by default.

chris-celerdata:
Any preference on the `Show all databases` approach? I can have either option ready quickly. Also happy to address anything else needed for re-review; I'm really excited about using DBeaver with StarRocks and want to help move this PR along in any way I can!

carc-connor-moore:
hi @Matvey16 any chance you can take a look? Lot of people in the community (and our team) is really excited about this one.

plugins/org.jkiss.dbeaver.ext.mysql/plugin.xml
```diff
@@ -260,10 +260,9 @@
                 </driver>
                 <driver
                         id="starRocks"
-                        label="StarRocks"
-                        icon="icons/starRocks_icon.png"
-                        iconBig="icons/starRocks_icon_big.png"
-                        logoImage="icons/starRocks_logo.png"
+                        label="StarRocks (Legacy)"
```
serge-rider: It is better to hide old driver completely, just leave one line:

<driver id="starRocks"/>

serge-rider: Also delete starrocks icons from mysql plugin

plugins/org.jkiss.dbeaver.ext.starrocks/plugin.xml
```diff
@@ -0,0 +1,87 @@
...
+                class="org.jkiss.dbeaver.ext.starrocks.StarRocksDataSourceProvider"
+                description="StarRocks Database"
+                id="starrocks"
+                parent="generic"
+                dialect="starrocks"
+                label="StarRocks"
+                icon="icons/starrocks_icon.svg">
+
+            <!-- Define StarRocks tree with catalog layer -->
+            <tree path="starrocks" label="StarRocks" icon="#folder_database">
```
serge-rider: It is better to reuse Generic tree structure.
It is almost identical to generic, one exception is that `databases` are named as `catalogs` and `schemas` named as `databases`. This is really confusing. But should work ok.

The only non-standard part is materializedView - we can add it with treeInject.

plugins/org.jkiss.dbeaver.ext.starrocks/plugin.xml
```diff
@@ -0,0 +1,87 @@
...
+                        iconBig="icons/starrocks_icon_big.svg"
+                        class="com.mysql.cj.jdbc.Driver"
+                        sampleURL="jdbc:mysql://{host}[:{port}]/[{database}]"
+                        defaultPort="9030"
+                        defaultUser="root"
+                        webURL="https://docs.starrocks.io/"
+                        description="StarRocks using MySQL 8 driver"
+                        categories="sql">
+                    <replace provider="mysql" driver="starRocks"/>
+                    <file type="jar" path="maven:/com.mysql:mysql-connector-j:RELEASE[8.3.0]" bundle="!drivers.mysql"/>
```
serge-rider: There is real startRocks driver: https://docs.mirrorship.cn/docs/3.5/integrations/JDBC_driver/

It is better to use it instead of legacy MySQL driver.

plugins/org.jkiss.dbeaver.ext.starrocks/plugin.xml
serge-rider: Thus we can delete most of localizations in bundle.properties.

serge-rider (changes requested):

Destrolaric:
Hi, sorry for the delay. We rechecked and found some minor issues. Can you please take a look? `Show all databases` is not a problem, I think. It might be better to show them by default if the database doesn't have issues with it. After that, I think we might continue.

chris-celerdata:
Thank you for the review! I have updated to the StarRocks JDBC driver; it was not released until recently. Excited to get this completed!

plugins/org.jkiss.dbeaver.ext.starrocks/plugin.xml
```diff
@@ -27,18 +27,15 @@
...
+                        sampleURL="jdbc:starrocks://{host}[:{port}]/[{database}]"
                         defaultPort="9030"
                         defaultUser="root"
                         webURL="https://docs.starrocks.io/"
-                        description="StarRocks using MySQL 8 driver"
+                        description="StarRocks JDBC driver"
                         categories="sql">
                     <replace provider="mysql" driver="starRocks"/>
-                    <file type="jar" path="maven:/com.mysql:mysql-connector-j:RELEASE[8.3.0]" bundle="!drivers.mysql"/>
-                    <file type="license" path="https://www.gnu.org/licenses/old-licenses/lgpl-2.0.txt" bundle="!drivers.mysql"/>
```
Destrolaric: Could you add a path to a license file? It will be great.

Destrolaric:
Fixed merge conflict, also added some minor comments regarding a license file. For now, I am probably going to move it back to testing. It might take a while, as some parts of the testing team are on vacation.

DariyaFedorova:
Hi @chris-celerdata,
Can't build this PR because of compilation error:
```
[ERROR] /var/jenkins_home/jobs/DBeaverDesktop/jobs/build-copy/workspace/dbeaver/plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/StarRocksDataSourceProvider.java:[31,8] constructor GenericDataSourceProvider in class org.jkiss.dbeaver.ext.generic.GenericDataSourceProvider<DATASOURCE> cannot be applied to given types;
[ERROR]   required: java.lang.Class
[ERROR]   found:    no arguments
[ERROR]   reason: actual and formal argument lists differ in length
[ERROR] /var/jenkins_home/jobs/DBeaverDesktop/jobs/build-copy/workspace/dbeaver/plugins/org.jkiss.dbeaver.ext.starrocks/src/org/jkiss/dbeaver/ext/starrocks/StarRocksDataSourceProvider.java:[42,26] openDataSource(org.jkiss.dbeaver.model.runtime.DBRProgressMonitor,org.jkiss.dbeaver.model.DBPDataSourceContainer) in org.jkiss.dbeaver.ext.starrocks.StarRocksDataSourceProvider cannot override openDataSource(org.jkiss.dbeaver.model.runtime.DBRProgressMonitor,org.jkiss.dbeaver.model.DBPDataSourceContainer) in org.jkiss.dbeaver.ext.generic.GenericDataSourceProvider
[ERROR]   return type org.jkiss.dbeaver.model.DBPDataSource is not compatible with org.jkiss.dbeaver.ext.generic.model.GenericDataSource
```

chris-celerdata:
@DariaMarkaryan Sorry about that, the merge caused that compilation error. It should be fixed now.

DariyaFedorova:
@chris-celerdata 
Now all top-level folders have disappeared from the navigation tree (Advanced view is on)
image
And what it looked like:
image

chris-celerdata:
This is expected behavior. The top level folders were from the MySQL tree structure and this plugin uses the generic tree structure. 

Some of those folders (Users, Session/Global Status) were either broken or unused. However, Session Variables and Global Variables were functional and are now missing. Re-implementing this is non-trivial so a follow-up may be better. For example, users in StarRocks are queried differently than MySQL and requires custom parsing and tree injection (for generic).

DariyaFedorova:
Hi @chris-celerdata 
Sorry for the delay. 
I tried to connect to the external Hive catalog, but an error occurs when I expand it
```
java.sql.SQLSyntaxErrorException: Failed to getAllDatabases, msg: Unable to instantiate org.apache.hadoop.hive.metastore.HiveMetaStoreClient
	at com.starrocks.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:112)
	at com.starrocks.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:114)
	at com.starrocks.cj.jdbc.ClientPreparedStatement.executeInternal(ClientPreparedStatement.java:991)
	at com.starrocks.cj.jdbc.ClientPreparedStatement.executeQuery(ClientPreparedStatement.java:1059)
	at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCPreparedStatementImpl.executeQuery(JDBCPreparedStatementImpl.java:229)
	at org.jkiss.dbeaver.ext.starrocks.model.StarRocksMetaModel.loadSchemas(StarRocksMetaModel.java:115)
	at org.jkiss.dbeaver.ext.generic.model.GenericCatalog.getSchemas(GenericCatalog.java:86)
	at org.jkiss.dbeaver.ext.generic.model.GenericCatalog.getSchemaList(GenericCatalog.java:77)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.jkiss.dbeaver.model.navigator.DBNDatabaseNode.extractPropertyValue(DBNDatabaseNode.java:1055)
	at org.jkiss.dbeaver.model.navigator.DBNDatabaseNode$PropertyValueReader.run(DBNDatabaseNode.java:1127)
	at org.jkiss.dbeaver.model.navigator.DBNDatabaseNode$PropertyValueReader.run(DBNDatabaseNode.java:1112)
	at org.jkiss.dbeaver.model.exec.DBExecUtils.tryExecuteRecover(DBExecUtils.java:210)
	at org.jkiss.dbeaver.model.navigator.DBNDatabaseNode.loadTreeItems(DBNDatabaseNode.java:657)
	at org.jkiss.dbeaver.model.navigator.DBNDatabaseNode.loadChildren(DBNDatabaseNode.java:470)
	at org.jkiss.dbeaver.model.navigator.DBNDatabaseNode.getChildren(DBNDatabaseNode.java:229)
	at org.jkiss.dbeaver.model.navigator.DBNDatabaseNode.getChildren(DBNDatabaseNode.java:55)
	at org.jkiss.dbeaver.model.navigator.DBNUtils.getNodeChildrenFiltered(DBNUtils.java:106)
	at org.jkiss.dbeaver.ui.navigator.database.load.TreeLoadService.evaluate(TreeLoadService.java:50)
	at org.jkiss.dbeaver.ui.navigator.database.load.TreeLoadService.evaluate(TreeLoadService.java:31)
	at org.jkiss.dbeaver.ui.LoadingJob.run(LoadingJob.java:99)
	at org.jkiss.dbeaver.model.runtime.AbstractJob.run(AbstractJob.java:110)
```

chris-celerdata:
Hey @DariaMarkaryan, this problem is likely due to the connection between StarRocks and the Hive database failing. You can test this by connecting to StarRocks using `mysql` and running `SET CATALOG <hive_catalog>; SHOW DATABASES;`. I have a local Hive catalog and I was able to connect on DBeaver (see image below). 
image

Matvey16:
Verified

PR #39998, opened 2025-12-25
#39884 Allow exporting just fetched rows

### LINKED ISSUE #39884: [Export] Clarify possibility to export without re-running query
As explained in issue #39692, users sometimes need to export data without re-running the query.

Like the user who opened that issue, I wanted that feature but was unaware it already existed:
'Selected rows only' option

It's easy to get lost among the many options when exporting data, and I think the link between the "Selected rows only" option and the fact that the query won't be run again is far from obvious.

Maybe the options could be reorganized a bit? I personally see the fact of re-running the query or not as an important distinction, so I would put it at the top, maybe as a radio button with 2 options: "Export from memory" vs. "Re-run query". I can't find a tool to quickly create UX mockups so I'll just draw a table to show how the options might then be visually rearranged:
| ◉ Export from memory | ◎ Re-run query |
|--------|--------|
| ☑ Selected columns only | ☑ Open new connection |
|  | ☐ Select row count |
|  | Fetch size: [input] |

The "Selected rows only" would therefore no longer be necessary. Probably likewise for "Selected columns only" actually?

ShadelessFox:
The whole page has been completely redesigned:

image

Please thoroughly test whether everything works.

_"Why is this PR 1.5k LoC"_, you may think. It's totally an expected reaction. Let me explain: it introduces a new Forms API that I've been working on for some time now. It implements a builder-like pattern, similar to one seen in the new [ClassFile API](https://openjdk.org/jeps/484). It still has a lot of rough edges, but without actually using it, it's impossible to complete it. So I assumed the issue this PR resolves is a perfect candidate to finally reveal it and merge. :^)


IvanGorshechnikov (dismissed):

uslss:
works as expected, but the extraction settings page looks bad on win 
image
image

uslss:
verified

PR #40020, opened 2025-12-30
7710 dbvr cli connection management command

alexander-skoblikov:
closes https://github.com/dbeaver/pro/issues/7710


plugins/org.jkiss.dbeaver.model.cli/META-INF/MANIFEST.MF
```diff
@@ -13,6 +13,7 @@ Export-Package: org.jkiss.dbeaver.model.cli,
  org.jkiss.dbeaver.model.cli.model.option,
  org.jkiss.dbeaver.model.cli.command
 Require-Bundle: org.jkiss.dbeaver.model;visibility:=reexport,
+ org.jkiss.dbeaver.registry;visibility:=reexport,
```
serge-rider: Remove

serge-rider (changes requested):

PR #40055, opened 2026-01-07
dbeaver/dbeaver#39210 Sybase SQL dialect

### LINKED ISSUE #39210: Double-Clicking on database table to view data fails in Sybase 11 due to alias keyword "AS"
### Description

Double-Clicking on database table to view data fails in Sybase 11 due to alias keyword "AS"

### DBeaver Version

Community Edition 25.2.0

### Operating System

Debian 13 Trixie

### Database and driver

Database: Sybase ASE 11.0.3
JDBC Driver: Sybase jTDS driver (jtds-1.3.1.jar)

### Steps to reproduce

1. Connect to a Sybase ASE 11.0.3 database
2. Double-click to open a table to view its Data

This error is displayed (spanish): 
```
SQL Error [107] [S1000]: El prefijo de columna 'm' no coincide con ningún nombre de tabla o nombre de alias utilizado en la consulta. La tabla no se ha especificado en la cláusula FROM o tiene un nombre de correlación que debe utilizarse en su lugar.
```

This is the SELECT built by DBeaver, which does not work because "AS" is not supported for tables in Sybase ASE 11.0.3:

```sql
SELECT m.* FROM mydb.dbo.mytable AS m
```

This SELECT works fine with the alias but without the "AS" keyword:
```sql
SELECT m.* FROM mydb.dbo.mytable m
```

### Additional context

It worked on DBeaver 24.1.0 Community Edition

serge-rider:


Matvey16:
Verified

PR #40058, opened 2026-01-08
Fix: Cache entity attributes to avoid repeated metadata queries (#40057)

### LINKED ISSUE #40057: Timeouts and slowness when we run redshift datashare queries with lot of columns (30+)
### Description

Unnecessary metadata calls cause severe slowdown on Datashare connections in DBeaver 25.x

Environment

DBeaver versions affected:

25.1.4.202508031529

25.3.2

Database: Amazon Redshift Datashare consumer

Driver: Redshift JDBC 2.2

Connection mode: Read-only, server enforced

Problem Description

Unnecessary metadata calls are being triggered for simple SELECT queries in DBeaver 25.x.
On Datashare read-only connections, this causes queries to become very slow because each metadata call results in additional catalog access across the Datashare boundary.

On non-Datashare connections, the same behavior introduces only a minor lag. However, the extra metadata calls are still present.

This indicates that additional metadata processing was introduced in recent versions

Reproduction Steps
Query 1: Single column
select col1 from sample_schema.sample_table limit 10;

Observed behavior in DBeaver 25.x

A metadata query against pg_catalog is executed three times to inspect column properties such as nullability and generated or identity behavior.

The following metadata query is executed:

SELECT c.oid, a.attnum, a.attname, c.relname, n.nspname,
       a.attnotnull OR (t.typtype = 'd' AND t.typnotnull),
       pg_catalog.pg_get_expr(d.adbin, d.adrelid) LIKE '%nextval(%'
FROM pg_catalog.pg_class c
JOIN pg_catalog.pg_namespace n ON (c.relnamespace = n.oid)
JOIN pg_catalog.pg_attribute a ON (c.oid = a.attrelid)
JOIN pg_catalog.pg_type t ON (a.atttypid = t.oid)
LEFT JOIN pg_catalog.pg_attrdef d
       ON (d.adrelid = a.attrelid AND d.adnum = a.attnum)
JOIN (SELECT <table_oid> AS oid, <attnum>) vals
       ON (c.oid = vals.oid AND a.attnum = vals.attnum);

Query 2: Two columns
select col1, col2 from sample_schema.sample_table limit 10;

Observed behavior in DBeaver 25.x

The same metadata query is executed six times

The number of metadata calls scales linearly with the number of projected columns

Observed Pattern

The same metadata lookup is executed approximately three times per statement

Each execution scales linearly with the number of selected columns

One column results in three executions

Two columns result in six executions

Metadata inspection includes checks against column default expressions to detect generated or identity columns

This occurs even when the connection is strictly read-only and result sets are not editable

Expected Behavior

Should run fast.

Minimal metadata introspection for simple SELECT queries

Impact

Severe performance degradation on Datashare consumer clusters

Queries that were previously fast become noticeably slow

Performance impact increases with the number of selected columns

Cannot be mitigated at the SQL level because all operations are read-only

Negative impact on analyst and exploratory query workflows

Notes

The same queries executed using Python and SQLAlchemy do not trigger these metadata calls

Disabling editable result sets and column reference resolution reduces but does not eliminate the behavior in DBeaver 25.x

This strongly suggests that additional result set metadata processing is the issue.

### DBeaver Version

25.3.2

### Operating System

Windows 11

### Database and driver

redshift-jdbc42 2.2.0

### Steps to reproduce

1. Create 2 redshift clusters (Cluster A, ClusterB)
2. In cluster A, create schema1.table1 with 30 columns
3. Do datashare from cluster A to cluster B
4. Create external schema to point to schema1
5. Access table1 with select * from schema1.table1

### Additional context

This error was not there in 23.3.0

deepaksenthilkumar:
When executing SELECT queries, DBeaver was calling getAttribute() for each column individually, which triggered metadata queries to pg_catalog for every column. On Redshift Datashare connections, this caused severe performance degradation as each metadata call must cross the Datashare boundary.

This fix implements attribute caching per entity:
- Loads all entity attributes once using getAttributes()
- Stores them in a HashMap for O(1) lookup
- Reuses cached attributes for subsequent columns from the same entity

Impact:
- Before: N columns = ~3N metadata queries
- After: N columns = 1 metadata query per unique entity
- Eliminates performance issues on Datashare and other slow metadata sources

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
@@ -730,6 +730,9 @@ public class DBExecUtils {
 
             boolean needsTableMetaForColumnResolution = dataSource.getInfo().needsTableMetaForColumnResolution();
 
+            // Cache for entity attributes to avoid repeated getAttribute() calls (#40057)
+            Map<DBSEntity, Map<String, DBSEntityAttribute>> entityAttributesCache = new HashMap<>();
+
             monitor.subTask("Discover attributes");
             for (DBDAttributeBinding binding : bindings) {
                 monitor.subTask("Discover attribute '" + binding.getName() + "'");
@@ -825,12 +828,30 @@ public class DBExecUtils {
                     if (bindingMeta.getPseudoAttribute() != null) {
                         tableColumn = bindingMeta.getPseudoAttribute().createFakeAttribute(attrEntity, attrMeta);
                     } else if (columnName != null) {
+                        // Use cached attributes to avoid repeated database metadata queries (#40057)
+                        Map<String, DBSEntityAttribute> attributesMap = entityAttributesCache.get(attrEntity);
+                        if (attributesMap == null) {
+                            // Load all attributes once for this entity
+                            attributesMap = new HashMap<>();
+                            try {
+                                Collection<? extends DBSEntityAttribute> attributes = attrEntity.getAttributes(mdMonitor);
+                                if (attributes != null) {
+                                    for (DBSEntityAttribute attr : attributes) {
+                                        attributesMap.put(attr.getName(), attr);
+                                    }
+                                }
+                            } catch (DBException e) {
+                                log.debug("Error loading entity attributes: " + e.getMessage());
+                            }
+                            entityAttributesCache.put(attrEntity, attributesMap);
+                        }
+                        
                         if (sqlQuery == null) {
-                            tableColumn = attrEntity.getAttribute(mdMonitor, columnName);
+                            tableColumn = attributesMap.get(columnName);
                         } else {
                             boolean isAllColumns = sqlQuery.getSelectItemAsteriskIndex() != -1;
                             if (isAllColumns || (selectItem != null && (selectItem.isPlainColumn() || selectItem.getName().equals("*")))) {
-                                tableColumn = attrEntity.getAttribute(mdMonitor, columnName);
+                                tableColumn = attributesMap.get(columnName);
                             }
                         }
                     }


deepaksenthilkumar:
@serge-rider  - Can you please help in looking into this

arhayka:
Hi @deepaksenthilkumar! Thank you for the contribution.
I will pass your PRs to the team for review and testing.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
```diff
@@ -825,12 +828,30 @@ public static void bindAttributes(
...
+                                if (attributes != null) {
+                                    for (DBSEntityAttribute attr : attributes) {
+                                        attributesMap.put(attr.getName(), attr);
+                                    }
+                                }
+                            } catch (DBException e) {
+                                log.debug("Error loading entity attributes: " + e.getMessage());
+                            }
+                            entityAttributesCache.put(attrEntity, attributesMap);
+                        }
```
E1izabeth, line 847: Hi, @deepaksenthilkumar!
Could you explain the approach under these changes? 

Also, could you refactor this code into a new method to improve readability?

ged-yuko:
I personally don't experience such effects with all the databases i use, which leads me to the thought that the problem is maybe redshift-specific, and so the fix should rather be in the redshift integration plugin than in the common code applied for all the databases. @deepaksenthilkumar did you look into the `getAttributes(..)`/`getAttribute(..)`  implementation? It typically already uses cache inside, which basically is literally the same Map collection. The usage scenario details required to understand why the existing cache doesn't help. 

This should have been asked in the original issue, but stiil:
@deepaksenthilkumar do you experience the same problems with repeated queries when executing the same SELECT statement multiple times? 
Like this: run your complex SELECT-something, observe unwanted effects, wait a bit, then run the same SELECT-something again. Are unwanted effects still presented?

I understand that short-living per-query cache may have a point for certain optimizations, but the original issue looks like the _alredy existing cache_ doesn't work. This is _really_ strange in the first place.

SolovevAnton:
Thank you for your contribution. With the second look at your PR we discovered the cache issue `org.jkiss.dbeaver.ext.postgresql.model.PostgreTableBase#getAttribute`.  So we decided to fix the issue at the PostgreSQL table class level ourselves.

PR #40059, opened 2026-01-08
feat: Add preference to skip metadata reading for resultset

### LINKED ISSUE #40057: Timeouts and slowness when we run redshift datashare queries with lot of columns (30+)

deepaksenthilkumar:
Adds user preference 'Skip Metadata Reading For Resultset' to optionally bypass metadata reading when not needed.

This addresses issue #40057 by providing an alternative approach:
- Adds RESULT_SET_SKIP_METADATA_READING preference (default: false)
- When enabled, skips entire metadata binding loop in DBExecUtils
- Eliminates ~3N metadata queries (where N = column count)
- Maintains backward compatibility with false default

Use cases:
- Redshift Datashare with slow cross-database metadata queries
- Read-only query scenarios
- Performance-critical applications

Files modified:
- ModelPreferences.java: Added RESULT_SET_SKIP_METADATA_READING
- ResultSetPreferences.java: Added RESULT_SET_SKIP_METADATA_READING
- DataEditorsPreferencesInitializer.java: Set default value
- DBExecUtils.java: Added preference check before binding loop


deepaksenthilkumar:
@serge-rider - Can you please help in looking into this

serge-rider:
We already have metadata disable option:
image

Does it really make sense to add yet another one?

deepaksenthilkumar:
It's not working

image


image

Can you check why and combine this skip there or want me to check that?

deepaksenthilkumar:
@serge-rider This has been affecting us for a over a year now. Pls help.

SolovevAnton:
Thank you for your contribution. With the second look at your PR we discovered the cache issue org.jkiss.dbeaver.ext.postgresql.model.PostgreTableBase#getAttribute. So we decided to fix the issue at the PostgreSQL table class level ourselves.

PR #40073, opened 2026-01-09
Dbeaver/pro#7899 support user preferences desktop

SolovevAnton:
closes dbeaver/pro#7899


kseniaguzeeva:
**A**. Context menu: _use current user setting_ becomes marked after changing the connection view.
ER: it's disabled, when a default view is used for the connection and becomes active if a user has changed the view via the context menu.

STR:
1. Login to Desktop TE
2. Select in a connection context menu Connection view - use current user setting

https://github.com/user-attachments/assets/7cb12bf5-2b20-44cf-8fe7-65305f7f0ad6

**B**. impossible to change global connection view with enabled user connection view. STR:
1. Login by admin and change a connection view via the context menu
2. Go to the connection settings and open General tab - the user connection view is displayed there instead of global
3. Change global connection view - the database is reconnected and changes the view accordingly, _use current user setting_ isn't marked in the context menu and the view from connection settings is selected
4. Relaunch the app and check connection view in the context menu and settings

AR: 

- context menu shows the selected view by the user again and its really applied (correctly)
- global view change is reset to previous in the connection settings

https://github.com/user-attachments/assets/85be788a-e92b-4807-bf8b-315a9085cd63

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/actions/datasource/ConnectionViewSettingsContributor.java
```diff
@@ -227,4 +238,44 @@ public void run() {
...
+            }
+            if (dsContainer instanceof DataSourceDescriptor descriptor) {
+                descriptor.setNavigatorSettings(DataSourceNavigatorSettings.getDefaultSettings());
+            }
+            askToReconnectIfNeeded();
+            enableCheck();
+        }
+
+        private void enableCheck() {
+            setEnabled(dsContainer.getNavigatorSettings().isUserSettings());
```
ShadelessFox: Perhaps the option itself should be rather hidden than disabled? I expect it might be confusing

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/actions/datasource/ConnectionViewSettingsContributor.java
```diff
@@ -227,4 +238,44 @@ public void run() {
...
+            refreshNavigator();
+        }
+
+        @Override
+        void updateSettings(DBNBrowseSettings settings) {
+            super.updateSettings(settings);
+            enableCheck();
+        }
+
+        void clearCurrentUserSettings() {
```
ShadelessFox: Can be `private`

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/actions/datasource/ConnectionViewSettingsContributor.java
```diff
@@ -227,4 +248,57 @@ public void run() {
...
+            refreshNavigator();
+        }
+
+        public void setSeparator(Separator separator) {
+            this.separator = separator;
+        }
+
+        public void setClearUserSettings(ActionContributionItem clearUserSettings) {
+            this.clearUserSettings = clearUserSettings;
+        }
```
ShadelessFox: Could be constructor parameters instead

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/actions/datasource/ConnectionViewSettingsContributor.java
```diff
@@ -227,4 +248,57 @@ public void run() {
...
+
+        @Override
+        void updateSettings(DBNBrowseSettings settings) {
```
ShadelessFox: Please annotate the base method with `@NotNull`

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/dialogs/connection/ConnectionPageGeneral.java
```diff
@@ -690,6 +689,16 @@ public void saveSettings(@NotNull DBPDataSourceContainer dataSource) {
         }
     }
 
+    private void processNavigatorSettings(DataSourceDescriptor dsDescriptor) {
```
ShadelessFox:
```suggestion
    private void processNavigatorSettings(@NotNull DataSourceDescriptor dsDescriptor) {
```

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceNavigatorSettings.java
```diff
@@ -231,6 +231,10 @@ public boolean equals(Object obj) {
     private static final String DEFAULT_HIDE_VIRTUAL_MODEL = "navigator.settings.default.hideVirtualModel";
 
     public static DBNBrowseSettings getDefaultSettings() {
+        return getDefaultDataSourceNavigatorSettings();
+    }
+
+    public static DataSourceNavigatorSettings getDefaultDataSourceNavigatorSettings() {
```
ShadelessFox:
```suggestion
    @NotNull
    public static DataSourceNavigatorSettings getDefaultDataSourceNavigatorSettings() {
```

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/actions/datasource/ConnectionViewSettingsContributor.java
SolovevAnton: Unfortunately, impossible, since ActionContributionItem, which is used here, has ClearCurrentUserSettings  as a constructor parameter. could do only for separator - but decided to stay consistent with set function here

ShadelessFox (dismissed):

IvanGorshechnikov (dismissed):

IvanGorshechnikov (dismissed):

PR #40091, opened 2026-01-12
dbeaver/dbeaver#39979 Transaction mode detect fix

### LINKED ISSUE #39979: Although the transaction mode is set to manual commit, it is displayed as auto commit.
### Description

In version 25.3.1.202512211813, after updating to this version and opening an existing SQL console that previously had its transaction mode set to manual, I noticed that the transaction mode displayed at the top was changed to auto instead of manual.

### DBeaver Version

Community edition 25.3.1.202512211813

### Operating System

Windows 11 pro 25H2

### Database and driver

database Driver for MySQL 8

### Steps to reproduce

Below is a screenshot showing the transaction mode displayed as auto after the update.
When clicking the Commit or Rollback button, the previously displayed None appears. From that point on, the number corresponding to the commit history is displayed correctly and behaves as expected.

Image

Image

I am not sure whether the same behavior occurs when connecting to other types of databases, but in the case of MySQL, it does not require any special reproduction steps—if it is an existing MySQL console that was previously used with manual commit mode, it appears to occur consistently.

### Additional context

_No response_

serge-rider:


Matvey16:
Transaction box doesn't say `Auto` anymore, but still doesn't show pending queries in the active trancasction. It just says `None` instead

https://github.com/user-attachments/assets/47348301-a66a-4211-94f2-1c78eb6d709b

Matvey16:
Verified

PR #40093, opened 2026-01-12
#40082 Fix "Pin tab" shortcut to always act on active tab

### LINKED ISSUE #40082: "Pin tab" shortcut logic should always pin or unpin the currently active tab, not some other
### Description

Steps to reproduce:

- press ctrl-shift-p again

Expected result: 

Actual result:

### DBeaver Version

### Operating System

Debian forky

### Database and driver

SQLite

### Steps to reproduce

_No response_

### Additional context

_No response_

HellAmbro:
This PR fix issue  #40082


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -1469,10 +1469,12 @@ public void mouseUp(MouseEvent e) {
...
 
         {
-            resultTabs.addMouseListener(new MouseAdapter() {
+            resultTabs.addSelectionListener(new SelectionAdapter() {
                 @Override
-                public void mouseDown(MouseEvent e) {
-                    activeResultsTab = resultTabs.getItem(new Point(e.x, e.y));
+                public void widgetSelected(SelectionEvent e) {
+                    if (e.item instanceof CTabItem) {
+                        activeResultsTab = (CTabItem) e.item;
```
SolovevAnton:
```suggestion
                    if (e.item instanceof CTabItem cbItem) {
                        activeResultsTab = cbItem;
```

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -1880,8 +1882,18 @@ private ToolItem getViewToolItem(@NotNull String commandId) {
     }
 
     private CTabItem getActiveResultsTab() {
```
SolovevAnton, line 1887:
```suggestion
    @Nullable
    private CTabItem getActiveResultsTab() {
```

SolovevAnton (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
HellAmbro: I used `cTabItem` instead of `cbItem`, is that okay?

SolovevAnton (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -1469,10 +1469,12 @@ public void mouseUp(MouseEvent e) {
         }
 
         {
-            resultTabs.addMouseListener(new MouseAdapter() {
+            resultTabs.addSelectionListener(new SelectionAdapter() {
```
ShadelessFox: While we're still here, let's also replace this adapter with `SelectionListener.widgetSelectedAdapter`

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -1879,9 +1881,20 @@ private ToolItem getViewToolItem(@NotNull String commandId) {
...
+        if (resultTabs == null || resultTabs.isDisposed()) {
+            return null;
+        }
+
+        CTabItem selected = resultTabs.getSelection();
+        if (selected != null && !selected.isDisposed()) {
+            return selected;
+        }
+
+        return (activeResultsTab != null && !activeResultsTab.isDisposed())
```
ShadelessFox: Redundant parentheses


```suggestion
        return activeResultsTab != null && !activeResultsTab.isDisposed()
```

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -1879,9 +1881,20 @@ private ToolItem getViewToolItem(@NotNull String commandId) {
...
+                ? activeResultsTab
+                : null;
```
ShadelessFox: We use 4-space indentation


```suggestion
            ? activeResultsTab
            : null;
```

ShadelessFox (commented):
Small nitpicks

HellAmbro:
@ShadelessFox thanks for your review, i've updated the code accordling.

ShadelessFox (approved):
Looking good. Thanks for the contribution :^)

SolovevAnton (approved):

Matvey16:
With this change, you transfer the problem to a context menu. With current changes any action (including pinning) to a tab from a context menu ignores the actual tab from which the context menu was opened, the action is applied to currently selected tab


https://github.com/user-attachments/assets/d702868a-df82-4410-b512-13af2a8d0ca7

HellAmbro:
Hi @Matvey16 , thanks for your help i noticed the error. I found a little time to work on that issue, i created a new PR #40223. 
I'm closing this PR since i did a different work. I would appreciate it if you could take a look at it. 
Thanks.

PR #40114, opened 2026-01-15
dbeaver/pro#7877 temporarily use getString instead of getArray for UUID, IPv4, IPv6 and Map

Destrolaric:
Closes dbeaver/pro#7877


plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/data/ClickhouseArrayValueHandler.java
```diff
@@ -59,6 +60,25 @@ protected boolean useSetArray(@NotNull DBCSession session, @NotNull DBSTypedObje
...
+    ) throws DBCException, SQLException {
+        // Remove after https://github.com/ClickHouse/clickhouse-java/issues/2711 is fixed
+        if (type.getTypeName().toLowerCase().contains("ipv4")
+            || type.getTypeName().toLowerCase().contains("ipv6")
+            || type.getTypeName().toLowerCase().contains("uuid")
+            || type.getTypeName().toLowerCase().contains("map")
+        ) {
+            return getValueFromObject(session, type, resultSet.getString(index), false, false);
+        }
+        return super.fetchColumnValue(session, resultSet, type, index);
```
ShadelessFox: This is a suboptimal solution, as it affects every version of the driver, including those that work just fine.

Wrap `super.fetchColumnValue` in a try-catch statement and fallback to `getValueFromObject(session, type, resultSet.getString(index), false, false)` in case of an exception. Don't forget to log the exception as well.

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/data/ClickhouseArrayValueHandler.java
```diff
@@ -59,6 +60,25 @@ protected boolean useSetArray(@NotNull DBCSession session, @NotNull DBSTypedObje
...
+        @NotNull DBCSession session,
+        @NotNull JDBCResultSet resultSet,
+        @NotNull DBSTypedObject type,
+        int index
+    ) throws DBCException, SQLException {
+        // Remove after https://github.com/ClickHouse/clickhouse-java/issues/2711 is fixed
+        if (type.getTypeName().toLowerCase().contains("ipv4")
+            || type.getTypeName().toLowerCase().contains("ipv6")
+            || type.getTypeName().toLowerCase().contains("uuid")
+            || type.getTypeName().toLowerCase().contains("map")
```
ShadelessFox: For the love of God, please cache `type.getTypeName().toLowerCase()`

ShadelessFox (changes requested):

ShadelessFox (approved):

Matvey16:
Verified

PR #40117, opened 2026-01-15
dbeaver/pro#7335 Specify custom instructions for different prompts

ShadelessFox:
Demonstration against the AI chat. AI chat is a PRO feature.

java_EFQUgWP244


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/registry/AIPromptGeneratorDescriptor.java
```diff
@@ -44,24 +45,25 @@ protected AIPromptGeneratorDescriptor(@NotNull IConfigurationElement config) {
         super(config);
         this.contributorConfig = config;
         this.objectType = new ObjectType(config, RegistryConstants.ATTR_CLASS);
-        this.icon = iconToImage(config.getAttribute(RegistryConstants.ATTR_ICON));
+        this.id = contributorConfig.getAttribute("id");
```
IvanGorshechnikov:
```suggestion
        this.id = contributorConfig.getAttribute(RegistryConstants.ATTR_ID);
```

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AISettings.java
```diff
@@ -61,7 +62,7 @@ public void setProperty(@NotNull String name, @Nullable Object value) {
 
     @NotNull
     public Set<String> getEnabledFunctions() {
-        return new HashSet<>(enabledFunctions);
+        return Set.copyOf(enabledFunctions);
```
IvanGorshechnikov, line 77: It could break somewhere, because you make this set immutable

IvanGorshechnikov (changes requested):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AISettings.java
ShadelessFox: I've checked every call site before making this change :^)

IvanGorshechnikov (dismissed):

PR #40140, opened 2026-01-19
Dbeaver/pro#7925 save navigator filters for current user only

SolovevAnton:
closes dbeaver/pro#7925: adds option to save current user filter in TE.

_For reviewers:_ now it saves user-defined filters in cb_object_settings as json array string value


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/json/JSONUtils.java
```diff
@@ -50,6 +50,8 @@ public class JSONUtils {
         .ofPattern("yyyy-MM-dd['T'HH:mm:ss['.'SSS]['Z']]")
         .withZone(ZoneId.of("UTC"));
     public static final Type MAP_TYPE_TOKEN = new TypeToken<Map<String, Object>>() {}.getType();
+    public static final Type LIST_TYPE_TOKEN = new TypeToken<List<Map<String, Object>>>() {
+    }.getType();
```
ShadelessFox: Consider using `TypeToken#getParameterized` instead of creating an anonymous class.


```suggestion
    public static final Type LIST_TYPE_TOKEN = TypeToken.getParameterized(List.class, MAP_TYPE_TOKEN).getType();
```

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/FilterSerializer.java
```diff
@@ -0,0 +1,143 @@
...
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.OutputStreamWriter;
+import java.io.StringReader;
+import java.nio.charset.StandardCharsets;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+
+public class FilterSerializer<T extends DataSourceDescriptor> {
```
ShadelessFox, line 37: If it's not intended to be subclassed, consider making it `final`
```suggestion
public final class FilterSerializer<T extends DataSourceDescriptor> {
```

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/FilterSerializer.java
```diff
@@ -0,0 +1,143 @@
...
+
+    protected static final Gson CONFIG_GSON = new GsonBuilder()
+        .setStrictness(Strictness.LENIENT)
+        .serializeNulls()
+        .create();
```
ShadelessFox: Pretty sure we have a hundred of these now. Can we reuse those that are available from utility classes?

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/UserDBSObjectFilterUtils.java
```diff
@@ -0,0 +1,109 @@
...
+import org.jkiss.dbeaver.model.app.DBPProject;
+import org.jkiss.dbeaver.model.security.SMObjectType;
+
+import java.io.IOException;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class UserDBSObjectFilterUtils {
```
ShadelessFox:
```suggestion
public final class UserDBSObjectFilterUtils {
```
And also make a private constructor

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/UserDBSObjectFilterUtils.java
```diff
@@ -0,0 +1,109 @@
...
+
+    private static final Log log = Log.getLog(UserDBSObjectFilterUtils.class);
+
+    public static final String USER_FILTER_KEY = "navigator-filters";
+
+    protected static final FilterSerializer<DataSourceDescriptor> filterSerializer = new FilterSerializer<>();
```
ShadelessFox: Should likely be `private`. It's also created another time in `DataSourceSerializerModern`

ShadelessFox (changes requested):
It's really bad that TE-specific user filters are literally leaking into the CE codebase. Can we do something about it?

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/FilterSerializer.java
SolovevAnton: I'm not sure about this, actually. Maybe we should keep the option to subclass this serializer, in case we will want to create some other implementations, for special filter serialization use cases. Not totally sure

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/UserDBSObjectFilterUtils.java
SolovevAnton: changed in UserDBSObjectFilterUtil. In the data serializer, modern it remained protected to be consistent with protected static final Gson CONFIG_GSON. If I understand correctly, it is done on purpose, since DataSourceSerializerModern should be able to be subclassed with updated behavior

IvanGorshechnikov:
Unfortunately, this PR has conflicts

IvanGorshechnikov (dismissed):

IvanGorshechnikov (approved):

PR #40081, opened 2026-01-11
#40065 Implement copy/cut whole line without selection

jihoon0413:
### Description
This PR implements the "Smart Copy/Cut" feature for the SQL Editor.
Currently, pressing `Ctrl+C` or `Ctrl+X` without selecting any text does nothing. With this change, the operation is applied to the **whole line** where the caret is currently located, which aligns with the standard behavior of modern IDEs (VS Code, IntelliJ, Eclipse, etc.).

Resolves dbeaver/dbeaver#40065

### Changes
- Modified `SQLEditorBase.java`:
  - Overridden `createActions` to register a custom `SmartEditAction` instead of the default Copy/Cut actions.
  - Added `SmartEditAction` inner class to handle the logic:
    - Checks if the selection is empty.
    - If empty, calculates the current line range (including the line delimiter).
    - Performs Copy or Cut on the whole line.
    - Restores caret position after Copy.
    - Includes a check for Read-Only mode before performing Cut.

### Verification (How to test)
1. Open the SQL Editor.
2. **Case 1: Smart Copy**
   - Place the caret on any line without selecting text.
   - Press `Ctrl+C`.
   - Paste (`Ctrl+V`) somewhere else -> Verify the **entire line** is pasted.
   - Verify the caret remains at the original position.
3. **Case 2: Smart Cut**
   - Place the caret on any line without selecting text.
   - Press `Ctrl+X`.
   - Verify the **entire line is removed**.
   - Paste (`Ctrl+V`) -> Verify the removed line is pasted.
4. **Case 3: Standard Selection**
   - Select a specific part of the text.
   - Press `Ctrl+C` or `Ctrl+X`.
   - Verify that only the **selected text** is affected

https://github.com/user-attachments/assets/99e665ba-034e-4a1e-8bf2-2ca6a3c19a27

PR #40152, opened 2026-01-20
#40065 Fixed: Copy/Cut items missing from context menu

### LINKED ISSUE #40065: Text Editor: Ctrl-X/Ctrl-C operate on whole line if no selection
It is currently standard in almost all IDEs, that when you press ctrl-c or ctrl-x and _no text is selected_, the operation is applied on the whole line.
I find myself making errors multiple times when editing as this behavior is so strong in muscle memory, that it's hard to edit/move code lines in the dbeaver editor.

As described, pressing Ctrl-X without any text selected, shall put the entire line to clipboard. Same for ctrl-c.

Selecting text then pressing Ctrl-X. Feels clumsy nowadays.

**Additional context**
--

jihoon0413:
Description:
Previously, PR #40081 fixed the issue of the copy/cut item disappearing from the context menu.

Resolves dbeaver/dbeaver#40065

Modifications included in this PR:
Modify context menu: You have fixed the issue where the "copy" and "cut" labels/icon are missing from the context menu by explicitly setting the text and image descriptors in createActions.


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorBase.java
```diff
@@ -770,9 +772,36 @@ protected void createActions() {
...
+        action = createSmartCopyCutAction(ITextOperationTarget.COPY, IWorkbenchCommandConstants.EDIT_COPY);
+        setAction(ITextEditorActionConstants.COPY, action);
+
+        action = createSmartCopyCutAction(ITextOperationTarget.CUT, IWorkbenchCommandConstants.EDIT_CUT);
+        setAction(ITextEditorActionConstants.CUT, action);
+
         SQLEditorCustomActions.registerCustomActions(this);
     }
 
+    private IAction createSmartCopyCutAction(int operationCode, String actionDefinitionId) {
```
ShadelessFox:
```suggestion
    @NotNull
    private IAction createSmartCopyCutAction(int operationCode, @NotNull String actionDefinitionId) {
```

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorBase.java
```diff
@@ -770,9 +772,36 @@ protected void createActions() {
...
+
+        if (operationCode == ITextOperationTarget.CUT) {
+            action.setText("Cut");
+            action.setToolTipText("Cut selection or current line");
+            action.setImageDescriptor(sharedImages.getImageDescriptor(ISharedImages.IMG_TOOL_CUT));
+            action.setDisabledImageDescriptor(sharedImages.getImageDescriptor(ISharedImages.IMG_TOOL_CUT_DISABLED));
+        } else {
+            action.setText("Copy");
+            action.setToolTipText("Copy selection or current line");
+            action.setImageDescriptor(sharedImages.getImageDescriptor(ISharedImages.IMG_TOOL_COPY));
```
ShadelessFox: Don't use hardcoded values.

Instead, retrieve an existing action using `getAction(String id)` and copy over its attributes.

Here's how the original action is initialized in `AbstractTextEditor#createActions`:

```java
action = new TextOperationAction(EditorMessages.getBundleForConstructedKeys(), "Editor.Copy.", this, //$NON-NLS-1$
    ITextOperationTarget.COPY, true);
action.setHelpContextId(IAbstractTextEditorHelpContextIds.COPY_ACTION);
action.setActionDefinitionId(IWorkbenchCommandConstants.EDIT_COPY);
updateImages(action, IWorkbenchCommandConstants.EDIT_COPY);
setAction(ITextEditorActionConstants.COPY, action);
```

The `updateImages` method also sets these:
```java
actions.setImageDescriptor(...)
actions.setDisabledImageDescriptor(...)
actions.setHoverImageDescriptor(...)
```

You should copy them as well.

ShadelessFox (changes requested):

ShadelessFox:
While looking at `TextViewer`'s source code, I noticed that COPY/CUT actions delegate to this method: `TextViewer#doOperation` which calls `TextViewer#copyMarkedRegion` **if no selection provided**. So I think the best solution is to override `TextViewer#copyMarkedRegion` and do selection shenanigans there. This way, actions won't have to be overridden at all.

See `org.jkiss.dbeaver.ui.editors.sql.SQLEditorSourceViewer`

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorBase.java
ShadelessFox: Outdated; please see my comment below

jihoon0413:
@ShadelessFox Thank you for pointing that out. I realized that your suggestion aligns much better with the JFace architecture. I've updated the PR to use the override method (copyMarkedRegion), which makes the code significantly cleaner.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorSourceViewer.java
```diff
@@ -186,4 +187,49 @@ public void removeVerifyKeyListener(VerifyKeyListener listener) {
...
+            return isEditable();
+        }
+        return super.canDoOperation(operation);
+    }
+
+    @Override
+    protected void copyMarkedRegion(boolean delete) {
+        Point selection = getSelectedRange();
+
+        if (selection.y == 0) {
```
ShadelessFox: I suggest inverting this condition and adding an early `return;` so we can strip one level of indentation.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorSourceViewer.java
```diff
@@ -186,4 +187,49 @@ public void removeVerifyKeyListener(VerifyKeyListener listener) {
...
+
+                setSelectedRange(lineInfo.getOffset(), lineInfo.getLength() + delimiterLength);
+
+                if (delete) {
+                    getTextWidget().cut();
+                } else {
+                    getTextWidget().copy();
+                }
+            } catch (BadLocationException e) {
+                super.copyMarkedRegion(delete);
```
ShadelessFox: Please log the exception

jihoon0413:
@ShadelessFox I've applied your suggestions!

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditorSourceViewer.java
```diff
@@ -42,6 +43,8 @@
 
 public class SQLEditorSourceViewer extends ProjectionViewer {
 
+    static protected final Log log = Log.getLog(SQLEditorSourceViewer.class);
```
ShadelessFox: Wrong modifier order; it should be `private` as well
```suggestion
    private static final Log log = Log.getLog(SQLEditorSourceViewer.class);
```

jihoon0413:
@ShadelessFox Fixed — updated to private static final.
Sorry for the noise on this PR, and thanks for your patience and review.

ShadelessFox (approved):
Looking good :^)

IvanGorshechnikov (approved):

Matvey16:
Verified

PR #40159, opened 2026-01-21
dbeaver/pro#7736 Fix inserted rows not being committed

ShadelessFox:
Reproducible against any DB-DB transfer when "Selected rows only" option is checked


Matvey16:
Verified

PR #40162, opened 2026-01-21
#40151 Hide status bar when running multiple queries at once

### LINKED ISSUE #40151: Query results toolbar does not appear in results window until the window is clicked on
### Description

The toolbar that appears at the bottom of the query results window used to appear automatically once the query was finished running.  Now, you have to click into the window for the bar to appear.

### DBeaver Version

25.3.3.202601181752

### Operating System

Windows 11

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

ShadelessFox:


Matvey16:
Verified

PR #40170, opened 2026-01-22
dbeaver/dbeaver#11453 search path fix

### LINKED ISSUE #11453: [postgresql] How to not change search_path when selecting database

serge-rider:
We won't call `set search_path` if it is not needed (selected schema is already there).

Also now we show default schema for the selected database so users won't need to guess what to choose:

image

Also please test on Redshift


uslss:
verified

checked: PG and Redshift

PR #40186, opened 2026-01-23
7734 core cli ability to specify network handlers ssh

alexander-skoblikov:
closes https://github.com/dbeaver/pro/issues/7734


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNBrowseSettings.java
```diff
@@ -40,4 +40,15 @@ public interface DBNBrowseSettings {
      */
     boolean isUserSettings();
 
+    void setShowSystemObjects(boolean showSystemObjects);
```
serge-rider, line 51: Why setters here? Just cast to DataSourceNavigatorSettings

serge-rider (changes requested):

PR #40191, opened 2026-01-24
dbeaver/pro#8063 fix bedrock issues

Destrolaric:
Closes dbeaver/pro#8063

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIMessage.java
+++ b/plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIMessage.java
@@ -34,7 +34,7 @@ public class AIMessage {
     @Nullable
     private final String displayMessage;
     @Nullable
-    private final String functionCallID;
+    private final String customResultInfo;
     @NotNull
     private final LocalDateTime time;
     @Nullable
@@ -59,7 +59,7 @@ public class AIMessage {
         this.meta = meta;
         this.functionCall = null;
         this.functionResult = null;
-        this.functionCallID = functionCallID;
+        this.customResultInfo = functionCallID;
     }
     /**
      * Creates AI message
@@ -78,7 +78,7 @@ public class AIMessage {
         this.meta = meta;
         this.functionCall = null;
         this.functionResult = null;
-        this.functionCallID = null;
+        this.customResultInfo = null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIMessage.java
```diff
@@ -33,6 +33,8 @@ public class AIMessage {
     private final String content;
     @Nullable
     private final String displayMessage;
+    @Nullable
+    private final String functionCallID;
```
serge-rider: Lets name more generic (e.g. customResultInfo).
And maybe move into AIFunctionResult?

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/registry/AIEngineDescriptor.java
```diff
@@ -36,6 +36,7 @@ public class AIEngineDescriptor extends AbstractDescriptor {
     private final ObjectType propertiesType;
     private final boolean supportsFunctions;
     private final boolean providesMetadata;
+    private final boolean passFunctionsCallBackToAI;
```
serge-rider: I think we don't need it in descriptor.
It is better to use market interface + move it to PRO.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIMessage.java
```diff
@@ -42,6 +44,23 @@ public class AIMessage {
     @Nullable
     private final AIMessageMeta meta;
 
+    public AIMessage(
```
IvanGorshechnikov, line 47: I don't understand what this new constructor is for. Before, there were two: one for regular messages and another for functions. So what is this one for? It seems to be called in the `functionMessage` method, but we also have the `functionCall` method, and it's not clear what the difference is between them. This is very confusing for me.

Destrolaric: Message with the result of the function call. Contains the name of the function it was produced from

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIMessage.java
```diff
@@ -148,6 +180,11 @@ public String getContent() {
         return content;
     }
 
+    @Nullable
+    public String getToolUseID() {
+        return customResultInfo;
```
ShadelessFox, line 185: Why is the field called `customResultInfo` but this getter returns `toolUseID`?

PR #40193, opened 2026-01-25
#40144 Fix hashcode display when table name is empty

### LINKED ISSUE #40144: Object hashcode is dipsplayed when table name is empty
### Description

When i try to create a new table and delete the whole text in the `Table Name` input box the hashcode of the object is displayed.

### DBeaver Version

Community Edition 25.3.4

### Operating System

Windows 11 x64

### Database and driver

Database: SQLite
Driver: SQLite JDBC driver

### Steps to reproduce

1. Open a connection
2. Right click on Tables or `Alt + Insert`
3. Delete the full text from `Table Name`

### Additional context

Image
Image

dahyeong-yun:
### Changes

Improved the display of unnamed objects (tables, indexes, etc.) which were previously showing internal hashcodes (e.g., `ClassName@Hashcode`). This behavior was confusing to users as it exposed internal implementation details.

*   **Removed Hashcode Exposure:** Prevents the default `object.toString()` implementation from leaking into the UI.
*   **Applied Empty String (`""`):** Replaced the hashcode with an empty string for unnamed objects. This minimizes visual noise and avoids exposing technical data.
*   **Minimized Overhead:** By using an empty string instead of a specific label (like `<unnamed>`), we avoid the need for new localization (i18n) resources and maintenance across multiple languages.

### Keywords/Ticket

Closes #40144

### Screenshots
Screenshot 2026-01-25 at 8 20 20 AM

### Note to Reviewers

I implemented this using an empty string (`""`) rather than a placeholder label like `<unnamed>`.

**Reasoning:**
1.  **Consistency:** Currently, different objects have inconsistent behaviors for empty names (e.g., Columns might show `{Table}.` while others show hashcodes). An empty string seemed the most neutral approach.
2.  **No i18n Overhead:** Introducing a new label text would require translation updates for all supported languages.

However, if there is an established project convention for handling empty names, or if you prefer a standardized label (such as <unnamed>), I am more than happy to adjust the implementation to align with DBeaver's standards. Please let me know your thoughts during the review!


ShadelessFox (approved):
Thanks for the contribution!

I think it looks okay. I'll have to discuss it with my team next week, but I give my initial approval.

E1izabeth (approved):

MashaKorax:
verified
DBeaver ce win

PR #40198, opened 2026-01-26
dbeaver/pro#7910: fix clear old secrets

SolovevAnton:
to delete


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceDescriptor.java
```diff
@@ -1003,6 +1003,12 @@ public void resolveSecrets(DBSSecretController secretController) throws DBExcept
         secretsResolved = true;
     }
 
+    public void clearOldSecrets(DBSSecretController secretController) throws DBException {
```
ShadelessFox:
```suggestion
    public void clearOldSecrets(@NotNull DBSSecretController secretController) throws DBException {
```

And please add Javadoc describing what it does. It feels like it should be called just `clearSecrets`; there are no "old" secrets.

ShadelessFox (dismissed):

SolovevAnton:
all moved to pro

PR #40201, opened 2026-01-26
dbeaver/dbeaver#11453 Verify that database/schema selector dialog has…

### LINKED ISSUE #11453: [postgresql] How to not change search_path when selecting database

serge-rider:
… valid selection


uslss:
issue with two bolded db/schema (chosen and default is still reproducible
image

uslss:
default (not current) schema is always selected in schema selector (but current database is always selected in database selector). Probably, also current schema should be selected on dialog opening

image

uslss:
verified (CE, win 11)
image

PR #40208, opened 2026-01-27
dbeaver/pro#8112 Hide "Use fetched rows" if no fetched data available

ShadelessFox:
|Export from table|Export from result set|
|---|---|
|image|image|


uslss:
verified (Navigator, tasks, data editor, data editor in sql editor)

PR #40216, opened 2026-01-28
dbeaver/pro#7913 Fix FileSystem lifecycle for CLI file options

vkirst:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/fs/DBFPath.java
```diff
@@ -0,0 +1,58 @@
...
+package org.jkiss.dbeaver.model.fs;
+
+import org.jkiss.code.NotNull;
+import org.jkiss.code.Nullable;
+
+import java.io.IOException;
+import java.nio.file.FileSystem;
+import java.nio.file.Path;
+
+public final class DBFPath implements AutoCloseable {
```
ShadelessFox, line 24: This is so weird... NIO's `Path` already knows about its filesystem. Why are we doing this?

alexander-skoblikov (approved):

PR #40223, opened 2026-01-29
dbeaver/dbeaver#40082 Rework of the selection handling of tab

### LINKED ISSUE #40082: "Pin tab" shortcut logic should always pin or unpin the currently active tab, not some other

HellAmbro:
This PR improves the management of results tabs.
For example, it resolves the error where the keyboard shortcut operates on an unselected tab; the context menu is now consistent with the selected tab.
Closes #40082


ShadelessFox (commented):
Looking at the code, it seems like `activeResultsTab` is just redundant. We can retrieve the active tab using `resultTabs.getSelection()`, and it's not even performance-critical. After it's done, it's no longer required to track the active and selected result tabs.

PR #40256, opened 2026-02-02
dbeaver/dbeaver#40253 Fix local file handling

### LINKED ISSUE #40253: [Open File] Change file path to temp when open saved file.
### Description

Hi,
When I open file by menu: File -> OpenFile e.q. "C:\sql_scripts\test.sql" on the SQL file bar I see other file path:
Image

This is a serious problem!
I need to remember to save this file back to its original location, otherwise all the SQL code disappears when I close it!

### DBeaver Version

CommunityuEdition v25.3.4.202602011558

### Operating System

Win 11 Pro

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

IvanGorshechnikov:
Closes https://github.com/dbeaver/dbeaver/issues/40253

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/ShellUtils.java
+++ b/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/ShellUtils.java
@@ -22,6 +22,7 @@ import org.eclipse.ui.internal.ide.IDEInternalPreferences;
 import org.eclipse.ui.internal.ide.IDEWorkbenchPlugin;
 import org.jkiss.code.NotNull;
 import org.jkiss.dbeaver.Log;
+import org.jkiss.dbeaver.runtime.DBWorkbench;
 import org.jkiss.dbeaver.utils.RuntimeUtils;
 import org.jkiss.utils.IOUtils;
 
@@ -52,13 +53,21 @@ public final class ShellUtils {
     public static boolean openExternalFile(@NotNull Path path) {
         try {
             if (RuntimeUtils.isMacOS()) {
-                executeWithReturnCodeCheck("open", path.toAbsolutePath().toString());
+                try {
+                    // For known associations
+                    executeWithReturnCodeCheck("open", path.toAbsolutePath().toString());
+                } catch (IOException e) {
+                    // For unknown associations
+                    executeWithReturnCodeCheck("open", "-a", "Finder.app", path.toAbsolutePath().toString());
+                }
                 return true;
             } else if (RuntimeUtils.isLinux()) {
                 executeWithReturnCodeCheck("xdg-open", path.toAbsolutePath().toString());
                 return true;
             }
         } catch (IOException | InterruptedException e) {
+            DBWorkbench.getPlatformUI()
+                .showError("Unable to open external program", "Unable to open external program in a platform-specific way", e);
             log.debug("Unable to open external program in a platform-specific way: " + e.getMessage());
         }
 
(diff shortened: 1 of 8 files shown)


plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/ShellUtils.java
```diff
@@ -52,7 +52,7 @@ public static boolean launchProgram(@NotNull String path) {
     public static boolean openExternalFile(@NotNull Path path) {
         try {
             if (RuntimeUtils.isMacOS()) {
-                executeWithReturnCodeCheck("open", "-a", "Finder.app", path.toAbsolutePath().toString());
+                executeWithReturnCodeCheck("open", path.toAbsolutePath().toString());
```
ShadelessFox: Why this change, again?

IvanGorshechnikov: Because it breaks opening files in the external apps on Mac, I tested this method on other releases, and it didn't work too

Destrolaric (dismissed):
Can't verify information about MacOS, so I rely on approve from @ShadelessFox

ShadelessFox:
We figured out what's wrong with the `open -a Finder.app <path>` approach: it works for unknown file associations, but doesn't for known ones. On the other hand, the `open <path>` works for known associations, but not for unknown ones. So we decided on the following flow:
1. Try `open -a Finder.app <path>`
2. If it returns a non-zero code, try `open <path>`
3. If it returns a non-zero code, show an error dialog instead of silently failing

IvanGorshechnikov:
I have to change the order, because `open -a Finder.app <path>` returns 0 code when it fails with a system error. So we can't do anything for that case. But `open <path>` returns a non-zero code when we try to open unknown file association. So this is the best way:

1. Try `open <path>`
2. If it returns a non-zero code, try `open -a Finder.app <path>`
3. If it returns a non-zero code, show an error dialog instead of silently failing

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/ShellUtils.java
```diff
@@ -52,13 +53,21 @@ public static boolean launchProgram(@NotNull String path) {
     public static boolean openExternalFile(@NotNull Path path) {
         try {
             if (RuntimeUtils.isMacOS()) {
-                executeWithReturnCodeCheck("open", "-a", "Finder.app", path.toAbsolutePath().toString());
+                try {
+                    // For known associations
```
ShadelessFox: > In recent versions of macOS, `open -a Finder.app <path>` no longer works for known file associations, so we have to rely on `open <path>`, but we can't use it for unknown associations either, as it doesn't know how to deal with them. So probe the latter first, and fall back to the former. If both fail, show an error.

ShadelessFox (dismissed):

Destrolaric (dismissed):

uslss:
verified (DBeaver UE, win)

uslss:
changes aren't saved for scripts opened from cloud storage. No error is shown.

ShadelessFox (dismissed):

serge-rider:
I generally don't like the approach (passing source node all over thru all abstractions).

IvanGorshechnikov:
I agree, but we don't have a better solution

Matvey16:
Incorrect path for db files opened from cloud explorer

dbeaver_910V7oMFbA

Matvey16:
Verified

DB CE, UE Linux, Windows, MacOS

PR #40263, opened 2026-02-03
dbeaver/pro#8128 fix anthropic function calling

Destrolaric:
Closes dbeaver/pro#8128

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIMessageMeta.java
+++ b/plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIMessageMeta.java
@@ -25,7 +25,7 @@ public record AIMessageMeta(
     @NotNull String type,
     @NotNull String engineId,
     @Nullable String modelId,
-    @NotNull AIUsage usage,
+    @Nullable AIUsage usage,
     @NotNull Duration timeSpent,
     int systemPromptLength
 ) {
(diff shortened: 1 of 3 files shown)


IvanGorshechnikov:
Copilot has good comments

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AIFunctionCall.java
```diff
@@ -62,6 +74,15 @@ public void setArguments(@NotNull Map<String, Object> arguments) {
         this.arguments = arguments;
     }
 
+    @Nullable
+    public Map<String, String> getAdditionalProperties() {
+        return messageMetadata;
+    }
+
+    public void setAdditionalProperties(Map<String, String> additionalProperties) {
```
ShadelessFox: Missing nullability annotations. I would also have an explicit copy of the passed map using `Map.copyOf`.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIMessageMeta.java
```diff
@@ -25,7 +25,7 @@ public record AIMessageMeta(
     @NotNull String type,
     @NotNull String engineId,
     @Nullable String modelId,
-    @NotNull AIUsage usage,
+    @Nullable AIUsage usage,
```
ShadelessFox, line 28: We actually have an ongoing discussion regarding this issue: https://dbeaver.slack.com/archives/CBZ75PGD6/p1770647413931889

I think you should revert this change for now.

ShadelessFox (approved):

PR #40277, opened 2026-02-04
#39887 Show multiple foreign keys as a single action

### LINKED ISSUE #39887: Multiple virtual foreign keys on the same column confuse the UI
### Description

Some columns in my DB can reference different foreign tables depending on some other column.

When I define a virtual foreign key for these, the UI shows multiple “navigate to” buttons, one per foreign key, but:
- They all have the same tooltip (“navigate to community” in this case)
- Only the rightmost one does something on click (in my case, it navigates to the `community` table)

Image

### DBeaver Version

25.2.5.202511161745

### Operating System

Windows 10.0.26200 Build 26200

### Database and driver

- Database : SQLite
- Driver : SQLite JDBC driver 3.50.3.0

### Steps to reproduce

1. Have a SQLite database
2.  Have a table `a` with an `id` column, a table `b` with an `id` column, and a table `reference` with a `ref_id` column
3. In DBeaver, define two virtual foreign keys for `reference.ref_id`, pointing respectively to `a.id` and `b.id`
4. Attempt to use the two “navigate to” widgets in the view: only one does anything, and both have the same exact tooltip
5. Ideally, each widget maps to a different table, both are interactive and both navigate to the adequate one

### Additional context

_No response_

ShadelessFox:
java_MXAj2ng6NQ

Keyboard buttons 1-9 can be used to quickly select the corresponding menu item.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/hints/DBDValueHint.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/hints/DBDValueHint.java
@@ -1,6 +1,6 @@
 /*
  * DBeaver - Universal Database Manager
- * Copyright (C) 2010-2024 DBeaver Corp and others
+ * Copyright (C) 2010-2026 DBeaver Corp and others
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


Destrolaric (approved):

IvanGorshechnikov (approved):

Matvey16:
I feel like since this feature replaces the basic navigation link with mouse click, it should also replace this item in the context menu


https://github.com/user-attachments/assets/4fef64dd-d4f0-4118-bbe9-6a28f3846aee

As well as an Alt+Space shortcut

Matvey16:
Verified

DB CE, EE Windows

PR #40289, opened 2026-02-05
dbeaver/pro#8047 exposed resultset SQL generators to CB

Nexus6v2:
Closes dbeaver/pro#8047


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/DBDDataProvider.java
```diff
@@ -0,0 +1,50 @@
...
+import org.jkiss.dbeaver.model.data.DBDValueRow;
+import org.jkiss.dbeaver.model.struct.DBSEntity;
+
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * Represents the context for SQL generation from a result set.
+ */
+public interface DBDDataProvider {
```
serge-rider: move in model.data

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -473,6 +473,36 @@ public void shellDeiconified(ShellEvent e) {
         });
     }
 
+    @NotNull
```
serge-rider, line 476: IT looks like you are trying to expose DBDResultSetModel here.
Maybe we just add method getModel to DataProvider?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
Nexus6v2: I had this idea, but it would be an excessive method.
We would still need all the accessors for Web because the logic is different there (We don't have a similar model class to extract from).

I think it's better to have the same set of methods, even when one implementation is overly simplistic like here.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/DBDDataProvider.java
Nexus6v2: Done

PR #40305, opened 2026-02-09
dbeaver/pro#7820 AI functions enabled by default

IvanGorshechnikov:
Closes https://github.com/dbeaver/pro/issues/7820


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/registry/AISettingsManager.java
```diff
@@ -162,10 +168,22 @@ private static AISettings loadSettingsFromConfig() {
...
+                AIFunctionCategoryDescriptor category = entry.getKey();
+                List<AIFunctionDescriptor> functions = entry.getValue();
+
+                if (category.isEnabledByDefault() && !settings.isCategoryInitialized(category.getId())) {
+                    settings.enableFunctionCategory(category.getId());
+                    for (AIFunctionDescriptor function : functions) {
+                        settings.enableFunction(function.getId());
+                    }
+                    settings.markCategoryAsInitialized(category.getId());
+                }
```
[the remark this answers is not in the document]
IvanGorshechnikov: This is what we want

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AISettings.java
```diff
@@ -141,6 +142,26 @@ public void disableFunctionCategory(@NotNull String category) {
...
+        }
+    }
+
+    public void markCategoryAsInitialized(@NotNull String categoryId) {
+        this.initializedDefaultCategories.add(categoryId);
+    }
+
+    public boolean isCategoryInitialized(@NotNull String categoryId) {
+        return initializedDefaultCategories.contains(categoryId);
+    }
```
ShadelessFox: This new property will live with us forever, and its only purpose is to mimic `enabledFunctionCategories` if it were a clean installation. We can simply rename the existing `enabledFunctionCategories` to something else, as Serge suggested.

IvanGorshechnikov: I don't like the renaming thing. It's okay now, but we'll definitely want to enable readSampleData by default later, and then we'll have this problem where we can't change the name.
And basically, I'm now enabling it through a property so that I don't have to do anything later.

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIFunctionSettings.java
```diff
@@ -0,0 +1,116 @@
...
+ * Stores function-related configuration: enabled/disabled states,
+ * initialized categories tracking.
+ */
+public class AIFunctionSettings {
+    private boolean functionsEnabled = true;
+    private final Set<String> enabledFunctionCategories = new LinkedHashSet<>();
+    private final Set<String> enabledFunctions = new LinkedHashSet<>();
+    private final Set<String> initializedDefaultCategories = new LinkedHashSet<>();
+
+    public AIFunctionSettings() {
```
ShadelessFox:
```suggestion
    AIFunctionSettings() {
```

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIFunctionSettings.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+/**
+ * AI function settings.
+ * Stores function-related configuration: enabled/disabled states,
+ * initialized categories tracking.
+ */
+public class AIFunctionSettings {
```
ShadelessFox:
```suggestion
public final class AIFunctionSettings {
```

ShadelessFox (approved):

PR #40317, opened 2026-02-10
dbeaver/pro#7807 check platform initialized

alexander-skoblikov:
closes dbeaver/pro#7807


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/core/DBeaverActivator.java
```diff
@@ -108,7 +108,9 @@ public void stop(BundleContext context)
     }
 
     private void shutdownUI() {
-        DesktopUI.disposeUI();
+        if (DesktopPlatform.instance != null) {
```
serge-rider, line 111: But this is DBeaver-specific, how it is related to dbvr?

PR #40320, opened 2026-02-10
#39027 Add back P2 Sites preference page

### LINKED ISSUE #39027: BUG - Manage install sources doesn't actually work
### Description

Clicking the manage button in the install software context does not actually open the source manager where you can edit or remove sources. See below.

Image

### DBeaver Version

25.1.5.202508171444

### Operating System

Windows 11

### Database and driver

_No response_

### Steps to reproduce

1. Open the install software menu
2. Clicking the manage button 
3. It doesn't open the menu for editing, removing, adding sources.

### Additional context

_No response_

ShadelessFox:
I couldn't find an easy way to conditionally reveal an externally defined preference page without modifying the registry, so I followed a rather simple approach of putting **Sites** under **User Interface** and removing the two other unrelated pages:

dbeaver_gUluUNUZnY

I think it will do for now.


Matvey16:
Verified

DB CE, EE Windows

PR #40330, opened 2026-02-11
dbeaver/pro#7713 Resolve linked resources

ShadelessFox:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNProject.java
```diff
@@ -195,6 +195,20 @@ public DBNNode refreshNode(@NotNull DBRProgressMonitor monitor, @Nullable Object
         return this;
     }
 
+    @Nullable
+    public DBNNode findResource(@NotNull DBRProgressMonitor monitor, @NotNull String path) throws DBException {
```
serge-rider: Why it is called findResource? It has to be findNodeByPath.
btw we already have getNodeByPath in DBNModel

ShadelessFox: > Why it is called findResource? It has to be findNodeByPath.

It mimics the other method next to it

IvanGorshechnikov (approved):

PR #40337, opened 2026-02-12
dbeaver/pro#8126 create temp folders in cli

alexander-skoblikov:
closes dbeaver/pro#8126


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/BaseLocalPlatform.java
```diff
@@ -0,0 +1,96 @@
...
+import org.jkiss.utils.CommonUtils;
+import org.jkiss.utils.StandardConstants;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+public abstract class BaseLocalPlatform extends BasePlatformImpl {
```
serge-rider: Why new class?
Can we move it to BasePlatformImpl?

serge-rider (changes requested):

Issue #8746, opened 2020-05-19
Fails to start with error "Java was started but returned exit code=13" on Windows (Zip install)


Chealer:
#### System information: 
- Windows 10
- DBeaver 7.0.4
- No additional extensions

I have Java 8 and JDK 13 installed.

#### Describe the problem you're observing:
DBeaver fails to initially start on this machine, showing instead the following dialog:
![image](https://user-images.githubusercontent.com/486141/82365953-69341880-99df-11ea-887f-a962fe29a5d5.png)

#### Steps to reproduce, if exist:
I do not know what steps are needed. What I did:
1. Download the zip file from https://github.com/dbeaver/dbeaver/releases/download/7.0.4/dbeaver-ce-7.0.4-win32.win32.x86_64.zip
1. Extract it (to C:\Program Files (x86)\dbeaver\)
1. Run dbeaver.exe

#### Include any warning/errors/backtraces from the logs
As this is an initial install, the log directory was not yet created.

---
Issue #5678 suggests this problem is not new.

vl2376:
Same issue here only its exit code = 1 for me

rafliher:
so no solution yet?

meriturva:
same here with version 7.1.1. Version 7.0.4 is ok!

vl2376:
So if you download the installation folder as a zip you'll see a dbeaver.ini file in which changing the path to Java and JDK installations may help you. That was the fix for me.

rafliher:
all i do is delete the java path which leads to oracle folder and it works. for now.

meriturva:
Fixed installing Java 64bit

aniston:
> Fixed installing Java 64bit

Hi, I would like to confirm that having only Java JRE x86 or i586 was not sufficient on my W7x64 OS so I had to install a JRE x64 as well and that solved the startup problem with the .zip package.

Chealer:
Why is this ticket closed and labelled as a question?

Chealer:
@serge-rider , you marked this report as a question and removed its _bug_ label on 2020-07-24. Can you explain?

serge-rider:
@Chealer Because @meriturva's reply answers the question about DBeaver start problems. DBeaver doesn't support 32-bit Java anymore.
To solve this problem just install 64-bit Java.

Chealer:
What question are you talking about @serge-rider ? Do you know what an ITS is? This ticket is a report of a bug in DBeaver, it is not a support request.

[...]

PR #40341, opened 2026-02-12
dbeaver/pro#7791 custom exit code

alexander-skoblikov:
closes dbeaver/pro#7791

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBConstants.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBConstants.java
@@ -173,9 +173,4 @@ public class DBConstants {
     public static final String CERTIFICATE_STORAGE_FOLDER = "security";
 
     public static final String PROP_RESOURCE_DEFAULT_DATASOURCE = "default-datasource"; //$NON-NLS-1$
-
-
-    /// //////////////////////////////
-    // Eclipse
-    public static final String ECLIPSE_EXIT_DATA = "eclipse.exitdata";
 }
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -230,11 +232,20 @@ public Object start(IApplicationContext context) {
...
             } catch (Exception e) {
                 log.error("Error processing command line parameters", e);
-                return IApplication.EXIT_OK;
+                shutdown = true;
+                exitCode = CLIConstants.EXIT_CODE_ERROR;
+            }
+            if (shutdown) {
+                // hide standard Eclipse exit message if exit code is not OK (otherwise it may be confusing)
+                if (!EXIT_OK.equals(exitCode)) {
+                    System.setProperty(DBConstants.ECLIPSE_EXIT_DATA, "");
```
serge-rider: Maybe it is better to set this propertyy always?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBConstants.java
```diff
@@ -173,4 +173,9 @@ public class DBConstants {
     public static final String CERTIFICATE_STORAGE_FOLDER = "security";
 
     public static final String PROP_RESOURCE_DEFAULT_DATASOURCE = "default-datasource"; //$NON-NLS-1$
+
+
+    /// //////////////////////////////
+    // Eclipse
+    public static final String ECLIPSE_EXIT_DATA = "eclipse.exitdata";
```
serge-rider: Let's move to BaseApplicationImpl

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
alexander-skoblikov: not sure because this problem sometimes occurs for DBeaver users

see https://github.com/dbeaver/dbeaver/issues/8746

PR #40347, opened 2026-02-13
dbeaver/pro#7233 fix icons for distributed driver lib folders

yagudin10:


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/driver/DriverLibraryLocal.java
```diff
@@ -212,6 +212,12 @@ public String getId() {
     @NotNull
     @Override
     public DBIcon getIcon() {
+        if (DBWorkbench.isDistributed()) {
```
Nexus6v2: If think you can join conditions here `DBWorkbench.isDistributed() && isDistributedDriverLibraryFolder()`

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/driver/DriverLibraryLocal.java
```diff
@@ -225,6 +231,20 @@ public DBIcon getIcon() {
         }
     }
 
+    private boolean isDistributedDriverLibraryFolder() {
+        if (driver == null) {
+            return false;
+        }
+        List<DriverFileInfo> fileInfos = driver.getDefaultDriverLoader().getResolvedFiles().get(this);
+        if (fileInfos.size() != 1) {
```
Nexus6v2: I assume the idea is to check if list contains multiple values. 
What if size is 0? Can it happen?
Maybe `fileInfos.size() > 1` is better?

Nexus6v2 (dismissed):

PR #40359, opened 2026-02-16
dbeaver/dbeaver#40353 Fix active schema selection

### LINKED ISSUE #40353: Wrong schema selected by default
### Description

The bug is in database/schema selector in SQL Editor.
When schema name matches the database name then we switch to it instead of the default schema.

Reproducible in PostgreSQL but not only.

Originated from https://github.com/dbeaver/dbeaver/issues/11453#issuecomment-3897818571

serge-rider:


uslss:
cannot change database for sql server

https://github.com/user-attachments/assets/64230885-0171-48d9-b1c4-b1beeb29d6bd

PR #40360, opened 2026-02-16
dbeaver/dbeaver#38299 Multiple policy roles support

### LINKED ISSUE #38299: Bug: "Role" field in table policy view shows only a single role, even when multiple roles are assigned
### Description

In the table Policies view within DBeaver, the Role column displays only a single role, even when the policy is configured with multiple roles.

The same behavior occurs in the Policy Properties dialog: the Role field shows only one role.

However, when querying the database directly with:

`SELECT * FROM pg_policies;`

the role column correctly returns an array of roles, even when there is only one role (as an array of size 1). If multiple roles are assigned to the policy, all of them are correctly listed in this array.

Currently, DBeaver's interface appears to display only a single role, not accounting for the possibility that the policy applies to multiple roles.

**Expected behavior:**

Both the Policies list and the Policy Properties dialog should display all roles associated with the policy. These could be presented, for example, as a comma-separated list or using another suitable format to indicate that multiple roles are involved.

### DBeaver Version

Community Edition 25.1.0.202506011722

### Operating System

Windows 11 24H2

### Database and driver

Database: PostgreSQL 13.3 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-16), 64-bit

### Steps to reproduce

1. Create a policy on a PostgreSQL table with more than one role.
2. Open DBeaver and navigate to the table's Policies tab.
3. Observe that the Role column displays only a single role.
4. Open the Policy Properties dialog for the policy.
5. Note that the Role field also shows only one role.
6. Execute SELECT * FROM pg_policies in the SQL Editor.
7. Confirm that the role column contains an array with all assigned roles.

### Additional context

This behavior may lead to confusion, as users might incorrectly assume that the policy applies to only one role when, in fact, it applies to multiple. PostgreSQL represents the role attribute as an array of roles, as documented in its system catalogs.

serge-rider:


Matvey16:
Verified

DB CE Windows

PR #40369, opened 2026-02-17
dbeaver/pro#8023 added nullable exception message handling

Nexus6v2:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -1015,6 +1016,12 @@ public static String makeStandardErrorMessage(@NotNull Throwable error) {
             return "Class not found: " + cnfe.getMessage();
         } else if (error instanceof NoClassDefFoundError ncdf) {
             return "Class definition not found: " + ncdf.getMessage();
+        } else if (error instanceof DBCConnectException dbce) {
```
serge-rider, line 1019: This is our own exception. Just fix the error message in the place where it is thrown.

serge-rider (changes requested):

PR #40373, opened 2026-02-17
dbeaver/dbeaver#40342: fix bug with new column in record mode

### LINKED ISSUE #40342: Can't add a new row in multi-record mode
### Description

Clicking a button to add a new row in multi-record mode just opens another existing record instead of the new one. This happens only if the rows have other rows between them, so in some cases this will not reproduce (e.g. adding a new row when the last of the selected records is in focus)

### DBeaver Version

25.3.5

### Operating System

_No response_

### Database and driver

Any

### Steps to reproduce

1. Select a few rows that are not near each other
2. Open record mode
3. Click `add row`

**Expected:** A new row is added
**Actual:** Existing row is added to the view

### Additional context

https://github.com/user-attachments/assets/6390f5da-5cc6-4280-877d-b6003cf2671b

SolovevAnton:
closes dbeaver/dbeaver#40342: fixes bug with new columns creation


IvanGorshechnikov (approved):

Matvey16:
Verified

DB CE Windows

PR #40381, opened 2026-02-18
dbeaver/pro#8164 Fix NPE for Periods in DB2

IvanGorshechnikov:
Closes https://github.com/dbeaver/pro/issues/8164


Matvey16:
Verified

DB CE Windows

PR #40391, opened 2026-02-19
dbeaver/pro#4050 Fix XML export element name escaping

ivgag:
Prepend underscore to element names that do not start with an alphabetic character.


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterXML.java
```diff
@@ -170,6 +170,12 @@ else if (buffer[i] == '>') {
     }
 
     private String escapeXmlElementName(String name) {
-        return name.replaceAll("[^\\p{Alpha}\\p{Digit}]+","_");
+        String escapedName = name.replaceAll("[^\\p{Alpha}\\p{Digit}]+", "_");
+        char firstCharacter = escapedName.charAt(0);
```
ShadelessFox, line 174: This will fail on empty strings. Also, missing nullability annotations

PR #40392, opened 2026-02-19
dbeaver/pro#8154 Reflect modifications to SSH jump servers

ShadelessFox:
When a jump server configuration is modified, it should now be reflected in the table immediately.


IvanGorshechnikov (dismissed):

Matvey16:
Clicking on an already focused entry in the list will make unexpected changes to setting (cnage authentication method, change passwords etc) https://github.com/dbeaver/pro/issues/8154#issuecomment-3934182609

Matvey16:
Verified

DB CE Windows

PR #40400, opened 2026-02-20
dbeaver/pro#8256 Fix JSCH key conversion resulting in NPE

ShadelessFox:


E1izabeth (approved):

Matvey16:
Verified

DB CE Windows

PR #40401, opened 2026-02-20
dbeaver/pro#8324 Return tool result for OpenAI

IvanGorshechnikov:
Closes https://github.com/dbeaver/pro/issues/8324


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/dto/OAIMessage.java
```diff
@@ -38,12 +44,28 @@ public class OAIMessage {
...
     public List<OAIMessageContent> content;
 
     public OAIMessage() {
     }
 
     public OAIMessage(@NotNull AIMessage msg) {
+        this(msg, null);
+    }
+
+    public OAIMessage(@NotNull AIMessage msg, @Nullable String toolCallId) {
```
ShadelessFox: Can we please start making static factories instead? We're polluting the class with non-final fields that are actually supposed to be final, just because we're reassigning them in one of the constructors.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/dto/OAIMessage.java
```diff
@@ -53,6 +75,19 @@ public OAIMessage(@NotNull AIMessage msg) {
         content = List.of(new OAIMessageContent(input, msg.getContent()));
     }
 
+    private void mapFunctionCall(@NotNull AIFunctionCall functionCall) {
```
ShadelessFox: Same here. It shouldn't be an instance method, really

ShadelessFox:
@IvanGorshechnikov there are conflicts

ShadelessFox (approved):

PR #40403, opened 2026-02-20
dbeaver/pro#8450 Unload editor on error instead of closing it

ShadelessFox:
https://github.com/user-attachments/assets/fe5caddb-bd21-4c36-8c3c-b116b11c7ab1


uslss:
The OK button as an answer to "What do you want to do?" looks strange. Probably Close Button should be used instead 
image

ShadelessFox:
Yes, probably. Or even "Cancel", as one might assume it's addressing the editor, not the dialog.

ShadelessFox:
Replaced with "Abort"

uslss:
verified (win11, CE)
image

PR #40414, opened 2026-02-23
dbeaver/dbeaver#40062 Fix autocomplete for fabric

### LINKED ISSUE #40062: Table name auto-completion does not work (Azure SQL Fabric)
### Description

DBeaver currently does not support table name suggestions (auto-completion) when querying tables using schema-qualified names.

When a schema is specified (for example: schema_name.table_name), DBeaver does not suggest or autocomplete the available table names under that schema.

This makes writing SQL queries slower and more error-prone, especially in environments with many schemas and tables.

### DBeaver Version

Community edition 25.3.2

### Operating System

Window 11

### Database and driver

Azure SQL Server (Using connection string from Fabric Lakehouse)

### Steps to reproduce

_No response_

### Additional context

_No response_

IvanGorshechnikov:
Close https://github.com/dbeaver/dbeaver/issues/40062


E1izabeth (approved):

uslss:
The issue is still reproducible for fabric. table name suggestion is shown only for database_name.schema_name.table
Also, database/schema selector isn't shown in toolbar. probably separate ticket is needed?
image

uslss:
image

PR #40420, opened 2026-02-23
Dbeaver/pro#7836 function statistics

Destrolaric:
Closes Dbeaver/pro#7836


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIBaseFeatures.java
```diff
@@ -16,13 +16,44 @@
...
+
+
+    DBRFeature AI_CHAT_FUNCTION_CALL = DBRFeature.createFeature(CATEGORY_AI_CALLS, "AI Chat function call");
+
+    @NotNull
+    static Map<String, Object> buildFeatureParameters(
+        @Nullable DBPDataSourceContainer container,
+        @NotNull Map<String, Object> additionalInfo
+    ) {
+        HashMap<String, Object> featureInfoMap = new HashMap<>(Map.of(
```
serge-rider: Use LinkedHashMap as order matters

serge-rider: Use Map for var

serge-rider: Save AIUtils.getActiveEngineDescriptor()  to variable

serge-rider (changes requested):

PR #40430, opened 2026-02-24
dbeaver/dbeaver#40297 Add supporting multi-object DROP statements for PostgreSQL

### LINKED ISSUE #40297: Deleting multiple tables/views in DBeaver should use "DROP TABLE/VIEW tableA, tableB, tableC" to allow dropping interdependent objects
When selecting multiple tables of views in DBeaver and deleting, them, it creates a list of individual `DROP TABLE tableA; DROP TABLE tableB;` queries. This can easily lead to errors due to foreign keys or views using other views when cascade isn't used.
 
I would like DBeaver to generate a single query which lists the selected objects in a comma-separated way: `DROP TABLE tableA, tableB;` - At least for PostgreSQL (i'm unsure about other DBMS), the DBMS understands that I'm dropping both tables at the same time, so foreign keys between them do not cause a error due to dependent objects.

Enabling CASCADE: This is the simple solution which exists currently, but it can lead to mistakes that would be prevented by the suggested solution. For exaple, if i'm dropping a table with CASCADE, it will also drop the corresponding views. This is implicit and can lead to views that were not considered at the time mistakenly getting dropped if they happen to depend on the table. With `DROP TABLE tableA, tableB` this case would still lead to an error, giving the chance to consider whether the view is okay to be dropped. This is espcially relevant if there are a lot of views, and a view being dropped by a CASCADE accidentally is not immediatly obvious.  

**Additional context**
I haven't checked if this syntax and behaviour exists in other DBMS besides PostgreSQL

statxc:
### Problem

Deleting multiple tables/views generates individual `DROP TABLE` statements, which fails when objects have inter-dependencies (foreign keys, view references) unless CASCADE is used. CASCADE can silently drop unrelated dependent objects, leading to accidental data loss.

### Solution

For PostgreSQL, generate a single combined DROP statement:

```sql
-- Before: fails if tableB has FK to tableA
DROP TABLE tableA;
DROP TABLE tableB;

-- After: resolves mutual dependencies atomically
DROP TABLE tableA, tableB;
```

Objects are grouped by type — tables and views get separate statements. Single-object deletes and non-PostgreSQL databases are completely unchanged.

### Changes

- **`SQLDialect.java`** — Added `supportsMultiObjectDrop()` interface method
- **`AbstractSQLDialect.java`** / **`BasicSQLDialect.java`** — Default `return false`
- **`PostgreDialect.java`** — Override `return true`
- **`NavigatorObjectsDeleter.java`** — Batch grouping, combined SQL generation, script preview, transaction rollback on failure

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialect.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLDialect.java
@@ -311,6 +311,8 @@ public interface SQLDialect {
 
     boolean supportsTableDropCascade();
 
+    // TODO: add supportsMultiObjectDrop() for DROP TABLE t1, t2, t3 syntax (see #40297)
+
     boolean supportsOrderByIndex();
 
     boolean supportsNestedComments();


statxc:
@davestokes @ShadelessFox could you please review this PR? thanks

statxc:
@davestokes please give me any feedback.

IvanGorshechnikov:
Hi @statxc 
Thanks for the contribution. We will review your PR and let you know if any corrections are required

IvanGorshechnikov (dismissed):

statxc:
Thanks for approving, @IvanGorshechnikov 
please any update for me. Hope to merge the pr if it has no problem. 🙏

statxc:
@SolovevAnton @serge-rider please review this PR.

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorObjectsDeleter.java
```diff
@@ -455,6 +546,285 @@ private Map<String, Object> collectObjectMakerOptionsMap(@NotNull DBSObject obje
...
+                monitor, first.context.getExecutionContext(), first.deleteOptions);
+            if (actions == null || actions.length == 0 || actions[0].getScript() == null) {
+                throw new DBException("No persist actions generated for batch DROP");
+            }
+
+            String firstSQL = actions[0].getScript();
+            String firstName = DBUtils.getEntityScriptName(first.object, first.deleteOptions);
+
+            // Extract prefix (e.g. "DROP TABLE ") and suffix (e.g. " CASCADE")
+            String[] parts = extractDropParts(firstSQL, firstName);
```
ShadelessFox: This part is quite fragile, as we're trying to extract different parts of a query using the object's name as the origin. Moreover, we assume that other commands are DROP statements with the same options (although the expectations may not hold, I believe?). Wouldn't it be more correct to delegate query generation to `SQLQueryGenerator` so it does everything for us, including single/multi-object DROP generation?

ShadelessFox:
I remind you that we review PRs when we have time. Using @ won't speed up the process.

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorObjectsDeleter.java
statxc: Good point - the string parsing approach is fragile. I'll refactor to delegate multi-object DROP generation to the object manager layer (e.g., extending SQLTableManager with a multi-object delete method) so each dialect can generate the correct SQL directly.

statxc:
I updated following @ShadelessFox feedbacks. I'd appreciate you could review again.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/SQLObjectEditor.java
```diff
@@ -222,6 +222,26 @@ protected abstract void addObjectDeleteActions(
...
+     * Generates persist actions for deleting multiple objects of the same type.
+     * Default implementation delegates to individual {@link #addObjectDeleteActions} calls.
+     * Subclasses may override to produce a combined statement (e.g. DROP TABLE t1, t2, t3).
+     */
+    @SuppressWarnings("unchecked")
+    public void addMultiObjectDeleteActions(
+        @NotNull DBRProgressMonitor monitor,
+        @NotNull DBCExecutionContext executionContext,
+        @NotNull List<DBEPersistAction> actions,
+        @NotNull List<? extends DBSObject> objects,
```
ShadelessFox:
```suggestion
        @NotNull List<OBJECT_TYPE> objects,
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/SQLObjectEditor.java
```diff
@@ -222,6 +222,26 @@ protected abstract void addObjectDeleteActions(
         @NotNull Map<String, Object> options
     ) throws DBException;
 
+    /**
+     * Generates persist actions for deleting multiple objects of the same type.
+     * Default implementation delegates to individual {@link #addObjectDeleteActions} calls.
+     * Subclasses may override to produce a combined statement (e.g. DROP TABLE t1, t2, t3).
+     */
+    @SuppressWarnings("unchecked")
```
ShadelessFox: Then this is not needed

```suggestion
```

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorObjectsDeleter.java
```diff
@@ -330,19 +344,29 @@ boolean showScriptWindow() {
...
         final StringBuilder sql = new StringBuilder();
         try {
             UIUtils.runInProgressService(monitor -> {
-                for (Object obj: selection) {
+                // Phase 1: Collect individual script data for each node
+                List<ScriptEntry> scriptEntries = new ArrayList<>();
+                for (Object obj : selection) {
                     if (obj instanceof DBNDatabaseNode) {
-                        appendScript(monitor, sql, (DBNDatabaseNode) obj);
+                        ScriptEntry entry = collectScriptEntry(monitor, (DBNDatabaseNode) obj);
```
ShadelessFox: Use instanceof pattern

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreTableManager.java
```diff
@@ -231,5 +231,37 @@ protected void addObjectDeleteActions(@NotNull DBRProgressMonitor monitor, @NotN
...
+        @NotNull DBCExecutionContext executionContext,
+        @NotNull List<DBEPersistAction> actions,
+        @NotNull List<? extends DBSObject> objects,
+        @NotNull Map<String, Object> options
+    ) throws DBException {
+        if (objects.size() <= 1) {
+            super.addMultiObjectDeleteActions(monitor, executionContext, actions, objects, options);
+            return;
+        }
+        PostgreTableBase first = (PostgreTableBase) objects.get(0);
```
ShadelessFox:
```suggestion
        PostgreTableBase first = (PostgreTableBase) objects.getFirst();
```

This also applicable to 10 other similar changes in this PR

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreViewManager.java
```diff
@@ -110,6 +110,33 @@ protected void addObjectDeleteActions(@NotNull DBRProgressMonitor monitor, @NotN
         );
     }
 
+    @Override
+    public void addMultiObjectDeleteActions(
```
ShadelessFox: I don't think you even need this override honestly, the logic looks pretty much the same as in SQLTableManager.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreTableManager.java
```diff
@@ -231,5 +231,37 @@ protected void addObjectDeleteActions(@NotNull DBRProgressMonitor monitor, @NotN
             new SQLDatabasePersistAction(ModelMessages.model_jdbc_drop_table, script);
         actions.add(action);
     }
-    
+
+    @Override
+    public void addMultiObjectDeleteActions(
```
ShadelessFox: Same here

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreViewManager.java
statxc: The override is needed because of CASCADE handling. `SQLTableManager` checks `!DBUtils.isView(first)` before adding CASCADE, so it never adds CASCADE for views. But PostgreSQL supports `DROP VIEW ... CASCADE`, and the existing individual `addObjectDeleteActions` in `PostgreViewManager` already adds it. Removing this override would break CASCADE for view batch drops.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreTableManager.java
statxc: Two reasons this override is needed:
1. External schema handling - it uses `SQLDatabasePersistActionAtomic` when any table is in an external schema. `SQLTableManager` doesn't have this logic.

2. `first.getTableTypeName()` returns the correct type like `FOREIGN TABLE`, while `getDropTableType()` in the base class may not. Without this, `DROP FOREIGN TABLE ...` would become `DROP TABLE ...`

statxc:
@ShadelessFox Good point! Thanks for reviewing. But we can't remove override because of CASCADE problems. I left reason for that. And I have fixed others. I'd appreciate you review once more.

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorObjectsDeleter.java
```diff
@@ -455,6 +580,250 @@ private Map<String, Object> collectObjectMakerOptionsMap(@NotNull DBSObject obje
...
+        }
+
+        private void executeBatchDrop(
+            @NotNull DBCExecutionContext executionContext,
+            @Nullable DBCTransactionManager txnManager,
+            @NotNull String batchSQL,
+            @NotNull DBRProgressMonitor monitor
+        ) throws DBException {
+            try (DBCSession session = executionContext.openSession(
+                    monitor, DBCExecutionPurpose.META_DDL, "Drop multiple objects")) {
```
ShadelessFox: General code style nitpick: when wrapping call arguments, each argument goes on its own line, similar to how method declarations are formatted.

This also applies to other changes in this PR

```suggestion
                    monitor,
                    DBCExecutionPurpose.META_DDL,
                    "Drop multiple objects"
                )) {
```

statxc:
Thanks @ShadelessFox I updated code style

ShadelessFox (dismissed):
I think everything is good now. Thank you for the contribution :^)

IvanGorshechnikov (dismissed):

IvanGorshechnikov:
Thank you 
We will pass your PR to our QA team

ShadelessFox:
@statxc your PR is failing to build:
```
15:51:25  [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:5.0.1:compile (default-compile) on project org.jkiss.dbeaver.ext.postgresql: Compilation failure: Compilation failure: 
15:51:25  [ERROR] /var/jenkins_home/jobs/DBeaverDesktop/jobs/build-copy/workspace/dbeaver/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreViewManager.java:[114,17] name clash: addMultiObjectDeleteActions(org.jkiss.dbeaver.model.runtime.DBRProgressMonitor,org.jkiss.dbeaver.model.exec.DBCExecutionContext,java.util.List<org.jkiss.dbeaver.model.edit.DBEPersistAction>,java.util.List<org.jkiss.dbeaver.ext.postgresql.model.PostgreViewBase>,java.util.Map<java.lang.String,java.lang.Object>) in org.jkiss.dbeaver.ext.postgresql.edit.PostgreViewManager and addMultiObjectDeleteActions(org.jkiss.dbeaver.model.runtime.DBRProgressMonitor,org.jkiss.dbeaver.model.exec.DBCExecutionContext,java.util.List<org.jkiss.dbeaver.model.edit.DBEPersistAction>,java.util.List<org.jkiss.dbeaver.ext.postgresql.model.PostgreTableBase>,java.util.Map<java.lang.String,java.lang.Object>) in org.jkiss.dbeaver.model.impl.sql.edit.struct.SQLTableManager have the same erasure, yet neither overrides the other
15:51:25  [ERROR] /var/jenkins_home/jobs/DBeaverDesktop/jobs/build-copy/workspace/dbeaver/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreViewManager.java:[113,5] method does not override or implement a method from a supertype
15:51:25  [ERROR] /var/jenkins_home/jobs/DBeaverDesktop/jobs/build-copy/workspace/dbeaver/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreViewManager.java:[122,83] incompatible types: java.util.List<org.jkiss.dbeaver.ext.postgresql.model.PostgreViewBase> cannot be converted to java.util.List<org.jkiss.dbeaver.ext.postgresql.model.PostgreTableBase>
```
Did you test it locally?

statxc:
@IvanGorshechnikov @ShadelessFox Could you please review and test again? Fixed all!

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreViewManager.java
```diff
@@ -110,6 +110,33 @@ protected void addObjectDeleteActions(@NotNull DBRProgressMonitor monitor, @NotN
...
+        @NotNull DBCExecutionContext executionContext,
+        @NotNull List<DBEPersistAction> actions,
+        @NotNull List<PostgreTableBase> objects,
+        @NotNull Map<String, Object> options
+    ) throws DBException {
+        if (objects.size() <= 1) {
+            super.addMultiObjectDeleteActions(monitor, executionContext, actions, objects, options);
+            return;
+        }
+        PostgreViewBase first = (PostgreViewBase) objects.get(0);
```
ShadelessFox:
```suggestion
        PostgreViewBase first = (PostgreViewBase) objects.getFirst()
```

statxc:
@ShadelessFox Updated! thank you 👍

ShadelessFox (approved):

statxc:
Thanks again @ShadelessFox.  @IvanGorshechnikov Could you please review this and would appreciate to approve again🙏

uslss:
verified (CE, win11)

statxc:
would you merge this if it has no problem? Does it need more checking step else?

ShadelessFox:
It will be merged later.

statxc:
@serge-rider would you merge this PR? 🙏

ShadelessFox:
@statxc It will be merged when the time comes. Please stop rushing us and pinging everyone, thanks :^)

statxc:
Thanks. I’ll wait. Sorry for the inconvenience.

statxc:
Hi maintainer team, sorry to bother again. It’s been a while since QA passed. How much longer should I wait? Is there anything else you’d like me to update?

ShadelessFox:
Hi,

> Is there anything else you’d like me to update?

No, everything's good. We'll merge it later this week.

statxc:
@ShadelessFox Sorry to ping you again 🙏 But It’s been 3 weeks since QA passed, and still I haven’t received any updates. I’m not sure why it’s taking so long even though all checks have passed. Could you please let me know? Thanks.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreTableManager.java
```diff
@@ -231,5 +231,37 @@ protected void addObjectDeleteActions(@NotNull DBRProgressMonitor monitor, @NotN
...
+        }
+        if (CommonUtils.getOption(options, OPTION_DELETE_CASCADE)) {
+            sql.append(" CASCADE"); //$NON-NLS-1$
+        }
+        boolean hasExternal = objects.stream()
+            .anyMatch(t -> t.getSchema().isExternal());
+        SQLDatabasePersistAction action = hasExternal ?
+            new SQLDatabasePersistActionAtomic(ModelMessages.model_jdbc_drop_table, sql.toString()) :
+            new SQLDatabasePersistAction(ModelMessages.model_jdbc_drop_table, sql.toString());
+        actions.add(action);
```
ShadelessFox: This code differs from its super implementation in this part. There's definitely no need to copy the entire method; instead, split it into two parts: one that generates the script, and the second that calls the first and makes an action from it.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/edit/PostgreViewManager.java
```diff
@@ -110,6 +110,33 @@ protected void addObjectDeleteActions(@NotNull DBRProgressMonitor monitor, @NotN
...
+        for (int i = 0; i < objects.size(); i++) {
+            if (i > 0) {
+                sql.append(", "); //$NON-NLS-1$
+            }
+            sql.append(((PostgreViewBase) objects.get(i)).getFullyQualifiedName(DBPEvaluationContext.DDL));
+        }
+        if (CommonUtils.getOption(options, OPTION_DELETE_CASCADE)) {
+            sql.append(" CASCADE"); //$NON-NLS-1$
+        }
+        actions.add(new SQLDatabasePersistAction("Drop view", sql.toString())); //$NON-NLS-1$
```
ShadelessFox: Same deal here. Moreover, you can check if an object is a view using `DBUtils#isView`. And it should be plural, not singular (views vs view).

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorObjectsDeleter.java
ShadelessFox, line 1: We don't want all these changes in the UI code. Please refactor it to a separate class with a minimalistic interface so it can be reused, e.g., in CloudBeaver.

ShadelessFox (changes requested):
It didn't get merged because we believed it's too complex. Please see my comments

SolovevAnton:
Hi @statxc, just checking — are you planning to continue working on this PR?

PR #40464, opened 2026-03-03
dbeaver/dbeaver#40354 Improve user cancel query

### LINKED ISSUE #40354: SQL editor marks entire script as error after canceling query (57014)
### Description

When a running query is canceled and the database returns:

SQL Error [57014]: ERROR: canceling statement due to user request

the SQL editor incorrectly underlines the entire script in red as if it contains syntax errors. This appears to be a parser/validation state issue triggered after cancellation.

Expected behavior:
If the error code is 57014 (user-initiated cancel), the editor should not mark the script as invalid or show syntax error highlights.

Actual behavior:
All lines are marked as errors until the editor is reset/reopened.

### DBeaver Version

Community edition 25.3.4.202602011558

### Operating System

macOS 15.6.1

### Database and driver

PostgreSQL 18.1

Image

### Steps to reproduce

_No response_

### Additional context

_No response_

IvanGorshechnikov:
Closes https://github.com/dbeaver/dbeaver/issues/40354

Please test against different DB: PostgreSQL, MySQL, SQL Server, Oracle and others


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/execute/SQLQueryJob.java
```diff
@@ -629,11 +630,16 @@ private boolean executeSingleElement(@NotNull DBCSession session, boolean fireEv
             }
         }
         catch (Throwable ex) {
-            if (!(ex instanceof DBException)) {
+            if (DBExecUtils.isExecutionCanceled(dataSource, ex) || monitor.isCanceled()) {
+                executionCanceled = true;
```
ShadelessFox, line 634: Don't we want to `curResult.setError(ex);` anyway?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBExecUtils.java
```diff
@@ -164,6 +166,13 @@ public static DBPErrorAssistant.ErrorType discoverErrorType(@NotNull DBPDataSour
         return DBPErrorAssistant.ErrorType.NORMAL;
     }
 
+    public static boolean isExecutionCanceled(@Nullable DBPDataSource dataSource, @NotNull Throwable error) {
+        return error instanceof InterruptedException ||
+            error instanceof DBInterruptedException ||
+            error instanceof ClosedByInterruptException ||
+            (dataSource != null && discoverErrorType(dataSource, error) == DBPErrorAssistant.ErrorType.EXECUTION_CANCELED);
```
ShadelessFox:
```suggestion
        return error instanceof InterruptedException
            || error instanceof DBInterruptedException
            || error instanceof ClosedByInterruptException
            || (dataSource != null && discoverErrorType(dataSource, error) == DBPErrorAssistant.ErrorType.EXECUTION_CANCELED);
```

ShadelessFox: But also notice that you're checking just the top-level exception. Perhaps it's the expected behavior, though

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/exec/JDBCStatementImpl.java
```diff
@@ -273,8 +274,13 @@ protected int handleExecuteResult(int result) {
...
 
-    protected SQLException handleExecuteError(Throwable ex) {
-        executeError = ex;
+    @NotNull
+    protected SQLException handleExecuteError(@NotNull Throwable ex) {
+        if (DBExecUtils.isExecutionCanceled(connection.getDataSource(), ex)) {
+            executeError = null;
+        } else {
+            executeError = ex;
+        }
```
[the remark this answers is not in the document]
ShadelessFox: I agree. It's weird that we just _consume_ the error.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/execute/SQLQueryJob.java
IvanGorshechnikov: No, otherwise the query in the SQL editor will be marked with an error

uslss:
verified (db ue, win11)

PR #40465, opened 2026-03-03
dbeaver/dbeaver#40147 Allow paste over datasource nodes (do not check…

### LINKED ISSUE #40147: Copy/paste of connection does not work anymore from CE 25.3.3
### Description

From CE 25.3.3, I cannot duplicate (copy / paste) connections anymore. When I try to paste a connection, it writes the connection name in the last selected field.

### DBeaver Version

Community Edition 25.3.3

### Operating System

Ubuntu 24.04.3

### Database and driver

_No response_

### Steps to reproduce

Create a connection
Copy the connection
Paste the connection

Expected behavior: a new connection must be created (with a slightly different name)
Current behavior: nothing really happens (except copied connection name that might be pasted in the last selected field)

### Additional context

It works with CE 25.3.2.

serge-rider:
… for read-only flag)


uslss:
Issue is still reproducible for me with active Athena connection

https://github.com/user-attachments/assets/b2699764-524a-447d-8982-11dcc05a9648

ShadelessFox:
Can't reproduce the original issue against `devel` on Manjaro KDE

Matvey16:
Reproducible against both `devel` and `40147-copypaste-fix` on Windows 11

SolovevAnton:
reproducable on devel win-11, only with athena active connection and with not empty buffer.
Other connections can be copied:
Athena
image

Other connection
image

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/actions/ObjectPropertyTester.java
```diff
@@ -308,13 +312,13 @@ public static boolean canCreateObject(DBNNode node, Boolean onlySingle) {
             if (dataSource != null && dataSource.getInfo().isReadOnlyMetaData()) {
                 return false;
             }
-            if (!(node instanceof DBNDataSource) && isMetadataChangeDisabled(dbNode)) {
```
ShadelessFox, line 311: Redundant check - it's never a data source because of

https://github.com/dbeaver/dbeaver/blob/9a2154976c45e14eb2141805f7e9fba7c61e90bb/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/actions/ObjectPropertyTester.java#L302-L305

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/actions/ObjectPropertyTester.java
```diff
@@ -308,13 +312,13 @@ public static boolean canCreateObject(DBNNode node, Boolean onlySingle) {
...
+            if (isMetadataChangeDisabled(dbNode)) {
                 return false;
             }
         }
         if (onlySingle == null) {
             // Just try to find first create handler
-            if (node instanceof DBNDataSource || node instanceof DBNLocalFolder) {
```
ShadelessFox, line 317: Same here

IvanGorshechnikov (approved):

Matvey16:
Verified

DB CE Windows

PR #40469, opened 2026-03-03
dbeaver/pro8565 filter panel expand bug

serge-rider:


Matvey16:
Verified

DB CE Windows
DB CE MacOS

PR #40476, opened 2026-03-03
dbeaver/dbeaver#35020 Fix CSV import for locale decimal/grouping separators

### LINKED ISSUE #35020: Is there no way to import CSV files with with numbers using comma as decimal char?
When you try to import a CSV file, with semi-colon separation, which is usually because you use comma for decimal numbers, there seems to be no way to import those correctly via dBeaver. And you cannot easily convert such a CSV file from semi-colon to comma separated.

it-education-md:
Fixes #35020

## Summary
- Add CSV import properties for decimal/grouping separators
- Apply locale-based defaults in settings when supported
- Normalize numeric values during CSV type detection/import
- CSV import now supports locale-specific numeric formats (US/EU style)

## Tests
- NumericFormatUtilsTest -> pass
- CSVImporterTest -> pass

## Manual verification
- EU locale
- Imported US-style sample
- Imported EU-style sample after separator change
Screenshot 2026-03-03 103909

https://github.com/user-attachments/assets/706d50b4-044e-4810-84b9-2272c69f415b


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,267 @@
...
+import org.jkiss.dbeaver.model.DBPDataKind;
+import org.jkiss.dbeaver.tools.transfer.stream.StreamDataImporterColumnInfo;
+import org.jkiss.utils.CommonUtils;
+
+import java.text.DecimalFormatSymbols;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+
+public final class NumericFormatUtils {
```
ShadelessFox, line 33: Is there any particular reason to use this over `DecimalFormat`?

it-education-md: The reason is that this class is not really doing `DecimalFormat`-style parsing or formatting.
Its job here is to normalize raw CSV numeric text before import and type detection. For example, it handles locale-specific decimal/grouping separators, rejects invalid mixed formats, validates exponent notation, and converts values into a canonical Java-parsable form.

We already use `DecimalFormatSymbols` to get the locale defaults. The rest of the logic is import-specific string normalization that is shared between CSV sampling and the actual import flow. `DecimalFormat` would not cover that behavior cleanly, and using it here would make the validation rules less explicit.

ShadelessFox: Everything that you have described can be achieved using `DecimalFormat`. It can both produce and consume formatted string values from and to a `Number`.

it-education-md: I'm not fully sure this covers every case yet, but here are some examples where `DecimalFormat` seems too lenient for strict CSV-import numeric validation, because it accepts a valid numeric prefix instead of rejecting the whole field.

```
CSV input -> DecimalFormat parse -> Result
```

- 123abc -> parses prefix -> 123

- 1,234.56xyz -> parses prefix -> 1234.56

- (decimal='.' grouping=',' pattern='#,##0.###') 12.3.4 -> parses prefix -> 12.3

- (decimal='.' grouping=',') 123. -> accepted -> 123

- (decimal='.' grouping=',' pattern='#,##0.###') 12,34,567 -> accepted -> 1234567

- (decimal=',' grouping='.' pattern='#,##0.###') 12,3,4 -> parses prefix -> 12.3

- (decimal='.' grouping=',' pattern='0.###E0') 1.2E3foo -> parses prefix -> 1200

My point is that we need stricter whole-field validation than plain `DecimalFormat`. What do you think?

ShadelessFox: For cases where a full match is expected, you can check `ParsePosition pos` and ensure it's at the end of the input.

it-education-md: Ah, got it. Let me check with it more.

it-education-md: @ShadelessFox It looks much cleaner. Would you review again?

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,173 @@
...
+        boolean scientific = containsExponent(trimmedValue, decimalSeparator, groupingSeparator);
+        BigDecimal parsedValue = parseBigDecimal(trimmedValue, decimalSeparator, groupingSeparator, scientific);
+        if (parsedValue == null) {
+            return null;
+        }
+
+        return scientific ? parsedValue.toString() : parsedValue.toPlainString();
+    }
+
+    private static boolean containsExponent(String value, char decimalSeparator, char groupingSeparator) {
```
IvanGorshechnikov:
```suggestion
    private static boolean containsExponent(@NotNull String value, char decimalSeparator, char groupingSeparator) {
```

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,173 @@
...
+            return Character.MIN_VALUE;
+        }
+        return groupingSeparator == decimalSeparator ? Character.MIN_VALUE : groupingSeparator;
+    }
+
+    public static String toPropertyValue(char separator) {
+        return separator == Character.MIN_VALUE ? "" : String.valueOf(separator);
+    }
+
+    public static void normalizeNumericValues(
```
IvanGorshechnikov, line 82: Also here please add nullability

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,173 @@
...
+    private static final String OPTIONAL_FRACTION = "#".repeat(MAX_FRACTION_DIGITS);
+    private static final String DECIMAL_PATTERN_NO_GROUPING = "0." + OPTIONAL_FRACTION;
+    private static final String DECIMAL_PATTERN_GROUPING = "#,##0." + OPTIONAL_FRACTION;
+    private static final String SCIENTIFIC_PATTERN_NO_GROUPING = "0." + OPTIONAL_FRACTION + "E0";
+    private static final String SCIENTIFIC_PATTERN_GROUPING = "#,##0." + OPTIONAL_FRACTION + "E0";
+
+    private NumericFormatUtils() {
+    }
+
+    public static char getDecimalSeparator(Map<String, Object> processorProperties) {
```
IvanGorshechnikov: Here

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,173 @@
...
+        if (processorProperties.containsKey(PROP_DECIMAL_SEPARATOR)) {
+            String decimalSeparator = CommonUtils.toString(processorProperties.get(PROP_DECIMAL_SEPARATOR)).trim();
+            return CommonUtils.isEmpty(decimalSeparator) ? Character.MIN_VALUE : decimalSeparator.charAt(0);
+        }
+        return getLocaleDecimalSeparator();
+    }
+
+    public static char getGroupingSeparator(Map<String, Object> processorProperties, char decimalSeparator) {
```
IvanGorshechnikov: And here

plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/pages/stream/StreamProducerPageSettings.java
```diff
@@ -270,6 +276,49 @@ public void activatePage() {
...
+    private boolean hasProperty(@NotNull DBPPropertyDescriptor[] properties, @NotNull String propertyId) {
+        for (DBPPropertyDescriptor property : properties) {
+            if (propertyId.equals(property.getId())) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private boolean hasStoredPropertyValue(Map<String, Object> processorProperties, String propertyId) {
```
IvanGorshechnikov: And here

IvanGorshechnikov (changes requested):

plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/pages/stream/StreamProducerPageSettings.java
it-education-md: Thanks, let me check and fix instantly.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,173 @@
...
+        for (int i = 0; i < Math.min(line.length, streamColumns.size()); i++) {
+            if (streamColumns.get(i).getDataKind() == DBPDataKind.NUMERIC && line[i] != null) {
+                String normalizedValue = normalizeNumberValue(line[i], decimalSeparator, groupingSeparator);
+                line[i] = normalizedValue == null ? line[i] : normalizedValue;
+            }
+        }
+    }
+
+    @Nullable
+    public static String normalizeNumberValue(@NotNull String value, char decimalSeparator, char groupingSeparator) {
```
ShadelessFox: A bit of Javadoc definitely won't hurt

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,173 @@
...
+    public static final String PROP_DECIMAL_SEPARATOR = "decimalSeparator";
+    public static final String PROP_GROUPING_SEPARATOR = "groupingSeparator";
+
+    // DecimalFormat patterns require an explicit upper bound for optional fraction digits.
+    private static final int MAX_FRACTION_DIGITS = 31;
+    private static final String OPTIONAL_FRACTION = "#".repeat(MAX_FRACTION_DIGITS);
+    private static final String DECIMAL_PATTERN_NO_GROUPING = "0." + OPTIONAL_FRACTION;
+    private static final String DECIMAL_PATTERN_GROUPING = "#,##0." + OPTIONAL_FRACTION;
+    private static final String SCIENTIFIC_PATTERN_NO_GROUPING = "0." + OPTIONAL_FRACTION + "E0";
+    private static final String SCIENTIFIC_PATTERN_GROUPING = "#,##0." + OPTIONAL_FRACTION + "E0";
```
ShadelessFox: I _think_ you can get away with the default decimal format that `NumberFormat` provides.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,173 @@
...
+
+    private NumericFormatUtils() {
+    }
+
+    public static char getDecimalSeparator(@NotNull Map<String, Object> processorProperties) {
+        if (processorProperties.containsKey(PROP_DECIMAL_SEPARATOR)) {
+            String decimalSeparator = CommonUtils.toString(processorProperties.get(PROP_DECIMAL_SEPARATOR)).trim();
+            return CommonUtils.isEmpty(decimalSeparator) ? Character.MIN_VALUE : decimalSeparator.charAt(0);
```
ShadelessFox: Don't use magic values unless it's at least documented. At most (the recommended way), use a nullable type, like `Character`. Don't forget to mark it as `@Nullable`.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    public static char getGroupingSeparator(@NotNull Map<String, Object> processorProperties, char decimalSeparator) {
+        if (processorProperties.containsKey(PROP_GROUPING_SEPARATOR)) {
+            String groupingSeparator = CommonUtils.toString(processorProperties.get(PROP_GROUPING_SEPARATOR)).trim();
+            if (CommonUtils.isEmpty(groupingSeparator)) {
+                return Character.MIN_VALUE;
+            }
+            char grouping = groupingSeparator.charAt(0);
+            return grouping == decimalSeparator ? Character.MIN_VALUE : grouping;
```
ShadelessFox: Same here

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,173 @@
...
+        }
+        return getLocaleGroupingSeparator(decimalSeparator);
+    }
+
+    public static char getLocaleDecimalSeparator() {
+        try {
+            return DecimalFormatSymbols.getInstance(Locale.getDefault()).getDecimalSeparator();
+        } catch (Exception ignored) {
+            return '.';
+        }
```
ShadelessFox: Why is this here? What can throw an exception?

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,173 @@
...
+        try {
+            groupingSeparator = DecimalFormatSymbols.getInstance(Locale.getDefault()).getGroupingSeparator();
+        } catch (Exception ignored) {
+            return Character.MIN_VALUE;
+        }
+        return groupingSeparator == decimalSeparator ? Character.MIN_VALUE : groupingSeparator;
+    }
+
+    public static String toPropertyValue(char separator) {
+        return separator == Character.MIN_VALUE ? "" : String.valueOf(separator);
```
ShadelessFox: See previous comment

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @NotNull
+    private static DecimalFormat createFormat(boolean scientific, char decimalSeparator, char groupingSeparator) {
+        boolean groupingEnabled = groupingSeparator != Character.MIN_VALUE;
+        String pattern = scientific
+            ? (groupingEnabled ? SCIENTIFIC_PATTERN_GROUPING : SCIENTIFIC_PATTERN_NO_GROUPING)
+            : (groupingEnabled ? DECIMAL_PATTERN_GROUPING : DECIMAL_PATTERN_NO_GROUPING);
+
+        DecimalFormat format = new DecimalFormat(pattern, createSymbols(decimalSeparator, groupingSeparator));
+        format.setParseBigDecimal(true);
```
ShadelessFox: Why do we want to parse the value as a big decimal?

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
it-education-md: The default localized decimal format is sufficient for plain numbers.
Scientific notation requires an explicit `DecimalFormat` pattern because exponent parsing is only enabled when the pattern contains `E` (for example, "0.###E0").
So it seems good to keep the dedicated pattern for scientific numbers.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
it-education-md: @ShadelessFox I still think that `setParseBigDecimal(true)` makes sense here, this method both validates and normalizes numeric input. `BigDecimal` lets us normalize from the parsed numeric value, not by rewriting the validated text.
That is safer for exact decimals and scientific notation. For example, with `DecimalFormat` validation followed by raw string replacement, 1,2E-3 could be corrupted if - were configured as a grouping separator, because the exponent sign would also be removed. `BigDecimal` avoids that risk. Am I missing something?

it-education-md:
@ShadelessFox @IvanGorshechnikov Would you mind taking another look at this PR? I’d really appreciate it.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
ShadelessFox: What's the difference between parsing to a `Float` or `BigDecimal`? Just the precision. Moreover, the code that later parses the value, in most cases, doesn't use `BigDecimal`. See `DBValueFormatting#convertStringToNumber`

From a performance point of view, `BigDecimal` is way slower to parse than a `float`. If a CSV contains millions of numbers, it might become a bottleneck.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
ShadelessFox: The default instance parses exponents with group separators just fine:
```java
jshell> DecimalFormat.getInstance().parse("1,000E3")
$1 ==> 1000000
```

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
it-education-md: @ShadelessFox Thanks for your feedback. I’ve addressed your comments. I won’t give up, LOL

it-education-md: @ShadelessFox Hi, the PR has been updated and is ready for review.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/NumericFormatUtils.java
```diff
@@ -0,0 +1,165 @@
...
+    @Nullable
+    public static String normalizeNumberValue(@NotNull String value, char decimalSeparator, @Nullable Character groupingSeparator) {
+        String trimmedValue = value.trim();
+        if (trimmedValue.isEmpty()) {
+            return null;
+        }
+
+        boolean scientific = containsExponent(trimmedValue, decimalSeparator, groupingSeparator);
+        String normalizedValue = scientific ? trimmedValue.replace('e', 'E') : trimmedValue;
+        DecimalFormat format = createFormat(decimalSeparator, groupingSeparator);
```
ShadelessFox: Please reuse these. Introduce a separate method that accepts these as parameters, and initialize them in `normalizeNumericValues` and in its overloaded `normalizeNumberValue`

it-education-md: Good catch on performance. fixed this.

it-education-md: @ShadelessFox Any further concerns?

it-education-md: A follow-up on this PR. Please let me know if anything else is needed from my side. @ShadelessFox

ShadelessFox (approved):
I think the current iteration is decent enough. Thanks :^)

it-education-md:
Thanks for your review efforts so far!

IvanGorshechnikov (approved):

IvanGorshechnikov:
Thank you @it-education-md 
We will pass your PR to QA team

DariyaFedorova:
Hi @it-education-md 
PR build fails with "NumericFormatUtilsTest.normalizeNumberValueSupportsScientificNotation:38 expected:<1.25E+3> but was:<null>"

E1izabeth:

PR #40479, opened 2026-03-04
dbeaver/dbeaver#40295 add support for version range

### LINKED ISSUE #40295: Support ranges for version of Maven dependencies
We need to support Maven dependencies with a range of versions like `[1.0,2.0)`. It breaks when we try to download the H2GIS driver v2.2.5. 

Documentation for it: https://maven.apache.org/pom.html#Version_Order_Specification

We need to use the latest version.

Destrolaric:
Closes dbeaver/dbeaver#40295
Adds extracting best version from metadata.xml if version range is present


ShadelessFox (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified

DB CE Windows

PR #40489, opened 2026-03-05
Dbeaver/dbeaver#39733 add proceed button at first step data export

### LINKED ISSUE #39733: When exporting data, enable Proceed button already at first step
I am often exporting the data from my queries to check something in Excel or so. To finally export the data, I have to click the 'Next' button several times in the 'Data Transfer' pop up before I can finally hit the 'Proceed' button. Since I am always exporting the data in the same fashion, it is unnecessary to always click Next.

It would be good if the Proceed button was enabled right away at the first step. When clicking the Next button, the selections and values are always pre-defaulted, so technically the Proceed button could already be enabled at the first step and also just use the defaulted values.

n/a

**Additional context**
n/a

SolovevAnton:
closes dbeaver/dbeaver#39733 : proceed button avaliable to cycle to last wizard page

### DIFF
--- a/plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/dialog/DataTransferConfigurationWizardDialog.java
+++ b/plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/dialog/DataTransferConfigurationWizardDialog.java
@@ -45,15 +45,15 @@ public class DataTransferConfigurationWizardDialog extends TaskConfigurationWiza
         super.buttonPressed(buttonId);
     }
 
-    public void cycleToLastPage() {
-        while (canShowNext()) {
-            nextPressed();
-        }
-    }
-
     @Override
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/dialog/DataTransferConfigurationWizardDialog.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    protected void buttonPressed(int buttonId) {
+        if (buttonId == IDialogConstants.OK_ID && !getWizard().canFinish() && canShowNext()) {
+            cycleToLastPage();
+            return;
+        }
+        super.buttonPressed(buttonId);
+    }
+
+    public void cycleToLastPage() {
```
ShadelessFox: Could be `private`

ShadelessFox (approved):

Destrolaric (approved):

uslss:
verified (win11, EE)

PR #40497, opened 2026-03-06
Dbeaver/pro#8396 fix delete column in record mode

SolovevAnton:
dbeaver/pro#8396: fix now only one added column is shown in record mode


IvanGorshechnikov (approved):

Matvey16:
Verified

DB CE Windows

PR #40499, opened 2026-03-06
dbeaver/pro#8417 Add support for short IDs in data export processors

labanovichttps:


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/registry/DataTransferRegistry.java
```diff
@@ -178,7 +178,7 @@ public List<DataTransferProcessorDescriptor> getAvailableProcessors(
                 }
             }
         }
-        return null;
+        return Collections.emptyList();
```
ShadelessFox:
```suggestion
        return List.of();
```

alexander-skoblikov (approved):

PR #40502, opened 2026-03-08
dbeaver/pro#8627 Model extender redesign

serge-rider:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/registry/DBNRegistry.java
```diff
@@ -57,32 +54,38 @@ private DBNRegistry(IExtensionRegistry registry) {
...
         }
         if (!CommonUtils.isEmpty(extraNodes)) {
-            for (DBNNode eNode : extraNodes) {
+            for (DBNNodeExtension eNode : extraNodes) {
                 parentNode.addExtraNode(eNode, reflect);
             }
         }
     }
 
+    public <T> DBNModelExtender findNodeExtender(@NotNull DBNNode parentNode, @NotNull Class<T> nodeType) {
```
ShadelessFox, line 81:
```suggestion
    @Nullable
    public <T> DBNModelExtender findNodeExtender(@NotNull DBNNode parentNode, @NotNull Class<T> nodeType) {
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNNodeExtension.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Nullable
+    public DBNNode resolveRealNode() {
+        DBNNode extension = createExtension();
+        if (extension == null) {
+            return null;
+        }
+
+        return ((DBNNodeExtendable)getParentNode()).resolveTargetNode(this, extension);
```
ShadelessFox, line 81: The constructor doesn't enforce the parent node to be `DBNNodeExtendable` assignable. Either make the constructor accept `DBNNodeExtendable`, or guard the cast with `instanceof`

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNNodeExtension.java
```diff
@@ -0,0 +1,119 @@
...
+    @Override
+    protected boolean allowsChildren() {
+        return true;
+    }
+
+    @Nullable
+    @Override
+    public DBNNode[] getChildren(@NotNull DBRProgressMonitor monitor) throws DBException {
+        return null;
+    }
```
ShadelessFox, line 98: Children are allowed, but no children?

serge-rider: This is a fake node. It will be replaced by a real one which will return real children

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNNodeExtension.java
serge-rider: Unfortunately DBNNode is not an interface and we cannot extend it in DBNNodeExtendable (which IS interface).
A bit dirty but I don't think it is critical.

ShadelessFox (dismissed):

PR #40529, opened 2026-03-12
dbeaver/pro#8354 Validate conflicting CLI parameters for update/create

labanovichttps:
execute update
`dbvr datasource update <DATASOURCE_ID> --url="jdbc:postgresql://host:port/db" --port=5433` for example
and create cases
`dbvr datasource create --driver=h2_embedded_v2 --url=jdbc:h2:mem:test --host=localhost`


plugins/org.jkiss.dbeaver.model.cli/src/org/jkiss/dbeaver/model/cli/model/option/DataSourceOptions.java
```diff
@@ -109,6 +110,17 @@ public boolean isSavePassword() {
         return savePassword;
     }
 
+    public void validate() throws CLIException {
```
ShadelessFox: Perhaps you could utilize https://picocli.info/#_mutually_exclusive_options?

alexander-skoblikov (changes requested):
let's use exclusive arguments, see CreateOrFindDataSource for example

alexander-skoblikov (approved):

PR #40534, opened 2026-03-14
Fix native tool host/port extraction for URL-based connections

### LINKED ISSUE #40485: Database Dump tool doesn't work with URL based connection
### Description

The error shows that the connection tries to use localhost whenever you are using a connection URL string in the connection setup. Would a PR for this be accepted to use what is in the URL connection string rather than the unused placeholder data? 

I have not tested for other DBs if this is the case so I can only really confirm for MySql as of right now. This dump works just moving the information to the connection fields.

Image
(this image is the defaults when you have url selected and fully input)

```
/opt/homebrew/Cellar/mysql-client/9.6.0/bin/mysqldump --skip-lock-tables --routines --add-drop-table --disable-keys --extended-insert -u root --host= test
2026-03-04 09:23:38.161 - Dump database into file:///Users/username/dump-test-202603040923.sql
Task 'MySQL dump' started at Wed Mar 04 09:23:38 EST 2026
mysqldump: Got error: 2002: Can't connect to local MySQL server through socket '/tmp/mysql.sock' (2) when trying to connect

Task 'MySQL dump' finished at Wed Mar 04 09:23:38 EST 2026
2026-03-04 09:23:38.371 - IO error: Process failed (exit code = 2). See error log.
2026-03-04 09:23:38.372 - java.io.IOException: Process failed (exit code = 2). See error log.
	at org.jkiss.dbeaver.tasks.nativetool.AbstractNativeToolHandler.validateErrorCode(AbstractNativeToolHandler.java:264)
	at org.jkiss.dbeaver.tasks.nativetool.AbstractNativeToolHandler.executeProcess(AbstractNativeToolHandler.java:242)
	at org.jkiss.dbeaver.tasks.nativetool.AbstractNativeToolHandler.doExecute(AbstractNativeToolHandler.java:285)
	at org.jkiss.dbeaver.ext.mysql.tasks.MySQLNativeToolHandler.doExecute(MySQLNativeToolHandler.java:47)
	at org.jkiss.dbeaver.tasks.nativetool.AbstractNativeToolHandler.lambda$executeTask$0(AbstractNativeToolHandler.java:88)
	at org.jkiss.dbeaver.runtime.RunnableContextDelegate.lambda$run$0(RunnableContextDelegate.java:39)
	at org.eclipse.jface.operation.ModalContext$ModalContextThread.run(ModalContext.java:123)
```

### DBeaver Version

Community Edition 26.0.0

### Operating System

Mac OS Tahoe 26.2

### Database and driver

1. MySql 8.0

### Steps to reproduce

1. Setup connection with MySQL DB with url connection string
2. Try to run a DB dump

### Additional context

_No response_

mango766:
## Summary

Fixes #40485

When a database connection uses URL configuration type (instead of manual host/port/database fields), the native tool commands (`mysqldump`, `pg_dump`, etc.) fail to extract the host and port from the JDBC URL. This results in an empty `--host=` parameter being passed to the tool, which causes it to fall back to localhost via socket and fail.

## Root Cause

In `NativeToolUtils.addHostAndPortParamsToCmd()`, the URI fallback parser was placed inside an `else` branch that only executed when `sampleUrl` was `null`. However, for MySQL connections the sample URL is always defined (`jdbc:mysql://{host}[:{port}]/[{database}]`), so when `DatabaseURL.extractConfigurationFromUrl()` failed to match the user's actual URL against the sample pattern (e.g., due to query parameters or slight format differences), the code had no fallback -- hostname remained empty.

## Changes

Three issues fixed in `NativeToolUtils.java`:

1. **URI fallback is now always attempted** when the sample URL extraction fails (not just when `sampleUrl == null`). This ensures that even when the regex-based extraction cannot parse the URL, the standard `java.net.URI` parser is tried.

2. **Empty `--host=` is no longer appended** to the command. Previously, `--host=` was always added even with an empty hostname, causing native tools to fail with socket connection errors. Now `--host=` is only added when a hostname was successfully extracted.

3. **Fixed port `-1` handling**. `URI.getPort()` returns `-1` when no port is present. The old code converted this to the string `"-1"` via `String.valueOf()`, which would pass `--port=-1` to the native tool. Now the port is only used when it is a positive integer.

## Test Plan

- [x] Configure a MySQL connection using URL mode (e.g., `jdbc:mysql://myhost:3306/mydb?useSSL=false`)
- [x] Run Database Dump -- should correctly extract host and port from the URL
- [ ] Configure a PostgreSQL connection using URL mode
- [ ] Run pg_dump -- should correctly extract host and port from the URL
- [ ] Verify that manual (field-based) connections still work as before


arhayka:
Hi @mango766! Thank you for the contribution. I’ve passed your PR to the development team for review and testing.

ShadelessFox (approved):

IvanGorshechnikov (approved):

uslss:
verified (db ce, win11)

PR #40543, opened 2026-03-16
#39743 Fix MariaDB zero date fetching

### LINKED ISSUE #39743: Date field showing as NULL
### Description

Hi

First, a big thank you to you all for providing DBeaver as a Community Tool - it is very much appreciated

The issue I have is concerning the handling of a MariaDB date field. 

The table defines the field as   `ReportingMonth` date DEFAULT NULL,

In some circumstances, the logic of the code writes a values of '0000-00-00' into this field which is accepted by the database and displays correctly if the information is retrieved using command line mysql.

However, via the DBeaver GUI view, values in this field are displayed as "NULL" which is misleading.

If I run a query such as SELECT * FROM mytable WHERE ReportingMonth = '0000-00-00', the correct number or rows are returned but the value displayed in the ReportingMonth column is "NULL"

### DBeaver Version

Community 25.2.0

### Operating System

Linux Mint 22.2 runs DBeavor, Raspian 12 bookwork runs MariaDB

### Database and driver

Native MariaDB driver supplied in DBeavor package

### Steps to reproduce

_No response_

### Additional context

_No response_

EastLord:
## How to test

1. Start a local MariaDB container and expose it to the host.

```bash
docker run --detach \
  --name some-mariadb \
  -e MARIADB_ROOT_PASSWORD=root \
  -p 3306:3306 \
  mariadb:latest
```

2. Create a test database/table and insert zero-date values.

```sql
CREATE DATABASE IF NOT EXISTS test_zero_date_db;
USE test_zero_date_db;

SET SESSION sql_mode = '';

DROP TABLE IF EXISTS test_zero_date;
CREATE TABLE test_zero_date (
    id INT PRIMARY KEY,
    d DATE,
    dt DATETIME
);

INSERT INTO test_zero_date (id, d, dt)
VALUES
    (1, '0000-00-00', '0000-00-00 00:00:00'),
    (2, '2024-01-02', '2024-01-02 03:04:05'),
    (3, NULL, NULL);

SELECT id, d, dt FROM test_zero_date ORDER BY id;
```


plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/data/MySQLDateTimeValueHandler.java
```diff
@@ -165,4 +168,11 @@ public Object getValueFromObject(@NotNull DBCSession session, @NotNull DBSTypedO
         }
         return super.getValueFromObject(session, type, object, copy, validateValue);
     }
+
+    private static boolean isMariaDBDateTimeStringRead(@NotNull DBSTypedObject type) {
```
SolovevAnton: lets make it non static

SolovevAnton (changes requested):
Thank you for your contribution

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/data/MySQLDateTimeValueHandler.java
EastLord: Thanks for the suggestion. have revised it.

SolovevAnton (approved):

IvanGorshechnikov (approved):

IvanGorshechnikov:

uslss:
verified (win11, CE)

PR #40553, opened 2026-03-18
Fix: column completion not working after WHERE clause in legacy SQL

jerateep:
Problem:

In the Legacy SQL completion engine, column suggestions were not shown when typing a partial column name after WHERE, or when using a table alias followed by . (e.g. h.), especially when the table is referenced with a fully-qualified name across databases (e.g. db_backoffice.dbo.MEM_Head).
image
image
image

Root cause:

In 

SQLCompletionAnalyzer.java
, the non-empty wordPart branch called 

getTableListFromAlias()
 and 

getTableFromAlias()
 with sc — the currently active/selected database container. When the table in the query belongs to a different catalog than the active one, the lookup returned nothing, causing the completion to fall through with no proposals.

Fix:

Changed the calls to use the root dataSource instead of sc, consistent with the emptyWord branch which already worked correctly. Also removed an early-return that aborted proposals when a partial wordPart matched a table name instead of a column prefix.

Test cases:

SELECT * FROM db_name.dbo.Table1 WHERE col → shows matching columns ✅
SELECT * FROM db_name.dbo.Table1 t WHERE t. → shows all columns ✅

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/completion/SQLCompletionAnalyzer.java
@@ -293,6 +293,7 @@ public class SQLCompletionAnalyzer implements DBRRunnableParametrized<DBRProgres
                 }
             } else if (!isInLiteral) {
                 DBSObject rootObject = null;
+                List<DBSObject> rootObjects = null;
                 if (queryType == SQLCompletionRequest.QueryType.COLUMN && dataSource instanceof DBSObjectContainer) {
                     // Part of column name
                     // Try to get from active object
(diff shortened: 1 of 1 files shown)


ged-yuko:
@jerateep Did you try to enable `Use global search (in all schemas)` option?

jerateep:
@ged-yuko I'm not entirely sure if that option covers this, but the issue I found is specifically that column names don't suggest after the WHERE clause when using this mode (Legacy engine).

ged-yuko:
> @ged-yuko I'm not entirely sure if that option covers this, but the issue I found is specifically that column names don't suggest after the WHERE clause when using this mode (Legacy engine).

My point is: it is intentional to resolve names only within current editor context by default. And there is a special option which regulates whether other schemas should be considered. It looks excessive to me to change default behavior from narrow-scoped lookup to globally scoped lookup. I noted many users switching from semantic-based completion to legacy precisely because it uses less database metadata and thus queries less information than more advanced one.

Or maybe I am missing what's going on while looking at the description of your modifications.

Ah, it needs more careful examination to remember the logic here. I have been reading these sources a pretty long ago. I am going to look at this stuff a bit later if i can. Thanks for bringing attention anyway.

E1izabeth (changes requested):
As I see it, the root cause of the issue is inside the `getTableFromAlias()` method, which fails to resolve the table by its fully qualified name. This happens because the call to `SQLSearchUtils.findObjectByFQN()` returns `null`, which is the result of an attempt to find a catalog reference within the schema.
The provided solution doesn't fix the root cause of the issue.

E1izabeth:

PR #40558, opened 2026-03-18
dbeaver/pro#8260 Add `SMTokenInfo` class and `readAccessTokenInfo` me…

ivgag:
…thod to enhance authentication management


plugins/org.jkiss.dbeaver.model.sm/src/org/jkiss/dbeaver/model/auth/SMTokenInfo.java
```diff
@@ -0,0 +1,84 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jkiss.dbeaver.model.auth;
+
+import org.jkiss.code.NotNull;
+import org.jkiss.code.Nullable;
+
+public class SMTokenInfo {
```
ShadelessFox, line 23: Feels like a good candidate to be `record`-ified

PR #40570, opened 2026-03-20
#40569 [CUBRID] Fix binding failure in PreparedStatement for table names containing single quote

### LINKED ISSUE #40569: [CUBRID] Fix binding failure in PreparedStatement for table names containing single quote
### Description

Opening the table column tab fails when the table name contains a single quote. The following error is thrown:

`Missing or invalid position of the bind variable provided.`

Image

### DBeaver Version

DBeaver 26.0.0

### Operating System

All

### Database and driver

_No response_

### Steps to reproduce

1. Create table following query below:

`CREATE TABLE DBA."abc'test" ( "column1" VARCHAR(100) );`
2. Refresh Database
3. Go to Navigator -> Schemas -> DBA -> Tables
4. Double to open the table

### Additional context

_No response_

longhaseng52:
Opening the table column tab fails when the table name contains a single quote. The following error is thrown:

`Missing or invalid position of the bind variable provided.`

Issue Ref: https://github.com/dbeaver/dbeaver/issues/40569


arhayka:
Hi @longhaseng52! Thank you for the contribution ❤️

serge-rider:
It is generally ok.
But metadata read performance is really bad because you execute a separate query for each table column. In big databases there may be thousands of columns.

Please try to use batch approach. Run `show columns` once per table or event schema.

longhaseng52:
@serge-rider 
Thank you for your feedback. now i’ve updated the implementation to use a batch approach.

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridUser.java
```diff
@@ -187,25 +191,47 @@ protected GenericTableColumn fetchChild(
...
         }
+
+        private static class ColumnExtraInfo {
+            String type;
+            String extra;
+            ColumnExtraInfo(String type, String extra) { 
+                this.type = type;
+                this.extra = extra;
+            }
+        }
```
ShadelessFox: Please use `record` if its fields are not intended to be modified

ShadelessFox: Also, missing nullability annotations

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridUser.java
```diff
@@ -187,25 +191,47 @@ protected GenericTableColumn fetchChild(
...
-                dbStat.setString(1, columnName);
-                try (JDBCResultSet result = dbStat.executeQuery()) {
-                    if (result.next()) {
-                        showDataType = JDBCUtils.safeGetString(result, "Type");
-                        autoIncrement = CubridConstants.AUTO_INCREMENT.equals(JDBCUtils.safeGetString(result, "Extra"));
+
+            // Batch load column info if we moved to a new table
+            if (!tableName.equals(lastTableName)) {
+                columnInfoMap.clear();
+                lastTableName = tableName;
```
ShadelessFox, line 201: Caches already cache the results. Why is this needed? I can imagine it's also prone to race conditions.

ShadelessFox (changes requested):

E1izabeth (approved):

ShadelessFox (approved):

uslss:
verified (ce, win11)

PR #40571, opened 2026-03-20
dbeaver/pro#8736 Proxy service activation fix

serge-rider:


Matvey16:
Verified

DB CE, EE Windows

PR #40576, opened 2026-03-20
dbeaver/pro#8772 Replace local file with an embedded server

### LINKED ISSUE #40581: OpenStreetMap tile request 403r Access Blocked error message
### Description

Using Dbeaver 26.0.0, I like looking at my data with the Map side panel with OpenStreetMap. 
But following the enhancement of  a new OSM Tile policy, users of Dbeaver and various other geospatial  tools, we are receiving tiles with  a 403r Access Blocked error message.

Image

This because the OSM team figths against OSM Tiles scrappers, the heavy or inappropriate use  of tiles and have applied a new Tile policy.

Applications that request the OSM tiles from tile.openstreetmap.org should include  Referer and User-Agent fields in the Tile request. 

For more details, See https://operations.osmfoundation.org/policies/tiles/ and the OSM Community  discussion
https://community.openstreetmap.org/t/access-blocked-on-osm-org/141419 

### DBeaver Version

Community Edition Version 26.0.0.202603011816

### Operating System

Windows 11

### Database and driver

1. PostgreSQL 16
2. PostgreSQL JDBC Driver

### Steps to reproduce

1. Visualize geospatial data with the Map View sidepanel
2.  Select OpenStreetMap background
3.  Zoom-in

### Additional context

- OpenStreetMap Tile policy https://operations.osmfoundation.org/policies/tiles/ 
- OSM Community  discussion about Blocked tiles https://community.openstreetmap.org/t/access-blocked-on-osm-org/141419

ShadelessFox:
This fixes a weird issue with OpenStreetMap since it now requires the Referer header to be set. For local files, that header is not set; as a workaround, we start a local server with a correct Host header, which is passed as the Referer header for outbound requests.

Please test against Windows, Linux, and macOS.


plugins/org.jkiss.dbeaver.data.gis.view/src/org/jkiss/dbeaver/ui/gis/panel/GISLeafletHttpServer.java
```diff
@@ -0,0 +1,99 @@
...
+import java.io.Closeable;
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.Executors;
+
+/**
+ * A simple server that serves the GIS viewer's HTML and related resources.
+ */
+final class GISLeafletHttpServer implements Closeable {
```
serge-rider: I think we need a single serve4r for all viewers. There may be tens or hundreds of opened gis viewers and thus we will pam with tons of opened ports + consume much more resources.

Let's use singleton server, start it on demand + shutdown once the last viewer is closed.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.data.gis.view/src/org/jkiss/dbeaver/ui/gis/panel/GISLeafletHttpServer.java
ShadelessFox: This was the plan, but unfortunately, it complicates things a lot. In this PR, I made it so that the consumer of this class is responsible for supplying the complete source code of the `index` page; it's not generated by the server, so the server can't be reused.

PR #40579, opened 2026-03-21
perf: fix slow navigator and editing for large MySQL databases (64K+ tables)

### LINKED ISSUE #40612: Performance: MySQL/MariaDB databases with large table counts are extremely slow
### Description

  DBeaver becomes extremely slow or unusable when connected to MySQL/MariaDB databases containing very large numbers of tables (e.g. 64K+ tables, common in WordPress multisite deployments). Three root causes have been identified:

  1. **Bulk constraint loading on navigator expansion** — `MySQLTable.getConstraints()` loads all unique key constraints for every table via the shared cache on each navigator expansion. With 64K+ tables this triggers massive bulk metadata queries on every tree interaction.

  2. **Search bar filter applied after pagination** — The SWT `PatternFilter` runs after pagination truncation in
  `DatabaseNavigatorContentProvider.getFinalNodes()`. When the table list exceeds `maxFetchSize`, matching items beyond the pagination cutoff are hidden behind "More..." nodes and never shown to the user.

  3. **Identifier quote escape character breaks SQL parsing** — Identifier quote rules (e.g. backticks in MySQL) incorrectly use the string escape character (backslash). This causes the tokenizer to consume past the closing backtick in expressions like `DATE_ADD(\`date\`, INTERVAL 1 DAY)`, breaking bracket tracking and SQL statement delimiter detection.

### DBeaver Version

Community Edition 26.0.2

### Operating System

macOS Sequoia 26.3 (Apple Silicon), also reproducible on intel silicon 

### Database and driver

  1. Database name and version: MySQL 8.x / MariaDB 10.x
  2. Driver name and version: MySQL Connector/J (bundled)
  3. Tested with 64K+ tables (WordPress multisite deployment)

### Steps to reproduce

  1. Connect to a MySQL/MariaDB database with 1,000+ tables
  2. Expand the Tables node in Database Navigator - observe long delay
  3. Type a table name in the navigator search bar - matching tables beyond the first page are not shown
  4. Open a SQL editor and run a script containing backtick-quoted identifiers inside function calls (e.g. `DATE_ADD(\`date\`, INTERVAL 1 DAY);`) - observe
  incorrect statement splitting

### Additional context

The combination of checking "Do not read tables information in SQL and data editors" (in Edit Connection → Metadata or Preferences → Metadata) together with the fixes in this PR makes navigation and editing extremely fast even with 64K+ tables. 

Without the PR fixes, that setting alone is not sufficient - constraint loading and pagination filtering still cause significant delays.

A PR with fixes is available: https://github.com/dbeaver/dbeaver/pull/40579

blakey0312:
## Summary

DBeaver is almost unusable with MySQL/MariaDB databases containing very large numbers of tables (e.g. 64K+ tables in WordPress multisite). This PR addresses three root causes:

- **Remove bulk constraint loading from `getTables()`** - previously loaded all unique key constraints for every table on each navigator expansion. Now constraints load lazily per-table when actually needed for editing.
- **Lazy per-table column and constraint loading for data editor** - the data editor uses `LocalCacheProgressMonitor` (force-cache mode) which skips database loads. Since bulk preloading is removed, `getAttribute()` and `getConstraints()` now trigger a real per-table load when the cache is empty, so the editor can resolve columns and primary keys correctly.
- **Pre-filter navigator children before pagination** - the search bar filter was applied after the pagination cutoff, so matching items in large lists could be hidden behind "More..." nodes. Now filters all children with fast `String.contains()` before pagination truncation.
- **Fix identifier quote escape character** - identifier quote rules (e.g. backticks in MySQL) incorrectly used the string escape character (backslash), which could cause the tokenizer to consume past the closing quote, breaking bracket tracking and delimiter detection.

## Test plan

- [ ] Connect to a MySQL/MariaDB database with 1000+ tables
- [ ] Expand Tables node - should load without long delay
- [ ] Use navigator search bar to filter tables - matching items should appear regardless of list size
- [ ] Open a table's Data tab and verify editing works (primary key detected, no "Table metadata not found")
- [ ] Run a SQL script with backtick-quoted identifiers in DATE_ADD() etc. - statements should split correctly on `;`
- [ ] Verify normal-sized databases still work as expected

**Recommended settings for large databases:** 
- In Preferences > User Interface > Navigator, uncheck "Show statistics info". 
- In Edit Connection > Metadata (or Preferences > Connections > Metadata), check "Do not read tables information in SQL and data editors". These skip `SHOW TABLE STATUS` queries that load engine type, row count, collation, and data sizes on every table interaction. Combined with the lazy loading changes in this PR, this eliminates all unnecessary metadata overhead.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLTable.java
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLTable.java
@@ -30,6 +30,7 @@ import org.jkiss.dbeaver.model.impl.jdbc.JDBCUtils;
 import org.jkiss.dbeaver.model.impl.jdbc.cache.JDBCObjectCache;
 import org.jkiss.dbeaver.model.meta.*;
 import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
+import org.jkiss.dbeaver.model.runtime.VoidProgressMonitor;
 import org.jkiss.dbeaver.model.struct.*;
 import org.jkiss.dbeaver.model.struct.cache.DBSObjectCache;
 import org.jkiss.dbeaver.model.struct.cache.SimpleObjectCache;
@@ -315,6 +316,10 @@ public class MySQLTable extends MySQLTableBase
     public Collection<MySQLTableConstraint> getConstraints(@NotNull DBRProgressMonitor monitor)
         throws DBException
     {
+        if (monitor.isForceCacheUsage()) {
+            getContainer().uniqueKeyCache.getObjects(
+                new VoidProgressMonitor(), getContainer(), this);
+        }
         List<MySQLTableConstraint> constraintObjects = getContainer().uniqueKeyCache.getObjects(monitor, getContainer(), this);
         if (getDataSource().supportsCheckConstraints()) {
             List<MySQLTableConstraint> checkConstraintObjects = getContainer().checkConstraintCache.getObjects(monitor, getContainer(), this);
--- a/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLCatalog.java
+++ b/plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLCatalog.java
@@ -330,12 +330,7 @@ public class MySQLCatalog implements
 
     @Association
     public Collection<MySQLTable> getTables(DBRProgressMonitor monitor) throws DBException {
-        List<MySQLTable> tables = getTableCache().getTypedObjects(monitor, this, MySQLTable.class);
-        if (getDataSource().readKeysWithColumns()) {
-            // Read constraints with columns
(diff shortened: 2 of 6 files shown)


E1izabeth:
@blakey0312 thanks for the contribution!

Could you create the issue or link the existing one and reference issue number at the beginning of the commit messages?

plugins/org.jkiss.dbeaver.ext.mysql/src/org/jkiss/dbeaver/ext/mysql/model/MySQLTable.java
```diff
@@ -315,6 +316,10 @@ public List<DBSEntityConstraintInfo> getSupportedConstraints() {
     public Collection<MySQLTableConstraint> getConstraints(@NotNull DBRProgressMonitor monitor)
         throws DBException
     {
+        if (monitor.isForceCacheUsage()) {
+            getContainer().uniqueKeyCache.getObjects(
+                new VoidProgressMonitor(), getContainer(), this);
+        }
```
Destrolaric: Same

Destrolaric:
The changes in PR look weird to me. 
1) RuleManager changes are a really bad idea. I am almost sure it will break something with different databases 
2) Cache logic might not lead to a significant improvement in performance. Also, the option to load them at start can be disabled

PR #40175, opened 2026-01-22
#39264 Add preference option to load lazy comments in ERD diagrams

jhjalison01:
## Summary

This PR adds a new preference option "Load lazy comments" to the Diagram Editor page. When enabled, the ERD will fetch and display lazy-loaded table and column comments/descriptions.

### New Preference

- Added "Load lazy comments" checkbox under the Advanced section in Diagram Editor
- Default value is `false` (disabled) to preserve performance for large schemas

### Implementation

- Extended `ERDAttributeSettings.java` to carry the `loadLazyDescriptions` flag from UI to Model layer
- Updated `ERDContentProviderDecorated.java` to read the preference and pass it via settings
- Updated `ERDContentProviderDefault.fillEntityFromObject()` to load lazy descriptions for both tables and columns when enabled

## Files Modified

- ERDUIConstants.java - Added preference constant
- ERDAttributeSettings.java - Added `loadLazyDescriptions` field
- ERDContentProviderDefault.java - Implemented lazy description loading logic
- ERDContentProviderDecorated.java - Read preference and pass to settings
- ERDPreferencePage.java - Added UI checkbox
- ERDPreferencesInitializer.java - Set default value
- ERDEditorPart.java - Added preference change listener
- ERDUIMessages.java / ERDUIMessages.properties - Added label text for the new preference

## Testing

1. Open Preferences > Editors > Diagram Editor
2. Select "Load lazy comments" checkbox
3. Open an ERD for a table with comments
4. Verify comments are displayed

https://github.com/user-attachments/assets/a3a6de61-d12a-4a37-85e6-8b767ee814ed

Issue #35071, opened 2024-08-05
[Manticore] Add new driver


KarelWintersky:
**Is your feature request related to a problem? Please describe.**

Auto view data of Manticore index produces syntax error:

![image](https://github.com/user-attachments/assets/811eb313-80e0-49a9-892f-c86ad741e68e)

because table alias used for single table (it's not required)

checkbox : "insert table aliases (in FROM clause)" is "N/A"

![image](https://github.com/user-attachments/assets/5ad42bba-d47b-4378-8e2b-4273a1973320)

I can assume that automatic query generation mechanisms ignore user settings for the query editor.

**Describe the solution you'd like**

the automatic query generator should use the settings for custom autocompletion mode OR do not complete the table alias if the query specifies a single table

**Describe alternatives you've considered**

`apt purge dbeaver`

E1izabeth:
Thank you for request.
Could you tell me your driver and server version? You can get this information by clicking `Test connection...` button in the `Connection configuration` dialog.
![image](https://github.com/user-attachments/assets/9f66445f-2ba8-4c37-b85c-9c89e5d9fcd9)

KarelWintersky:
![image](https://github.com/user-attachments/assets/fc11829f-6825-41e9-bf1e-02dcab72aa86)

![image](https://github.com/user-attachments/assets/ba99db86-42f3-48e0-bd06-fa2b53a424ed)

**Note:** 

ManticoreSearch can understand connection from DBEaver at host `127.0.0.1` port `9306`, but does not support table aliases.

looks like DBEaver's checkbox : "insert table aliases (in FROM clause)" is "N/A" used only for editor, not for auto generated requests.

MashaKorax:
The issue will be solved when the Manticore driver will be implemented.

sanikolaev:
@MashaKorax Is there any documentation on how to add a new plugin? I couldn't find one.

E1izabeth:
There is no general guide of how to add a new plugin, because every case is different. You can take a look at the examples of successfully merged PRs. It's easy to find them by looking at the connected issues with label "new driver".

The setting that you're referencing "Insert table aliases (in FROM clause)" is from Code Completion setting, so it works for autocomplete only. It's expected.

The behavior of inserting alias is regulated in `SQLDialect::supportsAliasInSelect`.

KarelWintersky:
@E1izabeth 

[...]

PR #40582, opened 2026-03-22
Fix #39264: Eagerly load table/column comments in ERD to fix lazy loading

### LINKED ISSUE #39264: No table/Column comments in ERD because of lazy comment loading
Currently, table and column comments are lazy-loaded whenever I open table and see column structure. This is good when I check individual tables. But when it comes to ERD and I check view style > show comments, No comments are shown because no comments are currently loaded due to lazy-loading so this option is meaningless.

I think it will be better all table and column comments are loaded before creating ERD or make an option to load comments before creating ERD.

KartikeyTripathi469:
When an ER Diagram opens, table and column comments are blank even though they exist in the database. Pressing Refresh makes them appear, confirming the cause is lazy loading. Two bugs were found: in ERDEntity.java line 193, reloadAttributes() was calling fillEntityFromObject() with a new VoidProgressMonitor(), a fake monitor that cannot trigger any real DB fetch; and in ERDEntityAttribute.java line 135, getDescription() was called without a monitor, which returns null for objects implementing DBPObjectWithLazyDescription. The fix replaces VoidProgressMonitor with a diagram.getMonitor() (falling back to void if null) in ERDEntity.java, and checks for DBPObjectWithLazyDescription in ERDEntityAttribute.java to pass context.getMonitor() so a real DB fetch is triggered. Note: the devel branch has a pre-existing unrelated build failure in model.sql (LSM API mismatch) — our changed files compile cleanly when their dependency chain is available.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.erd/src/org/jkiss/dbeaver/model/erd/ERDEntity.java
+++ b/plugins/org.jkiss.dbeaver.model.erd/src/org/jkiss/dbeaver/model/erd/ERDEntity.java
@@ -190,7 +190,7 @@ public class ERDEntity extends ERDElement<DBSEntity> {
             attributes.clear();
         }
         try {
-            diagram.getContentProvider().fillEntityFromObject(new VoidProgressMonitor(), diagram, Collections.emptyList(), this);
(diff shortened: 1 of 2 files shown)


ShadelessFox:
Thanks for the contribution, but unfortunately, there's already another PR submitted that is about to be merged: https://github.com/dbeaver/dbeaver/pull/40175.

KartikeyTripathi469:
That's okay. Please review my PR https://github.com/dbeaver/dbeaver/issues/35071#issuecomment-4090514543

PR #40586, opened 2026-03-23
dbeaver/pro#8095 Track canceled fetches in QM statement execution

vkirst:


plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/exec/JDBCResultSetImpl.java
```diff
@@ -190,6 +192,12 @@ public boolean nextRow()
             return this.next();
         }
         catch (SQLException e) {
+            if (!disableLogging && DBExecUtils.isExecutionCanceled(session.getDataSource(), e)) {
```
serge-rider: We cannot move QM object manipulations here.
Add function QMExecutionHandler.handleFetchError(DBCResultSet) and implement it in QMMCollectorImpl

serge-rider (changes requested):

PR #40604, opened 2026-03-25
dbeaver/pro#8825 SQL translator: add handling of ASCII in column specs

ivgag:


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/translate/SQLQueryTranslator.java
```diff
@@ -302,6 +261,64 @@ private boolean translateColumnDataType(ColumnDefinition cd, SQLDialectDDLExtens
...
+                            }
+                        } else if (extendedDialect != null) {
+                            int indexOf = cd.getColumnSpecs().indexOf(columnSpec);
+                            defChanged = true;
+                            cd.getColumnSpecs().set(indexOf, extendedDialect.getAutoIncrementKeyword());
+                        }
+                    }
+                    case "ASCII" -> {
+                        cd.getColumnSpecs().remove(columnSpec);
+                        if ("mysql".equals(targetDialect.getDialectId())) {
```
serge-rider: Lets instead create method in SQLDialectDDLExtension
String getColumnCharsetModifier(ColumnCharset charset);

enum ColumnCharset { NONE, ASCII, UTF8 }

Default implementation return null. MySQL returns CHARACTER SET LATIN1 for ASCII.

serge-rider (changes requested):

PR #40610, opened 2026-03-25
update sql server driver to 13.4.0

### LINKED ISSUE #40616: SQL Server: incorrect "Updated Rows" count after failed statement in multi-query execution
### Description

When using SQL Server with the Microsoft JDBC driver version currently bundled by DBeaver (`13.2.1.jre11`), execution statistics may report an incorrect `Updated Rows` value for a multi-statement execution that contains an error between successful statements.

In my case, DBeaver reports:

- `Updated Rows = 1`

But the correct value should be:

- `Updated Rows = 2`

### Steps to reproduce

Execute the following statements together:

```sql
insert into test_schema.table_name values (1, 'test');
insert into test_schema.table_name values (1, 'test');
insert into test_schema.table_name values (2, 'test');
select * from test_schema.table_name;
```

The second statement fails with a primary key violation, but the first and third `INSERT` statements succeed.

## Actual Result

DBeaver shows `Updated Rows = 1`.

## Expected Result

DBeaver should show `Updated Rows = 2`, because two `INSERT` statements were successfully applied.

## Additional Information

This seems to come from the SQL Server JDBC driver currently used by DBeaver rather than from DBeaver's own SQL execution logic.

Upstream JDBC issue:

- https://github.com/microsoft/mssql-jdbc/issues/2850

I tested locally with a newer driver build and the problem no longer reproduces there, but DBeaver currently still uses `13.2.1.jre11`.

## Attachment

I can provide a screenshot showing the incorrect `Updated Rows` value in the execution statistics panel.

Image

Image

### DBeaver Version

ce.26.0.1

### Operating System

_No response_

### Database and driver

Microsoft SQL Server 2016 (SP2) (KB4052908) - 13.0.5026.0 (X64)

### Steps to reproduce

_No response_

### Additional context

_No response_

EastLord:
This updates the Microsoft SQL Server JDBC driver from 13.2.1 to 13.4.0

Reason:
DBeaver currently inherits a driver bug in mixed statement execution flows:
when executing `INSERT -> INSERT(error) -> INSERT -> SELECT` via `Statement.execute()`,
the driver may swallow the valid update count after the failed statement.

In DBeaver this can lead to incorrect execution statistics, for example showing
`Updated Rows = 1` when the correct value should be `2`.

a

Upstream reference:
- https://github.com/microsoft/mssql-jdbc/issues/2850

Validation:
- Reproduced with the driver version currently used by DBeaver
- Verified that the problem no longer reproduces after upgrading to 13.4.0


E1izabeth:
@EastLord thanks for the contribution!

Could you create new issue about that and reference issue number at the beginning of the commit message?

EastLord:
> @EastLord thanks for the contribution!
> 
> Could you create new issue about that and reference issue number at the beginning of the commit message? https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code#commit-message-rules

@E1izabeth Thanks for the guidance. I created a new issue for this:
https://github.com/dbeaver/dbeaver/issues/40616

E1izabeth:
Thanks

IvanGorshechnikov (approved):

Destrolaric (approved):

PR #40634, opened 2026-03-29
#34431 Show type names in column headers

### LINKED ISSUE #34431: Show column data types under column names
I suggest to show column data types under column names.
Use this query for test:

```
select
    1::int as column1,
    2::smallint as column2,
    1.02::numeric(21,2) as column3,
    1.2::double precision as column4,
    1.6::real as column5

```

"As is" and "to be", for comparison:
изображение

Why?
1. The UI would look more strict, not so childish, teddy as it is now. After all, we are working with serious things, with DBMS.

2. The UI would look more informative. 

2.1. For example, user could see number of decimal places for data type "numeric" with arbitrary precision numbers, for example "numeric(15,2)", not just "numeric". 

2.2. User could distinguish visually between similar data types: "double precision", "float", "int", which now look the same: "123".

3. The column width will take less size (in width). It leads to more columns would fit in width, and more data can be visible for a user to get right decision based on data without need to use horizontal scroll bar or throwing away unwanted columns and re-execute the query. 

I know about a tooltip shown on mouse hover over column name. It shows column data type. But it is not so handy for a user. It's better to see this informatino on a screen without any action rather then use mouse to look at the tooltip.

seroperson:
Closes #34431 

## Summary

- Replace boolean "Show column description" preference with a 3-way dropdown: `Nothing` / `Description` / `Data type`
- When "Data type" is selected, show short DB-native type names with modifiers (like `varchar(50)`)
- Prepend "PK" for primary key columns (like `PK int4`)
- Backward compatible with old boolean preference

## Screenshots

New dropdown selector:

Screenshot 2026-03-28 030932

"Data type" option:

Screenshot 2026-03-28 030822

"Description" option:

Screenshot 2026-03-28 030839

Default:

Screenshot 2026-03-26 173118

With pre-existing option which disables icons:

Screenshot 2026-03-28 030909


IvanGorshechnikov (dismissed):

seroperson:
@IvanGorshechnikov @E1izabeth @SolovevAnton Hello! Sorry for pinging you, but would you mind reviewing this? It's 2 weeks old and contribution guide says I'm eligible for pinging now =)

IvanGorshechnikov (dismissed):

E1izabeth:
Thanks!

E1izabeth (dismissed):

E1izabeth (approved):

MashaKorax:
verified DBeaver ce win+linux

PR #40642, opened 2026-03-30
dbeaver/dbeaver#40607: fix revert functionality to open procedure

### LINKED ISSUE #40607: packaged procedure now opens in readonly view
### Description

26.0.1 introduced new behaviour. Probably linked to changelog note: 
"Oracle: improved viewing of packaged procedures. You can open a specific procedure directly instead of entire package"

For me it does not look like improvement, since previous useful behaviour was lost. 
Previous behaviour was _navigating straight to relevant code line_ in package body view. Extremely useful. Now I have to resort to Ctrl+F search...

Problem:
Image

While looking at packaged procedures in Database Navigator, double click on procedure name now opens only code for this single procedure, in a new (almost) _read_only_ tab. Exact same result as "View Procedure F4" _and_ "Go to source code" from context menu. I would expect those two actions were meant to be different?

Proposal: add option for choosing what happens on double-click. Perhaps in addition to User Interface->Navigator->Double click on node

I swear i looked, and could not find option in preferences to switch back to previous behaviour... My apologies if I missed this. Also tried to find issues/discussions where this feature was proposed, but no luck.

### DBeaver Version

Community Edition 26.0.1

### Operating System

Windows 11

### Database and driver

Oracle 19

### Steps to reproduce

1) Connect to Oracle
2) in Database Navigator expand packages node
3) double click on package (opens package body - so far so good)
4) expand package (list of procedures appears)
4) double click on packaged function/procedure (or F4, or "View Function/Procedure" from context menu, or "Go to source code" from context menu -> same result)

### Additional context

_No response_

SolovevAnton:
dbeaver/dbeaver#40607: old functionality restored.
For now, didn't revert many changes made in https://github.com/dbeaver/dbeaver/pull/40473, even though this functionality is now unused. 
Reason is since some of them might be useful in: https://github.com/dbeaver/pro/issues/8719

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.oracle.ui/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.oracle.ui/plugin.xml
@@ -210,7 +210,6 @@
             <objectType name="org.jkiss.dbeaver.ext.oracle.model.OracleTableIndex"/>
             <objectType name="org.jkiss.dbeaver.ext.oracle.model.OracleSequence"/>
             <objectType name="org.jkiss.dbeaver.ext.oracle.model.OracleTablespace"/>
-            <objectType name="org.jkiss.dbeaver.ext.oracle.model.OracleProcedurePackaged"/>
         </editor>
         <editor
                 id="org.jkiss.dbeaver.ext.oracle.ui.editors.SchedulerJobLogEditor"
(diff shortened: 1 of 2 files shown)


uslss:
verified (win, ce)

PR #40643, opened 2026-03-30
Fix Firebird embedded connection support

### LINKED ISSUE #3397: Firebird embedded driver
It would be great to have embedded driver to let users to open FB databases without server.
However it is not that easy (see #2620)

Current workaround is to manually add all needed DDLs in driver editor (as libraries).

### LINKED ISSUE #19960: Impossible to connect to Firebird Embedded
### Description

There is no way to use Firebird Embedded with DBeaver.

DBeaver always fails with 
```
Unable to complete network request to host "xnet://Global\FIREBIRD". [SQLState:08006, ISC error code:335544721]
```
Even when "Embedded" checkbox is set (this is plain wrong).

### DBeaver Version

Community Version 23.0.4.202305071722

### Operating System

Windows 10 Version 10.0.19044.2965

### Database and driver

org.firebirdsql.jdbc jaybird 4.0.0.java8

### Steps to reproduce

Steps to reproduce as reported in case #17908.

I collected every needed file for Firebird embedded in [this repository](https://github.com/fdcastel/firebird-binaries). The files in `master` are 64-bit but there is also a x86 branch with 32-bit binaries, if needed.

Just clone this repository into your local machine and (try to) connect DBeaver using any of Firebird versions there.

### Additional context

The case #17908 should not have been closed. The problem still exists (and, as it seems, nothing was done to solve it).

Even following everything described there and using latest DBeaver version it is impossible to access Firebird Embedded. 

1) The use of `embedded` as hostname and changing the URL template to `jdbc:firebirdsql:{host}:{file}` is a hack, at best. It should not be needed. The "Embedded" checkbox should work instead of this.

2) Firebird Embedded does not need `security.fdb`. I have several versions of embedded Firebird in distinct folders in my machine and can access every one of them with other platform drivers (Python, Delphi/FireDac to cite a few).

3) I believe the root cause of this problem may be that **jaybird doesn't allow you to inform the path to  `fbclient.dll`** (where the embedded server files are located). I have opened a issue for this [here](https://github.com/FirebirdSQL/jaybird/issues/745).

### LINKED ISSUE #40864: Firebird: Improve embedded database connection support
## Tasks

- ➖ **Improve embedded Firebird connection support** — Added `jaybird_embedded` driver with `jdbc:firebirdsql:embedded:{file}` URL template, JNA dependency, `nativeLibraryPath` forwarding via `jna.library.path`, and `supports-embedded-database-creation=true`. Also fixed server driver's `categories` from `"sql,embedded"` to `"sql"`.
- ➖ **Add file browser for database file path** — Firebird connections require a file path (`.fdb`). The connection dialog should have a file browser. Needs UI plugin.

Part of epic #40664.
Covered by #40643.

fdcastel:
Closes #3397, Closes #19960

## Problem

DBeaver has no way to connect to a Firebird database in embedded mode (local file, no
running server). Two issues have been open since 2017 (#3397) and 2023 (#19960)
reporting this.

The root causes were:

1. **No embedded URL template** — the only registered Firebird driver uses
   `jdbc:firebirdsql://{host}:{port}/{file}`, which always requires a server. The
   embedded form is `jdbc:firebirdsql:embedded:{file}`.

2. **No JNA dependency** — Jaybird's native/embedded protocol is JNA-based. Without
   `jna.jar` on the driver classpath the connection fails immediately.

3. **No way to specify the native library location** — Jaybird 5.x requires `jna.library.path`
   to be set to the directory containing `fbclient.dll` / `libfbclient.so` before the
   first native connection. DBeaver provided no mechanism to do this per-connection.

4. **Misleading category** — the server driver had `categories="sql,embedded"` despite
   not supporting embedded connections at all.

### New "Firebird Embedded" driver (`plugin.xml`)

A separate driver entry `jaybird_embedded` is added alongside the existing server
driver. It uses the correct URL template and includes JNA as a dependency:

```xml
<driver
    id="jaybird_embedded"
    label="Firebird Embedded"
    class="org.firebirdsql.jdbc.FBDriver"
    sampleURL="jdbc:firebirdsql:embedded:{file}"
    embedded="true"
    categories="sql,embedded">

    <file type="jar" path="maven:/org.firebirdsql.jdbc:jaybird:RELEASE[5.0.4.java11]" .../>
    <file type="jar" path="maven:/net.java.dev.jna:jna:RELEASE[5.14.0]" .../>
    ...
    <parameter name="supports-embedded-database-creation" value="true"/>
</driver>
```

The existing server driver's `categories` attribute is corrected from `"sql,embedded"` to
`"sql"`.

### Native library path forwarding (`FireBirdDataSourceProvider.java`)

`getConnectionURL()` is overridden to read the `nativeLibraryPath` driver/connection
property and forward its value to `jna.library.path` before the URL is returned. This
allows users to specify the directory containing `fbclient.dll` through the **Driver
Properties** tab without editing `dbeaver.ini` or touching system PATH.

```java
@Override
public String getConnectionURL(@NotNull DBPDriver driver, @NotNull DBPConnectionConfiguration connectionInfo) {
    String libPath = connectionInfo.getProviderProperty("nativeLibraryPath");
    if (CommonUtils.isEmptyTrimmed(libPath)) {
        libPath = CommonUtils.toString(driver.getDriverParameter("nativeLibraryPath"));
    }
    if (!CommonUtils.isEmptyTrimmed(libPath)) {
        // prepend to jna.library.path; effective before first native connection only (JNA limitation)
        ...
        System.setProperty("jna.library.path", trimmed + File.pathSeparator + existing);
    }
    return super.getConnectionURL(driver, connectionInfo);
}
```

> **Forward-compatibility note**: Jaybird 6 introduces a built-in `nativeLibraryPath`
> JDBC connection property (handled entirely by the driver). When DBeaver upgrades to
> Jaybird 6 this override can be removed, as the property name and semantics are the
> same.

## How to Use

1. **New Database Connection** → search for **"Firebird Embedded"**
2. Set **Database** to the absolute path of your `.fdb` file
3. In the **Driver Properties** tab, add:
   - `nativeLibraryPath` = `/path/to/firebird/dir` (the directory containing
     `fbclient.dll` or `libfbclient.so`)
4. Click **Test Connection**

### Requirements

- A Firebird 3.0+ installation directory (must contain `fbclient.dll` **and**
  `plugins/engine13.dll` or equivalent for your Firebird version).
- The `jaybird-fbclient` Maven artifact does **not** work here — it bundles `fbclient`
  for the `native`/`local` protocols but explicitly does not support `embedded`.

## Testing

Tested manually with:

- Firebird 5.0.2.1613 (`C:\fbodbc-tests\fb502`)
- Jaybird 5.0.4.java11 + JNA 5.14.0 (downloaded automatically by DBeaver)

Connection succeeded, system tables visible, SQL editor functional. Old server-mode
connections unaffected.

dbeaver-new-connection-firebird-embedded

dbeaver-new-connection-firebird-embedded-driver-properties

dbeaver-new-connection-firebird-embedded-connection-properties


fdcastel:
Updated commit message to conform with [Contribute your code](https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code) guidelines (issue prefix at start, AI disclosure).

IvanGorshechnikov:
Could you clarify what the difference is between `Firebird` and `Firebird Embedded` driver? They would refer to the one DataSourceProvider, so would they work the same?

fdcastel:
> Could you clarify what the difference is between `Firebird` and `Firebird Embedded` driver? 

Both drivers use the same `FireBirdDataSourceProvider`, but the differences are in `plugin.xml` configuration, not in the provider class itself. Here is a summary:



### 1. JDBC URL format (the functional core)

The most important difference is the `sampleURL` attribute, which drives what URL DBeaver generates for the connection:

| Driver | `sampleURL` | Protocol |
|---|---|---|
| `Firebird` | `jdbc:firebirdsql://{host}:{port}/{file}` | TCP/IP to a running Firebird server — Jaybird's pure-Java wire protocol, no native library needed |
| `Firebird Embedded` | `jdbc:firebirdsql:embedded:{file}` | Embedded/local file via JNA — Jaybird loads `fbclient.dll` (or `libfbclient.so`) directly in-process |

Jaybird routes the connection entirely differently based on the URL prefix. `embedded:` bypasses the TCP stack completely; the database file is opened directly via the native Firebird client library loaded by JNA.

### 2. Driver classpath (JNA dependency)

The embedded driver adds `jna-5.14.0.jar` to its classpath. The server driver does not include it because JNA is never needed for pure-Java TCP connections. Without JNA on the classpath, any attempt to use `jdbc:firebirdsql:embedded:` would fail with a `ClassNotFoundException` or `UnsatisfiedLinkError`.

### 3. UI behaviour (`embedded="true"`)

The `embedded="true"` attribute on the embedded driver tells DBeaver to present the connection dialog differently — it hides the host/port fields and shows a file picker for the database path instead, which is appropriate for a local-file protocol.

### 4. `getConnectionURL()` override — does it affect both?

Yes, both drivers go through the same `getConnectionURL()` override. However, it is effectively a no-op for the server driver: it only runs the `jna.library.path`-setting logic when the user has set a `nativeLibraryPath` property on the connection. No user of the TCP `Firebird` driver would ever set that property, so `CommonUtils.isEmptyTrimmed(libPath)` is always `true` for it, and the method falls straight through to `super.getConnectionURL()` — identical behaviour to before this PR.

The `nativeLibraryPath` forwarding exists solely to help embedded users point Jaybird at their Firebird installation directory without having to edit `dbeaver.ini` or modify the system PATH (which are the only alternatives in Jaybird 5.x).

### Summary

The two drivers are distinct connection modes in Firebird/Jaybird, not just cosmetic variants of the same thing. Sharing one `DataSourceProvider` is correct because both use the Firebird/Jaybird JDBC driver class (`org.firebirdsql.jdbc.FBDriver`) — the provider just needs to ensure the right URL and classpath are used, which the driver descriptor handles.

E1izabeth (dismissed):

IvanGorshechnikov:
Hi @fdcastel 
Your code is ok, but I don't like that solution. To fix the connection, the user should somehow know about this Driver property and go to this page to add it with a value.
The best variant for that is to add to the `Firebird Embedded` connection page one more file picker for `nativeLibraryPath`with tooltip
Ask if you have any questions about my suggestion

fdcastel:
> To fix the connection, the user should somehow know about this Driver property and go to this page to add it with a value. 

Agreed. Actually, that was just my initial idea to test whether it works.



> The best variant for that is to add to the `Firebird Embedded` connection page one more file picker for `nativeLibraryPath`with tooltip 

Yep. Much better!

Thanks for the insights. I'll work on this path in a few days (cannot do it right now).

fdcastel:
All done!

## Changes

- rebased onto latest `devel`
- Created the new `org.jkiss.dbeaver.ext.firebird.ui` plugin with:
  - `FireBirdConnectionPage` — extends `GenericConnectionPage`, adds the "Embedded Settings" tab only for the `jaybird_embedded` driver
  - `FireBirdEmbeddedConnectionPage` — an "Embedded Settings" tab with a "Native Library Path" label, text field, Browse button (directory picker), and tooltip
  - All OSGi/Maven boilerplate (MANIFEST.MF, pom.xml, build.properties, plugin.xml)
  - Registered in `plugins/pom.xml` and `features/org.jkiss.dbeaver.db.ui.feature/feature.xml`

## What to verify for this PR

1. **Firebird Embedded driver appears** — New Connection → search for "Firebird Embedded" (separate from the regular "Firebird" entry)

2. **Connection page shows the file picker** — Select "Firebird Embedded", click Next. On the connection page you should see the standard database file field. Then go to the **"Embedded Settings"** tab — it should have a "Native Library Path" field with a Browse button and a tooltip explaining the directory contains `fbclient.dll`.

3. **Round-trip persistence** — Set a path, click Finish, then Edit Connection → Embedded Settings tab → verify the path was saved.

4. **Regular Firebird driver unchanged** — Open the regular "Firebird" driver's connection page → it should **not** show the "Embedded Settings" tab.

image

_Screenshot updated as requested_

plugins/org.jkiss.dbeaver.ext.firebird.ui/src/org/jkiss/dbeaver/ext/firebird/ui/FireBirdEmbeddedConnectionPage.java
```diff
@@ -0,0 +1,111 @@
...
+import org.jkiss.dbeaver.ui.UIUtils;
+import org.jkiss.dbeaver.ui.dialogs.connection.ConnectionPageAbstract;
+import org.jkiss.utils.CommonUtils;
+
+public class FireBirdEmbeddedConnectionPage extends ConnectionPageAbstract {
+
+    private static final String PROP_NATIVE_LIBRARY_PATH = "nativeLibraryPath";
+
+    private static final String TOOLTIP_NATIVE_LIB_PATH =
+        "Directory containing the Firebird client library (fbclient.dll on Windows, " +
```
IvanGorshechnikov, line 39: Move it to localization

plugins/org.jkiss.dbeaver.ext.firebird.ui/src/org/jkiss/dbeaver/ext/firebird/ui/FireBirdEmbeddedConnectionPage.java
```diff
@@ -0,0 +1,111 @@
...
+        "libfbclient.so on Linux, libfbclient.dylib on macOS).\n" +
+        "Required for Jaybird 5.x embedded connections. Leave empty if the library " +
+        "is already on the system PATH or jna.library.path.";
+
+    private Text nativeLibPathText;
+
+    public FireBirdEmbeddedConnectionPage() {
+        setTitle("Embedded Settings");
```
IvanGorshechnikov: And this

plugins/org.jkiss.dbeaver.ext.firebird.ui/src/org/jkiss/dbeaver/ext/firebird/ui/FireBirdEmbeddedConnectionPage.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public void createControl(Composite parent) {
+        Composite container = new Composite(parent, SWT.NONE);
+        container.setLayout(new GridLayout(1, false));
+        container.setLayoutData(new GridData(GridData.FILL_BOTH));
+
+        Group group = new Group(container, SWT.NONE);
```
IvanGorshechnikov: Use `UIUtils.createTitledComposite` instead of group

IvanGorshechnikov (changes requested):

plugins/org.jkiss.dbeaver.ext.firebird.ui/src/org/jkiss/dbeaver/ext/firebird/ui/FireBirdEmbeddedConnectionPage.java
fdcastel: Moved to a new localization bundle.

plugins/org.jkiss.dbeaver.ext.firebird.ui/src/org/jkiss/dbeaver/ext/firebird/ui/FireBirdEmbeddedConnectionPage.java
fdcastel: The group title `"Native Library"` is now also in the message bundle

plugins/org.jkiss.dbeaver.ext.firebird.ui/src/org/jkiss/dbeaver/ext/firebird/ui/FireBirdEmbeddedConnectionPage.java
fdcastel: Replaced with `UIUtils.createTitledComposite`. The original 4 lines:
```java
Group group = new Group(container, SWT.NONE);
group.setText("Native Library");
group.setLayout(new GridLayout(3, false));
group.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
```
are now a single line:
```java
Composite group = UIUtils.createTitledComposite(container, FireBirdUIMessages.page_embedded_group_native_library, 3, GridData.FILL_HORIZONTAL);
```
The `Group` SWT import was removed; `FireBirdUIMessages` import added.

fdcastel:
@IvanGorshechnikov The remaining hardcoded strings in `FireBirdEmbeddedConnectionPage.java` have *not* been localized. Let me know if you’d like me to move those as well.

plugins/org.jkiss.dbeaver.ext.firebird.ui/src/org/jkiss/dbeaver/ext/firebird/ui/FireBirdEmbeddedConnectionPage.java
IvanGorshechnikov: But you didn't move it

plugins/org.jkiss.dbeaver.ext.firebird.ui/src/org/jkiss/dbeaver/ext/firebird/ui/FireBirdEmbeddedConnectionPage.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public void createControl(Composite parent) {
+        Composite container = new Composite(parent, SWT.NONE);
+        container.setLayout(new GridLayout(1, false));
+        container.setLayoutData(new GridData(GridData.FILL_BOTH));
+
+        Composite group = UIUtils.createTitledComposite(container, FireBirdUIMessages.page_embedded_group_native_library, 3, GridData.FILL_HORIZONTAL);
+
+        UIUtils.createControlLabel(group, "Native Library Path");
```
IvanGorshechnikov: Localize it please too

plugins/org.jkiss.dbeaver.ext.firebird.ui/src/org/jkiss/dbeaver/ext/firebird/ui/FireBirdEmbeddedConnectionPage.java
IvanGorshechnikov: Could you update screenshot after this change?

fdcastel: done!

plugins/org.jkiss.dbeaver.ext.firebird.ui/src/org/jkiss/dbeaver/ext/firebird/ui/FireBirdEmbeddedConnectionPage.java
fdcastel: Done in 931fd3cf — tooltip moved to `page_embedded_label_native_library_path_tip` in the message bundle.

plugins/org.jkiss.dbeaver.ext.firebird.ui/src/org/jkiss/dbeaver/ext/firebird/ui/FireBirdEmbeddedConnectionPage.java
fdcastel: Done in 931fd3cf — moved to `page_embedded_label_native_library_path`. Also localized the remaining strings (Browse, dialog title/message).

fdcastel:
Tracking issue: #40864

IvanGorshechnikov (approved):

E1izabeth (approved):

fdcastel:
Guys, is something holding this up?

E1izabeth:
Hi, @fdcastel!
I see that it's "In progress of testing". Thanks for bringing it up. I'll ask our QA team why it's taking so long to check this PR and let you know.

Matvey16:
Verified

PR #40644, opened 2026-03-30
Fix Firebird EXECUTE BLOCK with ? parameters

### LINKED ISSUE #37871: Firebird EXECUTE BLOCK not working with parameters
### Description

Trying to run this simple EXECUTE BLOCK (from [Firebird Documentation](https://firebirdsql.org/file/documentation/html/en/refdocs/fblangref50/firebird-50-language-reference.html#fblangref50-dml-execblock-exmpl))

```sql
EXECUTE BLOCK (x DOUBLE PRECISION = ?, y DOUBLE PRECISION = ?)
RETURNS (gmean DOUBLE PRECISION)
AS
BEGIN
  gmean = SQRT(x*y);
  SUSPEND;
END
```

Throws the error:

```
SQL Error [337248300] [07008]: Invalid number of parameters, expected 2, got 0 [SQLState:07008, ISC error code:337248300]
```

Even with "Anonymous SQL Parameters" set:

Image

### DBeaver Version

Community 25.0.3

### Operating System

Windows 11

### Database and driver

Database: `Firebird 3.0`
Driver: `org.firebirdsql.jdbc:jaybird:RELEASE [5.0.4.java11]`

### Steps to reproduce

1. Go to Preferences / Editors / SQL Editor / SQL Processing. In "Parameters" section, check "Anonymous SQL parameters";
2. Connect to a Firebird database;
3. Run the following SQL:
    ```sql
    EXECUTE BLOCK (x DOUBLE PRECISION = ?, y DOUBLE PRECISION = ?)
    RETURNS (gmean DOUBLE PRECISION)
    AS
    BEGIN
      gmean = SQRT(x*y);
      SUSPEND;
    END
    ```
4. A dialog will open asking for the parameter values. Enter x = 3, y = 4:
    Image
5. Click "ok". The following error will appear:
    
    Image
    
    ```
    SQL Error [335544634] [42000]: Dynamic SQL Error; SQL error code = -104; Token unknown - line 1, column 37; 3 [SQLState:42000, ISC error code:335544634]
    ```
        

### Additional context

For reference: An EXECUTE BLOCK without parameters works correctly:

```sql
EXECUTE BLOCK
RETURNS (gmean DOUBLE PRECISION)
AS
BEGIN
  gmean = SQRT(3*4);
  SUSPEND;
END
```

Output: 

Image

### LINKED ISSUE #40853: Firebird: Support ? parameter binding in EXECUTE BLOCK
## Tasks

- 👍 **`EXECUTE BLOCK` with `?` parameter binding** — Implemented `needsNativeParameterBinding()` dialect hook. When an `EXECUTE BLOCK (` query contains `?` placeholders, DBeaver now shows the parameter dialog and binds values via `PreparedStatement.setString()` instead of text substitution.

Part of epic #40664.
Covered by #40644.

fdcastel:
Closes #37871

## Problem

Firebird's `EXECUTE BLOCK` statement accepts input parameters via JDBC `?`
placeholders in the parameter-definition clause:

```sql
EXECUTE BLOCK (x DOUBLE PRECISION = ?, y DOUBLE PRECISION = ?)
RETURNS (gmean DOUBLE PRECISION)
AS BEGIN
  gmean = SQRT(x * y);
  SUSPEND;
END
```

Running this in DBeaver fails with one of two errors depending on configuration:

- `Invalid number of parameters, expected 2, got 0` — the `?` markers reach
  Firebird unbound because DBeaver never prompted for values.
- `Token unknown - line 1, column 37; 3` — DBeaver prompted for values but
  replaced `?` with the literal `3`, which Firebird rejects; only a JDBC `?`
  placeholder is valid in that position.

As confirmed by [@mrotteveel](https://github.com/mrotteveel) in the issue
thread: *"The value must be sent as a parameter to a prepared statement."*

## Root cause

Two independent problems compound each other:

**1. Parameter detection skipped.** `EXECUTE` is listed in `DDL_KEYWORDS` (to
support `EXECUTE PROCEDURE`). `SQLScriptParser.parseParametersAndVariables()`
skips `?`-detection for DDL queries, so no parameter dialog appears and the raw
`?` is forwarded to Firebird.

**2. Text substitution is incompatible.** Even when the parameter dialog
appeared (with "Anonymous SQL parameters" preference enabled), DBeaver replaced
`?` with a literal value. Firebird's DSQL parser does not accept literals in
`EXECUTE BLOCK` parameter slots — only `?` is valid.

## Solution

Introduced a dialect hook `SQLDialect.needsNativeParameterBinding(String query)`
(default `false`). `FireBirdSQLDialect` overrides it to return `true` when the
query matches `EXECUTE BLOCK (`.

When the hook returns `true`:

1. **`SQLScriptParser`** runs a dedicated `parseNativeBindingParameters()` pass
   that detects `?` directly in the token stream, bypassing the DDL-skip logic
   and the "anonymous parameters" user preference. This ensures the parameter
   dialog always appears for `EXECUTE BLOCK (`.

2. **`SQLScriptContext`** sets `SQLQuery.nativeParameterBinding = true` and
   skips text substitution in `fillQueryParameters()`.

3. **`SQLQueryJob` / `SQLQueryDataContainer`** use `DBCStatementType.QUERY`
   (which produces a `PreparedStatement`) and bind each parameter value via
   `ps.setString()` / `ps.setNull()` rather than embedding it in the SQL text.

## Testing

Manually verified against Firebird 5.0:

```sql
-- Should show parameter dialog for x and y; result: GMEAN = 3.4641016151377544
EXECUTE BLOCK (x DOUBLE PRECISION = ?, y DOUBLE PRECISION = ?)
RETURNS (gmean DOUBLE PRECISION)
AS BEGIN
  gmean = SQRT(x * y);
  SUSPEND;
END
```

image

Regression tests: all existing unit tests pass (`mvn verify`, 0 failures).

No change in behaviour for other databases — `needsNativeParameterBinding()`
returns `false` by default so all non-Firebird and non-`EXECUTE BLOCK` paths
are unchanged.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.firebird/src/org/jkiss/dbeaver/ext/firebird/model/FireBirdSQLDialect.java
+++ b/plugins/org.jkiss.dbeaver.ext.firebird/src/org/jkiss/dbeaver/ext/firebird/model/FireBirdSQLDialect.java
@@ -28,8 +28,16 @@ import org.jkiss.dbeaver.model.impl.jdbc.JDBCDataSource;
 import org.jkiss.dbeaver.model.struct.rdb.DBSProcedure;
 
 import java.util.Arrays;
+import java.util.regex.Pattern;
 
 public class FireBirdSQLDialect extends GenericSQLDialect {
+
+    // EXECUTE BLOCK (param = ?) requires a '?' placeholder in the parameter slot —
+    // Firebird's DSQL parser rejects a literal there. EXECUTE PROCEDURE p(?) also
+    // has to skip DBeaver's anonymous-parameter preference gate so '?' is bound
+    // out-of-the-box. Both need native JDBC binding instead of text substitution.
+    private static final Pattern EXECUTE_BLOCK_OR_PROCEDURE =
+        Pattern.compile("^\\s*EXECUTE\\s+(BLOCK|PROCEDURE)\\b", Pattern.CASE_INSENSITIVE | Pattern.DOTALL);
 
     private boolean supportsAsBeforeTableAlias = true;
 
@@ -311,5 +319,10 @@ public class FireBirdSQLDialect extends GenericSQLDialect {
     @Override
     public DBDBinaryFormatter getNativeBinaryFormatter() {
         return BinaryFormatterHexString.INSTANCE;
+    }
+
+    @Override
+    public boolean needsNativeParameterBinding(@NotNull String queryText) {
+        return EXECUTE_BLOCK_OR_PROCEDURE.matcher(queryText).find();
     }
 }
--- a/plugins/org.jkiss.dbeaver.ext.firebird/src/org/jkiss/dbeaver/ext/firebird/model/FireBirdSQLDialect.java
+++ b/plugins/org.jkiss.dbeaver.ext.firebird/src/org/jkiss/dbeaver/ext/firebird/model/FireBirdSQLDialect.java
@@ -28,8 +28,12 @@ import org.jkiss.dbeaver.model.impl.jdbc.JDBCDataSource;
 import org.jkiss.dbeaver.model.struct.rdb.DBSProcedure;
 
 import java.util.Arrays;
+import java.util.regex.Pattern;
 
 public class FireBirdSQLDialect extends GenericSQLDialect {
+
+    private static final Pattern EXECUTE_BLOCK_WITH_PARAMS =
+        Pattern.compile("^\\s*EXECUTE\\s+BLOCK\\s*\\(", Pattern.CASE_INSENSITIVE | Pattern.DOTALL);
 
     private boolean supportsAsBeforeTableAlias = true;
 
@@ -148,5 +152,10 @@ public class FireBirdSQLDialect extends GenericSQLDialect {
     @Override
     public DBDBinaryFormatter getNativeBinaryFormatter() {
         return BinaryFormatterHexString.INSTANCE;
+    }
+
+    @Override
+    public boolean needsNativeParameterBinding(@NotNull String queryText) {
+        return EXECUTE_BLOCK_WITH_PARAMS.matcher(queryText).find();
     }
 }
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/SQLScriptContext.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/SQLScriptContext.java
@@ -345,6 +345,15 @@ public class SQLScriptContext implements DBCScriptContext {
             }
         }
 
+        // Check if dialect requires native parameter binding for this query
+        DBCExecutionContext execCtx = getExecutionContext();
+        if (execCtx != null) {
+            SQLDialect dialect = execCtx.getDataSource().getSQLDialect();
+            if (dialect.needsNativeParameterBinding(query.getText())) {
+                query.setNativeParameterBinding(true);
+            }
+        }
+
         SQLUtils.fillQueryParameters(query, parameters);
 
         return true;
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/data/SQLQueryDataContainer.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/data/SQLQueryDataContainer.java
@@ -34,6 +34,8 @@ import org.jkiss.dbeaver.model.struct.DBSDataContainer;
 import org.jkiss.dbeaver.model.struct.DBSObject;
 import org.jkiss.utils.CommonUtils;
 
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
@@ -138,11 +140,18 @@ public class SQLQueryDataContainer implements DBSDataContainer, SQLQueryContaine
         try (final DBCStatement dbcStatement = DBUtils.makeStatement(
             source,
             session,
-            DBCStatementType.SCRIPT,
+            sqlQuery.isNativeParameterBinding() ? DBCStatementType.QUERY : DBCStatementType.SCRIPT,
             sqlQuery,
             firstRow,
             maxRows))
         {
+            if (sqlQuery.isNativeParameterBinding() && dbcStatement instanceof PreparedStatement ps) {
+                try {
+                    SQLUtils.bindNativeParameters(ps, sqlQuery);
+                } catch (SQLException e) {
+                    throw new DBException("Failed to bind native parameters", e);
+                }
+            }
             DBExecUtils.setStatementFetchSize(dbcStatement, firstRow, maxRows, fetchSize);
 
             // Execute statement
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/data/SQLQueryDataContainer.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/data/SQLQueryDataContainer.java
@@ -34,7 +34,11 @@ import org.jkiss.dbeaver.model.struct.DBSDataContainer;
 import org.jkiss.dbeaver.model.struct.DBSObject;
 import org.jkiss.utils.CommonUtils;
 
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Types;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 /**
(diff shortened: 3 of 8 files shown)


fdcastel:
### Added: **Named parameter** support for Firebird EXECUTE BLOCK

The initial fix handled only positional (`?`) parameters. This update extends native parameter binding to also support **named parameters** (e.g. `:x`, `:y`).

**Example that now works:**

```sql
EXECUTE BLOCK (x DOUBLE PRECISION = :x, y DOUBLE PRECISION = :y)
RETURNS (gmean DOUBLE PRECISION)
AS
BEGIN
  gmean = SQRT(x * y);
  SUSPEND;
END
```

**Changes (2 files):**

1. **`SQLScriptParser.parseNativeBindingParameters()`** — When the scanner classifies a token as `T_PARAMETER` (e.g. `:x`), the named-parameter prefix (`:`) is now stripped so the parameter dialog displays the bare variable name (`x`) and lookup matches EXECUTE BLOCK variable declarations correctly.

2. **`SQLUtils.fillQueryParameters()`** — When `
ativeParameterBinding` is enabled and the SQL contains named parameters (e.g. `:x` instead of `?`), those tokens are replaced with positional `?` placeholders before sending the query to JDBC. Values are then bound positionally — identical to the existing `?` parameter path.

Both positional (`?`) and named (`:name`) syntax continue to work correctly.

image

fdcastel:
Updated to conform with [Contribute your code](https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code) guidelines:
- Fixed commit message format (issue prefix at start)
- Squashed commits into one
- Fixed line exceeding 140-character limit in SQLScriptParser.java
- Added AI disclosure

IvanGorshechnikov:
Your PR adds lots of Firebird-specific code to the generic model classes. Please, don't do this
The main root of your problem with the parser is that if I execute something like `select ?`, it will work correctly

image

fdcastel:
> Your PR adds lots of Firebird-specific code to the generic model classes. Please, don't do this 

Agreed, @IvanGorshechnikov -- and apologies for that. I’m still getting familiar with this codebase, and I should have brought this up earlier to ask for guidance.

I’ve converted the PR to draft and will spend the next few days reworking it.

In the meantime, I’d really appreciate any suggestions on the best way to approach this implementation.



> The main root of your problem with the parser is that if I execute something like `select ?`, it will work correctly

I’m not sure I fully understand your last point—could you clarify what you mean here? Or it was a question?

fdcastel:
@IvanGorshechnikov — agreed, and I apologize for polluting generic core classes. I have been re-thinking the approach.

Looking at `SQLScriptParser.parseParametersAndVariables()`, the DDL check is already dialect-aware — it calls `sqlDialect.getDDLKeywords()`. `FireBirdSQLDialect` already overrides `getDDLKeywords()`. So the fix could live entirely within `FireBirdSQLDialect` without touching any generic code:

**Proposed approach:** Remove `EXECUTE` from `FireBirdSQLDialect.DDL_KEYWORDS`.

Neither `EXECUTE BLOCK` nor `EXECUTE PROCEDURE` is truly DDL — they are procedural/DML constructs. With `EXECUTE` no longer classified as DDL, the standard `?`-detection will run for `EXECUTE BLOCK (x = ?)`, and the existing JDBC parameter binding path handles the rest correctly.

Before implementing this, a few questions:

1. Is removing `EXECUTE` from the Firebird DDL keywords the right call, or is there a reason it was added there?
2. Is there any concern about `EXECUTE PROCEDURE` behavior changing as a result?
3. Your earlier screenshot showed `SELECT ?` working correctly — was that a hint that the fix should follow a similar (non-DDL) code path?

Happy to implement once I have your guidance.

fdcastel:
Thank you @IvanGorshechnikov — your `SELECT ?` hint pointed exactly the right way. The generic parameter-detection path already handles this correctly; the only thing standing in the way was that `EXECUTE` was classified as a DDL keyword in `FireBirdSQLDialect.DDL_KEYWORDS`, which made `SQLScriptParser.parseParametersAndVariables()` skip `?`-detection for every `EXECUTE BLOCK (...)` statement.

`EXECUTE` isn't really DDL in Firebird (both `EXECUTE BLOCK` and `EXECUTE PROCEDURE` are procedural/DML), so removing it from the dialect's DDL list is the principled fix.

I've reworked the PR accordingly:

- All changes to generic core classes have been reverted (`SQLDialect`, `SQLQuery`, `SQLUtils`, `SQLScriptContext`, `SQLScriptParser`, `SQLQueryDataContainer`, `SQLQueryJob`).
- The entire fix is now a single line in `FireBirdSQLDialect`: drop `"EXECUTE"` from `DDL_KEYWORDS`.
- The branch has also been rebased onto the latest `devel`.

Diff shrank from **+241 / −2 across 8 files** to **+1 / −1 in one Firebird-only file**.

Moving out of draft — ready for your review whenever you have a moment.

E1izabeth (dismissed):
Thanks!

fdcastel:
Found a problem that the tests didn't catch. Working on it...

fdcastel:
@IvanGorshechnikov updates after trying the path you hinted at.



### First attempt — the one-line fix. 

 - Removing `EXECUTE` from `FireBirdSQLDialect.DDL_KEYWORDS` alone. I shipped that, built a pre-release, and ran an SWTBot UI-test harness against it. 
   - Result: the parameter dialog appears for `:name` but the query still fails end-to-end, and for `?` the dialog never appears at all. 
   - Same underlying problem in both cases — `SQLUtils.fillQueryParameters()` substitutes literals (`?` → `3`, `:x` → `9`) and Firebird's DSQL grammar rejects a literal in the `EXECUTE BLOCK (var = …)` slot with `Token unknown — line 1, column N; <value>`. That's the "Root cause #2" the original PR description called out; it doesn't go away by adjusting the DDL classification because literal substitution runs regardless.

So some generic-code change is unavoidable for this fix — the question was **how small it could be**.



### Second attempt — this push.

  - Kept the principled hook idea (`SQLDialect.needsNativeParameterBinding(query)`, default `false`) but trimmed the plumbing:
    - **Rebased onto latest `devel`.**
    - **+148 / −5 across 7 files** (was +241 / −2 across 8 files — **~40% smaller**).
    - Dropped the duplicated `parseNativeBindingParameters()` method (~100 lines); the existing `parseParametersAndVariables()` loop now has a single extra branch that scans non-string/non-comment tokens for `?` when the hook returns `true`.
    - Dropped the `SQLScriptContext` change; the flag is set once in `SQLScriptParser` where parsing already happens, and `SQLUtils.fillQueryParameters()` reads it.
    - Extracted the native binding into `SQLUtils.bindNativeParameters()`, shared by both `SQLQueryJob` and `SQLQueryDataContainer`.
    - Firebird-side is ~10 lines: one regex (`EXECUTE (BLOCK|PROCEDURE)`) and the hook override.

**What the hook does, one sentence each:**
  1. `SQLScriptParser` — detect `?` regardless of the `anonymous parameters` preference, and bypass the DDL / execute-keyword skips for hook'd queries.
  2. `SQLUtils.fillQueryParameters` — skip literal substitution; rewrite any `:x` → `?` so JDBC positional binding lines up.
  3. `SQLQueryJob` / `SQLQueryDataContainer` — use `DBCStatementType.QUERY` (produces a `PreparedStatement`) and bind values via the shared helper.

> **Non-Firebird dialects see zero behavior change**:  the hook returns `false` by default. Firebird queries that aren't `EXECUTE BLOCK` / `EXECUTE PROCEDURE` are also unaffected.

## Testing

Built an [unofficial pre-release](https://github.com/fdcastel/dbeaver/releases) from my fork, and ran [an SWTBot UI-test](https://github.com/fdcastel/dbeaver-ui-test) harness against it. 

Three shapes covered, matching the scenarios from the issue plus an implicit regression guard:

- `EXECUTE BLOCK (x = ?, y = ?)` — positional, the main issue
- `EXECUTE BLOCK (x = :x, y = :y)` — named
- `EXECUTE PROCEDURE SP_GET_CUSTOMER(?)` — stored procedure invocation

**All three pass on this branch and fail on `devel` (verified both directions).** Assertions scan the Results panel, not just modal dialogs, so "dialog appeared but query silently failed" no longer produces a false positive.

Happy to split this into smaller commits if that helps review, or to shrink it further if specific lines look unnecessary. 

The goal was to keep the generic change principled and self-contained while genuinely closing #37871.

fdcastel:
One deliberate scope note: the original PR's regex matched only `EXECUTE BLOCK`.

I widened it to `EXECUTE (BLOCK|PROCEDURE)` because `EXECUTE PROCEDURE p(?)` hits the same tokenizer-level preference gate — the `?` is never detected by default, so the user has to enable "Anonymous parameters" just to get a bind dialog. 

Same mechanism, one-word regex change, and it's a no-op for procedure calls without parameters. 

Happy to drop `PROCEDURE` back out if you'd prefer to keep this PR strictly to `EXECUTE BLOCK` — the `EXECUTE PROCEDURE` case is arguably a separate issue.

E1izabeth (changes requested):
We discussed the proposed changes and agreed that it would be worth creating a new way to bind parameters for statements that require more than just inline replacement. Our current approach with anonymous parameters is not suitable for this scenario, as it only allows for simple substitution.

In the EXECUTE BLOCK with parameters, we need a more flexible solution that allows for complex logic and multiple values. Therefore, it's better to create a new parameter-binding mechanism that better suits this use case.

Additionally, we do not currently execute prepared statements in the SQL editor within our workflow.

fdcastel:
Thank you, @E1izabeth, for sharing your insights.

I agree that the current situation isn’t ideal either.

One question: **Would you like me to draft a proposal (perhaps in a separate PR)?**; or is this something you’re already planning to address? If so, is there a way for me to follow its progress?

I use DBeaver regularly in my day-to-day work, and this PR would resolve a significant issue I encounter. I completely understand if my proposed solution isn’t viable or if you’re planning a different refactor. In that case, I’d just appreciate being kept in the loop.

E1izabeth:
@fdcastel thanks! Yes, I'm going to discuss it  and let you know if we can fit this issue to the nearest plans or we will try to guide you.

E1izabeth:
We have discussed this feature request further. We can implement it ourselves. However, we cannot take it earlier than July.

fdcastel:
@E1izabeth, just a quick nudge 

Has this work progressed? Is there anything I can help with?

E1izabeth:
Hi, @fdcastel! Thanks for the follow-up. We plan to get this to work in the next milestone

PR #40647, opened 2026-03-30
#40210 Allow count query for non-standard SELECT statements

### LINKED ISSUE #40183: SQL parser error while calculating the row count in SQL result grid
### Description

Given an SQL query with particular PostgreSQL syntax, calculating the row count in SQL result grid fails with an error.

### DBeaver Version

CE 25.3.3

### Operating System

Windows 11

### Database and driver

PostgreSQL 17.5
Driver version: 42.7.2

### Steps to reproduce

1. Open a new SQL script tab and enter the following query:
```
with t as (select n from generate_series(1, 500) s(n))
select n from generate_series(1, 1000) s(n)
where n in (table t)
```
2. Execute the query. The first 200 rows will be fetched.

3. Click on the "Calculate total row count" button under the SQL result grid

4. The operation fails with the following error:

```
Error executing row count
  Error parsing SQL query: net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "t" <S_IDENTIFIER>
    at line 3, column 19.

Was expecting one of:

    ")"
    net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: "t" <S_IDENTIFIER>
    at line 3, column 19.

Was expecting one of:

    ")"
      Encountered unexpected token: "t" <S_IDENTIFIER>
    at line 3, column 19.

Was expecting one of:

    ")"
```

### Additional context

An equivalent SQL with the following syntax:

```
with t as (select n from generate_series(1, 500) s(n))
select n from generate_series(1, 1000) s(n)
where n in (select n from t)
```

...works fine in the same scenario. So, it appears as if the parser gets tripped up by the unconventional (but legal) syntax of the first SQL query while attempting to get the row count, even if the first 200 rows were successfully fetched.

### LINKED ISSUE #40210: "Calculate total row count" fails to identify a correct SELECT statement ("Cannot calculate row count for non-SELECT statement")
### Description

Trying to run a query like 

```sql
-- DuckDB
FROM "/temp/lego_sets.parquet"
```

with DuckDb driver works and it fetches the first 10000 rows.

However, when clicking in the "Calculate total row count" button 

Image

it raises the error:

Image

In DuckDb dialect this is a valid SELECT statement called [FROM-First Syntax](https://duckdb.org/docs/stable/sql/query_syntax/from#from-first-syntax).

### DBeaver Version

Community Edition Version 25.3.3.202601181752

### Operating System

Windows 11

### Database and driver

DuckDb v1.4.4
duckdb_jdbc-1.4.4.0.jar
(latest version)

### Steps to reproduce

1. Create a DuckDb connection
2. Run the following query
    ```sql
    FROM "https://raw.githubusercontent.com/fdcastel/lego-sets/refs/heads/master/lego_sets.csv"
    ```
3. Click in "Calculate total row count" button 

### Additional context

Detecting whether a SQL command is a SELECT statement is not a simple task. There are many edge cases, for example:

- The DuckDB-specific FROM first clause mentioned above is just one such case.
- Queries that use CTEs start with WITH, not SELECT.
- Some databases, such as Firebird, can return row sets from stored procedures.

I believe DBeaver should avoid trying to determine whether a query is a SELECT statement at all.

A better approach would be to assume it is, attempt to execute a `SELECT COUNT(*) FROM (...)` around it, and only if that fails, treat the statement as a non-SELECT query.

### LINKED ISSUE #40515: "Calculate total row count" returns error
### Description

**Calculate total row count** returns error.

### DBeaver Version

Community Edition 26.0.0

### Operating System

Windows

### Database and driver

ClickHouse with the latest driver version.

### Steps to reproduce

Run query:
`SELECT *
FROM (SELECT number FROM numbers(10)) t1
	JOIN (
		SELECT number FROM numbers(10)
	) t2 
	USING number `

Click "Calculate total row count":
Image

### Additional context

_No response_

### LINKED ISSUE #40653: Calculate total row count: Encountered unexpected token: "@" "@"
### Description

When using the "Calculate total row count" button, the following error occurs:
~~~
Encountered unexpected token: "@" "@"
at line 6, column 17.

Was expecting one of:

"!"
~~~~
### DBeaver Version

CE 26.0.1.202603221810

### Operating System

Windows 10

### Database and driver

Database name and version: PostgreSQL 16.3
Driver name and version: PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

1) Execute this query:
~~~ sql
         select *, cost-discount,  abs(cost-discount)
           from (
                  values 
                    (1, 100.00, 70.00),
                    (1, 10.00, 70.00)
                ) as tab (id, cost, discount)
          where @(cost-discount) > 0
~~~

2) And click the button:  "Calculate total row count"

### Additional context

If executed this way, the SQL query produces the same result, but clicking the button does not display an error.
~~~ sql
         select *, cost-discount,  abs(cost-discount)
           from (
                  values 
                    (1, 100.00, 70.00),
                    (1, 10.00, 70.00)
                ) as tab (id, cost, discount)
          where abs(cost-discount) > 0
~~~

fdcastel:
Closes #40210

## Problem

When using non-standard SQL syntax that returns result sets — such as DuckDB's `FROM`-first syntax (`FROM table_name`) or PostgreSQL's `TABLE` command (`TABLE my_table`) — clicking **Calculate total row count** produced: *Cannot calculate row count for non-SELECT statement*.

This happened because `SQLQueryTransformerCount.wrapSourceQuery()` rejected any query that JSQLParser did not parse as a `PlainSelect`, even though these are valid result-returning statements in their respective dialects.

## Fix

Instead of throwing an exception for non-`PlainSelect` statements or unparseable queries, the method now wraps the original query text in `SELECT COUNT(*) FROM (...)` and lets the database decide whether the query is valid.

This also handles the case where JSQLParser throws a parse exception for non-standard syntax (e.g., DuckDB's `FROM`-first) — the query is still wrapped and forwarded to the database.

## Changes

- `plugins/org.jkiss.dbeaver.model.sql/.../transformers/SQLQueryTransformerCount.java`
  - Removed the `throw new DBCException(...)` for non-`PlainSelect` statements
  - Wrapped the `parseQuery()` call in try-catch so unparseable SQL is handled gracefully
  - In both cases, the original query is wrapped in `SELECT COUNT(*) FROM (...)`

## Testing

- All existing tests pass (`org.jkiss.dbeaver.model.sql.test` and all other test modules)
- Full `mvn clean package -Pproduct-dbeaver-ce` build succeeds

Manual test steps:
1. Launch the built DBeaver.
2. Create a new DuckDB connection (or use an existing one).
3. Open an SQL editor and run:
   ```sql
   FROM "https://raw.githubusercontent.com/fdcastel/lego-sets/refs/heads/master/lego_sets.csv"
   ```
4. Verify the query executes and returns rows.
5. Click the **"Calculate total row count"** button in the results panel (the button with a `#` / row count icon in the bottom status bar of the result set).
6. **Expected:** The total row count is calculated and displayed without error.
7. **Before fix:** Error dialog: *"Cannot calculate row count for non-SELECT statement"*.


fdcastel:
Updated commit message to conform with [Contribute your code](https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code) guidelines (issue prefix at start of commit message, AI disclosure added).

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/transformers/SQLQueryTransformerCount.java
```diff
@@ -75,14 +75,21 @@ private SQLQuery wrapSourceQuery(
...
             }
-        } else {
-            throw new DBCException("Cannot calculate row count for non-SELECT statement '" + query.getText() + "'");
+            // For non-PlainSelect statements (e.g. DuckDB FROM-first syntax, PostgreSQL TABLE command,
+            // CTEs starting with WITH), still attempt to wrap in SELECT COUNT(*) and let the database
+            // decide if it's valid (#40210)
+        } catch (DBCException e) {
+            // Query could not be parsed (e.g. uses non-standard syntax like DuckDB FROM-first).
+            // Still attempt to wrap it and let the database handle validation (#40210).
+            log.debug("Could not parse query for count transformation, wrapping as-is: " + e.getMessage());
```
E1izabeth: Please modify debug log message
```suggestion
            log.debug("Could not parse query to remove orderings, wrapping as-is: " + e.getMessage());
```

E1izabeth (changes requested):

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/transformers/SQLQueryTransformerCount.java
```diff
@@ -75,14 +75,21 @@ private SQLQuery wrapSourceQuery(
...
             }
-        } else {
-            throw new DBCException("Cannot calculate row count for non-SELECT statement '" + query.getText() + "'");
+            // For non-PlainSelect statements (e.g. DuckDB FROM-first syntax, PostgreSQL TABLE command,
+            // CTEs starting with WITH), still attempt to wrap in SELECT COUNT(*) and let the database
+            // decide if it's valid (#40210)
+        } catch (DBCException e) {
+            // Query could not be parsed (e.g. uses non-standard syntax like DuckDB FROM-first).
+            // Still attempt to wrap it and let the database handle validation (#40210).
+            log.debug("Could not parse query to remove orderings, wrapping as-is: " + e.getMessage());
```
ShadelessFox: Let's not omit the original exception
```suggestion
            log.debug("Could not parse query to remove orderings, wrapping as-is: ", e);
```

fdcastel: Good catch, @ShadelessFox! My bad. 😅

ShadelessFox (approved):

E1izabeth (approved):

E1izabeth:
Thanks for the contribution!

Matvey16:
Verified

PR #40656, opened 2026-03-31
dbeaver/pro#8660 Change databricks icons

IvanGorshechnikov:
Closes https://github.com/dbeaver/pro/issues/8660


plugins/org.jkiss.dbeaver.ext.databricks/plugin.xml
```diff
@@ -14,7 +14,7 @@
                 id="databricks"
                 parent="generic"
                 label="Databricks"
-                icon="platform:/plugin/org.jkiss.dbeaver.ext.generic/icons/databricks_icon.png"
+                icon="icons/Databricks_icon.svg"
```
E1izabeth: shouldn't it be so?

```suggestion
                icon="icons/databricks_icon.svg"
```

IvanGorshechnikov: fixed

E1izabeth (approved):

PR #40668, opened 2026-04-01
dbeaver/pro#8896 Drop h2_server_v3 driver and update h2_server to 2.4.x

vkirst:


plugins/org.jkiss.dbeaver.ext.h2/plugin.xml
```diff
@@ -38,17 +38,19 @@
...
 
-                    <file type="jar" path="drivers/h2_v2" bundle="drivers.h2_v2"/>
-                    <file type="license" path="drivers/h2_v2/LICENSE.txt" bundle="drivers.h2_v2"/>
+                    <file type="jar" path="drivers/h2_v3" bundle="drivers.h2_v3"/>
+                    <file type="license" path="drivers/h2_v3/LICENSE.txt" bundle="drivers.h2_v3"/>
 
                     <parameter name="query-get-active-db" value="SELECT SCHEMA()"/>
                     <parameter name="query-set-active-db" value="SET SCHEMA = ?"/>
                     <parameter name="active-entity-type" value="schema"/>
+                    <parameter name="query-shutdown" value="shutdown"/>
```
serge-rider: We don't need shutdown parameter for non-embedded drivers

serge-rider (changes requested):

PR #40552, opened 2026-03-18
#40537 http-request-timeout

dev-miro26:
Fixes #40537 

## Problem
Users behind slow corporate proxies or firewalls sometimes hit timeout errors when DBeaver attempts to download driver JARs or fetch Maven repository metadata.
The timeout was previously hardcoded at 10 seconds with no way to override it without patching the source.
The new setting is exposed at **Window → Preferences → Connections → Drivers → Network settings → Driver download timeout (seconds)** (range: 1–300 s, default: 10 s).
For headless or automated environments the value can also be set without the UI:
- System property: `-Ddbeaver.http.request.timeout=<seconds>`
- Environment variable: `DBEAVER_HTTP_REQUEST_TIMEOUT=<seconds>`
Both are checked before the preference value. An invalid (non-numeric) value is logged as a warning and the next source is tried.

## Changes
- **`ModelPreferences`** — Added `UI_DRIVERS_UPDATE_TIMEOUT` preference key  (`"ui.http.request.timeout"`, stored in ms, default 10 000)
- **`DriverUtils`** — Added `getDownloadTimeout()` which resolves the effective  `Duration` from system property → environment variable → preference → hardcoded
  default; this is the single authoritative source for all driver download callers
- **`WebUtils`** — Removed the driver-specific `getDefaultHttpRequestTimeout()` helper (model classes should not hold driver-specific logic); `downloadRemoteFile` gained explicit `Duration timeout` overloads while keeping the existing no-arg signatures (used by the app-update path) unchanged
- **`PrefPageDrivers`** — New **Network settings** group added between _HTTP Proxy_  and _Drivers location_, containing the timeout spinner
- **`PrefPageNetworkExpert`** — Timeout spinner removed (it belonged with driver settings, not global network expert settings)
- **`MavenArtifact`**, **`MavenArtifactVersion`**, **`DriverLibraryAbstract`**, **`NativeClientDistributionDescriptor`**,
 All driver download call sites updated to pass `DriverUtils.getDownloadTimeout()`

## Testing
1. Open **Preferences → Connections → Drivers** and verify the **Network settings**
2. Set **Driver download timeout** to `1` second, click **Apply and Close**
3. Create a new connection that requires a driver download — a timeout error appears immediately
4. Set timeout to `60` seconds — download completes normally on a slow/proxied network

## Screenshots

Screenshot_5

PR #40680, opened 2026-04-01
#40615 sql is quoted string insert export escaping

### LINKED ISSUE #40615: DBeaver Export Issue: Single Quotes Split Field Values in GaussDB Inserts
### Description

When using DBeaver to connect to a GaussDB database and export data as INSERT statements, if a field value contains single quotes (e.g., '03','04','05','10','11'), the value will be split into multiple field values during export.

***Expected result***: '''03'',''04'',''05'',''10'',''11'''

***Actual result***: '03','04','05','10','11'

As shown in the figure below.

Image

### DBeaver Version

26.0.1.202603221810

### Operating System

Windows 10

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

dev-miro26:
## Summary

When exporting data as INSERT statements, string values containing single quotes were not always escaped correctly. For example, a value `'03','04','05','10','11'` was emitted as-is, splitting one value into multiple columns. A value like `'kkk''pp'` was also emitted unescaped, silently losing data after re-import.

## Problem

`SQLUtils.convertValueToSQLFormat()` called `isQuotedString()` on the display string before quoting it. If the raw value happened to start and end with `'`, the method assumed it was already a properly quoted SQL literal and returned it without escaping. But `getValueDisplayString()` always returns the raw value, never a pre-quoted SQL string, so this check was incorrect for the `STRING` data kind.

**Expected output:** `'''03'',''04'',''05'',''10'',''11'''`
**Actual output:** `'03','04','05','10','11'`

## Solution

For `STRING` (and text `CONTENT` falling through to it), always call `getQuotedString()` unconditionally. The `ROWID` case is left unchanged since its values may behave differently. No changes to `isQuotedString()` itself — it is still valid for its other callers (SQL parser, `getUnquotedString`).

## Changes

- `SQLUtils.convertValueToSQLFormat` — `STRING` case now returns `getQuotedString(strValue)` directly, without the `isQuotedString` guard
- `AbstractSQLDialectTest` — new test class verifying `getQuotedString` handles plain values, embedded quotes, multi-fragment values, and values that look like pre-quoted SQL literals

## Testing

- [ ] Export table data containing single-quote characters as SQL INSERT and verify the output is a valid single escaped literal
- [ ] Verify values like `'kkk''pp'` round-trip correctly (export as INSERT, re-import, compare)
- [ ] Regression: ordinary string values still generate valid SQL

Closes #40615


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
```diff
@@ -538,7 +538,19 @@ public String getUnquotedIdentifier(String identifier, boolean unescapeQuotesIns
...
+        }
+        // Verify internal single quotes are properly escaped (doubled)
+        for (int i = 1, len = string.length() - 1; i < len; i++) {
+            if (string.charAt(i) == '\'') {
+                if (i + 1 >= len || string.charAt(i + 1) != '\'') {
+                    return false;
+                }
+                i++; // skip the paired quote
+            }
+        }
```
E1izabeth: Thanks for the contribution!

It doesn't work right for the case when the field contains such value `'kkk''pp'`
Generated query would be `INSERT INTO "Elizabeth".employee (id) VALUES('kkk''pp');`
And after executing it, we get `kkk'pp` in the database with the proposed solution

E1izabeth (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/AbstractSQLDialect.java
dev-miro26: You're right, the `isQuotedString `check can never work here because the input is always a raw value, not SQL. Fixed by removing the guard entirely and always calling `getQuotedString()`.
please check again.

E1izabeth: I suppose changes in `isQuotedString` are not actual then.
Also I'm unsure we should remove this check for rowid data kind, let's keep the case with rowid untouched.

dev-miro26: done!

test/org.jkiss.dbeaver.test.platform/src/org/jkiss/dbeaver/model/sql/AbstractSQLDialectTest.java
```diff
@@ -0,0 +1,53 @@
...
+            dialect.getQuotedString("'03','04','05','10','11'")
+        );
+    }
+
+    @Test
+    public void getQuotedStringValueLooksLikeQuotedLiteralTest() {
+        Assert.assertEquals(
+            "'''kkk''''pp'''",
+            dialect.getQuotedString("'kkk''pp'")
+        );
```
E1izabeth, line 51: Please add the following scenarios to test cases:
kkk'''''pp
'kkk'''''pp'
kkk''pp

E1izabeth:
Also I would ask to update the PR description according to the solution that is present now

dev-miro26:
> Also I would ask to update the PR description according to the solution that is present now

All done.

dev-miro26:
@E1izabeth 
I am sorry I have one question. I have find out my commit message must start with `dbeaver/dbeaver#40615` or `#40615`.
In this case, Do I have to use `git reset` and `git force push`?
Also I have opened a PR(#40552). For this case approved. but did not passed QA test.
Same operation need to fix the commit message?
Please let me know kindly. I appreciate you.

E1izabeth:
Yes, you're right. It seems those who approved your another PR missed the failing workflow step.

dev-miro26:
Could you allow the CI test again?

E1izabeth (approved):

ShadelessFox (approved):

ShadelessFox:
Thanks! Passing to QA

dev-miro26:
@serge-rider 
I know you are so busy. sorry bother you.
If you don't mind, Could you please review this PR?

Matvey16:
Verified

PR #40705, opened 2026-04-06
dbeaver/pro#8949 Fix splash screen on MacOS in dev env

IvanGorshechnikov:
Closes https://github.com/dbeaver/pro/issues/8949

### DIFF
--- a/plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/Constants.java
+++ b/plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/Constants.java
@@ -32,7 +32,7 @@ public class Constants {
     public static final String INTERNAL_OS_ZOS = "z/OS"; //$NON-NLS-1$
 
     public static final String ARCH_X86 = "x86";//$NON-NLS-1$
-    public static final String ARCH_X86_64 = "x86_64";//$NON-NLS-1$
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/Constants.java
```diff
@@ -32,7 +32,7 @@ public class Constants {
     public static final String INTERNAL_OS_ZOS = "z/OS"; //$NON-NLS-1$
 
     public static final String ARCH_X86 = "x86";//$NON-NLS-1$
-    public static final String ARCH_X86_64 = "x86_64";//$NON-NLS-1$
+    public static final String ARCH_X86_64 = "aarch64";//$NON-NLS-1$
```
ShadelessFox: No, x86_64 is not aarch64.

Just remove this check completely, it is seemingly redundant:
https://github.com/dbeaver/dbeaver/blob/cfc5d5a908c9f93d7ac89e5401122a06b3e39bea/plugins/org.jkiss.dbeaver.launcher/src/org/jkiss/dbeaver/launcher/DBeaverLauncher.java#L462-L464

ShadelessFox (changes requested):

ShadelessFox (approved):

E1izabeth (approved):

PR #40708, opened 2026-04-06
#40527  Fix false "Invalid column reference" error for UNSIGNED type modifier in CREATE TABLE

### LINKED ISSUE #40527: UNSIGNED in DDL show Invalid column reference in sql script editor
### Description

In the SQL script editor UNSIGNED is underlined with the error "Invalid column reference" in the DDL query:
CREATE TABLE test_table (
	test_table_id INT UNSIGNED NOT NULL,
	test_name VARCHAR(10) NOT NULL
);
when this is in fact perfectly valid in mysql

### DBeaver Version

Community edition 26.0.0.202603011816

### Operating System

Windows 11

### Database and driver

Database: MySQL 8.4.8
Driver: mysql-connector/J 8.4.0

### Steps to reproduce

_No response_

### Additional context

_No response_

dev-miro26:
## Summary

- Fixes false "Invalid column reference" error for MySQL type modifiers like `UNSIGNED` and `ZEROFILL` in DDL statements (e.g. `CREATE TABLE test (id INT UNSIGNED NOT NULL)`)
- Introduces a dedicated `typeModifier` grammar rule in `columnDefinition` to accept dialect-specific type modifiers between the data type and constraints
- Makes `DEFAULT` keyword mandatory in `defaultClause` to resolve grammar ambiguity between type modifiers and default value expressions
- Registers `typeModifier` in `LSMInspections` for proper autocompletion and syntax classification

## Problem

In the SQL script editor, `UNSIGNED` is underlined with the error **"Invalid column reference"** in DDL queries:

```sql
CREATE TABLE test_table (
    test_table_id INT UNSIGNED NOT NULL,
    test_name VARCHAR(10) NOT NULL
);
```
---
## How to test

1. Connect to a MySQL 8.x database
2. Open a SQL script editor
3. Type: `CREATE TABLE test_table (test_table_id INT UNSIGNED NOT NULL, test_name VARCHAR(10) NOT NULL);`
4. Verify `UNSIGNED` is **not** underlined with "Invalid column reference"
5. Verify `CREATE TABLE test_table (col1 INT DEFAULT 5 NOT NULL);` still works correctly (default value recognized)
6. Verify `ALTER TABLE test_table ADD COLUMN col2 BIGINT UNSIGNED NOT NULL;` does not show the error


dev-miro26:
@E1izabeth 
Could you please review this PR?

E1izabeth (changes requested):
You've correctly found the root cause of the issue. However, the fix is masking the problem instead of fixing the root cause. So, we can not accept the proposed solution.

dev-miro26:
> You've correctly found the root cause of the issue. However, the fix is masking the problem instead of fixing the root cause. So, we can not accept the proposed solution.

Let me  solve again.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/model/ddl/SQLQueryColumnSpec.java
```diff
@@ -121,6 +121,11 @@ public static SQLQueryColumnSpec recognize(SQLQueryModelRecognizer recognizer, S
             .orElse(null);
         String typeName = Optional.ofNullable(node.findFirstChildOfName(STMKnownRuleNames.dataType))
             .map(STMTreeNode::getTextContent).orElse(null);
+        for (STMTreeNode child : node.findChildrenOfName(STMKnownRuleNames.typeModifier)) {
+            if (typeName != null) {
+                typeName = typeName + " " + child.getTextContent();
+            }
+        }
```
E1izabeth, line 128: I don't like this change.
- Are we sure that introducing any amount of actual-identifiers as `typeModifier` won't effectively consume _any_ not-yet-covered keywords, turning it into nonsense?
- Having the above, are we sure that type name construction by concatenation with a modifier won't break constraint expression analysis in terms of types?
- Consider such a line:
	`columnName customTypeName someDbSpecificModifiers CHECK (columnName.compositeField < 10)`
Having `someDbSpecificModifiers` treated as part of the `customTypeName` will break the composite type lookup and thus break `compositeField` completion and validation.

The solution I see is to add type modifiers to the dialect and dynamically bring those constants into the grammar. You can take a look at how it's done with `knownIdentifierQuotes`.

E1izabeth (changes requested):

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/semantics/model/ddl/SQLQueryColumnSpec.java
dev-miro26: I will update again.
Thanks

PR #40723, opened 2026-04-08
dbeaver/pro#8234 udbt only for admins

HocKu7:
Closes https://github.com/dbeaver/pro/issues/8234


yagudin10 (dismissed):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/network/NetworkHandlerRegistry.java
```diff
@@ -66,15 +69,22 @@ private NetworkHandlerRegistry(IExtensionRegistry registry) {
...
     }
 
+    @Nullable
     public NetworkHandlerDescriptor getDescriptor(@NotNull String id) {
+        NetworkHandlerDescriptor descriptor = getRawDescriptor(id);
+        return descriptor != null && isAvailable(descriptor) ? descriptor : null;
+    }
+
+    @Nullable
+    public NetworkHandlerDescriptor getRawDescriptor(@NotNull String id) {
```
serge-rider, line 83: What are the cases when we need it?
At least add description

serge-rider (changes requested):

PR #40725, opened 2026-04-08
dbeaver/pro#8688 Fix extraction settings page layout

ShadelessFox:
I also replaced the extraction mode combo with a checkbox:

image


uslss:
verified (ue,, win11)
image

PR #40737, opened 2026-04-09
#40715 Fix issue where row count icon doesn’t reset after counting

### LINKED ISSUE #40715: Calculate row count icon doesn't change
### Description

Calculate total row count button doesn;t change back from a red cross when the caclulation is finished

### DBeaver Version

26.0.2

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

1. Open any table
2. Click `Calculate total row count`

**Expected:** Icon changes to a red cross and than back to the original
**Actual:** Icon of the cross stays

### Additional context

https://github.com/user-attachments/assets/c685379d-01d1-4cfd-b37b-14ca82ee0f05

HellAmbro:
This PR closes https://github.com/dbeaver/dbeaver/issues/40715

I think that the regression was introduced in commit [a6ad9dc](https://github.com/dbeaver/dbeaver/commit/a6ad9dc7f99a7eb5e746a94f494522e95727f657#diff-plugins_org-jkiss-dbeaver-ui-editors-data_src_org-jkiss-dbeaver-ui_controls-resultset-ActiveStatusMessage-java). Previously, `actionItem.setImage(actionImage);` was always executed, but the commit moved it inside a conditional block. As a result, if the row count does not change, the if block is not executed and the icon is no longer updated.


E1izabeth (approved):

ShadelessFox (approved):
Thanks for the contribution

E1izabeth:
Please add the issue number to the beginning of the commit message.

Matvey16:
Verified

PR #40740, opened 2026-04-09
#40372 Skip redundant save dialog when executing DROP queries

### LINKED ISSUE #40372: "Save result set edit" dialog shows even when dropping that very table
### Description

Shows the "Save result set edit" dialog when dropping that very table. First it shows the "Execute drop queries" dialog, warning of possible data loss. Choosing OK leads to a second dialog, "Save result set edit", asking if we want changes saved to the database. Considering the table is being dropped, this question is redundant.

### DBeaver Version

Community Edition 25.35

### Operating System

Windows 10

### Database and driver

PostgreSQL 16.8 on x86_64-pc-linux-gnu, compiled by x86_64-pc-linux-gnu-gcc (GCC) 10.5.0, 64-bit
PostgreSQL JDBC Driver 42.7.2

### Steps to reproduce

1) CREATE TABLE A(A INT PRIMARY KEY);
2) INSERT INTO A SELECT 1;
3) Refresh the UI so the result set can be edited.
4) SELECT * FROM A;
5) Modify the result set.
6) DROP TABLE A;

### Additional context

_No response_

Crystora:
## Summary

When executing a DROP TABLE query while the result set has unsaved edits, DBeaver shows a redundant "Save result set edit" dialog after the drop confirmation.
Saving edits to a table that is about to be dropped makes no sense.

## Changes

Discard pending result set edits before refreshing the viewer when the query contains a DROP statement.
This skips the unnecessary save prompt while keeping the normal behavior for all other queries.

## Test plan

- [X] Execute SELECT, edit a cell, then run DROP TABLE - only drop confirmation dialog should appear
- [X] Execute SELECT, edit a cell, then run another SELECT - save dialog should still appear
- [X] Execute SELECT without edits, then run DROP TABLE - no extra dialogs

Fixes #40372

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/QueryProcessor.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/QueryProcessor.java
@@ -260,6 +260,10 @@ abstract class QueryProcessor implements SQLResultsConsumer, ISmartTransactionMa
                     curJob = job;
                     ResultSetViewer rsv = resultsContainer.getResultSetController();
                     if (rsv != null) {
+                        // Discard pending edits when the query will destroy the underlying data
+                        if (rsv.isDirty() && containsDropQuery(queries)) {
+                            rsv.rejectChanges();
+                        }
                         rsv.resetDataFilter(false);
                         rsv.resetHistory();
                         rsv.refresh();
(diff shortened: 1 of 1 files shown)


bittoby:
@SolovevAnton @ShadelessFox Could you please review this PR? 
It’s a simple bug fix PR. I tested it and confirmed it works well. I'd appreciate your feedback.

ShadelessFox (changes requested):
I can imagine a scenario when this won't work as expected.
- If a table has triggers, this fix would prevent changes from being propagated to the table before it's dropped.
- Another one is when executing two queries: one that inserts data into another table from the dirty table before dropping the dirty table.

There are other cases that I haven't covered, that's for sure.

bittoby:
That's a fair point about triggers and side effects. What approach would you prefer here?
1. Reorder the dialogs so the save prompt appears before the drop confirmation
2. Keep the current approach since these edge cases are uncommon
3. Something else

ShadelessFox:
I think everything should be kept as is, without proposed changes. It's such a rare scenario when you accidentally edit the result set of a table that you're about to drop anyway that it's not worth breaking other use cases.

bittoby:
OK. I will convert this PR as draft. If you are going to update this, then I will reopen this pr and update

E1izabeth:
This PR is stale, so I'm closing it.

PR #40748, opened 2026-04-11
#40753 Fix hex string data truncation in BinaryFormatterHexString.toBytes

### LINKED ISSUE #40753: BinaryFormatterHexString.toBytes silently truncates hex data
The toBytes method in BinaryFormatterHexString has two problems on the same two lines that cause silent data truncation when parsing hex strings.

The condition uses || instead of &&, so the substring logic runs even when only one side (prefix or postfix) is present. More critically, the end index of the substring call subtracts both HEX_PREFIX.length() and HEX_POSTFIX.length() from the string length. Since the prefix is at the front, there is no reason to subtract it from the end. For a value like x'1234ABCD' (length 11), the current code produces substring(2, 8) returning "1234AB" and silently drops the last two hex characters. The correct end index is string.length() - HEX_POSTFIX.length().

tejas-ae:
The toBytes method in BinaryFormatterHexString has two issues on the same two lines that cause silent data truncation when parsing hex strings.

The condition uses || (OR) instead of && (AND), so if only the prefix or only the postfix is present the substring logic still runs even though the assumption of both being there isn't met.

More critically, the end index of the substring call subtracts both HEX_PREFIX.length() and HEX_POSTFIX.length() from the string length. The prefix is at the front, so there's no reason to subtract it from the end. For a string like x'1234ABCD' (length 11) the current code produces substring(2, 8) which gives "1234AB", silently dropping the last two hex characters. The correct end index is string.length() - HEX_POSTFIX.length(), which gives substring(2, 10) and the full "1234ABCD".

You know the codebase far better than I do, so happy to adjust if I've missed any context here.


E1izabeth:

tejasae-afk:
Done, created #40753 and updated the commit message.

E1izabeth (approved):

ShadelessFox (approved):

ShadelessFox:
Thanks :^)

E1izabeth:
Hi, @tejas-ae! Please provide clear steps to reproduce the behavior that you have fixed.

ShadelessFox:
@tejas-ae we can't test and merge this ticket without clear instructions on how to reproduce it. Please provide them asap

E1izabeth:
It looks like this PR is stale, so I'm closing it. No steps to reproduce were provided

PR #40755, opened 2026-04-13
Fix: BinaryFormatterHexString.toBytes silently truncates hex data #40753

### LINKED ISSUE #40753: BinaryFormatterHexString.toBytes silently truncates hex data

enjoyandlove:
**Title**
Fix silent truncation when parsing x'…' hex strings in BinaryFormatterHexString
(Optional shorter title: Fix hex string toBytes prefix/postfix handling)

**Summary**
This updates BinaryFormatterHexString#toBytes so values in the form x'…' are parsed without dropping trailing hex digits, and so prefix/postfix stripping only runs when both wrappers are present.

**Problem** 
The guard used logical OR (||), so the unwrap logic could run when only the prefix or only the postfix matched, which is wrong for the x'…' format.
The end index for substring subtracted both HEX_PREFIX.length() and HEX_POSTFIX.length() from the string length. The prefix is only at the start; subtracting its length from the end incorrectly shortens the slice. For example, x'1234ABCD' (length 11) became substring(2, 8) → "1234AB", silently losing CD.

**Solution**
Require both prefix and postfix with && before stripping.
Set the exclusive end index to string.length() - HEX_POSTFIX.length() so the body is exactly the hex between x' and the closing '. 

**Example**
Input		
x'1234ABCD'

Before (buggy)
Parses as "1234AB"

After (fixed)
Parses as "1234ABCD"****


E1izabeth:
The issue was created specifically for PR #40748. So, we will review and accept the mentioned one if everything is fine

enjoyandlove:
> The issue was created specifically for PR #40748. So, we will review and if everything is fine to accept the mentioned one

Thanks for the clarification. I’ll wait for your review and feedback.

enjoyandlove:
Hi @E1izabeth 
I’ve addressed all of the requested changes in this PR. When you have a moment, could you please take another look and review it?
Thank you for your time.

enjoyandlove:
Hi, @serge-rider 
I already fixed  all of the requested changes in this PR.
When you have a chance, could you please review it ?
Thank you for your time

E1izabeth:
https://github.com/dbeaver/dbeaver/pull/40755#issuecomment-4243050467
As I said, another PR was created earlier. Most probably, we will accept that one.

ShadelessFox:
We decided to go with https://github.com/dbeaver/dbeaver/pull/40748.

PR #40762, opened 2026-04-14
dbeaver/pro#8132 Implement keywords completion when keyword is not completely written

E1izabeth:
image


IvanGorshechnikov (approved):

ShadelessFox (approved):

Matvey16:
Verified

DB CE Windows

Issue #38293, opened 2025-06-03
Wrong time formating in PostgreSQL


premyslc:
### Description

DBeaver displays the wrong time zone for the time data type, probably due to date formatting.

### DBeaver Version

25.1.0

### Operating System

Linux, Windows 11

### Database and driver

PostgreSQL SQL DB 17rc1
Driver:  JDBC 42.7.2

### Steps to reproduce

1. Execute Query: select current_time, current_timestamp;
2. Time zone is different for time a and time stamp (11:56:08 +0100	2025-06-03 12:56:08.150 +0200)
3. The workaround is to turn off date formatting. (13:04:32.662866+02	2025-06-03 13:04:32.662866+02)

E1izabeth:
Thanks for the bug report. We reproduced the issue

E1izabeth:
![Image](https://github.com/user-attachments/assets/55d41a09-0877-4d61-88ac-ebc469de9b0f)

labanovichttps:
https://github.com/pgjdbc/pgjdbc/issues/3692

PR #40763, opened 2026-04-14
#40757 Fix PostgreSQL TIME 24:00:00 display

### LINKED ISSUE #40757: postgres time column with value 2400 incorrectly shows as 0000
### Description

DBeaver accepts inputs 0000 and 2400 to a time column, when inserting a new record. But when displaying it shows both as 0000

### DBeaver Version

Community Edition Version 26.0.2.202604051735

### Operating System

Arch Linux

### Database and driver

1. Database PostgreSQL 18.3
2. Driver org.postgresql.Driver

### Steps to reproduce

1. Create a table with a field time(0)
2. Insert a row with value 0000
3. Insert a row with value 2400
4. query: > '2300'
5. query: < '0100'

Both queries work, but dbeaver shows both values in the response as 0000

### Additional context

_No response_

EastLord:
Closes: #40757 

test, Use the following SQL:

```sql
drop table if exists test_time_2400;
create table test_time_2400 (
    id int primary key,
    t time(0)
);

insert into test_time_2400(id, t) values (1, '0000');
insert into test_time_2400(id, t) values (2, '2400');

select id, t from test_time_2400;
```

Test environment：

image

before fix:
image

After fix:

image

pgAdmin:

image

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreDateTimeValueHandler.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreDateTimeValueHandler.java
@@ -85,6 +85,15 @@ public class PostgreDateTimeValueHandler extends JDBCDateTimeValueHandler {
                 try {
                     return jdbc.getObject(index + 1, OffsetTime.class);
                 } catch (SQLException e) {
+                    try {
+                        String rawString = jdbc.getString(index + 1);
+                        // PostgreSQL accepts 24:00:00 for TIME, but java.sql.Time can't display it distinctly from 00:00:00.
+                        if (PostgreConstants.TIME_END_OF_DAY.equals(rawString)) {
+                            return PostgreConstants.TIME_END_OF_DAY;
+                        }
+                    } catch (SQLException e1) {
+                        log.debug("Exception caught when reading PostgreSQL time value as string", e1);
+                    }
                     log.debug("Exception caught when fetching time value", e);
                 }
             }
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreConstants.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreConstants.java
@@ -120,6 +120,7 @@ public class PostgreConstants {
     public static final String TYPE_TIMESTAMP = "timestamp";
     public static final String TYPE_TIMETZ = "timetz";
     public static final String TYPE_TIMESTAMPTZ = "timestamptz";
+    public static final String TIME_END_OF_DAY = "24:00:00";
     public static final String TYPE_XML = "xml";
     public static final String TYPE_BOOLEAN = "boolean";
     public static final String TYPE_BYTEA = "bytea";


EastLord:
@E1izabeth Could you please review this PR? I'm not sure whether this change is acceptable.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreDateTimeValueHandler.java
```diff
@@ -83,9 +82,13 @@ public Object fetchValueObject(
         if (resultSet instanceof JDBCResultSet jdbc) {
             if (type.getTypeID() == Types.TIME && !formatSettings.isUseNativeDateTimeFormat()) {
                 try {
-                    return jdbc.getObject(index + 1, OffsetTime.class);
+                    String rawString = jdbc.getString(index + 1);
+                    // PostgreSQL accepts 24:00:00 for TIME, but java.sql.Time can't display it distinctly from 00:00:00.
+                    if (PostgreConstants.TIME_END_OF_DAY.equals(rawString)) {
+                        return PostgreConstants.TIME_END_OF_DAY;
+                    }
```
E1izabeth, line 89: I'm unsure about this change. We used to return Object of OffsetTime type here.Now we return String and ONLY if it's 24:00:00

EastLord: In practice, for PostgreSQL `TIME`, `jdbc.getObject(index + 1, OffsetTime.class)` was not working here. In my testing it always threw and fell back to the generic `java.sql.Time` path.

That fallback is exactly what causes the bug, because PostgreSQL allows `24:00:00`, but `java.sql.Time` cannot preserve it distinctly from `00:00:00`.

So this change only preserves the PostgreSQL-specific value `24:00:00` as text. Normal `TIME` values still follow the existing path.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreDateTimeValueHandler.java
```diff
@@ -83,9 +82,13 @@ public Object fetchValueObject(
         if (resultSet instanceof JDBCResultSet jdbc) {
             if (type.getTypeID() == Types.TIME && !formatSettings.isUseNativeDateTimeFormat()) {
                 try {
-                    return jdbc.getObject(index + 1, OffsetTime.class);
+                    String rawString = jdbc.getString(index + 1);
```
serge-rider: Let's keep old behavior by default. OffsetTime is better than string as it may keep timezone offset which is valuable and we don't want to loose it.

We can surround it with try/catch and try to handle this particular issue with weird 24:00. Return PostgreConstants.TIME_END_OF_DAY in this case.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreDateTimeValueHandler.java
EastLord: Thanks, I see the point now.

I checked pgjdbc metadata and noticed that `timetz` is also reported as `Types.TIME`, so this branch is used not only for regular `time`, but also for values like `current_time`. In that case `OffsetTime` can preserve the offset, so we should keep that path.

I will update the patch to keep the `OffsetTime` read first, and only handle the PostgreSQL `24:00:00` case in the fallback when `OffsetTime` cannot represent regular `time`.

https://github.com/dbeaver/dbeaver/issues/38293

EastLord:
I've updated  accordingly.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/data/PostgreDateTimeValueHandler.java
```diff
@@ -85,6 +85,15 @@ public Object fetchValueObject(
...
                     return jdbc.getObject(index + 1, OffsetTime.class);
                 } catch (SQLException e) {
+                    try {
+                        String rawString = jdbc.getString(index + 1);
+                        // PostgreSQL accepts 24:00:00 for TIME, but java.sql.Time can't display it distinctly from 00:00:00.
+                        if (PostgreConstants.TIME_END_OF_DAY.equals(rawString)) {
+                            return PostgreConstants.TIME_END_OF_DAY;
+                        }
+                    } catch (SQLException e1) {
+                        log.debug("Exception caught when reading PostgreSQL time value as string", e1);
```
ShadelessFox, line 95: Maybe we should log the original exception instead?

EastLord: Thanks for the suggestion.

The original exception is already logged after the fallback block:

`log.debug("Exception caught when fetching time value", e);`

ShadelessFox (approved):

EastLord:
Thanks everyone for the review and feedback.

Since #41044 now addresses the same PostgreSQL `TIME 24:00:00` issue as part of a broader datetime edge-case fix, I’ll close this PR to avoid duplicate work.

PR #40772, opened 2026-04-14
#39464 Fix force TOP injection for MSSQL table-valued function queries

### LINKED ISSUE #39464: [BUG] SQL query returns no results with default (200) row limit, but works with high limit (999999)
### Description

An SQL query using the function FUNC_PRODUTO with filters (e.g., `where T.CODIGO = 301089 order by T.CODIGO DESC`) returns correct results only when the row limit is set very high (e.g., 999999). When using the default row limit of 200, the query returns no results. The same issue occurs when selecting all rows with `order by desc`.

problem=
Image

success=
Image

### DBeaver Version

Community Edition 25.2.3.202510191611

### Operating System

MacOS Sierra

### Database and driver

Microsoft Sql Server

### Steps to reproduce

_No response_

### Additional context

_No response_

xingxing21:
...reference #39464 

### Problem
When querying a multi-statement Table-Valued Function (MSTVF) in SQL Server with the default row limit (200), DBeaver returns 0 results even though the same query works correctly with a high limit (e.g. 999999).

Root cause is a two-step failure:

1. By default (RESULT_SET_MAX_ROWS_USE_SQL = false), DBeaver does not inject TOP N into the SQL text. Instead it falls back to stmt.setMaxRows(200) via JDBC.
2. The SQL Server JDBC driver translates setMaxRows() into SET ROWCOUNT 200 — a session-level setting that limits rows for all  DML operations in the session, including INSERT statements executed inside the TVF body.

So the TVF's internal INSERT INTO @result_table SELECT ... stops after 200 rows from whatever order the source data happens to be in — before the outer WHERE and ORDER BY can be applied. If the target record falls outside those 200 rows, zero results are returned.

TOP N in the SQL text does not have this problem: it is applied only to the final output of the outer query, after the TVF completes execution in full.

### Change
SQLServerDataSource.isForceTransform()

Extended to return true when the query's FROM clause contains a TableFunction (jsqlparser term for a TVF reference). This forces QueryTransformerTop to inject TOP N directly into the SQL instead of falling back to setMaxRows(), so the TVF always executes without ROWCOUNT interference.


ShadelessFox:
Please update your commit message per our [contribution guide](https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code)

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDataSource.java
```diff
@@ -639,6 +640,15 @@ public void collectObjectStatistics(@NotNull DBRProgressMonitor monitor, boolean
     @Override
     public boolean isForceTransform(DBCSession session, SQLQuery sqlQuery) {
         try {
+            Statement statement = SQLSemanticProcessor.parseQuery(this.sqlDialect, sqlQuery.getText());
+            if (statement instanceof PlainSelect plainSelect
+                && plainSelect.getFromItem() instanceof TableFunction) {
```
ShadelessFox, line 649: Because of this, the method's javadoc is no longer accurate. Please update it accordingly

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDataSource.java
```diff
@@ -639,6 +640,15 @@ public void collectObjectStatistics(@NotNull DBRProgressMonitor monitor, boolean
     @Override
     public boolean isForceTransform(DBCSession session, SQLQuery sqlQuery) {
         try {
+            Statement statement = SQLSemanticProcessor.parseQuery(this.sqlDialect, sqlQuery.getText());
```
ShadelessFox, line 647: This invocation may throw an exception if the supplied query is incorrect or contains constructs that aren't recognized by JSQLParser, which we use under the hood, so the code after your additions won't execute as a fallback.

xingxing21:
@ShadelessFox 
I just fixed your two suggestions.
Please check it.
Thanks 😉

ShadelessFox:
Please clean up your commits. Drop everything redundant and rebase onto `devel`.

ShadelessFox (approved):

E1izabeth (approved):

E1izabeth:
Thanks for the contribution! We passed your PR to our QA team for testing.

Matvey16:
Verified

PR #40784, opened 2026-04-16
#40769 Fix transaction monitor showing N/A after toolbar drag-and-drop

### LINKED ISSUE #40769: Transaction Monitor shows "N/A" after dragging/dropping the Transaction Monitor toolbar when Auto-Commit is active
### Description

When the Transaction Monitor sub-toolbar is repositioned via drag-and-drop in the workbench, the monitor widget resets to "N/A" instead of reflecting the actual connection state (e.g., "Auto" when Auto-Commit is enabled).

Root cause: When the toolbar is dragged/dropped Eclipse calls ToolbarContribution.createControl() again (TransactionMonitorToolbar.java:271), which creates a new MonitorPanel instance. The new panel starts with txnState = null. The only automatic refresh triggers are:

A workbench part activation event (line 77) — but no part changes during a toolbar drag.
A QM transaction event (commit, rollback, statement execution, context open/close) — but none fire if the connection is in Auto-Commit with no pending activity.
Because no refresh is ever scheduled after recreation, the panel remains stuck displaying "N/A" (line 198) until the user switches to a different editor tab or executes a statement.

The issue is most visible with Auto-Commit enabled because that mode generates no transaction events, but the same stale "N/A" state occurs in any transaction mode after a toolbar move if no immediate activity follows.

### DBeaver Version

Version 26.0.3.202604141503

### Operating System

Windows 11

### Database and driver

_No response_

### Steps to reproduce

1. Open any SQL editor with an active database connection.
2. In the Database menu → Transaction Mode, set the mode to Auto-Commit. Verify the Transaction Monitor toolbar shows "Auto".
3. Right-click the toolbar area and unlock toolbars (if needed), then drag the Transaction Monitor sub-toolbar to a different position in the workbench.
4. Observe the Transaction Monitor widget.

### Additional context

https://github.com/user-attachments/assets/f9a7db97-b1a3-4d69-b822-99d9715dbb41

xingxing21:
… toolbar drag
... reference #40769 

## Problem

When the Transaction Monitor toolbar is repositioned via drag-and-drop,
Eclipse calls `ToolbarContribution.createControl()` again, which creates
a new `MonitorPanel` with `txnState = null`. The panel only refreshes on
workbench part activation or QM transaction events — neither of which
fires during a toolbar drag. As a result, the monitor stays stuck on
"N/A" until the user switches editor tabs or executes a statement.

This is most visible in Auto-Commit mode since no transaction events are
generated, but it affects all transaction modes.

## Fix

Trigger an immediate `refresh()` on the newly created `MonitorPanel`
inside `createControl()`, so the current transaction state is fetched and
displayed right after the toolbar is (re)created.

## Steps to reproduce

1. Open any SQL editor with an active database connection
2. Set transaction mode to Auto-Commit — verify monitor shows "Auto"
3. Drag the Transaction Monitor toolbar to a different position
4. Observe monitor resets to "N/A"

After the fix, the monitor correctly shows "Auto" (or the current state)
immediately after the toolbar is repositioned.

Assisted by: Claude Code


E1izabeth (approved):

ShadelessFox (approved):

ShadelessFox:
Thanks. Passing to QA

MashaKorax:
verified DBeaver ce win

PR #40816, opened 2026-04-20
dbeaver/pro#8245 Add ability to save sql to file from Generate SQL dialog

E1izabeth:
image

image


plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/dialogs/BaseSQLDialog.java
```diff
@@ -155,16 +165,47 @@ protected void createRefreshButton(Composite parent) {
...
+                DBWorkbench.getPlatformUI().showError(
+                    SQLEditorMessages.dialog_view_sql_save_to_file_failure_title,
+                    NLS.bind(SQLEditorMessages.dialog_view_sql_save_to_file_failure_message, file),
+                    e
+                );
+                return false;
+            }
+        } else {
+            return false;
+        }
```
ShadelessFox:
```suggestion
            }
        }
        return false;
```

IvanGorshechnikov (approved):

ShadelessFox (approved):

PR #40830, opened 2026-04-20
dbeaver/pro#8882 Fix truncate

IvanGorshechnikov:
Closes https://github.com/dbeaver/pro/issues/8882


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/impl/ChatTruncator.java
```diff
@@ -45,15 +47,41 @@ private ChatTruncator(Builder b) {
...
     }
 
+    /**
+     * Result of a truncation operation.
+     * {@link #wasTruncated()} returns {@code true} if any message was dropped or its content was cut.
+     */
+    public static final class TruncationResult {
+        private final List<AIMessage> messages;
+        private final boolean wasTruncated;
+
```
E1izabeth, line 57: It looks like a record

E1izabeth (dismissed):

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/impl/ChatTruncator.java
```diff
@@ -45,15 +47,25 @@ private ChatTruncator(Builder b) {
...
     /**
      * Truncate the conversation to fit into (maxDbSnapshotTokens - reserves).
      * Ordering in the result is chronological (SYSTEM first, then oldest->newest).
+     *
+     * @return a {@link TruncationResult} that carries the resulting message list
+     *         and a flag indicating whether anything was actually truncated.
      */
     @NotNull
-    public List<AIMessage> truncate(@NotNull List<AIMessage> input) {
+    public TruncationResult truncate(@NotNull List<AIMessage> input) {
```
ShadelessFox: Perhaps it should instead return `null` when nothing was truncated and rename it to `tryTruncate`? That auxiliary record seems redundant.

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/impl/ChatTruncator.java
```diff
@@ -74,47 +86,115 @@ public List<AIMessage> truncate(@NotNull List<AIMessage> input) {
...
+                }
+                AIMessage m = rest.get(idx);
+                int t = counter.count(m.getContent());
+                if (used + t <= budget) {
+                    pinned.put(idx, m);
+                    used += t;
+                } else {
+                    int remaining = Math.max(0, budget - used);
+                    AIMessage cut = truncateToTokens(m, remaining);
+                    cut = cut.withContent(cut.getContent() + ChatTruncator.DEFAULT_TRUNCATED_SUFFIX);
```
ShadelessFox: Redundant qualifier here and in two other places

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/impl/ChatTruncator.java
ShadelessFox: Would also be nice to have tests

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/impl/ChatTruncator.java
```diff
@@ -46,14 +46,17 @@ private ChatTruncator(Builder b) {
...
+     * Attempts to truncate the conversation so it fits within the configured token budget.
+     *
+     * @param input the full conversation history (may be unordered with respect to SYSTEM messages)
+     * @return the truncated message list if anything was cut or dropped,
+     *         or {@code null} if the entire input already fits within the budget unchanged
      */
-    @NotNull
-    public List<AIMessage> truncate(@NotNull List<AIMessage> input) {
+    @Nullable
+    public List<AIMessage> tryTruncate(@NotNull List<AIMessage> input) {
```
ShadelessFox, line 56: Why did it change so much? I don't understand changes

IvanGorshechnikov: These changes are needed to save the last user and assistant message; without them, AI go to inifinte cycle or loses any idea of what the user wants 
We have to do it because now AI will send many messages to the user, and we can lose some important messages, like what exactly the user wants

test/org.jkiss.dbeaver.model.ai.test/src/org/jkiss/dbeaver/model/ai/impl/ChatTruncatorTest.java
```diff
@@ -0,0 +1,193 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import static org.junit.Assert.*;
+
+public class ChatTruncatorTest {
```
ShadelessFox, line 30:
```suggestion
public class ChatTruncatorTest extends DBeaverProUnitTest {
```

test/org.jkiss.dbeaver.model.ai.test/src/org/jkiss/dbeaver/model/ai/impl/ChatTruncatorTest.java
ShadelessFox, line 1: Please fix nullability issues

plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/impl/ChatTruncator.java
```diff
@@ -74,47 +77,88 @@ public List<AIMessage> truncate(@NotNull List<AIMessage> input) {
...
         }
 
         result.addAll(pickedReverse);
-        return result;
+        return truncated ? result : null;
     }
 
-    private static AIMessage mergeSystems(List<AIMessage> systems) {
+    @NotNull
+    private static int[] getLastUserAndAssistantMessage(@NotNull List<AIMessage> rest) {
```
ShadelessFox, line 142: This should be a list. Negative values shouldn't be included in the result, instead of discarding them later

PR #40872, opened 2026-04-22
dbeaver/pro#8473 api for ai functions

yagudin10:


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/AIFunctionCall.java
```diff
@@ -79,6 +79,14 @@ public void setFunctionName(@NotNull String functionName) {
         this.functionName = functionName;
     }
 
+    @NotNull
+    public String getLabel() {
```
serge-rider: getFunctionDisplayName

serge-rider (changes requested):

PR #40873, opened 2026-04-22
dbeaver#15127 Colorize arrows in ERD by on-delete action

### LINKED ISSUE #15127: Configure arrow line in ER Diagram to represent foreign key
Thank you for the awesome project first!!

We use ER diagram to assist in database design and one of common problems is to introduce multiple cascade path suddenly when adding a new table/field. We need to identify possible cycles and determine which foreign key to set to "on delete set null". This is particularly difficult given current ERD because the lines between table does not indicate referential relationship clearly.

In the generated ER Diagram, provide some options to configure the link between tables to be like this:

Show referential relationship. If one field in A is a foreign key referring to B, show an arrow line between the two tables.

If the deletion relationship is cascade, use solid line, anything else (e.g., on delete do nothing, on delete set null), use dotted line.

Currently I had to draw the referential relationship of the whole ERD on paper to manually identify multiple cascade paths. No better way.

**Additional context**
 Something like this:

image

with all arrow line like this cycle or multiple cascade path will be really easy to identify.

SimonSimCity:
This is just a very rough first attempt of solving #15127 and rather meant to get feedback.

Since this is my first contribution to this project, let me outline a bit:

I started off searching for a proper way to set the color. As a first attempt, and for a proof of concept, I went for only implementing it for crowsfoot-style.

When it came to adjusting the color, I found a class called `colorRegistry` which is related to the theme, but couldn't quite find out where the colors were defined - I'd need a hint here to get to a good place to store this.

What I want is some early feedback on this POC:

* UI/UX: Is colorizing the arrows a good way to go? Are the colors themselves fine? I just prompted Perplexity about some well-distinguishable colors that would fit the delete-action.
* Code: Where would you put code like this? Where's a good place to start implementing?
* Code: How would you define colors? Have a predefined palette, or let the user choose in the settings?

Outlook:
When starting with `OnDelete`, we could even push further for an `OnUpdate` as well - whereby you could argue that `OnDelete`, specially the destructive actions, are something you want visually see in the design.


ShadelessFox:
Hi, thanks for your contribution!

> Is colorizing the arrows a good way to go? Are the colors themselves fine?

I don't know :) Please provide a screenshot in both light/dark themes, so we have something to judge.

> Where would you put code like this? Where's a good place to start implementing?

As long as you reference colors by their IDs via Java constants, it should be fine to have this logic duplicated across notations.

> When it came to adjusting the color, I found a class called colorRegistry which is related to the theme, but couldn't quite find out where the colors were defined - I'd need a hint here to get to a good place to store this.

See `org.jkiss.dbeaver.ui.editors.erd.ERDColors`. Colors are defined by extending the `org.eclipse.ui.themes` extension point. For example:

https://github.com/dbeaver/dbeaver/blob/03e0a7a75ba446f36e47503011274878579a3ebe/plugins/org.jkiss.dbeaver.ui.editors.erd/plugin.xml#L252-L264

Dark theme colors are overridden like so:

https://github.com/dbeaver/dbeaver/blob/035eb400cbee824d3b6a938fa69a2de21574576a/plugins/org.jkiss.dbeaver.ui.editors.erd/css/e4-dark-erd-editor.css#L1-L4

Next, we have `DBSForeignKeyModifyRule`. Looking at it, it weirdly feels like it should be an enum instead. So instead of switching on its (hardcoded) ID, you could instead switch on the rule itself.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/rdb/DBSForeignKeyModifyRule.java
```diff
@@ -25,25 +25,27 @@
  */
 public class DBSForeignKeyModifyRule implements DBPNamedObject
 {
-    public static final DBSForeignKeyModifyRule UNKNOWN = new DBSForeignKeyModifyRule("UNKNOWN", "?", null); //$NON-NLS-1$ //$NON-NLS-2$
-    public static final DBSForeignKeyModifyRule NO_ACTION = new DBSForeignKeyModifyRule("NO_ACTION", ModelMessages.model_struct_No_Action, null); //$NON-NLS-1$
-    public static final DBSForeignKeyModifyRule CASCADE = new DBSForeignKeyModifyRule("CASCADE", ModelMessages.model_struct_Cascade, "CASCADE"); //$NON-NLS-1$ //$NON-NLS-3$
-    public static final DBSForeignKeyModifyRule SET_NULL = new DBSForeignKeyModifyRule("SET_NULL", ModelMessages.model_struct_Set_NULL, "SET NULL"); //$NON-NLS-1$ //$NON-NLS-3$
-    public static final DBSForeignKeyModifyRule SET_DEFAULT = new DBSForeignKeyModifyRule("SET_DEFAULT", ModelMessages.model_struct_Set_Default, "SET DEFAULT"); //$NON-NLS-1$ //$NON-NLS-3$
-    public static final DBSForeignKeyModifyRule RESTRICT = new DBSForeignKeyModifyRule("RESTRICT", ModelMessages.model_struct_Restrict, "RESTRICT"); //$NON-NLS-1$ //$NON-NLS-3$
+    public enum ID { UNKNOWN, NO_ACTION, CASCADE, SET_NULL, SET_DEFAULT, RESTRICT, };
```
ShadelessFox: I'm fine with having *both* `name()` and `getName()`, as both have different origins. `DBSForeignKeyModifyRule` has a fixed subset of possible values it could have, so it *makes* sense to make it an enum instead, not have its `id` an enum, as it would open a _possible_ opportunity to introduce a new instance of `DBSForeignKeyModifyRule`, which shouldn't happen.

SimonSimCity:
<details>

<summary>I created a sample SQL structure, containing all possible relations in PostgreSQL.</summary>

```sql
-- Parent: Companies
CREATE TABLE companies (
    id SERIAL PRIMARY KEY,
    name VARCHAR(100) NOT NULL
);

-- Parent: Departments
CREATE TABLE departments (
    id SERIAL PRIMARY KEY,
    name VARCHAR(100) NOT NULL,
    company_id INTEGER REFERENCES companies(id) ON DELETE CASCADE
);

-- CASCADE: Deletes child rows when parent is deleted
CREATE TABLE employees (
    id SERIAL PRIMARY KEY,
    name VARCHAR(100) NOT NULL,
    dept_id INTEGER REFERENCES departments(id) ON DELETE CASCADE
);

-- RESTRICT: Prevents parent deletion if child exists
CREATE TABLE managers (
    id SERIAL PRIMARY KEY,
    name VARCHAR(100) NOT NULL,
    dept_id INTEGER REFERENCES departments(id) ON DELETE RESTRICT
);

-- SET NULL: Sets FK to NULL when parent deleted (column allows NULL)
CREATE TABLE contractors (
    id SERIAL PRIMARY KEY,
    name VARCHAR(100) NOT NULL,
    dept_id INTEGER REFERENCES departments(id) ON DELETE SET NULL
);

-- SET DEFAULT: Sets FK to default when parent deleted
CREATE TABLE interns (
    id SERIAL PRIMARY KEY,
    name VARCHAR(100) NOT NULL,
    dept_id INTEGER DEFAULT 1 REFERENCES departments(id) ON DELETE SET DEFAULT
);

-- Parent: Projects
CREATE TABLE projects (
    id SERIAL PRIMARY KEY,
    name VARCHAR(100) NOT NULL,
    company_id INTEGER REFERENCES companies(id) ON DELETE CASCADE
);

-- NO ACTION: Like RESTRICT but deferrable (default behavior)
CREATE TABLE tasks (
    id SERIAL PRIMARY KEY,
    description TEXT NOT NULL,
    project_id INTEGER REFERENCES projects(id) ON DELETE NO ACTION
);
```
</details>

## Visual comparison

Given a database containing tables created by the SQL statements above, the diagram of the schema would look as following:

| Theme  | Before | After |
| ------------- | ------------- | ------------- |
| Light  |  |  |
| Dark  |   |  |
| Classic  |   |  |

Just one thing I want to note as to consider:

Both in dark and light mode, an arrow is colorized as green and the thickness is increased, likewise are the two columns colorized in the related tables. This clashes with the idea of colorizing by delete-action, and should be reconsidered:



One possible solution would be to change the highlight-color (both for the arrow and the related columns) to the color we use for the arrow.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/rdb/DBSForeignKeyModifyRule.java
SimonSimCity: Thanks for your feedback. As in adherence [to the rules](https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code#commit-message-rules), I didn't force-push an edited commit (as I would usually), but rather added a new one at the end on what I was working on.

See f8bca10.

SimonSimCity:
## Code style

* I've moved the colors to the color-registry and added some translations to where I saw necessary. Please let me know if there's anything else I should have considered (e.g. in where the XML is used for example).
* I could move the colorization of the arrows to the `AssociationPart.java` file, and this would centralize it, but would split the concern of styling the arrows between this generalized class and the specific notations. I chose to leave it in the specific annotation files, as it creates are more comprehensible picture of the full styles applied by just looking at `IDEF1XDiagramNotation` and `CrowsFootDiagramNotation`.

**EDIT:**

I just found this section in the settings:



With this in mind I would rather remove my changes from `e4-dark-erd-editor.css` so the user can fully customize the colors without an override on the dark theme. Would be interesting to hear your opinion about it, as I see that there are also some other overrides of colors in that CSS file.

plugins/org.jkiss.dbeaver.ui.editors.erd/src/org/jkiss/dbeaver/ui/editors/erd/notation/crowsfoot/CrowsFootDiagramNotation.java
```diff
@@ -96,7 +100,27 @@ public void applyNotationForArrows(
...
+
+        // Adjust the color of the arrow based on the delete-rule of the foreign key.
+        if (association.getObject() instanceof DBSTableForeignKey) {
+            DBSForeignKeyModifyRule deleteRule = ((DBSTableForeignKey) association.getObject()).getDeleteRule();
+            conn.setForegroundColor(ERDColors.getForeignKeyModifyRuleColor(deleteRule));
+        }
+    }
+
+    /*
+    private String getDeleteRuleColorKey(DBSForeignKeyModifyRule rule) {
```
IvanGorshechnikov: Please remove commented code block

IvanGorshechnikov (changes requested):

ShadelessFox:
> One possible solution would be to change the highlight-color (both for the arrow and the related columns) to the color we use for the arrow.

That's what I thought when looking at the gif :) please try and share the results

plugins/org.jkiss.dbeaver.ui.editors.erd/src/org/jkiss/dbeaver/ui/editors/erd/notation/idef1x/IDEF1XDiagramNotation.java
```diff
@@ -74,6 +77,12 @@ public void applyNotationForArrows(
             }
             conn.setLineDash(constraintType.isLogical() ? new float[] { 4 } : new float[] { 5 });
         }
+
+        // Adjust the color of the arrow based on the delete-rule of the foreign key.
+        if (association.getObject() instanceof DBSTableForeignKey) {
+            DBSForeignKeyModifyRule deleteRule = ((DBSTableForeignKey) association.getObject()).getDeleteRule();
```
ShadelessFox:
```suggestion
        if (association.getObject() instanceof DBSTableForeignKey foreignKey) {
            DBSForeignKeyModifyRule deleteRule = foreignKey.getDeleteRule();
```

plugins/org.jkiss.dbeaver.ui.editors.erd/src/org/jkiss/dbeaver/ui/editors/erd/ERDColors.java
```diff
@@ -55,4 +56,16 @@ public static Color getHeaderColor(int index) {
         return UIUtils.getColorRegistry().get(EXTRA_HEADER_COLORS[index]);
     }
 
+    public static Color getForeignKeyModifyRuleColor(DBSForeignKeyModifyRule rule) {
```
ShadelessFox:
```suggestion
    @NotNull
    public static Color getForeignKeyModifyRuleColor(@NotNull DBSForeignKeyModifyRule rule) {
```

SimonSimCity:
> > One possible solution would be to change the highlight-color (both for the arrow and the related columns) to the color we use for the arrow.
> 
> That's what I thought when looking at the gif :) please try and share the results

@ShadelessFox, I think I need your help on this:

I searched for a color that could be suitable and ended up at the `ERDHighlightingManager` , used in `AssociationPart.setSelected()` and `AttributePart.setSelected()`. Judging by `ERDEditPartFactory`, I could get the name of the models and could confirm by their class-comments, that `ERDAssociation` are about relations (knew that kind-of already by the previous changes), and `ERDEntityAttribute` is about columns.

That's also what I could see in the UI: You can select relations and columns. Selecting tables doesn't apply any highlighting.

Now to the implementation:
- For `ERDAssociation`, it shouldn't be all to hard. We can get the color of the foreign-key relation and colorize all by that. The arrow is also bold, which indicates to the user which relation was selected.
- For `ERDEntityAttribute`, we have to choose which key to colorize in which color. What if an entity-attribute has multiple relations with different on-delete actions? You can even have two foreign key definitions on a single column - even though it might not be a sane choice for the majority of cases, it is possible, and I think the implementation should consider it.

Other than the color: When selecting `ERDAssociation`, they are bold. But when selecting `ERDEntityAttribute`, only the color changes, which is what we want to keep, so we'll have to set other indications of what has been selected and what is highlighted.

There's a need for a better UX concept than what I can as of now come up with. And either way, it might require some substantial changes on the `ERDHighlightingManager` class, which currently assumes all highlighted elements share the same color.

That said, here is how the highlighting on `ERDAssociation` looks. Keep in mind that the name of the column might get barely readable as a result of the chosen highlight-color (to worsen this problem: we have light and dark mode, where the text is the opposite to the mode).






**Note:** I tried to override the color, and it overrides the color also for dark-mode. Means, if the user defines his own color, it will be used for both light- and dark-mode, while the program can have different values for both light and dark. You can see both as a preset, that the user can override in one setting. So, no need to remove our color-definition from `e4-dark-erd-editor.css`. It actually gives us a flexibility the user doesn't have 😊

plugins/org.jkiss.dbeaver.ui.editors.erd/src/org/jkiss/dbeaver/ui/editors/erd/notation/crowsfoot/CrowsFootDiagramNotation.java
SimonSimCity: Done. See 09bebe8

plugins/org.jkiss.dbeaver.ui.editors.erd/src/org/jkiss/dbeaver/ui/editors/erd/notation/idef1x/IDEF1XDiagramNotation.java
SimonSimCity: Done. See 09bebe8

plugins/org.jkiss.dbeaver.ui.editors.erd/src/org/jkiss/dbeaver/ui/editors/erd/ERDColors.java
SimonSimCity: Done. See 09bebe8

ShadelessFox:
You can try blending the color between the background color and highlight color in a 50/50 ratio. Should yield acceptable results. See `org.jkiss.dbeaver.ui.UIStyles#mix`

SimonSimCity:
@ShadelessFox, thanks, this resolves the problem of the bad readability 🎉 So, this resolves the `ERDAssociation` part.

We're still left with colorizing on `ERDEntityAttribute`:



Either we leave it as is, choose a different mechanism than colorizing for highlighting selection, or we will have to make very fundamental changes to `ERDHighlightingManager`.

Leaving it as is to me sounds acceptable, how about you?

ShadelessFox:
I'm fine with leaving it as-is :^)

Destrolaric (approved):

SimonSimCity:
@ShadelessFox, just because you were the one assisting me on the changes - is there something I can do to help this change being merged? 🤗

ShadelessFox:
Hi, @SimonSimCity 

Your PR currently awaits QA. Our QA engineers are quite busy right now due to a high volume of AI PRs requiring their attention. It will be reviewed and merged eventually; it just takes a very long time right now.

uslss:
verified

uslss:
@SimonSimCity, we've discussed this feature. Could you make it an option so it can be enabled or disabled?

E1izabeth:
@SimonSimCity would you add a setting for your solution?
https://github.com/dbeaver/dbeaver/pull/40873#issuecomment-4934293814

SimonSimCity:
@E1izabeth, @uslss, thanks for following up - my time is quite limited these days, but I'll see when I can finish this 🎉

PR #40875, opened 2026-04-22
#40874 Fix clearing driver proxy credentials

### LINKED ISSUE #40874: Driver proxy username and password can't be cleared
### Description

I've noticed that the driver proxy username and password are not being cleared properly/saving the state if the inputs are cleared/emptied

### DBeaver Version

26.0.3

### Operating System

Windows

### Database and driver

_No response_

### Steps to reproduce

1. Go to: Window>Preferences>Drivers
2. Setup a url, username, password. Apply and close.
3. Then go back and remove the username and password, apply and close.
4. Open again to verify the old username and password are removed, but they are populated with the old populated values.

### Additional context

_No response_

EastLord:
Closes: #40874


EastLord:
@E1izabeth Could you please review this PR? Thanks !

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/runtime/net/GlobalProxyAuthenticator.java
```diff
@@ -161,14 +162,19 @@ public static DBPAuthInfo readCredentials() throws DBException {
...
-            if (CommonUtils.isNotEmpty(password)) {
-                secrets.setPrivateSecretValue(ModelPreferences.UI_PROXY_PASSWORD, password);
-            }
-            secrets.flushChanges();
-        }
+        DBPPreferenceStore store = DBWorkbench.getPlatform().getPreferenceStore();
+        store.setToDefault(ModelPreferences.UI_PROXY_USER);
+        store.setToDefault(ModelPreferences.UI_PROXY_PASSWORD);
+        PrefUtils.savePreferenceStore(store);
+
```
E1izabeth, line 169: I don't understand why do you set the proxy credentials to default in the `saveCredentials` method. Could you explain the approach?

EastLord: The reason is the backward-compatibility fallback in `readCredentials()`.

`readCredentials()` first reads the secret store, but if the username is absent there, it falls back to `DBPPreferenceStore` (`UI_PROXY_USER` / `UI_PROXY_PASSWORD`).

Without clearing those legacy preference values, removing the proxy credentials from the current UI would only clear the secret-store values, and the old credentials could still be read back through the legacy fallback path on the next open.

So the intention here was to keep the effective stored state consistent: when the current proxy credentials are saved, the obsolete legacy credential entries should no longer participate in credential resolution.

That said, I agree the placement inside `saveCredentials()` is not very obvious. If you prefer, I can extract this into a separate helper (e.g. `clearLegacyCredentials()`) or move the cleanup closer to the compatibility/migration path to make the intent clearer.

E1izabeth: I see. Thanks for the explanation

E1izabeth (approved):

ShadelessFox (approved):

RadAnton:
verified on Win 11

PR #40876, opened 2026-04-22
#40833 Fix IDEA connection import for non-standard datasource pages

### LINKED ISSUE #40833: Improve IDEA connection import for non-standard datasource pages (e.g Snowflake, Bigquery)
I noticed problems with importing connections that use non-standard connection pages  (e.g Snowflake, Bigquery). 

Currently, if a datasource has a custom or non-standard connection configuration page (e.g., Snowflake, BigQuery), the import does not work correctly. After import, the connection configuration may be empty or partially missing.

Observed issues include:

- Connection configuration is empty after import
- Port is not transferred and falls back to the default value
- Other connection parameters may also be lost

https://github.com/user-attachments/assets/5f568455-e73b-4e09-8d74-0edfc3afdf56

a3894281:
### Summary

- What: Fix JetBrains/IDEA connection import for non-standard datasource pages (Snowflake, etc.) so host, port, database, warehouse, schema, role, and authenticator are populated after import (fixes #40833).
- Why: Imported connections lost most of their configuration; fields fell back to driver defaults or showed empty in the connection settings dialog.

### Root Cause

- `plugins/org.jkiss.dbeaver.ext.snowflake/plugin.xml` sample URL used `[:port]` (no braces), so the regex had no `port` named group and `matcher.group("port")` threw `IllegalArgumentException` on successful host matches.
- `ConfigImportWizard.parseUrlAsDriverSampleUrl` used `matches()`, which requires the entire URL to match. Trailing query parameters like `&warehouse=...&schema=...` caused full parse failure, leaving host/port/database unset.
- URL query parameters were never captured into provider properties, so non-standard fields (warehouse, schema, role, authenticator) were lost even when the URL carried them.
- `SnowflakeConnectionPage` and `SnowflakeAuthSnowflakeConfigurator` only read these fields from their canonical slots (`serverName`, `PROP_SCHEMA`, `authProperty`), not from provider properties populated by the importer.

### Change

- `plugin.xml`: `[:port]` → `[:{port}]` so `port` is a named capture group.
- `ConfigImportWizard`:
  - Retry the sample-URL match with the query string stripped when the full-URL match fails.
  - Add a generic URI-based fallback parser for non-standard shapes (e.g. `jdbc:scheme://host:port?params` with no `/` before `?`).
  - `safeGroup()` helper tolerates sample URLs that do not declare `host`/`port`/`database` groups.
  - Always extract URL query parameters into provider properties regardless of whether the sample-URL match succeeded.
- `SnowflakeConnectionPage.loadSettings`: fallback to `providerProperty("warehouse")` / `providerProperty("schema")` when the canonical slots are empty.
- `SnowflakeAuthSnowflakeConfigurator.loadSettings`: same idea for role and authenticator.
- Commit: `#40833 Fix IDEA connection import for non-standard datasource pages`
- Branch: `fix/issue_40833`

### Testing

Manual: imported four Snowflake URL shapes via `Database > Import > DataGrip > Paste configuration`, then opened each connection's settings dialog.

- [x] Standard `jdbc:snowflake://host:port/?db=...&warehouse=...&schema=...&role=...&authenticator=...` → all fields populated.
- [x] No-slash `jdbc:snowflake://host:port?params` (matches the GitHub issue screenshot form) → host/port extracted via URI fallback; warehouse/schema/role populated.
- [x] No port `jdbc:snowflake://host/?params` → host + driver default port; warehouse/schema/role populated.
- [x] Minimal `jdbc:snowflake://host:port/` → host/port only, defaults elsewhere (no regression for simple URLs).

### Notes

- The parser changes in `ConfigImportWizard` benefit every import wizard that inherits from it (DataGrip/JetBrains, DBVis, Navicat, pgAdmin, Squirrel, SQL Workbench), not only Snowflake.

image


E1izabeth (dismissed):

ShadelessFox (dismissed):
Thanks :^) passing to QA

a3894281:
@ShadelessFox 
@E1izabeth 

I rebased on `devel` to fix out-of-date status. Only conflict was a Javadoc comment in ConfigImportWizard.java — kept upstream wording. No logic changes. Could you re-approve so QA can proceed?

Thanks.

E1izabeth (approved):

ShadelessFox (approved):

PR #40879, opened 2026-04-22
#40902 XML data export: use column label consistently and guard empty element names

### LINKED ISSUE #40902: XML data export: DOCTYPE uses column labels while row tags use raw names, and export crashes on empty column names
### Description

`DataExporterXML` in
`plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterXML.java`
has two user-visible bugs in its XML serialisation logic.

--- Bug 1: Label/name inconsistency across DOCTYPE and row tags ---

`printHeader` picks `columns[i].getLabel()` (with fallback to
`getName()`) for the DATA_RECORD content list on lines 78-86, but the
per-column `!ELEMENT` declarations on lines 89-91 and the actual
element tags emitted in `exportRow` on lines 103-104 use only
`column.getName()`.

For an aliased SELECT like

    SELECT price AS "Unit Price" FROM products

the exported XML declares

    <!ELEMENT DATA_RECORD (Unit_Price?)>
    <!ELEMENT Unit_Price (#PCDATA)>

but every row emits `<price>100</price>` — the DOCTYPE references
`<Unit_Price>` while the rows emit `<price>`. The document fails DTD
validation and does not round-trip through a standards-compliant XML
parser.

The JSON exporter already uses label-first (after merged PR #2731,
`DataExporterJSON.java:112`). The XML exporter got partial adoption
of that pattern — only in the content list — leaving the three-way
mismatch between the content list, the ELEMENT declarations, and the
row tags.

--- Bug 2: `escapeXmlElementName` crashes on empty column names ---

The function calls `escapedName.charAt(0)` on line 174 without first
checking whether the post-regex string is empty. If a column's label
AND name are both empty strings (which JDBC drivers can legitimately
return for computed columns — e.g. `SELECT 1+1 FROM dual` in some
driver implementations),

    name.replaceAll("[^\\p{Alpha}\\p{Digit}]+", "_")

returns `""` and `charAt(0)` throws `StringIndexOutOfBoundsException`,
aborting the export mid-row. The partially-written XML file is then
truncated and unusable.

Fix submitted in #40879: extracts `getXmlColumnName(DBDAttributeBinding)`
which returns `escapeXmlElementName(label.isEmpty() ? name : label)`
and wires all three call sites through it; guards
`escapeXmlElementName` to return `"_"` on empty input so the export
never crashes mid-row.

### DBeaver Version

Community Edition 26.0.4

### Operating System

Windows 11 Pro 24H2

### Database and driver

1. Database name and version: PostgreSQL 16.2 (any JDBC-backed DB reproduces — the bug is in the exporter layer, not in any driver)
2. Driver name and version: PostgreSQL JDBC 42.7.3

### Steps to reproduce

Bug 1 — DOCTYPE/row tag mismatch
1. Connect to any database with a table (e.g. `products(price numeric)`).
2. In the SQL editor run: `SELECT price AS "Unit Price" FROM products LIMIT 3`
3. Right-click results → Export data → XML (Comma-Separated Values).
4. On the exporter settings page, leave "Include DOCTYPE declaration" ON (the default).
5. Open the generated `.xml` in a strict XML/DTD validator (e.g. `xmllint --valid file.xml` or https://www.w3.org/2001/03/webdata/xsv).
6. Expected: document validates — DOCTYPE element names and row tag names match.
   Actual: DOCTYPE declares `<!ELEMENT DATA_RECORD (Unit_Price?)>` but rows emit `<price>…</price>`. Validator reports "element price not declared".

Bug 2 — empty-name crash
(Harder to reproduce without a specific driver — any JDBC driver that
returns empty strings from `ResultSetMetaData.getColumnLabel()` AND
`getColumnName()` for a computed column triggers
StringIndexOutOfBoundsException at
DataExporterXML.escapeXmlElementName line 174. Verifiable by stepping
through the exporter with a synthetic DBDAttributeBinding whose both
label and name are `""`.)

### Additional context

Precedent for the label-first idiom: merged PR #2731
(DataExporterJSON — "use column label instead of column name"). The
fix in #40879 just extends that same pattern to the XML exporter's
remaining two call sites.

boskodev790:
The XML exporter's printHeader used column labels for the DATA_RECORD content list but then switched to raw column names for the per-column ELEMENT declarations and for every element tag written in exportRow. For an aliased SELECT such as `SELECT price AS "Unit Price"` this produced XML whose DOCTYPE referenced `<Unit_Price>` while the rows emitted `<price>` — the resulting document failed DTD validation and did not round-trip through standards-compliant XML parsers.

Extract `getXmlColumnName(DBDAttributeBinding)` which returns `escapeXmlElementName(label.isEmpty() ? name : label)` and wire all three call sites through it. Matches the label-first idiom the JSON exporter already uses after #2731.

Also guard `escapeXmlElementName` against empty / null input: the previous body called `escapedName.charAt(0)` directly and threw `StringIndexOutOfBoundsException` if the column's label and name were both empty (which can happen for unnamed computed columns from JDBC drivers that return empty strings from `getColumnLabel` / `getColumnName`). Return `"_"` in that case so the export never crashes mid-row.

## Reproduction of bug #1

1. Run a query with aliased columns: `SELECT product_id AS "Product ID", price AS "Unit Price" FROM products`
2. Export results → XML, toggle "Include DOCTYPE declaration" on
3. The generated `<!DOCTYPE>` block lists `Product_ID?, Unit_Price?` but every `<DATA_RECORD>` emits `<product_id>…</product_id>, <price>…</price>` instead — the document fails DTD validation.

## Reproduction of bug #2

A driver returning an empty-string label and empty-string name for a computed column (unusual but legitimate per JDBC spec) crashes the export with `StringIndexOutOfBoundsException` at `escapeXmlElementName:174`.


E1izabeth:
Thanks for the contribution!

E1izabeth (changes requested):
Please edit your commit message to add the issue number to the beginning.
It should start with #40902

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterXML.java
```diff
@@ -178,4 +177,28 @@ private String escapeXmlElementName(String name) {
...
+     * Before this helper the three call sites disagreed: the DATA_RECORD
+     * content list used the label, but the per-column ELEMENT declarations
+     * and the row tags used the raw column name, so an aliased SELECT
+     * (e.g. `SELECT price AS "Unit Price"`) produced XML whose DOCTYPE
+     * referenced `<Unit_Price>` while the rows emitted `<price>` — the
+     * resulting document failed DTD validation.
+     *
+     * Matches the label-first idiom the JSON exporter already uses (merged
+     * in #2731).
+     */
```
SolovevAnton: remove comment please

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterXML.java
```diff
@@ -178,4 +177,28 @@ private String escapeXmlElementName(String name) {
...
+    private String getXmlColumnName(DBDAttributeBinding column) {
```
SolovevAnton: Please add nullability annotations

SolovevAnton (changes requested):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterXML.java
SolovevAnton: Or make it shorter and more readable, please

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterXML.java
```diff
@@ -178,4 +177,12 @@ private String escapeXmlElementName(String name) {
 
         return "_" + escapedName;
     }
+
+    private String getXmlColumnName(DBDAttributeBinding column) {
```
IvanGorshechnikov: Add nullability annotation here

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterXML.java
```diff
@@ -170,6 +166,9 @@ else if (buffer[i] == '>') {
     }
 
     private String escapeXmlElementName(String name) {
```
IvanGorshechnikov: And here please

IvanGorshechnikov (changes requested):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterXML.java
boskodev790: Done in cfd5fc7. `@Nullable String name` on the parameter — the `CommonUtils.isEmpty(name)` branch already handles a null caller — and `@NotNull` on the return since the method always produces a string.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterXML.java
boskodev790: Done in cfd5fc7. `@NotNull DBDAttributeBinding column` (the body dereferences `column.getLabel()` / `column.getName()` without a null guard, so non-null is the real contract) and `@NotNull` on the return since it just forwards to `escapeXmlElementName`.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterXML.java
boskodev790: Added in cfd5fc7. Same shape as the JSON exporter helpers — `@NotNull` on the return and `@NotNull DBDAttributeBinding column` on the parameter, and `@NotNull` / `@Nullable` on the sibling `escapeXmlElementName`.

E1izabeth (approved):

ShadelessFox (approved):

ShadelessFox:
Thanks, passing to QA

E1izabeth:
Unfortunately, we can't reproduce the reported issue with the provided steps.
Can you please provide footage showing your issue? Thank you.

E1izabeth:
This PR is stale, so I'm closing it.
We were unable to reproduce the reported issue, so we can't accept it without any additional information

PR #40880, opened 2026-04-22
#40899 HTML data export: emit matched <tr> tags when column-header and table-header flags differ

### LINKED ISSUE #40899: HTML data export emits unmatched <tr> tags when tableHeader and columnHeaders flags differ
### Description

The HTML data exporter produces malformed HTML when the "Table header"
and "Column headers" options in the export settings are toggled
independently. `DataExporterHTML.printHeader` opens a second `<tr>`
inside the `if (outputHeader)` branch and closes it inside the
`if (outputColumnHeaders)` branch, so the two tags only pair up when
both flags are true. Two of the four combinations produce broken HTML:

- Table header ON, Column headers OFF
    -> `<tr>title</tr><tr>`  (opening `<tr>` with no closing)
- Table header OFF, Column headers ON
    -> `<td>col1</td>…</tr>` (closing `</tr>` with no opening)

Browsers silently rewrap the stray tag, but strict consumers (XHTML
parsers, HTML linters like the W3C Nu validator, email clients) reject
the output.

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterHTML.java
Method: printHeader

Fix submitted in #40880.

### DBeaver Version

Community Edition 26.0.4

### Operating System

Windows 11

### Database and driver

1. Database name and version: SQLite 3.46.0 (any driver reproduces — the bug is in the HTML exporter, not the database layer)
2. Driver name and version: SQLite JDBC 3.46.0.0

### Steps to reproduce

1. Open any result set in DBeaver
2. Export → HTML
3. On the export settings page, enable "Table header" but uncheck "Column headers" (or vice versa)
4. Open the output .html in a strict validator (e.g. https://validator.w3.org/nu/)

### Additional context

Fix PR: #40880 (balances the <tr> tags by moving the column-header <tr>
into the same `if` block that closes it).

boskodev790:
```
`printHeader` in `DataExporterHTML` opened a second `<tr>` inside the
`if (outputHeader)` branch and closed it inside the
`if (outputColumnHeaders)` branch. The two tags only paired up when both
flags were true. The export panel exposes the flags independently via
the `tableHeader` and `columnHeaders` properties, so two of the four
combinations produced malformed HTML:

- `tableHeader=true, columnHeaders=false`:
      `<tr>title</tr><tr>`   ← opening `<tr>` with no closing
- `tableHeader=false, columnHeaders=true`:
      `<td>col1</td>…</tr>`  ← closing `</tr>` with no opening

Both emit invalid HTML that browsers silently rewrap (moving the stray
`<tr>` out of the `<table>` DOM) and that strict consumers (XHTML
parsers, automated HTML linters, email clients) reject or mis-render.

## Fix

Move the `<tr>` that opens the column-header row into the branch that
closes it. Each `<tr>` now opens and closes inside the same `if` block,
so the output is well-formed for every combination of the two flags.

## Reproduction

1. Export a result set as HTML.
2. In the export dialog, enable "Header" but disable "Column headers".
3. Inspect the saved `.html` file — the `<tr>` right after the title row
   is never closed: `<tr>…</tr><tr></table></body>`.

Or the reverse: disable "Header" and enable "Column headers" — the
column-header row produces an unmatched `</tr>` after the cells.
```


E1izabeth:
Thanks for the contribution!

ShadelessFox (approved):

E1izabeth:
Please edit your commit message to add the issue number to the beginning.
It should be
#40899 HTML data export: emit matched <tr> tags when column-header and table-header flags differ

E1izabeth (approved):

E1izabeth:

MashaKorax:
verified DBeaver ce win

PR #40883, opened 2026-04-23
#40900 SQL: guard delimiter bounds check so single-token scripts don't crash transformers

### LINKED ISSUE #40900: StringIndexOutOfBoundsException in SQLUtils when a script consists solely of the statement delimiter
### Description

`SQLUtils.needQueryDelimiter` and `SQLUtils.removeQueryDelimiter` both
compute `query.charAt(query.length() - delimiter.length() - 1)` to
verify the trailing delimiter is tokenised as its own SQL token (not
the tail of a larger identifier). When the entire query equals the
delimiter — e.g. a bare `GO` against SQL Server — the index is `-1`
and the call throws `StringIndexOutOfBoundsException: Index -1 out of
bounds for length N`.

Both methods are reached from real user flows:
- `removeQueryDelimiter` is called by `QueryTransformerLimit.addLimit`,
  so every Auto-Limit rewrite on a single-token script crashes.
- `removeQueryDelimiter` is also called by
  `GenericSourceViewEditor`, so viewing the source of a stored
  routine whose body is just a delimiter (common for empty/test
  procedures) crashes.

The script silently fails with a cryptic "Index -1 out of bounds"
entry in the error log instead of producing a result.

Files:
- plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
  (methods `needQueryDelimiter`, `removeQueryDelimiter`)

Arithmetic: query = "GO", delimiter = "GO"
  -> query.length() - delimiter.length() - 1 = 2 - 2 - 1 = -1
  -> charAt(-1) throws.

Fix submitted in #40883.

### DBeaver Version

Community Edition 26.0.4

### Operating System

Windows 11 Pro 24H2

### Database and driver

1. Database name and version: Microsoft SQL Server 2019 (reproduces on any MSSQL — the bug is in SQL pre-processing, driver-agnostic)
2. Driver name and version: Microsoft JDBC Driver 12.8

### Steps to reproduce

1. Connect to any Microsoft SQL Server instance.
2. In the SQL editor, type exactly one line containing only the statement delimiter `GO`.
3. Execute the script (Alt+X) with any script-level transformer enabled (Auto-Limit is on by default).
4. Observe in the Error Log view: `StringIndexOutOfBoundsException: Index -1 out of bounds for length 2` originating from `org.jkiss.dbeaver.model.sql.SQLUtils.needQueryDelimiter`.

### Additional context

Also reachable via "View Source" on a stored routine whose body is
just the delimiter (empty/test procs). Fix PR: #40883.

boskodev790:
```
`SQLUtils.needQueryDelimiter` and `SQLUtils.removeQueryDelimiter` both
computed `query.charAt(query.length() - delimiter.length() - 1)` to
check that the trailing delimiter is a standalone SQL token. When the
entire query equals the delimiter — e.g. a bare `GO` against SQL
Server — the computed index is `-1` and the call throws
`StringIndexOutOfBoundsException`.

`removeQueryDelimiter` is invoked from `QueryTransformerLimit.addLimit`
and `GenericSourceViewEditor`, so the crash reaches the Auto-Limit
rewrite codepath and the stored-routine "View Source" editor: the
script silently fails with a cryptic `Index -1 out of bounds` entry in
the log instead of producing a result.

## Fix

Extract `isDelimiterStandalone(query, delimiter)` which returns `true`
when there is no preceding character (delimiter IS the whole query —
no identifier could surround it), or when the preceding character is
not a letter or digit. Route both call sites through it so the bounds
check can never dereference a negative index.

## Reproduction (pre-fix)

1. Connect to any SQL Server instance.
2. Execute the single statement `GO` as a script.
3. Log shows `StringIndexOutOfBoundsException: Index -1 out of bounds
   for length 2` from `SQLUtils.needQueryDelimiter` (or from
   `QueryTransformerLimit` with Auto-Limit enabled).

After the fix, the transformer returns cleanly (`removeQueryDelimiter`
yields `""` — delimiter stripped — and `needQueryDelimiter` reports
`true` since the delimiter is present and standalone), and the outer
execution pipeline continues without throwing.
```

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
@@ -1242,10 +1242,10 @@ public final class SQLUtils {
         for (String delimiter : scriptDelimiters) {
             if (!delimiter.isEmpty()) {
                 if (Character.isLetterOrDigit(delimiter.charAt(0))) {
-                    if (query.toUpperCase().endsWith(delimiter.toUpperCase())) {
-                        if (!Character.isLetterOrDigit(query.charAt(query.length() - delimiter.length() - 1))) {
-                            return true;
-                        }
+                    if (query.toUpperCase().endsWith(delimiter.toUpperCase())
+                        && isDelimiterStandalone(query, delimiter))
+                    {
+                        return true;
                     }
                 } else {
                     return !query.endsWith(delimiter);
@@ -1261,10 +1261,10 @@ public final class SQLUtils {
             if (!delimiter.isEmpty() && query.contains(delimiter)) {
                 String queryWithoutDelimiter = query.substring(0, query.lastIndexOf(delimiter));
                 if (Character.isLetterOrDigit(delimiter.charAt(0))) {
-                    if (query.toUpperCase().endsWith(delimiter.toUpperCase())) {
-                        if (!Character.isLetterOrDigit(query.charAt(query.length() - delimiter.length() - 1))) {
-                            return queryWithoutDelimiter;
-                        }
+                    if (query.toUpperCase().endsWith(delimiter.toUpperCase())
+                        && isDelimiterStandalone(query, delimiter))
+                    {
+                        return queryWithoutDelimiter;
                     }
                 } else if (query.endsWith(delimiter)) {
                     return queryWithoutDelimiter;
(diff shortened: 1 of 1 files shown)


E1izabeth:
Thanks for the contribution!

E1izabeth (changes requested):
Please fix the commit message, and I'm ready to approve the changes. The commit message should be:
`#40900 SQL: guard delimiter bounds check so single-token scripts don't crash transformers`

E1izabeth (approved):
thanks

IvanGorshechnikov (approved):

E1izabeth:

E1izabeth:
@boskodev790 we can't reproduce the reported issue with the provided steps. So, we can't test it. Please add additional info about connection settings (are all defaults or not).

E1izabeth:

PR #40885, opened 2026-04-23
#40901 SQL: Fold case with explicit locale so case-insensitive lookups work on Turkish JVMs

### LINKED ISSUE #40901: Turkish JVM locale breaks AbstractObjectCache lookups and causes QueryTransformerLimit to append a second LIMIT
### Description

Two separate user-visible bugs in `org.jkiss.dbeaver.model` share a
common root cause: `String.toUpperCase()` / `toLowerCase()` is called
without an explicit `Locale`. On Turkish / Azeri / Lithuanian JVMs,
ASCII `i` folds to `İ` (U+0130, LATIN CAPITAL LETTER I WITH DOT ABOVE)
and ASCII `I` folds to `ı` (U+0131, dotless). Any codepath that
compares a default-locale-folded string against an ASCII literal — or
against a string folded in a different session / call site — breaks.

--- Bug 1: AbstractObjectCache misses every identifier containing i/I
      on Turkish JVMs ---

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/cache/AbstractObjectCache.java

`getObjectName` (line 307) builds the map key via `name.toUpperCase()`
and `getCachedObject` (line 93) looks it up via `name.toUpperCase()`.
When the stored name and the looked-up name differ in case — e.g. DB
stores `items` and the caller asks for `ITEMS` — the two default-
locale folds are NOT equal on a Turkish JVM:

    "items".toUpperCase(Locale.forLanguageTag("tr")) == "İTEMS"
    "ITEMS".toUpperCase(Locale.forLanguageTag("tr")) == "ITEMS"

The map lookup returns `null` and DBeaver re-fetches the object from
the database (or raises "object not found" upstream) for every
identifier containing i / I. `AbstractObjectCache` is inherited by
every metadata cache in the model (tables, columns, schemas, indexes,
procedures, …), so the regression touches virtually every navigator
action on a Turkish JVM.

--- Bug 2: QueryTransformerLimit appends a second LIMIT on Turkish JVMs ---

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/QueryTransformerLimit.java

`transformQueryString` tests whether the user's query already has a
LIMIT / OFFSET / INTO / UPDATE / PROCEDURE / FETCH clause via

    String testQuery = query.getText().toUpperCase().trim();
    plainSelect = !NON_LIMIT_QUERY_PATTERN.matcher(testQuery).find();

`NON_LIMIT_QUERY_PATTERN` contains ASCII literals. On a Turkish JVM,
`"select 1 limit 10".toUpperCase()` yields `"SELECT 1 LİMİT 10"`
(dotted I's) which does NOT match the ASCII `"LIMIT"` token. The
transformer concludes the query has no limit and appends a second
one, producing invalid SQL like

    select 1 limit 10
    LIMIT 200

which the driver rejects with a syntax error.

Fix submitted in #40885: fold with `Locale.ROOT` at all three call
sites (both AbstractObjectCache sites and the QueryTransformerLimit
site), plus a Mockito-backed test that pins the cache behaviour under
Turkish default locale.

### DBeaver Version

Community Edition 26.0.4

### Operating System

Windows 11 Pro 24H2

### Database and driver

1. Database name and version: PostgreSQL 16.2 (any DB reproduces — the bugs are in SQL pre-processing and model-layer caching, not in any driver)
2. Driver name and version: PostgreSQL JDBC 42.7.3

### Steps to reproduce

Both bugs require the JVM default locale to be Turkish (or Azeri /
Lithuanian). Easiest way: launch DBeaver with JVM args
`-Duser.language=tr -Duser.country=TR` (add to dbeaver.ini).

Bug 1 — AbstractObjectCache miss storm
1. Connect to any case-insensitive metadata provider (Postgres, Oracle, MSSQL).
2. Open a table whose name contains the letter i (e.g. `items`) in the Database Navigator.
3. Right-click the table several times and expand columns.
4. Observe in the SQL log: DBeaver re-issues the metadata query on every expansion instead of hitting the cache. Cache lookups for `getCachedObject("ITEMS")` miss because the stored key is `İTEMS`.

Bug 2 — QueryTransformerLimit double-LIMIT
1. SQL editor → Preferences → "Auto-Limit" enabled, default 200.
2. Execute: `select 1 limit 10`
3. Observe the rewritten SQL sent to the driver:
     select 1 limit 10
     LIMIT 200
   The driver rejects it with a syntax error (e.g. Postgres: `syntax error at or near "LIMIT"`).

### Additional context

Neither bug reproduces on an en_US / en_GB / any non-dotted-i JVM
locale, because `toUpperCase()` then behaves the same as
`toUpperCase(Locale.ROOT)`. Fix in #40885 uses `Locale.ROOT` (the
canonical Java idiom for ASCII-keyword folding per the `Locale`
Javadoc). New unit test `AbstractObjectCacheLocaleTest` saves /
restores `Locale.setDefault` around the suite so no leakage into
sibling tests.

boskodev790:
```
## Reproduction (pre-fix)

### Cache miss (Bug 1)

- Launch DBeaver on a JVM with `-Duser.language=tr -Duser.country=TR`.
- Connect to any case-insensitive metadata provider (Oracle, Postgres, etc).
- Open a table named `items` in the Database Navigator.
- Internal cache lookups via `getCachedObject("ITEMS")` miss on every
  column expansion or right-click action — DBeaver re-queries the DB
  each time (visible in the SQL log).

### Double-LIMIT rewrite (Bug 2)

- Same Turkish JVM.
- SQL editor → Preferences → "Auto-Limit" on, default limit 200.
- Run `select 1 limit 10`.
- DBeaver rewrites to `select 1 limit 10\nLIMIT 200\n` and the driver
  rejects the double LIMIT with a syntax error.

After this PR both scenarios succeed regardless of the JVM default
locale.
```

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/QueryTransformerLimit.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/QueryTransformerLimit.java
@@ -24,6 +24,7 @@ import org.jkiss.dbeaver.model.sql.SQLQuery;
 import org.jkiss.dbeaver.model.sql.SQLQueryType;
 import org.jkiss.dbeaver.model.sql.SQLUtils;
 
+import java.util.Locale;
 import java.util.regex.Pattern;
 
 /**
@@ -65,7 +66,12 @@ public class QueryTransformerLimit implements DBCQueryTransformer {
     @Override
     public String transformQueryString(SQLQuery query) throws DBCException {
         String newQuery;
-        String testQuery = query.getText().toUpperCase().trim();
+        // Use an explicit locale so the ASCII keyword list in NON_LIMIT_QUERY_PATTERN
+        // matches regardless of the JVM default locale. In Turkish/Azeri locales plain
+        // toUpperCase() folds a lowercase "limit" to "LİMİT" (with a combining dot above),
+        // which does not match the ASCII pattern literal "LIMIT" - Auto-Limit then
+        // appends a second LIMIT clause and the driver rejects the rewritten query.
+        String testQuery = query.getText().toUpperCase(Locale.ENGLISH).trim();
         SQLDialect dialect = SQLUtils.getDialectFromDataSource(query.getDataSource());
         boolean plainSelect = query.isPlainSelect();
         if (!plainSelect && query.getType() == SQLQueryType.UNKNOWN) {
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/cache/AbstractObjectCache.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/cache/AbstractObjectCache.java
@@ -90,7 +90,9 @@ public abstract class AbstractObjectCache<OWNER extends DBSObject, OBJECT extend
     @Override
     public OBJECT getCachedObject(@Nullable String name) {
         synchronized (cacheSync) {
-            return objectList == null || name == null ? null : getObjectMap().get(caseSensitive ? name : name.toUpperCase());
+            return objectList == null || name == null
+                ? null
+                : getObjectMap().get(caseSensitive ? name : name.toUpperCase(Locale.ENGLISH));
         }
     }
 
@@ -304,7 +306,12 @@ public abstract class AbstractObjectCache<OWNER extends DBSObject, OBJECT extend
             return null;
         }
         if (!caseSensitive) {
-            return name.toUpperCase();
+            // Use an explicit locale (matches renameObject above) so the case-folded
+            // map key is stable. With the JVM default locale, a Turkish/Azeri JVM
+            // folds "items" to "İTEMS" while a later getCachedObject("ITEMS") folds
+            // to "ITEMS" - the keys do not match and the cache misses every table,
+            // column or schema whose name contains i/I.
+            return name.toUpperCase(Locale.ENGLISH);
         }
         return name;
     }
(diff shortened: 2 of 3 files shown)


E1izabeth:
Thanks for the contribution!

E1izabeth:
Why is it closed? Are you going to open a new PR with a new commit message following our standards?

boskodev790:
@E1izabeth thanks for the nudge - sorry for the closed/reopened churn, that was on me.

I just rewrote the branch on top of current `devel`:

- Reworded the commit message to start with `#40901` per the contributor guide.
- Dropped the stale merge commit so the diff is just the actual fix.
- Switched both folds to `Locale.ENGLISH` instead of `Locale.ROOT` to match the existing `renameObject` site in the same `AbstractObjectCache` (lines 141-142), so the whole class folds names with one locale.

Head is `261e2b63e3`. The fix itself is unchanged - just the two `String.toUpperCase()` calls in `AbstractObjectCache#getCachedObject` and `#getObjectName`, plus the one in `QueryTransformerLimit#transformQueryString`, with the new `AbstractObjectCacheLocaleTest` pinning the Turkish-locale behaviour.

Could you re-run CI when you get a chance? Happy to follow up on anything Codacy flags.

E1izabeth (approved):

E1izabeth (approved):

ShadelessFox (commented):
I'm totally fine with this fix, and I understand its purpose, but I'm not fond of such long comments that try to explain every single detail of a single-line fix, while also going deep into details.

Also, per [String#toLowerCase](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#toLowerCase--)'s documentation, it's recommended to use `Locale.ROOT` instead.

Please remove comments (since the reasoning is clear without them) and replace `ENGLISH` locale with `ROOT`.

E1izabeth:
@boskodev790 would you address the review comment?

PR #40897, opened 2026-04-24
#40894 Fix plain text record view for multiple selected rows

### LINKED ISSUE #40894: RECORD mode: GRID vs TEXT view does not match
### Description

Dear friends,

Dbeaver is a great product that help users to read easily the data in large tables. 

Let's say we have a table with large number of columns like this ...
Image

to see all columns at once we use RECORD view and sometimes with more rows at once ... like this two rows ...
Image

and sometimes we want to copy those data, so we switch to TEXT view ... However only one record is displayed, this is anoying and lowers the value of dBeaver product :-(
Image

Would be nice to fix it.

Thanks.

### DBeaver Version

Version 26.0.2.202604051735

### Operating System

macOS latest

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

EastLord:
Closes #40894

## Summary

Plain Text presentation in Record mode only rendered the current row, while the Spreadsheet/Grid record view can show all selected rows. This made the Text view inconsistent and prevented users from copying multiple record rows in the same layout.

This change makes Plain Text Record mode render all selected records as separate `Row #N` value columns, matching the Grid record view semantics.

## Changes

- Render all selected records in Plain Text Record mode instead of only `currentRow`.
- Use `Row #N` headers for record value columns to match the Grid record presentation.
- Update cursor handling so selecting a value column in Plain Text Record mode updates the current row and panels.

## After fix 
image

image


E1izabeth (approved):

ShadelessFox (approved):
Thanks :^) passing to QA

Matvey16:
Looks like this feature doesn't work for me. Built from this PR. No errors in the log



https://github.com/user-attachments/assets/d7167868-f538-40fc-9e90-aea322f42919

EastLord:
@Matvey16 I have updated to the latest code and tested the fix locally with MySQL.
It works as expected on my side: in Record mode, switching from Grid to Text now shows all selected rows as `Row #N` columns. I also recorded a video showing the behavior.


https://github.com/user-attachments/assets/6ff70dee-6b96-4060-afb2-0347c4b1f55d

Matvey16:
@EastLord I'll try to rebuild it and check again. Maybe there was a problem on Jenkins. Or may be there is something in the code that makes this work only in dev enviroment

Matvey16:
Verified

It was actually an issue with the build

EastLord:
> Verified
> 
> It was actually an issue with the build

@Matvey16  Thank you for your reply.

PR #40903, opened 2026-04-24
Fix/postgresql search path missing schemas

### LINKED ISSUE #40898: AWS RedShift: Can't change default schema
### Description

Hi,

We're working on an AWS RedShift instance that does _not_ have a `public` schema. When I try to change the default schema frop the dropdown in the menu bar, I get an error:

```
Error changing default database
  Error setting search path
    SQL Error [3F000]: ERROR: schema "public" does not exist
      ERROR: schema "public" does not exist
```

When querying `SHOW search_path` manually, I get `$user, public`, so it seems our Redshift instance's default search path is somewhat misconfigured, because `public` doesn't exist. Still, I wouldn't expect DBeaver to choke on this.

The relevant bit is here: https://github.com/dbeaver/dbeaver/blob/83e5c3eab7c79ba59565503f28cecba3a4c40f32/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java#L234-L271

Clearly non-existing schemas should be ignored when setting `search_path` again.

But before we get to the fix, I noticed that the Postgresql is the only implementation that prepends the selected value to the existing search path value. Other implementations (e.g. MySQL), explicitly sets the search path to the newly selected value and only that. I can't see a reason why one would want to stack up the schemas and I believe it can be misleading. Consider this:

- user chooses `abc` as his default schema: search path will be `abc, $user, public`
- user executes stored proc `foo` which exists in `abc`
- user chooses `xyz` as his default schema: search path will now be `xyz, abc, $user, public`
- user attempts to execute `foo` in `xyz`, which he believes has a different implementation. But let's assume `xyz.foo` doesn't exist.
- This will inadvertently execute `abc.foo`, which is unintended.

I notice that @serge-rider did work on this section in https://github.com/dbeaver/dbeaver/pull/40170 so I'm hesitant just to undo that without his input, but I want to propose that we align the PG implementation to the other implementations: if the user choses `abc` as his default schema, then the search path becomes only `abc`. That implicitly solves the error that I receive above too.

### DBeaver Version

Community Edition 26.0.2

### Operating System

Windows 10 64b

### Database and driver

AWS RedShift

### Steps to reproduce

_No response_

### Additional context

_No response_

thomascolden585-svg:
### Summary

When setting the default schema on PostgreSQL/Redshift, the driver built a `search_path` that could retain stale or invalid entries (for example, `public` on Redshift when that schema is absent). The execution context now issues `SET search_path` with only the chosen default schema and keeps the in-memory `searchPath` list aligned with that behavior.

### Problem

- `addSearchPath` duplicated logic and could leave the client-side `searchPath` out of sync with what was actually set on the server, or include schemas that are not present on the current database.
- The previous `setSearchPath` logic merged the new default with the old path list, which could propagate invalid path components.

### Solution

- Remove the redundant `addSearchPath` call and delete the `addSearchPath` helper.
- Simplify `setSearchPath` so the session `search_path` is set to the quoted default schema only, then `searchPath` is cleared and repopulated with that single schema to match the server.

### Files changed

- `plugins/org.jkiss.dbeaver.ext.postgresql/.../PostgreExecutionContext.java`

### Testing

- Connect to a Redshift (or other PostgreSQL-compatible) database where the `public` schema is missing; set a non-`public` default schema and confirm no errors from references to a non-existent `public` in `search_path`.
- For a normal PostgreSQL database, set the default schema and confirm metadata and object resolution behave as before.

### Related

- Fixes / closes #40898

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
@@ -145,7 +145,6 @@ public class PostgreExecutionContext extends JDBCExecutionContext implements DBC
 
         if (!schema.isSystem() && !schema.isPublicSchema()) {
             setSearchPath(monitor, schema.getName());
-            addSearchPath(schema.getName());
         }
 
         final PostgreSchema oldActiveSchema = getDefaultSchema();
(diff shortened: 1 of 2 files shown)


serge-rider:
Proposed solution may break active schema switch in PostgreSQL and other its forks.

We'll try another approach to fix this issue.
Thanks for the contribution in any case

PR #40904, opened 2026-04-24
SourceCode data export: escape chosen quote character and guard missing property

boskodev790:
```
## Summary

DataExporterSourceCode (the PHP / array-literal exporter) has two
user-visible bugs around the `quoteChar` setting exposed in plugin.xml
(valid values: `"` and `'`):

- **NullPointerException during init** when PROP_QUOTE_CHAR is absent
  from the property map (programmatic invocations / saved task specs).
  The peer CSV exporter handles the same property with a null-check;
  this PR adopts the same pattern.
- **Invalid PHP output when the user picks `'` as quoteChar.** Values
  were piped through `JSONUtils.escapeJsonString` which does not
  escape `'`, so `O'Brien` became `'O'Brien'` (syntax error). The JSON
  escape also inserts `\n`/`\"`/`\t` sequences that PHP single-quoted
  strings treat as literal two-character pairs, corrupting backslash
  values like `C:\path`.

Extracts `escapePhpString(value, quoteChar)` and routes all three
escape sites (column key, scalar cell, streamed content) through it.

## Reproduction

### Bug 1 — NPE on programmatic export

Invoke the SourceCode exporter with a property map that omits
`quoteChar` (e.g. via the data-transfer API or a saved export task
whose non-default property set has been trimmed). `init` throws
NullPointerException on the `.equals(...)` call at line 92.

### Bug 2 — syntax-invalid PHP

1. Result set containing `{"name": "O'Brien"}`.
2. Export → Source code, pick PHP and set Quote char to `'`.
3. Open the output file; it contains:
   `'name' => 'O'Brien'`
   — run `php -l` or open in any PHP parser: "Parse error: unexpected
   identifier Brien". Expected: `'name' => 'O\'Brien'`.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

## Testing

- [x] `DataExporterSourceCodeTest` added (5 cases, mirrors the Mockito
      pattern in the existing DataExporterDbUnitExpertRowTest).
- [x] Manually traced: programmatic invocation without `quoteChar`
      property no longer throws; `'O\'Brien'` round-trips through
      `php -l`.
```


E1izabeth:
It's required to reference the issue and add the issue number as a prefix to each commit message.
github.com/dbeaver/dbeaver/wiki/Contribute-your-code#commit-message-rules

E1izabeth:
This PR is stale, so I'm closing it. Please, let me know if you continue working on this - we can reopen it.

PR #40905, opened 2026-04-25
#40874 Clear stored proxy credentials when user empties the fields

### LINKED ISSUE #40874: Driver proxy username and password can't be cleared

boskodev790:
```
## Summary

`GlobalProxyAuthenticator.saveCredentials` short-circuited on an empty
username or password, so clearing the proxy fields on the Drivers
preference page never removed the stored secret. Next time the page
was opened, `readCredentials` returned the stale values and the UI
re-populated the user's just-cleared fields.

Fix: always write both values. Mapping empty strings to null triggers
`LocalSecretController.setPrivateSecretValue` to call
`preferences.setToDefault`, which removes the stored entry — the same
contract the interface already documents via `@Nullable String
secretValue`.

Extracts a package-visible overload that takes the
`DBSSecretController` directly so the clear-on-empty behaviour can be
pinned by a unit test without standing up the platform workbench.

## Related Issues

Closes #40874.

## Reproduction (pre-fix)

1. Window → Preferences → Drivers.
2. Set proxy URL, username, password. Apply + close.
3. Reopen preferences, remove username and password, Apply + close.
4. Reopen preferences — old username and password reappear.

After the fix both entries are cleared when the user empties the
fields, so the page reopens with blank inputs.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation

## Testing

- [x] GlobalProxyAuthenticatorTest added (4 JUnit cases covering
      non-empty save, full clear, cleared-password-only,
      cleared-username-only — all backed by an in-memory secret
      controller that mirrors LocalSecretController semantics).
- [x] Manually traced the call path from PrefPageDrivers.performOk
      through the cleared-text case to confirm
      LocalSecretController.setPrivateSecretValue(id, null) hits
      preferences.setToDefault on both keys.
```


E1izabeth:
There is another PR that was created to solve #40874
It was created earlier, so we will accept that one if it passes review and testing.

E1izabeth:
Another PR was merged, so I'm closing this one.

PR #40918, opened 2026-04-27
#36267 Disable Save/Cancel buttons in Data Editor for read-only databases

### LINKED ISSUE #36267: Disable Save/Cancel buttons in Data Editor for read-only databases
When a database is read-only, we can double-click a cell and Save and Cancel buttons become active, the message to create a Virtual key appears.

We should disable the buttons, because these actions aren't possible.

MilosM348:
```
## Summary

`ResultSetPropertyTester.PROP_CAN_PERSIST_DATA` only checked
`isUpdateInProgress`, so the apply/reject toolbar handlers stayed
enabled the moment a row went dirty even when the underlying data
source was read-only. Double-clicking a cell on a read-only connection
therefore lit up Save and Cancel even though neither action could ever
succeed.

## Fix

Tighten `PROP_CAN_PERSIST_DATA` to also require a non-null execution
context and a clear model-level read-only status. The check delegates
to `ResultSetModel.getReadOnlyStatus(container)`, which already
accounts for connection read-only mode, the `DBExecUtils` policy gate
and the "unique key required but missing" condition — so the buttons
disable themselves consistently for every read-only scenario the model
recognises.

The check is intentionally model-level rather than viewer-level
(`ResultSetViewer.getReadOnlyStatus()`), so the buttons remain enabled
when only the active *presentation* lacks edit support — dirty changes
can still be saved after switching back to an editable presentation.

`PROP_CAN_PERSIST_DATA` is referenced only by the apply/reject handler
`<enabledWhen>` clauses (`plugin.xml` lines 746 and 764), so the
tighter semantics are scoped to those two buttons.

## Reproduction

1. Connect to any database with the connection marked read-only
   (Connection settings → "Connection read-only" or any driver/role
   that enforces it).
2. Open a table editor and double-click a cell to edit a value.
3. Note the toolbar: **before** the fix, Save and Cancel become enabled
   the moment the cell goes dirty; **after** the fix, both buttons stay
   greyed out so the user is not misled into a save attempt that can
   never succeed.

## Closes

Closes #36267.
```

Closes #36267


E1izabeth (approved):

IvanGorshechnikov (approved):

IvanGorshechnikov:
@MilosM348 Thanks for your contribution
We pass your PR to the QA team

Matvey16:
Verified

PR #40919, opened 2026-04-27
#36721 Hide Delete in navigator context menu when metadata changes are disabled

### LINKED ISSUE #36721: Remove Delete context menu item in simple connection view mode
Based on the definition of a simple mode, you cannot delete objects when it is applied. None objects except databases don't have the Delete context menu item.
(You can't delete object because finally you'll get error message)

MilosM348:
```
## Summary

`ObjectPropertyTester.PROP_CAN_DELETE` permitted Delete on every
database object the active user could otherwise edit, even when the
connection's navigator settings disabled metadata changes (the
"simple-view mode" — `isHideFolders` / `isShowOnlyEntities`, captured
by `isMetadataChangeDisabled`). The matching `PROP_CAN_CREATE` branch
already short-circuits via the same predicate, so users in simple-view
mode saw a Delete action that always failed at execution time with
*"Object create not available in simple view mode"*-style errors.

## Fix

Tighten `PROP_CAN_DELETE` to mirror the create check: when the node
is a `DBNDatabaseNode` whose container has metadata changes disabled,
return false so Eclipse hides the menu item entirely. The guard slots
into the existing `node instanceof DBSWrapper` arm — `DBNDatabaseNode`
implements `DBSWrapper`, so the path is reached for every database
object that previously fell through to `DBEObjectMaker.canDeleteObject`.

`DBNDataSource` and `DBNLocalFolder` retain their dedicated branch at
the top of `PROP_CAN_DELETE` — they are governed by the project-level
`PROJECT_DATASOURCES_EDIT` permission and remain deletable in
simple-view mode, matching the issue's stated expectation that only
*"databases"* (i.e. connection nodes) keep the Delete action.

## Reproduction

1. Open a connection's settings → Navigator → enable any of
   "Hide folders" or "Show only entities" (the configurations
   `isMetadataChangeDisabled` covers).
2. Right-click any database object inside that connection — table,
   schema, view, index, etc.
3. Before: the Delete menu item is shown but invoking it surfaces
   an error dialog. After: the Delete menu item is hidden, matching
   the existing behaviour of New / Create in the same mode.

## Closes

Closes #36721.
```

Closes #36721


E1izabeth (approved):
Thanks for the contribution!

IvanGorshechnikov (approved):

IvanGorshechnikov:

PR #40921, opened 2026-04-27
#34402 Suppress read-only badge in data editor when only a unique key is missing

### LINKED ISSUE #34402: Confusing tooltip 'read-only' for tables without PK/VirtualPK
### Description

1. Open a table without any keys in postgres
2. edit some value and save changes
3. After virtual key creation the table is marked as editable

https://github.com/dbeaver/dbeaver/assets/43252954/7b7a7254-977d-426f-9537-7297186fbc4b

### DBeaver Version

24.1.1

### Operating System

win11

### Database and driver
Cockroach, PostgreSQL 

### Steps to reproduce

_No response_

### Additional context

_No response_

MilosM348:
```
## Summary

The Read-only status column in the spreadsheet data editor showed the
`BUTTON_READ_ONLY` icon whenever `ResultSetController.getReadOnlyStatus()`
returned anything non-null. `ResultSetModel.getReadOnlyStatus` encodes
several distinct reasons through one return value: container/connection
read-only mode, an in-flight update, and the soft "no unique key
defined" state that the user can resolve by creating a virtual key.

The Edit Key status column right next to it already signals the
missing-key state explicitly via `META_KEY_NA` and the inline
*"No unique key was found. Data modification is not possible."* text,
so users saw two read-only badges and a "Read-only" tooltip even when
the table was not actually read-only at the database level. After
virtual key creation the row identifier becomes valid,
`isUniqueKeyUndefinedButRequired` flips to false, and the editable
state is restored — but the misleading badge in the meantime hid the
recovery path from users.

## Fix

Suppress the `BUTTON_READ_ONLY` icon in the Read-only column when the
only reason for the read-only status is a missing-but-required unique
key. The Edit Key column remains the single source of truth for that
state, and a virtual key correctly flips the row set back to
editable.

Genuine read-only causes (connection-level, in-flight update) still
render the badge, because in those cases
`isUniqueKeyUndefinedButRequired` is short-circuited by an earlier
non-null branch in `ResultSetModel.getReadOnlyStatus`
(`ResultSetModel.java:142`).

The Read-only status text returned by `getStatusText()` is unchanged,
so the inline column text still shows the *"No unique key defined"*
detail when the user expands the status panel — only the badge icon
is suppressed.

## Reproduction

1. Connect to a PostgreSQL / Cockroach (or any) instance and open a
   table that has no primary key and no virtual key defined.
2. Note the bottom status bar: **before** the fix, both the Read-only
   badge AND the Edit Key META_KEY_NA badge appear, with a
   *"Read-only"* tooltip on the first one — making it look as if the
   database is read-only and hiding that the user can still edit by
   creating a virtual key.
3. **After** the fix, only the Edit Key META_KEY_NA badge is shown.
   Edit a value and save → virtual key is auto-created → both badges
   clear and the row set is editable, with no transient confusing
   tooltip.

## Closes

Closes #34402.
```

Closes #34402


E1izabeth (approved):

IvanGorshechnikov (approved):

IvanGorshechnikov:
@MilosM348 Thanks for your contribution
We pass your PR to the QA team

E1izabeth:
Hi, @MilosM348!
Please provide a clear description of the changed behavior for our QA team. It would be better to provide screenshots.
At least, please mention what badge you're talking about. Where to find it.

E1izabeth:
The original issue is not reproducible anymore
 So we don't understand what's the purpose of the provided changes.

PR #40923, opened 2026-04-27
dbeaver/pro#8775 Fix record mode for composites and arrays

E1izabeth:
It should touch only Record mode.
Please check tables with complex structure in PostgreSQL, Clickhouse, and MongoDB (at least).

Note: I haven't intentionally fixed the update operation in record mode. We're gonna do it in a separate ticket.


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/ResultSetValuePath.java
```diff
@@ -0,0 +1,61 @@
...
+) {
+
+    public interface PathItemVisitor<T, TRet> {
+        @Nullable
+        TRet visitAttributeItem(@NotNull PathAttributeItem attrItem, @NotNull T arg);
+        @Nullable
+        TRet visitIndexItem(@NotNull PathIndexItem indexItem, @NotNull T arg);
+    }
+
+    public interface PathItem {
```
ShadelessFox: Make it sealed
```suggestion
    public sealed interface PathItem {
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/ResultSetValuePath.java
```diff
@@ -0,0 +1,61 @@
...
+        @Nullable
+        <T, TRet> TRet apply(@NotNull PathItemVisitor<T, TRet> visitor, @NotNull T arg);
```
ShadelessFox: `R` is a de facto standard name for return generics
```suggestion
        <T, R> R apply(@NotNull PathItemVisitor<T, R> visitor, @NotNull T arg);
```

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/DBSContextBoundAttribute.java
```diff
@@ -39,4 +39,9 @@ String formatMemberReference(
             @Nullable String containerAliasOrNull,
             @NotNull DBPAttributeReferencePurpose purpose
     );
+
+    /**
+     * Returns real underlying attribute represented by this object
+     */
+    DBSEntityAttribute getUnderlyingAttribute();
```
ShadelessFox, line 47: Missing nullability annotation. I assume it's not null?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -768,6 +835,72 @@ public static Object getAttributeValue(
...
+        }
+
+        if (rootIndex < 0) {
+            log.debug("Failed to resolve root attribute index by its binding.");
+            return null;
+        }
+
+        Object value = row[rootIndex];
+        int currentIndex = 0;
+        for (ListNode<DBDAttributeBinding> node = path.next; node != null;) {
```
ShadelessFox, line 865: It partially reflects what `RSV_PATH_ITEM_VALUE_RESOLVER` does, and I dislike it, as it introduces duplication of already complex, hard-to-read code. :(

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -809,7 +942,17 @@ public static Object getAttributeValue(
 
             if (!(curValue instanceof DBDCollection)) {
                 if (remainingAttributes == 0) {
-                    return DBDVoid.INSTANCE;
+                    if (remainingIndices == 0){
```
ShadelessFox:
```suggestion
                    if (remainingIndices == 0) {
```

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/spreadsheet/SpreadsheetPresentation.java
```diff
@@ -2054,6 +2077,53 @@ public Object[] getChildren(@NotNull IGridItem item) {
...
+                    } else if (compositeAttr.getDataKind() == DBPDataKind.ARRAY && attributeValue instanceof Collection<?> collection) {
+                        if (subpathNode.next == null) {
+                            return collection.toArray();
+                        } else {
+                            // should never happen, because collection items already represented explicitly
+                            //                      and thus shouldn't be traversed through
+                            return null;
+                        }
+                    } else {
+                        return null; // should never happen
```
ShadelessFox, line 2136: Let's at least log these cases. If you want to go extreme, throw exceptions :^)

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/spreadsheet/SpreadsheetPresentation.java
```diff
@@ -2054,6 +2077,53 @@ public Object[] getChildren(@NotNull IGridItem item) {
             } else if (item.getElement() instanceof DBDComposite composite) {
                 // This happens in record mode and dynamic databases
                 return composite.getAttributes();
+            } else if (controller.isRecordMode()) {
```
ShadelessFox, line 2080: I don't understand why we need so much logic for handling a simple transpose operation :(

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
E1izabeth: As a concept - yes, but it works in different situations. While we didn't drop the existing getAttributeValue(..) implementation, which is used in non-record mode, we can face failures of its indexing approach, so this method tries to compensate. There is a comment about the corresponding situation where it is called.
We can get rid of this code as soon as the old getAttributeValue(..) method will be deprecated and removed.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/spreadsheet/SpreadsheetPresentation.java
E1izabeth: Because unfortunately, we don't really have an explicit contract on what can be carried by a grid item's value in which situations.
Simplifying all the learned stuff down to two points:
 - for addressing value locations, it is almost everywhere assumed that there are some of DBDAttributeBinding objects, but this is not true in fact (from what i happened to learn - there might be "leaf values" of typically trivial types, collection or composite instance itself, or even raw database-model-specific entity attribute - and i don't even know what i didn't face yet);
 - the record mode is not a transpose operation, because it still tries to use the same bindings hierarchy (which describes non-record-mode's columns!), while collection's elements are not being transposed at all, and composite fields from one column of non-record mode go to different rows of record mode, so "columns become rows and rows become columns" is not true at all.

ShadelessFox: But it feels like it should be a plain transpose where rows become columns and vice versa, at least that's what my intuition says. I don't understand the full picture, so I might be missing something obvious that would make my intuition wrong.

ShadelessFox (approved):

IvanGorshechnikov (approved):

PR #40929, opened 2026-04-27
#36224 Support SQLite system tables in query analyzer

### LINKED ISSUE #36224: Support SQLite system tables
The following well-known system tables is not correctly handled by query analyzer now
sqlite_master, sqlite_schema, sqlite_temp_master, sqlite_temp_schema, dbstat, sqlite_stat1, sqlite_stat2, sqlite_stat3, sqlite_stat4

We don't want to execute any metadata queries to know if the table is present and what are its columns. We just want to say that it has any potential columns.

The previous approach based on sending metadata queries https://github.com/dbeaver/dbeaver/pull/36060/commits/15ad9c33add8c296b79deae8dfb82ddcd51705c0

dhgoal:
## Summary

Adds support for well-known SQLite system tables in the SQL semantic analyzer so query references to them resolve cleanly even when the table is not currently present in the database. Notable cases: `dbstat` is a virtual table available only when SQLite is built with `SQLITE_ENABLE_DBSTAT_VTAB`; `sqlite_stat1..4` only materialize after an `ANALYZE`; `sqlite_sequence` only appears once the first `AUTOINCREMENT` column is created.

## Fix

`SQLiteDataSource` now implements `DBSVisibilityScopeProvider` and exposes a single `SQLiteSystemTablesContainer` of well-known system tables. Each table declares its columns statically from the SQLite documentation — **no metadata queries are issued**. The semantic analyzer (`SQLQueryConnectionRealContext.findRealObjectsImpl`) falls through to this scope only when its primary FQN lookup against the regular datasource containers comes up empty, so it never overrides real tables that are present in the database.

Tables covered:

- `sqlite_master`, `sqlite_schema`
- `sqlite_temp_master`, `sqlite_temp_schema`
- `sqlite_sequence`
- `dbstat`
- `sqlite_stat1`, `sqlite_stat2`, `sqlite_stat3`, `sqlite_stat4`

This addresses the design constraint articulated in the issue and on the prior reverted attempt (#35430, commit `15ad9c3`): "we don't want to execute any metadata queries to know if the table is present and what are its columns. We just want to say that it has any potential columns."

## Reproduction

In a fresh SQLite database (or any database that has not yet been `ANALYZE`d):

```sql
SELECT * FROM sqlite_stat1;            -- previously highlighted as TABLE NOT FOUND
SELECT * FROM dbstat;                  -- previously highlighted as TABLE NOT FOUND
SELECT type, name, tbl_name FROM sqlite_master;  -- column references previously unresolved
```

After the change, each listed table — and its listed columns — resolves cleanly under the SQL editor highlighter.

## Closes

Closes #36224


E1izabeth (approved):
I don't like that with these changes, we will always pretend that these tables exist.
However, another approach with metadata reading was declined. So, let's move with these changes for now.

ShadelessFox (commented):
It's indeed weird that we include tables that might not exist due to a lack of installed extensions or a database version mismatch. Why are we against issuing database queries? @E1izabeth

ged-yuko:
I would rather prefer potentially missing tables as errors than having them proposed in autocompletion. Especially for tables which don't always exist by default. 

Don't see any reasons why its ok to query other table's metadata but not these. Or why its ok to query for the public scope in oracle but not this.

E1izabeth:
This PR cannot be built due to
`dbeaver/plugins/org.jkiss.dbeaver.ext.sqlite/src/org/jkiss/dbeaver/ext/sqlite/model/SQLiteSystemTablesContainer.java:[270,25] org.jkiss.dbeaver.ext.sqlite.model.SQLiteSystemTablesContainer.SQLiteSystemTableColumn is not abstract and does not override abstract method getDataKind() in org.jkiss.dbeaver.model.struct.DBSTypedObject`

Please fix it

PR #40931, opened 2026-04-27
#40922 Fix script execution breaking SQL Server variable scope

### LINKED ISSUE #40922: Cannot execute script with variable declarations
### Description

When I have a script (batch) for SQL Server that contains variable declaration

```
DECLARE @month DATETIME = '2026-03-01';
SELECT @month;
```

I can execute it by selecting everything with CTRL+A and then issuing CTRL+Enter but this is not optimal (too much hassle having to select everything every time). So I have tried the option Execute SQL Script (ALT+X) but this gives error:

```
Error occurred during SQL query execution
SQL Error [137] [S0002]: Must declare the scalar variable "@month".
Must declare the scalar variable "@month".
```

### DBeaver Version

Community Edition 26.0.1

### Operating System

Windows 11

### Database and driver

com.microsoft.sqlserver.jdbc.SQLServerDriver 13.2.1.jre11

### Steps to reproduce

Create script

```
DECLARE @month DATETIME = '2026-03-01';
SELECT @month;
```

Issue ALT+X

alonfaraj:
## Summary

Fixes #40922 

Spliting scripts on `;`, sending each statement as a separate JDBC call. This breaks T-SQL batch semantics and variables declared with `DECLARE` are only visible within the same batch.

SQL Server use `GO` as the batch delimiter, not `;`. Semicolons are statement terminators within a batch, not batch separators.

## Changes

- Add `SQLDialect.usesOnlyNativeScriptDelimiters()` - lets dialects opt out of the preference-based default `;` delimiter
- Override it in `SQLServerDialectBase` to return `true`, and change `getScriptDelimiters()` to return only `GO`
- Guard the extra-delimiter logic in `SQLSyntaxManager.init()` with the new flag
- Update existing tests and add a regression test for the scenario from the issue

## Tests

- [x] `parseSqlServerVariableDeclarationBatch` - `DECLARE @x; SELECT @x;` stays as one batch
- [x] `parseSqlServerVariableDeclarationBatch` - `GO` still splits into separate batches
- [x] `parseSqlServerVariableDeclarationBatch` - PostgreSQL still splits on `;`
- [x] `parseBeginTransaction` - updated for new SQL Server behavior
- [x] `parseFromCursorPositionBeginTransaction` - updated for new SQL Server behavior

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDialectBase.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDialectBase.java
@@ -117,12 +117,13 @@ public abstract class SQLServerDialectBase extends JDBCSQLDialect implements TPR
     @NotNull
     @Override
     public String[] getScriptDelimiters() {
-        return new String[]{"GO"};
+        return new String[]{";", "GO"};
     }
 
+    @Nullable
     @Override
-    public boolean usesOnlyNativeScriptDelimiters() {
-        return true;
+    public String[] getScriptBatchSeparators() {
+        return new String[]{"GO"};
     }
 
     @Override
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDialectBase.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDialectBase.java
@@ -117,7 +117,12 @@ public abstract class SQLServerDialectBase extends JDBCSQLDialect implements TPR
     @NotNull
     @Override
     public String[] getScriptDelimiters() {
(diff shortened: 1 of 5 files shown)


E1izabeth (approved):

E1izabeth (changes requested):
It leads to all statements that go without a blank line being treated as one: take a look at the folding.

Image

alonfaraj:
Thanks for catching that! 
You're right - removing `;` from the delimiters broke folding. 
I reworked the approach: 
`;` is now kept as a delimiter for the editor (so folding and highlighting work normally), but during script execution (ALT+X), statements between `GO` separators merged back into a single batch before being sent to JDBC.
So now the editor still sees individual statements, but the execution engine respects T-SQL batch scope.

E1izabeth (commented):
Thanks for trying to solve the issue!
I have tested the provided changes. However, I'm unsure we can accept it as it breaks the current behavior. In DBeaver, the new line is a statement separator by default. With the provided changes, only GO separate queries in script mode. So, if I have scripts that worked in DBeaver with only newlines as separators, they may stop executing as before.

PR #40934, opened 2026-04-28
dbeaver/pro#5090 Make AWS fs id case-insensitive

Nexus6v2:
Closes dbeaver/pro#5090


serge-rider:
This is quite odd fix.
It affects only file systems but not the rest. Thus it may lead to a mess in cloud configs.

At least we should restrict creation of clouds with the same id/name in different cases.
And fix CPCloudUtils

Nexus6v2:
@serge-rider I've added new PR adressing the improvements
https://github.com/dbeaver/dbeaver-ee/pull/4499

PR #40944, opened 2026-04-28
#21830 Disable Copy button while Generate DDL not completed

### LINKED ISSUE #21830: Disable Copy button while Generate DDL not completed
If you press the Copy button when DDL is not already generated, then the copied content would be 'Loading DDL ...'. This is a bit incorrect behavior

Better to disable this button while editor displays 'Loading DDL ...'
image

dhgoal:
## Summary
Disable the Copy button in the Generate DDL dialog while the initial DDL generation is still running, so it cannot copy the `Loading DDL...` placeholder text.

## Fix
`startGenerateJob()` is invoked from `createDialogArea()`, which JFace runs **before** `createButtonsForButtonBar()`. Its `getButton(IDialogConstants.DETAILS_ID)` call therefore returned `null` on initial open and the `setEnabled(false)` was silently dropped — the Copy button stayed enabled until the async job completed. (The disable did work for subsequent regenerations triggered by the option checkboxes, because by then the buttons existed.)

This change disables the Copy button right after `super.createButtonsForButtonBar(parent)` creates it. The existing job-completion handler enables it once the DDL is ready.

## Reproduction
1. Open any table → Generate SQL → DDL.
2. While the editor still shows `Loading DDL...`, click **Copy**.
3. Before this change, the clipboard contains the literal string `Loading DDL...`.

## Closes
Closes #21830.


Matvey16:
Verified

PR #40948, opened 2026-04-28
#40920 Preserve newly-created object in navigator on parent-node refresh

### LINKED ISSUE #40920: Edited object editor closes and disappears from the navi tree (if newly crated) when tree node is refreshed
### Description

This is an old bug. Reproduced on 25.0.0
https://github.com/user-attachments/assets/018a1254-d319-46cc-af6b-c6ea016d0169

### DBeaver Version

26.0.3

### Operating System

win11

### Database and driver

not depends

### Steps to reproduce

1. Start a new object creation or edit existing
2. Refresh some tree node
Result: the new object disappreared from the tree (if it was newly created) and editor also closed (with the exception of Views: when editing a View, the new object’s editor remains open)

### Additional context

No errors in log

Daedalus-Icarus:
Closes #40920

### Problem

When the user starts creating a new object (table, sequence, FK, index, etc.)
and then refreshes a parent tree node before saving, the navigator drops the in-progress object and the entity editor closes — losing the user's edits.

The new object is wrapped in a `DBNDatabaseItem` whose underlying `DBSObject`  returns `isPersisted() == false`. On refresh, `DBNDatabaseNode.loadTreeItems()` re-queries the children from the database, which can't see the unpersisted
object yet. The matching loop fails to pair it with anything in the fresh result set, so the cleanup loop calls `DBNUtils.disposeNode(...)` on it. That fires a removal event, which closes any editor whose input was that node.

There is already a precedent guard against this in `NavigatorHandlerRefresh.execute()`: when the user explicitly refreshes the editor's own node, it short-circuits with "Do not refresh non-persistent objects". The same principle was missing one level deeper, in the parent-node cleanup path.

### Fix

In `DBNDatabaseNode.loadTreeItems()`, before disposing an old child that isn't present in the freshly-fetched item list, check whether its underlying `DBSObject` is non-persisted. If so, preserve it in the new child list. The cross-cutting check is extracted into a small `isNonPersistedChild` helper for readability and to mirror the surrounding `equalObjects` helper style.

### Verification

Manual repro on a generic JDBC connection (steps from the issue):

1. Right-click a schema → Create New Table → entity editor opens for the new   non-persisted table.
2. Right-click the parent schema → Refresh.

- **Before this change:** editor closes, new object disappears from the tree.
- **After this change:** editor stays open, new object remains in the tree.
  Saving then refreshing then continues to work correctly (no duplicate node);
  closing the editor without saving correctly removes the unpersisted node.

Same result for sequence/index/FK creation, which all flow through the same `loadTreeItems` cleanup path.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNDatabaseNode.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNDatabaseNode.java
@@ -764,6 +764,14 @@ public abstract class DBNDatabaseNode extends DBNNode implements DBNLazyNode, DB
                     // Wrong type
                     continue;
                 }
+                if (isNonPersistedChild(oldChild)) {
+                    // Preserve newly-created (non-persisted) child across refresh.
+                    // Its underlying DBSObject isn't visible to the database query yet,
+                    // so the matching loop above can't find it. Disposing it would close
+                    // the in-progress entity editor and lose the user's work (#40920).
+                    toList.add(oldChild);
+                    continue;
+                }
                 boolean found = false;
                 for (Object childItem : itemList) {
                     if (childItem instanceof DBSObject && equalObjects(oldChild.getObject(), (DBSObject) childItem)) {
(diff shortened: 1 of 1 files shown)


Daedalus-Icarus:
Hi @DariaMarkaryan please review this.

E1izabeth (approved):

ShadelessFox (approved):
Thanks, it's a nice addition (or a fix, if you like). Passing to QA

PR #40958, opened 2026-04-29
dbeaver/pro#8899 disable autocommit toggle for db2i

Nexus6v2:
Closes dbeaver/pro#8899


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/exec/DBCNativeBatchAutoCommit.java
```diff
@@ -0,0 +1,24 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.exec;
+
+/**
+ * Interface for data sources that must not have their autocommit state
+ * toggled off/on around batch updates.
+ */
+public interface DBCNativeBatchAutoCommit {
```
serge-rider: It is not clear what is so special in DB2i.
What exactly operation breaks data edit and why it is in CB only.

And I'd better use driver properties to manage such things than new interfaces.

serge-rider (changes requested):

yagudin10 (approved):

PR #40967, opened 2026-04-30
dbeaver/dbeaver#39968 Shell command authentication model for database connections

### LINKED ISSUE #39968: Dynamic Password (Shell Command) for Database Native Authentication
The possibility of supplying the password as a Shell Command, thus making it dynamically use whatever the shell command returns with native db authentication would open up a lot of possibilities for custom integrations that are not supported by DBeaver Community or DBeaver Pro yet

kgns:
This PR adds the ability to provide a shell command that will be run and output the required db password to stdout, which will be used by DBeaver while providing credentials to the database server. 

This adds a kebab menu next to the password field in **Connection Settings** dialog, when clicked shows two options:
- **Type password manually**: default selected, which is the current way of providing a types static password
- **Use dynamic password from shell command output**: this is the newly added functionality, replaces password textbox with a textarea to enter a shell command, which will be evaluated any time we send credentials to the database server.

closes #39968

> Static password input (default)
> dbeaver_Ir0PuBDkNf

> Password provider command (empty textarea with hint)
> dbeaver_44Z1DJiRZY

> Password provider command
> dbeaver_52TKBqhl6K


E1izabeth:
Hi, @kgns! Sorry for long delay. We will review your PR and let you know

kgns:
meanwhile, for people that want to test it, I have a [github workflow](https://github.com/kgns/dbeaver/blob/devel/.github/workflows/build-from-upstream.yml) that builds on top of every new DBeaver release with these changes in my [fork](https://github.com/kgns/dbeaver/releases). no signatures or installers of course, so they are only for testing purposes

E1izabeth (dismissed):

E1izabeth:
We've discussed the provided solution. At first glance, it seems useful, but it’s potentially dangerous. It affects all password‑based authentications and is too intrusive as the button might confuse users.

We would like this implemented as a separate auth model that doesn’t include an explicit “password” field. Instead, it would use a command.

kgns:
Alright, can you give a little more detail so I can implement it?

At first I thought adding a new shell command to the shell commands dialog for this, but then there had to be a switch of some kind (checkbox?) in the credentials page telling DBeaver to use this shell command as password, or implicitly use it if the shell command is not empty. Both of those options seemed like a bad user experience, plus there are no authentication method list available to choose from, only native password authentication. That's why I added only a small button on top of the default single auth method dialog (also to not waste time on complex flows until I got a PR review).

If you can suggest a flow, I can implement it

E1izabeth:
The desired solution is to have a special Authentication method in this switcher
image

So when you switch from Username/password to the new one, it would have fields Username and Command.

You can take a look at `org.jkiss.dbeaver.model.impl.auth.AuthModelDatabaseNative`, and how the authentication methods are declared in `plugin.xml` with `<extension point="org.jkiss.dbeaver.dataSourceAuth">`

E1izabeth (dismissed):

kgns:
I thought the authentication method dropdown did not exist in community edition, but I guess it was hidden due having only 1 auth method registered. I created a new auth method for this type of authentication, force pushed over the old commits as they did something totally different, easier to review this way without creating a new PR.

Authentication dropdown:
image

Credentials for the new auth method
image

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/ShellCommandAuthModelConfigurator.java
```diff
@@ -0,0 +1,101 @@
...
+        final String hintText = UIConnectionMessages.dialog_connection_auth_shell_command_hint;
+        commandText.addPaintListener(e -> {
+            if (commandText.getCharCount() > 0) {
+                return;
+            }
+            Color hintColor = commandText.getDisplay().getSystemColor(SWT.COLOR_WIDGET_DISABLED_FOREGROUND);
+            Color savedFg = e.gc.getForeground();
+            e.gc.setForeground(hintColor);
+            e.gc.drawText(hintText, 3, 0, true);
+            e.gc.setForeground(savedFg);
```
ShadelessFox: You don't need to restore the old state within your paint callback

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/auth/AuthModelDatabaseShellCommand.java
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * The username is stored like the native model; the command is stored as an auth property.
+ * No password is ever persisted - it is resolved on each connect from the command's stdout.
+ */
+public class AuthModelDatabaseShellCommand<CREDENTIALS extends AuthModelDatabaseShellCommandCredentials>
+    extends AuthModelDatabaseNative<CREDENTIALS> {
+
+    public static final String ID = "native_shell_command";
+
+    private static final int COMMAND_TIMEOUT_MS = 60_000;
```
ShadelessFox: This should probably become configurable at some point.

plugins/org.jkiss.dbeaver.registry/plugin.xml
```diff
@@ -82,6 +82,17 @@
             </propertyGroup>
 
         </authModel>
+        <authModel
+            id="shell_command"
+            label="Password from shell command"
+            description="Database authentication; the password is read from the stdout of a shell command run at connect time"
+            class="org.jkiss.dbeaver.model.impl.auth.AuthModelDatabaseShellCommand">
```
ShadelessFox: Consider also specifying

```suggestion
            class="org.jkiss.dbeaver.model.impl.auth.AuthModelDatabaseShellCommand"
            desktop="true">
```

so it won't appear as a possible auth option in CloudBeaver. Pretty sure it's quite insecure... :^)

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/auth/AuthModelDatabaseShellCommand.java
kgns: made this configurable, also added working directory to make it more aligned with the existing shell commands feature

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/dialogs/connection/ShellCommandAuthModelConfigurator.java
kgns: this was a hacky fix remnant from the previous implementation, was no longer needed, removed it

plugins/org.jkiss.dbeaver.registry/plugin.xml
kgns: done (I did not know what cloudbeaver was, sorry 😄)

ShadelessFox:
Can you please include fresh screenshots?

kgns:
sure
image

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/internal/UIConnectionMessages.properties
```diff
@@ -61,6 +61,10 @@ dialog_connection_auth_group_user_cridentials = User Credentials:
 dialog_connection_auth_label_password = Password:
 dialog_connection_auth_label_username = Username:
 dialog_connection_auth_label_show_password = Show password on screen
+dialog_connection_auth_shell_command_label = Command
+dialog_connection_auth_shell_command_hint = The output (stdout) of this shell command will be used as the password.
```
ShadelessFox: Maybe it's also worth specifying that variables can be used there?

kgns: I did that initially, but then realized it would be redundant because there already was one at the bottom of the panel and removed it

kgns: ah you meant in the hint text, sure I'll add it

I added an extra link for the popup dialog, thought you also were also suggesting that

ShadelessFox: My fault for not clarifying that explicitly - yes, in the tooltip. I'm not sure what popup dialog you're talking about, though...

kgns: this one:
image

ShadelessFox: Right, I don't know what I was thinking. I think it's fine without a tooltip, then.

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/internal/UIConnectionMessages.properties
kgns: Since the hint text is just a drawText call, currently it does not redraw according to the width of the command multiline text box. Adding more information to it requires word-wrapping to be visible. This can be done in the paint listener by calculating where to add line breaks after every width change (resize), redrawing it, but this might be an unwanted implementation. The alternative is to add the hint text below the command text box using a Label only

drawText with word wrapping:
image

Label only:
image

(the hint text can be shorter, I left it long to test it visually)
which one would you prefer? or do you have any other suggestions?

ShadelessFox: I'd pick the former option because it just looks good. But I don't think it's worth the effort to implement custom text wrap logic, though it might be useful elsewhere, too. Please put it in `UIUtils` if you're willing to continue with that solution.

kgns: done

image

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/UIUtils.java
```diff
@@ -2422,6 +2422,58 @@ public static void drawTextWithBackground(@NotNull GC gc, @NotNull String text,
         gc.drawRoundRectangle(centerX - 3, centerY - 3, size.x + 5, size.y + 5, 5, 5);
     }
 
+    /**
+     * Word-wraps {@code text} so that no line exceeds {@code maxWidth} pixels when measured with
+     * {@code gc}. Existing newlines are preserved; words longer than the line are left intact.
+     * Returns the text unchanged when {@code maxWidth} is not positive.
+     */
+    @NotNull
+    public static String wrapText(@NotNull GC gc, @NotNull String text, int maxWidth) {
```
ShadelessFox, line 2542: We already have `org.jkiss.utils.StringUtils#wrap`

kgns: `org.jkiss.utils.StringUtils#wrap` function wraps the string according to a max character length, but has no idea what the font is, what the characters' widths are in pixels, or whether they are monospaced or not. `UIUtils.wrapText` function takes a GC object and does a pixel calculation to wrap the string so it fits a UI element according to its width in pixels, works with any font (i.e. any platform/OS). Since it is a wrap text function for a UI element, I thought it was best placed in UIUtils as well

While they are not interchangeable, I can delete the wrapText function and try to make `org.jkiss.utils.StringUtils#wrap` work for this case, but it may work a bit wonky

ShadelessFox (approved):

E1izabeth (approved):

E1izabeth:
Thanks! Passed to QA team for testing

kgns:
hello, any love from the QA team yet?

E1izabeth:
Unfortunately, it has not been tested yet :(

uslss:
verified
image

kgns:
what's the next step?

PR #40978, opened 2026-05-02
#40950 Add optional SSH agent socket path configuration for SSH tunnel

### LINKED ISSUE #40950: Add optional custom SSH agent socket path for SSH tunnel authentication (macOS / Bitwarden agent)
Yes. On macOS, when DBeaver is launched from Finder/Dock/Spotlight, the app process may not inherit the expected SSH_AUTH_SOCK value.
As a result, SSH Agent authentication in SSH tunnels can fail or use the wrong agent socket.

This is a practical issue when using third-party agents such as Bitwarden or 1Password SSH Agent, where I need to explicitly select a specific socket path.
Currently, the SSH Agent configuration relies on environment discovery and does not allow overriding the socket path in UI.

When DBeaver is launched via Spotlight, SSH_AUTH_SOCK is not available in the app process, and SSH agent authentication fails.

Image

Image

When I launch DBeaver from Terminal with SSH_AUTH_SOCK exported, SSH agent authentication works correctly.

Image

Image

Image

Adds an optional configuration in SSH tunnel settings (Connection Configuration, Agent auth mode):

A toggle: Use custom SSH agent socket
A text field: SSH socket path
Behavior proposal:

If toggle is OFF: keep current behavior (auto-detect via environment, e.g. SSH_AUTH_SOCK)
If toggle is ON: use the configured socket path explicitly
This should improve compatibility with Some third-party SSH Agent.

Launching DBeaver from terminal
(works, but is inconvenient)
Switching to key-file auth instead of agent auth
(loses agent-based workflow and key management benefits)

Relying on macOS launchctl environment propagation
(doesn't work, unknown why, and it’s also inconvenient and not easy to use.)

Image

**Additional context**

Environment: macOS, DBeaver Community Edition 26.0.3.202604191653 
Scenario: SSH tunnel using Agent authentication
Agent: Bitwarden SSH Agent

LystranG:
Closes #40950 

Adds an optional `authSockPath` field to the SSH Agent authentication configuration, allowing users to specify a custom SSH agent socket path. When left empty, the default agent behavior is preserved.

The ssh auth socket path field is **not applicable on Windows** because:

1. Windows uses Pageant (PuTTY's agent) via Windows IPC, not Unix domain sockets
2. The bundled `junixsocket` native library does not include Windows DLLs

For i18n, all translations except for English and Chinese are AI-generated.

Updated the copyright year information.

Effect:
image
image
image
image


plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/AbstractSessionController.java
```diff
@@ -131,34 +133,38 @@ public DBPDataSourceContainer[] getDependentDataSources(@NotNull SSHSession sess
     }
 
     @NotNull
-    protected IdentityRepository createAgentIdentityRepository() throws DBException {
-        if (agentIdentityRepository == null) {
-            AgentConnector connector = null;
+    protected IdentityRepository createAgentIdentityRepository(@Nullable String authSockPath) throws DBException {
```
ShadelessFox, line 147: There are two things about these changes:
1. You broke caching. A single controller can be used for multiple sessions for tunneling traffic over the same tunnel. This means that each connection would now instantiate a new identity repository. Maybe it's fine, but it's worth thinking about.
2. ~~The socket path is only respected when `PageantConnector` has failed to instantiate; otherwise, the provided socket path is left unused.~~ It's worth adding a comment that it's not supported on Windows.

LystranG: for 1:
Thanks for the review. The original cache was valid before introducing `authSockPath`, because without that parameter, every call was equivalent and caching was safe. With `authSockPath` now configurable, the cache would need to be invalidated whenever the path changes, adding complexity with little benefit.

In my opinion, the actual session reuse happens through `ShareableSession`: once an SSH tunnel is established, subsequent requests for the same destination reuse the existing `ShareableSession` via reference counting — they never call `createNewSession()` or `createAgentIdentityRepository()` again. 

for 2:
Good point. I have added a Javadoc comment on `createAgentIdentityRepository`.

ShadelessFox: Okay, maybe the lack of caching is fine. You are right about reused sessions; it really only affects new connections that do NOT reuse sessions. But it shouldn't have any major impact on the overall performance, according to the source code of each connector.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/AbstractSessionController.java
```diff
@@ -130,35 +132,51 @@ public DBPDataSourceContainer[] getDependentDataSources(@NotNull SSHSession sess
...
+        AgentConnector connector = null;
 
-            try {
-                connector = new PageantConnector();
-                log.debug("SSHSessionController: connected with pageant");
-            } catch (Exception e) {
-                log.debug("SSHSessionController: pageant connect exception", e);
-            }
+        // PageantConnector only works on Windows.
+        // On Windows, Pageant takes priority and authSockPath is not applicable.
```
ShadelessFox, line 151: Since it's only available on Windows, maybe it would be better to also guard it with `RuntimeUtils.isWindows()` and put initialization of `SSHAgentConnector` in its `else` branch? It's optional, and you can ignore this suggestion. But since we're already here, it would be nice to have, and also to avoid redundant errors in the log caused by attempts to use Pageant on non-Windows platforms.

LystranG: Good idea, I have optimized this.

ShadelessFox (dismissed):
Thanks!

E1izabeth (dismissed):

aleksandrs-ledovskis:
Just to chime in that this is eagerly awaited from my end. Use case is similar to @LystranG, but I'm using the [Secretive](https://github.com/maxgoedjen/secretive) which also exposes custom `SSH_AUTH_SOCK`.

LystranG:
@E1izabeth  Hi, just following up on this PR. Any update on the status, or is there something blocking it? Would really appreciate this feature being merged — let me know if there's anything I can do to help move it forward. Thanks!

E1izabeth:
Hi! We're waiting for our QA team to test it. After that we merge it, if they verify that it's working fine.

LystranG:
> Hi! We're waiting for our QA team to test it. After that we merge it, if they verify that it's working fine.

Got it,thanks!

PR #40981, opened 2026-05-03
dbeaver/pro#8935 Navigator views, project selector

serge-rider:


plugins/org.jkiss.dbeaver.data.transfer/OSGI-INF/l10n/bundle_tw.properties
```diff
@@ -6,7 +6,7 @@ Bundle-Vendor	=	DBeaver Corp
 dataTransfer.consumer.database.description	=	資料庫表格
 dataTransfer.consumer.database.name	=	資料庫
 dataTransfer.consumer.stream.description	=	串流平面化檔案
-dataTransfer.consumer.stream.name	=	串流檔案
+dataTransfer.consumer.stream.name	=	文件
```
E1izabeth, line 9: Is it expected change?

PR #40988, opened 2026-05-04
#36891 Keep info label content when ProgressMonitor.done() clears curStatus

### LINKED ISSUE #36891: Items count in the metadata editor disappear after refresh
### Description

Steps to reproduce:
1. Open metadata editor for an object (table, schema etc.)
2. Items count label is displayed at the bottom toolbar
3. Press F5 or Refresh button
Actual result: Items count label disappeared.
To return the label with items count, click one of the buttons on the bottom toolbar

https://github.com/user-attachments/assets/0f1c5a4f-a3c4-48f0-b542-0c080337f1bd

### DBeaver Version

24.3.4

### Operating System

win10

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

davidv399:
## Summary
Closes #36891.

The items-count label at the bottom of the metadata editor toolbar disappears after pressing F5. Reporter's steps:

1. Open the metadata editor for any object (table, schema, etc.).
2. The `X items` label is shown in the bottom toolbar.
3. Press F5 - the label disappears. Clicking any other toolbar button (which calls `ProgressPageControl.activate(true)` and restores the cached `curInfo`) brings it back.

## Root cause

In `ProgressPageControl$ProgressVisualizer`, `visualizeLoading()` calls `setInfo(curStatus)` once more during `completeLoading()`. By the time `completeLoading` runs, the wrapped `ProgressMonitor.done()` has already set `curStatus = ""`. `ProgressPageControl.setInfo("")` then:

- does NOT update `curInfo` (the empty-info branch leaves it untouched),
- does call `listInfoLabel.setVisible(false)` and `setText("")`.

`ObjectListControl.setListData()` immediately afterwards calls `setInfo(getItemsLoadMessage(count))` to set the post-refresh count. That correctly delegates to the owner control and updates `curInfo`, but with the metadata editor's substituted progress panel the layout pass scheduled by the earlier `setInfo("")` / `hideControls(true)` leaves `listInfoLabel` hidden until the next `activate(true)` (which fires only when the user clicks a toolbar button or otherwise triggers focus-gained on the items viewer).

## Change

`plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/controls/ProgressPageControl.java`:

- `ProgressVisualizer.visualizeLoading()` now skips `setInfo(curStatus)` when `curStatus` is empty. The empty value is only ever produced by `ProgressMonitor.done()` and represents "no current task" rather than a status the user should see - so propagating it to the label is always wrong, not just in the metadata-editor flow.
- After this change the label keeps its previous content until downstream code posts a new `setInfo()` with the legitimate post-load message.

No change to `getItemsLoadMessage(0) == ""` (an empty count is still an empty label, matching the existing behaviour for empty folders).

## Test plan

1. Build the affected plugin:
   ```
   mvn -pl plugins/org.jkiss.dbeaver.ui -am clean verify -DskipTests
   ```
2. Launch DBeaver with the bundled `DBeaver Sample Database (SQLite)`.
3. Open any table from the navigator and switch to a tab that uses the item-list metadata view (Columns, Constraints, References).
4. Confirm the bottom toolbar shows e.g. `8 items`.
5. Press **F5**.
   - **Before this PR:** label disappears until you click another toolbar button.
   - **After this PR:** label stays visible and reflects the refreshed count.
6. Repeat for an empty folder (e.g. a schema with no procedures) - the label remains empty, as before.
7. Repeat after a structural change (add a column externally → F5) - the count updates to the new value.

## Notes

The fix is in shared `ProgressPageControl` infrastructure, so it benefits any view that uses the same item-list / progress chain (database object lists, ER diagram lists, search results), not just the metadata editor reported in the issue. No regression expected: the only behaviour that changes is that an empty `curStatus` no longer wipes the info label - which was never useful information for the user and was already overridden in every observable code path by the next non-empty `setInfo()` call.

## Attachment

https://github.com/user-attachments/assets/3ad1fd05-f6df-40f2-8a6d-a33d96d43f85


davidv399:
Hi @E1izabeth, could you please take a look at this when you have a moment? It's a small fix for #36891. Thank you very much!

E1izabeth (approved):
Thanks for the contribution!

ShadelessFox (approved):

davidv399:
Hi @E1izabeth @ShadelessFox , just a gentle ping on this PR since it has been about 1 week without updates. Whenever you have time, could you please take a look? Thanks!

MashaKorax:
verified DBeaver ce win

PR #40989, opened 2026-05-04
#34418 Make jump-server table editing discoverable and refresh on the fly

### LINKED ISSUE #34418: Jump servers UX improvements
### Description

1. The user was trying to double-click on table cells in order to edit the jump server host and other information. However, it was not obvious to the user that the text fields above the table had changed and that they should be writing information there. When the user tried to edit the host by double-clicking, they became confused and lost.

2. More a bug actually - last jump server host is not shown right in the table, until you reopen the wizard (video below)

https://github.com/dbeaver/dbeaver/assets/28875055/fc706899-fdb4-4d89-84b5-cbf83aeb0444

### DBeaver Version

DBeaver 24.1.0

davidv399:
## Summary
Closes #34418.

The SSH **Jump servers** table in the connection wizard has two related UX problems:

1. Users naturally double-click a row in the table expecting to edit the cell inline. Nothing happens — editing is actually done in the credentials fields *above* the table, but there is no visual cue. The user gets stuck.
2. The table column for the currently-selected jump host sometimes does not visibly reflect the new value typed in the fields above until the wizard is reopened. (Demonstrated in the video on the issue.)

## Change

`plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/SSHTunnelDefaultConfiguratorUI.java`:

- Add a single-line hint label directly above the jump-host table:
  *"Select a row and edit its details using the fields above. Double-click a row to focus the host field."*
  This makes the editing surface discoverable without any guesswork.
- Add a `DoubleClickListener` on `hostsViewer` that calls `credentialsPanel.focusHostField()` for the selected row, providing the obvious affordance the user was already trying to use. Focus moves to the host field and its current value is selected, so the user can immediately overwrite it.
- Add a public `focusHostField()` method on `CredentialsPanel` that delegates to `hostNameText.setFocus()` + `selectAll()`, with the usual disposed-widget safety check.
- In `refreshActiveConfiguration()`, follow the existing per-element `hostsViewer.refresh(wrapper)` with `hostsViewer.getTable().redraw()` as a defensive paint flush, addressing the stale-cell behaviour from the second part of the issue (the JFace per-element refresh occasionally does not propagate to the underlying SWT `Table` repaint on Windows when modify events arrive faster than the layout pass).

`plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/SSHUIMessages.java`:

- Add the new NLS key `model_ssh_configurator_label_jump_server_edit_hint`.

`plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/SSHUIMessages.properties`:

- Add the English text for the new key.

## Test plan

1. Build the affected plugin:
   ```
   mvn -pl plugins/org.jkiss.dbeaver.net.ssh.ui -am clean verify -DskipTests
   ```
2. Database → New Database Connection → any driver → SSH tab → enable SSH tunnel → expand the **Jump servers** group → click **+** to add a jump host.
3. The hint label is visible above the table.
4. Double-click any row in the table:
   - **Before this PR:** nothing happens; the user has no idea where to type.
   - **After this PR:** focus jumps to the **Host/IP** field above, current text selected.
5. Type a new hostname into the field:
   - **Before this PR:** the table column for that row may stay stale until the wizard is reopened.
   - **After this PR:** the row reflects each keystroke immediately.
6. Sanity:
   - Add several jump hosts, reorder with the ↑/↓ toolbar, delete one, switch between rows — fields and table stay in sync.
   - Save the connection, reopen the edit dialog — all jump hosts reload with the typed values (no change to `saveSettings()` path).
   - Switch the Auth method between Password / Public Key / Agent — table column updates, no NPE.

## Notes

- No translation updates in this PR; DBeaver falls back to English when a key is missing in locale `.properties` files. The i18n team can add translations after the English text is merged.
- No public API change — `focusHostField()` is a method on the package-private `CredentialsPanel` inner class, only called from `SSHTunnelDefaultConfiguratorUI` itself.
- Behaviour for empty / non-applicable cases is preserved: if no row is selected (`selected == null` in the double-click handler), the focus call is skipped; if `hostNameText` is disposed, `focusHostField()` is a no-op.

## Attachment

https://github.com/user-attachments/assets/29224aae-eef3-4d00-846b-b093812a5b9f


davidv399:
Hi @E1izabeth, when you have a moment, could you please take a look at this one? It addresses #34418, the jump-server UX issue you filed. Happy to adjust based on your feedback. Thank you!

plugins/org.jkiss.dbeaver.net.ssh.ui/src/org/jkiss/dbeaver/ui/net/ssh/SSHTunnelDefaultConfiguratorUI.java
```diff
@@ -183,11 +183,24 @@ public void expansionStateChanged(ExpansionEvent e) {
 
             UIUtils.createLabelSeparator(client, SWT.HORIZONTAL);
 
+            final Label editHint = new Label(client, SWT.NONE);
+            editHint.setText(SSHUIMessages.model_ssh_configurator_label_jump_server_edit_hint);
+            final GridData editHintLayout = new GridData(GridData.FILL_HORIZONTAL);
+            editHintLayout.horizontalIndent = 4;
+            editHintLayout.verticalIndent = 2;
+            editHint.setLayoutData(editHintLayout);
```
ShadelessFox: I don't think we're happy with its placement. It visually separates the action toolbar from the table itself. Perhaps moving it BELOW the table would be nicer? Also, consider using `UIUtils.createInfoLabel` instead.

davidv399: Yes, you are right, It would be nicer using another UIutil component as well as moving label below the table.
Thanks for your kind feedback! ❤️

davidv399:
Hi @ShadelessFox, thanks for the feedback! Pushed a follow-up commit that moves the hint below the table and switches it to `UIUtils.createInfoLabel` as you suggested. Could you take another look when you have a moment?

ShadelessFox:
Can you please show footage of its new look?

davidv399:
Sure, Please find out attached video.

https://github.com/user-attachments/assets/dee6f791-730c-448f-923b-8f6f30c63f50

ShadelessFox (commented):
OK, it's definitely better. Thanks :^)

ShadelessFox (approved):

E1izabeth (approved):

E1izabeth:

davidv399:
Hi @E1izabeth @ShadelessFox , just a gentle ping on this PR since it has been about 1 week without updates. Whenever you have time, could you please take a look? Thanks!

ShadelessFox:
> Whenever you have time, could you please take a look? Thanks!

We don't do QA; we have a dedicated QA team. They have about 30 external PRs at the moment, so it will take a while to review them all.

Matvey16:
Verified

AnnaPozdeeva:
Hi @jack-stormentswe! Thanks for being part of the DBeaver community!  We really appreciate your support and would love to send you a small thank-you gift. Just send a quick email to anna.pozdeeva@dbeaver.com with the following info: your GitHub nickname, full name, email, phone number, T-shirt size, and complete address (including apartment number, street, city, state, country, and postal code). Looking forward to hearing from you!

PR #40992, opened 2026-05-04
dbeaver/pro#9224: add ui for EntraID

SolovevAnton:
decided not to change api


SolovevAnton:
decided not to change api

PR #41000, opened 2026-05-05
dbeaver/pro#9018 Fix toolbar buttons rendering in dark theme

E1izabeth:
We intercept the window function to handle the WM_PAINT message of the toolbar UI control, switching theming off and rendering items we're unhappy with with the default renderer instead of a bad-looking one.

image

https://github.com/user-attachments/assets/6612fad3-e850-4f68-bc4f-602425bb4b1e


plugins/org.jkiss.dbeaver.ui.swt.windows/fragment.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
serge-rider, line 1: Add it to build.xml (now it probably works only in dev mode)

serge-rider (changes requested):

serge-rider:
Doesn't work on 4k monitor
image

plugins/org.jkiss.dbeaver.ui.swt.windows/src/org/jkiss/dbeaver/ui/swt/windows/ToolBarRenderFix.java
```diff
@@ -0,0 +1,162 @@
...
+    private void addDarkThemeToolbarsFix() {
+        int[] eventsToHandle = new int[]{ SWT.Paint, SWT.Resize };
+
+        Listener listener = event -> {
+            if (event.widget instanceof ToolBar t && t.getData(DBEAVER_TOOLBAR_SUBCLASS_HANDLER_PROP_NAME) == null) {
+                t.setData(DBEAVER_TOOLBAR_SUBCLASS_HANDLER_PROP_NAME, new ToolbarSubclassHandler(t));
+            }
+        };
+
+        Display display = PlatformUI.createDisplay();
```
serge-rider: PlatformUI.createDisplay() looks odd. There already may be a display.
Plugin activation order is not predetermined.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.swt.windows/src/org/jkiss/dbeaver/ui/swt/windows/ToolBarRenderFix.java
serge-rider: Also new error appears in Eclipse log
```
!ENTRY org.eclipse.ui.workbench 2 0 2026-05-06 06:43:02.785
!MESSAGE swt.autoScale.updateOnRuntime is configured (e.g., via the INI), but the according preference should be preferred instead.
```

Matvey16:
Verified

DB UE, CE Windows 11 1080p, 2k scaling 150, 175, 200
DB CE Linux

PR #41004, opened 2026-05-05
Implement SSH tunnel handler with settings

saiganesh47:
Added SSH tunnel handler with various configuration properties.


ShadelessFox:
I'm not sure what it's supposed to be.

PR #41015, opened 2026-05-06
dbeaver/pro#9248 Do not close data transfer wizard on error

serge-rider:


Matvey16:
Verified

DBEE Windows

PR #41018, opened 2026-05-06
#41021 Group multi-object delete commands into a single undo batch

### LINKED ISSUE #41021: Multi-object delete in navigator requires N undos instead of one
### Description

When the user selects several database objects in the navigator (for example several columns of the same table) and presses Delete in one action, each object becomes its own undo entry. To restore the whole selection, the user must press Ctrl+Z once per deleted object.

The user issued the deletion as a single action, so the undo should also be a single action.

### DBeaver Version

26.0.5 (devel)

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

1. Connect to any database that supports column deletion (e.g. SQLite).
2. Create a table with multiple columns:
   ```sql
   CREATE TABLE foo (a INT, b INT, c INT, d INT, e INT, f INT, g INT, h INT, i INT, j INT);


### Additional context

_No response_

pandadev66:
## Summary

When the user selects several objects in the database navigator and presses Delete (for example 10 columns at once on a table), each object delete becomes its own undo entry. As a result, restoring all of them requires pressing Ctrl+Z N times instead of a single Ctrl+Z, even though the user issued the deletion as a single action.

This change groups every command queued during one navigator multi-delete into a single undo batch, so one Ctrl+Z restores the whole selection at once. The framework already supports batched undo through `CommandInfo.linkedCommand`, but `NavigatorObjectsDeleter` was not threading the link between successive deletes.

## What changed

- `org.jkiss.dbeaver.model/.../DBECommandContext.java`: two new default-no-op interface methods, `getCommandCount()` and `linkLastCommands(int count)`. Default implementations keep all existing implementers (e.g. `SimpleCommandContext`, `TestCommandContext`) compiling without changes.
- `org.jkiss.dbeaver.model/.../impl/edit/AbstractCommandContext.java`: real implementations. `getCommandCount` returns the size of the internal command list. `linkLastCommands(N)` chains `commands[size-N..size-1]` together via `linkedCommand`, which is what the existing `undoCommand()` walks during a batch undo.
- `org.jkiss.dbeaver.ui.navigator/.../actions/NavigatorObjectsDeleter.java`: track per-context starting count in an `IdentityHashMap`, snapshot it the first time a context is touched, then after each `objectMaker.deleteObject(...)` call `linkLastCommands(diff)` on that context. The chain grows monotonically, idempotent re-applies of earlier links are harmless.

## Test plan

1. Open or create any database connection with editable tables (SQLite is fine, no server required).
2. In a SQL editor: `CREATE TABLE foo (a INT, b INT, c INT, d INT, e INT, f INT, g INT, h INT, i INT, j INT);` and refresh the navigator.
3. Expand `foo` -> Columns. Multi-select 10 columns (Ctrl+click).
4. Right-click -> Delete -> confirm.
5. The columns are now marked for deletion. Press Ctrl+Z.

   Before this PR: only one column is restored. You must press Ctrl+Z 10 times to restore everything.

   After this PR: a single Ctrl+Z restores all 10 columns.

6. Press Ctrl+Y once after step 5 -> all 10 deletions are re-applied as a batch (existing redo logic also walks the `linkedCommand` chain).
7. Sanity check on a single delete: delete one column on its own, Ctrl+Z restores it. No regression.
8. Cross-context safety: select objects from two different connections in the navigator, multi-delete them. Each context is linked independently via the per-context `IdentityHashMap`; undo on one connection's editor is independent.

## Notes

- No `Closes #...` link yet -- I'll report this as a fresh issue and link it once the PR title is right. Happy to amend the PR description with the issue number once filed.
- `DBECommandContext` is implemented by a small number of classes. Default no-op implementations on the new methods mean only the `AbstractCommandContext` path actually batches; alternate implementations stay unchanged.
- The bug is observable in any database (PostgreSQL, MySQL, etc.) since the fix is at the navigator/framework layer, not in any DB-specific manager.

Closes #41021

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/edit/DBECommandContext.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/edit/DBECommandContext.java
@@ -69,6 +69,13 @@ public interface DBECommandContext extends DBPContextProvider {
 
     //void addCommandBatch(List<DBECommand> commands, DBECommandReflector reflector, boolean execute);
 
+    default int getCommandCount() {
+        return 0;
+    }
+
+    default void linkLastCommands(int count) {
+    }
+
     void removeCommand(@NotNull DBECommand<?> command);
 
     void updateCommand(@NotNull DBECommand<?> command, @Nullable DBECommandReflector commandReflector);
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/edit/AbstractCommandContext.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/edit/AbstractCommandContext.java
@@ -333,6 +333,29 @@ public abstract class AbstractCommandContext implements DBECommandContext {
         }
     }
 
+    @Override
+    public int getCommandCount() {
+        synchronized (commands) {
+            return commands.size();
+        }
+    }
+
+    @Override
+    public void linkLastCommands(int count) {
+        if (count < 2) {
+            return;
+        }
+        synchronized (commands) {
+            int size = commands.size();
+            if (size < count) {
+                return;
+            }
+            for (int i = size - 1; i > size - count; i--) {
+                commands.get(i).linkedCommand = commands.get(i - 1);
+            }
+        }
+    }
+
     @NotNull
     @Override
     public Collection<? extends DBECommand<?>> getUndoCommands()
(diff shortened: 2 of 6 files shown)


pandadev66:
Hi @E1izabeth @ShadelessFox 
How about you? This is my first PR and hope you review it. ❤️ 
Please find out attached recording.

https://github.com/user-attachments/assets/ee07db0c-4ce3-458a-ad9f-dd9054734260


Thanks.

E1izabeth:
Thanks for the contribution!
Please add the issue number to the beginning of the commit message.

pandadev66:
Hi @E1izabeth 
Done!  Commit message updated. Thanks for the review!

pandadev66:
Hi @E1izabeth 
When you have a moment, please take a look at this PR and let me know your recommendation.
Thanks.

E1izabeth:
@pandadev66 thanks for the contribution! We will review your PR and let you know

ShadelessFox (commented):
I don't understand how it works, and the lack of comments/docs doesn't improve the situation...

E1izabeth:
It looks like this PR is stale, so I'm closing it. Please, let me know if you continue working on this - we can reopen it.

PR #41028, opened 2026-05-07
#35904 Show full preview for small arrays in result grid

### LINKED ISSUE #35904: Show full values preview for small arays
How can I restore the representation to the way it was in version 24.1.0? I need to always display all values in an array, at least when the count is less than 10 or so. It should look like this:

_Originally posted by @btanner2 in https://github.com/dbeaver/dbeaver/issues/35403#issuecomment-2396127703_

omipheo:
Closes #35904.

```markdown
## Summary

Restores the pre-24.1.1 result-grid behavior of expanding small arrays
inline (e.g. `ARRAY[1,2,3,4]` → `{1,2,3,4}`) instead of collapsing them
to a `[+N]` size hint. The cap is configurable and can be disabled to
keep the current always-collapse behavior.

## Fix
- New preference `resultset.inline.collection.element.limit` (default
  `10`) declared in `ModelPreferences`. The constant is registered with
  a default value of 10 alongside the other result-set preferences.
- `ArrayCellHintProvider` now reads the limit from the data-source
  preference store (falling back to the global store) and skips emitting
  the size hint when `collection.size() <= limit`. With no hint the
  cell falls back to the value handler's full string, which matches the
  24.1.0 rendering shown in the issue screenshots.
- Setting the preference to `0` disables the feature and reproduces the
  current always-collapse behavior, so users who prefer the compact
  presentation lose nothing.
- Wired into the existing `Editors → Data Editor → Result Sets →
  Presentation` page next to *Visualize complex columns*, with load /
  save / clear / restore-defaults handling and an entry in
  `hasDataSourceSpecificOptions` so the override badge appears at the
  data-source scope.

## Reproduction
1. Run `SELECT ARRAY[1,2,3,4] AS a, ARRAY[1,2,3,4] AS b` against any
   datasource that surfaces native arrays (e.g. PostgreSQL).
2. Before: each cell renders as `1[+3]`. After: each cell renders as
   `{1,2,3,4}`.
3. Open *Window → Preferences → Editors → Data Editor → Result Sets →
   Presentation* and lower *Inline array preview limit* to `0` to fall
   back to the previous `[+N]` collapsed view.

## Files touched
- `plugins/org.jkiss.dbeaver.model/.../ModelPreferences.java`
- `plugins/org.jkiss.dbeaver.model/.../data/hints/standard/ArrayCellHintProvider.java`
- `plugins/org.jkiss.dbeaver.ui.editors.data/.../preferences/PrefPageResultSetPresentation.java`
- `plugins/org.jkiss.dbeaver.ui.editors.data/.../internal/DataEditorsMessages.java`
- `plugins/org.jkiss.dbeaver.ui.editors.data/.../internal/DataEditorsResources.properties`

## Closes
Closes #35904
```


E1izabeth:
Thanks for the contribution! We will review your PR and let you know.

IvanGorshechnikov:
Generally, I don't like this fix
We could have an array with long strings. And we shouldn't check the size of the array, we should check how long the previews of objects are.

omipheo:
@IvanGorshechnikov thanks for the review — that's a fair point. The current element-count gate mirrors the explicit suggestion in the issue body (*"at least when the count is less than 10 or so"*), and I picked it for the small-diff/behaviour-restoring scope, but you're right that a short array of long strings would still blow up the cell.

Three directions I'm happy to ship as a follow-up if @E1izabeth agrees on which fits best:

1. **Replace** the element-count gate with a max-rendered-length gate (suppress the size hint only when the combined preview string is under a configurable budget).
2. **Combine** both: keep `resultset.inline.collection.element.limit` and add a sibling `resultset.inline.collection.max.length`; suppress the hint only when *both* are satisfied.
3. **Keep as-is** if the element-count knob is the simpler/desired user surface.

@E1izabeth let me know which direction you'd prefer and I'll update the implementation.

E1izabeth:
Hi, @omipheo! Could you resolve the conflict?

E1izabeth:
This PR is stale, so I'm closing it. Please, let me know if you continue working on this - we can reopen it.

PR #41036, opened 2026-05-07
dbeaver/dbeaver#40898 Fix redshift search path

### LINKED ISSUE #40898: AWS RedShift: Can't change default schema

IvanGorshechnikov:
Closes https://github.com/dbeaver/dbeaver/issues/40898


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
```diff
@@ -223,28 +224,35 @@ public boolean refreshDefaults(DBRProgressMonitor monitor, boolean useBootstrapS
...
         List<String> newSearchPath = new ArrayList<>(searchPath);
         int schemaIndex = newSearchPath.indexOf(defSchemaName);
         {
             if (schemaIndex < 0) {
                 // Remove from previous position
+                if (getDataSource().getServerType() instanceof PostgreServerRedshift
+                    && getDefaultCatalog().getSchema(monitor, PostgreConstants.PUBLIC_SCHEMA_NAME) == null
+                    && !PostgreConstants.PUBLIC_SCHEMA_NAME.equals(defSchemaName)) {
+                    newSearchPath.removeIf(PostgreConstants.PUBLIC_SCHEMA_NAME::equals);
+                }
```
E1izabeth: Please add a comment here to explain why it's so only for Redshift

E1izabeth:
And please close another PR for this issue, then.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
```diff
@@ -223,28 +224,35 @@ public boolean refreshDefaults(DBRProgressMonitor monitor, boolean useBootstrapS
...
     public List<String> getSearchPath() {
         return searchPath;
     }
 
-    private void addSearchPath(String path) {
+    private void addSearchPath(@NotNull String path) {
         searchPath.clear();
         searchPath.add(path);
-        if (!path.contains(activeUser)) {
+        if (!path.contains(activeUser) && !(getDataSource().getServerType() instanceof PostgreServerRedshift)) {
```
serge-rider: Don't do instanceof PostgreServerRedshift
Add new function to PostgreServerExtension

serge-rider (changes requested):

uslss:
verified

PR #41040, opened 2026-05-08
#19247 Defer datasource connect in metadata search until tree expansion

### LINKED ISSUE #19247: Improve metadata search behaviour
When focusing on a connection in a metadata search, DBeaver immediately tries to connect to it in order to display the types of objects from this database to the right of the connections tree
image

But sometimes it can be annoying. For example, if you accidentally click or move through the connections using the keyboard, you will be prompted to download drivers, although you don't need them at all at the moment.
Maybe you should not show the window with object types until the user expands the tree. In this case, we can only request a driver download after the tree expands

MilosM348:
## Summary

`SearchMetadataPage` triggers `DBNDataSource.initializeNode(...)` from
the tree's `addSelectionChangedListener`. Because that fires on *any*
selection change — including arrow-key navigation through the connection
list and stray clicks — users get prompted with the driver-download
dialog for connections they never intended to open.

## Fix

Move the `initializeNode(...)` call from the selection listener to a new
`ITreeViewerListener.treeExpanded` handler. The driver download / connect
now runs only when the user explicitly expands a connection node, which
matches the maintainer's suggested behaviour on #19247:

> "Maybe you should not show the window with object types until the user
> expands the tree. In this case, we can only request a driver download
> after the tree expands."

The selection listener still calls `fillObjectTypes()` + `updateEnablement()`
unconditionally, so already-connected data sources continue to populate
the object-types panel the moment the user clicks them — only the
**not-yet-initialized** code path is deferred to expansion. The
`treeExpanded` handler short-circuits when the data source already has a
live `getDataSource()` so no redundant `initializeNode` call is made when
the user expands an already-open connection.

The `private static final boolean CONNECT_ON_CLICK = true` constant is
removed because it was the guard for the deleted code path and had no
other references.

## Reproduction

1. Open Search → Database objects search.
2. Use the up/down arrow keys to navigate through the connection list,
   or click any connection whose driver isn't already downloaded.
3. **Before the fix:** the driver-download dialog appears the moment the
   selection lands on that connection, even if the user is just browsing.
4. **After the fix:** selecting / navigating leaves the connection alone.
   Expanding the connection node (clicking the chevron, double-click, or
   pressing Right arrow) is what triggers the connect, matching the
   pattern the user expects from a lazy tree.

## Closes

Closes #19247.


arhayka:
@MilosM348 Thank you for the contribution!

IvanGorshechnikov (approved):

IvanGorshechnikov:

MashaKorax:
@MilosM348 Please fix the behavior: object types appear only after the second mouse click on the database.

https://github.com/user-attachments/assets/9bdf4967-ad09-4b16-a210-76a5afb15c2e

E1izabeth:
The bug found with the provided solution has not yet been resolved. So, it looks like this PR is stale, and I'm closing it. Please, let me know if you continue working on this - we can reopen it.

PR #41042, opened 2026-05-08
#40779 Fix Oracle ORA-00911 on CREATE VIEW with trailing CASE...END

### LINKED ISSUE #40779: When creating a view in Oracle 11g, it prompts ORA-00911: invalid character.
### Description

```
CREATE OR REPLACE view VM_TEST as(
	SELECT C,Q
	FROM (         
		SELECT 1 AS C,0 AS Q FROM dual       
	) A WHERE CASE WHEN C = 1 THEN Q ELSE 1 END <> 0
);
```

Executing the above script will prompt ORA-00911: invalid character.

### DBeaver Version

Community 26.0.2.202604051735

### Operating System

windows

### Database and driver

com.oracle.database.jdbc:ojdbc 11:RELEASE [23.2.0.0]

### Steps to reproduce

_No response_

### Additional context

_No response_

a3894281:
## Summary

Fixes #40779 — Oracle returns `ORA-00911: invalid character` when running `CREATE OR REPLACE VIEW` whose body ends with a `CASE ... END` expression.

## Root cause

`SQLScriptParser.parseQueryImpl` tokenizes the bare `END` of a `CASE` expression as `T_BLOCK_END`, identical to the `END` that closes a real PL/SQL block. The token's text overwrites `lastKeyword` to `"END"`. When the trailing `;` is reached, `needsDelimiterAfterBlock("CREATE", "END", oracle)` returns `true` because:

- Oracle dialect has `isDelimiterAfterBlock() == true`
- `lastKeyword == "END"`
- `firstKeyword == "CREATE"` — present in `getDDLKeywords()`

The delimiter is therefore re-appended to the extracted query text. Oracle JDBC rejects the trailing `;` with `ORA-00911`.

## Fix

Track the opening token of every block in `ScriptBlockInfo.beginToken`. When a `T_BLOCK_END` pops a block whose `beginToken` is `CASE`, set a local flag and skip the `lastKeyword = "END"` overwrite. The block-end delimiter retention path then only fires for genuine PL/SQL blocks (`BEGIN`, `LOOP`, `IF`, ...), which is the intended behaviour.

`CASE` statement form (`END CASE`) inside `BEGIN..END` is unaffected — the outer `BEGIN..END` still terminates with a normal `END` token and triggers delimiter retention as before.

## Reproduction

```sql
CREATE OR REPLACE VIEW VM_TEST AS (
  SELECT C, Q
  FROM (SELECT 1 AS C, 0 AS Q FROM dual) A
  WHERE CASE WHEN C = 1 THEN Q ELSE 1 END <> 0
);

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLScriptParser.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLScriptParser.java
@@ -133,6 +133,7 @@ public class SQLScriptParser {
         int lastTokenLineFeeds = 0;
         SQLTokenType prevNotEmptyTokenType = SQLTokenType.T_UNKNOWN;
         String firstKeyword = null, lastKeyword = null;
+        boolean endClosedCaseBlock = false;
         for (; ; ) {
             TPToken token = ruleScanner.nextToken();
             int tokenOffset = ruleScanner.getTokenOffset();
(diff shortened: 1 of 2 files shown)


arhayka:
@a3894281 Thank you for the contribution!

DariyaFedorova:
Verified

PR #41044, opened 2026-05-08
dbeaver/pro#8881 Refix zero datetimes formatting to not operate with strings

E1izabeth:
The bug was caused by https://github.com/dbeaver/dbeaver/issues/39743


Matvey16:
Verified

DB CE Windows

PR #41047, opened 2026-05-08
dbeaver/dbeaver#40759 Fix UI freeze on multi-object delete

### LINKED ISSUE #40759: Deleting multiple table columns at once causes inaccessible GUI (MariaDB)
### Description

1. Create a table
2. Create some columns (about 5  or more)
3. Select the columns and select delete

It seems the a confirmation message is hidden or not shown , do the UI is stuck on the dialog.
Go to logout from Ubuntu, the a message appears to confirm to commit the changes. This works.
Deleting 2 columns or 1 , this issue does not occur.
MariaDB db

### DBeaver Version

26.0.1.202603221810

### Operating System

Ubuntu 25.10

### Database and driver

mySQL => Actually it is  MARIADB

### Steps to reproduce

As above

### Additional context

_No response_

a3894281:
Closes #40759

## Summary

When deleting multiple objects (e.g. 5+ table columns at once on MariaDB), the UI froze because N "Change is not persisted" warning dialogs were spawned — one per selected object — and the underlying error from MariaDB rejecting the drop-all-columns statement ended up hidden behind them.

Reproduce video and full bug report are in the linked issue: #40759

## Root cause

In `NavigatorObjectsDeleter.deleteDatabaseNode()`, the guard flag `deleteWarningShowed` was set **after** `showMessageBox()` returned. SWT modal dialogs keep pumping events while open, so each pending `asyncExec` delete saw the flag still unset and opened its own dialog — stacking N dialogs and freezing the UI.

## Fix

Move the `deleteWarningShowed = true` assignment to **before** the dialog call so the next queued delete sees the flag already set and skips the warning. Only one dialog now appears regardless of selection size.

## Test plan

- [x] Create a table with 5+ columns on MariaDB, persist
- [x] Reopen, select all columns, Delete → only one warning dialog, UI responsive
- [x] Verify single-object delete still shows the warning once

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorObjectsDeleter.java
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/navigator/actions/NavigatorObjectsDeleter.java
@@ -275,13 +275,13 @@ public class NavigatorObjectsDeleter {
                 tasksToExecute.add(deleter);
             }
             if (commandTarget.getEditor() != null && selectedFromNavigator && !deleteWarningShowed) {
+                deleteWarningShowed = true;
                 UIUtils.getActiveWorkbenchWindow().getActivePage().activate(commandTarget.getEditor());
                 DBWorkbench.getPlatformUI().showMessageBox(
                     UINavigatorMessages.actions_navigator_persist_delete_in_the_editor_title,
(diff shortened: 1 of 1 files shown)


arhayka:
Hi @a3894281, thank you for the contribution! I'll pass your PR to the development team for review and testing.

E1izabeth (approved):

ShadelessFox (approved):
Thanks, passing to QA

PR #41050, opened 2026-05-09
#38547 Add result filter panel visibility toggle

### LINKED ISSUE #38547: Option To Hide Filter Row On Result Panel
This would be nice for those who have never used it.

Image

EastLord:
Closes: #38547

https://github.com/user-attachments/assets/fb4de42c-17d3-4dcd-a32c-a0c507da9360


arhayka:
Hi @EastLord, thank you again for the contribution! I'll pass this PR to the development team for review and testing.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/editors/data/preferences/PrefPageResultSetPresentation.java
```diff
@@ -88,6 +91,11 @@ protected Control createPreferenceContent(@NotNull Composite parent) {
             showFiltersInSingleTabMode = UIUtils.createCheckbox(uiGroup,
                 DataEditorsMessages.pref_page_database_resultsets_label_filters_panel_in_singletab_mode,
                 null, true, 2);
+            if (!isDataSourcePreferencePage()) {
```
IvanGorshechnikov: Why did you add it? It could be a datasource-specific property

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -1946,6 +1944,22 @@ private void createFilterPanel0(@NotNull Composite parent) {
...
+        UIUtils.setControlVisible(
+            filtersPanelComposite,
+            !statisticsPresentation && getFilterPanelPreferenceStore().getBoolean(ResultSetPreferences.RESULT_SET_SHOW_FILTER_PANEL)
+        );
+        mainPanel.layout(true, true);
+    }
+
+    @NotNull
+    private static DBPPreferenceStore getFilterPanelPreferenceStore() {
+        return DBWorkbench.getPlatform().getPreferenceStore();
```
IvanGorshechnikov: So here you should use preference store from datasource

IvanGorshechnikov (changes requested):

EastLord:
@IvanGorshechnikov Thanks for the review.

My initial implementation treated this as a global UI preference because the filter panel visibility is a result-set presentation option, and I expected the same visibility state to be shared across result tabs. That is why I originally kept the Preferences entry global-only and read it from the platform preference store.

After your comment, I agree that this should follow the existing Result Set preference model and support datasource-specific overrides. I have updated the code accordingly.

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -1946,6 +1944,17 @@ private void createFilterPanel0(@NotNull Composite parent) {
         );
     }
 
+    private void updateFilterPanelVisibility(boolean statisticsPresentation) {
```
ShadelessFox: Maybe swap this `boolean statisticsPresentation` with an actual check `getActivePresentation() instanceof StatisticsPresentation`? It's currently duplicated in several places.

EastLord:
@ShadelessFox Thanks for the code review. I agree that this check should be centralized inside updateFilterPanelVisibility(). I’ve updated the method to derive the state from the active presentation directly and adjusted the call order so the visibility is recalculated after the active presentation is updated.

ShadelessFox (approved):

E1izabeth (approved):

E1izabeth:
Passed to QA

MashaKorax:
Verified DBeaver CE
win
mac
linux

PR #41052, opened 2026-05-09
Dbeaver/pro#9018 fix background of toolbar items on dark theme

serge-rider:
How to test:
open lots of (15-20) data editors, property editors for different connections of different connection types (with different backgrounds)


Matvey16:
Verified

DB CE Windows

PR #41057, opened 2026-05-11
#40757 Fix handling of 24:00 in PostgreSQL 

### LINKED ISSUE #40757: postgres time column with value 2400 incorrectly shows as 0000

E1izabeth:


uslss:
verified (win11, CE)
image

PR #41058, opened 2026-05-11
dbeaver/pro#9269 Expandable composite styling and persistence

ShadelessFox:


Matvey16:
Verified

DB CE, EE Windows

PR #41060, opened 2026-05-11
dbeaver/pro#8916 Fix save password check during test connection

IvanGorshechnikov:
Closes https://github.com/dbeaver/pro/issues/8916


Matvey16:
Verified

DB CE, UE Windows

PR #41063, opened 2026-05-11
dbeaver/pro#9203 Fix oracle legacy indexes and privileges

labanovichttps:
Check Oracle privileges and indexes
image


plugins/org.jkiss.dbeaver.ext.oracle/src/org/jkiss/dbeaver/ext/oracle/model/OracleSchema.java
```diff
@@ -1201,19 +1201,41 @@ class IndexCache extends JDBCCompositeCache<OracleSchema, OracleTableBase, Oracl
 
         @NotNull
         @Override
-        protected JDBCStatement prepareObjectsStatement(@NotNull JDBCSession session, @NotNull OracleSchema owner, @Nullable OracleTableBase forTable)
-            throws SQLException
-        {
+        protected JDBCStatement prepareObjectsStatement(
```
serge-rider, line 1204: 1. Add comments regarding index DDL.
2. Use legacy join syntax for all versions

serge-rider (changes requested):

PR #41064, opened 2026-05-11
dbeaver/pro#9098 Redesign ODBC connection page

ShadelessFox:


plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
```diff
@@ -1020,6 +1023,9 @@ public static String getDatabaseNameFromConfiguration(@NotNull DBPConnectionConf
                 activeDatabaseName = matcher.group(1).replace("%2F", "/");
             }
         }
+        if (CommonUtils.isEmpty(activeDatabaseName)) {
```
serge-rider: No sure it is a good place. We probably should check for default name in the end of initializeRemoteInstance. Because connect behavior depends on database name. We mustn't change behavior for empty db name. Only provide it as a fallback for final URL.

E1izabeth (approved):

PR #41065, opened 2026-05-11
dbeaver/pro#8683 Add error state to the function results

IvanGorshechnikov:
Closes https://github.com/dbeaver/pro/issues/8683


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/impl/AIAssistantImpl.java
```diff
@@ -236,7 +236,19 @@ protected AIFunctionResult callFunction(
                 AIBaseFeatures.PROMPT_TYPE, context.getPrompt().generatorId()
             )
         ));
-        return function.getToolbox().callFunction(context, function, arguments);
+        AIFunctionResult result;
+        try {
+            result = function.getToolbox().callFunction(context, function, arguments);
+        } catch (DBException e) {
+            result = new AIFunctionResult(
+                AIFunctionType.ACTION,
```
serge-rider: But this code works for INFO function as well?

PR #41074, opened 2026-05-13
dbeaver/pro#9266 Add env variabled for DBeaverData and workspace6

serge-rider:


alexander-skoblikov (dismissed):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/registry/EclipsePluginApplicationImpl.java
```diff
@@ -63,9 +63,9 @@ public String getDefaultProjectName() {
         return "DBeaver";
     }
 
-    @Nullable
+    @NotNull
     @Override
-    public Path getDefaultWorkingFolder() {
+    public Path getWorkspacePath() {
         return null;
```
alexander-skoblikov: NotNull but returns null

plugins/org.jkiss.dbeaver.headless/src/org/jkiss/dbeaver/headless/DBeaverHeadlessApplication.java
```diff
@@ -65,8 +65,9 @@ public void stop() {
         super.stop();
     }
 
+    @NotNull
     @Override
-    public @Nullable Path getDefaultWorkingFolder() {
+    public Path getWorkspacePath() {
         return null;
```
alexander-skoblikov: same

PR #41080, opened 2026-05-14
dbeaver/dbeaver#41034 List parameters per overload for Informix procedures

### LINKED ISSUE #41034: Overloaded Stored Procedures in Informix 12.10 Don't List the Parameters Correctly
### Description

I am using several versions DBeaver, and it is a great product.  On Informix 12.10, you can have overloaded stored procedures.  The procedure parameters tab on the properties tab don't list all the IN parameters.  It seems to list the last set of IN parameters.  The code below are examples.

```
drop procedure if exists findtableswithvalue(lvarchar, lvarchar);
create procedure findtableswithvalue(
	columnName lvarchar,
	queryValue lvarchar
)
  -- SQL commands
end procedure;

drop procedure if exists findtableswithvalue(lvarchar, integer);
create procedure findtableswithvalue(
	columnName lvarchar,
	queryValue integer
)
  -- SQL commands
end procedure;
``` 

### DBeaver Version

26.0.2.202604051735

### Operating System

Windows Server 2019 Standard

### Database and driver

_No response_

### Steps to reproduce

1. Use the example code.
```
drop procedure if exists findtableswithvalue(lvarchar, lvarchar);
create procedure findtableswithvalue(
	columnName lvarchar,
	queryValue lvarchar
)
  -- SQL commands
end procedure;

drop procedure if exists findtableswithvalue(lvarchar, integer);
create procedure findtableswithvalue(
	columnName lvarchar,
	queryValue integer
)
  -- SQL commands
end procedure;
``` 
2. open up the procedure > properties > procedure parameters

### Additional context

_No response_

a3894281:
Fixes #41034. (Added reproduce results on [issue](https://github.com/dbeaver/dbeaver/issues/41034) ticket.

## Problem
Informix supports overloaded stored procedures (same name, different parameter signatures), but DBeaver only ever displayed one node per name and merged the parameters of every overload into it. The procedure source pane had the same issue, returning the concatenated source of all overloads.

Root cause:
- The Informix JDBC driver collapses overloaded procedures in `DatabaseMetaData.getProcedures()` (one row per name, `SPECIFIC_NAME` is null).
- `getProcedureColumns()` does return rows for every overload, but it only distinguishes them in the `REMARKS` text (e.g. `create procedure null.findtableswithvalue(lvarchar,integer)`).
- `GenericProcedure.loadProcedureColumns` advances through the local procedures list whenever `ORDINAL_POSITION` resets; with only one local procedure that branch never fires and all parameters end up on the same object.

## Fix
- Add `InformixProcedure` (subclass of `GenericProcedure`) holding `procid` and `paramTypes` from `sysprocedures`. It overrides `loadProcedureColumns` to call `getProcedureColumns()` and only keep rows whose REMARKS signature matches this overload's `paramTypes`.
- Override `InformixMetaModel.loadProcedures` to query `informix.sysprocedures` directly (filtered by `internal = 'f'` and the owning schema) so each `procid` becomes its own procedure node. `isproc` selects PROCEDURE vs FUNCTION.
- Update `InformixUtils.getProcedureSource` to filter `sysprocbody` by `procid` when the procedure is an `InformixProcedure`, so DDL is returned per overload instead of concatenated.

## Reproduction
```sql
create procedure findtableswithvalue(columnName lvarchar, queryValue lvarchar)
end procedure;
create procedure findtableswithvalue(columnName lvarchar, queryValue integer)
end procedure;
```
Before: tree shows a single `findtableswithvalue` node; the Procedure Parameters tab shows 4 mashed rows from both overloads.
After: tree shows two nodes (one per signature); each node lists its own two parameters and its own source.

Verified against Informix 14.10 (developer edition Docker image).


E1izabeth:
Thanks! We will review your PR and let you know

plugins/org.jkiss.dbeaver.ext.informix/src/org/jkiss/dbeaver/ext/informix/InformixUtils.java
```diff
@@ -75,11 +76,20 @@ public static String listToString(List<String> value, String delimiter) {
     }
 
     public static String getProcedureSource(DBRProgressMonitor monitor, GenericProcedure procedure) throws DBException {
-        String sqlProcedure = String.format("select b.data "
-            + "from sysprocbody b "
-            + "join sysprocedures p on b.procid=p.procid "
-            + "where datakey='T' and p.procname = '%s'"
-            + "order by b.procid, b.seqno", procedure.getName());
+        String sqlProcedure;
+        if (procedure instanceof InformixProcedure ip) {
```
E1izabeth: I believe it can't be not InformixProcedire if we're in InformixUtils, isn't it?
If so, maybe we don't need else branch

a3894281: Good catch. Only caller is `InformixMetaModel.getProcedureDDL`, so procedure is always `InformixProcedure`. Dropped else branch and changed param type to `InformixProcedure` in ef5635120e.

plugins/org.jkiss.dbeaver.ext.informix/src/org/jkiss/dbeaver/ext/informix/model/InformixProcedure.java
```diff
@@ -0,0 +1,149 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2025 DBeaver Corp and others
```
E1izabeth:
```suggestion
 * Copyright (C) 2010-2026 DBeaver Corp and others
```

E1izabeth:
Hi, @a3894281! Please update the copyright year, as it's 2026 already =)

plugins/org.jkiss.dbeaver.ext.informix/src/org/jkiss/dbeaver/ext/informix/model/InformixProcedure.java
a3894281: Done in 3a8fbe06fc. Bumped the copyright year to 2026 on all three touched files (InformixProcedure, InformixMetaModel, InformixUtils).

a3894281:
Hi @E1izabeth! Updated the copyright year to 2026 on all three touched files in 3a8fbe06fc. Thanks!

E1izabeth (dismissed):

ShadelessFox (dismissed):
Thanks! Passing to QA

a3894281:
Thanks.

uslss:
Procedures aren't visible in DB Navigator (refresh, reconnect do nothing)
image
image

a3894281:
Thanks for catching that, @uslss! Fixed in the latest commit.

The previous `loadProcedures` queried `sysprocedures` directly, which (1) listed callable built-in routines (`internal='f'` means *callable from SQL*, not *user-defined*) and (2) misread the BOOLEAN `isproc` column so every routine was treated as a function — leaving the Procedures node empty.

The list now comes from the JDBC driver's `getProcedures()`/`getFunctions()` (built-ins stay hidden, schema scoping preserved), and `sysprocedures` is used only to split overloaded routines into one node per `procid`. Overloaded-parameter listing from #41034 still works.

E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.informix/src/org/jkiss/dbeaver/ext/informix/model/InformixMetaModel.java
```diff
@@ -179,4 +183,140 @@ public boolean isTrimObjectNames() {
...
+                        : DBSProcedureType.FUNCTION;
+                    overloads.computeIfAbsent(procname, k -> new ArrayList<>())
+                        .add(new InformixProcedureMeta(procid, trimmedParams, procedureType));
+                }
+            }
+        }
+        return overloads;
+    }
+
+    private static class InformixProcedureMeta {
```
ShadelessFox, line 311: Could be a `record`

ShadelessFox (changes requested):
Most methods are missing nullability annotations; please add them. See https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code#nullability

E1izabeth:
@a3894281 would you address the review comments?

PR #41081, opened 2026-05-14
dbeaver/pro#9108 Replace checkbox Show connected only to navigator tree button about that

E1izabeth:
What changed - visual state management of the "Show connected only" toolbar item.
What intentionally didn't change - filter application logic. Where folders were expanded, they're expanded. Where they were not expanded, they're not.

### FOR QA
Dialogs to check:
- [ ] Select data source dialog: switch current script database in the main toolbar by clicking the schema name button itself (the one having the attached drop-down menu)
image

- [ ] Metadata Search
image

- [ ] DB full text search
image

- [ ] Create new task wizard, "Common" - "Data export", "Add Table..." or "Add Query..."
image

- [ ] In any tool task (for example, "Oracle" - "Tools" - "Gather table statistics")
image


IvanGorshechnikov (approved):

ShadelessFox (approved):

Matvey16:
Verified

DBeaver UE Windows

PR #41084, opened 2026-05-15
Expose DDL transaction support for generic drivers

### LINKED ISSUE #40889: Expose supports-ddl-transactions driver parameter to UI
I am working with database (don't have separate plugin for DBeaver, therefore I've created generic entry for driver) which don't support DDL in transactions.
However when I try to do any DDL-like manipulation through UI it gives driver error that following operation should not be called with explicit transaction. Example of such operation - deleting table from object tree list (right click on table -> delete). 
I've learned that if edit DBeaver configuration file `%APPDATA%\DBeaverData\workspace6\.metadata\.config\driver.xml` by adding `<parameter name="supports-ddl-transactions" value="false"/>` to my driver parameters that fixes issue I have with manipulating objects via UI.

I'd like to have means to configure mentioned driver parameter through already existing UI:

Image

popsiclelmlm:
## Summary
- add `supports-ddl-transactions` to the generic driver DDL parameter list
- allow custom generic drivers to configure DDL transaction support from the existing driver Parameters tab

Closes #40889

## Tests
- `git diff --check`


ShadelessFox:
Please change commit messages according to https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code#commit-message-rules

popsiclelmlm:
Updated the commit message to follow the DBeaver convention: #40889 Expose DDL transaction driver option. I also included the Assisted by trailer per the contribution guide.

E1izabeth:
Thanks for the contribution!

ShadelessFox (approved):
Thanks

PR #41087, opened 2026-05-15
dbeaver/pro#9316 add long option name for quiet mode

alexander-skoblikov:
Closes dbeaver/pro#9316


Matvey16:
Verified

DBeaver CE Windows

PR #41090, opened 2026-05-15
dbeaver/pro#9310 Fix tasks layout

serge-rider:


Matvey16:
Verified

DB CE, EE Windows

PR #41092, opened 2026-05-15
9316 cli  help command doesnt work global fix

alexander-skoblikov:
closes https://github.com/dbeaver/pro/issues/9316


Matvey16:
Verified

DBeaver CE Windows, Linux

PR #41102, opened 2026-05-18
7408 core te tasks and metadata editor use a first team creds to connect to db

alexander-skoblikov:
closes https://github.com/dbeaver/pro/issues/7408


plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/DatabaseLazyEditorInput.java
```diff
@@ -283,6 +284,17 @@ public IDatabaseEditorInput initializeRealInput(@NotNull DBRProgressMonitor moni
...
+            boolean cancelled = false;
+            if (dataSourceContainer.isSharedCredentials()) {
+                UIServiceConnections serviceConnections = DBWorkbench.getService(UIServiceConnections.class);
+                if (serviceConnections != null) {
+                    cancelled = !serviceConnections.resolveSharedCredentials(dataSourceContainer, null);
+                }
+            }
+
+            if (cancelled) {
+                return this;
```
vkirst: On cancel we return this. Won't this reload the editor and show the dialog again? Or is it on purpose? Maybe we should return unloadInput() like line 319.

plugins/org.jkiss.dbeaver.ui.editors.base/src/org/jkiss/dbeaver/ui/editors/DatabaseLazyEditorInput.java
```diff
@@ -283,6 +284,17 @@ public IDatabaseEditorInput initializeRealInput(@NotNull DBRProgressMonitor moni
         long connectionTimeout = dataSourceContainer.getPreferenceStore().getInt(ModelPreferences.CONNECTION_VALIDATION_TIMEOUT);
         long connectionStart = System.currentTimeMillis();
         while (!dataSourceContainer.isConnected()) {
+            boolean cancelled = false;
```
vkirst: Same block is duplicated in DBUtils.findObjectById. The logic is identical, only the return differs. Could we move it to a shared helper?

PR #41117, opened 2026-05-19
dbeaver/dbeaver#37519: add escaped char

### LINKED ISSUE #37519: When exporting json/jsonb data to xml attributes like when exporting to dbunit xml escape the json
### Description

Jsonb/json columns should be xml escaped when export to dbunit xml files.

### DBeaver Version

Community Edition 25.0.0.202503021833

### Operating System

Macos 15.3.1 (24D70)

### Database and driver

PostgreSQL 17.4
Postgreql jdbc 42.7.5

### Steps to reproduce

1. Navigate to table data for a table that has json/jsonb column
2. Export to DBUnit XML File(s)
3. Click Next 3 times
4. Click Proceed

Json/Jsonb columns are not escaped.

Interestingly converting the columns to text via a cast or writing to a new table with columns converted to text has no impact: the "s and such are not escaped but other text fields are.

### Additional context

_No response_

SolovevAnton:
closes dbeaver/dbeaver#37519


plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterDbUnit.java
```diff
@@ -53,6 +54,14 @@ public class DataExporterDbUnit extends StreamExporterAbstract {
     private static final String PROP_UPPER_CASE_COLUMN_NAMES = "upperCaseColumnNames";
     private static final String PROP_INCLUDE_NULL_VALUES = "includeNullValues";
 
+    private static final Map<Character, String> DBUNIT_ESCAPE_CHARS = Map.of(
```
ShadelessFox: Map lookups can be quite slow in this particular usage. Moreover, it boxes every `char` it encounters, and it's only cached for values in the Latin-1 range.

Matvey16:
Single quote escape is fixed for DbUnit files, but still reproducible for regular xml file export

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterDbUnit.java
SolovevAnton: refactored to XMLUtils

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/exporter/DataExporterXML.java
```diff
@@ -140,8 +141,10 @@ public void exportFooter(DBRProgressMonitor monitor) {
     private void writeTextCell(@Nullable String value)
     {
         if (value != null) {
-            value = value.replace("&", "&amp;").replace("<", "&lt;").replace(">", "&gt;");
-            getWriter().write(value);
+            String escapedValue = XMLUtils.escapeXml(value);
+            if (escapedValue != null) {
+                getWriter().write(escapedValue);
+            }
```
ShadelessFox, line 149:
```suggestion
            if (escapedValue != null) {
                getWriter().write(escapedValue);
            } else {
                getWriter().write(value);
            }
```

ShadelessFox (changes requested):
And also please add a test

ShadelessFox (approved):

E1izabeth (approved):

Matvey16:
Verified 

DBeaver CE Windows

PR #41118, opened 2026-05-19
#40230 Generate random color based on color value if it can't be parsed

### LINKED ISSUE #40230: GIS viewer geometries grayed and linestring not visible
### Description

Since v24.5 (maybe before) geometries appeared as grayed in the GIS viewer : 

Image

The linestring are not displayed anymore
Polygons are grayed too.

This is only when consulting all the columns from a table (`SELECT *` or `SELECT each column` or map mode when consulting data table.).
If result comes from a request that display only a part of the columns (LIKE a `SELECT geom`) then the geometries are diplayed in blue (or other colors if there is multiples geom columns)

### DBeaver Version

Community edition 25.3.3.202601181752

### Operating System

Windows 11

### Database and driver

PostgreSQL JDBC Driver from local PostgreSQL v17 install.
=> Same thing with DBeaver PostgreSQL binaries and local PostgreSQL v15 install.

### Steps to reproduce

_No response_

### Additional context

Reproduced on every database I use (from PG v11 to PG v18)

ShadelessFox:
This PR changes how we interpret the values of the `color` attribute.

1. If it's a valid hexadecimal string in the form of `#{rr}{gg}{bb}`, we parse the color as-is.
2. If it can't be parsed, we compute a random seed from its hash code and use it to derive a new color.
3. If no attribute is present, we use one of the predefined colors based on the row index.


E1izabeth (approved):

Matvey16:
Verified

DBeaver CE Windows

PR #41119, opened 2026-05-19
dbeaver/pro#9326 RSV: paste new rows option

serge-rider:


Matvey16:
Verified

Dbeaver CE Windows

PR #41123, opened 2026-05-20
dbeaver/pro#9296 fix npe for retrieving ddl in cubrid for shared conn…

yagudin10:
…ections


plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceDescriptor.java
```diff
@@ -533,7 +533,7 @@ public void resetAllSecrets() {
         this.secretsContainsDatabaseCreds = false;
         this.availableSharedCredentials = null;
         this.selectedSharedCredentials = null;
-        if (sharedCredentials) {
+        if (sharedCredentials && !isConnected()) {
```
serge-rider, line 538: Why? Add some comment

PR #41125, opened 2026-05-20
#40521 Try refresh container if object not found

### LINKED ISSUE #40521: Database task fails to re-check for object existence
### Description

I often clear and reload the dev database and have a script that creates a comparison view and a database task that exports it. After a reload and manipulation, I run the script to make the comparison view and then run the export task. Because the view was made from a script, DBeaver doesn't know about the view and the task fails. In version 25, I worked around this by navigating to the views folder and refreshing so it knew that the view existed; the task would work after. However, I need to remove and re-add the view to the task in version 26.

Logs: (can't send emails outside the company anymore, so these will be shortened to save some typing)
!ENTRY org.jkiss.dbeaver.model 4 0 2026-03-11 09:46:32
!MESSAGE Child object '<the view>' not found in container <the schema>
!STACK 0
...DBException: <same as !MESSAGE>
  at org.jkiss.dbeaver.model.DBUtils.findByObjectId(...:606)
  ...chain of deserializing methods...
  ...tools.transfer.DataTransferSettings.getNodesFromLocation(...:737)
  ...tools.transfer.DataTransferSettings.<init>(...:151)
  ...tools.transfer.ui.wizard.DataTransferWizard.lambda$openWizard$1(...:712)
  ...iu.UIUtils.lambda$runWithDialog$20(...:2090) // 2000 lines?! Whoa
  at org.eclipse.jface.operation.ModalContext$ModalContextThread.run(...:123)

!ENTRY org.jkiss.dbeaver.model 4 0 2026-03-11 09:47:42
!MESSAGE Unsupported producer data container: null

### DBeaver Version

Community 26

### Operating System

Win 11 Enterprise

### Database and driver

PostreSQL 42.7.2

### Steps to reproduce

Steps to reproduce:
1. make a table or view
2. make a DB task to export it
3. remove the table or view
4. refresh so DBeaver knows that the table or view is gone
5. remake the table or view using a script (mimics the clear and reload)
6. try to run the DB task
7. Of course the refresh works as I type these up... but it would be nice if I didn't have to navigate to the views folder and refresh every time

### Additional context

Using SSH tunnel

ShadelessFox:
Not sure if this approach is ideal, as it has several drawbacks:
1. It clears all caches of a container, because we don't know what we're looking for, so this can't be avoided.
2. It doesn't update the navigator tree. Should it, really?

Please let me know what you think, dear reviewer :^)


E1izabeth (commented):
I would refresh metadata each time we run the task

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
```diff
@@ -658,6 +676,10 @@ public static DBSObject findObjectById(
             for (int i = 1; i < names.length - 1; i++) {
                 String name = names[i];
                 DBSObject child = sc.getChild(monitor, name);
+                if (child == null && tryRefreshContainers && sc instanceof DBPRefreshableObject ro) {
+                    // Try refreshing, maybe object was not cached
+                    ro.refreshObject(monitor);
```
serge-rider, line 681: Don't get it. You just refresh and don't try to find child again?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/DBUtils.java
ShadelessFox: Indeed. It's funny because it worked. Oops...

E1izabeth (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified

DBeaver CE Windows
Dbeaver EE Windows

serge-rider:
Still not sure, lets discuss before merge

PR #41129, opened 2026-05-21
dbeaver/pro#9354 Fix array size hint in record mode

E1izabeth:
Record mode:
image

Non-record mode:
(the same as it was before)
image


Matvey16:
Verified

DBeaver EE Windows

PR #41131, opened 2026-05-21
dbeaver/pro#9189 Fix invalid editor state on opening sql terminal from vqb view

E1izabeth:
Closes dbeaver/pro#9189


Matvey16:
Verified

DBeaver EE Windows

PR #41147, opened 2026-05-22
Clear dirty state when cell value equals original value

### LINKED ISSUE #41146: Data editor remains dirty when a cell value is changed back to its original value
## Description

In the Data Editor, when a cell value is changed and then changed back to its original value before saving, the cell/row still remains dirty and the Save/Revert actions remain enabled.

## Steps to reproduce

1. Open a table in the Data Editor.
2. Pick an editable cell with original value `A`.
3. Change the value from `A` to `B` and confirm the edit.
4. Change the same cell back from `B` to `A` and confirm the edit.
5. Observe that the cell/row is still marked as modified and Save/Revert remain enabled.

## Expected behavior

When the final value equals the original value, the pending cell change should be removed. If there are no other pending changes, Save/Revert should become disabled.

## Actual behavior

The cell/row remains dirty even though there is no effective change.

### DBeaver Version

25.3.5

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

Zxc4wonder:
## Summary

Fixes dbeaver/dbeaver#41146.

This PR clears a pending cell change in the Data Editor when the edited value is changed back to its original value before saving.

## Problem

Currently, if a user edits a cell from `A` to `B` and then edits it back to `A`, the cell/row may still remain dirty and Save/Revert remain enabled, even though there is no effective change.

## Changes

- Compare the pending change's original value with the current new value after a normal root cell update.
- Clear the cell change with `row.clearChange(topAttribute)` when the values are equivalent.
- Refresh the change count after clearing the pending change.
- Use the existing `DBDValueHandler.getComparator()` and `DBUtils.compareDataValues()` instead of string comparison.
- Keep NULL semantics:
  - `NULL -> X -> NULL` clears dirty.
  - `NULL -> ""` remains dirty.
- Keep the fix limited to ordinary root cell updates and avoid complex values, `DBDValue`, new rows, and deleted row flows.
- Increment `changesCount` only when a new pending change is created, avoiding duplicate increments for repeated edits of the same cell.

## Manual testing

- `A -> B`: the cell becomes dirty and Save/Revert are enabled.
- `A -> B -> A`: dirty state is cleared and Save/Revert are disabled when there are no other changes.
- `NULL -> X -> NULL`: dirty state is cleared.
- `NULL -> ""`: the cell remains dirty.
- Multiple edited cells: changing one cell back to its original value only clears that cell's pending change; other pending changes remain.
- Existing save behavior for normal edits still works.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
@@ -507,10 +507,35 @@ public class ResultSetModel implements DBDResultSetModel {
         row.values[rootIndex] = valueToEdit;
 
         if (updateChanges && row.getState() == ResultSetRow.STATE_NORMAL) {
-            changesCount++;
+            if (attr == topAttribute && !attr.getDataKind().isComplex() && !(valueToEdit instanceof DBDValue)) {
+                Object originalValue = row.getChange(topAttribute);
+                if (row.isChanged(topAttribute) && equalCellValues(attr, originalValue, valueToEdit)) {
+                    row.clearChange(topAttribute);
+                    refreshChangeCount();
+                    return true;
+                }
+            }
+            if (isOldHistoricValueAbsent) {
+                changesCount++;
+            }
         }
 
         return true;
+    }
+
+    private static boolean equalCellValues(
+        @NotNull DBDAttributeBinding attr,
+        @Nullable Object value1,
+        @Nullable Object value2
+    ) {
+        if (DBUtils.isNullValue(value1) || DBUtils.isNullValue(value2)) {
+            return DBUtils.compareDataValues(value1, value2) == 0;
+        }
+        Comparator<Object> comparator = attr.getValueHandler().getComparator();
+        if (comparator != null) {
+            return comparator.compare(value1, value2) == 0;
+        }
+        return DBUtils.compareDataValues(value1, value2) == 0;
     }
 
     void resetCellValue(@NotNull DBDAttributeBinding attr, @NotNull ResultSetRow row, @Nullable int[] rowIndexes) {


Zxc4wonder:
Hi, just checking in on this PR.

I noticed it has been labeled, so I wanted to ask whether there is anything I should adjust or add to help move it forward.

The change is intended to fix the dirty-state behavior when a cell value is changed back to its original value, while preserving NULL semantics and keeping the patch minimal.

Thanks!

E1izabeth:
Hi, @xxxz3! Thanks for the contribution! Sorry for the long delay. We'll review your PR and let you know

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetModel.java
```diff
@@ -507,12 +507,37 @@ boolean updateCellValue(
...
+        @Nullable Object value1,
+        @Nullable Object value2
+    ) {
+        if (DBUtils.isNullValue(value1) || DBUtils.isNullValue(value2)) {
+            return DBUtils.compareDataValues(value1, value2) == 0;
+        }
+        Comparator<Object> comparator = attr.getValueHandler().getComparator();
+        if (comparator != null) {
+            return comparator.compare(value1, value2) == 0;
+        }
```
ShadelessFox, line 537: Why not invert the null check so `compareDataValues` doesn't appear twice?


```suggestion
        if (!DBUtils.isNullValue(value1) && !DBUtils.isNullValue(value2)) {
            Comparator<Object> comparator = attr.getValueHandler().getComparator();
            if (comparator != null) {
                return comparator.compare(value1, value2) == 0;
            }
        }
```

Zxc4wonder: Agreed, this is cleaner.

IvanGorshechnikov: So, please, add it to your PR

E1izabeth: @Zxc4wonder please address the requested change

E1izabeth:

PR #41166, opened 2026-05-25
dbeaver/pro#9396 Fix export zero date and zero timestamp from db to db

E1izabeth:


Destrolaric (approved):

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/data/handlers/JDBCDateTimeValueHandler.java
```diff
@@ -255,6 +255,8 @@ protected static java.sql.Date getDateValue(Object value) {
             return java.sql.Date.valueOf(localDate);
         } else if (value instanceof LocalDateTime localDateTime) {
             return java.sql.Date.valueOf(localDateTime.toLocalDate());
+        } else if (value == DBDZeroTimestampValue.INSTANCE || value == DBDZeroDateValue.INSTANCE) {
+            return new java.sql.Date(0);
```
serge-rider, line 259: Date(0) is 1st January 1970
Doesn't look right

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.jdbc/src/org/jkiss/dbeaver/model/impl/jdbc/data/handlers/JDBCDateTimeValueHandler.java
E1izabeth: It used to be this way
https://github.com/dbeaver/dbeaver/commit/335e1643ab48385e44b6895ff198a4b458d0052a
We can fix it separately

E1izabeth: https://github.com/dbeaver/pro/issues/9403

Matvey16:
Verified

DBeaver CE Windows

PR #41172, opened 2026-05-26
dbeaver/pro#8619 Don't log file connections during data transfer

E1izabeth:


SolovevAnton (dismissed):

IvanGorshechnikov (dismissed):

Matvey16:
This creates a warning after every import. The import is successful, but the warnings is confusing. Maybe it's not actually a good idea to skip that logging then? Maybe it doesn't even need fixing?

image

SolovevAnton (dismissed):

IvanGorshechnikov (dismissed):

IvanGorshechnikov (approved):

SolovevAnton (approved):

PR #41177, opened 2026-05-26
dbeaver/pro#9409 Fix NPE

serge-rider:


Matvey16:
Verified

DBeaver CE Windows

PR #41191, opened 2026-05-28
dbeaver/pro#9412 Mitigate NPE by explicitly loading view's DDL if it's absent

ShadelessFox:
It's a blind fix, and I was not able to reproduce the original issue. Ask the issuer for more details.


Matvey16:
Verified

DBeaver EE Windows

Issue #41142, opened 2026-05-22
Grammar bug: "Refresh every 1 seconds" should be "1 second"


Jashan66:
### Description

The auto-refresh menu in the SQL editor displays "Refresh every 1 seconds" which is grammatically incorrect (plural form used with singular number).

**Expected behavior:**
- "Refresh every 1 second" (singular when interval = 1)
- "Refresh every 5 seconds" (plural when interval > 1)

**Current behavior:**
- "Refresh every 1 seconds" ❌
- "Refresh every 5 seconds" ✅

**Location:**
`plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages.properties`
Lines 39-40

**Suggested fix:**
Use Java's `ChoiceFormat` for proper pluralization:
```properties
sql_editor_resultset_filter_panel_menu_refresh_interval = Refresh every {0,choice,1#1 second|1<{0} seconds}
sql_editor_resultset_filter_panel_menu_refresh_interval_1 =                 ... {0,choice,1#1 second|1<{0} seconds}
```

**Note:** This would need to be applied to all localized property files (UIMessages_de.properties, UIMessages_fr.properties, UIMessages_ru.properties, etc.)

<img width="806" height="384" alt="Image" src="https://github.com/user-attachments/assets/31105aa5-1ecf-4aab-a008-4cb7fd1dd700" />

### DBeaver Version

25.1.1

### Steps to reproduce

1. Open any SQL editor with a result set
2. Click the auto-refresh configuration button in the result set toolbar
3. Observe the menu items showing refresh intervals
4. Note the first item displays "Refresh every 1 seconds"

### Additional context

This affects the English version and would need corresponding fixes in all localized versions to maintain consistency across languages.

E1izabeth:
Thanks for the report

Jashan66:
Thank you! I'll submit a PR for this.

awaissaeed530:
@Jashan66 Don't worry about my PR, I'm just learning the codebase for further contributions

AkshatRaj00:
Hi! I'd like to work on this fix.

I've traced the string to `plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages.properties`. The fix involves using Java's `MessageFormat` with `ChoiceFormat` pattern to handle singular/plural correctly:

**Current:**
```properties
[...]

PR #41199, opened 2026-05-29
fix: correct singular 'second' pluralization in auto-refresh menu labels

algojogacor:
This is an independent contribution and is not associated with any hackathon or competition.

## Summary

The auto-refresh interval menu in the SQL editor displays "Refresh every 1 seconds" which is grammatically incorrect. The singular form "second" should be used when the interval is 1.

## Root Cause

The i18n properties only had a single plural form ("{0} seconds") regardless of the numeric value, and the Java code directly substituted the interval value without any singular/plural logic.

## Fix

- Added singular i18n keys (`sql_editor_resultset_filter_panel_menu_refresh_interval` and `sql_editor_resultset_filter_panel_menu_refresh_interval_1_s`) with "second" (singular) text
- Updated `AutoRefreshControl.java` to select between singular and plural keys based on the timeout value

## Testing

- Interval = 1: Displays "Refresh every 1 second" ✓
- Interval = 5: Displays "Refresh every 5 seconds" ✓
- Custom intervals also correctly use singular/plural forms


ShadelessFox:
I assume this PR targets https://github.com/dbeaver/dbeaver/issues/41142. The issue already has an open PR. No need to submit even more AI slop. Even more, per our contribution policy, you're supposed to disclose AI usage, which you didn't.

PR #41203, opened 2026-05-30
dbeaver/pro#9463 Driver edit dialog UIX

serge-rider:


ShadelessFox:
Would be nice to see some screenshots

PR #41206, opened 2026-06-01
#21481 Support tilde (~) in SSH tunnel private key path

### LINKED ISSUE #21481: Relative paths not supported for SSH Tunnel Private Key
### Description

DBeaver doesn't support using relative paths relative to the home directory (e.g. `~/.ssh/key.pem`) on macOS.

### DBeaver Version

Community Edition Version 23.2.2.202310081633

### Operating System

macOS Ventura 13.5.1

### Database and driver

N/A

### Steps to reproduce

_No response_

### Additional context

Using `${HOME}/.ssh/key.pem` also works, but that is way less intuitive.

Laotree:
Closes #21481


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/GeneralUtils.java
```diff
@@ -560,6 +560,12 @@ public static String replaceVariables(@NotNull String string, @NotNull IVariable
         if (CommonUtils.isEmpty(string)) {
             return string;
         }
+        // Normalize Unix tilde prefix to ${home} so it goes through the same variable resolution
+        if (string.startsWith("~/") || string.startsWith("~\\")) {
+            string = "${home}" + string.substring(1);
+        } else if ("~".equals(string)) {
+            string = "${home}";
+        }
```
ShadelessFox: Could be simplified to this
```suggestion
        if (string.startsWith("~")) {
            string = "${home}" + string.substring(1);
        }
```

E1izabeth (commented):
I believe it's not DBeaver's responsibility to solve this prefix

Laotree:
@E1izabeth If you still believe this isn't within DBeaver's responsibility, I'll close the PR.
However, I'd kindly ask you to reconsider: ~ is a universal cross-platform shorthand for the home directory. Users naturally type ~/.ssh/id_rsa when entering a private key path. The expansion is now scoped strictly to the SSH key path input in SSHUtils.loadHostConfiguration(), so the risk surface is minimal while the UX improvement is tangible.

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -382,6 +382,29 @@ private static void saveHostConfiguration(
...
+            if (home != null) {
+                return home + path.substring(1);
+            }
+        } else if ("~".equals(path)) {
+            String home = System.getProperty("user.home");
+            if (home != null) {
+                return home;
+            }
+        }
+        if (RuntimeUtils.isWindows() && path.startsWith("%USERPROFILE%")) {
```
ShadelessFox, line 399: Environment variables on Windows are case-insensitive

plugins/org.jkiss.dbeaver.net.ssh/src/org/jkiss/dbeaver/model/net/ssh/SSHUtils.java
```diff
@@ -382,6 +382,29 @@ private static void saveHostConfiguration(
...
+            String home = System.getProperty("user.home");
+            if (home != null) {
+                return home + path.substring(1);
+            }
+        } else if ("~".equals(path)) {
+            String home = System.getProperty("user.home");
+            if (home != null) {
+                return home;
+            }
+        }
```
ShadelessFox, line 398: In contrast, Windows _does not_ have `~` alias.

E1izabeth:
We have once again discussed the provided changes and decided not to accept them. It's an undocumented feature that only appears in one place. Either it should be supported globally (not sure we really want it), or it should not be handled this way.

PR #41210, opened 2026-06-01
dbeaver/pro#9476 synchronized for constraints check

Destrolaric:
Closes dbeaver/pro#9476 
Ignore formatting because try{} finally
Also removed synchronized from some places used compareSet instead


serge-rider:
It is a potential behavior changer. Also I think changes in updateColorMapping may lead to deadlock.
For hotfix we need a change with minimal impact.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/ResultSetViewer.java
```diff
@@ -4674,7 +4674,6 @@ void releaseDataReadLock() {
     boolean acquireDataReadLock() {
         synchronized (dataPumpJobQueue) {
             if (dataPumpRunning.get()) {
-                log.debug("Internal error: multiple data reads started (" + dataPumpJobQueue + ")");
```
ShadelessFox, line 4677: Why did you remove it?

PR #41216, opened 2026-06-01
Fix pluralization grammar bug in refresh interval message

### LINKED ISSUE #41142: Grammar bug: "Refresh every 1 seconds" should be "1 second"

Saatake:
### Overview
Fixed a grammar/pluralization bug in the SQL Editor results panel where it would incorrectly display "Refresh every 1 seconds" instead of "1 second".

### Changes
- Changed the resource strings to use `{0,choice,1}` formatting to correctly handle singular and plural forms for the refresh interval.

Resolves #41142

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages.properties
+++ b/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages.properties
@@ -36,8 +36,8 @@ sql_editor_resultset_filter_panel_btn_stop_refresh = Stop auto-refresh
 sql_editor_resultset_filter_panel_btn_config_refresh = Configure auto-refresh
 sql_editor_resultset_filter_panel_menu_customize = Customize ...
 sql_editor_resultset_filter_panel_menu_stop = Stop
-sql_editor_resultset_filter_panel_menu_refresh_interval = Refresh every {0} seconds
-sql_editor_resultset_filter_panel_menu_refresh_interval_1 =                 ... {0} seconds
+sql_editor_resultset_filter_panel_menu_refresh_interval = Refresh every {0,choice,1#1 second|1<{0} seconds}
+sql_editor_resultset_filter_panel_menu_refresh_interval_1 =                 ... {0,choice,1#1 second|1<{0} seconds}
 
 controls_locale_selector_group_locale = Locale
 controls_locale_selector_label_country = Country
--- a/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages_fr.properties
+++ b/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages_fr.properties
@@ -26,8 +26,8 @@ sql_editor_resultset_filter_panel_btn_stop_refresh = Arrêter le rafraîchisseme
 sql_editor_resultset_filter_panel_btn_config_refresh = Configurer le rafraîchissement automatique
 sql_editor_resultset_filter_panel_menu_customize = Personnaliser ...
 sql_editor_resultset_filter_panel_menu_stop = Stop
-sql_editor_resultset_filter_panel_menu_refresh_interval = Rafraîchir toutes les {0} secondes
-sql_editor_resultset_filter_panel_menu_refresh_interval_1 =                 ... {0} secondes
+sql_editor_resultset_filter_panel_menu_refresh_interval = Rafraîchir {0,choice,1#toutes les 1 seconde|1<toutes les {0} secondes}
+sql_editor_resultset_filter_panel_menu_refresh_interval_1 =                 ... {0,choice,1#1 seconde|1<{0} secondes}
 
 controls_locale_selector_group_locale = Localisation
 controls_locale_selector_label_country = Pays
--- a/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages_de.properties
+++ b/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages_de.properties
@@ -24,8 +24,8 @@ controls_progress_page_progress_bar_loading_tooltip = Ladefortschritt
 controls_progress_page_toolbar_title = Suche Element(e)
 
 sql_editor_resultset_filter_panel_menu_customize = Personalisieren ...
-sql_editor_resultset_filter_panel_menu_refresh_interval = Alle {0} Sekunden aktualisieren
-sql_editor_resultset_filter_panel_menu_refresh_interval_1 = \                ... {0} Sekunden
+sql_editor_resultset_filter_panel_menu_refresh_interval = Alle {0,choice,1#1 Sekunde|1<{0} Sekunden} aktualisieren
+sql_editor_resultset_filter_panel_menu_refresh_interval_1 =                 ... {0,choice,1#1 Sekunde|1<{0} Sekunden}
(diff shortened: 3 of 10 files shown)


plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages.properties
```diff
@@ -36,8 +36,8 @@ sql_editor_resultset_filter_panel_btn_stop_refresh = Stop auto-refresh
 sql_editor_resultset_filter_panel_btn_config_refresh = Configure auto-refresh
 sql_editor_resultset_filter_panel_menu_customize = Customize ...
 sql_editor_resultset_filter_panel_menu_stop = Stop
-sql_editor_resultset_filter_panel_menu_refresh_interval = Refresh every {0} seconds
-sql_editor_resultset_filter_panel_menu_refresh_interval_1 =                 ... {0} seconds
+sql_editor_resultset_filter_panel_menu_refresh_interval = Refresh every {0,choice,1} seconds
```
sathish256: `{0,choice,1}` is incomplete — Java `ChoiceFormat` requires `limit#text` branches separated by `|`.

This still renders as **"Refresh every 1 seconds"** because the word `seconds` stays outside the choice and is always plural.

Please use the pattern from #41142 (and consistent with `ModelResources.properties`):

```properties
sql_editor_resultset_filter_panel_menu_refresh_interval = Refresh every {0,choice,1#1 second|1<{0} seconds}
sql_editor_resultset_filter_panel_menu_refresh_interval_1 =                 ... {0,choice,1#1 second|1<{0} seconds}
```

Reference in-repo:
```properties
duration_formatter_seconds = {0,number,integer} {0,choice,1#second|1<seconds}
```

The same full choice pattern should be applied in each translated `.properties` file (with locale-appropriate singular/plural forms), not `{0,choice,1}` by itself.

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages_fr.properties
```diff
@@ -26,11 +26,11 @@ sql_editor_resultset_filter_panel_btn_stop_refresh = Arrêter le rafraîchisseme
 sql_editor_resultset_filter_panel_btn_config_refresh = Configurer le rafraîchissement automatique
 sql_editor_resultset_filter_panel_menu_customize = Personnaliser ...
 sql_editor_resultset_filter_panel_menu_stop = Stop
-sql_editor_resultset_filter_panel_menu_refresh_interval = Rafraîchir toutes les {0} secondes
-sql_editor_resultset_filter_panel_menu_refresh_interval_1 =                 ... {0} secondes
+sql_editor_resultset_filter_panel_menu_refresh_interval = Rafraîchir toutes les {0,choice,1} secondes
+sql_editor_resultset_filter_panel_menu_refresh_interval_1 =                 ... {0,choice,1} secondes
 
 controls_locale_selector_group_locale = Localisation
```
sathish256, line 32: This looks unrelated to the refresh-interval fix — please revert:

```diff
- controls_locale_selector_label_country = Pay
+ controls_locale_selector_label_country = Pays
```

`Pay` is incorrect French; the original **Pays** (country) was right.

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages_fr.properties
```diff
@@ -26,11 +26,11 @@ sql_editor_resultset_filter_panel_btn_stop_refresh = Arrêter le rafraîchisseme
 sql_editor_resultset_filter_panel_btn_config_refresh = Configurer le rafraîchissement automatique
 sql_editor_resultset_filter_panel_menu_customize = Personnaliser ...
 sql_editor_resultset_filter_panel_menu_stop = Stop
-sql_editor_resultset_filter_panel_menu_refresh_interval = Rafraîchir toutes les {0} secondes
-sql_editor_resultset_filter_panel_menu_refresh_interval_1 =                 ... {0} secondes
+sql_editor_resultset_filter_panel_menu_refresh_interval = Rafraîchir toutes les {0,choice,1} secondes
```
sathish256: Same issue as English: `{0,choice,1}` without branches won't fix singular/plural. For French, something like:

```properties
sql_editor_resultset_filter_panel_menu_refresh_interval = Rafraîchir {0,choice,1#toutes les 1 seconde|1<toutes les {0} secondes}
```

(Exact wording can be tuned — the important part is `1#…|1<…` branches, not bare `{0,choice,1}`.)

sathish256 (commented):
Thanks for tackling #41142 — the problem and locale coverage look right, but the `MessageFormat` pattern needs to be completed before this will fix the bug.

**Summary**
- `{0,choice,1}` alone is not valid choice syntax; you still get "1 seconds" (or a formatting error).
- Use full branches like the issue suggests and like existing `ModelResources.properties` (`duration_formatter_seconds`).
- Please revert the unrelated `Pays` → `Pay` change in `UIMessages_fr.properties`.

plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/internal/UIMessages_fr.properties
Saatake: Thanks for the guidance, @sathish256! 

I just pushed the fixes correcting the `ChoiceFormat` syntax across all translated property files (including German, French, Italian, Japanese, Portuguese, Romanian, Russian, Ukrainian, and Chinese) to handle the sensory/plural forms properly. 

I also restored the original `Pays` string in `UIMessages_fr.properties`. Let me know if everything looks good now! 🙏

ShadelessFox:
Thanks, but this is a fourth PR already. We have already accepted one, and it is currently awaiting QA.

Moreover, your proposed fix doesn't even work - and you didn't even bother verifying it.

> `NLS#bind` does not support rich formatting like `MessageFormat` does. It is **explicitly** stated in its Javadoc:
> 
> > The `#bind` methods perform string substitution and should be considered a
> > convenience and _not_ a full substitute replacement for `MessageFormat#format`
> > method calls.
> 
> So it would just throw an exception:
> 
> image

PR #41236, opened 2026-06-03
Implement OAuth token handling in Snowflake provider

mencwelp:
Added OAuth token fetching logic to SnowflakeDataSourceProvider.

This Snowflake plugin update resolves the issue of rejected OAuth tokens ("Invalid OAuth access token") by bypassing the DBeaver Secure Storage mechanism and its internal connection properties caching. The authorisation token is now fetched in the background and injected directly into the JDBC URL.

HOW TO CONFIGURE SSO CONNECTION IN DBEAVER:

For OAuth authorisation to work correctly, configure your connection as follows:

1. "Main" tab (General connection settings):
   - Username: [Your login, e.g., dbtprod]
   - Password: MUST REMAIN COMPLETELY EMPTY! (Entering a password here will block SSO).

2. "Driver Properties" tab:
   - Right-click in the properties window -> "Add new property"
   - Name: oauth.password
   - Value: [Your actual password]

3. Other required Driver Properties:
   - oauth.token.endpoint (URL to your PingFederate / IDP server)
   - oauthClientId (Client ID)
   - oauthClientSecret (Client Secret)
   - oauth.scopes (e.g., session:role-any)

How it works under the hood:
The plugin automatically intercepts the `oauth.password`, makes a request to the IDP server, fetches a fresh token, removes all traces of the password from the properties map, and securely passes the pure token to the native Snowflake driver via a parameterised JDBC URL.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.snowflake/src/org/jkiss/dbeaver/ext/snowflake/SnowflakeDataSourceProvider.java
+++ b/plugins/org.jkiss.dbeaver.ext.snowflake/src/org/jkiss/dbeaver/ext/snowflake/SnowflakeDataSourceProvider.java
@@ -6,7 +6,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


ShadelessFox:
Thanks for the contribution.

DBeaver natively supports OAuth-based Snowflake authorization in PRO products via a dedicated auth model. Your changes are invasive and not user-friendly at all, so we can't accept your PR.

PR #41248, opened 2026-06-04
dbeaver/pro#7204 Easy config wizard mockup

ShadelessFox:


IvanGorshechnikov (dismissed):

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.debug.ui/plugin.xml
```diff
@@ -199,4 +199,18 @@
 
     </extension>
 
+    <extension point="org.jkiss.dbeaver.easyConfig">
+        <feature id="debug" label="Procedure debugger" enabledByDefault="false"/>
+    </extension>
+
+    <extension point="org.eclipse.ui.activities">
+        <activity id="dbeaver.activity.debug" name="Dashboards Activity">
```
serge-rider, line 207: Dashboard?

plugins/org.jkiss.dbeaver.debug.ui/plugin.xml
```diff
@@ -199,4 +199,18 @@
...
+            <enabledWhen>
+                <test property="org.jkiss.dbeaver.easyConfig.isFeatureEnabled" value="debug"/>
```
serge-rider: Let's trename org.jkiss.dbeaver.easyConfig into org.jkiss.dbeaver.config?

plugins/org.jkiss.dbeaver.ui.app.config/src/org/jkiss/dbeaver/ui/app/config/nls/ProductConfigMessages.java
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.ui.config.easy.nls;
+
+import org.jkiss.dbeaver.ui.forms.UIObservable;
+import org.jkiss.dbeaver.ui.forms.util.UIReloadableNLS;
+
+public final class EasyConfigMessages extends UIReloadableNLS {
+    public static UIObservable<String> welcome_title;
```
serge-rider, line 23: 😢

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.debug.ui/plugin.xml
ShadelessFox: I don't know. It's exclusively used for Easy Config-related stuff. Do you want me to rename all instances of the name everywhere else?

plugins/org.jkiss.dbeaver.ui.app.config/src/org/jkiss/dbeaver/ui/app/config/nls/ProductConfigMessages.java
ShadelessFox: Well, I'm yet to hear a better solution for hot-reloadable NLS strings :)

The one option you've proposed is too intrusive, contradicts the concepts of Forms, and kinda does the same thing as `UIObservable`, which just duplicates functionality without any real use, limiting it to language-related shenanigans.

Another option I can think of is a two-step process:
1. Define standard messages by extending the standard `NLS` class;
2. Create a global listener that is fired when the language is changed;
3. Construct `UIObservable` where needed by passing in the message field in a lambda alongside the listener so the observable knows when it needs to be invalidated when the language is changed.

Just a mockup:
```java
class EasyConfigWizard extends XXX {
    ...
    UIObservable<String> localizedText(Supplier<String> nlsTextSupplier) { ... }
}
```
And then it is used like:
```java
Consumer<UIPanelBuilder> buildLabel() {
    return pb -> pb
        .row(rb -> rb
            .label(getWizard().localizedText(() -> MyMessages.my_message)));
}
```

This way, the implementation is hidden, but the lifetime of each created `UIObservable` is tied to the wizard's lifetime, but it becomes pretty much specific to Easy Config as the helper method lives in the wizard.

It would be nice to have a static helper class designed exclusively for dealing with localized strings like:
```java
class UILocalizer {
    ...
    void setLocale(Locale locale);

    UIObservable<String> localize(Supplier<String> textSupplier) 
}
```

but there's no way I can easily refresh NLS classes because I can't know from which NLS class the supplied text is coming from. It also suffers from the same lifetime issue as UIReloadableNLS (see below)

Unfortunately, `UIReloadableNLS` is far from being ideal; see **Known issues** section in https://github.com/dbeaver/pro/issues/9347. TL;DR: it wouldn't be too bad if not for `UIObservable#addChangeListener`. Since instances of `UIObservable`' declared in subclasses of `UIReloadableNLS` never get disposed of because they are static, it is **mandatory** to manually remove added listeners when they are no longer needed, which is very bug-prone and may lead to various weird issues, including memory leaks.

plugins/org.jkiss.dbeaver.debug.ui/plugin.xml
serge-rider: Id better call it AppConfig.

plugins/org.jkiss.dbeaver.ui.app.config/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,19 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: DBeaver Product Configuration UI
+Bundle-SymbolicName: org.jkiss.dbeaver.ui.app.config;singleton:=true
+Bundle-Version: 1.0.0.qualifier
+Bundle-Release-Date: 20260622
+Require-Bundle: org.eclipse.core.resources,
```
serge-rider, line 7: No need to add eclipse deps, they come from model

serge-rider (changes requested):

PR #41254, opened 2026-06-06
#41253 Build CubridDataSource for generic driver connections

### LINKED ISSUE #41253: [CUBRID] ClassCastException (GenericDataSource cannot be cast to CubridDataSource) when browsing tables over a generic connection
### Description

When connecting to CUBRID through the **generic** JDBC driver (instead of the bundled CUBRID driver) and expanding the table list, DBeaver throws a ClassCastException and fails to load metadata.

CubridMetaModel is slected by driver-class match (`cubrid.jdbc.driver.CUBRIDDriver`), but the generic connection path builds a plain `GenericDataSource`. CubridMetaModel then casts the data source to `CubridDataSource` without checking, so it fails as soon as metadata loading starts.

Error:
```
  class org.jkiss.dbeaver.ext.generic.model.GenericDataSource cannot be cast to class
  org.jkiss.dbeaver.ext.cubrid.model.CubridDataSource
```

### DBeaver Version

Community Edition 26.1.0.202605311718

### Operating System

macOS

### Database and driver

1. Database: CUBRID
2. Driver: Generic JDBC connection using the CUBRID JDBC driver (cubrid.jdbc.driver.CUBRIDDriver)

### Steps to reproduce

1. Create a new connection using the Generic driver
2. Configure it with the CUBRID JDBC driver (class cubrid.jdbc.driver.CUBRIDDriver), URL jdbc:CUBRID:host:path:db:::
3. Connect and expand Tables
4. A ClassCastException is raised and the table list fails to load

### Additional context

Other generic-based extensions (e.g. Altibase) override `createDataSourceImpl()` in their metamodel to build their dedicated data source, but CUBRID misses this, causing the mismatch.
A fix has been prepared and a PR will follow shortly.

Srltas:
## Description
Connecting to CUBRID with the generic driver (instead of the bundled CUBRID driver) throws a ClassCastException as soon as you expand the table list:
```
class org.jkiss.dbeaver.ext.generic.model.GenericDataSource cannot be cast to class 
org.jkiss.dbeaver.ext.cubrid.model.CubridDataSource
```
CubridMetaModel is picked by driver class match (`cubrid.jdbc.driver.CUBRIDDriver`), but on the generic provider path the data source is created as a plain GenericDataSource. CubridMetaModel then casts it to CubridDataSource unconditionally, so it fails right when metadata loading starts. The underlying reason is that CubridMetaModel never overrode `createDataSourceImpl()`, unlike Altibase and most other generic extensions.

## Changes
- Override `createDataSourceImpl()` in CubridMetaModel so the generic path builds a CubridDataSource as well. This makes the cast safe on both the dedicated and generic provider paths.
- Remove the now redundant `openDataSource()` override in CubridDataSourceProvider, since the parent provider already routes through `createDataSourceImpl()`.
- Keep the metamodel passed into the CubridDataSource constructor instead of allocating a new one.

Fixes #41253


E1izabeth (approved):

ShadelessFox (approved):
Thanks! Passing to QA

uslss:
verified

PR #41258, opened 2026-06-08
dbeaver/dbeaver#38133 Introduce spreadsheet quick filter

### LINKED ISSUE #38133: Feature Request: Add Filter Rows Input for Query Result Set (Like MySQL Workbench)
Add a filter tab or inline filter bar on top of the query result set — similar to what MySQL Workbench offers. This allows to quickly filter the existing query results without rerunning the query.
after running a broad query i want to quickly search specific values
filtering in-place improves productivity especially when working with large datasets.

Image

E1izabeth:
Use `Ctrl+F` to activate or from the toolbar
image

https://github.com/user-attachments/assets/812e90ac-4a22-4976-9e2c-1d08d2c5f423

- Works in any grid, JSON/XML/Text view (find is everywhere, replace only where it is supported):
image

- It has history:
image

- Also autocomplete works for quick filter history
image


IvanGorshechnikov (dismissed):

ShadelessFox:
The overlay really needs a border. It blends badly with the spreadsheet behind it.

plugins/org.jkiss.dbeaver.ui.editors.data/OSGI-INF/l10n/bundle.properties
```diff
@@ -174,6 +174,8 @@ command.org.jkiss.dbeaver.ui.editors.data.forSelection.name=View Data
 command.org.jkiss.dbeaver.ui.editors.data.forSelection.description=View/edit table  data
 command.org.jkiss.dbeaver.ui.editors.data.generate.uuid.name=Generate UUID
 command.org.jkiss.dbeaver.ui.editors.data.generate.uuid.description=Generates UUID/GUID and inserts into active text editor or data editor
+org.jkiss.dbeaver.core.resultset.spreadsheet.openQuickFilter.name=Open Spreadsheet Quick Filter
```
ShadelessFox: Action's name doesn't match other action names' convention


```suggestion
org.jkiss.dbeaver.core.resultset.spreadsheet.openQuickFilter.name=Open spreadsheet quick filter
```

E1izabeth:
It actually has a border.. it's of the same color as for similar overlay in SQL Editor.
And this color is darker than our grid grey lines.
Maybe I could make it darker? But then both here and in SQL Editor..

ShadelessFox:
That gray border's color is just the composite color or whatever it is; it's unusable for border color. Consider using a single-pixel line of `SWT.COLOR_WIDGET_NORMAL_SHADOW`; see `org.jkiss.dbeaver.ui.CompositeBorderPainter`

E1izabeth:
imho the "border" of such color doesn't look good.
@ShadelessFox wdyt?
image
image

ShadelessFox:
Of course it doesn't, because it's the background, not a border! See https://github.com/dbeaver/dbeaver/pull/41258#issuecomment-4649698117

plugins/org.jkiss.dbeaver.ui.editors.data/plugin.xml
```diff
@@ -510,7 +510,7 @@
             </enabledWhen>
         </handler>
         <handler commandId="org.eclipse.ui.edit.findReplace" class="org.jkiss.dbeaver.ui.controls.resultset.handler.ResultSetHandlerMain">
-            <activeWhen><reference definitionId="org.jkiss.dbeaver.core.ui.resultset.part"/></activeWhen>
+            <activeWhen><reference definitionId="org.jkiss.dbeaver.core.ui.resultset.presentation.control"/></activeWhen>
```
serge-rider: Won't it make find/replace inaccessible if focus is in filter panel?

E1izabeth: fixed

E1izabeth:
Added the message that data loading on scroll won't work when find and teplace overlay is opened





Also made the toolbar button to be a toggle

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/controls/resultset/spreadsheet/SpreadsheetPresentation.java
```diff
@@ -227,6 +229,44 @@ public void controlResized(ControlEvent e) {
...
+            this.controller.getSite().getPart(),
+            this.spreadsheet,
+            this.findReplaceTarget,
+            this,
+            f -> {
+                this.controller.getModel().setQuickFilter(f);
+                this.findReplaceOverlay.setExtrasVisibility(f != null);
+                this.refreshData(false, false, false);
+            },
+            c -> UIUtils.createInfoLabel(c, "Data autoloading is temporarily disabled.").setToolTipText("Close quick search or reset 'Search all' filter with empty string to enable it back.")
```
ShadelessFox: I would like the text to be more straight - as of now, it sounds kinda cryptic.

Something like `New data won't be loaded while the quick search is open`, but it's too long and will just stretch the panel... 🤔

E1izabeth:
Updated icons, background color and tooltips
image
image

The `Show matching only` button can now be a toggle (default is a toggle)
image
However, I left old behavior when it can be a one-time-applied (button). Because in this case, we can hide unmatched rows, then continue searching using highlighting
image
Example of such a use case:
image

ShadelessFox (approved):

uslss:
verified (db ue)

PR #41261, opened 2026-06-08
dbeaver/pro#9295 POC metadata transfer

IvanGorshechnikov:
Closes https://github.com/dbeaver/pro/issues/9295


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/struct/rdb/DBSTableForeignKey.java
```diff
@@ -17,18 +17,24 @@
...
 import org.jkiss.dbeaver.model.struct.DBSEntityAssociation;
+import org.jkiss.dbeaver.model.struct.DBSEntityConstraint;
 import org.jkiss.dbeaver.model.struct.DBSEntityReferrer;
 
 /**
  * DBSTableForeignKey
  */
-public interface DBSTableForeignKey extends DBSEntityReferrer, DBSEntityAssociation
-{
+public interface DBSTableForeignKey extends DBSEntityReferrer, DBSEntityAssociation {
```
serge-rider, line 26: Currently all DBS elements interfaces are read-only. LEts keep it this way.
Lets create DBSTableForeignKeyEditable. All setXxx function must be there.
In editors use instanceof to check enablemnent. Implementations whcih do not support it won't support FK migration.

serge-rider (changes requested):

PR #41270, opened 2026-06-09
#41259 Fix navigator delete objects dialog layout

### LINKED ISSUE #41259: Make delete database list scrollable
### Description

A selection of many databases for deletion leads to a UI overflow preventing further interaction because buttons are outside screen.

App is installed via Fedora/Gnome Software Center as flatpak.

### DBeaver Version

Version 26.1.0.202605311718

### Operating System

Fedora 44 Workstation with Gnome and Wayland

### Database and driver

_No response_

### Steps to reproduce

- (irrelevant:) use a postgres container for testing purposes in Python
- (irrelevant:) If tests run through, the temporary database gets cleaned up properly, but if I have to interrupt with Keyboard Interrupt, then the database with random (uuid) name stays. I regularly clean these up manually in DBeaver. But today, after many test runs, I had dozens of such databases.
-  Selected dozens of databases, open context menu and press delete. 
- The list is so long, that the actual "delete" button of the popover/dialog does not appear on screen, because the database list is not scrollable and extended beyond the bottom of the screen.

### Additional context

_No response_

ShadelessFox:
Limited the height of the table to show 10 rows at most:

java_957vJWUd1A


E1izabeth (approved):

uslss:
verified

PR #41294, opened 2026-06-11
dbeaver/pro#9570 Fix NPE

ShadelessFox:
Not sure if it's worth testing https://github.com/dbeaver/pro/issues/9453 again, it's up to you.


E1izabeth (approved):

E1izabeth:
Agree. Let's add self-check label and move to merge after second review

PR #40683, opened 2026-04-02
#40681 Support parsing Array(JSON) types

EastLord:
## Reproduction
- ClickHouse JDBC `0.9.8`
- ClickHouse 26.3.3.20

image

SQL:

```sql
CREATE TABLE test_array_json (
    id UInt8,
    items Array(JSON)
) ENGINE = Memory;

INSERT INTO test_array_json VALUES
(1, ['{"a":1}', '{"b":[1,2,3]}']);
```

Then execute:

```sql
SELECT * FROM test_array_json;
```

image

## Result after the fix

image

PR #41324, opened 2026-06-17
#41315 Render and edit ClickHouse JSON columns via the JSON content viewer

### LINKED ISSUE #41315: ClickHouse JSON columns: render and edit via the JSON value viewer
ClickHouse has a native `JSON` data type (stable since 24.8). In DBeaver, columns of this type are shown as a raw `OTHER` value with no JSON viewer - the Value panel offers no formatted JSON/Text view, and the cell shows the driver's flattened `java.util.Map` representation (`{amount=…, user.id=…}`), which is not valid JSON. The data is hard to read, and the cell cannot be edited (writing the flattened form back fails server-side with `Code 117 INCORRECT_DATA`). PostgreSQL `json`/`jsonb` columns, by contrast, render in the JSON viewer and are editable.

ClickHouse `JSON` columns should render as formatted JSON in the Value panel (like PostgreSQL `json`/`jsonb`) - in both the table Data editor and SQL result grids - and be editable (ClickHouse applies edits via `ALTER TABLE … UPDATE`). This mirrors the existing `PostgreContentJSON` approach: map the type to `CONTENT` + a `text/json` content handler. I have a working implementation and would be glad to open a PR - could a maintainer confirm the approach first?

- Per-attribute `valueManager=json` override - doesn't work: DBeaver only builds the JSON/Text viewers when the column's data kind is `CONTENT`/`STRING`, and an `OTHER` column can't be promoted that way.
- A driver-side change - clickhouse-jdbc v2 maps `JSON -> java.sql.Types.OTHER` deliberately (same as pgjdbc for json/jsonb; there is no `Types.JSON`). So the display fix belongs in DBeaver's ClickHouse extension, not the driver.

**Additional context**
- This is the display/edit gap - distinct from the earlier read-crash issues (e.g. #36753) that were fixed in the driver.
- Environment: DBeaver CE (devel), clickhouse-jdbc 0.9.x (v2 driver), ClickHouse 24.8+/26.x.
- Before/after screenshots below.

**Current behavior**
ClickHouse `JSON` column - flattened, unreadable, no JSON viewer:
ClickHouse JSON as flat map, no viewer

**With this change (prototype)**
The same column rendered as formatted JSON in the Value panel - automatically, with no driver property or manual configuration:
ClickHouse JSON formatted in the JSON viewer

bakhtiiartashbolotov:
This PR was generated with AI assistance (Claude Code).

**What** — ClickHouse native `JSON` columns were shown as `OTHER` with no JSON viewer (the value came back as a flattened, non-JSON `java.util.Map`). This maps them to `CONTENT` + a `text/json` content handler so they render in the JSON value panel and are editable — mirroring the existing PostgreSQL `json`/`jsonb` support (`PostgreContentJSON`).

**How**
- `ClickhouseDataSource.resolveDataKind`: `JSON` → `CONTENT` (table data editor path).
- `ClickhouseDataTypeCache`: `JSON` entry DATA_TYPE → `Types.CLOB` (SQL result-grid path).
- `ClickhouseValueHandlerProvider` + `plugin.xml`: route `JSON` to a new `ClickhouseJSONValueHandler`.
- New `ClickhouseJSONValueHandler` (extends `JDBCContentValueHandler`) + `ClickhouseContentJSON` (extends `JDBCContentChars`, `text/json`); editable, writes go through `ALTER TABLE … UPDATE`.
- `ClickhouseDataSource.openConnection`: enables `output_format_binary_write_json_as_string` through the driver's own default query settings (reflection — same idiom as the existing `getLastQueryId` call) so `getString()` returns canonical JSON; gated on the server version (≥ 24.10, where the setting was introduced — read from the already-open connection) so servers that predate it are never sent it.

**Scope** — only the bare `JSON` type; `Nullable(JSON)` / `Array(JSON)` / `JSON(...)` / `Dynamic` are out of scope (related: #40683).

**Known limitation** — manually overriding `output_format_binary_write_json_as_string` per-query via a SQL `SETTINGS` clause desyncs the clickhouse-jdbc reader (the driver binds its JSON reader at the connection level). No normal usage hits this.

**Tested** — CH 26.3 + jdbc 0.9.8: formatted in the Data tab and `SELECT`; edit → save applies via `ALTER UPDATE`; invalid JSON rejected (Code 117). CH 23.8 (lacks the setting): connects and queries normally — the probe skips it. Full Tycho build green; unit test added.

**Screenshots**

Current behavior — flattened, unreadable, no JSON viewer:

ClickHouse JSON as flat map, no viewer

With this change — formatted JSON in the value panel, automatically:

ClickHouse JSON formatted in the JSON viewer

Closes #41315


plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseDataSource.java
```diff
@@ -370,6 +376,60 @@ protected Connection openConnection(@NotNull DBRProgressMonitor monitor, @Nullab
...
+     */
+    private synchronized boolean isJsonStringSerializationSupported(@NotNull Connection connection) {
+        if (jsonStringSerializationSupported == null) {
+            try (Statement stmt = connection.createStatement();
+                 ResultSet resultSet = stmt.executeQuery(
+                     "SELECT count() FROM system.settings WHERE name = '" + ClickhouseConstants.SETTING_JSON_AS_STRING + "'")) {
+                jsonStringSerializationSupported = resultSet.next() && resultSet.getInt(1) > 0;
+            } catch (SQLException e) {
+                log.debug("Can't detect support for the " + ClickhouseConstants.SETTING_JSON_AS_STRING + " setting", e);
+                jsonStringSerializationSupported = false;
```
chernser: SQLException here can be for different reasons. For example, it may fail for some readonly profiles.
however from user perspective it will be no error just not rendered value. 
besides access to system.settings also can be restricted. 

This method does call anyway - so checking server version is the only right option here.

bakhtiiartashbolotov: Makes sense, thanks for catching that. The real problem with the probe was exactly what you said - if it throws (read-only profile, restricted system.settings), the user just gets an unrendered value and no error. So I dropped the query and gate on the server version instead: it only turns on for ClickHouse 24.10+ (where the setting was added), read straight from the open connection's metadata, so there's nothing that can be restricted or fail silently. Checked it against real 24.10 and 24.9 servers - 24.10 renders JSON, 24.9 falls back to the old plain view with no error. Updated in 18322da.

E1izabeth (dismissed):

ShadelessFox (dismissed):
Thanks! Passing to QA

uslss:
verified

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseDataSource.java
```diff
@@ -370,6 +376,57 @@ protected Connection openConnection(@NotNull DBRProgressMonitor monitor, @Nullab
...
+     * Makes the driver serialize {@code JSON} columns as canonical JSON strings, so that
+     * {@code getString()} returns valid JSON instead of a flattened {@code java.util.Map}. This is what
+     * lets the JSON content viewer/editor render those columns (see {@code ClickhouseJSONValueHandler}).
+     * <p>
+     * The {@code output_format_binary_write_json_as_string} setting is applied through the driver's own
+     * default query settings rather than a {@code SET} statement: a server-side {@code SET} changes the
+     * wire format without the driver knowing, which desynchronizes its binary reader. The setting is only
+     * applied when the server actually exposes it, so connecting to servers that predate it is unaffected.
+     */
+    private void enableJsonStringSerialization(@NotNull Connection connection) {
```
serge-rider, line 400: Isn't it too fragile and non-documented?
From doc I see that this option can be enabled through setting driver property `jdbc_read_json_as_string` (https://clickhouse.com/docs/integrations/language-clients/java/jdbc)

serge-rider (dismissed):

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseDataSource.java
bakhtiiartashbolotov: Thanks for the reply and pointer. I dug into `jdbc_read_json_as_string` and it turned out to be a doc mistake - the driver rejects it with `Unknown and unmapped config properties: [jdbc_read_json_as_string]` (tested on 0.9.8 and the latest 0.10.0-rc1), and it's nowhere in the clickhouse-java source, only in the docs prose.

The example right under that sentence actually uses the real setting, `output_format_binary_write_json_as_string`. I tried passing it as a plain connection property, but on servers older than 24.10 it fails every query with `Code: 115 UNKNOWN_SETTING`, since it's sent before we know the server version - that's why it's applied after connecting and gated on the version.

If you'd prefer, I can set it as a plain connection property instead, but it'd require ClickHouse 24.10+ and break older servers.

chernser: @bakhtiiartashbolotov 
This is server settings and should have `clickhouse_setting_` prefix  or wrapped with `ClientConfigProperties.serverSetting` (https://clickhouse.com/docs/integrations/language-clients/java/jdbc#connection-properties)

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/ClickhouseConstants.java
```diff
@@ -34,9 +34,17 @@ public class ClickhouseConstants {
     public static final String DATA_TYPE_ARRAY = "Array";
     public static final String DATA_TYPE_TUPLE = "Tuple";
     public static final String DATA_TYPE_MAP = "Map";
+    public static final String DATA_TYPE_JSON = "JSON";
     public static final String CLICKHOUSE_SETTING_SESSION_ID = "clickhouse_setting_session_id";
     public static final String CLICKHOUSE_SETTING_SESSION_TIMEOUT = "clickhouse_setting_session_timeout";
+    // Server setting that makes the driver serialize JSON columns as canonical JSON strings (so getString()
+    // returns valid JSON instead of a flattened java.util.Map). Applied through the driver's default query
+    // settings, see ClickhouseDataSource#enableJsonStringSerialization.
+    public static final String SETTING_JSON_AS_STRING = "output_format_binary_write_json_as_string";
```
chernser, line 43: it should be `clickhouse_setting_output_format_binary_write_json_as_string` if passed directly to connection properties.

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseDataSource.java
```diff
@@ -370,6 +376,57 @@ protected Connection openConnection(@NotNull DBRProgressMonitor monitor, @Nullab
...
+        if (!isJsonStringSerializationSupported(connection)) {
+            return;
+        }
+        try {
+            Object querySettings = BeanUtils.invokeObjectMethod(
```
chernser: this should be set as property on connection level - as it is only way to parse JSON. 

Also using reflection each time looks  is not optimal
- it can be a simple cast to `com.clickhouse.jdbc.ConnectionImpl`. ( from driver side we may be need interfaces for proprietary method)
- cache method reference as class is not changing here.

test/org.jkiss.dbeaver.ext.clickhouse.test/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseJSONContentTest.java
```diff
@@ -0,0 +1,48 @@
...
+package org.jkiss.dbeaver.ext.clickhouse.model;
+
+import org.jkiss.dbeaver.ext.clickhouse.model.data.ClickhouseContentJSON;
+import org.jkiss.dbeaver.model.data.DBDDisplayFormat;
+import org.jkiss.dbeaver.utils.MimeTypes;
+import org.jkiss.junit.DBeaverUnitTest;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class ClickhouseJSONContentTest extends DBeaverUnitTest {
```
chernser, line 26: Tests with failures and unsupported data types also needed - it will help to "document" agreed behavior and find issues with handling these scenarios.

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseDataSource.java
bakhtiiartashbolotov: @chernser
It already goes through serverSetting here - the code calls querySettings.serverSetting("output_format_binary_write_json_as_string", "1") (via reflection, since the plugin doesn't build against the driver). That method takes the un-prefixed name, and I confirmed it returns canonical JSON. The clickhouse_setting_ prefix / ClientConfigProperties.serverSetting is the raw connection-property path, which we're not using here (a connection property can't be version-gated, so it would break every query on servers older than 24.10).

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/ClickhouseConstants.java
bakhtiiartashbolotov: Right, if it were passed directly as a connection property it would need the clickhouse_setting_ prefix.
Here the constant is used with serverSetting(name, value), which takes the un-prefixed name, so the bare output_format_binary_write_json_as_string is intentional.

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseDataSource.java
bakhtiiartashbolotov: Cached the methods now, resolved once and reused (cfdc47a).

On "connection property is the only way to parse JSON", I tested it. As a connection property it works on 24.10 but returns Code 115 on 24.9 and breaks every query, not just JSON, because it's set before we know the server version and can't be gated. The post-connect setDefaultQuerySettings call parses JSON fine too (canonical on 24.10, skipped on 24.9), and it's the only option that stays safe on older servers. I also tried default_query_settings, jdbc_type_mappings, setClientInfo and a plain SET, none of them work (invalid JSON, ignored, or they desync the reader).

The cast to com.clickhouse.jdbc.ConnectionImpl won't compile: the plugin doesn't depend on the driver at build time (manifest only requires ext.generic and slf4j, nothing imports com.clickhouse), so any driver method has to be reflective, same as the existing getLastQueryId.

So I'd keep the reflection for now, or switch to a public driver method once clickhouse-java exposes one (you mentioned the driver might need interfaces for this, would that be on the table?). A connection property would break everyone on ClickHouse < 24.10, and a lot of the community still runs older versions.

If part of the concern is that this happens implicitly, I can also make it respect a user-set property so it's visible and overridable, and document it, while keeping the safe default. Would that help?

test/org.jkiss.dbeaver.ext.clickhouse.test/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseJSONContentTest.java
bakhtiiartashbolotov: Added tests for malformed and empty JSON content (cfdc47a). Invalid JSON on save is rejected server-side with Code 117, verified against 26.3 (needs a live server, so it's not in the unit tests).

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseDataSource.java
chernser: Go it! 
Thank you for explaining!

uslss:
verified (ce)

E1izabeth (approved):

ShadelessFox (approved):
Passing to QA

uslss:
verified
image

PR #41329, opened 2026-06-17
dbeaver/pro#9644 DataFilter thread safety

serge-rider:


E1izabeth (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/utils/RuntimeUtils.java
```diff
@@ -192,7 +192,7 @@ public static void pause(int ms) {
     }
 
     /**
-     * @deprecated consider using {@link DurationFormatter#format(Duration, DurationFormat)} instead
+     * Consider using {@link DurationFormatter#format(Duration, DurationFormat)} instead
```
ShadelessFox, line 195: It's marked deprecated for a reason, please revert this change

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/DBDDataFilter.java
```diff
@@ -55,70 +55,108 @@ public DBDDataFilter(@NotNull DBDDataFilter source) {
         this.useDisjunctiveNormalForm = source.useDisjunctiveNormalForm;
     }
 
+    public int getConstraintsCount() {
```
ShadelessFox, line 58: Maybe instead of marking everything as synchronized, we could use `CopyOnWriteArrayList`?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/DBDDataFilter.java
```diff
@@ -55,70 +55,108 @@ public DBDDataFilter(@NotNull DBDDataFilter source) {
...
+        synchronized (constraints) {
+            return constraints.size();
+        }
+    }
+
     @NotNull
-    public List<DBDAttributeConstraint> getConstraints() {
-        return constraints;
+    public DBDAttributeConstraint[] getConstraints() {
```
ShadelessFox, line 65: I don't really like this change. You could've done `List.copyOf` that does basically the same, plus the produced List is immutable.

PR #41337, opened 2026-06-18
Fix compiler log viewer colors in dark theme (#18597)

### LINKED ISSUE #18597: Messages in dark theme
### Description

Messages in dark mode are hard to read.

### DBeaver Version

Community edition 22.3.1.202212251901

### Operating System

macOS 13.0.1 (22A400)

### Database and driver

Oracle XE 18.0.0.0.0
ojdc 12.2.0.1

### Steps to reproduce

I'm using DBeaver in dark mode. When I recompile an Oracle package, I get the following output:

Bildschirm­foto 2022-12-26 um 12 55 33

The blue text on dark gray background is hard to read and should use a lighter color.

### Additional context

_No response_

anantgupta001:
Fixes #18597.

### Problem
In dark themes, compiler log messages in `ObjectCompilerLogViewer` were unreadable:
- The initial fix used hardcoded colors
- A subsequent attempt used `BaseThemeSettings.colorError` / `colorWarning`, but those are
  transaction background colors (not foreground), making error text barely visible in dark mode

### Fix
Introduced two new generic, reusable foreground color definitions:
- `org.jkiss.dbeaver.ui.general.color.warning.foreground`
- `org.jkiss.dbeaver.ui.general.color.error.foreground`

Changes:
- Added `COLOR_WARNING_FOREGROUND` and `COLOR_ERROR_FOREGROUND` constants to `BaseEditorColors`
- Added `colorWarningForeground` and `colorErrorForeground` fields to `BaseThemeSettings`
- Registered both colors as `colorDefinition` entries in `plugin.xml` with light theme defaults (`COLOR_DARK_YELLOW`, `COLOR_DARK_RED`)
- Added dark theme overrides in `e4-dark-sql-editor.css` (`230,200,80` and `255,120,120`)
- Added labels/descriptions in `bundle.properties`
- Updated `ObjectCompilerLogViewer` to use the new foreground colors

### Result
Warning and error messages in the compiler log are now clearly readable in both light and dark themes. The new color IDs are generic enough to be reused in other log viewers in the future.

### Screenshots
Screenshot 2026-07-09 160809


anantgupta001:
I wasn't able to run a local DBeaver instance due to PDE launch configuration issues, but the change is limited to color selection in ObjectCompilerLogViewer::log and was implemented based on the maintainer's guidance in issue #18597.

E1izabeth (dismissed):

E1izabeth:
Thanks for the contribution

ShadelessFox (dismissed):
I would prefer the colors not to be hardcoded at all and instead be defined via `colorDefinition`, but the proposed approach is better than nothing. Thanks, passing to QA :^)

ShadelessFox:
> I wasn't able to run a local DBeaver instance due to PDE launch configuration issues

Are you using Eclipse or IntelliJ? We don't maintain launch configuration for Eclipse anymore, unfortunately, so you'll have to stick with the latter. See https://github.com/dbeaver/dbeaver/wiki/Develop-in-IDEA

anantgupta001:
@ShadelessFox Took your suggestion and reworked it to use colorDefinition instead — added trace/warn/error entries in plugin.xml, labels in bundle.properties, and dark theme overrides in e4-dark-sql-editor.css. Pushed to the same branch. Also, switched to IntelliJ and got it running locally with the generator — verified the new colors show up correctly under Colors and Fonts. Thanks for the pointer!

evidence:
Screenshot 2026-06-20 061922

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/controls/ObjectCompilerLogViewer.java
```diff
@@ -106,20 +108,21 @@ protected void log(final int type, final Object message, final Throwable t)
             if (infoTable == null || infoTable.isDisposed()) {
                 return;
             }
-            int color = -1;
+            ITheme currentTheme = PlatformUI.getWorkbench().getThemeManager().getCurrentTheme();
+            org.eclipse.swt.graphics.Color themeColor = null;
             switch (type) {
                 case LOG_LEVEL_TRACE:
-                    color = SWT.COLOR_DARK_BLUE;
+                    themeColor = currentTheme.getColorRegistry().get("org.jkiss.dbeaver.sql.editor.color.compiler.trace");
```
ShadelessFox: Please see `org.jkiss.dbeaver.ui.controls.SQLEditorThemeSettings` and apply the same pattern for the new constants as well

anantgupta001: Done, added `compiler.trace/warn/error` constants to `ThemeConstants`, exposed them as `@ThemeColor` fields in `SQLEditorThemeSettings`, and updated `ObjectCompilerLogViewer` to use `SQLEditorThemeSettings.instance` instead of manual registry lookups. Pushed to the same branch. Thanks:)

ShadelessFox (approved):
Thanks!

anantgupta001:
> Thanks!

Could you please explain the typical process for getting a PR merged in DBeaver? I'm new to the open-source community, and this is my first contribution here, so I'd like to understand how things usually work.

From what I understand, a PR first goes through code review and approval by the maintainers. After that, does it go through QA/testing or any other validation steps before being merged?

I'd appreciate a brief overview of the workflow so I can better understand what to expect for this and future contributions.

ShadelessFox:
> From what I understand, a PR first goes through code review and approval by the maintainers. After that, does it go through QA/testing or any other validation steps before being merged?

Pretty much. Although it might take a long time to get merged because we have a lot of other PRs to verify as well :^)

ShadelessFox (changes requested):
After an internal discussion, we've concluded that we don't want to introduce new colors, since we already have them :^)

Please use these instead:
https://github.com/dbeaver/dbeaver/blob/9337fd34b2283fd648a1c41036b98bc2270ee98d/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/BaseThemeSettings.java#L49-L54

Also, null checks are not needed. Colors retrieved from `XXXThemeSettings` classes are guaranteed to never be null under normal conditions.

anantgupta001:
> After an internal discussion, we've concluded that we don't want to introduce new colors, since we already have them :^)
> 
> Please use these instead:
> 
> https://github.com/dbeaver/dbeaver/blob/9337fd34b2283fd648a1c41036b98bc2270ee98d/plugins/org.jkiss.dbeaver.ui/src/org/jkiss/dbeaver/ui/BaseThemeSettings.java#L49-L54
> 
> Also, null checks are not needed. Colors retrieved from `XXXThemeSettings` classes are guaranteed to never be null under normal conditions.

Addressed review feedback: removed new compiler color definitions and switched to BaseThemeSettings.colorWarning / colorError. Squashed to a single commit. Verified locally on Oracle dark theme


Screenshot 2026-07-08 115222
Screenshot 2026-07-08 115231
Screenshot 2026-07-08 115249
Screenshot 2026-07-08 115257

ShadelessFox (commented):
I believe your commit contains unwanted changes from another commits. Please either rebase/merge `devel`.

anantgupta001:
> I believe your commit contains unwanted changes from another commits. Please either rebase/merge `devel`.

done

ShadelessFox (approved):

E1izabeth (dismissed):

E1izabeth:
@anantgupta001 
I'm afraid the red color keeps being barely readable on your screens.
@ShadelessFox It doesn't look like a good solution. Could you suggest another approach?

ShadelessFox (changes requested):
I haven't noticed error messages on your screenshot because the color is too dark, so I assumed everything was okay 😅

Seems like exiting colors aren't really suitable. I'm really sorry to say this, but I would like you to revert changes to what you had before - separate color definitions, but please use a more generic name this time so we can reuse them in the future (something like `colorWarningForeground`/`org.jkiss.dbeaver.ui.general.warning.foreground`, etc.)

anantgupta001:
> I haven't noticed error messages on your screenshot because the color is too dark, so I assumed everything was okay 😅
> 
> Seems like exiting colors aren't really suitable. I'm really sorry to say this, but I would like you to revert changes to what you had before - separate color definitions, but please use a more generic name this time so we can reuse them in the future (something like `colorWarningForeground`/`org.jkiss.dbeaver.ui.general.warning.foreground`, etc.)

Done! Introduced two new generic foreground color definitions:
- `org.jkiss.dbeaver.ui.general.color.warning.foreground`
- `org.jkiss.dbeaver.ui.general.color.error.foreground`

Added constants to `BaseEditorColors`, wired them up in `BaseThemeSettings` as `colorWarningForeground`/`colorErrorForeground`, registered `colorDefinition` entries in `plugin.xml`, dark theme overrides in `e4-dark-sql-editor.css`, and updated `ObjectCompilerLogViewer` to use them.

Verified locally — warning and error messages are clearly readable in dark theme. Screenshot attached.

Screenshot 2026-07-09 160809

mariiachugunova:
@anantgupta001 Thanks for the update! I noticed one issue while testing.
Custom colors are preserved after restarting DBeaver, but the behavior when switching themes seems inconsistent.
Light → Dark → Light: the custom color is reset immediately after switching to the Dark theme.
Dark → Light → Dark: the custom color is only reset after switching back to the Dark theme.
Could you please check?

_Original color settings in Light theme_
before 2026-07-10 180233

_Changed color settings in Light theme_
after color change 2026-07-10 180303

_Switched to Dark theme — custom color is reset_
swith to the dark 2026-07-10 180348

_Switched back to Light theme — previous custom color is not restored_
swith light 2026-07-10 180441

anantgupta001:
> @anantgupta001 Thanks for the update! I noticed one issue while testing. Custom colors are preserved after restarting DBeaver, but the behavior when switching themes seems inconsistent. Light → Dark → Light: the custom color is reset immediately after switching to the Dark theme. Dark → Light → Dark: the custom color is only reset after switching back to the Dark theme. Could you please check?
> 
> _Original color settings in Light theme_ before 2026-07-10 180233
> 
> _Changed color settings in Light theme_ after color change 2026-07-10 180303
> 
> _Switched to Dark theme — custom color is reset_ swith to the dark 2026-07-10 180348
> 
> _Switched back to Light theme — previous custom color is not restored_ swith light 2026-07-10 180441

Thanks — I can reproduce the theme-switch wipe. Customs surviving restart but not Light↔Dark matches Eclipse re-applying dark CSS preferences on theme change.

I’ll verify the same steps with an existing SQL editor color (e.g. Keywords). If that behaves the same, this looks pre-existing for CSS-overridden colors, not specific to the new warning/error foreground definitions.

Separately I can move these general colorDefinitions into org.jkiss.dbeaver.core + e4-dark_dbeaver_prefstyle.css for consistency. Please confirm whether live theme-switch persistence is required for this PR or can be tracked separately.

E1izabeth:
@mariiachugunova @anantgupta001 I believe we can fix live theme switching separately as the main aim of this PR (to fix colors) is reached

mariiachugunova:

IvanGorshechnikov (approved):

ShadelessFox (approved):

PR #41340, opened 2026-06-19
Greenplum 6: avoid OOM when listing schema tables (replace pg_partitions join)

### LINKED ISSUE #41280: Greenplum 6: loading schema tables fails with OOM (pg_partitions join)
### Summary
On **Greenplum 6** (resource groups enabled), expanding a schema's **Tables** node fails to load the table list with:

```
SQL Error [53500]: ERROR: Out of memory
Detail: Resource group memory limit reached
```

### What actually fails
The error is **not** caused by the table data — it is raised by the **metadata query** that the Greenplum driver builds in `GreenplumSchema.GreenplumTableCache.prepareLookupStatement(...)`. On Greenplum 6 (`before7version` branch) that query joins the heavy `pg_catalog.pg_partitions` view:

```sql
...
LEFT OUTER JOIN pg_catalog.pg_partitions p
    on c.relname = p.partitiontablename and ns.nspname = p.schemaname
WHERE c.relnamespace = ? AND c.relkind not in ('i','c')
```

### Why it happens — an unoptimized query, not a server misconfiguration
`pg_catalog.pg_partitions` is a view that contains window functions (`rank() OVER (...)`, `min() OVER (...)`). Because of them the outer filter (by schema / relation) **cannot be pushed down**, so Greenplum must build the **entire** `pg_partitions` set for **all** partitions in the whole database — calling `pg_get_partition_rule_def(...)` / `pg_get_expr(...)` for every partition rule — *before* the join and the `WHERE` are applied. On databases with many partitions this blows past the resource group memory limit and the metadata query aborts.

Proof that it's the query, not the data:
- Simply removing the `pg_partitions` join makes the query return in **~1 second**.

So the root cause is the metadata query itself being unnecessarily expensive.

### Why "just raise the resource group limit" is not the answer
On a shared production cluster the resource group configuration **cannot be changed** just to let a GUI client browse a schema — it affects every workload and every user, and is owned by the DBA, not the client. A metadata query issued by an SQL tool must stay cheap regardless of cluster sizing. The correct fix is to **optimize the driver's query**, not to relax server-side memory limits.

### Steps to reproduce
1. Connect to a Greenplum 6 database with many partitioned tables (resource groups enabled).
2. Expand any schema → **Tables**.
3. The list fails with "Out of memory / Resource group memory limit reached".

### Proposed fix
The driver only needs two things from `pg_partitions`: whether a relation is a partition child (`partitiontablename`, used as a not-empty flag in `isPartitionTableRow`) and the partition boundary (`partition_expr`). Both can be read directly from `pg_catalog.pg_partition_rule` by `oid`, without the view, its window functions, or the full-database materialization:

```sql
-- SELECT list (before7version branch):
case when p.parchildrelid is not null then c.relname else null end as partitiontablename,
pg_catalog.pg_get_partition_rule_def(p.oid, true) as partition_expr,

-- join (before7version branch):
LEFT OUTER JOIN pg_catalog.pg_partition_rule p
    on p.parchildrelid = c.oid
```

Notes:
- `pg_partition_rule` and `pg_get_partition_rule_def(oid, bool)` are exactly what `pg_partitions` uses internally and are available on GP4/5/6.
- Output column names are preserved, so `isPartitionTableRow(...)` and the rest of the model keep working unchanged.
- The join by `oid` is also more correct than the previous name-based join, which could collide across schemas.
- The GP7 / native-partitioning branch is unaffected.

With this change the "Tables" node loads in ~1s instead of failing with OOM.

### DBeaver Version

Community Edition 26.0.5.

### Operating System

Windows 10

### Database and driver

Greenplum 6
org.jkiss.dbeaver.ext.greenplum_1.0.180.202605171705.jar

### Steps to reproduce

1. Connect to a Greenplum 6 database with many partitioned tables (resource groups enabled).
2. Expand any schema → **Tables**.
3. The list fails with "Out of memory / Resource group memory limit reached".

### Additional context

_No response_

vaefremov95:
### Summary

Fixes #41280.

Expanding the Tables node of a schema on Greenplum 6 can fail with
`53500 Out of memory: Resource group memory limit reached`. The failure comes
from the metadata lookup query built in
`GreenplumSchema.GreenplumTableCache.prepareLookupStatement` (the
`before7version` branch), not from the table data itself.

### Root cause

The query `LEFT OUTER JOIN`s the `pg_catalog.pg_partitions` view. That view
uses window functions, which prevent the planner from pushing the schema
(`c.relnamespace = ?`) and relation (`relname = ?`) predicates down into it.
As a result Greenplum materializes the full partition set of the entire
database before filtering. On clusters with many partitions and resource
groups enabled this exhausts the group memory limit.

Raising the resource-group limit is not a viable workaround, since it affects
every workload on the cluster.

### Fix

Read the two values the cache actually needs directly from the base catalog
table `pg_catalog.pg_partition_rule`, joined by oid, instead of going through
the `pg_partitions` view:

- `partitiontablename` ← `c.relname` when a matching partition rule exists
- `partition_expr` ← `pg_get_partition_rule_def(p.oid, true)`

This removes the window functions, so the existing schema/relation predicates
push down normally. Output column names are unchanged, `isPartitionTableRow`
keeps working (`partitiontablename` is non-null exactly for partition child
relations), and the join stays `LEFT OUTER`, so non-partitioned tables are
unaffected.

### Testing

Verified on a production Greenplum 6 cluster with resource groups enabled
(DBeaver CE 26.0.5):

- Before: expanding the Tables node fails with the OOM error above.
- After: the same query returns in ~1s and the table list loads normally.
- Spot-checked that partitioned tables still show their child partitions and
  partition expressions.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumSchema.java
+++ b/plugins/org.jkiss.dbeaver.ext.greenplum/src/org/jkiss/dbeaver/ext/greenplum/model/GreenplumSchema.java
@@ -114,7 +114,7 @@ public class GreenplumSchema extends PostgreSchema {
             boolean supportsRelStorageColumn = dataSource.isServerSupportsRelstorageColumn(session);
             before7version = !dataSource.isGreenplumVersionAtLeast(7, 0);
             String sqlQuery = "SELECT c.oid,c.*,d.description,\n" +
-                (before7version ? "p.partitiontablename,p.partitionboundary as partition_expr," :
+                (before7version ? "case when p.parchildrelid is not null then c.relname else null end as partitiontablename,pg_catalog.pg_get_partition_rule_def(p.oid, true) as partition_expr," :
                     "pg_catalog.pg_get_expr(c.relpartbound, c.oid) as partition_expr, pg_catalog.pg_get_partkeydef(c.oid) as partition_key,") +
                 (hasAccessToExttable ? "CASE WHEN x." + uriLocationColumn + " IS NOT NULL THEN array_to_string(x." + uriLocationColumn +
                     ", ',') ELSE '' END AS urilocation,\n" +
@@ -137,8 +137,7 @@ public class GreenplumSchema extends PostgreSchema {
                 "INNER JOIN pg_catalog.pg_namespace ns\n\ton ns.oid = c.relnamespace\n" +
                 "LEFT OUTER JOIN pg_catalog.pg_description d\n\tON d.objoid=c.oid AND d.objsubid=0\n" +
                 (hasAccessToExttable ? "LEFT OUTER JOIN pg_catalog.pg_exttable x\n\ton x.reloid = c.oid\n" : "") +
-                (before7version ? "LEFT OUTER JOIN pg_catalog.pg_partitions p\n\ton c.relname = p.partitiontablename " +
-                    "and ns.nspname = p.schemaname\n" : "") +
+                    (before7version ? "LEFT OUTER JOIN pg_catalog.pg_partition_rule p\n\ton p.parchildrelid = c.oid\n" : "") +
                 (before7version ? "" : "\nLEFT JOIN pg_catalog.pg_am pa ON pa.oid = c.relam") +
                 "\nWHERE c.relnamespace= ? AND c.relkind not in ('i','c') " +
                 (object == null && objectName == null ? "" : " AND relname=?");


IvanGorshechnikov (approved):

IvanGorshechnikov:
Thanks for the contribution!

We will review your PR and let you know

Destrolaric (approved):

SolovevAnton (approved):

vaefremov95:
Hi! Just checking on the status of this PR. Is there anything else needed from my side, or is it just waiting to be merged? Thanks!

uslss:
verified

PR #41341, opened 2026-06-19
#35817 Fix Informix procedure navigation

### LINKED ISSUE #35817: Informix - open declaration for procedures fail
### Description

When try to open the object declaration using the F4 shortcut or ctrl+mouse click over the **procedure** name , they open the wrong object properties.
Always open the first object on the procedure tree list 

### DBeaver Version

CE 24.2.1.202409221725

### Operating System

Windows 10

### Database and driver

IBM Informix 4.50.10.1

### Steps to reproduce

1. put the cursor over the proc name and press f4 OR move the mouse cursor over the name holding ctrl and click

image

2. They always open the first proc in the tree list , here is the internal database proc called addtoset2_combine 
 
image

### Additional context

_No response_

EastLord:
Closes: #35817

### Implementation

This fixes Informix procedure navigation by avoiding the generic JDBC `SPECIFIC_NAME` lookup path for procedures.

Informix now loads procedures from `informix.sysprocedures` and uses `procid` as the stable `GenericProcedure` unique name. SQL editor object lookup also resolves procedure references through `sysprocedures.procid`, so overloaded or driver-reported procedures no longer depend on invalid `SPECIFIC_NAME` values such as `"null"`.

The change is scoped to the Informix extension. Table and schema lookup continue to use the generic structure assistant behavior.

after fix:

https://github.com/user-attachments/assets/c193fbb0-d74b-4e25-a464-6ba72689b1d8


E1izabeth (approved):

EastLord:
I reproduced the issue using the IBM Informix official Docker installation guide.

One additional thing I noticed is that DBeaver still references `com.ibm.informix:jdbc:RELEASE[4.50.10.1]` as the default Informix JDBC driver. This seems older than the driver used with the current Informix Docker setup.

Do you think we should update the default driver version, or keep this PR focused only on the procedure metadata/navigation fix?

ShadelessFox:
> Do you think we should update the default driver version, or keep this PR focused only on the procedure metadata/navigation fix?

The driver should be updated in a separate PR if it's really desired. Otherwise, the testing scope would increase drastically because it would require performing regression testing.

EastLord:
> > Do you think we should update the default driver version, or keep this PR focused only on the procedure metadata/navigation fix?
> 
> The driver should be updated in a separate PR if it's really desired. Otherwise, the testing scope would increase drastically because it would require performing regression testing.

@ShadelessFox Makes sense. I will keep this PR focused on the Informix procedure metadata/navigation fix and leave the default JDBC driver version unchanged.

ShadelessFox (approved):
Thanks, passing to QA

uslss:
verified

EastLord:
cc @serge-rider

E1izabeth:
Unfortunately, we can't accept the changes you provided as they conflict with the PRO editions - we've already supported it there.

EastLord:
> Unfortunately, we can't accept the changes you provided as they conflict with the PRO editions - we've already supported it there.

OK

amivanovano:
So my 8 years expectations have been in vain

E1izabeth:
@amivanovano you can join early access program and use this feature for free https://dbeaver.com/docs/dbeaver/Early-Access-Program-license/

PR #41347, opened 2026-06-20
#41346 Fix object search to match partial names

### LINKED ISSUE #41346: [CUBRID] Object search returns no results for partial name
### Description

On CUBRID, searching for a databasae object (Goto Object / metadata search) by a partial name returns no results.
`CubridStructureAssistant` runs `... WHERE class_name = ?` and binds the search text directly. DBeaver passes a wildcard pattern (for example `emp%`), so using `=` instead of `LIKE` means the patterns never matches.

### DBeaver Version

26.1.0.202605311718

### Operating System

macOS 26.5.1(25F80)

### Database and driver

CUBRID, official CUBRID JDBC driver

### Steps to reproduce

1. Connect to a CUBRID database.
2. Open the object search (Goto Object).
3. Type part of a table name. No results are shown.

### Additional context

Other extensions (Generic, MySQL, Oracle, PostgreSQL) use `LIKE` for thie search.
CUBRID is the only one using `=`.

Srltas:
Fixes #41346

The CUBRID object search used an exact match (`WHERE class_name = ?`) while the search passes a wildcard pattern, so partial-name returned no results. Changed the query to use `LIKE`.

It also closes the statement and result set in the same method with try-with-resources (they were left open on each search). If you'd prefer to keep that out of this PR, let me know and I'll remove it.


E1izabeth (approved):

ShadelessFox (approved):
Thanks, passing to QA

Issue #16364, opened 2022-04-29
Background color for even/odd rows


hoppfrosch:
#### System information: 
- Win 10 Ultimate
- DBeaver 22.0.3

#### Describe the problem you're observing:
After trying out "dark theme" and switching back to "classic" or "light" theme, the background color of the odd rows in result view remained black. Also the basic background color of the result view remained dark.

**Expected behaviour**
As I want true "light" colors again within the result grid,  the themes should be able to set the alternating colors for odd/even rows. Or at least there should be a possibility for users to define the colors of odd/even rows in result grid.

**What I tried, to solve the problem**
* Searched within configuration dialog for setting the color manually, but failed.
* Switching theme from "dark" to "light" (or "classic")(and vice versa) does not change the color from dark to light again for odd
rows ...

Is there a solution for setting the bg color for odd rows? Or have I overseen somnething?

LonwoLonwo:
Hello @hoppfrosch 

Did you restart DBeaver after the thema switching?

uslss:
There have been no updates on the issue for a long time.
If it is still actual - feel free to reopen it with additional info

hoppfrosch:
Sorry - Problem still exists.

For sure I restarted DBeaver after switching - bg color of odd rows is still black, which is annoying (I use dbeaver frequently - and in 2 months I restarted it quite often ;-))

Dont know, which additional info to give beside above info that would help you

LonwoLonwo:
Maybe you can provide a video of this issue?
We can't reproduce your issue for now.

hoppfrosch:
Sorrry for coming back too late - just for information: 

DBeaver 22.2.3 does NOT show this behaviour anymore - after frist tests, switching between  dark-light-default colorscheme works correctly.  :thumbsup:

PR #41349, opened 2026-06-20
Refresh UI colors on theme switch (#16364, #41250)

### LINKED ISSUE #41250: Theme remains dark for grid and sql editor
### Description

I used the dark theme for a while, but I can't get used to it so I switched the theme back to Light.
But my grid and SQL editor remain to use the dark theme.

I seem not to be able to reset the complete theme from dark to light again, so it smells like a bug, but is it?

### DBeaver Version

Community Edition Version 26.1.0.202605311718

### Operating System

Windows 11 

### Database and driver

Not related

### Steps to reproduce

When you are on a dark theme:
1. Go to Window \ Preferences \User Interface \ Apperance
2. Select theme Light
3. Restart application

result:
Main UI is light except for the datagrid and editors

Image

Image

Image

### Additional context

My other appearance settings,

Image

Also tried to change color and font theme, but without effect.
Also disabled theming, restart and enable theming again dit not have any effect

anantgupta001:
fix: refresh stale colors on theme switch in folder tab list and result grid

## Problem

Switching themes (e.g. Dark -> Light -> Dark) leaves several UI
elements stuck on the previously active theme's colors instead of
updating to match the new one:

- The object editor's left-hand page list (Columns / Constraints /
  Partitions / etc.) keeps the selection highlight from whichever
  theme was active when that widget was first created.
- Result set grids retain stale default cell background/foreground
  colors after a theme switch, until something unrelated happens to
  trigger a repaint with the correct colors.

Related: #16364, #41250

## Root cause

1. `TabbedFolderList` computes all of its colors once, in
   `initColours()`, called only from the constructor. The class never
   listens for theme-change events, so colors are frozen at whatever
   theme was active when the widget was constructed.

2. `SpreadsheetPresentation.applyThemeSettings(ITheme)` is the proper
   theme-refresh hook invoked on theme change, and it does reset most
   colors (fonts, line colors, header selection background) - but it
   never clears the lazily-cached `backgroundDefault` /
   `foregroundDefault` fields used by `getBackground()` /
   `getForeground()`. Those fields were previously only cleared as a
   side effect of an unrelated `LightGrid#setForeground()` call,
   never as part of the actual theme-switch path.

## Fix

- `TabbedFolderList`: register an `IPropertyChangeListener` on the
  workbench `IThemeManager` in the constructor. On
  `IThemeManager.CHANGE_CURRENT_THEME`, re-run `initColours()` and
  `redraw()`. The listener is removed in the existing dispose
  listener to avoid leaking it.

- `SpreadsheetPresentation`: reset `backgroundDefault` and
  `foregroundDefault` to `null` inside `applyThemeSettings()`, so
  they get recomputed against the new theme on next paint - mirroring
  the existing (but differently-triggered) `resetColors()` reset
  pattern already used elsewhere in this class. These two fields are
  not disposed here, since they hold shared colors obtained from the
  controller rather than colors owned/allocated by this class.


IvanGorshechnikov (approved):

IvanGorshechnikov:
Thanks for the contribution!

We will review your PR and let you know

Destrolaric (approved):

PR #41353, opened 2026-06-21
dbeaver/pro#9662 Entity editor label background color fix

serge-rider:


plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/properties/TabbedFolderPageForm.java
```diff
@@ -218,10 +215,7 @@ private void refreshProperties(@Nullable Runnable afterRefresh) {
 
             // Create edit forms
             for (DBPPropertyDescriptor prop : sortedProps) {
-                var placeholder = CompositeFactory.newComposite(SWT.NONE)
-                    .layoutData(new ColumnLayoutData())
-                    .layout(GridLayoutFactory.fillDefaults().numColumns(2).create())
-                    .create(propertiesGroup);
+                Composite placeholder = UIUtils.createComposite(propertiesGroup, 2);
```
ShadelessFox: No, this is wrong? It has special layout data to dynamically resize columns depending on available space, and its own layout is also special, using fill by default, which `UIUtils.createComposite` doesn't do.

ShadelessFox (changes requested):

ShadelessFox (approved):

PR #41357, opened 2026-06-22
dbeaver/pro#8899 use same model for updating result set

yagudin10:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/data/resultset/ISmartTransactionManager.java
```diff
@@ -15,7 +15,7 @@
  * limitations under the License.
  */
 
-package org.jkiss.dbeaver.ui;
+package org.jkiss.dbeaver.model.data.resultset;
```
serge-rider, line 18: Rename it into DBCSmartTransactionManager

serge-rider (changes requested):

serge-rider:
@codex review

PR #41363, opened 2026-06-23
dbeaver/pro#9649 Fix TipOfTheDay layout

serge-rider:


IvanGorshechnikov:
Veirfied DBeaver UE

Image

PR #41366, opened 2026-06-23
dbeaver/dbeaver#41362 Handle and quote text/* content types 

### LINKED ISSUE #41362: 'Generate SQL (INSERT and UPDATE)' returns null::jsonb even when the JSONB column has a value.
### Description

The issue happens in Community Edition Version 26.1.1.202606211757 -- not in the previous versions such as Community Edition Version 26.1.0.202605311718

### DBeaver Version

Community Edition Version 26.1.1.202606211757

### Operating System

Windows 11 Pro

### Database and driver

1. Database name and version: PostgreSQL 16.11
2. Driver name and version: PostgreSQL JDBC Driver 42.7.11

### Steps to reproduce

1. Right-click on a table record that contains a JSONB value.  
2. Select the "Generate SQL" menu.  
3. Select "INSERT" or "UPDATE".  
4. In the SQL preview, "null::jsonb" will be generated.

### Additional context

_No response_

HellAmbro:
Closes #41362

This affect also the import from CSV. 

The issue occurs because dataKind is set to CONTENT, while contentType is text/json (or text/xml, etc.). As a result, the current condition does not match and returns directly  `SQLConstants.NULL_VALUE`;.
At this point, it's unclear whether the additional check for "text" is still required, since it seems redundant given the CONTENT handling.
An alternative approach would be to explicitly quote all values when dataKind = CONTENT, ensuring consistent handling regardless of the contentType (e.g. JSON, XML, etc.).
Screenshot 2026-06-23 130322


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/SQLUtils.java
```diff
@@ -749,6 +749,8 @@ private static String convertValueToSQLFormat(
                     String contentType = contentValue.getContentType();
                     if (!contentType.startsWith("text") && CommonUtils.isNotEmpty(strValue)) {
```
HellAmbro, line 750: Why this startsWith("text") is required?

IvanGorshechnikov (approved):

Destrolaric (approved):

mariiachugunova:
Verified for DBeaver CE Version 26.1.2 on Windows 11 pro

PR #41375, opened 2026-06-24
dbeaver/pro#9526 handle network profiles changes

yagudin10:


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/net/DBWNetworkProfile.java
```diff
@@ -198,4 +198,35 @@ private void loadFromLegacySecret(DBSSecretController secretController) throws D
         }
     }
 
+    public boolean equalConfigurations(@Nullable DBWNetworkProfile other) {
+        if (this == other) {
```
serge-rider: Looks like AI generated. Why not use `||`?

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceRegistry.java
```diff
@@ -1131,9 +1139,17 @@ public ProjectNetworkProfileManager(@NotNull DBPProject project) {
...
+        @Override
+        public void reloadProfiles() {
+            project.getDataSourceRegistry().refreshConfig();
+        }
+
         @Override
         public void saveSettings() {
-            project.getDataSourceRegistry().flushConfig();
+            // TODO: find a better way to save network profiles.
+            if (project.getDataSourceRegistry() instanceof DataSourcePersistentRegistry dspr) {
```
serge-rider: It is double save, flushConfig already saved datasources. Why do we need it?

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceParseResults.java
```diff
@@ -29,4 +30,7 @@ public class DataSourceParseResults {
     public final Set<DBPDataSourceFolder> addedFolders = new LinkedHashSet<>();
     public final Set<DBPDataSourceFolder> updatedFolders = new LinkedHashSet<>();
     public final Set<DBPDataSourceFolder> removedFolders = new LinkedHashSet<>();
+    public final Set<DBWNetworkProfile> addedProfiles = new LinkedHashSet<>();
```
serge-rider: Isn't it too much?
Profiles are rarely changed. Can we just forcibly reload ALL profiles on datasources refresh (e.g. cleanup them before DS load)?

plugins/org.jkiss.dbeaver.model.event/META-INF/MANIFEST.MF
```diff
@@ -11,6 +11,7 @@ Export-Package: org.jkiss.dbeaver.model.websocket,
  org.jkiss.dbeaver.model.websocket.event,
  org.jkiss.dbeaver.model.websocket.event.datasource,
  org.jkiss.dbeaver.model.websocket.event.permissions,
+ org.jkiss.dbeaver.model.websocket.event.profile,
```
serge-rider: Lets keep profile events in .datasource package

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceRegistry.java
yagudin10: ?

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceParseResults.java
yagudin10: Profiles won't be deleted because we don't do in in serializer

PR #41382, opened 2026-06-25
#41325 [CUBRID] Distinguish user triggers and table triggers in the navigator

### LINKED ISSUE #41325: [CUBRID] Distinguish user triggers and table triggers in the navigator
### **Is your feature request related to a problem? Please describe.**
CUBRID has two kinds of triggers.
- Table triggers, which fire on `INSERT`/`UPDATE`/`DELETE` of a specific table (ON `<table>`).
- User triggers, which fire on `COMMIT`/`ROLLBACK` and are not bound to any table.

In the current CUBRID navigator, table triggers and user triggers are all mixed together in a single list without any distinction by type. As a result, `COMMIT`/`ROLLBACK` triggers look identical to table triggers in DBeaver, making it confusing to see which is which.

### **Describe the solution you'd like**
Show the two trigger types under separate nodes, the same way Oracle and SQL Server already do (Oracle uses Table Triggers and Schema Triggers, SQL Server uses Table Triggers and Database Triggers).
- A Table Triggers node for DML(`INSERT`/`UPDATE`/`DELETE`) triggers.
- A User Triggers node for `COMMIT`/`ROLLBACK` triggers.

Each node lists only its own type. If there are none, the node stays empty, which is consistent with Oracle and SQL Server. Node labels should follow CUBRID terminology, so "Table Triggers" and "User Triggers".

### **Describe alternatives you've considered**
A single unified "Triggers" node, like MySQL and MariaDB. This was rejected because MySQL only has table triggers, so a single node is enough there. CUBRID has two distinct categories, so single node would hide the distinction and leave all triggers cluttered without any proper classification.

### **Additional context**
The two types can be told apart reliably from the CUBRID catalog. A table trigger has a target table and a user trigger does not. CUBRID enforces this, because creating a `COMMIT`/`ROLLBACK` trigger with a target table is rejected with the following error.
```
ERROR: Inappropriate trigger target class "dba.trg_src".
```
So the presence or absence of a target table classifies a trigger without ambiguity.
Notes:
- The Triggers folder is only visible to DBA group users, so testing must be done with a DBA account.
- A deferred table trigger (for example `DEFERRED UPDATE ON t ...`) fires at commit time but still has a target table, so it is still a table trigger and not a user trigger.

Repro and test triggers (run as DBA):
```sql
CREATE TABLE trg_src (id INTEGER PRIMARY KEY, name VARCHAR(100), qty INTEGER);

-- Table trigger
CREATE TRIGGER t_table AFTER INSERT ON trg_src EXECUTE PRINT 'table trigger';

-- User triggers
CREATE TRIGGER t_commit   STATUS INACTIVE AFTER COMMIT   EXECUTE PRINT 'user trigger';
CREATE TRIGGER t_rollback STATUS INACTIVE AFTER ROLLBACK EXECUTE PRINT 'user trigger';
```
Expected result: `t_table` appears under Table Triggers, and `t_commit` and `t_rollback` appear under User Triggers.

CUBRID trigger manual: https://www.cubrid.org/manual/en/11.4/sql/trigger.html

This feature will be implemented by @longhaseng52.

longhaseng52:
### Implementation

- Added two new folders: **Table Triggers** and **User Triggers**.
- Separated triggers into the appropriate folder based on their type.
- Hid the **Target Table** and **Target Column** fields for **User Triggers**, as these fields are not applicable.

Issue reference: https://github.com/dbeaver/dbeaver/issues/41325


plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridUser.java
```diff
@@ -175,6 +175,35 @@ public List<GenericTableIndex> getIndexes(@NotNull DBRProgressMonitor monitor) t
         return indexes;
     }
 
+    public List<CubridTrigger> getUserTriggers(@NotNull DBRProgressMonitor monitor) throws DBException {
+        List<CubridTrigger> userTriggers = new ArrayList<>();
+        for (GenericTrigger<?> trigger : super.getTriggers(monitor)) {
+            if (trigger instanceof CubridTrigger) {
+                CubridTrigger cubridTrigger = (CubridTrigger) trigger;
```
SolovevAnton:
```suggestion
            if (trigger instanceof CubridTrigger cubridTrigger) {             
```

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridTrigger.java
```diff
@@ -320,4 +321,11 @@ public Object[] getPossibleValues(CubridTrigger object) {
         }
     }
 
+    public static class TargetTableValidator implements IPropertyValueValidator<CubridTrigger, Object> {
+        @Override
+        public boolean isValidValue(@NotNull CubridTrigger object, @Nullable Object value) throws IllegalArgumentException {
+            String event = object.getEvent();
+            return !"COMMIT".equals(event) && !"ROLLBACK".equals(event);
```
SolovevAnton: Let's move this method and its negation to the CubridUtils helper class. Here and all other occurrences

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridUser.java
```diff
@@ -175,6 +175,35 @@ public List<GenericTableIndex> getIndexes(@NotNull DBRProgressMonitor monitor) t
...
+        }
+        return userTriggers;
+    }
+
+    @Override
+    public List<CubridTrigger> getTableTriggers(@NotNull DBRProgressMonitor monitor) throws DBException {
+        List<CubridTrigger> tableTriggers = new ArrayList<>();
+        for (GenericTrigger<?> trigger : super.getTriggers(monitor)) {
+            if (trigger instanceof CubridTrigger) {
+                CubridTrigger cubridTrigger = (CubridTrigger) trigger;
```
SolovevAnton:
```suggestion
            if (trigger instanceof CubridTrigger cubridTrigger) {
```

SolovevAnton (changes requested):
Thank you for your contribution. Some nitpicks

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ext.cubrid/src/org/jkiss/dbeaver/ext/cubrid/model/CubridTrigger.java
longhaseng52: Thank you for the suggestion! I've done extracted the logic and its negation into a shared static method to eliminate the duplication.

SolovevAnton (approved):
Thank you for your contribution

IvanGorshechnikov (approved):

PR #41391, opened 2026-06-26
Set AppUserModelID to prevent duplicate taskbar icon (#40782)

### LINKED ISSUE #40782: When DBeaver application is pinned to taskbar, duplicate icon is getting created while launching the application.
### Description

When DBeaver application is pinned to taskbar, duplicate icon is getting created while launching the application.

### DBeaver Version

25.3.4

### Operating System

Windows 11

### Database and driver

_No response_

### Steps to reproduce

1. Install latest OS on your machine
2. Install the latest version of DBeaver application from [HERE](https://dbeaver.io/download/)
3. Launch the Application.
4. Pin the application to taskbar 
5. Observe Duplicate icon is getting created while launching the Application

### Additional context

_No response_

anantgupta001:
Fixes #40782

## Problem

When DBeaver is pinned to the Windows taskbar, launching the application creates a duplicate taskbar icon instead of reusing the pinned one.

This happens because the Equinox launcher often starts the UI in a separate `javaw.exe` process while the pin targets `dbeaver.exe`. Without a shared process identity / AppUserModelID (AUMID), Windows treats them as different applications and shows a second taskbar button.

## Solution

Two complementary changes:

1. **In-process JVM on Windows**  
   Product configs now pass:
   `-vm jre/bin/server/jvm.dll`
   
   so the JVM is loaded inside `dbeaver.exe` instead of spawning a separate `javaw` UI process. Updated:
- `product/community/DBeaver.product`
- `product/community-msstore/DBeaverCommunityMSStore.product`
- `product/testing/DBeaverUnitTest.product`

2. **Explicit AppUserModelID before first SWT Display**  
On Windows, `DBeaverApplication.start()` sets AUMID `DBeaverCorp.DBeaverCE` before any workspace dialog / Display is created.  
The call is delegated via reflection to `WindowsAppUserModelId` in the Windows SWT fragment (`org.jkiss.dbeaver.ui.swt.windows`), which invokes `SetCurrentProcessExplicitAppUserModelID` with a **null-terminated** buffer (required by the Win32 API).  
Failures are logged at debug level and do not affect startup.

## Why reflection?

`DBeaverApplication` is platform-independent. Direct references to Windows-only APIs would break non-Windows builds. Reflection through `Display.class.getClassLoader()` resolves the Windows SWT fragment correctly in OSGi while keeping compile-time platform neutrality.

## Testing

**Automated**
- Packaged `dbeaver.exe` launches with UI hosted by `dbeaver.exe` (not a separate `javaw` window).
- `dbeaver.ini` contains `-vm jre/bin/server/jvm.dll`.

**Manual**
1. Unpin any existing DBeaver icons.
2. Launch packaged `dbeaver.exe` (not IntelliJ Run/Debug).
3. Pin the running taskbar button.
4. Close DBeaver completely.
5. Relaunch from the pinned icon.

**Before:** a duplicate taskbar icon appears.  
**After:** the pinned icon is reused; only one DBeaver icon is shown.

**evidence:**
appIconFix

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
+++ b/plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
@@ -185,6 +185,11 @@ public class DBeaverApplication extends DesktopApplicationImpl implements DBPApp
         // hide standard Eclipse exit message if exit code is not OK (otherwise it may be confusing)
         System.setProperty(ECLIPSE_EXIT_DATA, "");
 
+        // Must run before any SWT Display / dialog is created (workspace lock UI, splash, etc.)
+        if (RuntimeUtils.isWindows()) {
+            setWindowsAppUserModelID();
+        }
+
         var args = preprocessCommandLine();
         Location instanceLoc = Platform.getInstanceLocation();
         Path defaultHomePath = getDefaultInstanceLocation();
@@ -294,10 +299,6 @@ public class DBeaverApplication extends DesktopApplicationImpl implements DBPApp
 
         // Write version info
         writeWorkspaceInfo();
-
-        if (RuntimeUtils.isWindows()) {
-            setWindowsAppUserModelID();
-        }
 
         // Initialize display early
         // It sets main windows name and images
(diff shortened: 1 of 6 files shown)


Phillipus:
`org.eclipse.swt.internal.win32.OS.SetCurrentProcessExplicitAppUserModelID` won't compile on non-Windows OS. `DBeaverApplication` is OS agnostic. You might be able to do it with Java reflection.

anantgupta001:
> `org.eclipse.swt.internal.win32.OS.SetCurrentProcessExplicitAppUserModelID` won't compile on non-Windows OS. `DBeaverApplication` is OS agnostic. You might be able to do it with Java reflection.

Thanks for pointing that out. I've updated the implementation to use reflection so `DBeaverApplication` remains platform-agnostic and doesn't introduce a compile-time dependency on the Windows-only SWT class.

The implementation now resolves `org.eclipse.swt.internal.win32.OS` using `Display.class.getClassLoader()`, which works correctly in the OSGi environment, and the code is still guarded by `RuntimeUtils.isWindows()`.

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -510,6 +514,29 @@ protected void initializeApplication() throws DBException {
...
+            ClassLoader swtClassLoader = Display.class.getClassLoader();
+            Class<?> osClass = Class.forName(
+                    "org.eclipse.swt.internal.win32.OS",
+                    true,
+                    swtClassLoader
+            );
+            java.lang.reflect.Method method = osClass.getMethod(
+                    "SetCurrentProcessExplicitAppUserModelID", char[].class
+            );
+            method.invoke(null, (Object) "DBeaverCorp.DBeaverCE".toCharArray());
```
ShadelessFox: This will conflict with other editions and versions. Consider using at least `org.jkiss.dbeaver.utils.GeneralUtils#getProductTitle`, which produces a full name-and-version tuple.

plugins/org.jkiss.dbeaver.ui.app.standalone/src/org/jkiss/dbeaver/ui/app/standalone/DBeaverApplication.java
```diff
@@ -510,6 +514,29 @@ protected void initializeApplication() throws DBException {
...
+        try {
+            ClassLoader swtClassLoader = Display.class.getClassLoader();
+            Class<?> osClass = Class.forName(
+                    "org.eclipse.swt.internal.win32.OS",
+                    true,
+                    swtClassLoader
+            );
+            java.lang.reflect.Method method = osClass.getMethod(
+                    "SetCurrentProcessExplicitAppUserModelID", char[].class
+            );
```
ShadelessFox, line 532: Please use 4-space continuous indentation

product/community/DBeaver.product
```diff
@@ -10,6 +10,8 @@
 
     <launcherArgs>
         <programArgs></programArgs>
+        <!-- Load JVM in-process so the taskbar identity matches dbeaver.exe (avoids duplicate pinned icons). -->
+        <programArgsWin>-vm jre/bin/server/jvm.dll</programArgsWin>
```
ShadelessFox: I don't understand how this contributes to the fix. Is this required in addition to `SetCurrentProcessExplicitAppUserModelID`?

ShadelessFox:
Apparently, I forgot to submit my review, sorry! Please take a look

anantgupta001:
Yes, both are intentional and complementary.

-vm jre/bin/server/jvm.dll loads the JVM in-process inside dbeaver.exe (same approach Eclipse uses on Windows; macOS already had an equivalent via programArgsMac). That addresses the core issue: the pinned shortcut targets dbeaver.exe, but without -vm the visible UI often runs in a separate javaw.exe.

SetCurrentProcessExplicitAppUserModelID sets an explicit process identity before any SWT Display is created (required by the Win32 API). It ensures consistent taskbar grouping even when process/shortcut identity would otherwise diverge.

ShadelessFox (approved):

SolovevAnton (approved):
Approved, passed to QA team

uslss:
The issue is still reproducible if the icon was pinned during application loading

anantgupta001:
> The issue is still reproducible if the icon was pinned during application loading

The previous pinned icon needs to be unpinned first. Then run the application with my PR, pin the new icon, close the application, and launch it again. After that, you shouldn't see duplicate icons anymore.

PR #41396, opened 2026-06-26
Fix logic for active results tab selection in SQL editor (#40082 ,#40223)

### LINKED ISSUE #40082: "Pin tab" shortcut logic should always pin or unpin the currently active tab, not some other

anantgupta001:
fixes #40082, #40223

### Problem
`getActiveResultsTab()` sometimes returned the wrong tab when tabs were changed via keyboard or programmatic actions. This was due to relying on a stale `activeResultsTab` field.

Additionally, the right-click context menu (rename, detach, close tabs to the left/right, etc.) relied on this same active-tab derivation, which caused it to act on the wrong tab when right-clicking an unselected tab — since right-click does not change `CTabFolder` selection on most platforms.

### Fix
- Removed the `activeResultsTab` field entirely. `getActiveResultsTab()` now derives the active tab directly from `resultTabs.getSelection()`, used for keyboard/programmatic actions (e.g. the pin-tab shortcut) where "active" correctly means "currently selected tab."
- Added a `MouseListener.mouseDownAdapter` on `resultTabs` that records the tab under the cursor at mouse-down time into a local `CTabItem[] clickedTab` array. The context menu listener now uses `clickedTab[0]` as the target, falling back to `getActiveResultsTab()` only when nothing was clicked (e.g. menu invoked via keyboard). This avoids the multi-monitor coordinate-mapping risk of querying `Display.getCursorLocation()` at menu-open time.
- Updated `SQLEditorQueryListener.processQueryResult()` to use `getActiveResultsTab()` instead of the removed field when restoring tab selection after query execution.

### Before
https://github.com/user-attachments/assets/e20ac559-498e-4f6f-bbd2-947b1e1de395

### After (pin-tab shortcut fix — #40082/#40223)
https://github.com/user-attachments/assets/90cd4d46-d305-4d82-812c-b02aea1f8c50

### After (right-click rename/detach on unselected tab)
https://github.com/user-attachments/assets/9a2bb61d-8c65-44f2-b1e5-c15c84034466

### After (keyboard-invoked context menu fallback)
https://github.com/user-attachments/assets/1d7beca4-7d90-463f-97fe-5ab1d47f902a


ShadelessFox:
See https://github.com/dbeaver/dbeaver/pull/40223#pullrequestreview-3750565561

And also please change your commit messages in accordance to our contribution guide: https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code#commit-message-rules

anantgupta001:
> See [#40223 (review)](https://github.com/dbeaver/dbeaver/pull/40223#pullrequestreview-3750565561)
> 
> And also please change your commit messages in accordance to our contribution guide: https://github.com/dbeaver/dbeaver/wiki/Contribute-your-code#commit-message-rules

My bad, I thought the issue was still open and that I could contribute to it.

I'll make sure to follow the commit message guidelines from now on. As for this PR, I don't think changing the commit message is necessary since it won't be merged anyway.

ShadelessFox:
No, no - that linked PR was closed for a reason. Please update your PR, and hopefully it will get merged.

HellAmbro:
Hi @anantgupta001, the issue with your solution is that when you right click a non activeTab, for example for rename or detach the action is applied on the current active tab. That's why my attempt, that i honestly forgot about, to use two distinct  instances of ResultsTab, not the clever solution. 
I think that can be quite annoying to select each tab for renaming or detach when now with mouse right click works properly.

anantgupta001:
> Hi @anantgupta001, the issue with your solution is that when you right click a non activeTab, for example for rename or detach the action is applied on the current active tab. That's why my attempt, that i honestly forgot about, to use two distinct instances of ResultsTab, not the clever solution.
> I think that can be quite annoying to select each tab for renaming or detach when now with mouse right click works properly.

Thanks for catching this, [@HellAmbro](https://github.com/HellAmbro) — you were right, and I've fixed it without reintroducing the two-instance shadow state.

The issue was that `getActiveResultsTab()` (used by the context menu) only derives from `resultTabs.getSelection()`, but right-click doesn't change `CTabFolder` selection on most platforms — so the menu was always resolving to whatever tab was previously selected, not the one actually right-clicked.

Instead of tracking a second `ResultsTab` reference, I added a `getContextMenuTargetTab()` helper that resolves the tab under the cursor at the moment the menu is opened (via `resultTabs.getItem(cursorPos)`), falling back to `getActiveResultsTab()` only when there's no tab under the cursor (e.g. menu invoked via keyboard). The context menu listener now uses this for rename/detach/close-tabs-to-left-right, while keyboard-driven actions like the pin shortcut still use `getActiveResultsTab()` as before.

I tested:
- Right-click an unselected tab → Rename → correctly renames the right-clicked tab, not the active one
- Right-click an unselected tab → Detach → same, correct target
- Switching active tab, then right-clicking a *different* unselected tab → still resolves correctly each time
- Pin shortcut on the active tab → still pins the correct (active) tab, unaffected

Pushed the update — would appreciate another look when you have a chance.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -1910,8 +1908,22 @@ private ToolItem getViewToolItem(@NotNull String commandId) {
     }
 
     private CTabItem getActiveResultsTab() {
-        return activeResultsTab == null || activeResultsTab.isDisposed() ?
-            (resultTabs == null ? null : resultTabs.getSelection()) : activeResultsTab;
+        if (resultTabs == null || resultTabs.isDisposed()) return null;
+        CTabItem sel = resultTabs.getSelection();
+        return (sel != null && !sel.isDisposed()) ? sel : null;
```
ShadelessFox:
```suggestion
        if (resultTabs == null || resultTabs.isDisposed()) {
            return null;
        }
        CTabItem sel = resultTabs.getSelection();
        if (sel == null || sell.isDisposed()) {
            return null;
        }
        return sel;
```

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -1910,8 +1908,22 @@ private ToolItem getViewToolItem(@NotNull String commandId) {
...
+    }
+
+    @Nullable
+    private CTabItem getContextMenuTargetTab() {
+        if (resultTabs == null || resultTabs.isDisposed()) {
+            return null;
+        }
+        Point pt = resultTabs.toControl(resultTabs.getDisplay().getCursorLocation());
```
ShadelessFox: I'm not sure whether this will work correctly against a multi-monitor setup...

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -1910,8 +1908,22 @@ private ToolItem getViewToolItem(@NotNull String commandId) {
     }
 
     private CTabItem getActiveResultsTab() {
-        return activeResultsTab == null || activeResultsTab.isDisposed() ?
```
ShadelessFox, line 1914: There are still places where `activeResultsTab` is used. Make it so that it can be _removed_ completely.

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
ShadelessFox: I suggest bringing back the original `MouseListener.mouseDownAdapter` but store the active tab in a local variable like so:
```java
CTabItem[] clickedTab = new CTabItem[1];
resultTabs.addMouseListener(MouseListener.mouseDownAdapter(e ->
    clickedTab[0] = resultTabs.getItem(new Point(e.x, e.y))));
```
Then you can remove this helper method completely.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
anantgupta001: Fixed the brace style in getActiveResultsTab()

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
anantgupta001: > I'm not sure whether this will work correctly against a multi-monitor setup...

Good catch on the multi-monitor concern — reverted to the mouse-down listener approach you suggested, storing the clicked tab in a local CTabItem[] array instead of querying cursor location at menu-open time. getContextMenuTargetTab() is removed entirely; the menu listener now reads clickedTab[0] directly, falling back to getActiveResultsTab() when nothing was clicked (e.g. keyboard-invoked menu).

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
anantgupta001: > There are still places where `activeResultsTab` is used. Make it so that it can be _removed_ completely.

The activeResultsTab field is fully removed — including from SQLEditorQueryListener.processQueryResult(), which was the other place still reading it; it now uses getActiveResultsTab() there too.

ShadelessFox (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -1503,12 +1502,17 @@ public void widgetSelected(SelectionEvent e) {
...
+
             MenuManager menuMgr = new MenuManager();
             Menu menu = menuMgr.createContextMenu(resultTabs);
             menuMgr.addMenuListener(manager -> {
-                final CTabItem activeTab = getActiveResultsTab();
+                CTabItem target = clickedTab[0];
+                final CTabItem activeTab = (target != null && !target.isDisposed())
+                        ? target
+                        : getActiveResultsTab();
                 boolean activeTabHasSingleResult = activeTab != null && activeTab.getData() instanceof QueryResultsContainer;
```
E1izabeth, line 1524:
```suggestion
                final CTabItem activeTab = (target != null && !target.isDisposed())
                    ? target
                    : getActiveResultsTab();
                boolean activeTabHasSingleResult = activeTab != null && activeTab.getData() instanceof QueryResultsContainer;
```

E1izabeth (commented):
Please use 4-spaces indentation

anantgupta001:
> Please use 4-spaces indentation

mb, applied your suggested changes

E1izabeth (dismissed):

E1izabeth:
Thanks. Passed to QA for testing

mariiachugunova:
@anantgupta001  Keyboard shortcut behavior looks correct (uses active tab), and context menu actions like Rename/Detach correctly apply to the right-clicked tab. But Pin/Unpin from the context menu still act on the active tab instead of the tab under the cursor, making context menu behavior inconsistent.

https://github.com/user-attachments/assets/7ddc1a9e-f198-4bc7-99df-a35913fe5296

anantgupta001:
> @anantgupta001 Keyboard shortcut behavior looks correct (uses active tab), and context menu actions like Rename/Detach correctly apply to the right-clicked tab. But Pin/Unpin from the context menu still act on the active tab instead of the tab under the cursor, making context menu behavior inconsistent.
> 
>  20260706-1014-20.1460712.mp4


fixed

https://github.com/user-attachments/assets/062437cd-2a69-4b7a-8096-480ab1f322b1

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -1503,12 +1502,17 @@ public void widgetSelected(SelectionEvent e) {
         }
 
         {
+            final CTabItem[] clickedTab = new CTabItem[1];
             resultTabs.addMouseListener(MouseListener.mouseDownAdapter(e ->
-                activeResultsTab = resultTabs.getItem(new Point(e.x, e.y))));
+                    clickedTab[0] = resultTabs.getItem(new Point(e.x, e.y))));
```
E1izabeth: Please use 4-spaces indentation
```suggestion
                clickedTab[0] = resultTabs.getItem(new Point(e.x, e.y))));
```

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -1503,12 +1502,17 @@ public void widgetSelected(SelectionEvent e) {
...
+
             MenuManager menuMgr = new MenuManager();
             Menu menu = menuMgr.createContextMenu(resultTabs);
             menuMgr.addMenuListener(manager -> {
-                final CTabItem activeTab = getActiveResultsTab();
+                CTabItem target = clickedTab[0];
+                final CTabItem activeTab = (target != null && !target.isDisposed())
+                        ? target
+                        : getActiveResultsTab();
```
E1izabeth:
```suggestion
                final CTabItem activeTab = (target != null && !target.isDisposed())
                    ? target
                    : getActiveResultsTab();
```

E1izabeth (dismissed):

E1izabeth:
Passed to QA

mariiachugunova:
@anantgupta001 I noticed a small inconsistency here. Delete always removes the active result tab, even if the context menu was opened on a different tab. The other context menu actions, however, are applied to the tab that was right-clicked

https://github.com/user-attachments/assets/01339475-ad2b-4d3c-a70a-92d1b36611ee

anantgupta001:
@mariiachugunova thanks — fixed. Close from the context menu now targets the right-clicked tab (same as Rename/Detach/Pin). The keyboard shortcut still closes the currently selected tab.

https://github.com/user-attachments/assets/5f10a70d-89a0-4a6d-9758-6493f30f2c97

E1izabeth (dismissed):

mariiachugunova:
@anantgupta001 Thanks for the fix! I found another issue related to this fix.

**Steps to reproduce:**

1. Open several result tabs.
2. Click tab 3 with the mouse.
3. Switch to tab 1 using the keyboard.
4. Press Shift+F10 to open the context menu and choose any action.

**Actual:** The action is applied to tab 3 (the last tab clicked with the mouse).
**Expected:** Since the context menu is opened via the keyboard, the action should be applied to the currently selected tab (tab 1)


https://github.com/user-attachments/assets/465a840c-d5eb-4b3d-bba2-93c3c816314c

anantgupta001:
@mariiachugunova kindly retest

E1izabeth (approved):

IvanGorshechnikov (approved):

IvanGorshechnikov:

mariiachugunova:

ShadelessFox (approved):

PR #41410, opened 2026-06-29
#6731 Add escaping of Connection CLI Parameter

dhufnagel:
Resolves #6731 by adding escaping of CLI Connection Parameter with Backslash for the Pipe Character.

This MR is created with the use of AI (Claude Code).


Matvey16:
Verified

DBeaver CE Windows
Dbeaver CE Linux

PR #41429, opened 2026-07-01
dbeaver/pro#9757 CustomSashForm fix

serge-rider:


Matvey16:
Verified

DBeaver UE Windows
Dbeaver UE Linux

PR #41433, opened 2026-07-01
#38554 Honor "Use fully qualified names" option for foreign keys and constraints in DDL

### LINKED ISSUE #38554: "Use fully qualified names" does not work on foreign keys
### Description

I have a table with foreign keys. When right clicking that table => Generate SQL => DDL, the dialogue window shows an option "Use fully qualified names". When enabled, the DDL reads

`CREATE TABLE [database].schema.table`

When disabled, this changes to

`CREATE TABLE table`

However, when in addition I enable the "Separate foreign keys" option, the "use fully qualified names" option does not affect the DDL for the foreign keys. They read like

`ALTER TABLE [database].schema.table ADD CONSTRAINT ...`

no matter if "use fully qualified names" is selected or not, this line does not change.

Expected behaviour: 
The fully qualified name setting should also affect the foreign keys (and potentially indexes etc.) of a table, also when "Separate foreign keys" is selected

### DBeaver Version

25.1.1

### Operating System

Win 11

### Database and driver

Azure SQL

### Steps to reproduce

_No response_

### Additional context

_No response_

gaurav0107:
### Implementation

When generating table DDL (Generate SQL -> DDL) with **Separate foreign keys** /
separate constraints enabled, the standalone `ALTER TABLE ... ADD CONSTRAINT ...`
statement always used the fully qualified table name, ignoring the dialog's
**Use fully qualified names** (`useFQN`) option. Toggling the option had no
effect on the foreign-key / constraint lines (reported in #38554).

The create-action path in the two generic edit managers hardcoded
`table.getFullyQualifiedName(DBPEvaluationContext.DDL)`. This routes the target
table name through the option-aware helper `DBUtils.getEntityScriptName(table, options)`
instead, which returns the fully qualified name when `useFQN` is set (the
default, so existing output is unchanged) and the plain quoted name otherwise:

- `SQLForeignKeyManager.addObjectCreateActions`
- `SQLConstraintManager.addObjectCreateActions`

This matches the pattern already used by `SQLIndexManager`, `SQLTableManager`,
and the referenced-table branch of `SQLForeignKeyManager` itself, so foreign
keys and constraints now behave consistently with the rest of the generated DDL.

Only the create path is changed; the drop/delete paths are left as-is.

**Manual verification:** right-click a table that has a foreign key ->
Generate SQL -> DDL; with **Separate foreign keys** enabled, toggle **Use fully
qualified names**. The `ALTER TABLE ... ADD CONSTRAINT` line now drops the
catalog/schema prefix when the option is off and keeps it when on.

### Issue reference

Closes dbeaver/dbeaver#38554

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLConstraintManager.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/sql/edit/struct/SQLConstraintManager.java
@@ -64,7 +64,7 @@ public abstract class SQLConstraintManager<OBJECT_TYPE extends AbstractTableCons
         actions.add(
             new SQLDatabasePersistAction(
                 ModelMessages.model_jdbc_create_new_constraint,
-                "ALTER TABLE " + table.getFullyQualifiedName(DBPEvaluationContext.DDL) + " ADD " + getNestedDeclaration(monitor, table, command, options)));
+                "ALTER TABLE " + DBUtils.getEntityScriptName(table, options) + " ADD " + getNestedDeclaration(monitor, table, command, options)));
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


gaurav0107:
Closing this draft. On a closer read, the **REFERENCES** (referenced-table) part
of the generated foreign-key clause already routes through the "Use fully
qualified names" option in `devel`
(`SQLForeignKeyManager#getNestedDeclarationScript` uses
`DBUtils.getEntityScriptName(refConstraint.getParentObject(), options)`), and per
the clarification on #38554 the `ALTER TABLE` / CONSTRAINT owner-table part is
intentionally kept fully qualified. This change modified the CONSTRAINT part, so
it is not the right scope for the issue. Withdrawing.

PR #41452, opened 2026-07-05
dbeaver/dbeaver#23137 Add code folding for --region/--endregion marke…

### LINKED ISSUE #23137: Code folding by "special" words/comments
At this point there is no way to add folding markers manualy

I would like to have markers, ideally to do not break compatibility as "special" comments
```sql

--region some code group
code1

--region some code subgroup

code2
--endregion 

code3
--endregion 
```
expanded
image
folded
image
image

no ? 

**Additional context**
this feature is allready avaiable in Aginity Workbench, Coginiti Pro, JustyBase (my code), simmilar feature is in c# (#region / #endregion)
accualy brackets and BEGIN/END have this funcionality. bur region does not :(
image
collapses to
image

KrzysztofDusko:
Closes dbeaver/dbeaver#23137

## Problem

DBeaver supports code folding for multi-line SQL statements (parser-based), but not for explicit section markers `--region` / `--endregion` commonly used in Visual Studio, VS Code, and Rider to organize large scripts. Issue #23137 requests parity with those editors.

The existing folding pipeline (`SQLReconcilingStrategy` + Eclipse `ProjectionAnnotationModel`) was built only around `extractScriptQueries()` results. Region markers are comments and are invisible to the SQL parser as fold targets, so a separate detection path is required.

## Scope of changes

Three files in `org.jkiss.dbeaver.ui.editors.sql`:

### 1. `SQLReconcilingStrategy.java` — region detection and reconcile integration

**What changed**

- Added `extractRegionFoldingPositions()` — a line-by-line document scan for `--region` / `--endregion`.
- Region fold annotations are merged into the same reconcile cycle as parser-based folds.
- Parser-based folds strictly enclosed inside a region are excluded from the annotation model.
- Reworked collapsed-state restore: `markCollapsed()` on new annotations replaced with `model.collapse()` after `modifyAnnotations()`.
- Added `syncCollapsedProjectionAnnotations()` to re-apply collapse on all marked annotations after reconcile.
- Extended `saveState()` to persist deferred inner collapse offsets from `SQLProjectionAnnotationModel`.
- Fixed `getNumberOfLines()` boundary handling for exclusive end offsets.

**Why these decisions**

| Decision | Rationale |
|----------|-----------|
| **Separate scanner instead of extending SQL parser** | Region markers are editor-organizational comments, not SQL syntax. A lightweight line scan keeps the change localized, avoids dialect/parser coupling, and does not affect query execution or completion. |
| **Match only SQL comment partitions** | Before pattern matching, the scanner checks that `--` is in `CONTENT_TYPE_SQL_COMMENT`. This prevents false positives inside string literals (e.g. `'--region'`) or other partitions. |
| **Case-insensitive patterns (`(?i)`)** | Matches VS Code / Visual Studio behavior where `--REGION`, `--EndRegion`, etc. are accepted. |
| **Stack-based pairing (`Deque`)** | Supports nested regions naturally: inner `--endregion` closes the nearest `--region`, outer pairs remain valid. Unmatched `--endregion` is ignored (no orphan end). |
| **Region span includes marker lines through end of `--endregion` line** | Fold range starts at `--region` line offset and ends at the start of the line *after* `--endregion` (or EOF). This keeps both marker lines visible when expanded and hides the full section when collapsed — consistent with how users expect region folding to behave. |
| **Skip single-line regions** | Folding a one-line region provides no benefit and clutters the gutter; same rule as `deservesFolding()` for statements. |
| **Suppress parser folds strictly inside regions** | Without this, a region containing multiple statements would show overlapping +/- handles (region + each statement). Eclipse projection folding does not handle overlapping annotations well. The region becomes the single fold handle for that section — intentional trade-off for clarity. |
| **`isStrictlyEnclosed` (not merely overlapping)** | Folds with identical bounds as the region are not suppressed (edge case). Folds that share a boundary with the region but extend outside remain visible. |
| **`model.collapse()` instead of `markCollapsed()`** | During testing, region folds created with `markCollapsed()` did not reliably show collapse UI (+/- icons, hidden lines). Calling `collapse()` on the projection model after annotations are registered correctly wires the visual projection layer. |
| **Preserve collapsed offsets of deleted annotations** | When reconcile removes and recreates annotations (e.g. after edit), offsets of previously collapsed folds are collected from deleted elements and reapplied to new annotations at the same document offsets. |
| **`syncCollapsedProjectionAnnotations()` after reconcile** | Ensures all annotations marked collapsed are fully applied in the projection model, including cases where reconcile recreates annotation instances. |
| **Exclusive-end fix in `getNumberOfLines()`** | Region length uses an exclusive end offset (start of next line). Using `offset + length` directly as a document offset could point past the last included character and miscount lines for regions ending at file boundary. |

### 2. `SQLProjectionAnnotationModel.java` (new) — nested fold behavior

**What changed**

New subclass of `ProjectionAnnotationModel` overriding `collapse()`, `expand()`, and `toggleExpansionState()`.

When collapsing a parent fold:
1. Collect offsets of strictly enclosed *collapsed* child folds.
2. Store them in `deferredCollapsedEnclosedOffsets` keyed by parent offset.
3. Expand those children before collapsing the parent.

When expanding the parent:
1. Restore previously deferred children, collapsed in ascending order by fold length (inner/smaller folds first).

`getDeferredCollapsedOffsets()` exposes deferred state for editor persistence.

**Why a separate class**

Eclipse's default `ProjectionAnnotationModel` does not handle nested collapsed projections: collapsing an outer region while an inner region/statement is already collapsed leads to inconsistent UI (hidden inner state, broken +/- icons). This is a known limitation of stacking projection collapses.

The custom model solves two scenarios discovered during manual testing:
- **OUTER collapsed alone** — works with default model.
- **INNER collapsed, then OUTER collapsed** — default model breaks; inner collapse must be temporarily undone and remembered.

Keeping this logic in a dedicated class avoids polluting `SQLReconcilingStrategy` (annotation lifecycle) with projection UI behavior (collapse/expand interaction).

**Why restore inner folds sorted by length**

When re-collapsing multiple nested folds, smaller (inner) folds must be collapsed before larger enclosing ones; otherwise the outer collapse logic would re-trigger deferral. Sorting by annotation position length ensures stable restoration order.

### 3. `SQLEditorSourceViewer.java` — install custom projection model

**What changed**

Override `createVisualAnnotationModel()` to replace Eclipse's internal `fProjectionAnnotationModel` field with `SQLProjectionAnnotationModel` via reflection.

**Why reflection**

`ProjectionViewer` creates and owns `fProjectionAnnotationModel` internally; there is no public extension point or setter to substitute a subclass. Reflection is a pragmatic workaround used only at viewer initialization. If reflection fails, the editor falls back to the default model (region folding still appears, but nested collapse behavior degrades gracefully — a warning is logged).

**Why check `visualModel == wiredModel`**

`ProjectionViewer` may use the same instance as both the wired model and the visual model. The override returns the SQL model only when they are the same object, avoiding incorrect model substitution in configurations where visual and wired models differ.

## What is intentionally out of scope

- No changes to SQL parsing, execution, or formatting.
- No folding for `(` `)` brackets — DBeaver never had parenthesis-level folds; regions do not affect bracket matching or auto-close.
- Parser-based folds inside a region are hidden, not nested under the region fold in the gutter. Users choose region-level folding *instead of* per-statement folding within that section.
- No new preferences — regions use the existing **Enable code folding** setting.

## Test plan

- [x] Multi-line `--region` / `--endregion` — gutter fold appears and collapses/expands correctly
- [x] Case variants: `--REGION`, `--EndRegion`
- [x] Nested regions — independent fold at each level
- [x] Collapse inner region, then outer — inner expands visually; expanding outer restores inner state
- [x] Multiple statements inside one region — only region +/- visible (no duplicate statement folds)
- [x] Marker inside string literal — no spurious fold
- [x] Save/reopen script file — collapsed state preserved (including deferred inner folds)
- [x] Statement folding outside regions — unchanged

## AI disclosure

This PR was assisted by AI (Cursor).

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRegionMarkerFolding.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRegionMarkerFolding.java
@@ -1,0 +1,247 @@
+/*
+ * DBeaver - Universal Database Manager
+ * Copyright (C) 2010-2026 DBeaver Corp and others
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jkiss.dbeaver.model.sql.parser;
+
+import org.eclipse.jface.text.BadLocationException;
+import org.eclipse.jface.text.IDocument;
+import org.eclipse.jface.text.IDocumentExtension3;
+import org.eclipse.jface.text.IDocumentPartitioner;
+import org.eclipse.jface.text.IRegion;
+import org.jkiss.code.NotNull;
+import org.jkiss.dbeaver.Log;
+
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Deque;
+import java.util.List;
+import java.util.Locale;
+import java.util.regex.Pattern;
+
+/**
+ * Scans {@code --region}/{@code --endregion} script markers and decides when folding gutter refresh is needed.
+ */
+public final class SQLRegionMarkerFolding {
+
+    private static final Log log = Log.getLog(SQLRegionMarkerFolding.class);
+
+    private static final Pattern REGION_START_PATTERN = Pattern.compile("^\\s*--\\s*region\\b", Pattern.CASE_INSENSITIVE);
+    private static final Pattern REGION_END_PATTERN = Pattern.compile("^\\s*--\\s*endregion\\b", Pattern.CASE_INSENSITIVE);
+    private static final Pattern REGION_MARKER_NAME_PATTERN = Pattern.compile("^\\s*--\\s*region\\s*(.*)", Pattern.CASE_INSENSITIVE);
+
+    private SQLRegionMarkerFolding() {
+    }
+
+    public record RegionFold(@NotNull String regionKey, int offset, int length) {
+    }
+
+    private record RegionStartFrame(int offset, @NotNull String regionKey) {
+    }
+
+    @NotNull
+    public static List<RegionFold> scanRegions(@NotNull IDocument document) {
+        List<RegionFold> regions = new ArrayList<>();
+        if (document.getLength() == 0) {
+            return regions;
+        }
+        IDocumentPartitioner partitioner = getSqlDocumentPartitioner(document);
+        Deque<RegionStartFrame> regionStarts = new ArrayDeque<>();
+        int lineCount = document.getNumberOfLines();
+        for (int line = 0; line < lineCount; line++) {
+            collectRegionFoldingOnLine(document, line, lineCount, partitioner, regionStarts, regions);
+        }
+        return regions;
+    }
+
+    @NotNull
+    public static List<RegionFold> scanFoldableRegions(@NotNull IDocument document) {
+        List<RegionFold> foldable = new ArrayList<>();
+        for (RegionFold region : scanRegions(document)) {
+            if (isValidDocumentRange(document, region.offset(), region.length())
+                && getRegionNumberOfLines(document, region) > 1
+            ) {
+                foldable.add(region);
+            }
+        }
+        return foldable;
+    }
+
+    public static boolean needsFoldingGutterRefresh(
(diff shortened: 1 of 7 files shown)


KrzysztofDusko:
**Manual testing**

I tested the changes manually in the SQL editor. From my side, the behavior looks correct (nested collapse/expand, gutter icons, editing above a region).

**Sample script:**

```sql
SELECT 0;

--REGION OUTER

SELECT 1;

--REGION INNER1

SELECT 2;

--REGION INNER11

SELECT 3;

--ENDREGION

--ENDREGION

SELECT 4;

--REGION INNER2

SELECT 5;

--ENDREGION

SELECT 6;

--ENDREGION

SELECT 7;
```

**Screenshot (folding in the editor):**

SQL region folding in DBeaver editor

**Manually verified scenarios:**
- Collapse/expand of individual regions (`OUTER`, `INNER1`, `INNER11`, `INNER2`)
- Nested regions — collapse `INNER` → collapse `OUTER` → expand `OUTER` (inner fold state restored)
- Gutter stays in sync when editing above a region (Enter above `SELECT 0`)

KrzysztofDusko:
Superseded by #23137

PR #41457, opened 2026-07-06
dbeaver/pro#9778 Fix shortcuts resolution

E1izabeth:
**For reviewers: most of diff are just spaces, hide them:**
image

**For QA:**
Test that shortcuts are fine in tooltips there:
image

image

Check that the overlay still appears normally by shortcut, from the context menu, and from the toolbar button in result sets (different presentations), Reference panel.


IvanGorshechnikov (approved):

ShadelessFox (approved):

uslss:
verified (db ue)

PR #41459, opened 2026-07-06
dbeaver/pro#9656 easy config next steps

ShadelessFox:


plugins/org.jkiss.dbeaver.ui.app.config/src/org/jkiss/dbeaver/ui/app/config/ProductConfigWizardHandler.java
```diff
@@ -0,0 +1,49 @@
...
+public final class ProductConfigWizardHandler extends AbstractHandler {
+    @Nullable
+    @Override
+    public Object execute(@NotNull ExecutionEvent event) throws ExecutionException {
+        var dialog = new ProductConfigWizardDialog(HandlerUtil.getActiveWorkbenchWindow(event));
+        if (dialog.open() != IDialogConstants.OK_ID) {
+            return null;
+        }
+        if (dialog.isRestartRequired()) {
+            if (UIUtils.confirmAction(
```
serge-rider, line 39: I don't think we need to confirm this. Usually we cannot continue without restart in any case.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ui.app.config/src/org/jkiss/dbeaver/ui/app/config/ProductConfigWizardHandler.java
ShadelessFox: It will require user confirmation only when initiated from the GUI, not when it's automatically shown at startup. It's totally fine to restart later - the user might still have some unsaved work.

PR #41467, opened 2026-07-06
dbeaver/pro#8749 Ask user choose function

IvanGorshechnikov:
Closes https://github.com/dbeaver/pro/issues/8749


plugins/org.jkiss.dbeaver.ui.ai/plugin.xml
```diff
@@ -2,6 +2,12 @@
 <?eclipse version="3.4"?>
 
 <plugin>
+    <extension-point id="messageList" name="AI Chat Message List" schema="schema/org.jkiss.dbeaver.ui.ai.messageList.exsd"/>
+
+    <extension point="org.jkiss.dbeaver.ui.ai.messageList">
```
serge-rider: Let's rename it to org.jkiss.dbeaver.ui.ai.chatView. And we can omit default implementation declaration - create WebViewMessageList if no extension found.

serge-rider (changes requested):

PR #41496, opened 2026-07-10
dbeaver/pro#9491 Introduce new database URL parser and use it to fix snowflake url template

E1izabeth:
Old and very old snowflake connections should work as well as newly created ones.
Please also test
- various custom URL configurations and databases.
- new connection from url creation
- generic connection page
- IoTDB connection
- connections importing


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/StringTemplate.java
```diff
@@ -0,0 +1,1655 @@
...
+
+import java.util.*;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+
+public class StringTemplate {
```
serge-rider, line 34: Is it full refactoring of DatabaseURL? If so then it is better to h=keep this code in DatabaseURL.
And why DatabaseURL.MetaURL is still used?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/StringTemplate.java
E1izabeth: That's what I wanted to discuss

uslss:
verified (UE)

PR #41515, opened 2026-07-12
dbeaver/dbeaver#23137 Add folding for SQL region markers

### LINKED ISSUE #23137: Code folding by "special" words/comments

KrzysztofDusko:
Closes dbeaver/dbeaver#23137

  ## Problem

  DBeaver supports parser-based folding for multi-line SQL statements, but
  it does not recognize explicit `--region` / `--endregion` markers commonly
  used to organize large scripts.

  Region markers are SQL comments and are therefore not returned as foldable
  elements by the existing query extraction pipeline. They require a
  separate detection path integrated with the existing projection annotation
  model.

  ## Changes

  ### Region marker scanner

  Added `SQLRegionMarkerFolding` in the SQL model plugin.

  It:

  - scans SQL documents line by line for `--region` / `--endregion`;
  - matches markers case-insensitively;
  - verifies SQL comment partitions to avoid matching markers inside string
  literals;
  - uses a stack to support nested regions;
  - creates hierarchical region keys such as `OUTER/INNER/CHILD`;
  - ignores orphan `--endregion` and unclosed regions;
  - validates document ranges and excludes non-foldable single-line ranges;
  - determines when document edits require a projection gutter refresh.

  The scanner is independent of SQL dialect parsing because region markers
  are editor organization comments rather than SQL syntax.

  ### Reconcile integration

  Updated `SQLReconcilingStrategy` to:

  - combine region folds with existing parser-based folds;
  - suppress parser folds strictly enclosed by a region, avoiding
  overlapping projection handles;
  - cache regions by their hierarchical keys;
  - preserve collapsed state when annotations are recreated or shifted after
  edits;
  - restore collapsed regions through
  `ProjectionAnnotationModel.collapse()`;
  - correctly calculate line counts for ranges with an exclusive end offset;
  - schedule UI-thread refreshes when the projection gutter needs
  repainting;
  - coalesce repeated refresh requests;
  - keep the original edit offset separate from the widened damaged region
  used for SQL reconciliation.

  ### Nested folding state

  Added `SQLProjectionAnnotationModel`, which preserves nested collapse
  state.

  When an outer region is collapsed, already-collapsed enclosed folds are
  temporarily expanded and remembered. When the outer region is expanded,
  the enclosed folds are restored in ascending order of their range length.

  This avoids inconsistent projection state when nested collapsed
  annotations overlap.

  ### Projection model installation

  Updated `SQLEditorSourceViewer` to install `SQLProjectionAnnotationModel`
  during visual annotation model initialization.

  The internal Eclipse projection model is accessed through
  `MethodHandles.privateLookupIn()` and a `VarHandle`. The implementation
  does not use `java.lang.reflect.Field` or `setAccessible()`.

  If the custom model cannot be installed, a warning is logged and the
  standard Eclipse projection model remains available.

  ### Gutter refresh

  Added `SQLEditorBase.refreshProjectionFoldingPresentation()` to refresh
  projection annotations and repaint the vertical ruler without invalidating
  the presentation of the entire SQL document.

  ## Tests

  Added unit tests covering:

  - nested regions and hierarchical region keys;
  - correct marker offsets;
  - orphan `--endregion` markers;
  - unclosed regions;
  - markers inside SQL string literals;
  - two-line marker-only regions;
  - renamed region keys;
  - gutter refresh decisions for edits before, on, inside, and after
  regions.

  Manual verification covered:

  - collapsing and expanding nested regions;
  - restoring an inner collapsed state after expanding an outer region;
  - projection gutter synchronization after edits above a region;
  - parser-based folding outside regions.

  The full aggregate build is currently blocked locally by an unrelated
  missing upstream bundle: `com.dbeaver.rest.client`.

  ## Out of scope

  - SQL parsing, execution, and formatting behavior;
  - folding based on parentheses;
  - new user preferences;
  - displaying parser folds inside an enclosing explicit region.

  ## AI disclosure

  This PR was assisted by AI (Cursor).

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRegionMarkerFolding.java
+++ b/plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRegionMarkerFolding.java
@@ -21,8 +21,12 @@ import org.eclipse.jface.text.IDocument;
 import org.eclipse.jface.text.IDocumentExtension3;
 import org.eclipse.jface.text.IDocumentPartitioner;
 import org.eclipse.jface.text.IRegion;
+import org.eclipse.jface.text.ITypedRegion;
 import org.jkiss.code.NotNull;
+import org.jkiss.code.Nullable;
 import org.jkiss.dbeaver.Log;
+import org.jkiss.dbeaver.model.impl.sql.BasicSQLDialect;
+import org.jkiss.dbeaver.model.sql.SQLDialect;
 
 import java.util.ArrayDeque;
 import java.util.ArrayList;
@@ -30,18 +34,15 @@ import java.util.Collection;
 import java.util.Deque;
 import java.util.List;
 import java.util.Locale;
-import java.util.regex.Pattern;
+import java.util.NavigableMap;
+import java.util.TreeMap;
 
 /**
- * Scans {@code --region}/{@code --endregion} script markers and decides when folding gutter refresh is needed.
+ * Scans SQL comment region markers and provides indexes used by the SQL editor folding reconciler.
  */
 public final class SQLRegionMarkerFolding {
 
     private static final Log log = Log.getLog(SQLRegionMarkerFolding.class);
-
-    private static final Pattern REGION_START_PATTERN = Pattern.compile("^\\s*--\\s*region\\b", Pattern.CASE_INSENSITIVE);
-    private static final Pattern REGION_END_PATTERN = Pattern.compile("^\\s*--\\s*endregion\\b", Pattern.CASE_INSENSITIVE);
-    private static final Pattern REGION_MARKER_NAME_PATTERN = Pattern.compile("^\\s*--\\s*region\\s*(.*)", Pattern.CASE_INSENSITIVE);
 
     private SQLRegionMarkerFolding() {
     }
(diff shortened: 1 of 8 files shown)


plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRegionMarkerFolding.java
```diff
@@ -0,0 +1,247 @@
...
+            return regions;
+        }
+        IDocumentPartitioner partitioner = getSqlDocumentPartitioner(document);
+        Deque<RegionStartFrame> regionStarts = new ArrayDeque<>();
+        int lineCount = document.getNumberOfLines();
+        for (int line = 0; line < lineCount; line++) {
+            collectRegionFoldingOnLine(document, line, lineCount, partitioner, regionStarts, regions);
+        }
+        return regions;
+    }
```
ged-yuko, line 120: - There is no reason to do full scan each time in the first place. There is modified region information most of the time.
 - There is even no reason to scan the document line-by-line, because comments are already highlighted, so the information on where to look them should be in the partitions model already. 🤔
 - The pattern used for scanning is also wrong, it should be based on the current dialect, see how comments are handled everywhere already.

plugins/org.jkiss.dbeaver.model.sql/src/org/jkiss/dbeaver/model/sql/parser/SQLRegionMarkerFolding.java
```diff
@@ -0,0 +1,247 @@
...
+    ) {
+        for (RegionFold region : regions) {
+            int outerStart = region.offset();
+            int outerEnd = outerStart + region.length();
+            if (isStrictlyEnclosed(innerStart, innerEnd, outerStart, outerEnd)) {
+                return true;
+            }
+        }
+        return false;
+    }
```
ged-yuko: Custom annotation model is being introduced, so why iterating through the regions of interest linearly and multiple times? Did you try to use special data structures to make it closer to O(log(n)) instead of O(n)?

KrzysztofDusko:
Thanks, these were valid concerns. Addressed in 83e81bc700.

The scanner now uses the document partition model and the current SQL dialect instead of hardcoded comment parsing. Ordinary SQL insertions reuse the cached Position-backed regions; a full rescan is performed only when markers or string/multiline-comment partitions may have changed.

Reconcile now scans regions once and reuses the result. Region containment is handled by an offset-indexed RegionIndex with parent links, so we no longer iterate over the complete region collection for every parser fold.

I also fixed duplicate region-name collisions, preserved nested collapse state using Annotation identity instead of offsets, and synchronized projection refresh scheduling. Added tests cover dialect-specific comments, duplicate names, region indexing, and affected-range detection.

PR #41531, opened 2026-07-14
#41530 [CUBRID] Connection fails when metadata loading throws exceptions

### LINKED ISSUE #41530: [CUBRID] Connection fails when metadata loading throws exceptions
### Description

When connecting to a CUBRID database server, any exception thrown during the initial metadata loading phases (such as loading charsets, collations, privileges, etc.) completely blocks and aborts the connection establishment.

Specifically, in upcoming pre-release versions of CUBRID (version 11.5), catalog schema changes cause the queries in `loadPrivilege` to fail with a semantic error. Because these metadata queries are not wrapped in exception handlers, the entire connection attempt fails.

Image

### Expected behavior
Even if optional metadata (such as charsets, collations, privileges, or tracking) fails to load due to catalog changes in newer database versions, the connection should still succeed so users can run queries. Catalog load errors should be caught and logged as warnings instead of blocking the connection.

### DBeaver Version

Community edition 26.0.3

### Operating System

macOS

### Database and driver

1. Database name and version: CUBRID Server 11.5 (Pre-release build)
2. Driver name and version: CUBRID JDBC Driver 11.3.2

### Steps to reproduce

1. Attempt to connect to a CUBRID Server 11.5 (pre-release) instance using DBeaver
2. During the connection initialization phase, a semantic query execution error occurs
3. The connection dialog presents a blocker error popup and fails to establish the connection

### Additional context

_No response_

longhaseng52:
Closes https://github.com/dbeaver/dbeaver/issues/41530


IvanGorshechnikov (approved):

E1izabeth (approved):
Thanks for the contribution! Passed to QA team for testing

uslss:
verified

PR #41549, opened 2026-07-16
Fix error dialog parent shell during data transfer

JackyLin0123:
## Summary

- Parent execution-error dialogs to the active SWT shell.
- Fall back to the active workbench shell when no active shell is available.

## Root cause

Data-transfer errors were shown as children of the workbench shell instead of the active transfer/progress shell. On Linux this could let the modal transfer window obscure the error dialog.

## Validation

- `git diff --check`
- Full DBeaver CE aggregate build in Colab: `BUILD SUCCESS`


E1izabeth:
Another PR is opened #41565

PR #41563, opened 2026-07-18
#41553 Fix GaussDB schema switch from the SQL Editor toolbar

### LINKED ISSUE #41553: [GaussDB] Schema isn't changed from the toolbar for the SQL Editor
### Description

Schema stays the same after changing it in the toolbar

### DBeaver Version

26.1.3

### Operating System

Windows 11

### Database and driver

openGauss 5.0.0 build a07d57c3

gsjdbc4-1.1

### Steps to reproduce

1. Open SQL Editor
2. In the toolbar switch to `public` schema
3. Run `select current_schema();`

**Expected:** Result shows `public`
`Actual:` Result shows the default schema that was first selected

### Additional context

_No response_

augustocbx:
On GaussDB/openGauss, selecting `public` in the SQL Editor toolbar left `current_schema()` pointing at the schema chosen on connect (the user's own schema), so the switch had no effect.

Root cause is in the shared `PostgreExecutionContext`:
- `changeDefaultSchema()` skipped `SET search_path` whenever the target was the `public` schema. That is valid for standard PostgreSQL, where `public` is already the default, but GaussDB defaults to the connected user's schema, so switching to `public` must actually update the search path.
- `setSearchPath()` only prepended a schema when it was absent from the current path; a schema already present (like `public` in `"$user", public`) was never moved to the front, so `current_schema()` never changed.

The fix adds a `PostgreServerExtension.isDefaultSchemaPublic()` flag (default `true`, so every existing driver is unchanged) and overrides it to `false` for GaussDB, which lets the search path be applied when switching to `public`. `setSearchPath()` now also moves an already-present schema to the front so the selection takes effect.

Verification: added a unit test asserting the default flag value in `PostgreServerExtensionBaseTest`; the standard-PostgreSQL code path is unchanged (`!(isPublicSchema && isDefaultSchemaPublic)` reduces to the previous `!isPublicSchema`). Manual before/after on GaussDB: switch toolbar to `public`, run `select current_schema();` — now returns `public` instead of the default schema.

Closes dbeaver/dbeaver#41553

This PR was prepared with AI assistance (Claude Code); the change is small, understood, and verified by the author.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
@@ -143,7 +143,7 @@ public class PostgreExecutionContext extends JDBCExecutionContext implements DBC
             return false;
         }
 
-        if (!schema.isSystem() && !schema.isPublicSchema()) {
+        if (!schema.isSystem() && !(schema.isPublicSchema() && getDataSource().getServerType().isDefaultSchemaPublic())) {
             setSearchPath(monitor, schema.getName());
             addSearchPath(schema.getName());
         }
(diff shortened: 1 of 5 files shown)


E1izabeth (approved):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreExecutionContext.java
```diff
@@ -244,16 +244,16 @@ private void addSearchPath(@NotNull String path) {
...
-                }
-                newSearchPath.addFirst(defSchemaName);
+        if (schemaIndex < 0) {
+            if (!getDataSource().getServerType().supportsStandardSearchPath()
+                && getDefaultCatalog().getSchema(monitor, PostgreConstants.PUBLIC_SCHEMA_NAME) == null
+                && !PostgreConstants.PUBLIC_SCHEMA_NAME.equals(defSchemaName)) {
+                newSearchPath.removeIf(PostgreConstants.PUBLIC_SCHEMA_NAME::equals);
             }
+            newSearchPath.addFirst(defSchemaName);
+        } else if (schemaIndex > 0) {
```
IvanGorshechnikov, line 254: Why do you need this fix? I am afraid of it because it changes the logic for Postgres and all its forks

E1izabeth:
Still no response on the review comment.
I assume this PR is stale, so I'm closing it. Please, let me know if you continue working on this - we can reopen it.

PR #41565, opened 2026-07-19
dbeaver/dbeaver#38531 Parent execution-error dialog to the active shell

### LINKED ISSUE #38531: Issue with lost focuys on main window during data export and active popup being obfuscated by expoprt window
### Description

The issue is being caused if a 'table is full" error is encountered during an export job from one database to another.

You'll note the dialog prompt for the next course action can't be reached because its hidden when the app is active. This occurs when switching windows.

Image
Image

### DBeaver Version

Commnity Edition 25.1.1

### Operating System

Fedora 41

### Database and driver

The mysql and sqlite3 drivers were in use.

### Steps to reproduce

_No response_

### Additional context

_No response_

JackyLin0123:
Data-transfer/export batch-insert errors are shown via ExecutionQueueErrorJob, which parented the Stop/Retry/Ignore dialog to the workbench shell. On Linux/GTK the modal transfer progress window sits above the workbench, so the error dialog was rendered behind it and could not be reached. Parent the dialog to the active shell instead so it appears in front of the transfer window.

Closes #38531


E1izabeth (approved):

IvanGorshechnikov (approved):

IvanGorshechnikov:

PR #41590, opened 2026-07-21
dbeaver/pro#9872 Nullability checks and annotations

ShadelessFox:
Original STR:
1. Create a new table in SQL Server, **don't** persist it
2. Switch to Properties -> Statistics
3. Observe NPEs in the log

Now the Statistics tab should not be visible for non-persisted tables at all.


E1izabeth (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified

DBeaver CE Windows

PR #41595, opened 2026-07-22
dbeaver/dbeaver#41566 Restore Value viewer find routing

### LINKED ISSUE #41566: Find in Value window no longer working
### Description

Find (Ctrl-F) in (text) Value window triggers find (rows fetch) in Record grid, not find in field as it used to. The only way to trigger find in field Value is: R-click field in record list, Edit / Edit Cell. Then in Value View, R-click Find/Replace to find in field.

Fist noticed in previous DBeaver update, continues in current release.

### DBeaver Version

Community 26.1.3.202607191738.

### Operating System

Win 11 (Pro) 25H2

### Database and driver

SQLite, SQLite JDBC driver

### Steps to reproduce

Find in (text) Value window. (Ctrl-F or R--click Find/Replace, or menu Edit Find/Replace). 
Invokes Find in Record grid, not find in field Value window.

Image

### Additional context

_No response_

fearyan:
## Summary

- Route Find/Replace according to focused result-set control.
- Use focused Value editor native find target.
- Keep result-set overlay available from grid and filter panel.
- Add find-target fallback for plain `StyledText` Value editors.

## Root cause

Spreadsheet quick-filter change replaced focus-aware `IFindReplaceTarget` dispatch with unconditional result-set overlay dispatch. Because result-set handler is active for whole result-set editor, it intercepted Find/Replace while focus was inside Value editor.

## User impact

Ctrl/Cmd+F, Edit → Find/Replace, and Value editor context-menu action now target focused Value text. Grid and filter-panel activation continue to open result-set overlay.

## Validation

- `xmllint --noout plugins/org.jkiss.dbeaver.ui.editors.data/plugin.xml`
- `git diff --check`
- `mvn package -f product/aggregate/pom.xml -Pproduct-dbeaver-ce,product-dbeaver-eclipse-ce` (203 modules, build successful)
- Manual macOS regression test:
  - Value editor focus → Value Find/Replace
  - Result grid focus → result-set overlay
  - Filter panel focus → result-set overlay

## AI assistance

This change was generated with assistance from OpenAI Codex, then reviewed and validated by contributor.

Closes dbeaver/dbeaver#41566

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/plugin.xml
@@ -510,7 +510,11 @@
             </enabledWhen>
         </handler>
         <handler commandId="org.eclipse.ui.edit.findReplace" class="org.jkiss.dbeaver.ui.controls.resultset.handler.ResultSetHandlerMain">
-            <activeWhen><reference definitionId="org.jkiss.dbeaver.core.ui.resultset.presentation.control"/></activeWhen>
+            <activeWhen>
+                <with variable="activePart">
+                    <test property="org.jkiss.dbeaver.core.resultset.active"/>
+                </with>
+            </activeWhen>
         </handler>
         <handler commandId="org.jkiss.dbeaver.core.resultset.toggleMode" class="org.jkiss.dbeaver.ui.controls.resultset.handler.ResultSetHandlerToggleMode">
             <activeWhen><reference definitionId="org.jkiss.dbeaver.core.ui.resultset.part"/></activeWhen>
--- a/plugins/org.jkiss.dbeaver.ui.editors.data/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ui.editors.data/plugin.xml
@@ -510,11 +510,7 @@
             </enabledWhen>
         </handler>
         <handler commandId="org.eclipse.ui.edit.findReplace" class="org.jkiss.dbeaver.ui.controls.resultset.handler.ResultSetHandlerMain">
-            <activeWhen>
-                <with variable="activePart">
-                    <test property="org.jkiss.dbeaver.core.resultset.active"/>
-                </with>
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.ui.editors.data/plugin.xml
```diff
@@ -510,11 +510,7 @@
             </enabledWhen>
         </handler>
         <handler commandId="org.eclipse.ui.edit.findReplace" class="org.jkiss.dbeaver.ui.controls.resultset.handler.ResultSetHandlerMain">
-            <activeWhen>
-                <with variable="activePart">
-                    <test property="org.jkiss.dbeaver.core.resultset.active"/>
-                </with>
-            </activeWhen>
+            <activeWhen><reference definitionId="org.jkiss.dbeaver.core.ui.resultset.presentation.control"/></activeWhen>
```
E1izabeth: It makes find/replace in result set inaccessible when focus is in the filter panel

E1izabeth (changes requested):

plugins/org.jkiss.dbeaver.ui.editors.data/plugin.xml
fearyan: Thanks, you’re right. I restored broad handler activation, so Find/Replace remains accessible from the filter panel. Routing now happens by focus: Value viewer uses its text search target, while grid and filter panel keep the result-set overlay. I verified both paths locally and tested the changes on UI as well.

fearyan:
Hey @E1izabeth, please review it once when you're free.

I've tested the changes locally and it does work as required, you can test it once.

E1izabeth (changes requested):
It doesn't fix the context menu action

Image

It looks like more context and awareness are needed to fix this bug. We want to fix it ourselves.

Thanks for the contriburion!

PR #41605, opened 2026-07-22
#41604 Add ability to open Parquet files via DuckDB

### LINKED ISSUE #41604: Parquet (and in general file-based db) open with DuckDB
Following the implementation of [SQLite open with DBeaver](https://github.com/dbeaver/dbeaver/issues/15395) and [DuckDB open with DBeaver](https://github.com/dbeaver/dbeaver/issues/38253) I'm now requesting a Parquet open with DBeaver.

This mean that, when you run DBeaver with a parquet file as its first command line argument, DBeaver automatically create the connection and open the file.

Nowadays when you click on a sqlite3 file or a duckdb file (provided you set up the OS to open those files with DBeaver, of course), DBeaver ask you whether you want to open it as a database or a text file (why on earth would you open it as a text file?)

Image

If you open as database it will create the connection into a "File Database" folder:

Image

But double clicking on a parquet file shows a un-readable text:

Image

raffaelemancuso:
Closes #41604

### What this does

Opening a `.parquet` file (via command-line argument / OS file association, or from DBeaver's file navigator) now automatically creates a DuckDB-backed connection in the **File Databases** folder, following the same pattern as the existing SQLite/DuckDB/MS Access file handlers.

Since a Parquet file is not a DuckDB database file (`jdbc:duckdb:file.parquet` is not valid), the new `ParquetFileDatabaseHandler` connects to an **in-memory DuckDB instance** (`jdbc:duckdb:`) and exposes the file through a view created by a connection bootstrap query:

```sql
CREATE OR REPLACE VIEW "filename" AS SELECT * FROM read_parquet('/path/to/filename.parquet')
```

Bootstrap queries run for every new execution context, so the view exists in each in-memory database instance the connection spawns.

### Changes

- **`AbstractFileDatabaseHandler`**: added a `configureConnection(configuration, fileList)` hook invoked before the temporary datasource is created (no-op by default, existing handlers unaffected). This also makes future file formats (CSV, JSON, ...) easy to add, per the "and in general file-based db" note in the issue.
- **`ParquetFileDatabaseHandler`** (new, in `org.jkiss.dbeaver.ext.duckdb.ui`):
  - `databaseName` is kept as the parquet path so connection dedup (`PROP_ORIGINAL_FILE_PATH`) and connection naming work per file, while the actual JDBC URL is the in-memory one via `configurationType=URL`.
  - `supportedActions()` returns only `DATABASE`, so the "open as database / text editor" dialog is skipped — viewing binary Parquet as text is not useful.
  - View names and file paths are escaped for safe SQL embedding.
- **`plugin.xml`** (duckdb.ui): registered the handler under `org.jkiss.dbeaver.fileTypeHandler` with `order="104"`, extension `parquet`, reusing the DuckDB icon.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.duckdb.ui/META-INF/MANIFEST.MF
+++ b/plugins/org.jkiss.dbeaver.ext.duckdb.ui/META-INF/MANIFEST.MF
@@ -6,6 +6,8 @@ Bundle-Version: 1.0.16.qualifier
 Bundle-Release-Date: 20260921
 Require-Bundle: org.jkiss.dbeaver.ui,
  org.jkiss.dbeaver.ui.editors.connection,
+ org.jkiss.dbeaver.ui.editors.data,
+ org.jkiss.dbeaver.model.sql,
  org.jkiss.dbeaver.ext.generic.ui,
  org.jkiss.dbeaver.ext.duckdb
 Bundle-ActivationPolicy: lazy
--- a/plugins/org.jkiss.dbeaver.ext.duckdb.ui/OSGI-INF/l10n/bundle.properties
+++ b/plugins/org.jkiss.dbeaver.ext.duckdb.ui/OSGI-INF/l10n/bundle.properties
@@ -1,1 +1,5 @@
 duckdb.dialog.connection.header = DuckDB Connection Settings
+
+category.duckdb.name = DuckDB
+command.duckdb.exportResultSetToParquet.name = Export to Parquet file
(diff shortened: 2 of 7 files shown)


raffaelemancuso:
I didn't test whether it breaks something else, but:

1. In the first run, when it asks you to create a sample database, it doesn't open .parquet files
2. After that, it does create a DuckDB in-memory connection and open the parquet file

raffaelemancuso:
I have rebased the PR. Is it useful? @E1izabeth

E1izabeth:
Thanks!

E1izabeth:
Could you provide a screenshots how does UI look like?

PR #41612, opened 2026-07-23
dbeaver/cloudbeaver#4438 Remove mysql comments

### LINKED ISSUE #4438: Unable to update data if the query contains a comment.
### Description

If you execute a SELECT query containing a comment, it will be impossible to modify the data via the interface (by directly editing a value in one of the result columns).

### Steps to reproduce

SELECT yourcolumn FROM yourtable

-> green dot at the top left of your result = you can update the data via the UI

SELECT yourcolumn FROM yourtable #comment

-> yellow dot at the top left of your result = you can't update the data via the UI

### Expected/Desired Behavior

No change in behavior when a request contains a comment.

### CloudBeaver Version

26.1.1

### Additional context

_No response_

IvanGorshechnikov:
Closes https://github.com/dbeaver/cloudbeaver/issues/4438

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/parser/SQLSemanticProcessor.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/parser/SQLSemanticProcessor.java
@@ -40,6 +40,7 @@ import org.jkiss.dbeaver.model.exec.DBCAttributeMetaData;
 import org.jkiss.dbeaver.model.exec.DBCEntityMetaData;
 import org.jkiss.dbeaver.model.exec.DBCException;
 import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
+import org.jkiss.dbeaver.model.sql.SQLConstants;
 import org.jkiss.dbeaver.model.sql.SQLDialect;
 import org.jkiss.dbeaver.model.sql.SQLQueryGenerator;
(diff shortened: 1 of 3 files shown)


SolovevAnton (dismissed):

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/sql/parser/SQLSemanticProcessor.java
```diff
@@ -93,6 +96,73 @@ private static CCJSqlParser buildParser(@Nullable SQLDialect dialect, @NotNull S
...
+            .filter(prefix -> !CommonUtils.isEmpty(prefix) && !prefix.startsWith(SQLConstants.SL_COMMENT))
+            .toList();
+        if (extraCommentPrefixes.isEmpty()) {
+            return sql;
+        }
+        StringBuilder result = new StringBuilder(sql);
+        int pos = 0;
+        while (pos < sql.length()) {
+            char c = sql.charAt(pos);
+            if (c == '\'' || c == '"' || c == '`') {
```
ShadelessFox: You're not checking whether `c` is escaped or not (can be the case if a part of a comment that is not recognized by the database's SQLDialect). Moreover, you're hardcoding a list of _potential_ quote characters.

ShadelessFox (changes requested):

ShadelessFox (approved):

PR #41621, opened 2026-07-26
Quote MySQL option file password

### LINKED ISSUE #41620: MySQL native tool (mysqldump) writes password unquoted in temporary .my.cnf → passwords containing # (and other special chars) fail with "Access denied"
### Description

When using Database → Tools → Dump database (which invokes the native `mysqldump` client), DBeaver writes the connection credentials into a temporary option file and passes it to `mysqldump` via `--defaults-file`:

```
--defaults-file=/tmp/.dbeaver-temp13176011087612274957/mysql-native-handler/.my.cnf
```
The password in that file is written without surrounding quotes, e.g.:

[client]
user=root
password=Super#Secret123

Because `#` starts a comment in MySQL option files (INI format), the client parses the line as `password=Super` and silently discards `#Secret123`. The result is an authentication failure:
```
mysqldump: Got error: 1045: Access denied for user 'root'@'172.18.0.1' (using password: YES) when trying to connect
```
This only affects the native tools path. Regular browsing/querying via the JDBC driver works fine, because JDBC sends the password over the wire as a raw string and never writes a `.my.cnf`.

### DBeaver Version

Community Edition 26.1.3.202607191930

### Operating System

Arch Linux (kernel 7.1.4-arch1-1)

### Database and driver

Database: MySQL 8.0.46 running in Docker (image mysql:8.x)
Driver: MySQL 8+

### Steps to reproduce

1. Create a MySQL user (e.g. `root`) whose password contains a `#` character (e.g. `Super#Secret123`).
2. Configure the DBeaver MySQL connection with that password (Host = `127.0.0.1`, port 3306).
3. Confirm the connection works (JDBC browse/query succeeds).
4. Database → Tools → Dump database → Next → Start.
5. The dump fails with error 1045 "Access denied ... (using password: YES)".

### Expected behavior
DBeaver should write the password quoted in the temporary `.my.cnf`, as MySQL's own documentation prescribes:
```ini
[client]
password="Super#Secret123"
```
The MySQL 8.4 reference manual explicitly states:
>"The value optionally can be enclosed within single quotation marks or double quotation marks, which is useful if the value contains a # comment character." — [Technical Difficulties](https://dev.mysql.com/doc/refman/8.2/en/option-files.html)

This is also the resolution of the long-standing MySQL bug [Technical Difficulties](https://bugs.mysql.com/bug.php?id=1158): "The fix is to use quotes in my.cnf around such arguments that include '#'."

Any password character that has special meaning in option files (`#`, `;`, `"`, `'`, `\`) should be safely escaped/quoted by DBeaver when it generates the temporary config.

### Actual behavior
The password is written unquoted, causing silent truncation for any value containing `#` and presumably similar breakage for other special characters. The dump fails with a misleading "Access denied" error even though the password is correct.

### Workaround (currently)
- Change the account password to one without `#` / `;` / quotes / backslash, or
- Use a dedicated dump user with a "clean" password.

**This should not be required: strong passwords legitimately contain special characters.**

### Suggestion
When generating the temporary `.my.cnf` for `mysqldump/mysql/mysqladmin`, DBeaver should wrap the `password` value in double quotes (and escape any embedded `"` / `\` ), exactly as MySQL's documentation recommends. This would make native-tool dumps work with arbitrary passwords.

### Additional notes
- Full command DBeaver executed (host fix already applied to `127.0.0.1`):
```
/home/code78/.local/share/containers/storage/overlay/.../diff/usr/bin/mysqldump --defaults-file=/tmp/.dbeaver-temp.../.my.cnf --skip-lock-tables --routines --add-drop-table --disable-keys --extended-insert --host=127.0.0.1 --port=3306 landes
```
- Related (closed, "can't reproduce" / "not planned"): #6034, #21694 — both about special characters in passwords being mishandled.

sakshichitnis27:
## Summary

- Quote passwords written to the temporary MySQL option file.
- Escape embedded backslashes and double quotes.
- Add a regression test covering option-file special characters.

## Why

MySQL treats `#` as the start of a comment in unquoted option-file values, which truncates passwords used by native tools such as `mysqldump`.

Closes dbeaver/dbeaver#41620

## Validation

- `MySQLNativeToolHandlerTest`
- MySQL test fragment: 5 tests passed
- Headless reactor: 91 modules passed
- Verified the generated value with MySQL 8.4.10 `my_print_defaults`

This PR was generated with AI assistance (OpenAI Codex) and reviewed by the contributor.


IvanGorshechnikov (dismissed):

IvanGorshechnikov:
Thanks for you contibution

E1izabeth (approved):

IvanGorshechnikov (approved):

IvanGorshechnikov:

PR #41622, opened 2026-07-26
#41488 Handle missing GBase read_defaultstr

### LINKED ISSUE #41488: GBase 8s / Informix: "Routine (read_defaultstr) can not be resolved" when opening table Properties if the routine is missing
### Description

Description:
When connecting DBeaver to a GBase 8s V8.8 database via the official gbasedbtjdbc driver,
querying table data works fine, but opening a table's Properties / Columns tab fails with:
SQL error [IX000]: Routine (read_defaultstr) can not be resolved.
  Routine (read_defaultstr) can not be resolved.
    SQLException
Steps to reproduce:
Create a GBase 8s (Informix-compatible) connection in DBeaver using gbasedbtjdbc (tested 3.3.0 / 3.5.1 / 3.6.3).
Connect successfully; plain SELECT on tables works.
Right-click a table → Properties → switch to the Columns tab (column metadata load).
The error is thrown and the Properties panel cannot load.
Environment:
DBeaver version: 26.1.2 (latest)
Database: GBase 8s V8.8
JDBC driver: gbasedbtjdbc_3.3.0, gbasedbtjdbc_3.5.1_3X2_1, gbasedbtjdbc_3.6.3 (all reproduce)
Error origin in log: org.jkiss.dbeaver.model (DBeaver's own metadata layer)
Root cause:
DBeaver's Informix/GBase metadata adapter hard-codes a call to the read_defaultstr() stored procedure
when reading column defaults. However, read_defaultstr is not guaranteed to exist in every GBase 8s V8.8 instance.
Evidence:
SELECT procname FROM sysprocedures WHERE LOWER(procname)='read_defaultstr'; returns 0 rows in the connected DB.
GBase's own native client tools (dbaccess / GBaseDataStudio native path) open the same table's properties
without any error, because they read syscolumns.default directly and never call read_defaultstr.
Toggling DBeaver metadata options (Generate DLLs with extra info, Do not read tables information, etc.) does not help.
The database cannot be modified (security policy), so creating the routine server-side is not an option.
Expected behavior:
DBeaver should not hard-fail the entire Properties panel when read_defaultstr is unresolvable.
It should either:
Read the column default directly from syscolumns.default (consistent with GBase's own tooling), or
Wrap the read_defaultstr call in a try/catch and gracefully degrade (show empty/default value) instead of blocking the whole panel.
Suggested fix:
Make the column-default resolution in the Informix/GBase meta model resilient:
Prefer reading syscolumns.default directly, OR
Fall back to it / leave the default blank when read_defaultstr cannot be resolved,
rather than surfacing IX000 -674 and preventing table properties from opening.
Workarounds tried (none sufficient):
Switched JDBC driver versions (3.3.0 → 3.5.1 → 3.6.3): same error.
Disabled DBeaver metadata options: no effect.
Cannot create the routine on the server (security policy).
Current usable workaround: query syscolumns.default directly via SQL — but this should not be necessary.

### DBeaver Version

26.1.2

### Operating System

window 11

### Database and driver

gbase 8s  V8.8

### Steps to reproduce

_No response_

### Additional context

_No response_

sakshichitnis27:
## Summary

- Detect whether the GBase read_defaultstr routine is available before building the column metadata query.
- Preserve expression-default decoding when the routine exists.
- Leave expression defaults blank when the routine is unavailable so table properties can still load.

## Root cause

The GBase metadata query always referenced read_defaultstr. Some GBase 8s V8.8 installations do not provide that routine, so the database rejected the complete metadata query and DBeaver could not open the Columns tab.

## Validation

- Reproduced the reported Routine (read_defaultstr) can not be resolved failure with GBase 8s lacking the routine.
- Verified the fallback metadata query completes and retains literal defaults.
- Verified the existing expression-default path against a GBase version that provides read_defaultstr.
- Built and packaged org.jkiss.dbeaver.ext.gbase8s successfully.
- Ran git diff --check.

Fixes #41488

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.gbase8s/src/org/jkiss/dbeaver/ext/gbase8s/model/meta/GBase8sMetaModel.java
+++ b/plugins/org.jkiss.dbeaver.ext.gbase8s/src/org/jkiss/dbeaver/ext/gbase8s/model/meta/GBase8sMetaModel.java
@@ -58,6 +58,7 @@ public class GBase8sMetaModel extends GenericMetaModel {
     private static final String[] VALID_TABLE_TYPES = { "T", "V" };
     // TABLE_TYPE: String {@code =>} table type. ('SYNONYM')
     private static final String[] VALID_SYNONYM_TYPES = { "S" };
+    private Boolean readDefaultStrAvailable;
 
     public GBase8sMetaModel() {
         super();
@@ -392,6 +393,9 @@ public class GBase8sMetaModel extends GenericMetaModel {
                 : JDBCUtils.escapeWildCards(session, owner.getSchema().getName());
         boolean isOracleMode = GBase8sUtils.isOracleSqlMode(owner.getDataSource().getContainer());
         String ownerPattern = "%s%s".formatted(isOracleMode ? schema : catalog, isOracleMode ? "." : ":");
+        String expressionDefault = hasReadDefaultStr(session)
+                ? "read_defaultstr(c.tabid, c.colno, c.coltype, c.collength, c.extended_id)::VARCHAR(32731)"
+                : "null::VARCHAR(32731)";
         String sql = """
                 SELECT
                     t.tabname::VARCHAR(128) AS TABLE_NAME,
@@ -404,7 +408,7 @@ public class GBase8sMetaModel extends GenericMetaModel {
                     schema_numprecradix(c.coltype)::INTEGER AS NUM_PREC_RADIX,
                     CASE d.type
                         WHEN 'L' THEN get_default_value(c.coltype, c.extended_id, c.collength, d.default::lvarchar(256))::VARCHAR(254)
-                        WHEN 'E' THEN read_defaultstr(c.tabid, c.colno, c.coltype, c.collength, c.extended_id)::VARCHAR(32731)
+                        WHEN 'E' THEN %s
                         WHEN 'C' THEN 'current' || replace(get_colname(c.coltype, c.collength, c.extended_id, 1), schema_coltypename(c.coltype, c.extended_id), '')::VARCHAR(254)
                         WHEN 'S' THEN 'dbservername'::VARCHAR(254)
                         WHEN 'U' THEN 'user'::VARCHAR(254)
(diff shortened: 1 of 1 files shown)


IvanGorshechnikov (dismissed):

IvanGorshechnikov:
Thanks for your contribution

E1izabeth (dismissed):

E1izabeth:
Thanks for the contribution! Passed to QA team for testing

plugins/org.jkiss.dbeaver.ext.gbase8s/src/org/jkiss/dbeaver/ext/gbase8s/model/meta/GBase8sMetaModel.java
```diff
@@ -392,6 +392,9 @@ public JDBCStatement prepareTableColumnLoadStatement(
                 : JDBCUtils.escapeWildCards(session, owner.getSchema().getName());
         boolean isOracleMode = GBase8sUtils.isOracleSqlMode(owner.getDataSource().getContainer());
         String ownerPattern = "%s%s".formatted(isOracleMode ? schema : catalog, isOracleMode ? "." : ":");
+        String expressionDefault = hasReadDefaultStr(session)
```
ShadelessFox, line 396: You're issuing an additional query for each table, basically. Please cache the result of `hasReadDefaultStr` and reuse it.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.ext.gbase8s/src/org/jkiss/dbeaver/ext/gbase8s/model/meta/GBase8sMetaModel.java
sakshichitnis27: Thank You @ShadelessFox ,
I’ve cached the hasReadDefaultStr result in GBase8sMetaModel and synchronized the initial check, so sysprocedures is queried at most once per data source and reused for subsequent table metadata loads. Addressed in commit 4b6a6c80d3.

ShadelessFox (dismissed):

E1izabeth (dismissed):

E1izabeth:
@sakshichitnis27 please address the review comment and fix commit messages. They should start with a prefix #41488

plugins/org.jkiss.dbeaver.ext.gbase8s/src/org/jkiss/dbeaver/ext/gbase8s/model/meta/GBase8sMetaModel.java
```diff
@@ -437,12 +441,26 @@ public JDBCStatement prepareTableColumnLoadStatement(
...
+                    SELECT FIRST 1 1
+                    FROM sysprocedures
+                    WHERE procname = 'read_defaultstr'
+                    """);
+                 JDBCResultSet dbResult = dbStat.executeQuery()) {
+                this.readDefaultStrAvailable = dbResult.next();
+            }
+        }
+        return this.readDefaultStrAvailable;
+    }
```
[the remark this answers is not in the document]
sakshichitnis27: Addressed in f9ef9bfb. The capability check now matches `LOWER(procname) = 'read_defaultstr'`, requires `isproc = 'f'`, and catches `SQLException`, caching `false` and logging at debug level so column metadata loading continues when `sysprocedures` is inaccessible. The existing expression-default path remains unchanged when the function is available. I also squashed the PR to a single commit with the required `#41488` prefix.

sakshichitnis27:
@E1izabeth @ShadelessFox The remaining review feedback is addressed in f9ef9bfb: the routine check is case-insensitive, verifies `isproc = 'f'`, and fails closed with debug logging on `SQLException`. The PR history is now a single commit with the required `#41488` prefix, and both review threads are resolved. The refreshed GitHub Actions runs are awaiting maintainer approval for this fork update. Please re-review when convenient.

sakshichitnis27:
@E1izabeth @ShadelessFox I refreshed the single commit onto current `devel` (`77f63935`) without changing the one-file functional scope. I also reproduced the upstream build locally using a freshly built `dbeaver-deps-ce` P2 repository: the complete 204-module `mvn clean verify` reactor passed, including `org.jkiss.dbeaver.ext.gbase8s`, all OSGi/unit-test bundles, CE product assembly, and P2 repositories (`BUILD SUCCESS`, 22:39). The previous hosted failure was the external `org.jkiss.bundle.gis:2.0.9` SHA-512 mismatch from `repo.dbeaver.net`, not a Java/source failure. The newly triggered workflows currently show `action_required` and need maintainer approval to run. Please approve the workflows and re-review when convenient.

ShadelessFox (approved):

E1izabeth (approved):

PR #41648, opened 2026-07-30
dbeaver/pro#9795:fix properties refresh

SolovevAnton:
closes dbeaver/pro#9795 . 
Bug description: 
Properties are accessed first during project creation. However, in the project import scenario, properties are populated after this step.
This results in properties being cached as empty in the project.
So we force reload properties, right after internal project creation


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/impl/app/BaseProjectImpl.java
```diff
@@ -478,7 +478,9 @@ public void moveResourceProperties(@NotNull String oldResourcePath, @NotNull Str
 
     @Override
     public void refreshProject(DBRProgressMonitor monitor) {
-
+        synchronized (this) {
```
IvanGorshechnikov:
```suggestion
        synchronized (metadataSync) {
```
Maybe this way is better

IvanGorshechnikov (dismissed):

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.model.rcp/src/org/jkiss/dbeaver/model/rcp/EclipseWorkspaceImpl.java
```diff
@@ -316,6 +317,8 @@ public void resourceChanged(IResourceChangeEvent event) {
                                     activeProject = projectMetadata;
                                     fireActiveProjectChange(null, activeProject);
                                 }
+                                // ensures properties are reloaded if project is being imported
+                                projectMetadata.refreshProject(new VoidProgressMonitor());
```
serge-rider: In fact we do not need monitor in refreshProject implementations. Just delete this param.

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model.rcp/src/org/jkiss/dbeaver/model/rcp/EclipseWorkspaceImpl.java
SolovevAnton: fixed

E1izabeth (approved):

PR #41650, opened 2026-07-30
Dbeaver/pro#4738 run script

### LINKED ISSUE #4738: Run selected script - without opening editor
Some scripts are very big and editor will take long time to load into UI -  for such scripts we save in drive and run them from workbench or some other tool - just by selecting the file and running it without loading it to editor.

Can we get that feature -  probably in tools or some relevant location when we right click on the database name -  to run the sql file on selected connection/DB instead of opening the file in editor.
untitled

Destrolaric:
Closes Dbeaver/dbeaver#4738 Opens an execute Task Wizard, if no datasources selected beforehand opens datasource select menu


IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tasks/ui/sql/script/SQLScriptTaskPageSettings.java
```diff
@@ -456,11 +456,14 @@ public void deactivatePage() {
 
     @Override
     protected boolean determinePageCompletion() {
-        if (selectedScripts.isEmpty()) {
+        SQLScriptExecuteSettings settings = getWizard().getSettings();
+        boolean hasScripts = !selectedScripts.isEmpty() || !CommonUtils.isEmpty(settings.getScriptFiles());
+        boolean hasDataSources = !selectedDataSources.isEmpty() || !CommonUtils.isEmpty(settings.getDataSources());
```
ShadelessFox: Why is this needed to check settings?

plugins/org.jkiss.dbeaver.ui.editors.sql/plugin.xml
```diff
@@ -401,6 +403,25 @@
                 </with>
             </enabledWhen>
         </handler>
+        <handler commandId="org.jkiss.dbeaver.core.sql.script.execute" class="org.jkiss.dbeaver.ui.editors.sql.handlers.SQLEditorHandlerExecuteScriptFile">
+            <activeWhen><reference definitionId="org.jkiss.dbeaver.core.ui.navigator.part"/></activeWhen>
+            <enabledWhen>
+                <and>
+                    <reference definitionId="org.jkiss.dbeaver.core.ui.navigator.part"/>
```
ShadelessFox: Why does it have to be limited to just the navigator? Let it be available anywhere where it's possible to adapt a DBNResource.

Destrolaric: Fixed

plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tasks/ui/sql/script/SQLScriptTaskPageSettings.java
```diff
@@ -403,7 +405,7 @@ public Image getImage(Object element) {
 
     private void refreshScripts() {
         scriptsViewer.refresh(true, true);
-        updateSelectedScripts();
+        updateSelectedScripts   ();
```
ShadelessFox:
```suggestion
        updateSelectedScripts();
```

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/handlers/SQLEditorHandlerExecuteScriptFile.java
ShadelessFox, line 1: Please reformat this file. It has multiple issues with the closing `)` placed in the wrong place.

Matvey16:
1. The wizard doesn't see variables, it ignores it and just replaces every variable with NULL.

1.1 `Show variables before run` also does nothing

2. There is no results shown anywhere. Wizard is immidiately closed after pressing proceed and nothing is recorded anywhere

image

IvanGorshechnikov (dismissed):

PR #41652, opened 2026-07-30
dbeaver/pro#8963: fix children refresh

SolovevAnton:
closes dbeaver/pro#8963

The fix defers context initialization and database metadata loading until refresh completes
without it - npe was cought on refresh on some postgres instances


E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNDatabaseNode.java
```diff
@@ -220,7 +220,7 @@ public boolean hasChildren(@NotNull DBRProgressMonitor monitor, @NotNull DBXTree
     public DBNDatabaseNode[] getChildren(@NotNull DBRProgressMonitor monitor) throws DBException {
         boolean needsLoad;
         synchronized (this) {
-            needsLoad = childNodes == null && hasChildren(false);
+            needsLoad = (childNodes == null || childNodes.length == 0) && hasChildren(false);
```
ShadelessFox: If a node is loaded but its children are empty, it must not try to load it again; that's the whole point of caching. Otherwise, we might end up loading the same node over and over again.

ShadelessFox (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNDatabaseNode.java
SolovevAnton: agreed, found core bug

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNDatabaseNode.java
```diff
@@ -497,8 +497,8 @@ private void loadChildren(
                         }
                     }
 
-                    if (oldList == null) {
-                        // Load new folders only if there are no old ones
+                    if (oldList == null || Arrays.stream(oldList).noneMatch(oldFolder -> oldFolder.getMeta() == child)) {
```
ShadelessFox: I don't understand this fix :(

SolovevAnton: the problem was following:
1. In simple mode, some nodes were filtered out (databases folder, scripts folder, etc)
2. We didn't refresh the connection on change -> filtered nodes stayed the same in the navigator
3. We change to advanced mode -> it allows all these folders, and works on the assumption they are present in nodes, and any additional filters are applied recursively, folder by folder
4. None of these folders were found -> all nodes filtered -> empty navigator

this fix says:
if none of currently cahced folders corresprond with meta to parent node, probably they were filtered away before. so lets reload node structure

ShadelessFox: Can't we just forcefully refresh the connection node so its children get invalidated?

SolovevAnton: but it would break the logic.
User specifically said after property change: i dont want to reload the connection
So the connection stays the same 

If we forcefully reload navigator nodes, the connection view will be changed
And with this fix, the connection will change only when and if the navigator is forcefully reloaded by the user.

ShadelessFox (dismissed):

IvanGorshechnikov (dismissed):

ShadelessFox (dismissed):

serge-rider:
Let's refresh (cleanup) entire tree (from DS level). Instead of incremental refresh.
This is a dangerous fix.

E1izabeth (approved):

E1izabeth (approved):

ShadelessFox (approved):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/model/navigator/DBNDataSource.java
```diff
@@ -269,6 +269,9 @@ public boolean moveToFolder(DBPProject project, DBPDataSourceFolder folder)
     @Nullable
     public DBNNode refreshNode(@NotNull DBRProgressMonitor monitor, @Nullable Object source) throws DBException
     {
+        if (!isLocked()) {
+            cleanupNode();
+        }
```
serge-rider, line 274: Again it looks it affects ALL tree refreshes on ds level, no matter how they were triggered.
The idea is to refresh entire tree only when we change connection view.

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDatabase.java
```diff
@@ -118,8 +118,13 @@ protected PostgreDatabase(DBRProgressMonitor monitor, PostgreDataSource dataSour
         // We need to set name first
         this.name = databaseName;
         this.initCaches();
-        checkInstanceConnection(monitor, false);
+        if (!dataSource.isConnectionRefreshing()) {
```
serge-rider, line 121: Why do we need to change PG behavior for this ticket? At least add a comment in code.

serge-rider (changes requested):

PR #41672, opened 2026-08-03
dbeaver/dbeaver#41665 Export binary values to XLSX

### LINKED ISSUE #41665: XLSX export incorrectly exports BINARY expression values as [BINARY]
### Description

When exporting query results to **XLSX**, columns selected using the MySQL `BINARY` operator are exported incorrectly.

For example, assume the `users` table contains:

| country |
|----------|
| India |
| India |
| india |
| USA |

Executing the following query:

```sql
SELECT BINARY country, COUNT(*)
FROM users
GROUP BY BINARY country;
```

produces the following result in DBeaver's result grid:

| country | COUNT(*) |
|----------|---------:|
| India | 2 |
| india | 1 |
| USA | 1 |

However, when exporting the same result set using **Export Data → XLSX**, the first column contains:

| country | COUNT(*) |
|----------|---------:|
| [BINARY] | 2 |
| [BINARY] | 1 |
| [BINARY] | 1 |

instead of the actual values (`India`, `india`, `USA`).

The issue appears to be specific to the **XLSX export**. Exporting the same result set as **CSV** works correctly and preserves the expected values.

### DBeaver Version

Community Edition 26.1.0

### Operating System

Garuda Linux

### Database and driver

- **Database:** MySQL 5.7.37
- **Driver:** MySQL JDBC Driver 8.2.0

### Steps to reproduce

1. Create a `users` table (or use an existing one) containing the sample data shown above.
2. Execute:

   ```sql
   SELECT BINARY country, COUNT(*)
   FROM users
   GROUP BY BINARY country;
   ```

3. Verify that the result grid displays the expected values.
4. Export the result using **Export Data → XLSX**.
5. Open the generated `.xlsx` file.
6. Observe that the first column contains `[BINARY]` instead of the expected country values.

### Additional context

- The issue is reproducible with MySQL.
- The query results are displayed correctly in the DBeaver result grid.
- **CSV export works correctly**; the issue only occurs with **XLSX export**.
- The problem appears to be related to how the XLSX exporter handles expressions using the MySQL `BINARY` operator.

EastLord:
Closes: #41665

Export binary values to XLSX using the selected encoding instead of the [BINARY] placeholder, while respecting Excel’s cell-length limit.

Test SQL:

```sql
SELECT
    BINARY 'India' AS text_binary,
    UNHEX('00FF10') AS raw_binary,
    CAST(CONVERT('印度' USING BINARY) AS BINARY) AS utf8_binary;
```


IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/stream/IStreamDataExporterSite.java
```diff
@@ -53,6 +54,8 @@ public interface IStreamDataExporterSite {
 
     void writeBinaryData(@NotNull DBDContentStorage cs) throws IOException;
 
+    void writeBinaryData(@NotNull DBDContentStorage cs, @NotNull Writer writer) throws IOException;
```
serge-rider: No need to add new function in exporter site.

If you need an utility to reuse in XLSX exporter then just pull this util function out of  StreamTransferConsumer.

serge-rider (changes requested):

EastLord:
@serge-rider Thanks, updated to reuse the utility without adding a new exporter-site method.

E1izabeth:
@codex review

PR #41673, opened 2026-08-03
#39972 Add support for Databricks data types

### LINKED ISSUE #39972: Databricks SQL: ARRAY<STRUCT> — display the array as json
### Description

The new interface for displaying arrays does not handle array of struct properly
When expanding the array the data is displayed as if it was an array of string which makes it very hard to use.
It was MUCH better to just display everything as json.

Image

On DBeaver 24.3.5, It looks like this, which is much clearer:

Image

### DBeaver Version

Community edition Version 25.3.0.202511301802

### Operating System

Windows 11

### Database and driver

Databricks 1.0.9

Image

### Steps to reproduce

Run this query against Databricks:
select array(struct("Hello",1),struct("hopp",2)) as myarray;

When expanding the array , the data is displayed as a badly formatted array of string

Image

### Additional context

_No response_

E1izabeth:
Added support for complex types https://docs.databricks.com/aws/en/sql/language-manual/sql-ref-datatypes#complex-types

For QA: if you find that something complex is not working (for example, some particular combination of nested complex types), please create a separate issue.


plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/model/types/DatabricksMapValueHandler.java
```diff
@@ -0,0 +1,195 @@
...
+        @Override
+        public String toString() {
+            return key + "=" + value;
+        }
+
+        @Override
+        public int compareTo(@Nullable EntryComposite other) {
+            return other == null
+                ? 1
+                : (
```
ShadelessFox: Please use an `if` statement with an early return, otherwise it's pretty much unreadable

plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/model/types/DatabricksMapValueHandler.java
```diff
@@ -0,0 +1,195 @@
...
+                    this.key instanceof Comparable<?> thisKey &&
+                    other.key instanceof Comparable<?> otherKey &&
+                    other.key.getClass() == this.key.getClass()
+                        ? ((Comparable<Object>) thisKey).compareTo(otherKey)
+                        : this.key.getClass().getName().compareTo(other.key.getClass().getName())
```
ShadelessFox: Quite a method to compare potentially non-comparable entities. Why do we want it to be Comparable in the first place?

plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/model/types/DatabricksDataType.java
```diff
@@ -0,0 +1,74 @@
...
+import org.jkiss.dbeaver.model.struct.DBSDataType;
+
+import java.sql.Types;
+import java.util.Map;
+import java.util.Objects;
+
+public class DatabricksDataType extends GenericDataType {
+
+    @Nullable
+    private final Integer srid;
```
ShadelessFox: Feels like it should be in a dedicated subclass rather than here in the base class that represents all data types.

plugins/org.jkiss.dbeaver.ext.databricks/src/org/jkiss/dbeaver/ext/databricks/model/types/DatabricksMapValueHandler.java
E1izabeth: Databricks documentation says maps are unordered, but it might be unusable to observe them in the UI this way. So we're just sorting map entries by key when we receive them, as well as possible.

ShadelessFox (approved):

SolovevAnton (approved):

Matvey16:
Verified

DBeaver CE Windows
Dbeaver EE Windows

PR #41695, opened 2026-08-05
#41509 Ensure changes to search_path are committed

### LINKED ISSUE #41509: DBeaver forgets the currently selected Postgres schema upon hitting a query error under manual commit mode
### Description

A video that demonstrates the issue:

https://github.com/user-attachments/assets/5994babb-6c5b-4ee3-bb78-1a77c3630237

The default schema for this database is `public`, which has no tables. 

Running `show search_path` with the same steps:

https://github.com/user-attachments/assets/c67d49c4-ce5a-4b5d-a8fe-93e23faf448e

As far as I recall this had been happening for quite a long time (probably several years), but I only managed to find a way to reliably reproduce it recently. In memory, this also happens when the connection is invalidated due to a network issue.  

### DBeaver Version

Community Edition 26.1.2.202607051737

### Operating System

Windows 11

### Database and driver

Database name: PostgreSQL 15.18
Database driver: PostgreSQL JDBC Driver 42.7.11

### Steps to reproduce

_No response_

### Additional context

_No response_

ShadelessFox:
- [ ] Test whether the search_path doesn't change when opening a new editor and running a faulty query with manual commit enabled
- [ ] Test whether it affects transactions made in other editors when this option is set to "Always":
image

- [ ] Changing the active catalog/schema from the editor will still result in the original issue (and it should not commit changes made in other editors, see the previous item). This PR does not fix it

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
@@ -396,14 +396,11 @@ public class PostgreDataSource extends JDBCDataSource implements DBSInstanceCont
         if (activeSchema != null) {
             postgreContext.setDefaultCatalog(monitor, activeSchema.getDatabase(), activeSchema, true);
 
-            // Commit possible changes to search_path
-            var txnManager = DBUtils.getTransactionManager(context);
-            if (txnManager != null && !txnManager.isAutoCommit()) {
-                try (DBCSession session = context.openSession(monitor, DBCExecutionPurpose.UTIL, "Commit transaction")) {
-                    txnManager.commit(session);
-                } catch (DBCException e) {
-                    log.warn("Error committing changes context defaults for new connection", e);
-                }
+            try {
+                // Commit possible changes to search_path
+                DBExecUtils.commitContextTransaction(monitor, context);
+            } catch (DBCException e) {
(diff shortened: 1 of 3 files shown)


E1izabeth (dismissed):

SolovevAnton (dismissed):

mariiachugunova:

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
```diff
@@ -395,6 +395,16 @@ protected void initializeContextState(@NotNull DBRProgressMonitor monitor, @NotN
         postgreContext.refreshDefaults(monitor, true);
         if (activeSchema != null) {
             postgreContext.setDefaultCatalog(monitor, activeSchema.getDatabase(), activeSchema, true);
+
+            // Commit possible changes to search_path
+            var txnManager = DBUtils.getTransactionManager(context);
```
serge-rider: Let's avoid copy-paste and move this code into DBExecUtils.commitContextTransaction

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/model/PostgreDataSource.java
ShadelessFox: Should've done that like 5 years ago xd

PR #41696, opened 2026-08-05
dbeaver/dbeaver#41548 Refresh table property colors after connection …

### LINKED ISSUE #41548: Connection type color is not applied to the entire Table Properties editor after changing connection settings
### Description

After changing the Connection Type color in the connection settings, the new color is only applied to some parts of the already opened Table Properties editor. Several UI areas keep the previous color even after refreshing the editor.

### DBeaver Version

CE 26.1.2

### Operating System

Windows 11 pro

### Database and driver

_No response_

### Steps to reproduce

1. Create a connection and assign a custom Connection Type color.
2. Open any table.
3. Open the Properties tab of the table editor.
4. Right-click the connection and select Edit Connection.
5. Go to the General tab.
6. Change the Connection Type color.
7. Click Apply and Close.
8. Return to the already opened Table Properties editor.
9. Click Refresh.

**Actual result:**
The updated connection color is applied only to part of the editor. Some UI elements (such as the background of individual panels) keep the previous color even after refreshing.

**Expected result:**
The new connection type color should be applied consistently to the entire Table Properties editor immediately after clicking Apply and Close, or at least after refreshing the editor.

Image

### Additional context

_No response_

EastLord:
Closes: #41548

Refresh open Table Properties when the connection type color changes, including labels and checkboxes.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/properties/ObjectPropertiesEditor.java
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/editors/entity/properties/ObjectPropertiesEditor.java
@@ -41,6 +41,10 @@ import org.jkiss.dbeaver.Log;
 import org.jkiss.dbeaver.ModelPreferences;
 import org.jkiss.dbeaver.model.DBIcon;
 import org.jkiss.dbeaver.model.DBPDataSource;
+import org.jkiss.dbeaver.model.DBPDataSourceContainer;
+import org.jkiss.dbeaver.model.DBPEvent;
+import org.jkiss.dbeaver.model.DBPEventListener;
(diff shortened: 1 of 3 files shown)


IvanGorshechnikov (dismissed):

IvanGorshechnikov:
Thanks for your contribution!

E1izabeth (dismissed):

EastLord:
I tracked this down to two issues. The color change event contains the datasource registry in event.data, but the editor was only checking event.object, so the update was ignored. The folder list also kept its previously calculated background colors.
The fix now handles registry update events and recalculates the folder colors before redrawing the editor. The new color should appear immediately after saving, without clicking inside the Properties page.

EastLord:
https://github.com/user-attachments/assets/857a8e00-43a5-4ec9-a913-6a0278375a11

Destrolaric (approved):

EastLord:
Rebased onto the latest devel, resolved the conflict

PR #41700, opened 2026-08-06
#41342 Fix clickhouse maps handling

### LINKED ISSUE #41342: Maps don't show actual value but "normalized one
### Description

connecting to clickhouse with DBeaver a query returning the result set does not show the actual values but a "normalized one"

the clickhouse map value contains twice "abc"="a" which is returned when querying via clickhouse play ui

beaver is showing just a single abc=a within the ui most likely through usage of a map class that deduplicates values.

### DBeaver Version

Version 26.1.0.202605311718

### Operating System

_No response_

### Database and driver

clickhouse driver

### Steps to reproduce

create a db with column type "Map" as used in clickhouse otel db schema

insert a string like 'abc=a, abc=a' in the column

query table 

run same query on clickhouse/play internal ui and validate results 

DBeaver does only return a single occurance of the abc=a in the result set

### Additional context

_No response_

E1izabeth:
Closes https://github.com/dbeaver/pro/issues/10217
Please test different combinations of maps and nested complex types.
For example:

```
CREATE TABLE `default`.`example_map_table`
(
    `id` UInt64,
    `attributes` Map(String, String)
)
ENGINE = MergeTree
ORDER BY `id`;
    
INSERT INTO `default`.`example_map_table` (`id`, `attributes`)
VALUES
    (1, map('color', 'blue', 'size', 'medium')),
    (2, map('color', 'red', 'size', 'large', 'in_stock', 'true')),
    (3, map('category', 'electronics', 'brand', 'example'));

select * from `default`.`example_map_table`;

select array(1,2,3);

select array(tuple('a','b',3),tuple('c','d',2));

select tuple('a','b',3);

select tuple('a','b',tuple('c','d',2));

-----------------------------

select tuple(array('a','b'),array(1,2,3),2,1);

select tuple(1,2,array('a','b'),array(1,2,3));

select tuple(array(array(1,2,3),array(4,5,6)))

select tuple(array(tuple('x','y'),tuple('y','x')),array(1,2,3),2,1);
```


SolovevAnton (approved):

ShadelessFox (approved):

Matvey16:
This fixes a few clickhouse issues mainly with tuples, but the original bug from the ticket is still present

image

Matvey16:
Verified

DBeaver CE Windows
Clichouse 22.1
Clichouse 26.5

PR #41707, opened 2026-08-07
dbeaver/dbeaver#41705 Preserve raw SHOW CREATE DDL output for StarRocks and Doris

### LINKED ISSUE #41705: Preserve raw SHOW CREATE DDL output for StarRocks and Doris
### Description
When viewing DDL for StarRocks or Apache Doris tables/views in DBeaver, the DDL returned by `SHOW CREATE TABLE` / `SHOW CREATE VIEW` is passed through DBeaver's generic SQL formatter.
This may change the database-provided DDL formatting unexpectedly. In particular, depending on the user's SQL formatter preferences, commas can be moved to separate lines or placed before line breaks, making the displayed DDL harder to read and different from the raw `SHOW CREATE` output.
For StarRocks and Doris, the database already returns the canonical DDL from `SHOW CREATE`, so DBeaver should display that DDL as-is instead of applying a second formatting pass.
### Affected drivers
- StarRocks
- Apache Doris
### Current behavior

Image

The StarRocks and Doris DDL loading utilities call:
```java
SQLFormatUtils.formatSQL(table.getDataSource(), definition)
```

As a result, raw DDL returned by the database can be modified by DBeaver's SQL formatter.

### Expected behavior
DBeaver should display the raw DDL returned by SHOW CREATE TABLE / SHOW CREATE VIEW for StarRocks and Doris without applying SQLFormatUtils.formatSQL(...). This would be consistent with how some other drivers display database-provided SHOW CREATE output, for example MySQL.

Image

### DBeaver Version

26.1.4

### Operating System

macOS Tahoe 26.5.2

### Database and driver

StarRocks Connector/J starrocks-connector-j-1.1.2
Doris: MySQL Connector/J mysql-connector-j-8.2.0 

### Steps to reproduce

1. Enable the SQL formatter option "line feed before comma".
2. Connect to a StarRocks or Doris database.
3. Open a table DDL/source view.
4. Observe that commas in the DDL are moved to separate lines.

### Additional context

This issue is different from cases where a database returns a single-line DDL and DBeaver needs to format it. For StarRocks and Doris, the DDL comes from explicit SHOW CREATE commands and should be treated as database-generated canonical output.

dupen01:
## Description
This PR preserves the raw `SHOW CREATE` DDL output for StarRocks and Apache Doris.
Previously, the DDL returned by `SHOW CREATE TABLE` / `SHOW CREATE VIEW` was passed through DBeaver's generic SQL formatter via `SQLFormatUtils.formatSQL(...)`. This could unexpectedly change the database-provided DDL formatting depending on local SQL formatter preferences, for example by moving commas to separate lines.
For StarRocks and Doris, the database already returns the canonical DDL from `SHOW CREATE`, so this PR returns that DDL directly without applying an additional formatting pass.
## Changes
- Removed `SQLFormatUtils.formatSQL(...)` from StarRocks `SHOW CREATE` DDL loading.
- Removed `SQLFormatUtils.formatSQL(...)` from Doris `SHOW CREATE` DDL loading.
- Updated method comments to reflect that DDL is loaded directly rather than formatted.
## Affected drivers
- StarRocks
- Apache Doris

## Related issue
Closes dbeaver/dbeaver#41705

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.doris/src/org/jkiss/dbeaver/ext/doris/DorisUtils.java
+++ b/plugins/org.jkiss.dbeaver.ext.doris/src/org/jkiss/dbeaver/ext/doris/DorisUtils.java
@@ -30,7 +30,6 @@ import org.jkiss.dbeaver.model.exec.jdbc.JDBCResultSet;
 import org.jkiss.dbeaver.model.exec.jdbc.JDBCSession;
 import org.jkiss.dbeaver.model.impl.jdbc.JDBCUtils;
 import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
-import org.jkiss.dbeaver.model.sql.format.SQLFormatUtils;
 
 import java.sql.SQLException;
 import java.sql.Statement;
@@ -59,14 +58,14 @@ public class DorisUtils {
     }
 
     /**
-     * Loads and formats DDL using SHOW CREATE commands for Doris tables and views.
+     * Loads DDL using SHOW CREATE commands for Doris tables and views.
      *
      * @param monitor progress monitor
      * @param table the table or view to load DDL for
      * @param sessionTitle title for the metadata session
      * @param showCommand the SHOW CREATE command (e.g., "SHOW CREATE TABLE", "SHOW CREATE VIEW")
      * @param columnName the result column name (e.g., "Create Table", "Create View")
-     * @return formatted DDL string, or null if not found
+     * @return DDL string, or null if not found
      * @throws DBCException if DDL loading fails
      */
     @Nullable
@@ -90,7 +89,7 @@ public class DorisUtils {
                     if (dbResult.next()) {
                         String definition = JDBCUtils.safeGetString(dbResult, columnName);
                         if (definition != null) {
-                            return SQLFormatUtils.formatSQL(table.getDataSource(), definition);
+                            return definition;
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


E1izabeth (commented):
There is another way to generate DDL in the UI `(right-click table in the navigator tree) --> Generate SQL-->DDL`, where DDL generator dialog window has an optional checkbox `Format SQL`.
Please validate that this modification doesn't break it. Maybe even fixes. 
If the `Format SQL` checkbox doesn't work with this fix, then the fix might need to be enhanced.

dupen01:
> There is another way to generate DDL in the UI `(right-click table in the navigator tree) --> Generate SQL-->DDL`, where DDL generator dialog window has an optional checkbox `Format SQL`. Please validate that this modification doesn't break it. Maybe even fixes. If the `Format SQL` checkbox doesn't work with this fix, then the fix might need to be enhanced.

Thanks for pointing this out.
I built the project locally and validated the `Generate SQL -> DDL` flow as well. The behavior looks correct with this change. I validated this with StarRocks and confirmed the same behavior for Doris.
1. In the table properties `DDL` tab, the DDL is displayed exactly as returned by `SHOW CREATE TABLE` / `SHOW CREATE VIEW`. It is not implicitly reformatted by DBeaver anymore.
image


2. In the navigator tree `Generate SQL -> DDL` dialog, when `Format SQL` is unchecked, the generated DDL keeps the raw database-provided format.
image


3. When `Format SQL` is checked in the same dialog, the generated DDL is still passed through the SQL formatter and gets formatted as expected.
image


So this change preserves the raw DDL in the object source view while keeping the explicit `Format SQL` option in the DDL generator dialog working.

E1izabeth:
Thanks

E1izabeth (approved):

ShadelessFox (approved):
Thanks, passing to QA

mariiachugunova:

PR #41711, opened 2026-08-07
dbeaver/pro#9992 [CB] rename field for OpenAI configuration:

dgr-dbeaver:
Endpoint -> API Base URL


dgr-dbeaver:
**Self checked, testing is not required**
The fields has already exists but there was incorrect label. Renamed it.
image

yagudin10 (approved):

E1izabeth (approved):

PR #41712, opened 2026-08-07
dbeaver/dbeaver#4979 Decompression of blobs

### LINKED ISSUE #4979: Decompression of blobs

Divahar2507:
Closes dbeaver/dbeaver#4979 Decompression of blobs

// Description
Adds decompression support (GZIP and ZLIB formats) for binary BLOB columns in DBeaver's data editor Value View Panel.

// Changes Included
- Added `GzipStreamValueManager` implementing `IStreamValueManager` to detect GZIP (`0x1F 0x8B`) and ZLIB (`0x78`) headers.
- Added `GzipPanelEditor` extending `TextPanelEditor` to decompress BLOB bytes via `GZIPInputStream` / `InflaterInputStream` and render decompressed text.
- Registered `<streamManager>` under `org.jkiss.dbeaver.dataManager` in `plugin.xml`.

Closes dbeaver/dbeaver#4979


IvanGorshechnikov:
Thanks for a contribution!

You have a merge conflict
Also, could you add some footage of your feature?

ShadelessFox:
@Divahar2507 please address comments before we can move on.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/stream/GzipPanelEditor.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.zip.GZIPInputStream;
+import java.util.zip.InflaterInputStream;
+
+/**
+ * GZIP/ZLIB panel editor
+ */
+public class GzipPanelEditor extends TextPanelEditor {
+
+    private static final Log log = Log.getLog(GzipPanelEditor.class);
+    private static final long MAX_DECOMPRESSED_SIZE = 50 * 1024 * 1024; // 50 MB limit against decompression bombs
```
ShadelessFox: It definitely needs to be configurable. We already have a similar preference for text data, and since this editor is for compressed (text) data, it can be used here as well. See `Editors | Data Editor | Binary Editor` under `Content` group `Maximum text editor content length`

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/stream/GzipPanelEditor.java
```diff
@@ -0,0 +1,133 @@
...
+            return;
+        }
+
+        try {
+            ByteArrayOutputStream rawBuffer = new ByteArrayOutputStream();
+            try (InputStream is = storage.getContentStream()) {
+                if (is != null) {
+                    ContentUtils.copyStreams(is, -1, rawBuffer, monitor);
+                }
+            }
```
ShadelessFox: You don't check the size of the input stream. It can totally be several GBs, which would crash DBeaver. Read some leading bytes to figure out the format, then feed the rest into a matching `OutputStream` implementation.

plugins/org.jkiss.dbeaver.ui.editors.data/src/org/jkiss/dbeaver/ui/data/managers/stream/GzipStreamValueManager.java
```diff
@@ -0,0 +1,79 @@
...
+                                int b1 = header[1] & 0xFF;
+                                // GZIP magic header: 0x1F, 0x8B
+                                if (b0 == 0x1F && b1 == 0x8B) {
+                                    return MatchType.PRIMARY;
+                                }
+                                // ZLIB magic header check (RFC 1950)
+                                if ((header[0] & 0x0F) == 8 && ((b0 << 8) | b1) % 31 == 0) {
+                                    return MatchType.PRIMARY;
+                                }
+                            }
```
ShadelessFox, line 50: Please move to `GzipPanelEditor`

ShadelessFox (changes requested):

ShadelessFox (approved):

PR #41713, opened 2026-08-07
dbeaver/dbeaver#41706 Fix DDL normalization for ClickHouse views

### LINKED ISSUE #41706: ClickHouse: StringIndexOutOfBoundsException when generating DDL for a view
### Description

`ClickhouseMetaModel.normalizeDDL()` assumes the statement returned by `SHOW CREATE TABLE` always contains both `(` and `) ENGINE`. `getViewDDL()` delegates to `getTableDDL()`, so a plain view reaches the same code — and a view definition frequently has neither marker.

`plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseMetaModel.java` (line 182 on `devel` @ 79c29fb):

```java
private String normalizeDDL(String ddl) {
    int declStart = ddl.indexOf("(");
    int declEnd = ddl.indexOf(") ENGINE");
    if (declEnd == -1) {
        declEnd = ddl.length() - 1;
    }
    return
        ddl.substring(0, declStart) + "(\n" +
        ddl.substring(declStart + 1, declEnd).replace(",", ",\n") + "\n" +
        ddl.substring(declEnd);
}
```

When the statement contains no `(` at all, `declStart` is `-1` and `ddl.substring(0, -1)` throws. The `declEnd == -1` fallback is also wrong: it splits the final character off the statement and re-appends it after an inserted newline.

I ran the current implementation against four inputs:

| Input | Current result |
|---|---|
| ``CREATE TABLE db.t (`a` Int32, `b` String) ENGINE = MergeTree ORDER BY a`` | correct |
| `CREATE VIEW db.v AS SELECT id FROM db.t` | **StringIndexOutOfBoundsException: Range [0, -1) out of bounds for length 39** |
| `CREATE VIEW db.v AS SELECT count(id), toInt32(x) FROM db.t` | `... FROM db.\nt` — a newline is inserted in the middle of the identifier |
| ``CREATE TABLE db.t\n(\n    `a` Int32\n)\nENGINE = MergeTree`` (already formatted, so `") ENGINE"` does not match) | `ENGINE = MergeTre\ne` — last character split off |

Three of the four cases are broken.

Suggested fix — skip normalization when there is no real column declaration block, which leaves the working table case byte-identical:

```java
int declStart = ddl.indexOf("(");
int declEnd = ddl.indexOf(") ENGINE");
if (declStart == -1 || declEnd == -1 || declStart > declEnd) {
    // Not a column declaration block (e.g. plain view definition) - leave it as is
    return ddl;
}
```

I have a fix with unit tests covering all four cases above and would be happy to open a pull request.

### DBeaver Version

Community Edition 26.1.5 (devel branch, commit 79c29fb)

### Operating System

Ubuntu 24.04.4 LTS

### Database and driver

1. Database name and version: ClickHouse
2. Driver name and version: ClickHouse (com.clickhouse.jdbc.ClickHouseDriver)

### Steps to reproduce

1. Connect to a ClickHouse server.
2. Create a view whose definition contains no parentheses, e.g.
     CREATE VIEW default.v AS SELECT id FROM default.t;
3. In the navigator, open the view and switch to the DDL tab
   (or right-click -> Generate SQL -> DDL).
4. The DDL fails to render; the log shows
   java.lang.StringIndexOutOfBoundsException: Range [0, -1) out of bounds for length N
   from ClickhouseMetaModel.normalizeDDL.

A view that does contain parentheses does not throw, but the returned DDL is corrupted -
a newline is inserted at the first "(" regardless of where it falls.

### Additional context

getViewDDL() delegates to getTableDDL(), so views and tables share the same
normalization path. System-schema objects take an early return and are unaffected.

SEPURI-SAI-KRISHNA:
`ClickhouseMetaModel.normalizeDDL()` assumed the statement returned by `SHOW CREATE TABLE` always contains both `(` and `) ENGINE`. Since `getViewDDL()` delegates to `getTableDDL()`, a view definition without parentheses reached it and `ddl.substring(0, -1)` threw `StringIndexOutOfBoundsException`. The `declEnd == -1` fallback additionally split the final character off the statement.

Normalization is now skipped when there is no real column declaration block, so the DDL is returned unchanged instead of being mangled. The working table case is byte-identical to before.

Added `ClickhouseMetaModelDDLTest` covering four cases: a normal table declaration, a view without parentheses, a view whose query contains parentheses, and an already-multi-line-formatted declaration. The last three all failed on `devel` — one threw, two silently corrupted the SQL.

`normalizeDDL` was made package-private and annotated `@ForTest` so the test fragment can call it directly.

Closes dbeaver/dbeaver#41706

This PR was generated with AI assistance (Claude Code); the change and its rationale have been reviewed by me.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseMetaModel.java
+++ b/plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseMetaModel.java
@@ -35,7 +35,6 @@ import org.jkiss.dbeaver.model.exec.jdbc.JDBCStatement;
 import org.jkiss.dbeaver.model.impl.jdbc.cache.JDBCBasicDataTypeCache;
 import org.jkiss.dbeaver.model.impl.jdbc.struct.JDBCDataType;
 import org.jkiss.dbeaver.model.impl.sql.QueryTransformerLimit;
-import org.jkiss.dbeaver.model.meta.ForTest;
 import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
 import org.jkiss.dbeaver.model.struct.DBSObject;
 import org.jkiss.utils.CommonUtils;
@@ -180,7 +179,6 @@ public class ClickhouseMetaModel extends GenericMetaModel implements DBCQueryTra
         return false;
     }
 
-    @ForTest
     static String normalizeDDL(String ddl) {
         int declStart = ddl.indexOf("(");
         int declEnd = ddl.indexOf(") ENGINE");
--- a/plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseMetaModel.java
+++ b/plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseMetaModel.java
@@ -35,6 +35,7 @@ import org.jkiss.dbeaver.model.exec.jdbc.JDBCStatement;
 import org.jkiss.dbeaver.model.impl.jdbc.cache.JDBCBasicDataTypeCache;
 import org.jkiss.dbeaver.model.impl.jdbc.struct.JDBCDataType;
 import org.jkiss.dbeaver.model.impl.sql.QueryTransformerLimit;
+import org.jkiss.dbeaver.model.meta.ForTest;
 import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
 import org.jkiss.dbeaver.model.struct.DBSObject;
 import org.jkiss.utils.CommonUtils;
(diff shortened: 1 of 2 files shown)


E1izabeth (dismissed):

E1izabeth:
Thanks for the contribution!

ShadelessFox (dismissed):
Thanks, passing to QA

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseMetaModel.java
```diff
@@ -179,11 +180,13 @@ public boolean supportsTableDDLSplit(@NotNull GenericTableBase sourceObject) {
         return false;
     }
 
-    private String normalizeDDL(String ddl) {
+    @ForTest
```
serge-rider: It is not supposed to be here, it is not just for tests

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/ClickhouseMetaModel.java
SEPURI-SAI-KRISHNA: You're right — `normalizeDDL` is real logic on the `getTableDDL` path, not something that exists for tests, so the annotation was misleading. Every other `@ForTest` in the codebase marks a genuinely test-only member (the `PostgreDataType`/`OracleDataSource` constructors, `GenericTableBase.getCachedAttributes`), and this isn't one of those.

Removed the annotation and its import. The method is still package-private rather than `private` so the test fragment can call it directly, but there's no longer anything claiming it's test-only. Nothing else changed — the guard and the tests are the same.

E1izabeth (approved):

PR #41726, opened 2026-08-10
dbeaver/pro#9973 Add confirmation for script connection binding reading

E1izabeth:
Since now this dialog appears
image

and the following dialog to double-confirm if you check Don't ask again 
image


IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -526,23 +527,94 @@ private boolean updateDataSourceContainer() {
...
+                } else {
+                    result = DataSourceBindingOperation.USE_NOW;
+                }
+            } else {
+                result = dontAskAnymore ? DataSourceBindingOperation.CANCEL_AND_REMEMBER : DataSourceBindingOperation.CANCEL_NOW;
+            }
+        } else {
+            result = DataSourceBindingOperation.USE_NOW;
         }
+        return result;
```
ShadelessFox, line 590: Why don't we use early returns again? This method would definitely benefit from it by at least reducing its nesting significantly

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
```diff
@@ -609,7 +681,14 @@ private void releaseContainer() {
         }
     }
 
-    private DBPDataSourceContainer getDataSourceFromContent() {
+    private record DataSourceBindingInfo(
+        @NotNull String specText,
+        @NotNull DBPDataSourceContainer dataSource
+    ){
+    }
```
ShadelessFox:
```suggestion
    ) {
    }
```

plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLEditor.java
E1izabeth: I prefer a single exit point approach here

E1izabeth: It's a decision tree. There is no "fail fast"

ShadelessFox (dismissed):

uslss:
DB Navigator doesn't work 

image

[log.log](https://github.com/user-attachments/files/30937872/log.log)

E1izabeth:
Are you sure it's connected with my PR?

uslss:
1. Save SQL
2. Choose No,ignore
3. Execute script - Connection created


https://github.com/user-attachments/assets/4c0d6710-1369-43fa-bc3c-1007bb4b8fb2

PR #41727, opened 2026-08-10
dbeaver/pro#9812 Auto open Pallete in Edit mode

E1izabeth:
Closes dbeaver/pro#9812

Unfortunately, the method to open the Palette is private. And there is no way to call it other than to use reflection or copy and paste the whole implementation of the corresponding class to make ONE method public


plugins/org.jkiss.dbeaver.ui.editors.erd/src/org/jkiss/dbeaver/ui/editors/erd/editor/ERDEditorPart.java
```diff
@@ -693,9 +696,50 @@ protected FlyoutPaletteComposite createPaletteComposite(Composite parent) {
...
         paletteComposite.setBackground(ERDThemeSettings.instance.diagramBackground);
+
+
+        if (Arrays.stream(paletteComposite.getChildren())
+                .filter(c -> c.getClass().getName().endsWith(".FlyoutPaletteComposite$Sash"))
+                .findFirst()
+                .orElse(null) instanceof Composite paletteFlyoutSash &&
+            Arrays.stream(paletteFlyoutSash.getChildren())
+                .filter(c -> c.getClass().getName().endsWith(".FlyoutPaletteComposite$ButtonCanvas"))
+                .findFirst().orElse(null) instanceof Widget paletteFlyoutSashButton
```
ShadelessFox: This is pretty much unreadable

ShadelessFox (changes requested):
I'm fine with using reflection (via BeanUtils). It's easier to understand and fil if it breaks than understand what this (vaguely hacky) workaround does

ShadelessFox (approved):

SolovevAnton (approved):

PR #41728, opened 2026-08-10
#41719 Remove project selector from status bar in non-standalone edition

### LINKED ISSUE #41719: Regression in 26.1.4: DBeaver adds a project selector to the global Eclipse status bar
### Description

DBeaver 26.1.4 adds an active-project selector to the Eclipse status bar; this is new in this version and shouldn't be there. The status bar is global across all Eclipse perspectives, so this control is visible and clickable even when working in perspectives that have nothing to do with DBeaver, making it easy to change the active DBeaver project by accident.

### DBeaver Version

CE 26.1.4.202608021747 (Eclipse plugin)

### Operating System

Debian GNU/Linux 13 (trixie)

### Database and driver

_No response_

### Steps to reproduce

1. Install the DBeaver plugin (26.1.4) in Eclipse.
2. Note the DBeaver project selector present in the Eclipse status bar.
3. Switch to any non-DBeaver perspective: the selector is still there.

### Additional context

This selector was not present in the status bar in 26.1.3. Regression introduced in 26.1.4.

E1izabeth:
This PR was generated after reviewing what an AI (OpenCode) has proposed.


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/controls/BreadcrumbTrim.java
```diff
@@ -52,9 +52,14 @@ public class BreadcrumbTrim {
 
     @PostConstruct
     public void createControls(@NotNull Composite parent) {
-        Composite composite = UIUtils.createPlaceholder(parent, 2, 0);
-        ((GridLayout)composite.getLayout()).marginLeft = 5;
-        ProjectsPanel projectsPanel = new ProjectsPanel(composite);
+        Composite composite;
+        if (DBWorkbench.getPlatform().getApplication().isStandalone()) {
+            composite = UIUtils.createPlaceholder(parent, 2, 0);
```
ShadelessFox: You can create a new placeholder with a single column unconditionally and conditionally increase its number of columns through `getLayout`, as already done with `marginLeft`.

ShadelessFox (approved):

IvanGorshechnikov (approved):

mariiachugunova:
The active project selector is now completely missing, including in the DBeaver perspective.
image

E1izabeth:
I think we should not show our DBeaver breadcrumb at the status bar for DBeaver plugin in Eclipse. Breadcrumb should remain in the editor's header as it used to be before we moved it to status bar.

PR #41733, opened 2026-08-10
#41730 Add Frostlake driver support

### LINKED ISSUE #41730: Add Frostlake driver support

Teams that write Snowflake SQL have no way to work against it locally in
DBeaver. Snowflake itself is cloud-only and metered, so local development and
CI runs either cost warehouse credits or happen with no database tooling at all.

Frostlake (https://frostlake.dev) is an open-source Java engine that implements
Snowflake's SQL — DDL/DML, CTEs, window functions, QUALIFY, PIVOT, semi-structured
VARIANT/OBJECT/ARRAY, streams, tasks, stages — and ships a JDBC driver. It can be
connected to today via a generic JDBC driver entry, but that loses everything
DBeaver is good at: the object tree shows only what plain JDBC metadata covers,
and the SQL editor has no dialect, so Snowflake-specific syntax and functions get
no highlighting or completion.

A Frostlake driver plugin, in the same shape as the existing lightweight driver
plugins (org.jkiss.dbeaver.ext.h2 as the model — generic model, no .ui bundle):

- Driver entries for Frostlake's three connection modes: a running server
  (jdbc:frostlake://host:port/db), an embedded persistent directory
  (jdbc:frostlake:file:<dir>), and in-memory (jdbc:frostlake:direct:<name>).
  The driver JAR resolves from Maven Central (dev.frostlake:frostlake-db).
- Navigator folders for the Snowflake-shaped objects generic JDBC metadata
  cannot expose: stages, streams, tasks, pipes, sequences, tags, and the
  account-level warehouses, roles and users.
- A SQL dialect carrying Frostlake's keyword, function and data-type
  vocabularies for highlighting and completion.

I have this implemented and working against DBeaver CE, and I'd like to
contribute it. The change is self-contained: one new plugin bundle, plus two
registration lines (plugins/pom.xml and org.jkiss.dbeaver.db.feature). No
existing behaviour is touched.

I'm happy to open the PR once you confirm you'd accept a driver for this
database — following the contribution guide, I wanted to check here first.

- A generic JDBC connection: works for running queries, but gives no
  Snowflake-shaped object tree and no dialect.
- Reusing the Snowflake driver entry: not viable — different JDBC URL scheme
  and driver class, and Frostlake is a separate engine, not a Snowflake endpoint.
- Keeping the plugin out-of-tree for users to install manually: possible, but
  DBeaver has no plugin marketplace for third-party drivers, so in practice it
  would not reach users.

**Additional context**

Frostlake is my own open-source project (Apache 2.0). The driver artifact is
published on Maven Central, so DBeaver's driver resolver and CI can fetch it
without any bundled binaries.

mlorek:
Closes #41730
Frostlake is a Java SQL engine that emulates core Snowflake features for local and CI testing. This adds a driver plugin so it can be browsed and queried like any other database.

The plugin follows the org.jkiss.dbeaver.ext.h2 pattern - generic model, no .ui bundle, so it inherits the standard connection page - and adds:

* Three driver entries: server (jdbc:frostlake://host:port/db), embedded file (jdbc:frostlake:file:<dir>) and in-memory (jdbc:frostlake:direct:<name>). The driver JAR resolves from Maven Central as dev.frostlake:frostlake-db.

* Navigator folders for the Snowflake-shaped objects the generic model does not cover - stages, streams, tasks, pipes, sequences and tags, plus the account-level warehouses, roles and users - each backed by the engine's own SHOW command and rendered by a single node class.

* A SQL dialect whose keyword, function and data-type vocabularies are generated from the engine (grammar tokens and SHOW BUILTIN FUNCTIONS) rather than hand-maintained, so they stay complete as the engine grows.

Registered in plugins/pom.xml and org.jkiss.dbeaver.db.feature.


E1izabeth:
Thanks for the contribution! We will review your PR snd let you know if any changes are needed

plugins/org.jkiss.dbeaver.ext.frostlake/src/org/jkiss/dbeaver/ext/frostlake/model/FrostlakeDataSource.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.EnumMap;
+import java.util.Map;
+
+/**
+ * Frostlake data source. Everything the generic model handles stays with the generic model; the
+ * Frostlake-specific parts are the schema class, which adds the schema-scoped SHOW-backed kinds, and
+ * the account-level kinds below — warehouses, roles and users belong to no schema, so they hang off
+ * the data source and appear as siblings of Databases in the tree.
+ */
+public class FrostlakeDataSource extends GenericDataSource {
```
E1izabeth, line 38: Have you considered inheriting from Snowflake?

plugins/pom.xml
```diff
@@ -74,6 +74,7 @@
         <module>org.jkiss.dbeaver.ext.postgresql</module>
 
         <module>org.jkiss.dbeaver.ext.snowflake</module>
+    <module>org.jkiss.dbeaver.ext.frostlake</module>
```
E1izabeth:
```suggestion
        <module>org.jkiss.dbeaver.ext.frostlake</module>
```

plugins/org.jkiss.dbeaver.ext.frostlake/src/org/jkiss/dbeaver/ext/frostlake/model/FrostlakeDataSource.java
mlorek: Yes, but the Snowflake plugin models a superset Frostlake can't back (warehouse, role, auth model, jdbc:snowflake:// URL building), so it extends the same GenericDataSource base instead and stays independent of that bundle.

E1izabeth (commented):
Could you provide icons

E1izabeth:
Could you provide icons in svg format? In this case, @2x icons are not needed

Also, if possible, could you adapt them a little bit more to the existing ones for other databases? For example, use the blue (#3387C5) color?

E1izabeth (dismissed):

IvanGorshechnikov:
Can you please use our icons for database objects and not add new ones?
You can look at any plugin.xml for other drivers

E1izabeth (dismissed):

PR #41758, opened 2026-08-12
#41687 Fix metadata panel column sort being overridden by ordinal pos…

### LINKED ISSUE #41687: Metadata panel sorting does not work in DBeaver 26.1.4
### Description

Please see the steps to reproduce.  The feature to click is there but the columns never re-order.

### DBeaver Version

DBeaver Community Edition 26.1.4.202608021747

### Operating System

_No response_

### Database and driver

_No response_

### Steps to reproduce

Run a query and open the result set.
Open the Metadata panel.
Click the Name column header.
Observe the sort arrow appears.
Click again to reverse the sort direction.
Observe the sort arrow changes direction.

Expected Result:

Columns in the Metadata panel are reordered alphabetically (ascending/descending).

Actual Result:

The sort indicator changes, but the Metadata list remains in the original table column order. No visible reordering occurs.

### Additional context

_No response_

gsastraw:
Closes #41687 

The default comparator's previous behavior states that if both items have an ordinal position, i.e. index exposed via DBDAttributeBinding::getOrdinalPosition(), sort by index and return immediately. 

So comparing e.g. column 1 vs. column 2 always returned 1 - 2 = -1 and stopped there, so the name comparison never actually ran. What is done now instead is that we check to see if no column header is selected first via ViewerColumnController::getColumnInfo, so that sorting only runs when no column header is selected.

Reproduced issue with mock data of the following SQLite schema:

```
CREATE TABLE repro (
    zeta_id     INTEGER PRIMARY KEY,
    alpha_name  TEXT,
    bravo_value INTEGER,
    kappa_flag  INTEGER,
    gamma_desc  TEXT
);

SELECT * FROM repro;
```

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/controls/ViewerColumnController.java
+++ b/plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/controls/ViewerColumnController.java
@@ -895,7 +895,7 @@ public class ViewerColumnController<COLUMN, ELEMENT> {
                 return cat1 - cat2;
             }
 
-            if (e1 instanceof DBPObjectWithOrdinalPosition p1 && e2 instanceof DBPObjectWithOrdinalPosition p2) {
+            if (getColumnInfo(viewer) == null && e1 instanceof DBPObjectWithOrdinalPosition p1 && e2 instanceof DBPObjectWithOrdinalPosition p2) {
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

E1izabeth:
Thanks for the contribution!

IvanGorshechnikov (approved):

IvanGorshechnikov:

gsastraw:
Thanks everyone for the quick feedback! <3

PR #41761, opened 2026-08-13
dbeaver/pro#10237 Fix wizard UI

serge-rider:
Created with OpenCode


Matvey16:
Verified

DBeaver CE WIndows
DBeaver EE Windows
DBeaver CE MacOS

Matvey16:
Verified

PR #41767, opened 2026-08-13
dbeaver/pro#10082 Add Copilot review instructions

IvanGorshechnikov:
## Summary
- add security-focused instructions for Copilot code review
- add a review skill to find copy-paste defects and exposed sensitive data

## Why
Helps Copilot identify actionable security and copy-paste risks during pull request reviews.

Closes dbeaver/pro#10082

### DIFF
--- a/.github/copilot-code-review-instructions.md
+++ b/.github/copilot-code-review-instructions.md
@@ -1,0 +1,77 @@
+# Code Review
+
+## Security Review
+
+Review every pull request for security issues in the changed code and its direct
+call paths. Report only actionable findings introduced by the pull request,
+including the affected file and line, attack scenario, and remediation.
+
+Pay particular attention to:
+
+- SQL, shell, file-path, URL, XML, and deserialization inputs that may be
+  controlled by users, database contents, configuration, or remote services.
+- Authentication, authorization, permission checks, and unsafe default access.
+- Secret, token, password, private key, connection string, or personal-data
+  exposure in code, tests, configuration, logs, exceptions, or documentation.
+- Unsafe handling of credentials and database connection settings.
+- Missing validation, encoding, escaping, canonicalization, or resource limits.
+- Insecure temporary files, archive extraction, redirects, TLS configuration,
+  and external-process execution.
+
+Do not report speculative concerns without a realistic attack path. Do not
+request broad refactoring when a focused fix is sufficient.
+
+## Interface Text
+
+Review changed user-visible interface text for punctuation and tone. Apply
+these rules to titles, headings, labels, controls, commands, messages,
+descriptions, and helper text. Do not apply them to technical documentation or
+release notes unless the text is presented in the interface.
+
+### Ellipses
+
+Use the single ellipsis character `…`, not three periods, at the end of a
+command label only when the user must enter information, make a choice, or
+confirm an action before it takes effect. Do not use an ellipsis when opening a
+window or page completes the command, even if the user can make optional
+changes there.
+
+For example, use `Export Data…` and `New Database Connection…`. A command can
+have different labels depending on its flow: use `Delete…` when the user must
+confirm the deletion, and use `Delete` when the deletion takes effect
+immediately. Do not use an ellipsis for `Preferences`, `Properties`, or `View
+Details` because opening those pages completes the command.
+
+### Periods
+
+Do not use periods in titles, headings, labels, controls, commands, or other
+text that names an interface element or action. In messages, descriptions, and
+other explanatory prose, use normal sentence punctuation: end complete
+sentences with a period, including standalone messages, and do not add a
+period to a fragment.
+
+### Colons
+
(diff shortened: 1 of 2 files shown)


.github/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,27 @@
...
+- Duplicated constants, identifiers, URLs, or messages that should differ in
+  the new context.
+- Pasted secrets, credentials, API keys, tokens, private keys, connection
+  strings, personal data, internal URLs, or other sensitive values in source,
+  tests, fixtures, configuration, logs, exceptions, and documentation.
+
+For each finding, identify the affected file and line, explain why the pasted
+content is incorrect or sensitive in this context, and recommend a focused
+fix. Do not report intentional duplication that is already correct and
+maintained independently.
```
E1izabeth, line 32: It won't cover cases when someone tries to add support for a new database by copy pasting the existing plugin instead of extending generic or base db classes

E1izabeth (dismissed):

.github/copilot-code-review-instructions.md
```diff
@@ -0,0 +1,74 @@
...
+### Ellipses
+
+Use the single ellipsis character `…`, not three periods, at the end of a
+command label only when the user must enter information, make a choice, or
+confirm an action before it takes effect. Do not use an ellipsis when opening a
+window or page completes the command, even if the user can make optional
+changes there.
+
+For example, use `Export Data…`, `New Database Connection…`, and `Delete…`.
+Do not use an ellipsis for `Preferences`, `Properties`, or `View Details`.
```
arturtrifonov: With these examples, it may be important to add details about each action.
Both "Delete" and "Delete…" are possible, so it's important to mention that in this example, the user must confirm the deletion

arturtrifonov (dismissed):

.github/copilot-code-review-instructions.md
IvanGorshechnikov: Fixed

.github/copilot-code-review-instructions.md
```diff
@@ -0,0 +1,77 @@
...
+## Interface Text
+
+Review changed user-visible interface text for punctuation and tone. Apply
+these rules to titles, headings, labels, controls, commands, messages,
+descriptions, and helper text. Do not apply them to technical documentation or
+release notes unless the text is presented in the interface.
+
+### Ellipses
+
+Use the single ellipsis character `…`, not three periods, at the end of a
```
ShadelessFox, line 33: This is incorrect. We don't use `…`, we use `...`.

.github/copilot-code-review-instructions.md
```diff
@@ -0,0 +1,77 @@
...
+Use a colon after a label that introduces an input field or a group of radio
+buttons or checkboxes. Do not use a colon when the label and the text inside
+the field form a single phrase.
+
+### Contractions
+
+Do not use contractions in interface text. Write words in full, such as `The
+connection cannot be established` rather than `The connection can't be
+established`. Contractions are acceptable in longer reading-oriented content,
+such as release notes.
```
ShadelessFox, line 64: Are we sure about this?

.github/copilot-code-review-instructions.md
arturtrifonov: Yes, but it's better to change this approach.
All systems use a single symbol, not 3 dots; there's no point in doing the same thing but slightly differently.

ShadelessFox: That would mean that we also need to address this change in other existing places throughout the UI.

.github/copilot-code-review-instructions.md
arturtrifonov: I didn't find any usage of contractions in the interface. It's easier to keep it this way.

ShadelessFox (dismissed):

.github/copilot-code-review-instructions.md
arturtrifonov: It shouldn't look very awful if they exist together in a non-monospace font: "Delete ..." and "Delete…".
But we can remove it for now and discuss what to do with it. There's no rush here.

My point is that there are only two options: use it or don't use it.
And if yes, we should do it the right way, I think.

E1izabeth (dismissed):

.github/skills/code-review/SKILL.md
```diff
@@ -0,0 +1,32 @@
+---
+name: code-review
+description: Review pull requests for copy-paste defects and accidentally committed sensitive data. Use this during code review when changed code may have been copied or pasted.
+---
+
+# Copy-Paste Review
```
serge-rider: Oh

serge-rider:
```suggestion
# Pull Requests Review
```

serge-rider (changes requested):

E1izabeth (approved):

PR #41774, opened 2026-08-14
Dbeaver/pro#6235 grok

Destrolaric:
Closes Dbeaver/pro#6235


plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/openai/OpenAiUtils.java
```diff
@@ -80,8 +80,8 @@ public static OAIResponsesRequest createOpenAiRequest(
                 tool.type = OAITool.TYPE_FUNCTION;
                 tool.name = fd.getFullId();
                 tool.description = fd.getAiDescription();
+                tool.parameters = new OAIToolParameters();
```
serge-rider, line 83: Why? It was explicitly set to null to reduce traffic.

IvanGorshechnikov: @Destrolaric

Destrolaric: Discussed, agreed that this one was necessary

PR #41775, opened 2026-08-14
dbeaver/pro#10163 Update BigQuery driver

IvanGorshechnikov:
Updates the recommended Google BigQuery JDBC driver version to 1.2.0 in CE
Added BigQuery partner attribution

Closes dbeaver/pro#10163
Closes dbeaver/pro#10164


E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.ext.bigquery/src/org/jkiss/dbeaver/ext/bigquery/BQDataSourceProvider.java
```diff
@@ -44,6 +46,17 @@ public String getConnectionURL(@NotNull DBPDriver driver, @NotNull DBPConnection
         if (!CommonUtils.isEmpty(connectionInfo.getHostPort())) {
             url.append(":").append(connectionInfo.getHostPort());
         }
+        if (CommonUtils.toBoolean(driver.getDriverParameter(DRIVER_PARAM_PARTNER_ATTRIBUTION), false)) {
+            String product = GeneralUtils.getProductName()
+                .trim()
+                .replace(" ", "+")
+                .replace("/", "-");
```
ShadelessFox, line 53: Isn't this what `java.net.URLEncoder` does (but better)?

plugins/org.jkiss.dbeaver.ext.bigquery/src/org/jkiss/dbeaver/ext/bigquery/BQDataSourceProvider.java
```diff
@@ -22,9 +22,11 @@
 import org.jkiss.dbeaver.ext.generic.GenericDataSourceProvider;
 import org.jkiss.dbeaver.model.connection.DBPConnectionConfiguration;
 import org.jkiss.dbeaver.model.connection.DBPDriver;
+import org.jkiss.dbeaver.utils.GeneralUtils;
 import org.jkiss.utils.CommonUtils;
 
 public class BQDataSourceProvider extends GenericDataSourceProvider<BQDataSource> {
+    private static final String DRIVER_PARAM_PARTNER_ATTRIBUTION = "supports-partner-attribution";
```
ShadelessFox, line 29: Both drivers have it marked. Do we have any other drivers that use the same `BQDataSourceProvider` but don't support attribution?

IvanGorshechnikov: Yes, we have the Simba driver for BigQuery

plugins/org.jkiss.dbeaver.ext.bigquery/src/org/jkiss/dbeaver/ext/bigquery/BQDataSourceProvider.java
IvanGorshechnikov: No, in that case, we will encode other symbols, which we don't need to encode like `/, (, ), :, ;`

ShadelessFox (approved):

SolovevAnton (approved):

uslss:
verified (db ue, ce)

PR #41781, opened 2026-08-15
dbeaver/pro#10290 Radio buttons with hints

serge-rider:


plugins/org.jkiss.dbeaver.ui.forms/src/org/jkiss/dbeaver/ui/forms/UIControlBuilderImpl.java
```diff
@@ -357,12 +357,14 @@ int toSWT() {
         }
 
         private final UIObservable<String> text;
+        private final UIObservable<String> tooltip;
         private final Consumer<SelectionEvent> onSelect;
         private final Kind kind;
         private UIObservable<Boolean> selected;
 
-        ButtonBuilderImpl(@Nullable UIObservable<String> text, @Nullable Consumer<SelectionEvent> onSelect, @NotNull Kind kind) {
+        ButtonBuilderImpl(@Nullable UIObservable<String> text, @Nullable UIObservable<String> tooltip, @Nullable Consumer<SelectionEvent> onSelect, @NotNull Kind kind) {
```
ShadelessFox, line 365: Don't, there's already `UIControlBuilder#tooltip`. Constructor parameters define properties that can't be changed and/or are required by default.

ShadelessFox: Use the following at the callsite instead
```java
rb.radioButton(
    "text",
    bb -> bb
        .selected(<selected observable>)
        .tooltip("tooltip"))
````

ShadelessFox (changes requested):

PR #41790, opened 2026-08-17
dbeaver/pro#10293 Fix toolbars background

serge-rider:


Matvey16:
Verified

DBeaver CE WIndows

PR #41793, opened 2026-08-17
dbeaver/pro#10295 Fix freeze when restoring splash on Linux

ShadelessFox:


Matvey16:
Verified

DBeaver CE WIndows
DBeaver CE Linux .deb
Dbeaver CE Linux Snap

PR #41797, opened 2026-08-17
dbeaver/pro#10272 Support AI account authentication

IvanGorshechnikov:
## Summary
- add ChatGPT account authentication for OpenAI through browser PKCE and device-code OAuth flows
- securely persist and refresh account credentials, route account requests to the Codex endpoint, and load the account model catalog
- add provider-neutral account authentication hooks used by the Grok/xAI integration
- add account sign-in/sign-out UI while preserving API-token profiles and preventing model refresh before sign-in
- adapt Responses API payloads and temperature handling for ChatGPT account models

## Testing
- 12 `OpenAIAccountAuthenticatorTest` tests passed
- changed model and CE UI classes compiled with Java 21
- `git diff --check` passed
- full Tycho compile is blocked in this checkout by the missing `org.jkiss.dbeaver.model.jdbc` target-platform bundle

## Related PR
- dbeaver/dbeaver-ee#4741

Closes dbeaver/pro#10272


serge-rider:
Too much code unfortunately.
Generally we should think about reuse of oauth flow components here. I also don't like how it parses json. Looks bit fragile.

But let's test it.

PR #41802, opened 2026-08-18
dbeaver/pro#10268 Fix refresh with references panel open

E1izabeth:
Steps to reproduce:
1. Open Oracle HR.EMPLOYEES table
2. Click on the first column, second row cell
3. Scroll down to load all 106 rows
4. Press Refresh


IvanGorshechnikov (approved):

SolovevAnton (approved):

Matvey16:
There are still cases where a table with references can hang, now it involves Value panel

1. Open a table with references
2. Click on columns fith a foreign key and open Value panel
3. Scroll to the bottom and refresh a few times

https://github.com/user-attachments/assets/bb6a9dcf-f83c-4868-b2a7-ddd5e0a4aac6

Another case doesn't involve foreign keys at all. Reproducible only in SQL Editor

1. Run a select of a big table
2. Reduce fetch size to about 20% of all rows
3. Keep quicly scrolling to the bottom

https://github.com/user-attachments/assets/e74b8ff7-4dae-4b57-a4b0-db6cf535b0ae

Matvey16:
Verified

DBeaver CE Windows

PR #41803, opened 2026-08-18
dbeaver/pro#10120 Fix database logos on macOS

IvanGorshechnikov:
## Summary
- Recalculate the title area layout after replacing the default JFace banner
- Prevent wide database logos from being clipped on macOS

## Testing
- `git diff --check`
- Standalone UI plugin build could not resolve the local `org.jkiss.dbeaver.model` bundle

Closes dbeaver/pro#10120

This PR was generated with AI assistance (OpenCode).


Matvey16:
Verified

DBeaver EE MacOS
DBeaver CE WIndows

PR #41807, opened 2026-08-18
dbeaver/pro#9789 Fix stale table header trim on Linux

ShadelessFox:


plugins/org.jkiss.dbeaver.ui.ai/src/org/jkiss/dbeaver/ui/ai/preferences/AIPreferencePageEngines.java
```diff
@@ -279,6 +279,12 @@ protected Control createPreferenceContent(@NotNull Composite parent) {
         });
         relayoutPage();
         UIUtils.packColumns(profilesViewer.getTable(), true);
+        UIUtils.asyncExec(() -> {
+            // SWT on Linux can keep a stale header trim unless the table is realized with header state toggled.
+            Table table = profilesViewer.getTable();
+            table.setHeaderVisible(true);
+            table.setHeaderVisible(false);
+        });
```
E1izabeth: Can we do it ONLY for Linux then?

ShadelessFox: We can. I thought it looked ugly, so I opted out.

E1izabeth (approved):

PR #41808, opened 2026-08-18
#41593 Fix enum editing in record mode

### LINKED ISSUE #41593: Data Editor: ENUM dropdown does not open in Record View after pressing "TAB" key or in the Value Panel
### Description

**Bug Description**
When pressing the Tab key in the Data Editor to switch from Grid View to Record View, fields of type ENUM do not display the dropdown list of available values.
The exact same issue occurs in the Value side panel on the right.

### DBeaver Version

Environment - DBeaver version: 26.1.3.202607191738

### Operating System

Operating System: Windows 10

### Database and driver

Database: MariaDB 11.8.5

### Steps to reproduce

**Steps to Reproduce**
1. Open a table that contains at least one column of type ENUM.
2. View the data in the Data Editor.
3. Press "TAB" key to switch to Record View (where fields are displayed vertically, one below another).
4. Select an ENUM field.
5. Press Enter or double-click to edit it and attempt to open the list of allowed values.
    * Actual result: The dropdown list of allowed values does not open/expand.
6. With the same ENUM field selected, check the Value side panel on the right.
    * Actual result: The Value panel also fails to display or expand the dropdown list of allowed values.

**Impact**
When using Record View ("TAB" key) or the Value panel, it is impossible to select or edit any value for ENUM fields.

### Additional context

**Expected Behavior**
When editing an ENUM field in Record View or from the Value panel, a dropdown list showing all allowed values should be displayed, matching the behavior seen when editing directly inside the Grid View cell.

ShadelessFox:
Regression introduced in https://github.com/dbeaver/pro/issues/8775, please check it as well


E1izabeth (approved):

IvanGorshechnikov (approved):

Matvey16:
Verified

DBeaver CE Windows

PR #41828, opened 2026-08-19
#41791 Individually restore widths per column

### LINKED ISSUE #41791: Data Editor resets manually adjusted column widths after hiding a column
### Description

When working in the Data Editor, manually adjusted column widths are lost after hiding a column.

Actual result:
The widths of the remaining visible columns are recalculated or reset.

Expected result:
Hiding one column should not change manually adjusted widths of the other visible columns.

### DBeaver Version

DBeaver 26.1.5

### Operating System

Microsoft Windows 11 Enterprise

### Database and driver

Database: SQlite 3.42.0
Driver: SQLite JDBC driver

### Steps to reproduce

1. Open any table in the Data Editor.
2. Manually resize several columns.
3. Right-click one of the column headers.
4. Select "Hide column".
5. Observe the widths of the remaining visible columns.

Actual result:
The manually adjusted widths of other columns are reset or recalculated.

Expected result:
The manually adjusted widths of all remaining columns should stay unchanged.

### Additional context

Column visibility is also not preserved after closing and reopening the table.

This behavior makes editing wide tables difficult because the user must repeatedly resize and hide columns.

gsastraw:
Fixes #41791 

When a user chooses to hide a column, a check is done to see if the widths from before are equal to the # of columns currently showing. However, when the user hides columns, this check does not trigger, causing column sizes to be recomputed. This is because the columns that are currently showing are not equal to the widths (columns) from before.

This check seems to be a regression from a216ac223800480abf0b4ec0aae37a1486ce031e where widths were originally stored in an int[], where afterwards a refactor was done to an element-keyed map.


E1izabeth (approved):

gsastraw:
@IvanGorshechnikov just wanted to check how this is going

ShadelessFox (approved):

ShadelessFox:
Thanks for the contribution, passing to QA

gsastraw:
> Thanks for the contribution, passing to QA

Thank you @ShadelessFox ! I notice that the build failed due to a test in BOMParser, doesn't seem like it's directly related to my changes. Would you like me to fix this test too? Thinking that it's a fix unrelated to my changes that it shouldn't belong in this branch.

E1izabeth:
I believe it's unrelated

PR #41839, opened 2026-08-20
#41838: GBase 8s: Fix "Copy As SQL" cannot handle JSON type value properly

### LINKED ISSUE #41838: GBase 8s: "Copy As SQL" command can not handle JSON type value properly.
### Description

When I use "Copy as SQL" command in GBase 8s, I found that when the table contains `JSON` type column, the result SQL contains syntax error, like this:

```sql
INSERT INTO testdb.json_copy_issue (json_content) VALUES ({"key":"value"});
```

Execute SQL above will produce syntax error, the JSON value should be handled properly like:

```sql
INSERT INTO testdb.json_copy_issue (json_content) VALUES ('{"key":"value"}');
```

After debugging, I may figured out where need to fix. I will fix it and open a PR for that :)

### DBeaver Version

Community Edition 26.1.5

### Operating System

Windows 11 25H2 Build 26200.9168

### Database and driver

1. Database name and version: GBase 8s
2. Driver: `com.gbasedbt:jdbc:3.5.1`

### Steps to reproduce

1. Initialize a GBase 8s table using SQL below: 

```sql
CREATE TABLE json_copy_issue (
	id SERIAL8 NOT NULL,
	json_content JSON(32739),
	CONSTRAINT json_copy_issue_pk PRIMARY KEY (id)
);
COMMENT ON TABLE testdb.json_copy_issue IS 'Demonstrate DBeaver ''Copy as SQL'' issue.';

INSERT INTO json_copy_issue (json_content) VALUES ('{"key":"value"}');
```

2. Right click the only row in data-editor and choose "Advanced Copy" -> "Copy As SQL"

Image

3. pasted it in SQL console.

Image

 

### Additional context

_No response_

Downstream1998:
Close #41838 

Fix this issue by extending `GenericSQLDialect` and registering `GBase8sDialect` to handle this case. Now the copied SQL will look like this:

```sql
INSERT INTO testdb.json_copy_issue (json_content) VALUES ('{"key":"value"}');
```


plugins/org.jkiss.dbeaver.ext.gbase8s/plugin.xml
```diff
@@ -75,4 +75,9 @@
             objectType="org.jkiss.dbeaver.ext.gbase8s.model.GBase8sUniqueKey" />
     </extension>
 
+    <extension point="org.jkiss.dbeaver.sqlDialect">
```
serge-rider, line 78: You also need to change dialect="generic" in driver definition

serge-rider (changes requested):

serge-rider:
@codex review

plugins/org.jkiss.dbeaver.ext.gbase8s/plugin.xml
Downstream1998: fixed it in latest commit

PR #41850, opened 2026-08-20
#39606 Fix empty request body against plaintext OpenAI-compatible end…

### LINKED ISSUE #39606: No response when using the OpenAI api provided by LM Studio
### Description

After I entered the URL provided by lmstudio in DBeaver's AI settings and clicked 'Test Connection', it just keeps sending request. But actually, if I try navigate to 'http://127.0.0.1:1234/v1/models' directly in Edge, Edge can display the model list results. I tried changing the port number and turning off the firewall, but it still doesn't work.

Image

Image

### DBeaver Version

25.2.4

### Operating System

win11

### Database and driver

_No response_

### Steps to reproduce

_No response_

### Additional context

_No response_

kezarjg:
### Description

AI requests sent to a self-hosted OpenAI-compatible endpoint over plaintext `http://` (vLLM, Ollama, LM Studio, ...) arrive at the server with an empty body. Depending on the server, the request is rejected or the UI hangs - "Test Connection" and model listing never return. HTTPS endpoints are unaffected.

vLLM/FastAPI answers `POST /v1/responses` with:

```
1 validation error:
{'type': 'missing', 'loc': ('body',), 'msg': 'Field required', 'input': None}
```

### Root cause

`AbstractHttpAIClient` built its client with `HttpClient.newHttpClient()`, which defaults to `HTTP_2`. Without TLS there is no ALPN, so the JDK attempts an h2c upgrade and adds `Connection: Upgrade, HTTP2-Settings` and `Upgrade: h2c` to the request. Servers such as uvicorn with the httptools parser treat that as an unsupported upgrade and stop parsing the request, handing the application an empty body. HTTPS endpoints negotiate via ALPN and never send those headers, which is why only plaintext self-hosted endpoints are affected.

### Fix

Pin the shared AI HTTP client to HTTP/1.1. None of the AI APIs called here benefit from HTTP/2 over an unencrypted local connection, and this covers every engine built on `AbstractHttpAIClient`.

`OpenAIRequestFilter` additionally rebuilt each request from scratch and silently dropped the caller's timeout and version, so the configured request timeout never reached OpenAI requests. Both are now carried over.

### Testing

Built DBeaver CE from this branch (win32/x86_64 product) and verified manually against a plaintext `http://` vLLM endpoint: before the change the server received `POST /v1/responses` with no body and answered `Field required`; after it, the body arrives and the request completes. Other OpenAI-compatible servers and HTTPS endpoints were not re-tested.

Adds `OpenAIRequestFilterTest` to `test/org.jkiss.dbeaver.model.ai.test`, covering the timeout and version that the filter used to drop, plus the authorization and content-type headers.

### Related

Very likely the same root cause as #39606 (LM Studio / vLLM over plaintext, closed as stale). That thread diagnosed HTTP/2 as the trigger and concluded it had to be worked around with a reverse proxy - this fixes it on the DBeaver side.

Closes dbeaver/dbeaver#39606

---

*This PR was generated with AI (Claude Code). The change has been reviewed and tested by a human contributor.*

### DIFF
--- a/plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AbstractHttpAIClient.java
+++ b/plugins/org.jkiss.dbeaver.model.ai/src/org/jkiss/dbeaver/model/ai/engine/AbstractHttpAIClient.java
@@ -36,7 +36,9 @@ public abstract class AbstractHttpAIClient implements AutoCloseable {
 
     public AbstractHttpAIClient() {
         this.client = new MonitoredHttpClient(
-            HttpClient.newHttpClient(),
+            // Force HTTP/1.1: over plaintext http:// the default HTTP_2 client attempts an h2c upgrade,
+            // and OpenAI-compatible servers like uvicorn/httptools then drop the request body
+            HttpClient.newBuilder().version(HttpClient.Version.HTTP_1_1).build(),
             this::mapHttpError,
             this::processErrors
         );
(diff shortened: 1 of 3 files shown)


E1izabeth (approved):

IvanGorshechnikov (approved):

IvanGorshechnikov:

uslss:
verified 
image

PR #41856, opened 2026-08-20
dbeaver/pro#10369 fix NPE in LogOutputStream

flambchap:
Resolves dbeaver/pro#10369.

## What Was Wrong

The original writer initialization combined a null check and rotation into one short-circuit expression:

```java
if (currentLogFileOutput == null || rotateCurrentLogFile(false)) {
    currentLogFileOutput = new FileOutputStream(currentLogFile, true);
}
return currentLogFileOutput;
```

The failure sequence was:

1. `currentLogFileOutput` was non-null, so the first condition evaluated to `false`.
2. The log exceeded its size limit, so rotation began.
3. Rotation called `close()`, which closed the stream and set `currentLogFileOutput` to null.
4. Renaming the current log failed, causing `rotateCurrentLogFile()` to return `false`.
5. The complete condition became `false || false`, so a new stream was not opened.
6. `getLogFileWriter()` returned null, and the subsequent `.write(...)` caused the reported NPE.

The same problem could happen if the rename succeeded but listing old log files failed.

Despite appearing race-related, this was not an ordinary thread race. `write()`, `close()`, and writer acquisition are synchronized, preventing another thread from closing the stream between acquisition and writing. The bug was state mutation hidden inside the boolean expression.

Scheduler executions make rename failure more likely because separate DBeaver processes may contend for the same log file. File locking, permissions, antivirus software, or other filesystem conditions can also cause `File.renameTo()` to return `false`.

**How The Fix Works**

The rotation and writer initialization are now separate:

```java
if (currentLogFileOutput != null) {
    rotateCurrentLogFile(false);
}
if (currentLogFileOutput == null) {
    currentLogFileOutput = new FileOutputStream(currentLogFile, true);
}
return currentLogFileOutput;
```

The second check observes the writer’s state after rotation:

- If no rotation is needed, the existing writer remains open and is reused.
- If rotation succeeds, the closed writer is replaced with one for the new active log.
- If renaming fails, the original log is reopened in append mode.
- If reopening fails, an appropriate `IOException` is propagated instead of an unrelated NPE.

The regression test forces `renameTo()` to fail, exceeds the log-size threshold, and verifies that the next write succeeds and all bytes remain in the active log.

### DIFF
--- a/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/LogOutputStream.java
+++ b/plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/LogOutputStream.java
@@ -17,6 +17,7 @@
 package org.jkiss.dbeaver;
 
 import org.jkiss.code.NotNull;
+import org.jkiss.code.Nullable;
 import org.jkiss.dbeaver.model.preferences.DBPPreferenceStore;
 
 import java.io.File;
@@ -47,6 +48,7 @@ public class LogOutputStream extends OutputStream {
     /**
      * The Writer to log messages to.
      */
+    @Nullable
     private volatile FileOutputStream currentLogFileOutput = null;
     private volatile long currentLogSize;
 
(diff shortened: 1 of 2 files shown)


IvanGorshechnikov (approved):

SolovevAnton (approved):

E1izabeth (changes requested):
> If renaming fails, the original log is reopened in append mode.

It shouldn't be so. The reason for all this code is to restrict files size. I believe the correct way to solve it is to add a suffix with increment.

> If reopening fails, an appropriate IOException is propagated instead of an unrelated NPE.

Logging should not affect the app functioning while there's enough space on disk. I think the correct approach is to retry.

Issue #38321, opened 2025-06-04
[Clickhouse] Cannot import or insert/update timestamp with enabled date formatting


Matvey16:
### Description

Incorrect insert query is generated when date formatting is enabled. DBeaver adds trailing zeroes to the timestamp even if there none in original input. Clickhouse doesn't support that. The same error leads to inability to insert timestamp data in data import. Even if original file have a timestamp without trailing zeroes. Dbeaver will still use them in generated insert statement

### DBeaver Version

25.1.0

### Database and driver

ClickHouse 22.1.3.7
ClickHouse JDBC Driver 0.8.5

### Steps to reproduce

1. Enable data formatting
2. Create a table with DateTime column
3. Type a timestamp in appropriate format into a cell e.g. `2013-05-26 14:47:57`
4. Click generate script

**Expected:** `UPDATE last_update='2013-05-26 14:47:57'`
**Actual:** `UPDATE last_update='2013-05-26 14:47:57.000'`

### Additional context

![Image](https://github.com/user-attachments/assets/7338ef05-0c5b-467c-8b53-ff0c507dbaf2)

![Image](https://github.com/user-attachments/assets/65fc3ba0-b293-4510-aad4-cc50890498af)

Hubbitus:
@Matvey16, does "Use custom data/time format" helps you?

<img width="1209" height="435" alt="Image" src="https://github.com/user-attachments/assets/b7b3c8c5-2294-4149-9106-c0e0c0a28cb2" />

<img width="1229" height="438" alt="Image" src="https://github.com/user-attachments/assets/34bf150c-e242-4539-a0be-dcf0eb0ce3d7" />
?

Matvey16:
@Hubbitus This is another case. On your screenshot it is a script generated from the context menu with `Generate SQL -> Insert` where it does indeed work. I'm talking about the case where script is generated for changes made in data editor as well as insert in data transfer

uslss:
verified

Issue #9855, opened 2020-09-23
Clickhouse: DateTime64  data type support


serge-rider:
DataTime64 type read fails with error
```

java.lang.NumberFormatException: For input string: "2020-08-17 01:43:50.000930756"
at java.base/jdk.internal.math.FloatingDecimal.readJavaFormatString(FloatingDecimal.java:2054)
at java.base/jdk.internal.math.FloatingDecimal.parseDouble(FloatingDecimal.java:110)
at java.base/java.lang.Double.parseDouble(Double.java:549)
at ru.yandex.clickhouse.response.ClickHouseResultSet.getDouble(ClickHouseResultSet.java:469)
at org.jkiss.dbeaver.model.impl.jdbc.exec.JDBCResultSetImpl.getDouble(JDBCResultSetImpl.java:394)
at org.jkiss.dbeaver.model.impl.jdbc.data.handlers.JDBCNumberValueHandler.fetchColumnValue(JDBCNumberValueHandler.java:200)
```

It happens because driver doesn't include this data type in supported type list.

serge-rider:
Fixed

kseniiaguzeeva:
It is impossible to create a table with column type DateTime64 to test it.
Steps:
Execute in SQL Editor:
CREATE TABLE default.hits_v2
(
    `WatchID` UInt64,
    `JavaEnable` UInt8,
    `Title` String,
    `GoodEvent` Int16,
    `EventTime` DateTime64,
    `EventDate` Date,
    `CounterID` UInt32,
    `ClientIP` UInt32,
    `ClientIP6` FixedString(16),
    `RegionID` UInt32,
    `UserID` UInt64,
    `CounterClass` Int8,
    `OS` UInt8,
    `UserAgent` UInt8,
    `URL` String,
    `Referer` String,
    `URLDomain` String,
    `RefererDomain` String,
    `Refresh` UInt8,
    `IsRobot` UInt8,
    `RefererCategories` Array(UInt16),
    `URLCategories` Array(UInt16),
    `URLRegions` Array(UInt32),
    `RefererRegions` Array(UInt32),
    `ResolutionWidth` UInt16,
    `ResolutionHeight` UInt16,
    `ResolutionDepth` UInt8,
    `FlashMajor` UInt8,
    `FlashMinor` UInt8,
    `FlashMinor2` String,
    `NetMajor` UInt8,
    `NetMinor` UInt8,
    `UserAgentMajor` UInt16,
[...]

PR #41865, opened 2026-08-21
dbeaver/pro#9841 Fix date and time values handling in Clickhouse

E1izabeth:
### For QA:
1. Check that https://github.com/dbeaver/dbeaver/issues/38321 and https://github.com/dbeaver/dbeaver/issues/9855 are not reproducible
2. There is also Time data type since Clickhouse 25.6, check that formatting, filtering, updating inline, and with value panel work correctly for all the date-time-related data types:
    - date, date32, datetime, datetime64 - before 25.6
    - date, date32, datetime, datetime64, time, time64 - since 25.6 (we have 26.5)


IvanGorshechnikov (approved):

plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/model/data/ClickHouseDateTimeValueHandler.java
```diff
@@ -43,7 +44,12 @@ public ClickHouseDateTimeValueHandler(DBDFormatSettings formatSettings) {
     @Nullable
     @Override
     protected Format getNativeValueFormat(DBSTypedObject type) {
-        return DEFAULT_DATETIME_FORMAT;
+        return switch (type.getTypeID()) {
+            case Types.TIMESTAMP -> DEFAULT_DATETIME_FORMAT;
+            case Types.TIME -> DEFAULT_TIME_FORMAT;
+            case Types.DATE -> DEFAULT_DATE_FORMAT;
+            default -> null;
```
serge-rider, line 51: What about other types? Maybe we should use impl from JDBCDateTimeValueHandler?

E1izabeth: > Maybe we should use impl from JDBCDateTimeValueHandler?

We are already using it:
`public class ClickHouseDateTimeValueHandler extends JDBCDateTimeValueHandler {`

>What about other types?

There are no "other types"
image

And if they appear, it'll be much better to consider their semantics instead of silently applying default behavior; and even for these, I am not sure we are happy with the default implementation, but this is a subject for testing

ShadelessFox (approved):

uslss:
verified (UE)

PR #41868, opened 2026-08-21
dbeaver/pro#9857 don't show the password prompt if only new connection is created

Destrolaric:
Closes dbeaver/pro#9857


IvanGorshechnikov (approved):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/ui/preferences/PrefPageConnectionTypes.java
```diff
@@ -549,6 +551,7 @@ public boolean performOk() {
                 hasChanges = true;
             } else if (!source.equals(changed)) {
                 // Changed type
+                hasExistingTypeChanges = true;
```
E1izabeth, line 554: I don't understand such naming. We already have a flag `hasChanges`. What does adding to that flag name "ExistingType" should mean?

Destrolaric: Yep, different. New type counts as `hasChanges`. But we do not need to update datasources for it. `hasExistingTypeChanges ` marks if any datasource updates are needed

E1izabeth (approved):

Matvey16:
Verified

DBeaver CE Windows
DBeaver EE Windows

PR #41877, opened 2026-08-22
dbeaver/dbeaver#41855 Fix truncation of large numeric values in the cell editor

### LINKED ISSUE #41855: PostgreSQL numeric values are truncated in the results cell value viewer
### Description

Very large numeric values are truncated at 100 digits in the results cell value viewer, although they are displayed correctly in the results grid.

### DBeaver Version

CE 26.1.5

### Operating System

Windows 11 Pro

### Database and driver

PostgreSQL 17.9
Driver 42.7.13

### Steps to reproduce

This PostgreSQL query:

```
with recursive fact as
(
	select 1 as n, 1::numeric as result
	union all
	select n+1 as n, result*(n+1) as result
	from fact
	where n+1<=100
)
select result
from fact
where n=100
```

...produces the value:

93326215443944152681699238856266700490715968264381621468592963895217599993229915608941463976156518286253697920827223758251185210916864000000000000000000000000

in the result grid (159 digits), but the **results cell value viewer displays just the first 100 digits**:

9332621544394415268169923885626670049071596826438162146859296389521759999322991560894146397615651828

Double clicking on the full 159-digit value in the grid cell will also produce the **truncated value**, so there is apparently also the risk of data corruption when editing and saving.

### Additional context

This is due to `MAX_NUMBER_LENGTH` being hardcoded at 100. That seems to be an arbitrary cutoff. PG's numeric data type can go up to 131k digits before decimal point, though.

ayush-singh-0601:
`NumberInlineEditor` is used for both inline cell editing and the results value viewer. It set `Text.setTextLimit(100)`.

On Windows and macOS, `Text.setText()` truncates to that limit, so a 159-digit PostgreSQL `numeric` (100!) is shown in full in the result grid but only the first 100 digits in the viewer. Double-clicking the cell loads that truncated string, so saving can write a wrong value back.

100 was an arbitrary cap, not a real type limit. PostgreSQL NUMERIC allows 131072 digits before the decimal point and 16383 after. The SWT limit is still there so the widget does not take unbounded input, but it now matches that range. If an existing value is still longer, the editor raises the limit before `setText` so we never clip data that is already on screen.

Linux/GTK does not truncate `setText` for `SWT.MULTI`, which is why this is easy to miss off Windows.

Closes dbeaver/dbeaver#41855

This PR was generated with AI (Grok).


ayush-singh-0601:
Hi maintainers, could someone please take a look when you have a chance? The latest PR Check, CodeQL, and commit-message validation runs are all showing action_required, so it looks like the workflows need maintainer approval rather than a code change on my side. Codacy is passing, maintainer edits are enabled, and I am happy to make any changes or add coverage you would prefer. Thanks!

E1izabeth:
Hi! Thanks for the contribution and waiting! Why have you closed your PR?

ayush-singh-0601:
@E1izabeth I'm working on this and fix the issue given in suggestion

ayush-singh-0601:
Hey @E1izabeth ,   Pushed an update addressing the review feedback. The editor now retains an editing buffer when raising the text limit for large numeric values instead of capping strictly at the existing length, preventing field locking during edits.

PR #41880, opened 2026-08-23
dbeaver/pro#10382 Fix fake drivers dialog

serge-rider:


Matvey16:
Verified

DBeaver CE Windows
Dbeaver EE Windows

PR #41884, opened 2026-08-24
dbeaver/pro#9965 Configuration import fix

ShadelessFox:
Please test both the preferences import/export functionality. To reproduce the reported issue, open a created archive and manually add a directory/file containing `../../../` in its name.

### DIFF
--- a/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/tools/configuration/CopyingFileVisitor.java
+++ b/plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/tools/configuration/CopyingFileVisitor.java
@@ -29,27 +29,37 @@ import java.nio.file.attribute.BasicFileAttributes;
 final class CopyingFileVisitor extends SimpleFileVisitor<Path> {
     private static final Log log = Log.getLog(CopyingFileVisitor.class);
 
-    private final Path source;
-    private final Path target;
+    private final Path sourceRoot;
+    private final Path targetRoot;
 
     /**
      * Constructs a new CopyingFileVisitor with the specified source and target paths.
      *
-     * @param source the source directory from which files will be copied
-     * @param target the target directory to which files will be copied
+     * @param sourceRoot the source root directory from which files will be copied
+     * @param targetRoot the target root directory to which files will be copied
      */
-    CopyingFileVisitor(@NotNull Path source, @NotNull Path target) {
-        this.source = source;
-        this.target = target;
+    CopyingFileVisitor(@NotNull Path sourceRoot, @NotNull Path targetRoot) {
+        this.sourceRoot = sourceRoot;
+        this.targetRoot = targetRoot;
     }
 
     @NotNull
     @Override
     public FileVisitResult visitFile(@NotNull Path file, @NotNull BasicFileAttributes attrs) throws IOException {
-        var result = target.resolve(source.relativize(file).toString());
+        var result = resolveTargetPath(file);
         log.debug("Copying file " + result);
         Files.createDirectories(result.getParent());
         Files.copy(file, result, StandardCopyOption.REPLACE_EXISTING);
         return FileVisitResult.CONTINUE;
+    }
+
+    @NotNull
+    private Path resolveTargetPath(@NotNull Path sourcePath) {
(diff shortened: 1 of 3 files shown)


plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/tools/configuration/CopyingFileVisitor.java
```diff
@@ -0,0 +1,65 @@
...
+import org.jkiss.dbeaver.Log;
+
+import java.io.IOException;
+import java.nio.file.*;
+import java.nio.file.attribute.BasicFileAttributes;
+
+/**
+ * A file visitor that copies files from a source directory to a target directory, preserving the directory structure.
+ */
+final class CopyingFileVisitor extends SimpleFileVisitor<Path> {
```
SolovevAnton, line 29: not sure about making it final, but ok

SolovevAnton, line 29: Also, not sure, but may be we want also to refactor this:
`org.jkiss.dbeaver.model.navigator.fs.DBNPathBase#walkFileTree`
impl to the same visitor to remove code duplication

SolovevAnton (dismissed):

E1izabeth (dismissed):

plugins/org.jkiss.dbeaver.core/src/org/jkiss/dbeaver/tools/configuration/CopyingFileVisitor.java
ShadelessFox: Any reasons against making it final?

ShadelessFox: I would leave that to a separate ticket. That method handles weird cases that this visitor doesn't. Will likely need to be adapted and extended to the point of being too generic

SolovevAnton: for example, here: org.jkiss.dbeaver.model.navigator.fs.DBNPathBase#walkFileTre i would reuse it, overrriding only resolveTargetPath(@NotNull Path sourcePath) method.

ShadelessFox: I see. It's package-private anyway, so it would not benefit from being non-final at the moment. Let's move it somewhere else and refactor in preparation for https://github.com/dbeaver/dbeaver/pull/41884#discussion_r3843723375

ShadelessFox:
Seems like me and @E1izabeth fixed the same issue in two different tickets/PRs. I think it's worth accepting this one either way, as @SolovevAnton would benefit from the changes proposed here in the future :^)

E1izabeth:
OK
As you wish =)

uslss:
error on attempt to export config
image

```
eclipse.buildId=unknown
java.version=25.0.3
java.vendor=Microsoft
BootLoader constants: OS=win32, ARCH=aarch64, WS=win32, NL=en
Framework arguments:  -eclipse.keyring C:\Users\Dziyana\AppData\Roaming\DBeaverData\secure\secure_storage
Command-line arguments:  -os win32 -ws win32 -arch aarch64 -eclipse.keyring C:\Users\Dziyana\AppData\Roaming\DBeaverData\secure\secure_storage

org.eclipse.core.jobs
Error
Mon Aug 24 17:43:41 CEST 2026
An internal error occurred during: "Copying workspace configuration".

java.nio.file.ProviderNotFoundException: Provider not found
	at java.base/java.nio.file.FileSystems.newFileSystem(Unknown Source)
	at java.base/java.nio.file.FileSystems.newFileSystem(Unknown Source)
	at org.jkiss.dbeaver.tools.configuration.ConfigurationExportWizard$1.run(ConfigurationExportWizard.java:83)

```

ShadelessFox:
Seems like we're packaging JRE without `jdk.zipfs`. Well...

PR #41900, opened 2026-08-25
dbeaver/pro#10403 Handle unsupported AI profiles across editions

IvanGorshechnikov:
## Summary
- skip the complete legacy AI engine configuration when its engine is unavailable in the current edition
- keep the JSON reader at the correct nesting level so supported profiles continue loading
- add a regression test covering an unsupported configuration with a nested array

## Testing
- `mvn verify -pl plugins/org.jkiss.dbeaver.model.ai,test/org.jkiss.dbeaver.model.ai.test -am -Dtest=AISettingsManagerTest`

Closes dbeaver/pro#10403


ShadelessFox (approved):

E1izabeth (approved):

Matvey16:
Verified

Dbeaver EE Windows
DBeaver CE Windows

PR #41907, opened 2026-08-25
dbeaver/pro#10389 fix deletion

Destrolaric:
Closes dbeaver/pro#10389


plugins/org.jkiss.dbeaver.ui.editors.connection/src/org/jkiss/dbeaver/ui/preferences/PrefPageProjectNetworkProfiles.java
```diff
@@ -195,7 +195,9 @@ protected void removeProfile(
     ) throws DBException {
         DBWNetworkProfileManager profilesRegistry = getProfilesRegistry();
         profilesRegistry.removeProfile(profile);
-        profilesRegistry.saveSettings();
+        if (!DBWorkbench.isDistributed()) {
+            profilesRegistry.saveSettings();
+        }
```
SolovevAnton, line 200: sry could you please explain or add a comment? I don't really get why it works. It seems that the bug can be in the saveSettings method in the distributed  rather than here, no?

PR #40508, opened 2026-03-09
Add GizmoSQL JDBC driver

prmoore77:
## Summary
- Add [GizmoSQL](https://github.com/gizmodata/gizmosql) as a recognized JDBC driver in the generic driver plugin
- Include 4 icon PNGs (16x16, 32x32, 64x64, 128x128) with transparent backgrounds
- Driver connects via `jdbc:gizmosql://` scheme, Maven artifact: `com.gizmodata:gizmosql-jdbc-driver`

## Details
[GizmoSQL](https://github.com/gizmodata/gizmosql-jdbc-driver) is a JDBC driver published on Maven Central. This PR adds it as a named driver so users can select it from the driver list instead of manually configuring a generic JDBC connection.

Driver properties surfaced in the connection dialog:
- `useEncryption` (default: `true`)
- `disableCertificateVerification` (default: `false`)
- `catalog`

Related: #15535 (Arrow Flight SQL support request)

Issue #15535, opened 2022-02-17
Future support for Arrow Flight SQL 


jmcastagnetto:
**Is your feature request related to a problem? Please describe.**

A solution being worked in the Apache Arrow community is the use of an SQL layer over the Arrow Flight protocol, which will allow a client to communicate in a simpler way with a DB backend w/o the need for a multiplicity of drivers.

See for example: https://github.com/apache/arrow/pull/10906 and https://arrow.apache.org/blog/2022/02/16/introducing-arrow-flight-sql/

**Describe the solution you'd like**

Use an Arrow Flight SQL driver to connect to multiple DBs:

![Image from https://www.dremio.com/subsurface/arrow-flight-sql-a-universal-jdbc-driver/](https://user-images.githubusercontent.com/364668/154537754-295ee9c3-1250-443d-bdea-f9ffff93d087.png)
(Image from: https://www.dremio.com/subsurface/arrow-flight-sql-a-universal-jdbc-driver/)

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

flambchap:
Hello @jmcastagnetto. Could you please tell me what benefits does it bring? What are the use cases? Thanks.

jmcastagnetto:
The Arrow Flight SQL is a layer on top of the Arrow Flight protocol [1], and provides a known interface (SQL) to the "on-the-wire" representation of the data, making it easy to interact with datastores that support Arrow Flight.

Arrow Flight has better performance than JDBC or ODBC [2], and allows accessing huge amounts of data in batch mode, apart from supporting parallel and partitioned data access [1].

IMHO those are enough benefits (and not the only ones planned, AFAICT) to think about supporting (in the future) data access using either the lower level Apache Flight or a higher level Apache Flight SQL layer. Not sure what would be the amount of work needed or what changes will be needed in dBeaver. It seems like there is a reference implementation of a JDBC-style driver [3]

HTH

- [1] https://arrow.apache.org/blog/2019/10/13/introducing-arrow-flight/
- [2] https://www.dremio.com/is-time-to-replace-odbc-jdbc/
- [3] https://www.dremio.com/subsurface/arrow-flight-sql-a-universal-jdbc-driver/

jmcastagnetto:
[...]

PR #41908, opened 2026-08-25
GizmoSQL: page result sets with LIMIT/OFFSET instead of re-execute-and-skip

prmoore77:
## Summary
- `GizmoSQLMetaModel` now implements `DBCQueryTransformProvider` and returns a `QueryTransformerLimit(false, true)` for `RESULT_SET_LIMIT`, so result-set paging is expressed as `LIMIT n OFFSET m` — the same approach the PostgreSQL extension uses.
- One file changed, no driver/plugin.xml changes.

## Details
GizmoSQL speaks Arrow Flight SQL, whose result streams are forward-only with no server-side cursor. Without a dialect-level limit transformer, DBeaver falls back to the generic strategy for every "fetch next page": re-execute the full statement, then `JDBCUtils.scrollResultSet()` skips the preceding rows client-side. For a wide table that means page *k* re-runs the query and transfers `k × fetchSize` rows over gRPC (quadratic in pages), visible server-side as one `CreatePreparedStatement` per page.

With this change each page becomes `SELECT ... LIMIT 200 OFFSET 400`, which both GizmoSQL backends (DuckDB and SQLite) push down with early termination. Ordering semantics are unchanged from the existing behaviour: both approaches re-execute the statement per page, so without an `ORDER BY` they carry the same (dialect-standard) caveat.

Verified end-to-end: with the patched `GizmoSQLMetaModel` spliced into a DBeaver 25.x install and the SQL-limit path active, paging `SELECT * FROM lineitem` in the SQL editor reached the GizmoSQL server as `SELECT * FROM lineitem LIMIT 200`, `… LIMIT 400`, `… LIMIT 600`, `… LIMIT 800` (one prepared statement per page, bounded scan) instead of the bare statement re-executed per page. The `GenericDataSource` change was compile-checked against the current `org.jkiss.dbeaver.*` bundles.

Follow-up to #40508 (GizmoSQL driver). Related: #15535.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericDataSource.java
+++ b/plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/GenericDataSource.java
@@ -44,6 +44,7 @@ import org.jkiss.dbeaver.model.meta.ForTest;
 import org.jkiss.dbeaver.model.runtime.DBRProgressMonitor;
 import org.jkiss.dbeaver.model.runtime.VoidProgressMonitor;
 import org.jkiss.dbeaver.model.sql.SQLDialect;
+import org.jkiss.dbeaver.model.sql.SQLQuery;
 import org.jkiss.dbeaver.model.struct.DBSDataType;
 import org.jkiss.dbeaver.model.struct.DBSObject;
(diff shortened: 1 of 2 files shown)


prmoore77:
Closing for now — on reflection the current generic re-execute-and-skip paging is acceptable for GizmoSQL in practice, and this change touches GenericDataSource for all generic dialects. May revisit if paging cost becomes a real problem.

PR #41911, opened 2026-08-25
dbeaver/pro#10173 Prompt to commit transaction after changing defaults

ShadelessFox:
It's required at least in PostgreSQL because it's done using `SET search_path = ...`, which is a transactional statement.

STR:
1. Enable manual commit
2. Change active schema

image


SolovevAnton (dismissed):

IvanGorshechnikov (dismissed):

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/internal/UINavigatorMessages.properties
```diff
@@ -288,3 +288,6 @@ navigator_provider_element_tooltip_datasource_description = Description: {0}
 navigator_provider_element_tooltip_datasource_read_only = Read-only connection
 navigator_provider_element_tooltip_datasource_provided = Provided connection
 navigator_provider_element_tooltip_datasource_error = Error: {0}
+
+confirm_commit_after_defaults_change_title = Change active database
+confirm_commit_after_defaults_change_message = Do you want to commit current transaction after changing active database?
```
serge-rider, line 293: We need better message. Current one is very misterious for users. Why they need to commit?

serge-rider (dismissed):

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/internal/UINavigatorMessages.properties
ShadelessFox: I agree. But I feel it might become filled with database-specific quirks if we attempt to clarify this. Do you have any suggestions on how to improve it?

SolovevAnton (dismissed):

plugins/org.jkiss.dbeaver.ui.navigator/src/org/jkiss/dbeaver/ui/actions/datasource/SelectActiveSchemaHandler.java
```diff
@@ -235,16 +240,65 @@ private static void changeDataBaseSelection(
...
+     */
+    private static void commitTransactionIfNeeded(
+        @NotNull DBRProgressMonitor monitor,
+        @Nullable DBCExecutionContext executionContext
+    ) throws DBCException {
+        var transactionManager = DBUtils.getTransactionManager(executionContext);
+        if (transactionManager == null || transactionManager.isAutoCommit()) {
+            return;
+        }
+        var contextDefaults = executionContext.getContextDefaults();
```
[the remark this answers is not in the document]
ShadelessFox: Wrong, `DBUtils.getTransactionManager` as well accepts a nullable reference and does a null check there.

Issue #35505, opened 2024-09-06
Shorten the name of a detached tab


ask9:
### Description

Perhaps it would be better to shorten the name of a detached tab.
Seems like it is not very useful to write the text of the whole query there.

- use code:
```
select "oid", relname, relnamespace, reltype, reloftype, relowner, relam, relfilenode, reltablespace, relpages, reltuples, relallvisible, reltoastrelid, relhasindex, relisshared, relpersistence, relkind, relnatts, relchecks, relhasrules, relhastriggers, relhassubclass, relrowsecurity, relforcerowsecurity, relispopulated, relreplident, relispartition, relrewrite, relfrozenxid, relminmxid, relacl, reloptions, relpartbound
from pg_class
```

https://github.com/user-attachments/assets/b4fc000f-04b3-4c87-a68e-ea05719185bd

### DBeaver Version

Community Edition 24.1.5-202408122322

### Operating System

Windows 10

### Database and driver

PostgreSQL 16
PostgreSQL JDBC Driver 42.7.4

anuanu0-0:
Hi, can I take up this issue?

E1izabeth:
@anuanu0-0 yes

E1izabeth:
The preferred solution would be to limit query text to 30 symbols

anuanu0-0:
Thanks will be taking this up then.

anuanu0-0:
Hi @E1izabeth can you guide me to a wiki or give me a general idea of the code to help me get started. (Any debugging related docs)

E1izabeth:
Hi, @anuanu0-0 
Here is our wiki page with guide to set-up development environment in IDEA https://github.com/dbeaver/dbeaver/wiki/Develop-in-IDEA or Eclipse https://github.com/dbeaver/dbeaver/wiki/Develop-in-Eclipse

0xgitpushpray:
It appears to me that this has not yet been worked on? @E1izabeth

E1izabeth:
It looks so

PR #41913, opened 2026-08-25
#35505 Truncate query text used as detached result tab name

0xgitpushpray:
## Problem

When a query result tab is detached into its own editor, the tab's title is generated from `SQLResultsEditorInput.getName()`, which uses `SQLQueryDataContainer.getName()`, the raw, untruncated SQL query text. For any non-trivial query, this makes the detached tab title unreadable.

## Fix
Truncate the name using the existing `CommonUtils.truncateString` utility (already used the same way elsewhere in the codebase, e.g. `ConnectionNameResolver`), capped at 50 characters.

## Scope
Single-method change in `SQLResultsEditorInput.getName()`. No other behavior is affected. The result tab's tooltip is unchanged. It does not currently show the full query text for SQL results (it shows a generic "SQL Query" label), so this is not a regression, but could be a good separate follow-up if useful.

## Testing
- Manually verified in a local dev build: detaching a long query's result tab now shows a truncated, readable tab title instead of the full query text.

## AI disclosure
This PR was generated with AI (Claude). I reviewed and understand the change before submitting.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLResultsEditorInput.java
+++ b/plugins/org.jkiss.dbeaver.ui.editors.sql/src/org/jkiss/dbeaver/ui/editors/sql/SQLResultsEditorInput.java
@@ -1,6 +1,6 @@
 /*
  * DBeaver - Universal Database Manager
- * Copyright (C) 2010-2024 DBeaver Corp and others
+ * Copyright (C) 2010-2025 DBeaver Corp and others
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


E1izabeth (approved):

ShadelessFox:
This is a weird workaround. Can't we fix the title to use the text of the originating result tab?

0xgitpushpray:
@ShadelessFox we could, however, I went the `truncateString` way, given that it's DBeaver's existing utility and already used many times in the codebase.

ShadelessFox:
What kind of an answer is that? I'm asking why we can't fix the root cause of the issue instead of fixing what it's caused.

ShadelessFox:
Closing in favor of https://github.com/dbeaver/dbeaver/pull/41916

PR #41919, opened 2026-08-26
Mark non-online SQL Server databases in the navigator

### LINKED ISSUE #22602: Better visual support for "offline databases" (as in MS SQL server)
Sometimes you take a "database offline" before deleting it from a database server.

When connecting to an Microsoft SQL Server instance there could be two possible database types:
 a) regular databases that are available (**online**)
 b) databases that are still on the Microsoft SQL Server - but have been taken **offline** (but still exist on DB server)
 
In DBeaver both types are shown in the left database tree of a connection but both types unfortunately with the same icon - 
 so it is really hard to distinguish them. 
If one double clicks such an offline database DBeaver is already able to show a message that such a database is offline - but it is cumbersome to go through all of them. 
 
Other DB tools show them with a different icon to mark such "offline databases". This way they are easier to recognize. 

Would be a nice addition to DBeaver too to use an own icon/icon color for offline DBs or mark them as "databaseName (offline)" at least.

image

See https://www.mssqltips.com/sqlservertip/6418/how-to-take-sql-server-database-offline/ for reference

michalpsyk:
Fixes #22602

Online and offline MS SQL Server databases currently share the same icon in the navigator, so the only way to tell them apart is to double-click each one and read the error.

### What this does

`SQLServerDatabase` now implements `DBPStatefulObject`. `DBNDatabaseNode.getNodeIcon()` already routes any stateful object through `DBNModel.getStateOverlayImage()`, so a database whose `sys.databases.state_desc` is not `ONLINE` gets an error overlay in the tree and in the object list. The raw state is also exposed as a viewable property, which adds a sortable **State** column to the connection's *Databases* tab.

The overlay title carries the actual `state_desc`, so `RESTORING`, `RECOVERY_PENDING` and `SUSPECT` are distinguishable rather than being collapsed into one generic "invalid".

### Notes

- No query change: the database cache already runs `SELECT db.* FROM sys.databases db`, so `state_desc` was in the result set and just was not being read.
- A missing `state_desc` maps to `DBSObjectState.NORMAL`, not `UNKNOWN`. `JDBCUtils.safeGetString` returns `null` on servers that do not have that column (SQL Server 2000 / jTDS), and `UNKNOWN` would put a "?" overlay on every database there.
- `refreshObjectState()` runs on the data source context rather than on the database itself, because an offline database cannot execute the query about its own state.
- `refreshObject()` re-reads the state, so refreshing a single database node updates the marker (the cached object is reused there, so without this the marker would go stale).
- The `State` column label follows the existing `meta.<class>.<property>.name` convention in `OSGI-INF/l10n/bundle.properties`; only the English bundle is updated.

### Testing

Built the CE product from this branch and ran it against SQL Server 2019 with 757 databases: 9 `OFFLINE` and 6 `RESTORING` databases are marked and sort together under the new column, the remaining `ONLINE` ones are unchanged.

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDatabase.java
+++ b/plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDatabase.java
@@ -23,6 +23,7 @@ import org.jkiss.dbeaver.Log;
 import org.jkiss.dbeaver.ext.mssql.SQLServerConstants;
 import org.jkiss.dbeaver.ext.mssql.SQLServerUtils;
 import org.jkiss.dbeaver.model.*;
+import org.jkiss.dbeaver.model.exec.DBCException;
 import org.jkiss.dbeaver.model.exec.jdbc.JDBCPreparedStatement;
 import org.jkiss.dbeaver.model.exec.jdbc.JDBCResultSet;
 import org.jkiss.dbeaver.model.exec.jdbc.JDBCSession;
@@ -38,6 +39,7 @@ import org.jkiss.dbeaver.model.runtime.VoidProgressMonitor;
 import org.jkiss.dbeaver.model.sql.DBSQLException;
 import org.jkiss.dbeaver.model.struct.DBSObject;
 import org.jkiss.dbeaver.model.struct.DBSObjectFilter;
+import org.jkiss.dbeaver.model.struct.DBSObjectState;
 import org.jkiss.dbeaver.model.struct.rdb.DBSCatalog;
 import org.jkiss.utils.ArrayUtils;
 import org.jkiss.utils.CommonUtils;
@@ -57,7 +59,8 @@ public class SQLServerDatabase
         DBPRefreshableObject,
         DBPSystemObject,
         DBPNamedObject2,
-        DBPObjectStatistics {
+        DBPObjectStatistics,
+        DBPStatefulObject {
 
     private static final Log log = Log.getLog(SQLServerDatabase.class);
 
@@ -67,6 +70,8 @@ public class SQLServerDatabase
     private boolean persisted;
     private String name;
     private String description;
+    private String stateDesc;
+    private DBSObjectState objectState = DBSObjectState.NORMAL;
     private DataTypeCache typesCache = new DataTypeCache();
     private SchemaCache schemaCache = new SchemaCache();
     private TriggerCache triggerCache = new TriggerCache();
@@ -84,6 +89,7 @@ public class SQLServerDatabase
         this.name = name;
         this.isTempDatabase = name.equalsIgnoreCase(SQLServerConstants.TEMPDB_DATABASE);
         //this.description = JDBCUtils.safeGetString(resultSet, "description");
+        setStateDesc(JDBCUtils.safeGetString(resultSet, "state_desc"));
 
         this.persisted = true;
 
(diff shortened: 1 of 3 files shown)


Destrolaric:
@michalpsyk Please provide the resulting UI of this feature

michalpsyk:
@Destrolaric

Setup: SQL Server 2022, three databases taken offline with `ALTER DATABASE ... SET OFFLINE`, and two left in `RESTORING` by a `RESTORE ... WITH NORECOVERY`, so these are real states rather than simulated ones.

The first screenshot is the navigator tree. `analytics` (online) sits directly above `archive_2023` (offline), so the icon difference is visible without comparing two images.

The second is the connection's *Databases* tab sorted by the new `State` column. The overlay is applied to every state other than `ONLINE`, and the column shows the raw `state_desc`, which is why `RESTORING` stays distinguishable from `OFFLINE` instead of both collapsing into one generic "not online" marker.

A database that reports no state at all (SQL Server 2000 / jTDS, where the column does not exist) is deliberately left unmarked rather than marked as unknown, so nothing changes visually on servers that cannot report it.

image

image

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDatabase.java
```diff
@@ -134,6 +140,47 @@ public void setDescription(String description) {
...
+            try (JDBCPreparedStatement dbStat = session.prepareStatement(
+                "SELECT state_desc FROM sys.databases WHERE database_id = ?"))
+            {
+                dbStat.setLong(1, databaseId);
+                try (JDBCResultSet dbResult = dbStat.executeQuery()) {
+                    if (dbResult.next()) {
+                        setStateDesc(JDBCUtils.safeGetString(dbResult, "state_desc"));
+                    }
+                }
+            }
```
ShadelessFox, line 168: You can use `org.jkiss.dbeaver.model.impl.jdbc.JDBCUtils#queryString`

ShadelessFox (dismissed):
Thanks for the contribution! Please see my comment :^)

E1izabeth:
@codex review

E1izabeth:
@michalpsyk please resolve conflicts

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDatabase.java
michalpsyk: Done in b30980433e.

One behaviour note worth flagging: `queryString` returns `null` when the query matches no row, where the previous code kept the last known state. A database that is gone from `sys.databases` does not exist any more, so clearing its marker is the better of the two.

michalpsyk:
@E1izabeth conflicts resolved, devel is merged in (7c61c19d45). All three conflicts were additions to the same regions rather than competing edits, so both sides are kept: `DBPStatefulObject` next to `DBSCollationProvider` in the interface list, the state fields next to the collation fields, and both cache resets in `refreshObject`. `BUILD SUCCESS` locally on the merged branch.

I merged current devel into #41921 and #41925 as well, so all three are up to date.

Unrelated, but it cost me a build: `tools/build.sh` clones `dbeaver/datadam-api` at `main`, and that repository has no `main` branch - its default is `devel` - so a fresh build stops at the clone step with `fatal: Remote branch main not found in upstream origin`. `DATADAM_API_REF=devel` works around it.

E1izabeth:
@codex review

E1izabeth:
@codex review

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDatabase.java
```diff
@@ -160,6 +166,41 @@ public void setDescription(String description) {
...
+    }
+
+    @Override
+    public void refreshObjectState(@NotNull DBRProgressMonitor monitor) throws DBCException {
+        // sys.databases is queried on the data source context: an offline database cannot host the query itself
+        try (JDBCSession session = DBUtils.openMetaSession(monitor, dataSource, "Read database state")) {
+            setStateDesc(JDBCUtils.queryString(
+                session,
+                "SELECT state_desc FROM sys.databases WHERE database_id = ?",
+                databaseId));
```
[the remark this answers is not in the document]
michalpsyk: Fixed in f4771560ff, though not for the reasons given - two parts of this are wrong, and the part that is right is broader than SQL Server 2000.

`refreshObjectState()` is not called by the entity editor after saving. The only generic call site is `AbstractCommandContext#saveChanges`, and it sits inside a commented-out block. For `SQLServerDatabase` the one real caller is `refreshObject()` in this class, which already catches `DBCException` and logs it at debug level, so no save can fail because of this.

SQL Server 2000 does not reach this class at all. `SQLServerDataSource.DatabaseCache` loads the list with `SELECT db.* FROM sys.databases db` unconditionally, and `sys.databases` only exists from 2005 on, so the database list cannot load there in the first place. The legacy drivers and Sybase/SAP ASE are registered under `SQLServerGenericDataSourceProvider`, a different model entirely.

What is real is an asymmetry I introduced: the database list reads the column through `JDBCUtils.safeGetString`, which returns `null` for a column that is not there, while the refresh selected `state_desc` by name and would fail hard. Any server whose `sys.databases` does not expose the column - Babelfish is the plausible candidate, not 2000 - would be marked on load and error on refresh. The refresh now uses the same shape as the list, so both leave such a database unmarked.

The comment about pre-2005 servers is gone too. It was wrong, and it is what made this finding land on the wrong server version.

plugins/org.jkiss.dbeaver.ext.mssql/src/org/jkiss/dbeaver/ext/mssql/model/SQLServerDatabase.java
michalpsyk: @ShadelessFox heads-up: f4771560ff partly reverts this, and I would rather flag it than have you find it.

A later Copilot pass pointed out that the refresh fails on a server whose `sys.databases` does not expose `state_desc`, while the database list reads that same column through `JDBCUtils.safeGetString` and just leaves the database unmarked. That asymmetry was real, so the refresh now reads the column the same way the list does.

The two cannot be combined: surviving a missing column means selecting `db.*`, and `queryString` reads column 1 by index, which would then be `name` instead of `state_desc`. If you would rather keep the helper and accept the hard failure on such servers, say so and I will put it back.

PR #41939, opened 2026-08-26
dbeaver/pro#10400 Fix statements separation for DDL containing IF (NOT) EXISTS for BigQuery

E1izabeth:
Closes https://github.com/dbeaver/pro/issues/10400


plugins/org.jkiss.dbeaver.ext.bigquery/src/org/jkiss/dbeaver/ext/bigquery/model/BQSQLDialect.java
```diff
@@ -84,4 +96,60 @@ public EnumSet<ProjectionAliasVisibilityScope> getProjectionAliasVisibilityScope
...
+
+    @Override
+    public void initDriverSettings(@NotNull JDBCSession session, @NotNull JDBCDataSource dataSource, @NotNull JDBCDatabaseMetaData metaData) {
+        super.initDriverSettings(session, dataSource, metaData);
+        cachedDialectSkipTokenPredicates = this.makeDialectSkipTokenPredicates(dataSource);
+    }
+
+    @NotNull
+    protected SQLTokenPredicateSet makeDialectSkipTokenPredicates(@NotNull JDBCDataSource dataSource) {
+        SQLSyntaxManager syntaxManager = new SQLSyntaxManager();
```
serge-rider: Can we move this part somewhere in GenericSQLDialect or JDBCSQLDialect?
We have similar thing in Oracle and probably will have it in other places.

So'd dialects have only implementation of makeDialectSkipTokenPredicatesImpl (which returns null by default)

serge-rider (changes requested):

Matvey16:
Verified

DBeaver CE Windows
DBeaver EE Windows

PR #41942, opened 2026-08-27
dbeaver/pro#10278 Support all data export formats

IvanGorshechnikov:
Closes dbeaver/pro#10278

## Summary
- resolve data transfer processors by their short ID when the format does not match the processor ID
- fall back to the default file extension for formats such as Markdown and PHP source code
- preserve direct processor ID lookup and prioritize short IDs over ambiguous extensions

## Testing
- `mvn clean verify` in `plugins/org.jkiss.dbeaver.data.transfer.ui`
- `git diff --check`

### DIFF
--- a/plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/SQLPragmaExport.java
+++ b/plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/SQLPragmaExport.java
@@ -49,6 +49,7 @@ public class SQLPragmaExport implements SQLPragmaHandler {
 
     public static final String PARAMETER_TYPE = "type";
     public static final String PARAMETER_INCLUDE_PIPES = "includePipesConfiguration";
+    public static final String PARAMETER_SEARCH_SIMILAR = "searchSimilar";
 
     private static final String PRODUCER_NODE_ID = "database_producer";
     private static final String CONSUMER_NODE_ID = "stream_consumer";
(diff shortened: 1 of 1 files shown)


plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/SQLPragmaExport.java
```diff
@@ -108,6 +109,30 @@ protected boolean includePipesConfigurationPage() {
         return RESULT_CONSUME_PRAGMA | RESULT_CONSUME_QUERY;
     }
 
+    @Nullable
+    private static DataTransferProcessorDescriptor findProcessor(
```
SolovevAnton, line 115: Are we sure to use a static method here?

SolovevAnton (dismissed):

plugins/org.jkiss.dbeaver.data.transfer.ui/src/org/jkiss/dbeaver/tools/transfer/ui/SQLPragmaExport.java
```diff
@@ -108,6 +109,30 @@ protected boolean includePipesConfigurationPage() {
...
+        @NotNull DataTransferRegistry registry,
+        @NotNull DataTransferNodeDescriptor consumerNode,
+        @NotNull String type
+    ) {
+        DataTransferProcessorDescriptor processor = registry.getProcessor(PROCESSOR_ID_PREFIX + type);
+        if (processor != null) {
+            return processor;
+        }
+
```
serge-rider, line 125: This is potentially errornous behavior in non-AI usages.
E.g. file extensions may conflict and identity processors in non-unique way.
Let's add a parameter like "searchSimilar", false by default and true for AI calls.

serge-rider (changes requested):

PR #41947, opened 2026-08-27
#41909 #41941 Read MariaDB check constraints without joining TABLE_CONSTRAINTS

### LINKED ISSUE #41909: MariaDB named CHECK constraints are not shown in the Constraints tab
### Description

Named table-level CHECK constraints exist in MariaDB and are returned by `information_schema.CHECK_CONSTRAINTS`, but DBeaver only shows the primary key in the table editor's **Constraints** tab.

The constraints are present in the generated DDL and are enforced by MariaDB.

Expected: all named CHECK constraints are listed with their expressions.

Actual: only the primary key is listed.

This appears related to #6069. It may also overlap with #38701, but this case uses explicitly named table-level constraints rather than inline column constraints.

### DBeaver Version

Community Edition 26.1.5

### Operating System

_No response_

### Database and driver

1. Database: MariaDB 10.11.10
2. Driver: MariaDB Connector/J 3.5.2

### Steps to reproduce

1. Create a table:

```sql
CREATE TABLE check_demo (
    id INT PRIMARY KEY,
    status VARCHAR(20) NOT NULL,
    payload LONGTEXT,
    CONSTRAINT chk_status CHECK (status IN ('ACTIVE', 'INACTIVE')),
    CONSTRAINT chk_payload CHECK (payload IS NULL OR JSON_VALID(payload))
);
```

2. Refresh the schema and open **Table → Properties → Constraints**.
3. Only the primary key is shown.
4. Confirm the constraints exist:

```sql
SELECT CONSTRAINT_NAME, CHECK_CLAUSE
FROM information_schema.CHECK_CONSTRAINTS
WHERE CONSTRAINT_SCHEMA = DATABASE()
  AND TABLE_NAME = 'check_demo';
```

Both `chk_status` and `chk_payload` are returned.

### Additional context

The same behavior occurs with named table-level checks and MariaDB-generated JSON validation checks.

### LINKED ISSUE #41941: MariaDB: Constraints tab can show the CHECK expression of a different table
### Description

`MySQLCatalog.CheckConstraintCache` joins `INFORMATION_SCHEMA.CHECK_CONSTRAINTS` with `INFORMATION_SCHEMA.TABLE_CONSTRAINTS` on `CONSTRAINT_SCHEMA` and `CONSTRAINT_NAME` only.

In MariaDB a check constraint name only has to be unique within its own table, so two tables in the same schema can reuse the same name. The join then mixes them up, and the Constraints tab can show a constraint with an expression that belongs to another table. Which one you get depends on row order, since the ORDER BY ties.

The same join also blows up the schema-wide read. In a schema with 30 tables and 30 check constraints it returns 634 rows.

MySQL is not affected. Its `CHECK_CONSTRAINTS` has no `TABLE_NAME` column, but it also refuses duplicate check constraint names inside a schema (error 3822), so matching on the name alone is already unambiguous there.

### DBeaver Version

Community Edition 26.1.5.202608161900

### Operating System

macOS 26.6.1 (aarch64)

### Database and driver

1. Database: MariaDB 10.11.10
2. Driver: MariaDB Connector/J 3.5.2

### Steps to reproduce

1. Create two tables that reuse a constraint name:

```sql
CREATE TABLE orders (
  id INT PRIMARY KEY, status VARCHAR(20),
  CONSTRAINT chk_status CHECK (status IN ('NEW','DONE'))
);
CREATE TABLE invoices (
  id INT PRIMARY KEY, status VARCHAR(20),
  CONSTRAINT chk_status CHECK (status IN ('PAID','VOID'))
);
```

2. Open either table, then **Properties → Constraints**.
3. `chk_status` can be listed with the other table's expression.

The query DBeaver runs for `orders` returns both rows:

```
orders  chk_status  `status` in ('NEW','DONE')
orders  chk_status  `status` in ('PAID','VOID')   <- belongs to invoices
```

### Additional context

MariaDB does expose `TABLE_NAME` in `CHECK_CONSTRAINTS`, so adding `AND tc.TABLE_NAME = cc.TABLE_NAME` to the join fixes both problems. With that predicate the schema-wide query above returns 30 rows instead of 634.

Since MySQL has no such column, the predicate should only be added when `MySQLDataSource.isMariaDB()` is true.

I already have a patch with tests and can open a PR if that helps.

zanlucathiago:
Fixes #41909 and fixes #41941. Both come from the same join, so they are fixed together.

### The problem

`CheckConstraintCache` reads check constraints by joining `CHECK_CONSTRAINTS` with `TABLE_CONSTRAINTS` on schema + constraint name. On MariaDB that causes two separate issues.

**#41909 - no check constraints for read-only users.** MariaDB returns no rows from `TABLE_CONSTRAINTS` to an account that only holds `SELECT` on the schema. `CHECK_CONSTRAINTS` has no such restriction, so the constraints are there but the join drops all of them and the Constraints tab shows only the primary key. Any extra privilege is enough to hide the problem, which is why it does not show up when testing as root:

| Grant on the schema | Rows in `TABLE_CONSTRAINTS` |
| --- | --- |
| `SELECT` | 0 |
| `SELECT, SHOW VIEW` | 8 |
| `SELECT, REFERENCES` | 8 |
| `SELECT, INDEX` | 8 |
| `ALL PRIVILEGES` | 8 |

**#41941 - wrong expression shown.** MariaDB scopes check constraint names per table, so two tables in one schema can both have a constraint named `chk_status`. Matching on schema + name alone cross-matches them, and since `fetchObject` only runs for the first row of each name, a table can end up displaying another table's CHECK expression. It also makes the result set grow quadratically (30 constraints produced 634 rows in a test schema).

### The fix

MariaDB exposes `TABLE_NAME` in `CHECK_CONSTRAINTS`, so `TABLE_CONSTRAINTS` is not needed at all there. Dropping it fixes both issues.

MySQL is left alone: `CHECK_CONSTRAINTS` has no `TABLE_NAME` column there, so the join is the only way to know the owning table, and MySQL neither filters `TABLE_CONSTRAINTS` by privilege nor allows duplicate check constraint names within a schema. The two variants are now separate methods behind the existing `isMariaDB()` check.

### Verification

Tested against MariaDB 10.11.10 and MySQL 8.0, using the query the patched code actually generates:

| Scenario | Before | After |
| --- | --- | --- |
| MariaDB, `SELECT`-only user | 0 constraints | 5 constraints |
| MariaDB, root | 5 constraints | 5 constraints |
| MySQL 8, `SELECT`-only user | 1 constraint | 1 constraint |
| MariaDB, same name in two tables | expression of the wrong table | correct expression |

Steps to reproduce #41909 on a fresh MariaDB:

```sql
CREATE DATABASE demo;
CREATE TABLE demo.t (id INT PRIMARY KEY, qty INT,
  CONSTRAINT chk_qty CHECK (qty > 0));
CREATE USER 'ro'@'%' IDENTIFIED BY 'ro';
GRANT SELECT ON demo.* TO 'ro'@'%';
```

Connect as `ro` and open the Constraints tab of `demo.t`: only `PRIMARY` is listed before the patch, `chk_qty` shows up after it.

Added `MySQLCheckConstraintCacheTest` with 6 regression tests covering both dialects and both parameter shapes.


E1izabeth (approved):
Thanks for the contribution!

uslss:
verified (MySQL 8.0.45)
image

PR #41976, opened 2026-08-31
dbeaver/pro#7922 Close data transfer producers after use

vkirst:


serge-rider:
@codex review

plugins/org.jkiss.dbeaver.data.transfer/src/org/jkiss/dbeaver/tools/transfer/DataTransferJob.java
```diff
@@ -129,47 +130,60 @@ protected void canceling() {
         super.canceling();
     }
 
-    private void transferData(
+    @ForTest
```
serge-rider: Why ForTest here?

serge-rider (changes requested):

PR #41981, opened 2026-09-01
dbeaver/pro#41967 DuckDB force SQL limit

### LINKED ISSUE #41967: No LIMIT is set for DuckDB data view queries
### Description

When opening a big table in a DuckDB database, DBeaver sends a `SELECT * FROM table`-like query which makes DuckDB read it all (or big part of it) and the DBeaver just discard almost all rows from the resultset on the client. For a 70M table, Eclipse peaks at ~12GiB memory usage (in DuckDB driver versions before 15 it stay on the peak, on 15 it drops to ~6GiB). For a 500M table, it thrashes the machine to halt. Basically #4361 for DuckDB.

### DBeaver Version

Community Edition 26.1.5

### Operating System

Linux Mint 22

### Database and driver

All downloadable DuckDB drivers, 13—15

### Steps to reproduce

1. Connect to a database (with "Use SQL to limit fetch size")
2. Open a big table
3. Open Query Manager

Expected result: `SELECT t.* FROM db.main.table_name AS t LIMIT 200`
Actual result: `SELECT t.* FROM db.main.table_name AS t`

### Additional context

_No response_

serge-rider:


IvanGorshechnikov (approved):

E1izabeth (approved):

Matvey16:
Still no SQL limit

image

Matvey16:
Now DuckDB specifically overrides global preferences with `Use SQL to limit fetch size` even if connection specific settings are disabled. The only way to not use SQL for limit on DuckDB is to disable it in connection settings.

STR
1. Disable SQL for limits in global preferences
2. Disable DuckDB connection settings for data editor
3. Run select on a big table

SQL would still be used


https://github.com/user-attachments/assets/0cf4518d-cf4d-4d05-acab-0a7572b10d74

serge-rider:
Yes, and this is how we manage this for all drivers, duckdb is not an exception.

PR #41982, opened 2026-09-01
dbeaver/pro#10369 Handle unability to create new log file

E1izabeth:
Closes dbeaver/pro#10369

Updated log-file rotation algorithm:
 1. Rename current log file to have a suffix of the current timestamp. Takes a number of attempts. If all attempts fails, then:
    - leave that initial file untouched for the remainder of the application run;
    - switch from default log file name to file name having timestamp and '-since' suffix (switch output to <name>-<stamp>-since.<ext>)
    Files with timestamped names, including -since files whose later rename fails, remain eligible for retention cleanup.
 3. List all existing log files, two cases possible:
    - rename was (always) successful, so all file names are of the form `<name>-<stamp>.<ext>` (a);
    - rename failed at certain point (maybe multiple times), so there's a file with initial name (which is ignored) and maybe certain files of the `<name>-<stamp>-since.<ext>` form (b).
 4. Try to delete all the extra files exceeding `logs.files.output.maxCount` setting.
Each filesystem-related operation uses retry attempts with small timeout for a filesystem to sync.


plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/LogOutputStream.java
```diff
@@ -16,15 +16,21 @@
  */
 package org.jkiss.dbeaver;
 
+import org.eclipse.osgi.internal.debug.Debug;
```
serge-rider, line 19: m?

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/LogOutputStream.java
```diff
@@ -36,18 +42,91 @@ public class LogOutputStream extends OutputStream {
     public static final String LOGS_MAX_FILE_SIZE = "logs.files.output.maxSize";
     public static final String LOGS_MAX_FILES_COUNT = "logs.files.output.maxCount";
 
+    private static final int FILE_OPERATIONS_RETRY_LIMIT = 5;
 
+    private static final Operations REAL_OPERATIONS = new Operations() {
```
serge-rider, line 47: Why do we need this abstraction - single implementation of interface with single purpose?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/LogOutputStream.java
E1izabeth: It is used to observe log file rotation failures. At least when it is possible _somehow_.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/LogOutputStream.java
E1izabeth: There are a number of tests covering all potential filesystem-related failure points in the logfile rotation logic. These tests use another implementation of this interface to isolate the logic from the real file system and test only the logic itself.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/LogOutputStream.java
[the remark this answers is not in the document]
E1izabeth: We are already using Eclipse internals in the other parts of the project, and our release cycle is ok with that.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/LogOutputStream.java
```diff
@@ -132,40 +217,189 @@ public synchronized void close() throws IOException {
...
+            }
+        }
+        if (errors != null && errors.size() > 1) {
+            this.operations.debugPrint("Multiple exceptions occurred during IO operation retry, propagating the first one.");
+        }
+        if (initialException != null) {
+            throw new IOException("Failed to retry on IO operation due to the underlying exception", initialException);
+        } else {
+            throw new IOException("Failed to retry on IO operation due having unsuccessful result.");
+        }
```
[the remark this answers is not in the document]
E1izabeth: Addressed.

plugins/org.jkiss.dbeaver.model/src/org/jkiss/dbeaver/LogOutputStream.java
```diff
@@ -80,15 +165,15 @@ public LogOutputStream(@NotNull File debugLogFile) throws IOException {
...
-            this.currentLogSize = this.currentLogFile.length();
+        if (operations.exists(debugLogFile)) {
+            this.currentLogSize = operations.length(this.currentLogFile);
             this.rotateCurrentLogFile(true);
         } else {
             this.currentLogSize = 0;
-            if (this.logFileLocation.mkdirs()) {
+            if (operations.makeDirectories(this.logFileLocation)) {
                 throw new IOException("Failed to initialize debug log output location: " + debugLogFile.getAbsolutePath());
             }
```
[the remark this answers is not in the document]
E1izabeth: Addressed.

PR #41987, opened 2026-09-01
dbeaver/pro#10472: minimal fix

SolovevAnton:
closes dbeaver/pro#10472


plugins/org.jkiss.dbeaver.model/schema/org.jkiss.dbeaver.app.config.exsd
```diff
@@ -87,6 +87,16 @@
                </appInfo>
             </annotation>
          </attribute>
+         <attribute name="availabilityTester" type="string">
+            <annotation>
+               <documentation>
+
+               </documentation>
+               <appInfo>
+                  <meta.attribute kind="java" basedOn="org.jkiss.dbeaver.model.config.ProductConfigFeatureAvailabilityTester:"/>
```
E1izabeth, line 96: Why `:` is at the end?

SolovevAnton: tbh, I just copied the same pattern as I saw above with `ProductConfigFeatureTester` registration.

Now, i also wasn't able to find any Eclipse documentation about it (I assume because modern eclipse approach is doing everything from wizards) , but from this for example:
https://github.com/opcoach/e4Preferences/blob/c35b443c4ab0298ec5e39ed674ce722d94a9f8c2/com.opcoach.e4.preferences/schema/e4PreferencePages.exsd#L85
I assumed that the eclipse will register any class that implemets this interface.
What's more strange - it works with and without ":"

PR #41997, opened 2026-09-02
dbeaver/dbeaver#41996 Do not retry a statement after a transaction invalidation that did nothing

### LINKED ISSUE #41996: An aborted transaction opened with BEGIN cannot be recovered from the UI in auto-commit mode
> Updated after the discussion below. The original text proposed rolling the transaction back
> automatically; that was the wrong remedy and is withdrawn. The defect described here is
> unchanged, and the evidence for it is unchanged.

## Description

In auto-commit mode a transaction opened with `BEGIN` is invisible to DBeaver's transaction UI.
If a statement inside it fails, the transaction is left aborted and every subsequent statement
on that connection fails with `25P02`. There is no way out from the UI — only typing
`ROLLBACK;` by hand, or closing the editor to drop the connection.

Any statement error inside the transaction produces this; it is not specific to read-only
databases.

**Nothing in the UI reports that there is a transaction to act on.** Commit and Rollback are
greyed out in the toolbar and in `Database → Transaction mode`. `Database → Pending
transactions` lists nothing and its own Commit and Rollback are disabled too. The enablement
of those actions is derived from the JDBC auto-commit flag rather than from whether a
transaction is actually open:

```java
// DataSourcePropertyTester, PROP_TRANSACTION_ACTIVE
DBCTransactionManager txnManager = DBUtils.getTransactionManager(context);
return txnManager != null && !txnManager.isAutoCommit();
```

`org.jkiss.dbeaver.core.commit` and `org.jkiss.dbeaver.core.rollback` are both wired to that
property in `plugin.xml`, so in auto-commit mode neither can become enabled.
`QMMConnectionInfo.transactional` is set from the same flag, which is why Pending transactions
is empty as well.

**The automatic recovery cannot recover this state.** `DBExecUtils.tryExecuteRecover` reaches
`InvalidateJob.invalidateTransaction` for an error classified as `TRANSACTION_ABORTED`, and
that method rolls back only when the transaction manager is not in auto-commit mode, because
`JDBCExecutionContext.rollback` reaches `Connection.rollback()`, which JDBC does not permit
while auto-commit is on. It then retries the statement regardless, so each rejected statement
is executed twice. That retry is what #41997 addresses; it does not change this state.

`DBExecUtils.recoverSmartCommit` also returns immediately unless smart commit is enabled —
`TRANSACTIONS_SMART_COMMIT` and `TRANSACTIONS_SMART_COMMIT_RECOVER` both default to `false`.

The error is already classified: `PostgreDataSource.discoverErrorType` returns
`ErrorType.TRANSACTION_ABORTED` for `25P02`. That result is not used to offer the user a way
out.

## DBeaver Version

Community Edition 26.1.5.202608161900

## Operating System

macOS 26.6.2, Apple Silicon

## Database and driver

1. Database name and version: PostgreSQL 14.22 (aarch64-unknown-linux-gnu)
2. Driver name and version: PostgreSQL JDBC Driver (bundled)

## Steps to reproduce

On any PostgreSQL, no special setup:

1. Transaction mode: **Auto-Commit**.
2. In a new SQL editor, run these one at a time:

```sql
BEGIN;
SELECT 1/0;   -- 22012, aborts the transaction
SELECT 1;     -- SQL Error [25P02]: current transaction is aborted, commands ignored until end of transaction block
```

3. Commit and Rollback in the toolbar are greyed out, and `Database → Transaction mode →
   Rollback` is disabled as well. `Database → Pending transactions` lists nothing, and its own
   Commit and Rollback buttons are disabled too.
4. Every further statement on that connection fails with `25P02`. Typing `ROLLBACK;` clears it
   immediately; closing the editor also clears it, by dropping the connection.

## Additional context

First hit on an AWS PostgreSQL read replica (`pg_is_in_recovery() = true`,
`transaction_read_only = on`), where an `UPDATE` inside `BEGIN` fails with `25006` and leaves
the same state. The `SELECT 1/0` form above reproduces it on an ordinary writable database, so
the replica is incidental.

Without `BEGIN` there is no issue: a single failing `UPDATE` in auto-commit returns `25006` and
the next `SELECT` runs normally, since each statement is its own transaction.

The Scripts preference "Error handling: Stop + rollback" does not apply here. `SQLQueryJob`
gates its end-of-script commit/rollback block on `!oldAutoCommit`, and the scenario above
executes statements individually rather than as a script, so the setting is out of scope in two
ways rather than one.

Related reports:

- #3541 — closed with the advice to commit/rollback from the toolbar or enable rollback-on-error.
  Both of those are unavailable in auto-commit mode.
- #6479 — the auto-recover request that `recoverSmartCommit` implements, scoped to smart commit mode.
- #8094 — the same `BEGIN` + failing statement shape, closed as not reproducible. Auto-commit
  mode may be the condition that was missing.
- #41038 — same configuration (auto-commit on, transaction started with `BEGIN`), where the
  aborting statement was the row-count query DBeaver generates itself. Fixed by #41491, which is
  present in 26.1.5; this report is about the state that remains once a transaction has been
  aborted, whatever aborted it.

ethznn:
Follows the discussion in #41996. This is the part of it that stands on its own; the recovery affordance is a separate question still open there.

`InvalidateJob.invalidateTransaction` rolls back only when the transaction manager reports it is not in auto-commit mode, because `JDBCExecutionContext.rollback` reaches `Connection.rollback()`, which JDBC does not permit while auto-commit is on. In auto-commit it therefore returns having done nothing.

Its only external caller is `DBExecUtils.tryExecuteRecover`, which calls it for an error classified as `TRANSACTION_ABORTED` and then retries the statement. `EXECUTE_RECOVER_ENABLED` defaults to true and `EXECUTE_RECOVER_RETRY_COUNT` to 1, so in auto-commit mode a statement rejected for an aborted transaction is executed a second time against a transaction nothing has ended. The second attempt fails with the same error.

The change makes `invalidateTransaction` report whether a transaction was actually rolled back, and `tryExecuteRecover` retry only when it was. Nothing is rolled back that was not rolled back before, and the manual-commit path is unchanged.

Measured on PostgreSQL 14.22, reading the Query Manager log while running `BEGIN;` / `SELECT 1/0;` / `SELECT 1;` one statement at a time in auto-commit mode on a fresh connection.

Community Edition 26.1.5 — four rows for three statements:

```
BEGIN        0s       Success
SELECT 1/0   0.003s   ERROR: division by zero
SELECT 1     0s       ERROR: current transaction is aborted...
SELECT 1     0s       ERROR: current transaction is aborted...
```

A build of this branch — three rows:

```
BEGIN        0.001s   Success
SELECT 1/0   0.001s   ERROR: division by zero
SELECT 1     0.001s   ERROR: current transaction is aborted...
```

`SELECT 1/0` fails with `22012`, which is not `TRANSACTION_ABORTED`, so it is not eligible for the retry either way. The statement this changes is `SELECT 1`, which now reaches the user after one execution instead of two. It is a per-statement cost while the connection stays in that state: on a connection that was already holding an aborted transaction, so that every statement including `BEGIN` came back with `25P02`, all three were doubled.

Manual-commit mode was checked separately on the same build: a statement failing inside a transaction is still recovered and the following statement still succeeds without the error surfacing, as before.


SolovevAnton:
Please see the comment: https://github.com/dbeaver/dbeaver/issues/41996#issuecomment-5527174362

ethznn:
Understood on the issue, and `END` does clear it — in PostgreSQL that is `COMMIT`, which on an aborted transaction rolls back. No argument with the decision there.

This PR is a separate change though, so I wanted to check whether closing it was intended. It rolls nothing back, and it does not alter the behaviour that decision is about — the transaction stays aborted and the user still ends it themselves.

What it removes is a pointless round trip to the server. `invalidateTransaction` returns without doing anything while the context is in auto-commit mode, and `tryExecuteRecover` retries the statement regardless, so a statement already rejected with `25P02` is sent to PostgreSQL a second time and rejected again.

Query Manager on 26.1.5, running `BEGIN;` / `SELECT 1/0;` / `SELECT 1;` one at a time on a fresh connection — four rows for three statements:

```
BEGIN        0s       Success
SELECT 1/0   0.003s   ERROR: division by zero
SELECT 1     0s       ERROR: current transaction is aborted...
SELECT 1     0s       ERROR: current transaction is aborted...
```

Three rows with this branch. It is a per-statement cost for as long as the connection stays in that state.

If the state itself is expected, the second execution still serves no purpose. Would you reconsider this one on its own? Happy to leave it closed if you would rather not carry the change.

PR #42005, opened 2026-09-03
#41183 Adjust tab highlighting for colored connections

### LINKED ISSUE #41183: Connection type colors and editor tabs
### Discussed in https://github.com/orgs/dbeaver/discussions/41108

<sup>Originally posted by **aasherber-fadv** May 18, 2026</sup>
If I have an open editor window associated with a connection that has a color assigned to it (like the default Production pink), then all of the chrome around the editor window picks up that color -- except for the tab of the active window, which gets a lighter version of that color. This can get confusing when there are also inactive editor widows associated with the same connection, since the inactive windows pick up the main color, not the lighter one.

Is there a reason that things were done this way, or a setting to change it? I would expect the active tab to match the connection/chrome color, and inactive tabs to be lighter. </div>

E1izabeth:
Closes https://github.com/dbeaver/dbeaver/issues/41183

- light theme
  * with vivid color
    * normal rendering
    * with mouse cursor over the inactive tab
  * with light color
    * normal rendering
    * with mouse cursor over the inactive tab
 - dark theme (same combinations as for light)

image

image


Matvey16:
Verified

DBeaver CE Windows
Dbeaver CE Linux

PR #41550, opened 2026-07-16
dbeaver/dbeaver#40575 Fix psql dump

IvanGorshechnikov:
Closes https://github.com/dbeaver/dbeaver/issues/40575

Looks like regression after https://github.com/dbeaver/dbeaver/issues/38860

My STR:
image

The command result on "PostgreSQL dump progress" was
/opt/homebrew/Cellar/postgresql@17/17.10/bin/pg_dump --verbose --host=localhost --port=5432 ****** --format=c --file /Users/ivangorsecnikov/dump-postgres-202607161408.sql ******                                                                        
But it should be
/opt/homebrew/Cellar/postgresql@17/17.10/bin/pg_dump --verbose --host=localhost --port=5432 ****** --format=c --file /Users/ivangorsecnikov/dump-postgres-202607161409.sql -t test."1" ******

Also, now close the table picker if a complete backup is checked
image

And look at this tickets https://github.com/dbeaver/dbeaver/issues/40575#issuecomment-4852997550, they should be fixed

PR #38959, opened 2025-08-15
#38860 Postgres: hide table section when complete backup flag is set

mikiubo:

PR #42008, opened 2026-09-03
#40575 Respect table selection in PostgreSQL backup wizard

### LINKED ISSUE #40575: Tools -> Backup functionality backup whole db
### Description

Backup with selected tables functionality doesn't work anymore. Whatever I choose, system backup entire db.

### DBeaver Version

Community Edition 26.0.0.202603011816

### Operating System

Windows 11

### Database and driver

Database: PostgreSQL 16.8.0
DriverName: PostgreSQL JDBC Driver

### Steps to reproduce

Go to your database,
right click -> Tools and select Backup,
select schema,
Complete backup passive,
Select a few tables and observe the results.

### Additional context

_No response_

tarikhagustia:
Fixes #40575

The PostgreSQL backup wizard ignores the table selection and dumps the whole database instead.

### What was wrong

`PostgreBackupWizardPageObjects.updatefullSchemaBackupState()` wrote the setting from the schema check state rather than from the checkbox:

```java
boolean allSchemasSelected = isAllSchemaSelected();
fullSchemaBackupCheck.setEnabled(allSchemasSelected);
wizard.getSettings().setFullSchemaBackup(allSchemasSelected);
```

`updateState()` calls this on every check event, so whenever all schemas happened to be checked — a database with a single `public` schema, for instance — `fullSchemaBackup` was silently `true` while the "Complete backup" checkbox rendered unchecked. `PostgreDatabaseBackupHandler.fillProcessParameters` then returned before adding any `-n`/`-t`, and pg_dump dumped everything.

The widget and the setting stopped agreeing in 7f27ae84f5 (#38959), which dropped `fullSchemaBackupCheck.setSelection(allSchemasSelected)` from that method but left the setter call in place.

### Changes

1. **The checkbox is the single source of truth.** `updateFullSchemaBackupState()` now only enables the checkbox when all schemas are checked, clears it when they are not, and writes the setting plus the tables-section visibility from the widget state. The checkbox listener just calls `updateState()`, so there is no longer a second path that can write a different value.
2. **`loadSettings()` syncs the checkbox with the persisted setting**, so a `pg.export.fullSchemaBackup` carried over from an earlier session or a saved task is visible instead of hidden.
3. **`isAllSchemaSelected()` returns `false` for an empty list**, so the flag cannot switch on before the schemas are loaded.
4. **The racy table collection in `updateTableCheckedStatus()` is gone.** `loadTables()` schedules an `AbstractJob` and returns its result list immediately, so the caller read a list that the job had not filled yet and could store a partial set in `checkedObjects` — which `saveState()` turns into a `-t` list covering only some of the tables. A checked schema without an entry in `checkedObjects` is already dumped whole via `-n`, so dropping the entry is both sufficient and race-free. `loadTables()` now returns `void`, since its result was never usable.
5. `exportViewsCheck` was initialised from `isFullSchemaBackup()` instead of `isShowViews()`.

### Relation to #41550 / #41683

#41550 fixed point 1 and was reverted in #41694 because of #41683, where a schema backup restored with many tables missing. Point 4 is, as far as I can tell, what caused that: while the flag was being forced to `true` the handler ignored `exportObjects` entirely, which masked the race. Removing the forcing without removing the race makes the partial `-t` list observable, and that is exactly the reported symptom. This PR addresses both, so the earlier regression should not come back.

### How to test

Against a PostgreSQL database with a single `public` schema holding several tables:

1. Tools → Backup → **None** → check one table → Start. The pg_dump command line on the log page must contain `-t "public"."<table>"` and the dump must hold only that table. "Complete backup" stays unchecked and stays enabled.
2. Tick "Complete backup" → the table list is hidden → the command line carries neither `-n` nor `-t` (whole database).
3. Untick it → the table list returns with the previous checkboxes intact → `-t` is used again.
4. With a multi-schema database, uncheck one schema: "Complete backup" becomes disabled and clears itself; the backup uses `-n`/`-t` according to the selection.
5. Regression check for #41683: check a schema with all of its tables and press "All" before the table list has finished loading. The command line must be `-n "public"`, never a partial `-t` list, and a restore of that dump must contain every table.

https://claude.ai/code/session_01JrpGArXT5CsD6ZJWcqouKA

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/tools/PostgreBackupWizardPageObjects.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql.ui/src/org/jkiss/dbeaver/ext/postgresql/tools/PostgreBackupWizardPageObjects.java
@@ -111,13 +111,9 @@ class PostgreBackupWizardPageObjects extends AbstractNativeToolWizardPage<Postgr
             schemasTable.setLayoutData(gd);
 
             Composite buttonsPanel = UIUtils.createComposite(catPanel, 3);
-            
-                        
+
             fullSchemaBackupCheck = UIUtils.createCheckbox(buttonsPanel, PostgreMessages.wizard_backup_page_object_checkbox_complete_backup, false);
-            fullSchemaBackupCheck.addSelectionListener(SelectionListener.widgetSelectedAdapter(e -> {
-                    wizard.getSettings().setFullSchemaBackup(fullSchemaBackupCheck.getSelection());
-                    tablesTable.setVisible(!fullSchemaBackupCheck.getSelection());
-                }));
+            fullSchemaBackupCheck.addSelectionListener(SelectionListener.widgetSelectedAdapter(e -> updateState()));
             fullSchemaBackupCheck.setLayoutData(new GridData(GridData.GRAB_HORIZONTAL));
             buttonsPanel.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
             createCheckButtons(buttonsPanel, schemasTable);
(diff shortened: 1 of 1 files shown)


IvanGorshechnikov:
Unfortunately, it doesn't fix the regression described in https://github.com/dbeaver/dbeaver/issues/41683
We will close your PR and make our own fix

PR #42018, opened 2026-09-04
dbeaver/pro#10545 Fix doris meta model creation

E1izabeth:
Closes dbeaver/pro#10467
Closes dbeaver/pro#10545


E1izabeth:
@codacy review

IvanGorshechnikov:
@codex review

plugins/org.jkiss.dbeaver.registry/schema/org.jkiss.dbeaver.dataSourceProvider.exsd
```diff
@@ -107,6 +107,13 @@
                </appInfo>
             </annotation>
          </attribute>
+         <attribute name="defaultMetaModel" type="string">
```
E1izabeth, line 110: Introduced default metamodel per datasource registration attribute.
It is required because, otherwise, the metamodel registry fails to resolve it correctly and forces driver implementations to instantiate duplicate metamodel instances instead of reusing the one provided by the registry, which breaks metamodel lifecycle management. This creates a high risk of metamodel clashes because certain logic relies on the registry for metamodel resolution, but the driver might create an instance of a different type, and later, different code ends up encountering different metamodels, which is just wrong. We already have per-driver model registration for that. Why create its instance manually? This also solves the problem of resolving the correct metamodel for user-created drivers that don't inherit the bundled driver's meta-model parameter otherwise (and their functioning might be broken as a result). Explicit driver-level settings continue to take precedence, preserving old behavior for all the existing driver and datasource registrations.

E1izabeth:
@codex review

plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/GenericMetaModelRegistry.java
```diff
@@ -91,7 +91,16 @@ public GenericMetaModel getMetaModel(DBPDataSourceContainer dataSourceContainer)
         if (metaModelId != null && !GenericConstants.META_MODEL_STANDARD.equals(metaModelId)) {
             metaModel = metaModels.get(metaModelId.toString());
             if (metaModel == null) {
-                log.warn("Meta model '" + metaModelId + "' not recognized. Default one will be used");
+                log.warn("Driver meta model '" + metaModelId + "' not recognized. Default one will be used");
+            }
+        }
+        if (metaModel == null && metaModelId == null) {
```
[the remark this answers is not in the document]
E1izabeth: The only intention of defaultMetaModel is to generalize metamodel configuration for partially overriden implementations reusing the same driver but giving distinct semantics (like VeloDb vs Doris case), there is no goal to put any constraints on top of previously existing mechanisms, because it might unwantedly and unknowingly break existing custom user configuration.

E1izabeth:
@codex review

plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/meta/GenericMetaModel.java
```diff
@@ -952,6 +952,20 @@ public String generateOnUpdateFK(@NotNull DBSForeignKeyModifyRule updateRule) {
...
 
+    @NotNull
+    public JDBCStatement prepareIndexesLoadStatement(
+        @NotNull JDBCSession session,
+        @NotNull GenericStructContainer owner,
+        @Nullable GenericTableBase forParent
+    ) throws SQLException {
+        return session.getMetaData().getIndexInfo(
+            owner.getCatalog() == null ? null : owner.getCatalog().getName(),
+            owner.getSchema() == null || DBUtils.isVirtualObject(owner.getSchema()) ? null : owner.getSchema().getName(),
```
serge-rider, line 963: DBUtils.isVirtualObject call looks like a behavior change. Why check it on so low level?

plugins/org.jkiss.dbeaver.registry/src/org/jkiss/dbeaver/registry/DataSourceProviderDescriptor.java
```diff
@@ -300,6 +303,18 @@ public String getName() {
         return CommonUtils.toString(name, id);
     }
 
+    @Nullable
+    @Override
+    public String getDefaultMetaModelId() {
```
serge-rider, line 308: What makes Doris so special that we need it?

serge-rider (changes requested):

plugins/org.jkiss.dbeaver.ext.generic/src/org/jkiss/dbeaver/ext/generic/model/meta/GenericMetaModel.java
E1izabeth: This code just moved here from IndexCache for the whole code paths to be overridable the same way as it works for tables (see prepareTableLoadStatement).

E1izabeth:
@codex review

PR #42030, opened 2026-09-07
dbeaver/dbeaver#42027 Format nullable and parameterized ClickHouse JSON columns

### LINKED ISSUE #42027: ClickHouse: format nullable and parameterized JSON columns

Support for the JSON value viewer was added for bare ClickHouse `JSON` columns in #41315 / #41324. Nullable and parameterized JSON types were outside that change's scope.

For `Nullable(JSON)` and parameterized types such as `JSON(a UInt64, items Array(Nullable(UInt64)))`, values can already be read as JSON text, but the Value panel does not offer the JSON viewer and formatting available for bare `JSON`. Nullable variants open in a plain-text panel; parameterized JSON opens in the Text content viewer. Nested documents are difficult to read without formatting.

Extend the existing ClickHouse JSON value viewer to scalar JSON variants, in both the table Data editor and SQL result grids:

- `Nullable(JSON)`
- `JSON(...)`, including `JSON()`
- `Nullable(JSON(...))`, including `Nullable(JSON())`

These values should use the same JSON viewer and Auto Format option as bare `JSON`, without manual configuration. SQL NULL should remain NULL, with no stale document left in the Value panel when switching rows.

Reuse the existing ClickHouse JSON content handler and recognize these types consistently during value-handler selection and local type-cache lookup. I have a local implementation with regression tests and would be glad to submit a focused PR.

- Copying the returned text into an external JSON formatter — possible, but interrupts browsing data in DBeaver.
- Extending only the value-handler registration — insufficient for nullable types: local type lookup reduces `Nullable(JSON)` to `Nullable`, whose cached data kind is `UNKNOWN`, so the JSON content viewer is still not selected.

**Additional context**

- Follow-up to #41315 and #41324, not a regression of bare `JSON` support.
- Reproduced in an unpatched DBeaver CE `devel` build, version `26.2.1.202609061336`, commit `ae49ed3dcf90aa106d238ab1f4d5b7fb63443f98` (2026-09-06), with ClickHouse `26.9.1.762`, clickhouse-jdbc `0.10.0`, and Java 21 on macOS arm64.
- Minimal reproduction: run the query below, select the result cell, and open the Value panel. The value is returned as JSON text, but JSON formatting is unavailable.

```sql
SELECT CAST(
    '{"a":1,"items":[1,null,2],"meta":{"text":"Привет","flag":true}}'
    AS Nullable(JSON)
) AS nullable_json;
```

- On the tested server, `JSON()` and `Nullable(JSON())` are normalized in metadata to `JSON` and `Nullable(JSON)`. Consequently, standalone `JSON()` already formats when normalized; literal empty-parameter type names should also be recognized consistently.
- This request covers reading and JSON viewer/formatter selection. It does not request disabling the existing editor, adding INSERT support, or fixing pre-existing UPDATE/binding issues. Top-level `Array(JSON)`, `Array(Nullable(JSON))`, Map and other collection types are outside scope.
- Current behavior: Nullable(JSON) returns JSON text, but the Value panel does not provide the JSON viewer or formatting controls.

Image

bakhtiiartashbolotov:
Closes dbeaver/dbeaver#42027.

Follow-up to [#41315](https://github.com/dbeaver/dbeaver/issues/41315) and
[#41324](https://github.com/dbeaver/dbeaver/pull/41324), which introduced the JSON content viewer for bare `JSON` columns.

## What

Extend the existing JSON viewer routing to scalar JSON variants in both the table Data editor and SQL result grids:

- `Nullable(JSON)`
- `JSON(...)`, including `JSON()`
- `Nullable(JSON(...))`, including `Nullable(JSON())`

Existing bare `JSON` support is retained.

## Why

Nullable JSON values can already be returned as JSON text but still open in a plain-text panel without JSON formatting.
The local type lookup strips `Nullable(JSON)` to `Nullable`, whose cached data kind is `UNKNOWN`.
Since `JDBCColumnMetaData` accepts that cached kind, the `resolveDataKind` fallback is not reached.
Parameterized JSON names also need to be recognized by the value-handler registration and routing.

## How

- Use one `isJsonType` predicate for handler selection, data-kind resolution and local type lookup. It handles nested parentheses and quoted type parameters without changing the existing ANTLR grammar.
- Resolve recognized scalar JSON variants through the existing cached `JSON` entry (`CLOB` / `CONTENT`) before the usual name truncation. Keep `getTypeNameWithoutModifiers()` and the lookup path for other types unchanged.
- Extend the ClickHouse-specific type registration using the existing wildcard mechanism.
- Handle the observed JDBC metadata spelling that omits the outer closing parenthesis of parameterized `Nullable(JSON(...))`, without rewriting the original metadata name or nullability.

The existing JSON value handler, content class, serialization settings and binding implementation are unchanged. The predicate classifies type metadata; it is not a JSON document validator or a full ClickHouse type-grammar parser.

## Scope

This PR covers reading and JSON viewer/formatter selection. It reuses the existing editable content handler and does not disable editing.
INSERT support and fixes to pre-existing UPDATE/binding issues are outside this change; the validation below does not establish lossless write support for every JSON type hint.
Top-level Array, Map, Tuple, Dynamic and LowCardinality types are not added to scalar JSON routing.

## Tested

- Ran the unfiltered aggregate reactor with Java 21 (`mvn verify -f product/aggregate/pom.xml -Dtycho.localArtifacts=ignore`): BUILD SUCCESS, 1,212 reported cases, 1,208 passed, 4 skipped, 0 failures/errors across 21 test-module summaries (20 DBeaver test bundles plus `dbeaver-common` utils). This run used the default headless test configuration, without the CE/Eclipse product profiles; the separate native CE product builds are described below.
- Rechecked the unchanged patch against upstream `devel` commit `ae49ed3dcf90aa106d238ab1f4d5b7fb63443f98` (2026-09-06): CE product build succeeded and all 278 ClickHouse tests passed. On the same commit without the production patch, the added regression suite produced 155 assertion failures (0 errors); the original 25 tests passed. The upstream SHA was verified again after building.
- Manually compared clean `26.2.1.202609061336` and patched `26.2.1.202609061339` builds of this commit. The clean build reproduces the nullable/parameterized viewer issue in both SQL result grids and the table **Data** tab; the patched build selects the JSON viewer and shows formatted documents in both paths. Bare JSON and the `Array(JSON)` collection viewer retain their existing behavior. A fresh post-UI run again passed all 278 ClickHouse tests.
- **278 ClickHouse test cases passed** across all seven ClickHouse test classes, also confirmed in the unfiltered reactor run above.
- Coverage includes supported and malformed type names, quoted/nested parameters, extension registration, real populated/empty type-cache lookup through `JDBCColumnMetaData`, preservation of metadata names/nullability, SQL NULL, read failures and unchanged collection-handler selection. The new cache regression cases failed for 12 nullable variants before the lookup fix.
- Environment: ClickHouse `26.9.1.762`, JDBC `0.10.0` and Java 21 on macOS arm64. Successful document → SQL NULL → document transitions were observed in the patched table grid; table data was unchanged after both builds' UI checks (byte-for-byte JSONEachRow comparison).
- The contributor confirmed the final manual checks on 2026-09-07 in patched build `26.2.1.202609061339`: Auto Format off/on (switching rows after disabling it), clearing the preview on SQL NULL, and restoring the correct document when returning from NULL to the first/third `nullable_quoted` row. These are direct manual results, not an automated UI-test pass.

The server normalizes `JSON()` / `Nullable(JSON())` in table metadata; literal empty-parameter metadata names are covered separately by routing tests. Manual validation covers reading and formatting, not saving edits.

Implementation and tests were developed with AI assistance (Codex).

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/ClickhouseTypeParser.java
+++ b/plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/ClickhouseTypeParser.java
@@ -1,6 +1,6 @@
 /*
  * DBeaver - Universal Database Manager
- * Copyright (C) 2010-2024 DBeaver Corp and others
+ * Copyright (C) 2010-2026 DBeaver Corp and others
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -62,6 +62,47 @@ public class ClickhouseTypeParser {
                 typeName.startsWith(ClickhouseConstants.DATA_TYPE_TUPLE) ||
                 typeName.startsWith(ClickhouseConstants.DATA_TYPE_ARRAY)
             );
+    }
+
+    /** Recognizes scalar JSON metadata, including type hints and an optional Nullable wrapper. */
+    public static boolean isJsonType(@NotNull String typeName) {
+        if (ClickhouseConstants.DATA_TYPE_JSON.equalsIgnoreCase(typeName)
+            || ClickhouseConstants.DATA_TYPE_NULLABLE_JSON.equalsIgnoreCase(typeName)
+        ) {
+            return true;
+        }
+
+        String nullablePrefix = "Nullable(";
+        boolean nullable = typeName.regionMatches(true, 0, nullablePrefix, 0, nullablePrefix.length());
+        int start = nullable ? nullablePrefix.length() : 0;
+        String jsonPrefix = ClickhouseConstants.DATA_TYPE_JSON + "(";
+        if (!typeName.regionMatches(true, start, jsonPrefix, 0, jsonPrefix.length())) {
+            return false;
+        }
+
+        // Type hints can contain nested types, quoted paths, enum labels and regular expressions.
+        int depth = 1;
+        char quote = 0;
+        for (int i = start + jsonPrefix.length(); i < typeName.length(); i++) {
+            char c = typeName.charAt(i);
+            if (quote != 0) {
+                if (c == '\\') {
+                    i++;
+                } else if (c == quote) {
+                    quote = 0;
+                }
+            } else if (c == '\'' || c == '"' || c == '`') {
+                quote = c;
+            } else if (c == '(') {
+                depth++;
+            } else if (c == ')' && --depth == 0) {
+                // JDBC v2 can omit the outer Nullable close in getColumnTypeName().
+                // Accept that metadata spelling without changing the original type name.
+                return i == typeName.length() - 1
+                    || (nullable && i == typeName.length() - 2 && typeName.charAt(i + 1) == ')');
+            }
+        }
+        return false;
     }
 
     @Nullable
--- a/plugins/org.jkiss.dbeaver.ext.clickhouse/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.clickhouse/plugin.xml
@@ -165,6 +165,9 @@
             <type name="TUPLE*"/>
             <type name="MAP*"/>
             <type name="JSON"/>
+            <type name="JSON(*"/>
+            <type name="NULLABLE(JSON)"/>
+            <type name="NULLABLE(JSON(*"/>
 
             <!-- https://clickhouse.com/docs/sql-reference/data-types/geo -->
             <type name="POINT"/>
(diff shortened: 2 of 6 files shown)


plugins/org.jkiss.dbeaver.ext.clickhouse/src/org/jkiss/dbeaver/ext/clickhouse/ClickhouseTypeParser.java
```diff
@@ -64,6 +64,47 @@ public static boolean isComplexType(@NotNull String typeName) {
             );
     }
 
+    /** Recognizes scalar JSON metadata, including type hints and an optional Nullable wrapper. */
+    public static boolean isJsonType(@NotNull String typeName) {
```
ShadelessFox, line 68: This class wraps an ANTL4 grammar for parsing ClickHouse types. Extend the grammar if necessary, and use it instead of rolling your own parser.

ShadelessFox (changes requested):

PR #42036, opened 2026-09-07
dbeaver/dbeaver#34950 Move Refresh Materialized View to first-level menu

### LINKED ISSUE #34950: A GUI button for refreshing materialized view in postgresql
I can't find a button to refresh materialized view in postgresql database. The context menu contains a "refresh" button but that doesn't work. 

Add an option in the context menu, toolbar or top menu dedicated for updating the materialized view. 

**Additional context**
Dbeaver 24.0.2
Windows 10

akmalsyrf:
## Summary

Closes dbeaver/dbeaver#34950

Moves PostgreSQL **Refresh Materialized View** from **Tools → Maintenance** to the first level of the Database Navigator context menu (after **Filter**), as discussed with the team in the issue.

### Motivation

The action already existed under Tools, but users often missed it and expected a top-level "refresh materialized view" entry (distinct from the ordinary navigator Refresh, which only reloads metadata).

### Changes

- Removed the Tools contribution `PostgreToolRefreshMView` from the Maintenance tool group in `org.jkiss.dbeaver.ext.postgresql.ui/plugin.xml`.
- Added Eclipse command `org.jkiss.dbeaver.ext.postgresql.tools.refreshMView` with an enabled-when condition for `PostgreMaterializedView` (adapt + instanceof pattern to avoid forced plugin activation).
- Added menu contribution `popup:org.eclipse.ui.popup.any?after=navigator.filter` so the action appears at the root context menu after Filter.
- New handler `PostgreToolRefreshMViewHandler` opens the existing task wizard via `TaskConfigurationWizardDialog.openNewToolTaskDialog(..., PostgreSQLTasks.TASK_MVIEW_REFRESH, ...)` — same execution path as before.
- Task definition `pgToolRefreshMView` / `PostgreToolMViewRefresh` unchanged.

### Notes

This intentionally does not change the dialog defaults (e.g. WITH DATA). Follow-up feedback in the issue about defaulting WITH DATA can be handled separately.

## Test plan

- [x] Code review: menu URI `after=navigator.filter` matches `navigator.filter` menu id in `org.jkiss.dbeaver.core`.
- [x] Code review: Tools contribution removed; task `pgToolRefreshMView` still registered; handler opens the same task id.
- [x] Code review: `enabledWhen` requires one-or-more selection of `PostgreMaterializedView` only (adapt + instanceof).
- [ ] Connect to PostgreSQL, create or select a materialized view in Database Navigator.
- [ ] Right-click the materialized view and verify **Refresh Materialized View** appears at the first menu level after **Filter** (not only under Tools).
- [ ] Run the action and confirm the existing refresh wizard/dialog still opens and executes `REFRESH MATERIALIZED VIEW` correctly (WITH DATA / WITH NO DATA options).
- [ ] Verify the action is **not** visible for ordinary tables/views.
- [ ] Confirm the action no longer appears under Tools → Maintenance (avoid duplicate entries).
- [ ] Multi-select multiple materialized views and verify the handler still opens the tool task dialog.

This PR was generated with AI (Cursor).

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql.ui/plugin.xml
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql.ui/plugin.xml
@@ -257,14 +257,6 @@
                     singleton="false">
                 <task id="pgToolTriggerEnable"/>
             </tool>
-            <tool
-                    description="%tools.refesh.mview.description"
-                    id="org.jkiss.dbeaver.ext.postgresql.tools.maintenance.PostgreToolRefreshMView"
-                    group="org.jkiss.dbeaver.ext.postgresql.tools.maintenance"
-                    label="%tools.refesh.mview.name"
-                    singleton="false">
-                <task id="pgToolRefreshMView"/>
-            </tool>
 
             <toolGroup
                 id="org.jkiss.dbeaver.ext.postgresql.tools.config"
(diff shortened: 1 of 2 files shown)


serge-rider:
Why adding new command and menu if we already have it?
image

Generally we don't want to pollute main menu, it is already too big. And MView refresh is not a task you do thousand times per day.

bayusamsudin:
Beware, this is classic vibecoding @akmalsyrf  🤖💀 Overclaimed issue, slopped changes everywhere, and somehow the solution is to pollute the main menu even more.

Closing this PR is the right choice @serge-rider 🚫

PR #42045, opened 2026-09-08
dbeaver/dbeaver#41993 Drain password command streams before waiting

### LINKED ISSUE #41993: Password from shell command hangs when command outputs a large value on Windows
## Description

DBeaver's Password from shell command authentication appears to hang when the command produces sufficiently large output on stdout.

This was discovered while trying to authenticate to Azure Database for PostgreSQL using a Microsoft Entra access token. The token is approximately 6 KB, and DBeaver consistently waits until the configured password-command timeout and then reports:

```
Authentication error: Password command exited with code -1
  Password command exited with code -1
```

I was able to reproduce the problem without Azure, PostgreSQL, or an access token.

Environment
OS: Windows
DBeaver: Community Edition
Database: PostgreSQL
Authentication: Password from shell command
Password command timeout: tested with 10 seconds and 60 seconds

Minimal reproduction

Create a batch file containing:

```
@echo off
powershell.exe -NoProfile -Command "[Console]::Write('A' * 6200)"
```

Configure DBeaver's PostgreSQL connection to use this BAT file as the Password from shell command.

DBeaver waits until the configured timeout and then reports:

```
Authentication error: Password command exited with code -1
  Password command exited with code -1
```

In contrast, a short output works immediately.

For example:

```
@echo off
echo abc
```

This immediately results in PostgreSQL rejecting abc as an invalid password/token:

FATAL: The access token has invalid format. Try again after making sure that the token is passed exactly as returned by the authentication endpoint.

This demonstrates that the command is being executed and its stdout is being passed to the database connection.

## Original use case

The original command was:

```
@echo off
"C:\Program Files\Microsoft SDKs\Azure\CLI2\wbin\az.cmd" account get-access-token --resource https://ossrdbms-aad.database.windows.net --query accessToken -o tsv
```

Running this BAT directly from PowerShell returns a 6,019-character access token in approximately one second.

Using the same BAT as DBeaver's password command causes DBeaver to wait until the password-command timeout.

I also confirmed that the token is actually generated successfully by redirecting Azure CLI output to a file. The complete 6,019-character token is present in the file.

Additional observations

DBeaver's environment contains the correct Azure CLI path:

```
PATH=C:\Program Files\Microsoft SDKs\Azure\CLI2\wbin;C:\Windows\system32;...
```

az.cmd can therefore be located by the process.

The failure is reproducible with arbitrary output and does not require Azure CLI. For example, 6,200 A characters trigger the same timeout.

### Expected behaviour

DBeaver should consume the first line of stdout from the password command and pass it to the database connection, regardless of the password length, within the configured command timeout.

### Actual behaviour

For sufficiently large stdout output, DBeaver waits until the password-command timeout and reports:

Password command exited with code -1

### DBeaver Version

Community 26.1.5.202608161900

### Operating System

Windows 11

### Database and driver

psql (PostgreSQL) 17.0
org.postgresql:postgresql:RELEASE [42.7.13]

### Steps to reproduce

```
# This fails because the token is incorrect
@echo off
echo abc
```

```
# This hangs out
@echo off
powershell.exe -NoProfile -Command "[Console]::Write('A' * 6200)"
```

If you try and reduce the length of the string, DBeaver will stop hanging and start failing again.

### Additional context

_No response_

dajiaohuang:
Closes dbeaver/dbeaver#41993

## Summary
- drain stdout and stderr concurrently while the password shell command runs
- preserve first-line extraction, timeout handling, and process cleanup semantics
- add a regression test for 6,200-character command output

## Testing
- `mvn -pl plugins/org.jkiss.dbeaver.model -DskipTests package` (BUILD SUCCESS)
- `git diff origin/devel...HEAD --check` (passed)
- test plugin reactor is blocked locally because `org.jkiss.dbeaver.headless` and `org.jkiss.dbeaver.model.rcp` are absent from the target platform

This PR was generated with AI (Codex).


dajiaohuang:
Follow-up: addressed the Copilot findings in the current PR head (`acece8269a9e4040a16a2e55b1b32fd15c54e1a8`). Stream reads now convert interruption/cancellation/reader failures to `DBException`, and the reader no longer closes the process-owned pipes. The focused model package build passes; the platform-test reactor remains locally blocked by missing `org.jkiss.dbeaver.headless` / `org.jkiss.dbeaver.model.rcp` bundles.

PR #42050, opened 2026-09-08
dbeaver/pro#10620 ResultSetViewer performance fix

serge-rider:


Matvey16:
Verified

DBeaver CE Windows

PR #42054, opened 2026-09-09
dbeaver/pro#10278 Use registered transfer formats in AI

IvanGorshechnikov:
Closes dbeaver/pro#10278

## Summary
- add extension-backed dynamic valid values for AI function parameters
- expose stable AI format identifiers from data transfer processors
- resolve AI export formats by explicit identifier while preserving legacy fallbacks

## Testing
- `mvn package -f product/aggregate/pom.xml -pl :org.jkiss.dbeaver.model.ai,:org.jkiss.dbeaver.data.transfer,:org.jkiss.dbeaver.data.transfer.ui -am -DskipTests`
- `git diff --check`


IvanGorshechnikov:
Replaced by #42056 so the CE and EE dependency branches use the same name.

PR #42072, opened 2026-09-10
dbeaver/pro#10649 Fix AI chat background blinking

serge-rider:
Closes dbeaver/pro#10649

Declares the active DBeaver color scheme in the chat HTML before external styles load, preventing the WebView's initial white paint in dark theme. The same generated theme value is now applied to both HTML and CSS.

Testing:
- `git diff --check`
- Targeted Maven build attempted; dependency resolution stopped before compilation because `org.jkiss.dbeaver.model.jdbc` was unavailable in the partial reactor.


serge-rider:
@codex review

PR #42091, opened 2026-09-11
dbeaver/dbeaver#11061 Fix PostgreSQL PUBLIC meta-role missing from Permissions tabs

### LINKED ISSUE #11061: PUBLIC meta role missing in Permissions
#### System information: 
- all
- DBeaver version: 7.3.2
- N/A

- postgresql v10, v12
- Driver name: PostgreSQL JDBC Driver
- No

Permission tabs contain no PUBLIC meta role. It's has key value for security settings - therefore missing it I consider as a bug of permission tabs

PUBLIC metarole gives particular grants to ANYONE connected. 
For example every function gets by default GRANT USAGE for PUBLIC therefore can be executed by anyone who is connected and has access to the schema the function is located in. This grant can be revoked of course.

In contrary schemas and tables have no public role granted by default. But it can be granted later.
Schema public has USAGE granted to PUBLIC meta role given by default (after creation of database)

PUBLIC role should be visible always. It does exists since postgres cluster creation.

Be aware however PUBLIC  meta role  doesn't exist in pg_authid nor pg_roles. But it appears in information_schema tables. 
It is manifested by 0 (zero) value in ACL object.
For example:
```
SELECT nspname, grants.*, grants.grantee::regrole::VARCHAR, pg_get_userbyid(grants.grantee)
FROM pg_catalog.pg_namespace 
JOIN aclexplode(nspacl) AS grants ON true
WHERE nspname = 'public'
```
Look how it behaves when resolving its id (zero) different way.

It's also important, PUBLIC in sql queries is a keyword not object name. Therefore cannot be escaped. 
While it works:
GRANT SELECT ON TABLE mytab TO "myrole";
it won't:
GRANT SELECT ON TABLE mytab TO "PUBLIC";
It must be written without doublequotes always:
GRANT SELECT ON TABLE mytab TO PUBLIC;

1. create new schema (to make sure there are no default privilege rules set for the schema)
2. create new function in this schema
3. check the function permission tab
4. The is no PUBLIC role

N/A

NikRom5531:
Closes dbeaver/dbeaver#11061

### Problem

The PostgreSQL PUBLIC pseudo-role does not appear in the Permissions tab for any database object (functions, schemas, tables, etc.), despite being a critical part of PostgreSQL's access control model. Every database connection includes PUBLIC implicitly, and PostgreSQL assigns default privileges to PUBLIC (e.g. `EXECUTE` on all functions, `USAGE` on the `public` schema). The inability to see or manage these grants through DBeaver's UI is a security concern.

PUBLIC does not exist in `pg_roles` or `pg_authid` — it is a virtual role with OID 0, represented as the string `'PUBLIC'` in `information_schema` tables and as the integer 0 in ACL objects.

### Solution

Synthesize the PUBLIC pseudo-role in the PostgreSQL role cache so it participates in the Permissions tab like any other role.

**Changes across 5 files:**

1. **`PostgreConstants.java`** — Add `PUBLIC_ROLE_NAME = "public"` constant.

2. **`PostgreDatabase.java`** — Override `addCustomObjects()` in `RoleCache` to inject a synthetic `PostgreRole` for PUBLIC (oid=0, canLogin=false, persisted=true). Add case-insensitive fallback in `getRoleByReference()` so that both `"public"` and `"PUBLIC"` resolve to the same cached role.

3. **`PostgreRole.java`** — Add `isPublicRole()` helper. In `getPrivileges()`, use uppercase `'PUBLIC'` as the grantee filter in `information_schema` queries (matching PostgreSQL's casing). Guard `getObjectDefinitionText()` to return a descriptive comment instead of invalid `CREATE ROLE "public"` DDL. Fix `otherObjectsSQL` to resolve the PUBLIC pseudo-role by OID 0 instead of `pg_get_userbyid()` (which returns `'unknown (OID=0)'` instead of `'PUBLIC'` on modern PostgreSQL). Fix default privileges query: add missing space in SQL text block concatenation (`where` + filter) that caused a syntax error at runtime (`wherepg_get_userbyid`).

4. **`PostgreCommandGrantPrivilege.java`** — Add `getGranteeName()` static helper that emits the unquoted keyword `PUBLIC` instead of the quoted identifier `"PUBLIC"` (which PostgreSQL rejects). Used in both the `PostgreRole` and `PostgreObjectPrivilege` code paths for DDL generation.

5. **`PostgreUtils.java`** — Fix NULL ACL handling. When an object's ACL is NULL (i.e. it has never been explicitly granted), PostgreSQL applies its default access privileges. For procedures/functions the default grants `EXECUTE` to PUBLIC, but DBeaver only synthesized the owner grant. Now the PUBLIC `EXECUTE` grant is also synthesized, so default function permissions match what `information_schema.routine_privileges` reports.

### Technical notes

- PUBLIC is a SQL keyword — `GRANT ... TO "PUBLIC"` is invalid; it must be `GRANT ... TO PUBLIC`. The fix ensures DDL generation never quotes it.
- `information_schema` reports PUBLIC in uppercase (`'PUBLIC'`), while `DBUtils.getQuotedIdentifier` produces lowercase `"public"` for cache lookup. The case-insensitive fallback in `getRoleByReference()` bridges this gap.
- The `addCustomObjects` callback in `JDBCObjectCache` runs after the database fetch from `pg_roles`, so the synthetic role is added regardless of query results.
- A NULL ACL (no explicit GRANT/REVOKE ever issued) means "default privileges" in PostgreSQL. `acldefault()` grants PUBLIC `EXECUTE` on functions/procedures and `USAGE` on types/languages by default, and this must be reflected when DBeaver loads privileges from likely-NULL ACL columns.

### Testing

Key checks:
- PUBLIC row visible in Permissions tab for functions, schemas, tables
- Granting/revoking privileges to/from PUBLIC via UI succeeds
- DDL for PUBLIC shows a comment, not invalid `CREATE ROLE`
- Generated `GRANT ... TO PUBLIC` uses unquoted keyword

### DIFF
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreConstants.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreConstants.java
@@ -41,6 +41,12 @@ public class PostgreConstants {
     public static final String DEFAULT_DATA_TYPE = "varchar";
     public static final String DEFAULT_USER = "postgres";
     public static final String USER_VARIABLE = "$user";
+    /**
+     * Name of the PUBLIC pseudo-role which exists in every database since cluster initialization.
+     * PUBLIC is not stored in pg_authid/pg_roles. It is manifested by 0 (zero) value in ACL objects
+     * and 'PUBLIC' string in information_schema views.
+     */
+    public static final String PUBLIC_ROLE_NAME = "public";
 
     public static final String PROP_CHOSEN_ROLE = DBConstants.INTERNAL_PROP_PREFIX + "chosen-role@";
     public static final String PROP_SHOW_NON_DEFAULT_DB = DBConstants.INTERNAL_PROP_PREFIX + "show-non-default-db@";
--- a/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
+++ b/plugins/org.jkiss.dbeaver.ext.postgresql/src/org/jkiss/dbeaver/ext/postgresql/PostgreUtils.java
@@ -742,7 +742,9 @@ public class PostgreUtils {
     ) throws DBException {
         if (!(acl instanceof java.sql.Array)) {
             if (acl == null) {
-                // Special case. Means ALL permissions are granted to table owner
+                // Special case. NULL ACL means the object has default privileges:
+                // all privileges are granted to the object owner plus the default PUBLIC
+                // privileges for the specific object type (e.g. EXECUTE for procedures).
                 PostgreRole objectOwner = owner.getOwner(monitor);
                 PostgreRoleReference granteeReference = objectOwner == null ? null : objectOwner.getRoleReference();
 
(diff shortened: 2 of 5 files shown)


E1izabeth:

serge-rider:
@codex review

E1izabeth:
@codex review

PR #42095, opened 2026-09-12
dbeaver/pro#9786 Fix Firestore data transfer

E1izabeth:
Closes dbeaver/pro#9786

Adds a transient object command context for database objects that must be used before metadata materialization. Dynamic document targets can now be bound directly and skip premature metadata refreshes.

Also avoids forced finish-time remapping for direct targets while retaining the normal mapping transition behavior.

Counterpart: dbeaver/dbeaver-ee#4835


E1izabeth:
@codex review

PR #42120, opened 2026-09-15
dbeaver/pro#9579 Add DataDam Access Point network handler

vkirst:
Summary: adds the DataDam Access Point tunnel network handler and settings UI, requests signed single-use bridge tickets, and rewrites datasource connections to a local forwarding endpoint. Related changes: DataDam backend https://github.com/dbeaver/datadam/pull/85 and IDEA workspace https://github.com/dbeaver/idea-workspace-pro/pull/200. Testing was not run as part of PR creation. This PR was created with AI assistance using OpenCode. Closes dbeaver/pro#9579


alexander-skoblikov (changes requested):
It looks like we shouldn't merge this

PR #42128, opened 2026-09-15
dbeaver/dbeaver#41858 Fix model listing for custom OpenAI endpoints

### LINKED ISSUE #41858: "Refresh models" returns no models for self-hosted OpenAI-compatible endpoints
### Description

"Refresh models" returns nothing for a self-hosted OpenAI-compatible endpoint (vLLM, Ollama, LM Studio, ...). The model list request succeeds, the server returns its models, but the dropdown stays empty and the notification reads "The engine returned no models." No error is shown and nothing is written to the log, so it looks like a connectivity problem when it is not.

The cause is model-name filtering. `OpenAIModels.isChatModel` only recognises a model as chat-capable if its id starts with `gpt-` or `o`:

```java
private static boolean isChatModel(@NotNull String modelName) {
    if (!(modelName.startsWith("gpt-") || modelName.startsWith("o"))) {
        return false;
    }
    ...
}
```

Any other id gets an empty feature set from `detectModelFeatures`, and `OpenAiConfigurator` then drops it because it has no `CHAT` feature:

```java
.filter(it -> it.features().contains(AIModelFeature.CHAT))
```

Self-hosted servers normally report the HuggingFace model path, which never starts with `gpt-` or `o`, so every model is filtered out and the user is left with an empty dropdown.

The chain is:

1. `OpenAiClientBase.getModels` - GET `{baseUrl}/models`, parses `data[]` (succeeds)
2. `OpenAIEngine.getModels` - maps each id through `KNOWN_MODELS.getOrDefault(id, new AIModel(id, null, detectModelFeatures(id)))`
3. `OpenAIModels.detectModelFeatures` -> `isChatModel` returns false -> empty feature set
4. `OpenAiConfigurator` filters on `AIModelFeature.CHAT` -> empty list
5. `ModelSelectorField` reports "The engine returned no models."

For comparison, `DDAIEngine.getModels` talks to the same OpenAI-compatible client and trusts the server's list instead of guessing from the name:

```java
.map(model -> new AIModel(model.id(), null, Set.of(AIModelFeature.CHAT, AIModelFeature.STREAMING)))
```

Workaround: the model combo is editable, so the model id can be typed in by hand and it is used correctly. Only the dropdown is affected.

### DBeaver Version

26.2.0 (build from `devel`, 202608201258)

### Operating System

Windows

### Database and driver

Not database-specific. AI engine: OpenAI (OpenAI-compatible), endpoint served by vLLM.

### Steps to reproduce

1. Serve any non-OpenAI model through an OpenAI-compatible server, e.g. vLLM with `nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-FP8`
2. Open Preferences -> AI, choose the OpenAI engine
3. Set the base URL to the local endpoint, e.g. `http://localhost:8000/v1/`, and enter any non-empty token (the field is required, the value is unused by vLLM)
4. Click "Refresh models"
5. The notification reads "The engine returned no models." and the dropdown stays empty

### Additional context

`GET /v1/models` on the same endpoint returns the model normally:

```json
{"object":"list","data":[{"id":"nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-FP8","object":"model","created":1787247052,"owned_by":"vllm","root":"nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-FP8","parent":null,"max_model_len":262144}]}
```

`isChatModel` was introduced in c586dcfebe ("dbeaver/pro#6610 Fix filtering of OpenAI models", 2025-08-21), so this affects every version with the dynamic model list.

IvanGorshechnikov:
Closes dbeaver/dbeaver#41858

## Summary
- Apply OpenAI model classification only to the default API endpoint, including empty URLs and URLs without a trailing slash.
- Treat models returned by custom OpenAI-compatible endpoints as chat and streaming models so they remain available in the model selector.
- Add regression coverage for self-hosted model IDs, names matching the OpenAI catalog, and default endpoint classification.

## Verification
- Built the AI plugin and ran OpenAIEngineTest and OpenAIModelsTest with Java 21: 10 tests passed.
- `git diff --check` passed.

```sh
mvn verify -f product/aggregate/pom.xml -pl :org.jkiss.dbeaver.model.ai,:org.jkiss.dbeaver.model.ai.test -am -Dtest=OpenAIEngineTest,OpenAIModelsTest -DfailIfNoTests=false -B
```

AI assistance: implemented and tested with OpenCode.


test/org.jkiss.dbeaver.model.ai.test/src/org/jkiss/dbeaver/model/ai/engine/openai/OpenAIEngineTest.java
```diff
@@ -0,0 +1,91 @@
...
+public class OpenAIEngineTest extends DBeaverUnitTest {
+    private static final List<String> MODEL_NAMES = List.of(
+        "nvidia/Nemotron-3-Nano-Omni-30B-A3B-Reasoning-FP8",
+        "llama3.2:latest",
+        "gpt-4o",
+        "gpt-4o-audio-preview",
+        "text-embedding-3-small"
+    );
+
+    @ParameterizedTest
```
serge-rider: IT looks like integration test, not unit test. Also I don't see this test run in build logs.

serge-rider (changes requested):

test/org.jkiss.dbeaver.model.ai.test/src/org/jkiss/dbeaver/model/ai/engine/openai/OpenAIEngineTest.java
serge-rider: In fact I cannot understand what this test tests.

PR #42147, opened 2026-09-17
dbeaver/dbeaver#13762 Fix function search in trigger create dialog

### LINKED ISSUE #13762: Function search doesn't work in trigger create dialog
#### System information: 
- Operating system (distribution) and version Mac
- DBeaver version 21.2 EE EA

1. Open Triggers tab in table Properties, open create trigger dialog, enter name and open dropdown to select a function
2. Type a function name in the search field

**Expected:** matched functions are displayed only
**Actual result:** no visible changes

https://user-images.githubusercontent.com/51405061/131799164-48ca83b3-322d-4e98-9934-d0435fdabf7b.mov

NikRom5531:
Closes #13762

## What was the problem?

The function search field in the "Create Trigger" dialog did not filter the list of functions when the user typed a name. Typing in the search field produced no visible changes - all functions remained visible.

## Root cause

The object browser dialog used to select a function for a trigger always forced the navigator tree search filter to the `connection` scope (`DatabaseNavigatorTreeFilterObjectType.connection`). This filter type only matches `DBPDataSourceContainer` (data source connection) objects against the search pattern. Since the function picker is rooted at a concrete database node and never displays data source containers, the typed pattern was never applied to database objects such as procedures/functions, so nothing was ever filtered out.

## What was changed

- `ObjectBrowserDialogBase` now keeps the filter object type configurable (default `connection` to preserve existing behavior) and uses it when building the navigator tree.
- `ObjectBrowserDialog` selects the `table` (database objects) filter scope when the dialog is rooted at a database node, so database objects like procedures/functions are matched against the typed search pattern. The `connection` scope is still used when the dialog browses data sources/connections.

## Affected functionality

- Function search in the PostgreSQL trigger create dialog now works as expected - typing filters the displayed functions.
- The same fix also benefits other object browser dialogs rooted at a database node (e.g. function selection in the debug configuration dialog and table selection when adding entities to an ER diagram).
- Object browser dialogs rooted at project/databases level (e.g. data transfer source selection) keep their previous behavior.

---

https://github.com/user-attachments/assets/47b40580-2158-46cc-bcaa-bd8317f30b3f


E1izabeth:

E1izabeth:
@codex review

Issue #41674, opened 2026-08-04
Unable to open previously stored data imprt tasks after update


JSKo13:
### Description

After update of dBeaver to version 26.1.4.202608021747 all the stored import tasks are missing target table and mapping of columns from csv to table. Even after updating the task, saving and running the updated target table and mapping is not stored.

<img width="682" height="361" alt="Image" src="https://github.com/user-attachments/assets/f9278088-f29c-4147-bd68-81cd0947d7a0" />

### DBeaver Version

Community Edition 26.1.4.202608021747

### Operating System

Windows 11 Pro 25H2

### Database and driver

Database:MySQL 8.4.9
Driver: com.mysql:mysql-connector-j:RELEASE [8.2.0]

### Steps to reproduce

1. Create data import task (csv>table) in previous dBeaver version
2. Update dBeaver
3. Try to run the data import task created before

JSKo13:
The issue is also present when creating new data import task. The target table and mapping is not stored.

E1izabeth:
Thanks for the bug report. I reproduced the issue

ShadelessFox:
Regression was introduced in https://github.com/dbeaver/dbeaver/commit/e0d43b9ec3e725f635930cba9f1a92b8d7ad46bf

mariiachugunova:
Verified for DBeaver CE on Windows 11 pro

PR #42148, opened 2026-09-17
dbeaver/pro#9896 Defer export target connection until mapping

IvanGorshechnikov:
Closes dbeaver/pro#9896

## Summary
- Defer loading the last database target for a new export until the mapping page, where Auto reconnect and explicit connection are handled.
- Load settings on demand when opening Configure columns.
- Track loaded producer/consumer settings independently and reinitialize pipes after format changes, preserving user edits when switching back.
- Add regression coverage for saved import targets and column mappings, including the failure reported in dbeaver/dbeaver#41674.

## Testing
- Built the Data Transfer model and UI bundles and ran the full Data Transfer test bundle: 52 tests passed, including 6 new regression scenarios.
- Confirmed the import regression tests fail when the reverted Auto reconnect condition is restored, and the format-switching test fails with the previous loading flag.
- Manual verification confirmed by the requester.
- `git diff --check` passed.

```sh
mvn -q verify -f product/aggregate/pom.xml -T1C \
  -pl :org.jkiss.dbeaver.data.transfer,:org.jkiss.dbeaver.data.transfer.ui,:org.jkiss.dbeaver.data.transfer.test -am
```

AI assistance: implementation and regression tests were prepared with OpenCode.


serge-rider:
@codex review

PR #42155, opened 2026-09-18
fix(ai): parse SSE data:/event: lines without requiring a space

### LINKED ISSUE #42096: OpenAI Responses API streaming: SSE parser requires a space after `data:`/`event:`, silently dropping events from spec-compliant servers
### Background

When using the built-in AI Chat (CE 26.2) with an OpenAI-compatible provider that implements the Responses API (e.g. Alibaba Cloud DashScope `https://dashscope.aliyuncs.com/compatible-mode/v1/`), prompts are sent and the HTTP request succeeds (200, SSE stream flowing), but no response is ever rendered and the token counter stays at 0. No error is shown.

### Root cause

In `org.jkiss.dbeaver.model.ai.engine.openai.OpenAiAPIStreamConsumer`:

```java
public static final String DATA_EVENT = "data: ";   // note the trailing space
private static final String EVENT_EVENT = "event: "; // note the trailing space
...
if (event.startsWith(DATA_EVENT)) { ... } else if (event.startsWith(EVENT_EVENT)) { ... } else { log.debug("Unknown OpenAI event: " + event); }
```

The parser only recognizes lines that start with `"data: "` / `"event: "` (colon **plus** a space).

Per the WHATWG SSE spec (https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation), after parsing the field name, "If the field value starts with a single U+0020 SPACE character, remove it from value." — the space is **optional**. `data:{json}` is perfectly valid SSE and is emitted by several major OpenAI-compatible providers (DashScope among them, verified byte-for-byte with `curl | od -c`).

Because such lines match neither prefix, every chunk falls into the `else` branch (logged only at debug level) and is silently discarded. From the user's perspective the chat just hangs with 0 tokens, which looks like a connectivity problem rather than a parsing one.

### Suggested fix

Parse SSE lines per spec instead of prefix-matching with a hardcoded space, e.g.:

```java
if (event.startsWith("data:")) {
    String data = event.substring(5).stripLeading();
    ...
} else if (event.startsWith("event:")) {
    String eventType = event.substring(6).stripLeading();
    ...
}
```

### Related (same streaming path, minor)

1. `AIHttpUtils.resolve(baseUrl, "models")` uses `URI.resolve`, so a Base URL without a trailing slash (`.../compatible-mode/v1`) resolves to `.../compatible-mode/models` (404). Normalizing the base URL (appending `/` when the path ends with a version segment, or documenting the requirement) would save a lot of debugging. The default OpenAI endpoint happens to end with `/v1/`, which hides this.
2. The legacy-chat fallback in `OpenAIClientResponses.mapHttpError` triggers only when the 400 body contains specific hardcoded strings (`"is not supported via Responses API"`, `"model_not_supported"`, ...). Providers that reject the Responses API with a different error wording (DashScope returns `<400> InternalError.Algo.InvalidParameter ...` for `function_call_output` input items) get no fallback and just error out. A configurable "use Chat Completions API instead of Responses API" switch would make the OpenAI engine work with any compatible provider.

### Environment

- DBeaver 26.2.0.202608301738 CE, macOS aarch64
- Provider: Alibaba Cloud DashScope (OpenAI-compatible mode), model qwen-plus
- Reproduced with plain curl: the SSE stream contains `data:{...}` (no space) while `OpenAiAPIStreamConsumer` requires `data: ` (with space)

Thanks!

Ashfaqbs:
## Issue

Fixes #42096

## Root cause

`OpenAiAPIStreamConsumer` matched the OpenAI Responses API SSE stream lines against `"data: "` and `"event: "` — colon **plus** a mandatory space. Per the [SSE spec](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation), the single space after the field-name colon is optional. Several OpenAI-compatible providers (Alibaba Cloud DashScope confirmed in the issue, verified byte-for-byte with `curl | od -c`) emit `data:{json}` with no space, which matches neither prefix and falls into the `else` branch — logged only at debug level, so every chunk is silently dropped. From the user's side, AI Chat just hangs at 0 tokens with no visible error.

## Change

- `OpenAiAPIStreamConsumer.java`: drop the trailing space from the `DATA_EVENT`/`EVENT_EVENT` constants (`"data: "` → `"data:"`, `"event: "` → `"event:"`). The existing `.substring(...).trim()` calls on the extracted payload already handle the optional space correctly either way, so this is the whole fix.

## Test

Added two cases to the existing `OpenAiAPIStreamConsumerTest`:
- `ignoresDoneMarkerWithoutSpaceAfterColon` — `data:[DONE]` (no space) is still recognized as the done marker.
- `preservesTextFromDataEventWithoutSpaceAfterColon` — a real delta chunk with no space after the colon is still parsed and delivered to the listener.

I could not run the module's Tycho/Eclipse test suite in this environment (it needs the full target-platform resolution), so I verified the exact fixed logic standalone against the JDK directly — same prefix constants, same `startsWith`/`substring`/`trim` sequence, against `data: [DONE]`, `data:[DONE]`, `data:   [DONE]  `, `event: response.created`, and `event:response.created` — all resolve as expected. Happy to have the added test cases re-verified in CI.

## Checklist

- [x] No breaking changes — spaced prefixes (the common case) are unaffected; this only widens what's accepted
- [x] Added unit test coverage for the fix
- [ ] Could not run the full module test suite locally (Tycho target platform not available in this environment) — verified the isolated logic manually instead (see above)


E1izabeth:

E1izabeth:
@codex review

E1izabeth (approved):

Issue #36951, opened 2025-01-29
[New feature] Local LLM Support for DBeaver Community


morandalex:
**Is your feature request related to a problem? Please describe.**

While DBeaver Enterprise Edition already includes AI capabilities, the community version lacks integration with local LLMs like Ollama. This limits the open-source community's ability to leverage AI features for database operations, especially for users who need to work with sensitive data locally or want to contribute to AI feature development.

**Describe the solution you'd like**

I propose adding support for local LLM integration in DBeaver Community Edition, specifically:
1. Integration with Ollama as the primary local LLM provider
2. An extensible API that allows the community to plug in different LLM providers
3. Basic AI features in the SQL editor such as:
   - Query generation assistance
   - SQL explanation
   - Query optimization suggestions
   - Schema analysis
   - create schemas quickly

**Describe alternatives you've considered**

1. Using DBeaver Enterprise Edition - but this limits community contributions and accessibility
2. External AI tools/plugins - but this breaks the workflow and doesn't provide tight integration
3. Cloud-based LLM services - but these don't address privacy concerns and offline usage needs

**Additional context**

- Ollama has become a popular choice for running local LLMs, supporting models like Llama 3.3 and others
- There's already community interest in this feature
- This would allow DBeaver to stay competitive while embracing the open-source community's contributions
- The implementation could start simple with Ollama and gradually expand based on community needs
- This would bridge the gap between enterprise and community editions while maintaining the commercial value of the enterprise version through advanced features

There is a great value in OSS community!

HocKu7:
@morandalex Thank you for your report.
So, for someone who is interested, please vote for the issue with 👍  under the issue description to promote it to a higher priority.

morandalex:
Hello @HocKu7  a lot of people voted for this issue! It can be game changer to allow people to setup their ollama/openai/anthropic/gemini  environment.

It can be a setting where you specify 

- the system prompt 
- the domain
- and the model name


Think to the fact that this can drive Dbeaver adoption for newbye in sql that are approaching for the first time programming .

Thank you

abeatbeyondlab:
I found  a guide  here : https://dbeaver.com/docs/dbeaver/AI-Smart-Assistance/

on ce edition is possible to setup Openai and Copilot 
<img width="428" height="272" alt="Image" src="https://github.com/user-attachments/assets/170a29b8-46cd-4ef0-a8ce-868c09635089" />

[...]

PR #42157, opened 2026-09-18
fix(ai): resolve relative paths correctly against a base URL with no trailing slash

### LINKED ISSUE #42156: OpenAI engine: self-hosted endpoint requests 404 when base URL has no trailing slash
### Description

Requests to a self-hosted OpenAI-compatible endpoint (Ollama, vLLM, LM Studio, ...) 404 when the configured base URL has no trailing slash — e.g. `http://localhost:11434/v1` instead of `http://localhost:11434/v1/`.

### Root cause

`AIHttpUtils.resolve(base, paths...)` builds the request URI with plain `URI.resolve(path)`. Per RFC 3986, relative-reference resolution treats the segment after the last `/` in the base as a file name, not a directory — resolving a relative path against a base without a trailing slash replaces that last segment instead of appending to it:

```java
new URI("http://localhost:11434/v1").resolve("models")
// -> http://localhost:11434/models  (drops "/v1")
new URI("http://localhost:11434/v1/").resolve("models")
// -> http://localhost:11434/v1/models  (correct)
```

Both `OpenAiClientBase.getModels` (`AIHttpUtils.resolve(baseUrl, "models")`) and `OpenAIClientChat` (`AIHttpUtils.resolve(baseUrl, "chat/completions")`) are affected, so a base URL without a trailing slash breaks both the model list and the chat request itself.

### Where this bites

This is the likely cause of the "connection test succeeds, but chat returns a 404" reports on #36951 — Ollama's OpenAI-compatible endpoint is conventionally documented and configured as `http://localhost:11434/v1` (no trailing slash), so pointing DBeaver's OpenAI engine at it hits this exact case.

### Reproduction

1. Preferences → AI → OpenAI engine
2. Base URL: `http://localhost:11434/v1` (no trailing slash) — or any self-hosted OpenAI-compatible server configured the same way
3. Refresh models / send a chat message
4. Request goes to `.../models` or `.../chat/completions` instead of `.../v1/models` / `.../v1/chat/completions` → 404

### Workaround

Add the trailing slash manually (`http://localhost:11434/v1/`) when configuring the base URL.

### Suggested fix

`AIHttpUtils.resolve` should normalize the base to end with `/` before resolving a relative path against it, whenever there's at least one path segment to append (a caller passing an already-complete URL as `base` with no extra `paths` — several `CopilotClientBase` call sites do this — needs to keep using it exactly as given, so the normalization should only apply when `paths.length > 0`).

I have this fix ready with test coverage — opening a PR shortly.

Ashfaqbs:
## Issue

Fixes #42156

## Root cause

`AIHttpUtils.resolve(base, paths...)` built the request URI with plain `URI.resolve(path)`. Per RFC 3986, relative-reference resolution treats the segment after the last `/` in the base as a file name, not a directory — resolving a relative path against a base without a trailing slash replaces that last segment instead of appending to it:

```java
new URI("http://localhost:11434/v1").resolve("models")
// -> http://localhost:11434/models   (drops "/v1")
new URI("http://localhost:11434/v1/").resolve("models")
// -> http://localhost:11434/v1/models   (correct)
```

`OpenAiClientBase.getModels` (`resolve(baseUrl, "models")`) and `OpenAIClientChat` (`resolve(baseUrl, "chat/completions")`) both go through this, so a base URL without a trailing slash breaks model listing and chat requests alike. Self-hosted OpenAI-compatible endpoints (Ollama, vLLM, LM Studio) are commonly configured/documented without the trailing slash (`http://localhost:11434/v1`), so this is a likely cause of the "connection test passes, chat returns 404" reports on the local-LLM feature request thread (#36951) — worth flagging since that issue has a lot of community demand riding on OpenAI-compatible endpoints working smoothly.

## Change

- `AIHttpUtils.resolve`: normalize `base` to end with `/` before resolving, but **only** when there's at least one relative path to resolve against it (`paths.length > 0`). Several `CopilotClientBase` call sites pass an already-complete URL as `base` with zero extra `paths` and rely on it being used exactly as given — those are untouched by this change.

## Test

New `AIHttpUtilsTest` covering:
- base without trailing slash + relative path → correctly appended, not replaced
- base with trailing slash + relative path → unchanged behavior
- the chat/completions case specifically (the one that 404s for Ollama-style configs)
- base with no path segment at all (`http://localhost:11434`)
- a complete URL passed with no extra paths → returned exactly as given (the Copilot device-code/login-URL case)

Couldn't run the module's Tycho/Eclipse test suite locally (needs the full target-platform resolution), so I also verified the exact fixed logic standalone against the JDK directly (same `resolve` implementation, same 5 cases) — all passed. Happy to have the added test re-verified in CI.

## Checklist

- [x] No breaking changes — a base URL that already ends with `/` is unaffected; a complete URL passed with no extra `paths` is unaffected
- [x] Added unit test coverage for the fix
- [ ] Could not run the full module test suite locally (Tycho target platform not available in this environment) — verified the isolated logic manually instead (see above)


E1izabeth:
@codex review

E1izabeth:

Ashfaqbs:
Good catch — fixed in e021eaca75. `getModels()` now resolves against "/v1/" explicitly instead of `RESPONSES_PATH`, so it correctly lands on `/v1/models` regardless of the base-URL normalization change.

